PR #3, opened 2017-10-06
Fix issue when setting instance group named ports

nicksardo:
Found an issue regarding the recent change to setting named ports. The wrong list was used when being appended to the instance group's existing list of ports. If you run the added test case, the unit test shows that ports get duplicated.


nicksardo:
cc @G-Harmon

nikhiljindal:
Thanks for catching the issue.
/lgtm

PR #10, opened 2017-10-09
Fix the glbc build by removing 'godeps' from command.

G-Harmon:
We no longer should invoke 'godeps' after the recent shuffling of repos (ingress -> ingress-gce/ingress-nginx split) and Makefiles.


nicksardo:
/lgtm
Thanks!

PR #15, opened 2017-10-11
Update build

bowei:
Look at the commits for review. The Makefile and rules.mk are from kube-dns and support a clean build environment with multi architectures.

The standard k8s projects have been following is to put library code in pkg/


Makefile
```diff
@@ -1,18 +1,44 @@
-all: push
+# Copyright 2016 The Kubernetes Authors.
```
G-Harmon: update to 2017?

bowei:
@G-Harmon you have the power...to make pull requests :-)

PR #57, opened 2017-10-25
Updating loadbalancer/fakes.go to return 404

nikhiljindal:
Same as https://github.com/kubernetes/ingress-gce/pull/56.
Missed `pkg/loadbalancers/fakes.go` in that PR.


nicksardo:
Love it, thanks!
/lgtm

PR #59, opened 2017-10-27
Extracting tlsloader into a separate package to enable reuse

nikhiljindal:
TlsLoader is the interface used to load TLSCerts for an ingress.

cc @bowei @nicksardo @csbell @G-Harmon @madhusudancs

### DIFF
--- a/pkg/tls/tls.go
+++ b/pkg/tls/tls.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package controller
+package tls
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


pkg/tls/tls.go
```diff
@@ -29,26 +29,31 @@ import (
...
 
-func (n *noOPValidator) validate(certs *loadbalancers.TLSCerts) error {
+func (n *noOPValidator) Validate(certs *loadbalancers.TLSCerts) error {
 	return nil
 }
 
-// apiServerTLSLoader loads TLS certs from the apiserver.
-type apiServerTLSLoader struct {
+// ApiServerTLSLoader loads TLS certs from the apiserver.
+type ApiServerTLSLoader struct {
```
nicksardo: Should we consider a better name for this while we're here? Thoughts? @bowei @freehan

nikhiljindal: Sure. Do you have some name in mind?
How about TLSCertsFromSecretLoader or SecretsTLSCertsLoader?

nicksardo: Something along those lines. Either your suggestion with a fixed pluralization: `TLSCertsFromSecretsLoader` or `SecretsToTLSCertsLoader`

nikhiljindal: Great, renaming to TLSCertsFromSecretsLoader

nikhiljindal: Done

nikhiljindal:
Pushed another commit to rename as per the comments. PTAL

PR #63, opened 2017-10-30
NEG: skip endpoint updates with no subset change

freehan:


bowei:
should this sometimes go through, just to make sure the cloud state is up to date?

freehan:
That is a good point. Syncer currently relying on the informer resync period to reassert NEG, just like ingress.

PR #70, opened 2017-11-04
Centralize more of the naming of GCE resources

bowei:
- Moves Namer out to its own source file.
- Improve some of the logging in the unit test.
- Lots of minor naming consistencies (look at the commits)


pkg/utils/namer.go
```diff
@@ -28,6 +28,15 @@ import (
 )
 
 const (
+	// A single target proxy/urlmap/forwarding rule is created per loadbalancer.
+	// Tagged with the namespace/name of the Ingress.
+	targetHTTPProxyPrefix     = "k8s-tp"
```
nikhiljindal: How about further extracting "k8s" prefix as a const (similar to what negPrefix() does)?
That way resource names can be reused by kubemci by replacing the prefix "k8s" with "mci1".

pkg/utils/namer.go
```diff
@@ -90,10 +106,11 @@ type NameComponents struct {
 	ClusterName, Resource, Metadata string
 }
 
-// SetClusterName sets the UID/name of this cluster.
-func (n *Namer) SetClusterName(name string) {
+// SetCluster sets the UID/name of this cluster.
+func (n *Namer) SetCluster(name string) {
```
nikhiljindal: This doesnt have to be a cluster name. It just needs to be a unique value. For ex: kubemci uses "load balancer" name instead of cluster name. How about keeping the name generic here?

pkg/utils/namer.go
```diff
@@ -210,60 +229,109 @@ func (n *Namer) BePort(beName string) (string, error) {
...
+
+// IsNEG returns true if the name is a NEG owned by this cluster.
+func (n *Namer) IsNEG(name string) bool {
+	return strings.HasPrefix(name, n.negPrefix())
 }
 
-func (n *Namer) NEGPrefix() string {
-	return fmt.Sprintf("k8s%s-%s", schemaVersionV1, n.GetClusterName())
+func (n *Namer) negPrefix() string {
+	return fmt.Sprintf("k8s%s-%s", schemaVersionV1, n.Cluster())
```
nikhiljindal: Should the k8s here be configurable?
Namer should probably take a prefix "k8s" and use that for all resources (similar to comment above about using k8s prefix for resource names).

nikhiljindal:
Thanks @bowei!
The changes look great. I have a few suggestions that will make it easier to share the namer with kubemci.
Would love to know if you think that is a viable goal.

pkg/utils/namer.go
bowei: sgtm -- on another PR though?

pkg/utils/namer.go
bowei: SetUID?

pkg/utils/namer.go
nikhiljindal: yes doing in another PR sounds good. You can optionally add a TODO to track

pkg/utils/namer.go
nikhiljindal: yes sg

nicksardo:
I'm so glad I don't have to look at FirewallName and FirewallSuffix anymore. 

/lgtm

pkg/utils/namer.go
bowei: done

PR #74, opened 2017-11-07
Add ability to change prefix in the Namer

bowei:
- This allows the multicluster code to share use of the Namer.
- Add more unit testing
- Coverage is now ~90% of namer code.

### DIFF
--- a/pkg/controller/cluster_manager.go
+++ b/pkg/controller/cluster_manager.go
@@ -36,16 +36,6 @@ const (
 	defaultPort            = 80
 	defaultHealthCheckPath = "/"
 
-	// A backend is created per nodePort, tagged with the nodeport.
-	// This allows sharing of backends across loadbalancers.
-	backendPrefix = "k8s-be"
-
-	// A single target proxy/urlmap/forwarding rule is created per loadbalancer.
-	// Tagged with the namespace/name of the Ingress.
-	targetProxyPrefix    = "k8s-tp"
-	forwardingRulePrefix = "k8s-fw"
-	urlMapPrefix         = "k8s-um"
-
 	// Used in the test RunServer method to denote a delete request.
 	deleteType = "del"
 
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -48,9 +48,8 @@ var (
 	testIPManager = testIP{}
 )
 
-// TODO: Use utils.Namer instead of this function.
 func defaultBackendName(clusterName string) string {
-	return fmt.Sprintf("%v-%v", backendPrefix, clusterName)
+	return fmt.Sprintf("%v-%v", "k8s-be", clusterName)
 }
 
 // newLoadBalancerController create a loadbalancer controller.
@@ -142,7 +141,7 @@ type nodePortManager struct {
 	portMap map[string]int
 	start   int
 	end     int
-	namer   utils.Namer
+	namer   *utils.Namer
 }
 
 // randPort generated pseudo random port numbers.
@@ -168,8 +167,8 @@ func (p *nodePortManager) toNodePortSvcNames(inputMap map[string]utils.FakeIngre
 	return expectedMap
 }
 
-func newPortManager(st, end int) *nodePortManager {
-	return &nodePortManager{map[string]int{}, st, end, utils.Namer{}}
+func newPortManager(st, end int, namer *utils.Namer) *nodePortManager {
+	return &nodePortManager{map[string]int{}, st, end, namer}
 }
 
 // addIngress adds an ingress to the loadbalancer controllers ingress store. If
@@ -222,7 +221,7 @@ func TestLbCreateDelete(t *testing.T) {
 			"/bar": "bar1svc",
 		},
 	}
-	pm := newPortManager(1, 65536)
+	pm := newPortManager(1, 65536, cm.Namer)
 	ings := []*extensions.Ingress{}
 	for _, m := range []map[string]utils.FakeIngressRuleValueMap{inputMap1, inputMap2} {
 		newIng := newIngress(m)
@@ -231,10 +230,10 @@ func TestLbCreateDelete(t *testing.T) {
 		lbc.sync(ingStoreKey)
 		l7, err := cm.l7Pool.Get(ingStoreKey)
 		if err != nil {
-			t.Fatalf("%v", err)
+			t.Fatalf("cm.l7Pool.Get(%q) = _, %v; want nil", ingStoreKey, err)
 		}
 		if err := cm.fakeLbs.CheckURLMap(l7, pm.toNodePortSvcNames(m)); err != nil {
-			t.Fatalf("%v", err)
+			t.Fatalf("cm.fakeLbs.CheckURLMap(l7, pm.toNodePortSvcNames(m)) = %v, want nil", err)
 		}
 		ings = append(ings, newIng)
 	}
@@ -311,7 +310,7 @@ func TestLbFaultyUpdate(t *testing.T) {
 		},
 	}
 	ing := newIngress(inputMap)
-	pm := newPortManager(1, 65536)
+	pm := newPortManager(1, 65536, cm.Namer)
 	addIngress(lbc, ing, pm)
 
 	ingStoreKey := getKey(ing, t)
@@ -321,7 +320,7 @@ func TestLbFaultyUpdate(t *testing.T) {
 		t.Fatalf("%v", err)
 	}
 	if err := cm.fakeLbs.CheckURLMap(l7, pm.toNodePortSvcNames(inputMap)); err != nil {
-		t.Fatalf("%v", err)
+		t.Fatalf("cm.fakeLbs.CheckURLMap(...) = %v, want nil", err)
 	}
 
 	// Change the urlmap directly through the lb pool, resync, and
@@ -334,7 +333,7 @@ func TestLbFaultyUpdate(t *testing.T) {
 
 	lbc.sync(ingStoreKey)
 	if err := cm.fakeLbs.CheckURLMap(l7, pm.toNodePortSvcNames(inputMap)); err != nil {
-		t.Fatalf("%v", err)
+		t.Fatalf("cm.fakeLbs.CheckURLMap(...) = %v, want nil", err)
 	}
 }
 
@@ -343,7 +342,7 @@ func TestLbDefaulting(t *testing.T) {
 	lbc := newLoadBalancerController(t, cm)
 	// Make sure the controller plugs in the default values accepted by GCE.
 	ing := newIngress(map[string]utils.FakeIngressRuleValueMap{"": {"": "foo1svc"}})
-	pm := newPortManager(1, 65536)
+	pm := newPortManager(1, 65536, cm.Namer)
 	addIngress(lbc, ing, pm)
 
 	ingStoreKey := getKey(ing, t)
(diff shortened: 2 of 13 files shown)


pkg/utils/namer.go
```diff
@@ -28,22 +28,24 @@ import (
...
+
 	// A single target proxy/urlmap/forwarding rule is created per loadbalancer.
 	// Tagged with the namespace/name of the Ingress.
-	targetHTTPProxyPrefix  = "k8s-tp"
-	targetHTTPSProxyPrefix = "k8s-tps"
-	sslCertPrefix          = "k8s-ssl"
+	targetHTTPProxyPrefix  = "tp"
+	targetHTTPSProxyPrefix = "tps"
+	sslCertPrefix          = "ssl"
 	// TODO: this should really be "fr" and "frs".
```
nikhiljindal: Doesnt firewall rule use "fr"?

nikhiljindal:
This is great! thx, lgtm

bowei:
fixing the test failure related to instance groups

bowei:
The second patch is a little large but it's basically making all of the unit tests and fakes use the Namer consistently.

pkg/controller/cluster_manager.go
```diff
@@ -36,16 +36,6 @@ const (
 	defaultPort            = 80
 	defaultHealthCheckPath = "/"
 
-	// A backend is created per nodePort, tagged with the nodeport.
```
nikhiljindal: Love this change!

pkg/controller/controller_test.go
```diff
@@ -50,7 +50,7 @@ var (
 
 // TODO: Use utils.Namer instead of this function.
 func defaultBackendName(clusterName string) string {
-	return fmt.Sprintf("%v-%v", backendPrefix, clusterName)
+	return fmt.Sprintf("%v-%v", "k8s-be", clusterName) // backendPrefix, clusterName) XXX
```
nikhiljindal: XXX is intentional? :)

nikhiljindal: Also dont like "k8s-be" being hardcoded here.
Remove this method altogether to use Namer?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -54,7 +54,7 @@ func TestHealthCheckAdd(t *testing.T) {
 }
 
 func TestHealthCheckAddExisting(t *testing.T) {
-	namer := &utils.Namer{}
+	namer := utils.NewNamer("uid1", "fw1")
```
nikhiljindal: Maybe define a utils.NewTestNamer() rather than duplicating "uid1", "fw1" it in all these tests?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -202,7 +202,7 @@ func TestHealthCheckDeleteLegacy(t *testing.T) {
 }
 
 func TestAlphaHealthCheck(t *testing.T) {
-	namer := &utils.Namer{}
+	namer := utils.NewNamer("uid1", "fw1")
```
nikhiljindal: Do these really need a separate instance?

pkg/loadbalancers/fakes.go
```diff
@@ -18,6 +18,8 @@ package loadbalancers
 
 import (
 	"fmt"
+	"github.com/golang/glog"
+	"testing"
```
nikhiljindal: No!!!!
This is reverting https://github.com/kubernetes/ingress-gce/pull/69. Please fix

pkg/healthchecks/healthchecks_test.go
nikhiljindal: You can define it one of the fakes.go?

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -89,13 +91,14 @@ func TestCreateHTTPLoadBalancer(t *testing.T) {
 func TestCreateHTTPSLoadBalancer(t *testing.T) {
 	// This should NOT create the forwarding rule and target proxy
 	// associated with the HTTP branch of this loadbalancer.
+	namer := utils.NewNamer("uid1", "fw1")
 	lbInfo := &L7RuntimeInfo{
-		Name:      "test",
```
nikhiljindal: Do things break without this change?
Seems like you have fixed all the instances. We can keep things simple where a namer generated name is not required?

pkg/controller/controller_test.go
bowei: yes, this needs to be fixed

pkg/loadbalancers/fakes.go
bowei: Can we move the method to the test or a different package?

pkg/loadbalancers/loadbalancers_test.go
bowei: Things do break and the inconsistent use of names is quite problematic.

pkg/loadbalancers/fakes.go
nikhiljindal: Moving the method to test is fine by me. As long as non test code does not depend on testing.

bowei: Quite funny I think I exactly reverted the change.

bowei:
@nikhiljindal -- took care of most of the comments. Some of the test clean I'm going to put in a follow on, the change is getting somewhat large.

pkg/controller/controller_test.go
```diff
@@ -48,9 +48,8 @@ var (
 	testIPManager = testIP{}
 )
 
-// TODO: Use utils.Namer instead of this function.
```
nikhiljindal: Why remove the TODO? It still seems relevant.

pkg/controller/controller_test.go
```diff
@@ -48,9 +48,8 @@ var (
...
 func defaultBackendName(clusterName string) string {
-	return fmt.Sprintf("%v-%v", backendPrefix, clusterName)
+	return fmt.Sprintf("%v-%v", "k8s-be", clusterName)
```
nikhiljindal: (repeat of previous comment). Do not hardcode this here and reuse namer method?

pkg/healthchecks/healthchecks_test.go
nikhiljindal: I see instances_test.go uses a single instance in all test cases. This uses separate. Lets be consistent.
Instantiate once and reuse it in all tests

pkg/healthchecks/healthchecks_test.go
nikhiljindal: ping, what do you think of instantiating once and reusing in all?

pkg/loadbalancers/fake/fakes.go
```diff
@@ -14,14 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package loadbalancers
+// Package fake contains fakes for loadbalancer.
+package fake
```
nikhiljindal: Keep package name and file name consistent. Call them both fakes?

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package loadbalancers
+package loadbalancers_test
```
nikhiljindal: Sorry not sure why is this a separate package? Unit tests are supposed to be in the same package.

If you are really moving to a separate package, then also move this to a different directory with the package name.

nikhiljindal:
cc @G-Harmon @csbell

pkg/controller/controller_test.go
bowei: It isn't actually a Namer resource (it is not  a GCP resource). The comment is incorrect.

pkg/controller/controller_test.go
bowei: See comment above.

pkg/loadbalancers/loadbalancers_test.go
bowei: I don't think there is actually a hard and fast rule about keeping the tests in the same package. (test via the front door). I hesitate to create a package just for the test.

bowei: Actually, let me see why I did that originally. Maybe it was a leftover from a different attempt to clean up

bowei: Ok, the problem is this: 

* fakes imports loadbalancers
* the test depends on fakes.

import cycle.

pkg/loadbalancers/fake/fakes.go
bowei: done

pkg/loadbalancers/loadbalancers_test.go
nikhiljindal: Agreed that there is not a hard and fast rule, but it is preferred so that you dont have to expose private members outside the package, just for unit tests.

The problem here is that fakes imports testing, which is not usual.
All of this seems like a hack just to be able to pass t to CheckURLMap and I am not convinced why do you want to do that. Why not let CheckURLMap return an error and handle it appropriately in tests (as it does now). You wont need separate fake and loadbalancer_tests packages then (we can still keep the fake package if we want, but it wont necessarily be required). Reverting that change will make this PR a lot simpler and focused only on the namer change.
What do you think?

pkg/controller/controller_test.go
nikhiljindal: ok got it, thx!

pkg/controller/controller_test.go
nikhiljindal: ok, resolved.

pkg/loadbalancers/fakes.go
```diff
@@ -38,46 +40,46 @@ func (t *testIP) ip() string {
 
 // Loadbalancer fakes
 
-// FakeLoadBalancers is a type that fakes out the loadbalancer interface.
+// LoadBalancers is a type that fakes out the loadbalancer interface.
```
nikhiljindal: Comment mismatches with code

pkg/loadbalancers/fakes.go
```diff
@@ -38,46 +40,46 @@ func (t *testIP) ip() string {
...
 type FakeLoadBalancers struct {
 	Fw    []*compute.ForwardingRule
 	Um    []*compute.UrlMap
 	Tp    []*compute.TargetHttpProxy
 	Tps   []*compute.TargetHttpsProxy
 	IP    []*compute.Address
 	Certs []*compute.SslCertificate
-	name  string
-	calls []string // list of calls that were made
+	Name  string
```
nikhiljindal: All these changes in this file seem like leftover from when the tests were not in same package and hence you had to export them. You dont need to export them anymore?

bowei: I'll change it back, but there shouldn't really be any issue with exporting stuff out of the fake...

pkg/loadbalancers/fakes.go
bowei: fixed

pkg/loadbalancers/fakes.go
bowei: fixed

nikhiljindal:
Thanks for all the fixes!
/lgtm

nikhiljindal:
unit test failures are legit:
```
pkg/loadbalancers/loadbalancers_test.go:415:24: f.Calls undefined (type *FakeLoadBalancers has no field or method Calls, but does have calls)
pkg/loadbalancers/loadbalancers_test.go:470:3: f.Name undefined (type *FakeLoadBalancers has no field or method Name, but does have name)
```

bowei:
hmm -- strange, thought I fixed all of the call sites, fixing

bowei:
should be fixed now

PR #78, opened 2017-12-05
Add ListGlobalForwardingRules to the LoadBalancers interface.

G-Harmon:
I have an upcoming commit in the k8s-multicluster-ingress repo where I will use this.

cc @bowei @nicksardo @nikhiljindal

### DIFF
--- a/pkg/loadbalancers/fakes.go
+++ b/pkg/loadbalancers/fakes.go
@@ -116,6 +116,12 @@ func (f *FakeLoadBalancers) GetGlobalForwardingRule(name string) (*compute.Forwa
 	return nil, utils.FakeGoogleAPINotFoundErr()
 }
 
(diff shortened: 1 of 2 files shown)


nikhiljindal:
I see ListGlobalForwardingRules is implemented in GCECloud already: https://github.com/kubernetes/kubernetes/blob/6d32783453a6f624e9c0e2d818f58bdaa65199b3/pkg/cloudprovider/providers/gce/gce_forwardingrule.go#L73, so just adding it to the interface seems enough.

lgtm, but will wait for @bowei or @nicksardo to take a look and lgtm and merge.

G-Harmon:
I updated the Fake so that it actually returns the current firewall rules. (I ended up using this logic in my commit for kubemci, which I haven't sent out yet.)

nikhiljindal:
thx, still lgtm

PR #79, opened 2017-12-05
WIP: Adds pkg/cache, a type-safe way to store GCP objects

bowei:
This commit adds the data structure, follow on commits will use the data
structure to implement both a cache of the cloud state as well as use it
to implement the synchronization routines. See the package docstring in
cache.go for more details.

The cache code is type-safe and generated from descriptions contains in
the pkg/cache/meta. The contents of `gen.go` are derived from `$ go run
gen/main.go > gen.go`

As the code is not expected to change often (only when types are
added/removed), the generated code is committed to the repo and the code
generation is expected to be run by hand. If there are significant
divergences from these assumptions, then full integration into the build
process will be needed.


bowei:
@nikhiljindal @nicksardo

nikhiljindal:
cc @csbell @G-Harmon

bowei:
@G-Harmon -- yes I realize this PR is kind of dropped without much explication.

The gist is that we need to be able to easily store a set of GCE objects to do a couple of things:
- cache GCP state, this both makes writing the controller logic easier and cleaner and also has performance improvements
- data structure for the result of a translation of ingress spec => GCE objects. This is sort of done already in loadbalancers.L7, but this struct is way more generic

Now the structure of the code:
- the explanation for usage is in the package comment in cache.go
- most of the code is autogenerated from gen/main.go -> gen.go, gen_test.go
- the backing store is in store.go

mosaa1212:
https://www.googleapis.com/auth/userinfo.profile
الدولة = e086dfcb289b5140e1b739add65c9903

PR #80, opened 2017-12-07
Update websocket example

### LINKED ISSUE #77: examples/websocket/server.go: concurrent write to socket
In the `handleWSConn` function, the goroutine calls `c.WriteMessage` and the other for loop also calls `c.WriteMessage`. But according to Gorilla WS doc, `WriteMessage` is not goroutine safe. We need to use a mutex to protect concurrent writes.

I saw a couple of goroutine panics because of this while doing some light stress tests.

nicksardo:
Changes:
- Fixes occasional panics due to concurrent writing
- Adds html page for connecting to the websocket
- Adds copyright header

Fixes #77

### DIFF
--- a/examples/websocket/README.md
+++ b/examples/websocket/README.md
@@ -9,25 +9,14 @@ Any websocket server will suffice; however, for the purpose of demonstration, we
 
 ### Containerize
 ```shell
-➜ docker build -t nicksardo/websocketexample .
-Sending build context to Docker daemon 6.134 MB
-Step 1 : FROM alpine:3.5
- ---> 4a415e366388
-Step 2 : COPY wsserver /wsserver
- ---> 8002887d752d
-Removing intermediate container 7772a3e76155
-Step 3 : CMD /wsserver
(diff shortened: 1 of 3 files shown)


examples/websocket/README.md
```diff
@@ -9,25 +9,14 @@ Any websocket server will suffice; however, for the purpose of demonstration, we
...
- ---> 8002887d752d
-Removing intermediate container 7772a3e76155
-Step 3 : CMD /wsserver
- ---> Running in 27c8ff226267
- ---> eecd0574e5d1
-Removing intermediate container 27c8ff226267
-Successfully built eecd0574e5d1
-
-➜ docker push nicksardo/websocketexample:latest
+➜ docker build -t nicksardo/websocketexample:v1 .
```
bowei: do we need your docker in the example?

bowei:
lgtm

examples/websocket/README.md
nicksardo: Had it for convenience. Removed.

PR #82, opened 2017-12-09
Add build, coverage and report badges

MrHohn:
As what we had for [kubernetes/dns](https://github.com/kubernetes/dns). The coverage result should be available at the next build.


MrHohn:
/hold

Coverage page needs more works.

PR #88, opened 2017-12-22
Add e2e make rule

rramkumar1:
Adds a new rule to the Makefile which pushes an image of glbc for e2e testing. This target basically does the same thing as a regular build and push but we need it for two reasons:

1. Modify the CONTAINER_PREFIX so that it is clear that image is for e2e testing.
2. In test-infra/kubetest, we need to build an image of glbc for testing and having a target called e2e so that we kubetest can just say "make e2e" makes it clearer what we are doing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,7 @@ VERBOSE ?= 0
 
 # Include standard build rules.
 include build/rules.mk
+
+# Additional rule to build an image of glbc for e2e testing.
+push-e2e:
+	@$(MAKE) --no-print-directory CONTAINER_PREFIX=ingress-gce-e2e containers push


freehan:
Can you squash the commits?

Makefile
```diff
@@ -42,3 +42,9 @@ VERBOSE ?= 0
 
 # Include standard build rules.
 include build/rules.mk
+
+# Additional rule to build an image of glbc for e2e testing.
+e2e:
```
bowei: push-e2e

rramkumar1: Done

Makefile
```diff
@@ -42,3 +42,7 @@ VERBOSE ?= 0
 
 # Include standard build rules.
 include build/rules.mk
+
+# Additional rule to build an image of glbc for e2e testing.
+push-e2e:
```
MrHohn: Probably put this into build/rules.mk?

rramkumar1: In rules.mk it says "These build rules should not need to be modified." I was under the impression that this meant new rules should not be added but I could be wrong.

PR #106, opened 2018-01-16
Sync one ingress

bowei:
A bunch of patches:
* A lot of clean up and movement
* Sync one ingress (as opposed to all) in Checkpoint
* Call GC explicitly when an Ingress is deleted
* Shutdown() firewall when there are no more Ingresses

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -36,14 +36,17 @@ type ControllerContext struct {
 
 // NewControllerContext returns a new shared set of informers.
 func NewControllerContext(kubeClient kubernetes.Interface, namespace string, resyncPeriod time.Duration, enableEndpointsInformer bool) *ControllerContext {
+	newIndexer := func() cache.Indexers {
+		return cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}
+	}
 	context := &ControllerContext{
-		IngressInformer: informerv1beta1.NewIngressInformer(kubeClient, namespace, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
-		ServiceInformer: informerv1.NewServiceInformer(kubeClient, namespace, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
-		PodInformer:     informerv1.NewPodInformer(kubeClient, namespace, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
-		NodeInformer:    informerv1.NewNodeInformer(kubeClient, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
+		IngressInformer: informerv1beta1.NewIngressInformer(kubeClient, namespace, resyncPeriod, newIndexer()),
+		ServiceInformer: informerv1.NewServiceInformer(kubeClient, namespace, resyncPeriod, newIndexer()),
+		PodInformer:     informerv1.NewPodInformer(kubeClient, namespace, resyncPeriod, newIndexer()),
+		NodeInformer:    informerv1.NewNodeInformer(kubeClient, resyncPeriod, newIndexer()),
 	}
 	if enableEndpointsInformer {
-		context.EndpointInformer = informerv1.NewEndpointsInformer(kubeClient, namespace, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+		context.EndpointInformer = informerv1.NewEndpointsInformer(kubeClient, namespace, resyncPeriod, newIndexer())
 	}
 	return context
 }
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -51,6 +51,26 @@ func NewControllerContext(kubeClient kubernetes.Interface, namespace string, res
 	return context
 }
 
+// HasSynced returns true if all relevant informers has been synced.
+func (ctx *ControllerContext) HasSynced() bool {
+
+	funcs := []func() bool{
+		ctx.IngressInformer.HasSynced,
+		ctx.ServiceInformer.HasSynced,
+		ctx.PodInformer.HasSynced,
+		ctx.NodeInformer.HasSynced,
+	}
+	if ctx.EndpointInformer != nil {
+		funcs = append(funcs, ctx.EndpointInformer.HasSynced)
+	}
+	for _, f := range funcs {
+		if !f() {
+			return false
+		}
+	}
+	return true
+}
+
 // Start all of the informers.
 func (ctx *ControllerContext) Start(stopCh chan struct{}) {
 	go ctx.IngressInformer.Run(stopCh)
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -19,19 +19,30 @@ package context
 import (
 	"time"
 
+	"github.com/golang/glog"
+
+	apiv1 "k8s.io/api/core/v1"
 	informerv1 "k8s.io/client-go/informers/core/v1"
 	informerv1beta1 "k8s.io/client-go/informers/extensions/v1beta1"
 	"k8s.io/client-go/kubernetes"
+	scheme "k8s.io/client-go/kubernetes/scheme"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 )
 
 // ControllerContext holds
 type ControllerContext struct {
+	kubeClient kubernetes.Interface
+
 	IngressInformer  cache.SharedIndexInformer
 	ServiceInformer  cache.SharedIndexInformer
 	PodInformer      cache.SharedIndexInformer
 	NodeInformer     cache.SharedIndexInformer
 	EndpointInformer cache.SharedIndexInformer
+
+	// Map of namespace => record.EventRecorder.
+	recorders map[string]record.EventRecorder
 }
 
 // NewControllerContext returns a new shared set of informers.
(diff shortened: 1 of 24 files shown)


bowei:
@G-Harmon

bowei:
Look at this change on a commit wise basis, otherwise will be hard to review.

pkg/controller/cluster_manager.go
```diff
@@ -172,15 +174,17 @@ func (c *ClusterManager) GC(lbNames []string, nodePorts []backends.ServicePort)
...
 		igName := c.ClusterNamer.InstanceGroup()
 		glog.Infof("Deleting instance group %v", igName)
-		igErr = c.instancePool.DeleteInstanceGroup(igName)
+		if err := c.instancePool.DeleteInstanceGroup(igName); err != err {
+			return err
+		}
 	}
-	if igErr != nil {
-		return igErr
+	if len(lbNames) == 0 {
```
freehan: merge it with the previous if statement?

Add a comment to say if there is no lbs. Remove instance group and firewall

pkg/controller/controller.go
```diff
@@ -92,26 +92,24 @@ type LoadBalancerController struct {
...
+	broadcaster := record.NewBroadcaster()
+	broadcaster.StartLogging(glog.Infof)
+	broadcaster.StartRecordingToSink(&unversionedcore.EventSinkImpl{
 		Interface: kubeClient.Core().Events(""),
 	})
 	lbc := LoadBalancerController{
 		client:              kubeClient,
+		ctx:                 ctx,
+		ingLister:           StoreToIngressLister{Store: ctx.IngressInformer.GetStore()},
+		nodeLister:          ctx.NodeInformer.GetIndexer(),
```
freehan: Probably makes sense to throw nodeLister and getReadyNodes into Translator.

pkg/context/context.go
```diff
@@ -19,35 +19,88 @@ package context
...
+	kubeClient kubernetes.Interface
+
 	IngressInformer  cache.SharedIndexInformer
 	ServiceInformer  cache.SharedIndexInformer
 	PodInformer      cache.SharedIndexInformer
 	NodeInformer     cache.SharedIndexInformer
 	EndpointInformer cache.SharedIndexInformer
+
+	// Map of namespace => record.EventRecorder.
+	recorders map[string]record.EventRecorder
```
freehan: EventRecorder per namespace?

pkg/controller/node.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/utils"
+)
+
+const (
+	nodeResyncPeriod = 1 * time.Second
```
nicksardo: Resync every second?

pkg/controller/utils.go
```diff
@@ -206,3 +208,19 @@ func uniq(nodePorts []backends.ServicePort) []backends.ServicePort {
 	}
 	return nodePorts
 }
+
+// getReadyNodeNames returns names of schedulable, ready nodes from the node lister.
+func getReadyNodeNames(lister listers.NodeLister) ([]string, error) {
+	nodeNames := []string{}
```
nicksardo: I know you didn't write this, but
Nit: var nodeNames []string

pkg/context/context.go
```diff
@@ -51,6 +51,26 @@ func NewControllerContext(kubeClient kubernetes.Interface, namespace string, res
 	return context
 }
 
+// HasSynced returns true if all relevant informers has been synced.
+func (ctx *ControllerContext) HasSynced() bool {
+
```
nicksardo: Extra space

pkg/controller/cluster_manager.go
```diff
@@ -105,6 +105,8 @@ func (c *ClusterManager) shutdown() error {
 // If in performing the checkpoint the cluster manager runs out of quota, a
 // googleapi 403 is returned.
 func (c *ClusterManager) Checkpoint(lbs []*loadbalancers.L7RuntimeInfo, nodeNames []string, backendServicePorts []backends.ServicePort, namedPorts []backends.ServicePort, firewallPorts []int64) ([]*compute.InstanceGroup, error) {
+	glog.V(4).Infof("Checkpoint %q, len(lbs)=%v, len(nodeNames)=%v, lne(backendServicePorts)=%v, len(namedPorts)=%v, len(firewallPorts)=%v", len(lbs), len(nodeNames), len(backendServicePorts), len(namedPorts), len(firewallPorts))
```
nicksardo: Provided only 5 values - expected 6.

Typo: "lne(backendServicePorts)"

pkg/context/context.go
nicksardo: I'm a bit confused by this too. The service controller doesn't specify the namespace:

https://github.com/kubernetes/kubernetes/blob/e35d67e032e6a064c0f867325ea2a48bca1ef9cf/pkg/controller/service/service_controller.go#L118

pkg/controller/translator/translator_test.go
```diff
@@ -0,0 +1,317 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
G-Harmon: s/2017/2018/

pkg/controller/node.go
```diff
@@ -0,0 +1,72 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
G-Harmon: 2018

pkg/controller/node.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+const (
+	nodeResyncPeriod = 1 * time.Second
+)
+
+// NodeController synchronizes the state of the nodes to the unmanaged instance
+// groups.
+type NodeController struct {
+	lister cache.Indexer
```
G-Harmon: newbie comment: is it worth commenting members? Is there style guidance on always commenting struct/class member variables? (In google c++, we say we should.)

pkg/controller/translator/translator.go
```diff
@@ -342,9 +342,10 @@ func (t *GCE) GatherFirewallPorts(svcPorts []backends.ServicePort, includeDefaul
 	if includeDefaultBackend {
 		svcPorts = append(svcPorts, *t.bi.DefaultBackendNodePort())
 	}
+
 	portMap := map[int64]bool{}
 	for _, p := range svcPorts {
-		if p.NEGEnabled {
+		if t.negEnabled && p.NEGEnabled {
```
G-Harmon: Is this a bug fix? If so, should it go in its own commit?

pkg/controller/cluster_manager.go
bowei: done

pkg/firewalls/firewalls.go
```diff
@@ -54,6 +54,7 @@ func NewFirewallPool(cloud Firewall, namer *utils.Namer) SingleFirewallPool {
 
 // Sync sync firewall rules with the cloud.
 func (fr *FirewallRules) Sync(nodePorts []int64, nodeNames []string) error {
+	glog.V(4).Infof("Sync(%v, %v)", nodePorts, nodeNames)
```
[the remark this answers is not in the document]
bowei: It should be clear from the filename in the log line

pkg/controller/node.go
bowei: ? what do you want me to do here? This is the existing policy.

pkg/controller/utils.go
bowei: done

pkg/context/context.go
bowei: Let's keep this for now, I'll remove in a later commit once we resolve the unit test log spam

pkg/context/context.go
bowei: done

pkg/controller/cluster_manager.go
bowei: done

pkg/controller/node.go
bowei: We typically don't update the year as it causes too much spam.

pkg/controller/node.go
bowei: good point, yes in general unless extremely obvious.

pkg/controller/node.go
```diff
@@ -0,0 +1,72 @@
...
+func NewNodeController(ctx *context.ControllerContext, cm *ClusterManager) *NodeController {
+	c := &NodeController{
+		lister: ctx.NodeInformer.GetIndexer(),
+		cm:     cm,
+	}
+	c.queue = utils.NewPeriodicTaskQueue("nodes", c.sync)
+
+	ctx.NodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.queue.Enqueue,
+		DeleteFunc: c.queue.Enqueue,
```
nicksardo: Can we add a TODO here that we need to improve this NodeController? 

We don't react on the UpdateFunc because it's supposedly very noisy. However, when new nodes are added, they are typically NotReady. This means in practice that instance groups are synced when the NodeInformer re-calls AddFunc (every 10 minutes).

pkg/controller/node.go
nicksardo: Could we rename to something else? Looks like this is just the frequency that the PeriodicTaskQueue checks the queue, not at which items in the queue are re-enqueued. Am I mistaken?

pkg/context/context.go
nicksardo: SGTM

pkg/controller/node.go
bowei: done

pkg/controller/node.go
bowei: I changed to nodeUpdatePeriod

pkg/controller/node.go
G-Harmon: This isn't an update. This is a new file. (I agree for changes to files the copyright isn't normally changed.)

PR #121, opened 2018-02-02
Introduce flag endpoint

nicksardo:
This endpoint currently prints the version and current verbosity level. You can also change the verbosity during runtime with a simple PATCH.  
```
curl -X PATCH http://localhost:8080/flag?v=10
```

More information can be added later.

### DIFF
--- a/cmd/glbc/app/handlers.go
+++ b/cmd/glbc/app/handlers.go
@@ -17,7 +17,9 @@ limitations under the License.
 package app
 
 import (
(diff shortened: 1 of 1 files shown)


bowei:
We should call this endpoint `/flag`.

bowei:
And have debug status show up on `/debug`

nicksardo:
Renamed to 'flag'. PTAL

PR #122, opened 2018-02-02
Sync entire nodeport range

nicksardo:
**WIP**

## Changes
- Firewall rule whitelists entire nodeport range instead of pin-holding nodeports used by ingresses. This is not a concern because the source IP ranges are only used by Google for health checkers and proxies.
- Begin enforcing target tags are up-to-date
- Use the `gce.LoadBalancerSourceRanges()` static func for specifying the GCE source ranges. This can be overwritten via flag. 

## Problems
The nodePort range is configurable via flag on the kubernetes API server. Is that list retrievable?


pkg/controller/cluster_manager.go
```diff
@@ -131,7 +131,7 @@ func (c *ClusterManager) Checkpoint(lbs []*loadbalancers.L7RuntimeInfo, nodeName
 		return igs, err
 	}
 
-	if err := c.firewallPool.Sync(firewallPorts, nodeNames); err != nil {
+	if err := c.firewallPool.Sync(nodeNames); err != nil {
```
bowei: Should we keep the existing behavior but as optional if the range is not set?

bowei: Actually on second thought, let's remove the old code -- it's a lot more complex.

pkg/firewalls/firewalls.go
```diff
@@ -190,3 +156,49 @@ type FirewallSyncError struct {
...
+func firewallAllowedEqual(allowListA []*compute.FirewallAllowed, allowListB []*compute.FirewallAllowed) bool {
+	if len(allowListA) != len(allowListB) {
+		return false
+	}
+
+	matchSet := sets.NewString()
+	for _, aAllow := range allowListA {
+		matchSet.Insert(firewallAllowedToString(aAllow))
+	}
+
```
bowei: why not use sets for both and call the set Equal?

pkg/firewalls/firewalls.go
```diff
@@ -190,3 +156,49 @@ type FirewallSyncError struct {
...
+		return false
+	}
+
+	// Ignore other firewall properties as the controller does not set them.
+	return true
+}
+
+func firewallAllowedToString(allowed *compute.FirewallAllowed) string {
+	sort.Sort(sort.StringSlice(allowed.Ports))
+	return strings.ToUpper(allowed.IPProtocol) + strings.Join(allowed.Ports, ",")
```
bowei: add a colon in the middle so if someone logs the result it looks ok to a human
`return strings.ToUpper(allowed.IPProtocol) + ":" + strings.Join(allowed.Ports, ",")`

pkg/firewalls/firewalls.go
nicksardo: Oops

pkg/controller/cluster_manager.go
```diff
@@ -226,6 +227,6 @@ func NewClusterManager(
 
 	// L7 pool creates targetHTTPProxy, ForwardingRules, UrlMaps, StaticIPs.
 	cluster.l7Pool = loadbalancers.NewLoadBalancerPool(cloud, defaultBackendPool, defaultBackendNodePort, cluster.ClusterNamer)
-	cluster.firewallPool = firewalls.NewFirewallPool(cloud, cluster.ClusterNamer)
+	cluster.firewallPool = firewalls.NewFirewallPool(cloud, cluster.ClusterNamer, gce.LoadBalancerSrcRanges(), nodePortRanges)
```
bowei: let's just access these as a global -- it is more like a constant than a parameter.

pkg/firewalls/fakes.go
```diff
@@ -120,3 +121,15 @@ func (ff *fakeFirewallsProvider) OnXPN() bool {
 func (ff *fakeFirewallsProvider) GetNodeTags(nodeNames []string) ([]string, error) {
 	return nodeNames, nil
 }
+
+func copyFirewall(f *compute.Firewall) *compute.Firewall {
+	enc, err := f.MarshalJSON()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to encode to json: %v", err))
```
bowei: return error instead of panic.

pkg/firewalls/firewalls.go
```diff
@@ -190,3 +160,35 @@ type FirewallSyncError struct {
 func (f *FirewallSyncError) Error() string {
 	return f.Message
 }
+
+func firewallsEqual(expected *compute.Firewall, existing *compute.Firewall) bool {
```
bowei: can just call this equal (firewalls.equal)

pkg/firewalls/firewalls.go
```diff
@@ -190,3 +160,35 @@ type FirewallSyncError struct {
...
+	if !sets.NewString(expected.SourceRanges...).Equal(sets.NewString(existing.SourceRanges...)) {
+		glog.V(5).Infof("Expected source ranges %v, actually %v", expected.SourceRanges, existing.SourceRanges)
+		return false
+	}
+
+	// Ignore other firewall properties as the controller does not set them.
+	return true
+}
+
+func firewallAllowedToString(allowed []*compute.FirewallAllowed) []string {
```
bowei: same, firewall prefix is not needed

pkg/flags/flags.go
```diff
@@ -92,11 +103,41 @@ the pod secrets for creating a Kubernetes client.`)
 		`Print the version of the controller and exit`)
 	flag.StringVar(&F.IngressClass, "ingress-class", "",
 		`If set, overrides what ingress classes are managed by the controller.`)
+	flag.Var(&F.NodePortRanges, "node-port-ranges", `Node port/port-ranges whitelisted for the
+L7 load balancing`)
```
bowei: might want to document the with an example

pkg/firewalls/fakes.go
```diff
@@ -55,7 +56,11 @@ func (ff *fakeFirewallsProvider) doCreateFirewall(f *compute.Firewall) error {
 	if _, exists := ff.fw[f.Name]; exists {
 		return fmt.Errorf("firewall rule %v already exists", f.Name)
 	}
-	ff.fw[f.Name] = f
+	cf, err := copyFirewall(f)
+	if err != nil {
+		return fmt.Errorf("Failed to copy firewall: %v", err)
```
bowei: why not propagate err directly?

pkg/firewalls/fakes.go
```diff
@@ -93,7 +98,11 @@ func (ff *fakeFirewallsProvider) doUpdateFirewall(f *compute.Firewall) error {
 		return fmt.Errorf("update failed for rule %v, srcRange %v ports %+v, rule not found", f.Name, f.SourceRanges, f.Allowed)
 	}
 
-	ff.fw[f.Name] = f
+	cf, err := copyFirewall(f)
+	if err != nil {
+		return fmt.Errorf("Failed to copy firewall: %v", err)
```
bowei: same here

pkg/controller/cluster_manager.go
nicksardo: On second thought, we'll need to pass in the list of ports used for NEG and merge them with the node port range.

nicksardo:
Do not merge

nicksardo:
I believe we have three options to handle NEG endpoints. 
1. Pass a separate set of ports to the firewall syncer and merge with the nodeport ranges. (Implemented in second commit)
2. Don't bother whitelisting individual ports. The L7 firewall rule opens up all TCP connections from the GCE src ranges. 
2. NEG use a separate instantiation of the firewall pool and creates a separate firewall rule.

nicksardo:
I need to update the static-ip E2E test.

PR #123, opened 2018-02-03
Sync ingress-specific backends and minor logging changes

nicksardo:
## Changes
- Backend service pool only syncs the list of service ports used for the respective ingress.
- Consolidate all ingress references to a single object which was deep copied.
- Changing log level of a frequently printed line from 4->5.
- Minor modifications to event handlers

#### Example with two ingresses in separate namespaces:
Before
```
# While syncing any ingress
I0205 23:03:18.031082       1 backends.go:216] Sync: backends [{30466 HTTP testing/my-echo-svc {1 0 my-http-port}} {32349 HTTP default/my-echo-svc {1 0 my-http-port}}]
```

After
```
# While syncing default/my-echo-svc
I0205 22:59:43.566574       1 backends.go:224] Sync: backends [{32349 HTTP default/my-echo-svc {1 0 my-http-port} 80 false}]
# While syncing testing/my-echo-svc
I0205 22:59:45.215794       1 backends.go:224] Sync: backends [{30466 HTTP testing/my-echo-svc {1 0 my-http-port} 80 false}]
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -119,19 +119,22 @@ func NewLoadBalancerController(kubeClient kubernetes.Interface, stopCh chan stru
 		AddFunc: func(obj interface{}) {
 			addIng := obj.(*extensions.Ingress)
 			if !isGCEIngress(addIng) && !isGCEMultiClusterIngress(addIng) {
-				glog.Infof("Ignoring add for ingress %v based on annotation %v", addIng.Name, annotations.IngressClassKey)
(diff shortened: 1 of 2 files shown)


nicksardo:
PTAL everyone

nikhiljindal:
/lgtm.
IIUC this should speed up the controller right? Only sync the ingress that changed!

nicksardo:
This should improve performance, but #106 was probably more impactful because it covered a few more kinds of resources.

PR #128, opened 2018-02-06
Rename Port to NodePort

nicksardo:
In order to prevent confusion with the `SvcPort` and `SvcTargetPort`, I suggest we be more specific by renaming `Port` to `NodePort`


nicksardo:
cc @rramkumar1

rramkumar1:
LGTM

PR #144, opened 2018-03-07
Adding information about using GCP SSL certs for frontend HTTPS

nikhiljindal:
Updating the documentation so that it clearly states the 2 options that users have to store their certs: as k8s secrets or as GCP SSL certs.

cc @csbell @G-Harmon @nicksardo @bowei


nikhiljindal:
PR updating kubemci documentation to point to these docs to keep them as source of truth: https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/pull/143

nicksardo:
/lgtm
Thanks!

PR #146, opened 2018-03-08
Ingress HTTP/2 Support

agau4779:
- Adds HTTP2 annotation to Service
- Creates alpha HealthCheck and BackendService resources when the protocol is HTTP/2, by using a composite BackendService type
- Tests for HealthCheck/BackendServices using HTTP/2

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -433,12 +433,7 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
 // edgeHop checks the links of the given backend by executing an edge hop.
 // It fixes broken links and updates the Backend accordingly.
 func (b *Backends) edgeHop(be *BackendService, igs []*compute.InstanceGroup) error {
-	addIGs := []*compute.InstanceGroup{}
-	if be.Alpha != nil {
-		addIGs = getAlphaInstanceGroupsToAdd(be, igs)
-	} else {
-		addIGs = getInstanceGroupsToAdd(be, igs)
-	}
+	addIGs := getInstanceGroupsToAdd(be, igs)
 	if len(addIGs) == 0 {
 		return nil
 	}
@@ -598,37 +593,11 @@ func getBackendsForNEGs(negs []*computealpha.NetworkEndpointGroup) []*computealp
 }
 
 func getInstanceGroupsToAdd(be *BackendService, igs []*compute.InstanceGroup) []*compute.InstanceGroup {
+	// A GA link can be used to reference an alpha object - so we only need to
+	// check the GA InstanceGroups.
 	beName := be.Ga.Name
 	beIGs := sets.String{}
 	for _, existingBe := range be.Ga.Backends {
-		beIGs.Insert(comparableGroupPath(existingBe.Group))
-	}
-
-	expectedIGs := sets.String{}
-	for _, ig := range igs {
-		expectedIGs.Insert(comparableGroupPath(ig.SelfLink))
-	}
-
-	if beIGs.IsSuperset(expectedIGs) {
-		return nil
-	}
-	glog.V(2).Infof("Expected igs for backend service %v: %+v, current igs %+v",
-		beName, expectedIGs.List(), beIGs.List())
-
-	var addIGs []*compute.InstanceGroup
-	for _, ig := range igs {
-		if !beIGs.Has(comparableGroupPath(ig.SelfLink)) {
-			addIGs = append(addIGs, ig)
-		}
-	}
-
-	return addIGs
-}
-
-func getAlphaInstanceGroupsToAdd(be *BackendService, igs []*compute.InstanceGroup) []*compute.InstanceGroup {
-	beName := be.Alpha.Name
-	beIGs := sets.String{}
-	for _, existingBe := range be.Alpha.Backends {
 		beIGs.Insert(comparableGroupPath(existingBe.Group))
 	}
 
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -94,13 +94,11 @@ type Backends struct {
 type BackendService struct {
 	Alpha *computealpha.BackendService
 	Ga    *compute.BackendService
-	// Flag to indicate we want to use the Alpha compute type
-	isAlpha bool
 }
 
 // GetProtocol gets the Protocol off the correct BackendService
 func (be *BackendService) GetProtocol() string {
-	if be.isAlpha {
+	if be.Alpha != nil {
 		return be.Alpha.Protocol
 	}
 
@@ -109,7 +107,7 @@ func (be *BackendService) GetProtocol() string {
 
 // GetDescription gets the Description off the correct BackendService
 func (be *BackendService) GetDescription() string {
-	if be.isAlpha {
+	if be.Alpha != nil {
 		return be.Alpha.Description
 	}
 
@@ -118,40 +116,34 @@ func (be *BackendService) GetDescription() string {
 
 // GetHealthCheckLink gets the Healthcheck link off the correct BackendService
 func (be *BackendService) GetHealthCheckLink() string {
-	if be.isAlpha && len(be.Alpha.HealthChecks) == 1 {
+	if be.Alpha != nil && len(be.Alpha.HealthChecks) == 1 {
 		return be.Alpha.HealthChecks[0]
-	} else if len(be.Ga.HealthChecks) == 1 {
+	}
+
+	if len(be.Ga.HealthChecks) == 1 {
 		return be.Ga.HealthChecks[0]
 	}
 
-	return ""
+	return "invalid-healthcheck-link"
 }
 
 // ensureProtocol updates the BackendService Protocol with the expected value
-func (be *BackendService) ensureProtocol(p ServicePort, hcLink string) (needsUpdate bool) {
+func (be *BackendService) ensureProtocol(p ServicePort) (needsUpdate bool) {
 	existingProtocol := be.GetProtocol()
 	if existingProtocol == string(p.Protocol) {
 		return false
 	}
 
-	// Using the HTTP2 Protocol requires that the BackendService is alpha.
-	// If the BackendService wasn't originally alpha, convert it to Alpha.
-	if !be.isAlpha && p.Protocol == annotations.ProtocolHTTP2 {
-		be.isAlpha = true
-		be.Alpha = toAlphaBackendService(be.Ga)
-	}
-
-	if be.isAlpha {
+	if be.Alpha != nil {
 		be.Alpha.Protocol = string(p.Protocol)
-	} else {
-		be.Ga.Protocol = string(p.Protocol)
 	}
 
+	be.Ga.Protocol = string(p.Protocol)
 	return true
 }
 
-// ensureHealthCheck updates the BackendService HealthCheck with the expected value
-func (be *BackendService) ensureHealthCheck(hcLink string) (needsUpdate bool) {
+// ensureHealthCheckLink updates the BackendService HealthCheck with the expected value
+func (be *BackendService) ensureHealthCheckLink(hcLink string) (needsUpdate bool) {
 	existingHCLink := be.GetHealthCheckLink()
 
 	// Compare health check name instead of health check link.
@@ -165,11 +157,11 @@ func (be *BackendService) ensureHealthCheck(hcLink string) (needsUpdate bool) {
 		return false
 	}
 
-	if be.isAlpha {
+	if be.Alpha != nil {
 		be.Alpha.HealthChecks = []string{hcLink}
-	} else {
-		be.Ga.HealthChecks = []string{hcLink}
 	}
+
+	be.Ga.HealthChecks = []string{hcLink}
 	return true
 }
 
@@ -179,42 +171,11 @@ func (be *BackendService) ensureDescription(description string) (needsUpdate boo
 	if existingDescription == description {
 		return false
 	}
-
-	if be.isAlpha {
+	if be.Alpha != nil {
 		be.Alpha.Description = description
-	} else {
-		be.Ga.Description = description
 	}
-
+	be.Ga.Description = description
 	return true
-}
-
-// ensureBackends generates backends with given instance groups with a specific mode
-func (be *BackendService) ensureBackends(bm BalancingMode, addIGs []*compute.InstanceGroup) {
-	if be.isAlpha {
-		originalIGBackends := []*computealpha.Backend{}
-		for _, backend := range be.Alpha.Backends {
-			// Backend service is not able to point to NEG and IG at the same time.
-			// Filter IG backends here.
-			if strings.Contains(backend.Group, "instanceGroups") {
-				originalIGBackends = append(originalIGBackends, backend)
-			}
-		}
-
-		newBackends := getAlphaBackendsForIGs(addIGs, bm)
-		be.Alpha.Backends = append(originalIGBackends, newBackends...)
-	} else {
-		originalIGBackends := []*compute.Backend{}
-		for _, backend := range be.Ga.Backends {
-			// Backend service is not able to point to NEG and IG at the same time.
-			// Filter IG backends here.
-			if strings.Contains(backend.Group, "instanceGroups") {
-				originalIGBackends = append(originalIGBackends, backend)
-			}
-		}
-		newBackends := getBackendsForIGs(addIGs, bm)
-		be.Ga.Backends = append(originalIGBackends, newBackends...)
-	}
 }
 
 // Backends is a BackendPool.
@@ -240,6 +201,12 @@ func (sp ServicePort) Description() string {
 		return ""
 	}
 	return fmt.Sprintf(`{"kubernetes.io/service-name":"%s","kubernetes.io/service-port":"%s"}`, sp.SvcName.String(), sp.SvcPort.String())
+}
+
+// isAlpha returns true if the ServicePort is using ProtocolHTTP2 - which means
+// we need to use the Alpha API.
+func (sp ServicePort) isAlpha() bool {
+	return sp.Protocol == annotations.ProtocolHTTP2
 }
 
 // NewBackendPool returns a new backend pool.
@@ -295,30 +262,27 @@ func (b *Backends) Init(pp ProbeProvider) {
 }
 
 // Get returns a single backend.
-func (b *Backends) Get(port int64) (*BackendService, error) {
+func (b *Backends) Get(port int64, isAlpha bool) (*BackendService, error) {
 	beGa, err := b.cloud.GetGlobalBackendService(b.namer.Backend(port))
 	if err != nil {
 		return nil, err
 	}
 
-	beAlpha := &computealpha.BackendService{}
-	isAlpha := false
-
+	var beAlpha *computealpha.BackendService
 	// If the Protocol is empty, this means this is a alpha BackendService and
 	// Protocol is expected to be HTTP2
 	// WARNING: If a user has created an alpha BackendService in the past but
 	// is no longer alpha whitelisted, this will always return an isForbidden err
 	// until the user can access the alpha APIs again.
-	if beGa.Protocol == "" {
+	if beGa.Protocol == "" || isAlpha {
 		beAlpha, err = b.cloud.GetAlphaGlobalBackendService(b.namer.Backend(port))
 		if err != nil {
 			return nil, err
 		}
-		isAlpha = true
 	}
 
 	b.snapshotter.Add(portKey(port), beGa)
-	return &BackendService{Ga: beGa, Alpha: beAlpha, isAlpha: isAlpha}, nil
+	return &BackendService{Ga: beGa, Alpha: beAlpha}, nil
 }
 
 func (b *Backends) ensureHealthCheck(sp ServicePort) (string, error) {
@@ -346,27 +310,36 @@ func (b *Backends) ensureHealthCheck(sp ServicePort) (string, error) {
 }
 
 func (b *Backends) create(namedPort *compute.NamedPort, hcLink string, sp ServicePort, name string) (*BackendService, error) {
-	bs := &compute.BackendService{
-		Name:         name,
-		Description:  sp.Description(),
-		Protocol:     string(sp.Protocol),
-		HealthChecks: []string{hcLink},
-		Port:         namedPort.Port,
-		PortName:     namedPort.Name,
-	}
-
-	if sp.Protocol != annotations.ProtocolHTTP2 {
-		if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
-			return nil, err
+	isAlpha := sp.isAlpha()
+	if isAlpha {
+		bsAlpha := &computealpha.BackendService{
+			Name:         name,
+			Description:  sp.Description(),
+			Protocol:     string(sp.Protocol),
+			HealthChecks: []string{hcLink},
+			Port:         namedPort.Port,
+			PortName:     namedPort.Name,
 		}
-	} else {
-		bsAlpha := toAlphaBackendService(bs)
+
 		if err := b.cloud.CreateAlphaGlobalBackendService(bsAlpha); err != nil {
 			return nil, err
 		}
+	} else {
+		bs := &compute.BackendService{
+			Name:         name,
+			Description:  sp.Description(),
+			Protocol:     string(sp.Protocol),
+			HealthChecks: []string{hcLink},
+			Port:         namedPort.Port,
+			PortName:     namedPort.Name,
+		}
+
+		if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
+			return nil, err
+		}
 	}
 
-	return b.Get(namedPort.Port)
+	return b.Get(namedPort.Port, isAlpha)
 }
 
 // Ensure will update or create Backends for the given ports.
@@ -415,7 +388,7 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
 
 	// Verify existance of a backend service for the proper port, but do not specify any backends/igs
 	beName := b.namer.Backend(p.NodePort)
-	be, _ = b.Get(p.NodePort)
+	be, _ = b.Get(p.NodePort, p.isAlpha())
 	if be == nil {
 		namedPort := &compute.NamedPort{
 			Name: b.namer.NamedPort(p.NodePort),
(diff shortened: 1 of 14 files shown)


pkg/backends/backends.go
```diff
@@ -262,14 +287,14 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
-	be, _ = b.Get(p.NodePort)
+	be, _ = b.GetAlpha(p.NodePort)
 	if be == nil {
 		namedPort := &compute.NamedPort{
 			Name: b.namer.NamedPort(p.NodePort),
 			Port: p.NodePort,
 		}
 		glog.V(2).Infof("Creating backend service for port %v named port %v", p.NodePort, namedPort)
-		be, err = b.create(namedPort, hcLink, p, beName)
+		be, err = b.createAlpha(namedPort, hcLink, p, beName)
```
bowei: we can't call alpha on the GA, beta paths.

agau4779:
@freehan - we're going to Feature Gate this.

pkg/backends/backends.go
```diff
@@ -168,13 +176,27 @@ func (b *Backends) Init(pp ProbeProvider) {
...
 	}
-	b.snapshotter.Add(portKey(port), be)
-	return be, nil
+
+	beAlpha := &computealpha.BackendService{}
+	isAlpha := false
+
+	// If the Protocol is empty, this means this is a alpha BackendService and
+	// Protocol is expected to be HTTP2
+	if beGa.Protocol == "" {
```
freehan: What happens if user manually change the protocol?

pkg/backends/backends.go
```diff
@@ -305,17 +353,80 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+	// If there are instance pools(node pool is synced) and NEG is not enabled,
+	// perform edgeHop to verify that BackendServices contains links to all
+	// backends/instancegroups
+	if len(igs) > 0 && !p.NEGEnabled {
+		return b.edgeHopAndUpdate(be, igs)
 	}
 
-	// If NEG is enabled, do not link backend service to instance groups.
-	if p.NEGEnabled {
+	return b.Update(be)
```
freehan: This seems wrong. It will always update backend service.  Please only update when necessary

Move this back to above

pkg/backends/backends.go
```diff
@@ -386,31 +517,31 @@ func getBackendsForNEGs(negs []*computealpha.NetworkEndpointGroup) []*computealp
...
 
-// edgeHop checks the links of the given backend by executing an edge hop.
-// It fixes broken links.
-func (b *Backends) edgeHop(be *compute.BackendService, igs []*compute.InstanceGroup) error {
+func getInstanceGroupsToAdd(be *BackendService, igs []*compute.InstanceGroup) []*compute.InstanceGroup {
+	beName := be.Ga.Name
 	beIGs := sets.String{}
-	for _, beToIG := range be.Backends {
-		beIGs.Insert(beToIG.Group)
+	if be.isAlpha {
```
freehan: Do you have to distinguish between Alpha and GA backend services here?

Or consider have some helper function with BackendService struct


IIUC, the URLs are the same except for the API version part.

pkg/backends/backends.go
```diff
@@ -305,17 +353,80 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
-	if p.NEGEnabled {
+	return b.Update(be)
+}
+
+// edgeHopAndUpdate checks the links of the given backend by executing an edge hop.
+// It fixes broken links and updates the Backend accordingly.
+func (b *Backends) edgeHopAndUpdate(be *BackendService, igs []*compute.InstanceGroup) error {
+	addIGs := getInstanceGroupsToAdd(be, igs)
+	if len(addIGs) == 0 {
+		return b.Update(be)
```
freehan: If no Instance Group needs to be added, why do you need to update?

edgeHop is mostly for linking backends and backend service

pkg/backends/backends.go
```diff
@@ -288,13 +324,25 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+	// Compare the Alpha fields if the BackendService has Alpha features, e.g. HTTP2
+	if be.isAlpha {
+		existingProtocol = be.Alpha.Protocol
+		existingDescription = be.Alpha.Description
+	}
+
+	if existingProtocol != string(p.Protocol) || existingHCName != expectedHCName || existingDescription != p.Description() {
+		glog.V(2).Infof("Updating backend protocol %v (%v) for change in protocol (%v) or health check", beName, existingProtocol, string(p.Protocol))
+		if be.isAlpha {
+			be.Alpha.Protocol = string(p.Protocol)
```
freehan: worth wraping this into a separate function that ensures protocol, hclink and description in a be

pkg/backends/backends.go
```diff
@@ -305,17 +353,80 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+	addIGs := getInstanceGroupsToAdd(be, igs)
+	if len(addIGs) == 0 {
+		return b.Update(be)
+	}
+
+	var errs []string
+	for _, bm := range []BalancingMode{Rate, Utilization} {
+		// Generate backends with given instance groups with a specific mode
+		if be.isAlpha {
+			originalIGBackends := []*computealpha.Backend{}
```
freehan: wrap this into a separate function to generate corresponding backends?

pkg/backends/backends.go
```diff
@@ -90,6 +90,14 @@ type Backends struct {
 	namer        *utils.Namer
 }
 
+// BackendService embeds both the GA and alpha compute BackendService types
+type BackendService struct {
```
freehan: Please add some helper functions to this struct. For instance,

- Ensure protocol, healthcheck and description, (return if there is difference)
- Ensure Backends (check if current backend instance groups are super set)

pkg/backends/backends.go
```diff
@@ -305,17 +353,80 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
 		return nil
 	}
-	// Verify that backend service contains links to all backends/instance-groups
-	return b.edgeHop(be, igs)
+
+	return fmt.Errorf("received errors when updating backend service: %v", strings.Join(errs, "\n"))
+}
+
+// Update calls either the GA or Alpha update path depending on Protocol
+func (b *Backends) Update(be *BackendService) error {
```
freehan: Move the log line here to log it is updating backend service

freehan: Does this function has to be public?

pkg/healthchecks/healthchecks.go
```diff
@@ -146,7 +147,10 @@ func (h *HealthChecks) create(hc *HealthCheck) error {
 }
 
 func (h *HealthChecks) update(oldHC, newHC *HealthCheck) error {
-	if newHC.ForNEG {
+	if newHC.isHttp2() {
+		glog.V(2).Infof("Updating health check with protocol %v", newHC.Type)
+		return h.cloud.UpdateAlphaHealthCheck(newHC.ToAlphaComputeHealthCheck())
```
freehan: Probably need to copy the same pattern for NEG health checks. Need to ensure existing settings are preserved. 
@nicksardo Do you think it is necessary for HTTP2?

pkg/backends/backends.go
agau4779: Oops, good point. Removing

pkg/healthchecks/healthchecks.go
nicksardo: Yea, settings should absolutely be preserved as they are done today. IIRC, we change some parameters for NEGs, but never the path value. 

We'll also need to consider the case of NEGs + HTTP2.

pkg/backends/backends.go
agau4779: Opted to add new logging for `update` here - https://github.com/kubernetes/ingress-gce/pull/146/commits/f6ca1c733bb0716a46450c278171c5f7adc23bcb

pkg/backends/backends.go
agau4779: https://github.com/kubernetes/ingress-gce/pull/146/commits/4de3e8b0788236ba2c6e68ca13a14def7ade0cea

pkg/backends/backends.go
```diff
@@ -168,13 +217,27 @@ func (b *Backends) Init(pp ProbeProvider) {
...
 	}
-	b.snapshotter.Add(portKey(port), be)
-	return be, nil
+
+	beAlpha := &computealpha.BackendService{}
+	isAlpha := false
+
+	// If the Protocol is empty, this means this is a alpha BackendService and
+	// Protocol is expected to be HTTP2
+	if beGa.Protocol == "" {
```
freehan: Is there any possibility that Protocol is empty and GetAlphaGlobalBackendService keeps returning error?

pkg/backends/backends.go
```diff
@@ -210,9 +273,20 @@ func (b *Backends) create(namedPort *compute.NamedPort, hcLink string, sp Servic
 		Port:         namedPort.Port,
 		PortName:     namedPort.Name,
 	}
-	if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
-		return nil, err
+
+	bsAlpha := &computealpha.BackendService{}
```
freehan: move to the following else block

pkg/backends/backends.go
```diff
@@ -288,12 +365,20 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
 	// TODO (mixia): compare health check link directly once NEG is GA
 	existingHCName := retrieveObjectName(existingHCLink)
 	expectedHCName := retrieveObjectName(hcLink)
-	if be.Protocol != string(p.Protocol) || existingHCName != expectedHCName || be.Description != p.Description() {
-		glog.V(2).Infof("Updating backend protocol %v (%v) for change in protocol (%v) or health check", beName, be.Protocol, string(p.Protocol))
-		be.Protocol = string(p.Protocol)
-		be.HealthChecks = []string{hcLink}
-		be.Description = p.Description()
-		if err = b.cloud.UpdateGlobalBackendService(be); err != nil {
+	existingProtocol := be.Ga.Protocol
```
freehan: existingProtocol := be.GetProtocol()
existingProtocol := be.GetDescription()

pkg/backends/backends.go
```diff
@@ -288,12 +365,20 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+	existingDescription := be.Ga.Description
+
+	// Compare the Alpha fields if the BackendService has Alpha features, e.g. HTTP2
+	if be.isAlpha {
+		existingProtocol = be.Alpha.Protocol
+		existingDescription = be.Alpha.Description
+	}
+
+	if existingProtocol != string(p.Protocol) || existingHCName != expectedHCName || existingDescription != p.Description() {
+		glog.V(2).Infof("Updating backend protocol %v (%v) for change in protocol (%v) or health check", beName, existingProtocol, string(p.Protocol))
```
freehan: remove this since b.update() has a log already.

pkg/backends/backends.go
```diff
@@ -288,12 +365,20 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+		be.ensureProtocol(p, hcLink)
```
freehan:
```
if be.ensureProtocol(p) || be.ensureHealthCheck(hcLink) {
     if err = b.update(be); err != nil {
 	return err
     }
}
```

`ensureProtocol` and `ensureHealthCheck` returns true if existing parameter does not match expected. Move all existingXXX into the helper functions.

pkg/backends/backends.go
```diff
@@ -277,8 +351,11 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
 
 	// Check that the backend service has the correct protocol and health check link
 	existingHCLink := ""
```
freehan: existingHCLink := be.GetHealthCheckLink()

pkg/backends/backends.go
```diff
@@ -90,6 +90,55 @@ type Backends struct {
...
+		be.Alpha.Description = p.Description()
+	} else {
+		be.Ga.Protocol = string(p.Protocol)
+		be.Ga.HealthChecks = []string{hcLink}
+		be.Ga.Description = p.Description()
+	}
+}
+
+// generateBackends generates backends with given instance groups with a specific mode
+func (be *BackendService) generateBackends(bm BalancingMode, addIGs []*compute.InstanceGroup) {
```
freehan: rename to ensureBackends?

pkg/backends/backends.go
```diff
@@ -386,67 +530,32 @@ func getBackendsForNEGs(negs []*computealpha.NetworkEndpointGroup) []*computealp
...
+func getInstanceGroupsToAdd(be *BackendService, igs []*compute.InstanceGroup) []*compute.InstanceGroup {
+	beName := be.Ga.Name
 	beIGs := sets.String{}
-	for _, beToIG := range be.Backends {
-		beIGs.Insert(beToIG.Group)
+	for _, existingBe := range be.Ga.Backends {
+		beIGs.Insert(comparableGroupPath(existingBe.Group))
 	}
+
 	igLinks := sets.String{}
```
freehan: s/igLinks/expectedIGs

pkg/healthchecks/healthchecks.go
freehan: @agau4779 need to rethink the logic here. 

1. Need to handle PortSpecification + HTTP2
2. Need to preserve whatever user healthcheck configuration even if protocol or NEG setting changes.

pkg/backends/backends.go
agau4779: `GetGlobalBackendService` gets the GA BackendService but returns an empty string in the Protocol if the Protocol is not HTTP or HTTPS - so here, we're anticipating that the user created an Alpha BackendService with an HTTP2 protocol (currently alpha-only), so calling the GA path will return a blank protocol.

`GetAlphaGlobalBackendService` may run into an `isForbidden` error if the user is no longer Alpha whitelisted - in which case, might cause the Ingress controller to be stuck forever. Will add this warning in a code comment.

pkg/backends/backends.go
```diff
@@ -90,6 +90,133 @@ type Backends struct {
...
+	}
+
+	return be.Ga.Description
+}
+
+// GetHealthCheckLink gets the Healthcheck link off the correct BackendService
+func (be *BackendService) GetHealthCheckLink() string {
+	if be.isAlpha && len(be.Alpha.HealthChecks) == 1 {
+		return be.Alpha.HealthChecks[0]
+	} else if len(be.Ga.HealthChecks) == 1 {
```
bowei: you don't need `else` here

pkg/backends/backends.go
```diff
@@ -90,6 +90,133 @@ type Backends struct {
...
+		return be.Ga.HealthChecks[0]
+	}
+
+	return ""
```
bowei: I would return "invalid-healthcheck link"

pkg/backends/backends.go
```diff
@@ -90,6 +90,133 @@ type Backends struct {
...
+		for _, backend := range be.Alpha.Backends {
+			// Backend service is not able to point to NEG and IG at the same time.
+			// Filter IG backends here.
+			if strings.Contains(backend.Group, "instanceGroups") {
+				originalIGBackends = append(originalIGBackends, backend)
+			}
+		}
+
+		newBackends := getAlphaBackendsForIGs(addIGs, bm)
+		be.Alpha.Backends = append(originalIGBackends, newBackends...)
```
bowei: return and outdent the else (remove else)

pkg/backends/backends.go
```diff
@@ -210,9 +354,18 @@ func (b *Backends) create(namedPort *compute.NamedPort, hcLink string, sp Servic
 		Port:         namedPort.Port,
 		PortName:     namedPort.Name,
 	}
-	if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
-		return nil, err
+
+	if sp.Protocol != annotations.ProtocolHTTP2 {
```
bowei: should the feature gate be here?

pkg/backends/backends.go
```diff
@@ -582,3 +748,10 @@ func retrieveObjectName(url string) string {
 	splited := strings.Split(url, "/")
 	return splited[len(splited)-1]
 }
+
+// comparableGroupPath trims project and compute version from the SelfLink
+// /zones/[ZONE_NAME]/instanceGroups/[IG_NAME]
+func comparableGroupPath(url string) string {
```
bowei: does this have a unit test

pkg/backends/backends.go
```diff
@@ -90,6 +90,14 @@ type Backends struct {
 	namer        *utils.Namer
 }
 
+// BackendService embeds both the GA and alpha compute BackendService types
+type BackendService struct {
+	Alpha *computealpha.BackendService
+	Ga    *compute.BackendService
+	// Flag to indicate we want to use the Alpha compute type
+	isAlpha bool
```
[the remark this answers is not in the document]
agau4779: There is not - when getting an Alpha BackendService, or promoting an GA BackendService to Alpha, isAlpha is set to true. Looks like it's redundant if we can just check `BackendService.Alpha == nil` - removing.

pkg/backends/backends.go
agau4779: done here - https://github.com/kubernetes/ingress-gce/pull/146/commits/f1b64173457b2b6347f90309fac2dec5bdb95b9e

pkg/backends/backends.go
agau4779: https://github.com/agau4779/ingress-gce/blob/f1b64173457b2b6347f90309fac2dec5bdb95b9e/pkg/backends/backends_test.go#L501-L528 uses `comparableGroupPath` and https://github.com/agau4779/ingress-gce/blob/be20ec01655b1e5ce7912174160ab477f19c20dc/pkg/backends/backends_test.go#L678-L702 explicitly tests it.

pkg/backends/backends.go
```diff
@@ -275,47 +425,75 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+func (b *Backends) edgeHop(be *BackendService, igs []*compute.InstanceGroup) error {
+	addIGs := getInstanceGroupsToAdd(be, igs)
+	if len(addIGs) == 0 {
 		return nil
 	}
 
-	// If NEG is enabled, do not link backend service to instance groups.
-	if p.NEGEnabled {
+	var errs []string
+	for _, bm := range []BalancingMode{Rate, Utilization} {
```
freehan: move the original comments here.

pkg/backends/backends.go
```diff
@@ -90,6 +90,133 @@ type Backends struct {
...
+
+	if len(be.Ga.HealthChecks) == 1 {
+		return be.Ga.HealthChecks[0]
+	}
+
+	return "invalid-healthcheck-link"
+}
+
+// ensureProtocol updates the BackendService Protocol with the expected value
+func (be *BackendService) ensureProtocol(p ServicePort, hcLink string) (needsUpdate bool) {
```
freehan: hcLink not used right?

pkg/backends/backends.go
```diff
@@ -90,6 +90,133 @@ type Backends struct {
...
+	}
+
+	// Using the HTTP2 Protocol requires that the BackendService is alpha.
+	// If the BackendService wasn't originally alpha, convert it to Alpha.
+	if !(be.Alpha != nil) && p.Protocol == annotations.ProtocolHTTP2 {
+		be.Alpha = toAlphaBackendService(be.Ga)
+	}
+
+	if be.Alpha != nil {
+		be.Alpha.Protocol = string(p.Protocol)
```
freehan: IIUC, if the existing backend-service is using HTTP2. If the new protocol set to HTTP. Then it will set protocol on the alpha object. Hence, during update, it will call Alpha API to update backend service. 

Try to avoid calling alpha API as much as possible. SLA is different.

pkg/healthchecks/healthchecks.go
```diff
@@ -146,9 +147,12 @@ func (h *HealthChecks) create(hc *HealthCheck) error {
 }
 
 func (h *HealthChecks) update(oldHC, newHC *HealthCheck) error {
-	if newHC.ForNEG {
+	if newHC.isHttp2() {
```
freehan: merge NEG and isHttp2 conditions

freehan:
Please also unit test the `ensurexxxxx` methods.

pkg/backends/backends.go
agau4779: Hmm, then should we convert it to a Ga backend service in this function?
e.g.

```
	if p.Protocol != annotations.ProtocolHTTP2 {
		be.Ga = toV1BackendService(be.Alpha)
		be.Alpha = nil
	}
```

pkg/backends/backends.go
```diff
@@ -210,9 +358,18 @@ func (b *Backends) create(namedPort *compute.NamedPort, hcLink string, sp Servic
...
 	}
-	if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
-		return nil, err
+
+	if sp.Protocol != annotations.ProtocolHTTP2 {
+		if err := b.cloud.CreateGlobalBackendService(bs); err != nil {
+			return nil, err
+		}
+	} else {
+		bsAlpha := toAlphaBackendService(bs)
```
freehan: `toAlphaBackendService` is mostly for testing. 

Are you confident about its reliability for production use?

pkg/backends/backends.go
```diff
@@ -90,6 +90,140 @@ type Backends struct {
...
+// ensureProtocol updates the BackendService Protocol with the expected value
+func (be *BackendService) ensureProtocol(p ServicePort) (needsUpdate bool) {
+	existingProtocol := be.GetProtocol()
+	if existingProtocol == string(p.Protocol) {
+		return false
+	}
+
+	// Using the HTTP2 Protocol requires that the BackendService is alpha.
+	// If the BackendService wasn't originally alpha, convert it to Alpha.
+	if be.Alpha == nil && p.Protocol == annotations.ProtocolHTTP2 {
```
freehan: Probably do this: `ensureProtocol` just ensures the protocol for both GA and Alpha objects if exists.

In `update`, pick the api to call.

pkg/backends/backends.go
```diff
@@ -275,47 +401,106 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
 		}
 	}
 
-	// Check that the backend service has the correct protocol and health check link
-	existingHCLink := ""
-	if len(be.HealthChecks) == 1 {
-		existingHCLink = be.HealthChecks[0]
-	}
+	needUpdate := be.ensureProtocol(p)
+	needUpdate = needUpdate || be.ensureHealthCheckLink(hcLink)
```
freehan: quick question: 
GA health checks supports HTTP2?

pkg/healthchecks/healthchecks.go
```diff
@@ -159,20 +160,20 @@ func (h *HealthChecks) update(oldHC, newHC *HealthCheck) error {
 	}
 }
 
-// mergeHealthcheckForNEG merges old health check configuration (potentially for IG) with the new one.
+// mergeHealthcheck merges old health check configuration (potentially for IG) with the new one.
 // This is to preserve the existing health check setting as much as possible.
 // WARNING: if a service backend is converted from IG mode to NEG mode,
 // the existing health check setting will be preserve, although it may not suit the customer needs.
-func mergeHealthcheckForNEG(oldHC, newHC *HealthCheck) *HealthCheck {
+func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
```
freehan: Seems incorrect. Does it preserve configs if oldHC is HTTPS or HTTP2?

pkg/backends/backends.go
```diff
@@ -275,47 +401,106 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
+	if len(igs) > 0 && !p.NEGEnabled {
+		return b.edgeHop(be, igs)
+	}
+
+	return nil
+}
+
+// edgeHop checks the links of the given backend by executing an edge hop.
+// It fixes broken links and updates the Backend accordingly.
+func (b *Backends) edgeHop(be *BackendService, igs []*compute.InstanceGroup) error {
```
freehan: Cannot think of a better way to avoid duplicate code. 

Do you think it is worthwhile to hide the complexity behind `ensureInstanceGroupBackends`? And do something clever behind the scene and unit test it thoroughly.

freehan: Is it possible to only use GA backend service to ensure backends?

Will it overwrite the HTTP2 config?

freehan: If GA Backend-service set to HTTP2, you probably gets a validation error. 

But if you get a GA backend-service, and update its backend without touching protocol. What will happen?

agau4779: We can't use the GA BackendService API. Verified that setting the Protocol on an HTTP or HTTPS BackendService to HTTP2 causes a validation error, and that modifying any other attribute on an HTTP2 BackendService causes the BackendService's Protocol to default to HTTP. The Sync loop eventually calls backends.go#Ensure then causes it to revert back to HTTP2.

pkg/backends/backends.go
agau4779: GA Healthchecks do not support HTTP2. HTTP2 healthchecks are created through the Alpha API.

pkg/backends/backends.go
```diff
@@ -90,6 +90,94 @@ type Backends struct {
...
+	// Hence, it will cause the health check links to be always different
+	// TODO (mixia): compare health check link directly once NEG is GA
+	existingHCName := retrieveObjectName(existingHCLink)
+	expectedHCName := retrieveObjectName(hcLink)
+	if existingHCName == expectedHCName {
+		return false
+	}
+
+	if be.Alpha != nil {
+		be.Alpha.HealthChecks = []string{hcLink}
```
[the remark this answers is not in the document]
agau4779: Shouldn't be possible -` GetHealthCheckLink` performs a check - `if be.Alpha != nil && len(be.Alpha.HealthChecks) == 1` then hcLink will return an Alpha healthcheck link; otherwise a Ga link is returned.

The compute API currently returns the alpha version of the Healthcheck link through the Alpha API and the GA version of the link through the GA API (`gcloud [alpha] compute backend-services describe [service-name]`), so I agree that it should be ok.

freehan:
LGTM

@bowei @nicksardo do you want another look?

PR #148, opened 2018-03-13
Initial implementation for ingress rate limiting

rramkumar1:
This PR adds support for rate limiting certain GCE API call's that are made by the controller. Currently, the cloud provider layer does not rate limit any standard controller calls (such as BackendService Get). As a result, in some cases when API calls are frequent, we are blowing through project quota and getting rate limited on the GCE side. This could also cause the controller to slow down. By allowing for customization of what calls can be rate limited, we can prevent blowing through project quota and we can intelligently rate limit so that we do not experience slowdown.  

Will add unit tests once the overall approach is LGTM'd.

/assign @nicksardo @bowei

### DIFF
--- a/cmd/glbc/app/clients.go
+++ b/cmd/glbc/app/clients.go
@@ -36,6 +36,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 
 	"k8s.io/ingress-gce/pkg/flags"
+	"k8s.io/ingress-gce/pkg/ratelimit"
 	"k8s.io/ingress-gce/pkg/utils"
 )
 
@@ -95,6 +96,12 @@ func NewGCEClient() *gce.GCECloud {
 		provider, err := cloudprovider.GetCloudProvider("gce", configReader())
 		if err == nil {
 			cloud := provider.(*gce.GCECloud)
+			// Configure GCE rate limiting
+			rl, err := ratelimit.NewGCERateLimiter(flags.F.GCERateLimit.Values())
+			if err != nil {
+				glog.Errorf("Error in configuring rate limiting: %v", err)
+			}
+			cloud.SetRateLimiter(rl)
 			// If this controller is scheduled on a node without compute/rw
 			// it won't be allowed to list backends. We can assume that the
 			// user has no need for Ingress in this case. If they grant
--- a/cmd/glbc/app/clients.go
+++ b/cmd/glbc/app/clients.go
@@ -99,7 +99,7 @@ func NewGCEClient() *gce.GCECloud {
 			// Configure GCE rate limiting
 			rl, err := ratelimit.NewGCERateLimiter(flags.F.GCERateLimit.Values())
 			if err != nil {
-				glog.Errorf("Error in configuring rate limiting: %v", err)
+				glog.Fatalf("Error configuring rate limiting: %v", err)
 			}
 			cloud.SetRateLimiter(rl)
 			// If this controller is scheduled on a node without compute/rw
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -43,6 +43,7 @@ var (
 		ConfigFilePath  string
 		DefaultSvc      string
 		DeleteAllOnQuit bool
+		GCERateLimit    RateLimitSpecs
 		HealthCheckPath string
 		HealthzPort     int
 		InCluster       bool
@@ -58,6 +59,7 @@ var (
 
 func init() {
 	F.NodePortRanges.ports = []string{DefaultNodePortRange}
+	F.GCERateLimit.specs = []string{}
 }
 
 // Register flags with the command line parser.
@@ -85,6 +87,11 @@ the default backend.`)
 external cloud resources as it's shutting down. Mostly used for testing. In
 normal environments the controller should only delete a loadbalancer if the
 associated Ingress is deleted.`)
+	flag.Var(&F.GCERateLimit, "gce-ratelimit",
+		`Optional, can be used to rate limit certain GCE API calls. Example usage:
+--gce-ratelimit=ga.Addresses.Get,qps,1.5,5
+(limit ga.Addresses.Get to maximum of 1.5 qps with a burst of 5).
+Use the flag more than once to rate limit more than one call.`)
 	flag.StringVar(&F.HealthCheckPath, "health-check-path", "/",
 		`Path used to health-check a backend service. All Services must serve a
 200 page on this path. Currently this is only configurable globally.`)
@@ -111,6 +118,25 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`This flag is deprecated. Use -v to control verbosity.`)
 	flag.Bool("use-real-cloud", false,
 		`This flag has been deprecated and no longer has any effect.`)
+}
+
+type RateLimitSpecs struct {
+	specs []string
+}
+
+// Part of the flag.Value interface.
+func (r *RateLimitSpecs) String() string {
+	return strings.Join(r.specs, ";")
+}
+
+// Set supports the flag being repeated multiple times. Part of the flag.Value interface.
+func (r *RateLimitSpecs) Set(value string) error {
+	r.specs = append(r.specs, value)
+	return nil
+}
+
+func (r *RateLimitSpecs) Values() []string {
+	return r.specs
 }
 
 type PortRanges struct {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -33,6 +33,8 @@ const (
 	// DefaultNodePortRange is the list of ports or port ranges used by kubernetes for
 	// allocating NodePort services.
 	DefaultNodePortRange = "30000-32767"
+	// DefaultGCERateLimit is the default rate limit spec.
+	DefaultGCERateLimit = "*.Operation.Get,qps,10,100"
 )
 
 var (
@@ -59,7 +61,7 @@ var (
 
 func init() {
 	F.NodePortRanges.ports = []string{DefaultNodePortRange}
-	F.GCERateLimit.specs = []string{}
+	F.GCERateLimit.specs = []string{DefaultGCERateLimit}
 }
 
 // Register flags with the command line parser.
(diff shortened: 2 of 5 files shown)


pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,122 @@
...
+		Operation: operation,
+		Version: version
+		Service: service,
+	}
+}
+
+func constructRateLimitImpl(params []string) flowcontrol.RateLimiter, error {
+	// For now, only the "qps" type is supported.
+	rlType := params[0]
+	implArgs := params[1:]
```
nicksardo: Are implArgs required? You assert the length of params >= 2, then send [1:] to this func.

pkg/flags/flags.go
```diff
@@ -85,6 +86,10 @@ the default backend.`)
 external cloud resources as it's shutting down. Mostly used for testing. In
 normal environments the controller should only delete a loadbalancer if the
 associated Ingress is deleted.`)
+	flag.StringVar(&F.GCERateLimit, "gce-ratelimit", "",
+		`Optional, can be used to rate limit certain GCE API calls. Example usage:
+--gce-ratelimit=ga.Addresses.Get,qps,30 (limit ga.Addresses.Get to maximum
```
nicksardo: Doesn't show an example using the semi-colon delimiter. 
Doesn't show the burst.

nicksardo: Let's also set the example to something logical: 1.25 or 1.5 QPS.

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/gce/support.go
```diff
@@ -62,5 +69,22 @@ func (l *gceRateLimiter) Accept(ctx context.Context, key *cloud.RateLimitKey) er
 			return ctx.Err()
 		}
 	}
+	if l.customRateLimiters != nil {
+		for _, c := range l.customRateLimiters {
+			if c.useLimiter(key) {
+				return c.rateLimiter.Accept(ctx, key)
```
nicksardo: As we discussed offline, I'm not seeing the immediate need of having a list over here and a map in `ratelimit.go`. Could it be simplified to a single map here?

pkg/flags/flags.go
rramkumar1: Done.

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/gce/support.go
rramkumar1: Done. Removed the usage of a map in ingress code, but left the list in vendored code. Still thinking about whether a map or list is better.

pkg/ratelimit/ratelimit.go
rramkumar1: If the length of the array passed in to constructRateLimitImpl is 1, then params[1:] will return empty. Is this what you are referring to?

nicksardo: Yea, that's fine.

Can you add the expected format as a comment above the func - same thing you did for `validateOperation`.

rramkumar1: Done.

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/gce/gce.go
```diff
@@ -510,6 +519,7 @@ func CreateGCECloud(config *CloudConfig) (*GCECloud, error) {
 		nodeInstancePrefix:       config.NodeInstancePrefix,
 		useMetadataServer:        config.UseMetadataServer,
 		operationPollRateLimiter: operationPollRateLimiter,
+		customRateLimiters:       make([]CustomRateLimiter, 0),
```
nicksardo: The `, 0` is unnecessary.

rramkumar1: Fixed

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/gce/gce.go
```diff
@@ -526,6 +536,13 @@ func CreateGCECloud(config *CloudConfig) (*GCECloud, error) {
 	return gce, nil
 }
 
+// AddRateLimiter adds a custom cloud.RateLimiter implementation and a
+// function that is used to check whether the custom rate limiter can be
+// used on a given RateLimitKey.
+func (g *GCECloud) AddRateLimiter(rl cloud.RateLimiter, useLimiter func(key *cloud.RateLimitKey) bool) {
```
nicksardo: Any reason to make this an `Add...` instead of `SetRateLimiters`?  I'm not suggesting a change, just wondering.

cmd/glbc/main.go
```diff
@@ -69,6 +70,10 @@ func main() {
 	}
 
 	cloud := app.NewGCEClient()
+	if err := ratelimit.ConfigureGCERateLimiting(cloud, flags.F.GCERateLimit); err != nil {
```
nicksardo: Should this be called within app.NewGCEClient()?

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/gce/gce.go
rramkumar1: I figured that since this could be called multiple times, that Add... makes more sense.

cmd/glbc/main.go
rramkumar1: Done.

cmd/glbc/app/clients.go
```diff
@@ -102,6 +103,11 @@ func NewGCEClient() *gce.GCECloud {
 			// manually to re-create the client.
 			// TODO: why do we bail with success out if there is a permission error???
 			if _, err = cloud.ListGlobalBackendServices(); err == nil || utils.IsHTTPErrorCode(err, http.StatusForbidden) {
+				rl, err := ratelimit.ConfigureGCERateLimiting(flags.F.GCERateLimit)
+				if err != nil {
+					glog.Infof("Error in configuring rate limiting: %v", err)
```
G-Harmon: not an error? (either glog.Errorf() or returning nil)

rramkumar1: Good catch, fixed.

rramkumar1: Regarding your other point, I'll add some context to the PR.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return &ingressRateLimiter{rateLimitImpls}, nil
+}
+
+// Implementation of cloud.RateLimiter
+func (l *ingressRateLimiter) Accept(ctx context.Context, key *cloud.RateLimitKey) error {
+	ch := make(chan struct{})
+	go func() {
+		// Call flowcontrol.RateLimiter implementation.
+		if l.rateLimitImpls[key] != nil {
```
bowei: This doesn't work -- the project ID will not be equal. You will have to write a routine that does the matching (can't use struct equality)

pkg/flags/flags.go
```diff
@@ -85,6 +86,11 @@ the default backend.`)
 external cloud resources as it's shutting down. Mostly used for testing. In
 normal environments the controller should only delete a loadbalancer if the
 associated Ingress is deleted.`)
+	flag.StringVar(&F.GCERateLimit, "gce-ratelimit", "",
```
bowei: There is a way to handle multiple instances of the same flag and have them append to a list. Easier than having to parse the semicolon ';'

rramkumar1: Done

pkg/ratelimit/ratelimit.go
rramkumar1: Done. In order to the keep the rateLimitImpls field as a map, the way I did it was kind of ugly :(

If I change that to a list, it can be a lot cleaner so let me know which way you prefer.

cmd/glbc/app/clients.go
G-Harmon: okay, thanks!

cmd/glbc/app/clients.go
```diff
@@ -102,6 +103,11 @@ func NewGCEClient() *gce.GCECloud {
 			// manually to re-create the client.
 			// TODO: why do we bail with success out if there is a permission error???
 			if _, err = cloud.ListGlobalBackendServices(); err == nil || utils.IsHTTPErrorCode(err, http.StatusForbidden) {
+				rl, err := ratelimit.ConfigureGCERateLimiting(flags.F.GCERateLimit.Values())
```
nicksardo: Can we reverse the conditions above move the ConfigureGCERateLimiting() call to the outer level. In other words, let's avoid nesting them.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,141 @@
...
+
+// ingressRateLimiter implements cloud.RateLimiter
+type ingressRateLimiter struct {
+	// Map a RateLimitKey to its rate limiter implementation.
+	rateLimitImpls map[*cloud.RateLimitKey]flowcontrol.RateLimiter
+}
+
+// ConfigureGCERateLimiting parses the value for the --gce-ratelimit
+// command line flag and returns a properly configured cloud.RateLimiter implementation.
+func ConfigureGCERateLimiting(specs []string) (*ingressRateLimiter, error) {
```
nicksardo: Might be more idiomatic to name it `NewGCERateLimiter` and have the struct be `GCERateLimiter`?

While this implementation exists in ingress, there's nothing ingress-specific about it, right? We could potentially move it to a more global library and use it in different places.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// ingressRateLimiter implements cloud.RateLimiter
+type ingressRateLimiter struct {
+	// Map a RateLimitKey to its rate limiter implementation.
+	rateLimitImpls map[*cloud.RateLimitKey]flowcontrol.RateLimiter
+}
+
+// ConfigureGCERateLimiting parses the value for the --gce-ratelimit
+// command line flag and returns a properly configured cloud.RateLimiter implementation.
```
nicksardo: I wouldn't mention the values coming in via flag. As far as this package knows, it's just a list of strings. You can give an example of the string format.

cmd/glbc/app/clients.go
rramkumar1: Done

pkg/ratelimit/ratelimit.go
rramkumar1: Done

pkg/ratelimit/ratelimit.go
rramkumar1: Done

PR #160, opened 2018-03-20
fix some grammar mistakes

jonyhy96:
fix some gramar mistakes.
```release-note
None
```

### DIFF
--- a/build/rules.mk
+++ b/build/rules.mk
@@ -77,7 +77,7 @@ else
 	VERBOSE_OUTPUT := >/dev/null
 endif
 
-# This MUST appear as the first rule in a Makefile
+# This MUST appears as the first rule in a Makefile
 all: build
 
 build-%:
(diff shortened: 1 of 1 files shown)


build/rules.mk
```diff
@@ -77,7 +77,7 @@ else
 	VERBOSE_OUTPUT := >/dev/null
 endif
 
-# This MUST appear as the first rule in a Makefile
+# This MUST appears as the first rule in a Makefile
```
rramkumar1: I think the way it was originally is correct...

jonyhy96: Thank you for your reply
I think the appear should in "The third person singular form"  
By the way,How do you think about the others mistakes on "Rule"

ihac: `appear` is correct, since it follows `must`, which is a modal verb~
[When do we not add -s to the third person singular in the simple present tense?](http://www.learnersdictionary.com/qa/third-person-singular-s-simple-present-tense)

jonyhy96: tahnks for correct me
it was my fault

jonyhy96:
some commits dont make sence

PR #161, opened 2018-03-20
Add Ingress HTTP2 feature gate

agau4779:
Feature gate for https://github.com/kubernetes/ingress-gce/pull/146


bowei:
preferable would be have a func in flags 

 ```
type Features struct {
  BackendHTTP2 bool
}
// DefaultFeatures can be stubbed out for unit testing.
var DefaultFeatures = &Features{
  BackendHTTP2: true,
}

func EnabledFeatures() *Features {
  return DefaultFeatures;
}
```

File: pkg/flags/flags.go [hunk 1]
```diff
@@ -60,6 +61,20 @@ var (
 func init() {
 	F.NodePortRanges.ports = []string{DefaultNodePortRange}
 	F.GCERateLimit.specs = []string{"alpha.Operations.Get,qps,10,100", "beta.Operations.Get,qps,10,100", "ga.Operations.Get,qps,10,100"}
+	F.Features = EnabledFeatures()
+}
+
+// Ingress feature flags
```
bowei: Features is

pkg/flags/flags.go
```diff
@@ -60,6 +61,20 @@ var (
...
+type Features struct {
+	Http2 bool
```
bowei: Http2 enables ...

PR #166, opened 2018-03-23
 reduce link jump

sunlintong:
I think it's more convenient.


spiffxp:
/assign
/close
@sunlintong the extra indirection is there on purpose, please see https://github.com/kubernetes/ingress-gce/pull/86 for details

PR #163, opened 2018-03-21
Add ServiceExtension CRD lifecycle management and empty spec definition

MrHohn:
Initial PR for ServiceExtension implementation. This hasn't touched any GLBC workflow. Just added CRD create/update logic and an empty spec definition & code-gen stuffs. New code is gated behind a flag.

NOTE: Code-gen scripts are derived from https://github.com/kubernetes/sample-controller. Longer-term we might want https://github.com/kubernetes-sigs/kubebuilder, but would like to get this in first.

P.S. `dep ensure -update` seems to pull a lot more things into vendor, not sure if I did it right. Dep version is v0.4.1.

@bowei @nicksardo @rramkumar1 @G-Harmon @nikhiljindal

PR #172, opened 2018-03-28
Bake ServiceExtension into GLBC workflow

MrHohn:
PR depends on https://github.com/kubernetes/ingress-gce/pull/163 (thinking of not distracting folks). Please only look at the last commit if you'd like to take a glance before that is merged.

This makes ServiceExtension create/update/delete trigger ingress sync. Also, the corresponding ServiceExtension will be save in `backends.ServicePort` if a service references it. This doesn't have actual impact on LB configuration yet.

Logic is gated behind a flag.

/hold


freehan:
I would recommend you just have a separate PR for vendoring.

MrHohn:
> I would recommend you just have a separate PR for vendoring.

Yeah I'd like to, though without updating the import in codes, dep might not pull in the new dependencies :/

MrHohn:
Closing this for now, the workflow modification are subject to change.

PR #175, opened 2018-03-28
behaviour --> behavior

sunlintong:
Maybe it's not an error, but it makes me not so comfortable.  :P


nicksardo:
/close
Covered by #177

PR #178, opened 2018-03-28
Updating FakeLoadBalancers.Delete to return NotFound when appropriate

nikhiljindal:
Ref https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/155

Fakes should return NotFound as the real code does.
This will help us catch issues like https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/155 in our unit tests.

cc @G-Harmon @nicksardo @bowei @MrHohn


nikhiljindal:
Updated to fix the tests

PR #183, opened 2018-03-29
Fix sync of multi-cluster ingress

### LINKED ISSUE #182: ingress controller should only manage instance groups for multicluster ingress
Forked from https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/131

ingress-gce controller is configuring the whole load balancer including creating url map, target proxy, etc for multicluster ingresses.
This is a regression. It should only manage instance groups.

Thanks to @MrHohn for finding the problem.
@nicksardo is working on a fix and I am going to send a PR to update the ingress-gce tests to catch this early.

Filing this issue to track

cc @csbell @G-Harmon @bowei

nicksardo:
Fixes #182
Also fixes an pre-existing issue of keeping around backend services if MCI and normal ingresses shared backend services. If the normal ingress was deleted, the backend service wouldn't be GC'd until MCI was deleted.

The logic for syncing MCI vs GCE-ingress was not easily distinguishable. I've re-ordered the sync logic and broke down `Checkpoint` into something I think is more clear. 
Checkpoint is broken up into three separate funcs and are called in the following order:
1. `EnsureInstanceGroupsAndPorts()`
    - MCI updates annotations and early returns.
2. `EnsureLoadBalancer()`
3. `EnsureFirewall()`

cc @nikhiljindal @csbell @bowei  @MrHohn  @rramkumar1

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -325,7 +325,7 @@ func (lbc *LoadBalancerController) sync(key string) (retErr error) {
 	negEndpointPorts := lbc.Translator.GatherEndpointPorts(gceNodePorts)
 	// Ensure firewall rule for the cluster and pass any NEG endpoint ports.
 	if err = lbc.CloudClusterManager.EnsureFirewall(nodeNames, negEndpointPorts); err != nil {
-		if fwErr, ok := err.(*firewalls.FirewallSyncError); ok {
+		if fwErr, ok := err.(*firewalls.FirewallXPNError); ok {
 			// XPN: Raise an event and ignore the error.
 			lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeNormal, "XPN", fwErr.Message)
 		} else {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -230,8 +230,8 @@ func (lbc *LoadBalancerController) Stop(deleteAll bool) error {
 	return nil
 }
 
-// sync manages Ingress create/updates/deletes.
-func (lbc *LoadBalancerController) sync(key string) (err error) {
+// sync manages Ingress create/updates/deletes
+func (lbc *LoadBalancerController) sync(key string) (retErr error) {
 	if !lbc.hasSynced() {
 		time.Sleep(storeSyncPollPeriod)
 		return fmt.Errorf("waiting for stores to sync")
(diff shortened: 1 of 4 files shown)


nikhiljindal:
Thanks for sending this @nicksardo 
Have added more e2e tests in https://github.com/kubernetes/kubernetes/pull/61909

pkg/controller/controller.go
```diff
@@ -272,62 +271,67 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
...
 			err = fmt.Errorf("error during sync %v, error during GC %v", syncError, deferErr)
 		}
 		glog.V(3).Infof("Finished syncing %v", key)
 	}()
 
-	singleIngressList := &extensions.IngressList{
-		Items: []extensions.Ingress{*ing},
+	igs, err := lbc.CloudClusterManager.EnsureInstanceGroupsAndPorts(nodeNames, allNodePorts)
+	if err != nil {
+		return err
```
nikhiljindal: Set syncError = err as at other places?

pkg/controller/controller.go
```diff
@@ -272,62 +271,67 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
...
+	}
+
+	if isGCEMultiClusterIngress(ing) {
+		// Add instance group names as annotation on the ingress and return.
+		if ing.Annotations == nil {
+			ing.Annotations = map[string]string{}
+		}
+		if err = setInstanceGroupsAnnotation(ing.Annotations, igs); err != nil {
+			return err
```
nikhiljindal: Same. set syncError = err?

pkg/controller/controller.go
```diff
@@ -272,62 +271,67 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
...
+		}
+		return updateAnnotations(lbc.client, ing.Name, ing.Namespace, ing.Annotations)
```
nikhiljindal: the returned error should be set as SyncError?

pkg/controller/controller.go
```diff
@@ -272,62 +271,67 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
 	// DeepCopy for assurance that we don't pollute other goroutines with changes.
 	ing = ing.DeepCopy()
 
-	// This performs a 2 phase checkpoint with the cloud:
-	// * Phase 1 creates/verifies resources are as expected. At the end of a
-	//   successful checkpoint we know that existing L7s are WAI, and the L7
-	//   for the Ingress associated with "key" is ready for a UrlMap update.
-	//   If this encounters an error, eg for quota reasons, we want to invoke
-	//   Phase 2 right away and retry checkpointing.
-	// * Phase 2 performs GC by refcounting shared resources. This needs to
```
nikhiljindal: fwiw, I think this comment was useful specially this explanation of why we always run phase 2, even when phase 1 (creating LB fails).
This helps understand why we have the defer func and tricky syncError tracking code.

nikhiljindal:
@nicksardo Were you able to find the PR that caused this regression? Will help in comparing how this was working before.

pkg/controller/controller.go
nicksardo: Yep. As I said, WIP.

nicksardo:
It was either the PR that stopped syncing multiple load balancers or stopped syncing multiple backend services. I don't think knowing which one caused the regression is important. You can examine the code at  v0.9.7 to see how MCI was separated from GCE-ingress. I think the division was too subtle and was bound to cause this break.

pkg/controller/controller.go
nicksardo: Will replace with a better comment elsewhere.

pkg/controller/controller.go
nicksardo: Deleted syncError

pkg/controller/controller.go
nikhiljindal: It should be retErr now?

pkg/controller/controller.go
```diff
@@ -247,89 +247,96 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
...
-		const eventMsg = "GCE"
-		if fwErr, ok := err.(*firewalls.FirewallSyncError); ok {
-			lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeNormal, eventMsg, fwErr.Message)
-		} else {
-			lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, eventMsg, err.Error())
-			syncError = err
-		}
+	// Create the backend services and higher-level LB resources.
+	if err = lbc.CloudClusterManager.EnsureLoadBalancer(lb, lbSvcPorts, igs); err != nil {
+		return err
```
nikhiljindal: set to retErr?

pkg/controller/controller.go
```diff
@@ -344,20 +351,24 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
 	// Update the UrlMap of the single loadbalancer that came through the watch.
 	l7, err := lbc.CloudClusterManager.l7Pool.Get(key)
 	if err != nil {
-		syncError = fmt.Errorf("%v, unable to get loadbalancer: %v", syncError, err)
-		return syncError
+		return fmt.Errorf("unable to get loadbalancer: %v", err)
```
nikhiljindal: This was being set to syncError earlier and hence we were generating event for this. Is the change to not generate event intentional?

nikhiljindal:
As discussed with @nicksardo, turns out I was understanding named returns wrong.
We were unnecessarily setting `syncErr = err` before, its not required :)

/lgtm

nicksardo:
There is one function difference in terms of syncError. Previously, if we failed "Checkpoint", it would continue on to update the URL map. I'm not sure that's behavior we want to keep. Now that it's only syncing GCP resources belonging to the ingress, updating the URL map seems pointless if there's an error earlier.

nikhiljindal:
sgtm

PR #194, opened 2018-04-06
Fix grammar mistake

mycapatin:
Fix grammar mistake


nicksardo:
Fixed by https://github.com/kubernetes/ingress-gce/pull/256

PR #199, opened 2018-04-10
Ensure only needed nodeports on instance group on ingress sync

MrHohn:
I might be missing something, but it seems `ensureInstanceGroupAndPorts()` is already doing what we need --- ensuring given ports are included in instance group, so the change is tiny?

@rramkumar1 @nicksardo


MrHohn:
Needs change.
/hold

MrHohn:
/hold cancel

Changed to get nodeports from the given ingress.

PR #200, opened 2018-04-10
Change naming of SSL certs

nicksardo:
Changes
 - Rename SSL certs from `k8s-ssl-{nameHash}--{clusterId}-{secretHash}` to `k8s-ssl-{nameHash}-{secretHash}--{clusterId}` to be consistent with other resources.
 - Get rid of `sslCertPrefix` for simplification.


nicksardo:
Thanks.

PR #207, opened 2018-04-11
Refactor translator.ToURLMap to not re-fetch backend services

rramkumar1:
/assign @nicksardo @MrHohn 

Will update tests once approach is LGTM'd.

### DIFF
--- a/pkg/loadbalancers/l7.go
+++ b/pkg/loadbalancers/l7.go
@@ -711,9 +711,9 @@ func (l *L7) UpdateUrlMap(ingressRules utils.GCEURLMap) error {
 	// backend, it applies to all host rules as well as to the urlmap itself.
 	// If it doesn't the urlmap might have a stale default, so replace it with
 	// glbc's default backend.
-	defaultBackend := ingressRules.GetDefaultBackend()
-	if defaultBackend != nil {
-		l.um.DefaultService = defaultBackend.SelfLink
+	defaultBackendName := ingressRules.GetDefaultBackendName()
+	if defaultBackendName != "" {
+		l.um.DefaultService = utils.BackendServiceRelativeResourcePath(defaultBackendName)
 	} else {
 		l.um.DefaultService = l.glbcDefaultBackend.SelfLink
 	}
@@ -726,7 +726,7 @@ func (l *L7) UpdateUrlMap(ingressRules utils.GCEURLMap) error {
 	l.um.HostRules = []*compute.HostRule{}
 	l.um.PathMatchers = []*compute.PathMatcher{}
 
-	for hostname, urlToBackend := range ingressRules {
+	for hostname, urlToBackendName := range ingressRules {
 		// Create a host rule
 		// Create a path matcher
 		// Add all given endpoint:backends to pathRules in path matcher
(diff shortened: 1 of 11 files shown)


pkg/utils/utils.go
```diff
@@ -135,3 +136,17 @@ func trimFieldsEvenly(max int, fields ...string) []string {
 
 	return ret
 }
+
+// BackendRelativeResourcePath returns a relative path of the link for a
+// BackendService given its name.
+func BackendRelativeResourcePath(name string) string {
```
nicksardo: Let's use BackendService to not get confused with BackendService.Backend

pkg/utils/utils.go
```diff
@@ -135,3 +136,17 @@ func trimFieldsEvenly(max int, fields ...string) []string {
...
+	return fmt.Sprintf("global/backendServices/%v", name)
+}
+
+// BackendComparableGroupPath trims project and compute version from the SelfLink
+// for a global BackendService.
+// /global/backendServices/[BACKEND_SERVICE_NAME]
+func BackendComparableGroupPath(url string) string {
```
nicksardo: Group was referring to InstanceGroup, which isn't applicable for the BackendService.
Maybe "BackendServiceComparablePath"

pkg/utils/utils.go
```diff
@@ -135,3 +136,17 @@ func trimFieldsEvenly(max int, fields ...string) []string {
...
+	path_parts := strings.Split(url, "/global/")
+	return fmt.Sprintf("/global/%s", path_parts[1])
```
nicksardo: We should handle the error case of len(url) != 2... Guessing we return empty.

pkg/utils/utils.go
rramkumar1: Done

pkg/utils/utils.go
rramkumar1: Done

pkg/utils/utils.go
rramkumar1: Done

rramkumar1:
Adding tests now.

pkg/loadbalancers/l7.go
```diff
@@ -824,7 +825,9 @@ func mapsEqual(a, b *compute.UrlMap) bool {
 					return false
 				}
 			}
-			if a.Service != b.Service {
+			// Trim down the url's for a.Service and b.Service to a comparable structure
+			// We do this because we update the UrlMap with relative links (not full) to backends.
+			if utils.BackendServiceComparablePath(a.Service) != utils.BackendServiceComparablePath(b.Service) {
```
nicksardo: There are two more lines with Service equality checks. 775 and 805.

pkg/utils/utils.go
```diff
@@ -135,3 +136,20 @@ func trimFieldsEvenly(max int, fields ...string) []string {
...
+
+// BackendServiceRelativeResourcePath returns a relative path of the link for a
+// BackendService given its name.
+func BackendServiceRelativeResourcePath(name string) string {
+	return fmt.Sprintf("global/backendServices/%v", name)
+}
+
+// BackendServiceComparablePath trims project and compute version from the SelfLink
+// for a global BackendService.
+// /global/backendServices/[BACKEND_SERVICE_NAME]
```
nicksardo: Nit: leading slash

pkg/loadbalancers/l7.go
rramkumar1: Done

pkg/utils/utils.go
rramkumar1: Done

PR #214, opened 2018-04-12
Code to setup removal of ServicePort logic out of translator 

rramkumar1:
A followup PR will modify the controller to use this code rather than the code in pkg/translator.


rramkumar1:
/assign @nicksardo 
For initial thoughts...

PR #217, opened 2018-04-13
Bootstrap multi-cluster controller

nicksardo:
All credit goes to @rramkumar1.


pkg/context/context.go
```diff
@@ -86,6 +92,9 @@ func (ctx *ControllerContext) HasSynced() bool {
 	if ctx.EndpointInformer != nil {
 		funcs = append(funcs, ctx.EndpointInformer.HasSynced)
 	}
+	if ctx.EndpointInformer != nil {
```
rramkumar1: Should be ClusterInformer :)

nicksardo: Nice catch!

PR #220, opened 2018-04-17
Typo fix in admin.md

AdamDang:
It's better to use same format for lb and LB
line 29: lb->LB


AdamDang:
no one care about this, closed.

PR #221, opened 2018-04-17
Update annotations.md

buzzedword:
This documentation mistakingly implies the prefix for `pre-shared-cert` is `ingress.kubernetes.io`, when it is actually `ingress.gcp.kubernetes.io`. Using the former will not create an HTTPS mapping in the GCLB, and does not emit any errors indicating there is an issue with the annotation.


buzzedword:
CLA has been signed.

nicksardo:
/lgtm
Thanks for your PR.

PR #222, opened 2018-04-17
Correct the returned message: does not exists->does not exist

AdamDang:
Correct the returned message


nicksardo:
Fixed by https://github.com/kubernetes/ingress-gce/pull/256

PR #226, opened 2018-04-18
Add support for logging latest commit hash of GLBC build being used

rramkumar1:
This will make it easier for us during debugging e2e test failures. Specifically, we now do not need to cross-reference with the latest e2e image push job run to find out the commit used in the test.

/assign @nicksardo


MrHohn:
Oh nice!

PR #235, opened 2018-04-19
Small addition to ClusterServiceMapper interface

rramkumar1:
This PR adds a new method to the interface. Specifically, now the list of expected services can be overwritten with a call to SetExpectedServices. 

/assign @nicksardo

### DIFF
--- a/pkg/mapper/mapper.go
+++ b/pkg/mapper/mapper.go
@@ -20,6 +20,7 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 	"k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/ingress-gce/pkg/utils"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/mapper/mapper.go
```diff
@@ -91,10 +87,28 @@ Loop:
...
 		return false
 	}
 	return true
 }
+
+func constructExpectedServices(expectedSvcs []string) map[string]bool {
+	es := make(map[string]bool)
+	if expectedSvcs == nil {
+		return es
+	}
```
nicksardo: This nil check isn't necessary.

nicksardo: This could be a generic helper func: `stringsToKeyMap`? If there's nothing already like it in utils, throw it in there?

rramkumar1: Done.

PR #239, opened 2018-04-25
BackendService Naming

agau4779:
- Use NEG naming schema for NEG Backends associated with NEG-enabled ServicePorts. Rely on garbage collection to remove Backends with the incorrect naming schema.
- Instead of md5 hash, use a sha256 hash of concatenated values for NEG name suffix
- Use BackendService name as the key in BackendPool.snapshotter

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -19,7 +19,6 @@ package backends
 import (
 	"fmt"
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
@@ -181,10 +180,6 @@ func (be *BackendService) ensureDescription(description string) (needsUpdate boo
 // Backends is a BackendPool.
 var _ BackendPool = (*Backends)(nil)
 
-func portKey(port int64) string {
-	return fmt.Sprintf("%d", port)
-}
-
 // ServicePort for tupling port and protocol
 type ServicePort struct {
 	SvcName       types.NamespacedName
@@ -222,12 +217,13 @@ func NewBackendPool(
 	healthChecker healthchecks.HealthChecker,
 	nodePool instances.NodePool,
 	namer *utils.Namer,
-	ignorePorts []int64,
+	ignorePorts []ServicePort,
 	resyncWithCloud bool) *Backends {
 
 	ignored := []string{}
 	for _, p := range ignorePorts {
-		ignored = append(ignored, portKey(p))
+		portName := namer.BackendNEG(p.SvcName.Namespace, p.SvcName.Name, p.SvcTargetPort)
+		ignored = append(ignored, portName)
 	}
 	backendPool := &Backends{
 		cloud:         cloud,
@@ -246,11 +242,7 @@ func NewBackendPool(
 		if !namer.NameBelongsToCluster(bs.Name) {
 			return "", fmt.Errorf("unrecognized name %v", bs.Name)
 		}
-		port, err := namer.BackendPort(bs.Name)
-		if err != nil {
-			return "", err
-		}
-		return port, nil
+		return bs.Name, nil
 	}
 	backendPool.snapshotter = storage.NewCloudListingPool("backends", keyFunc, backendPool, 30*time.Second)
 	return backendPool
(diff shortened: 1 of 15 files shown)


pkg/backends/backends.go
```diff
@@ -387,14 +397,20 @@ func (b *Backends) ensureBackendService(p ServicePort, igs []*compute.InstanceGr
...
 	if be == nil {
 		namedPort := &compute.NamedPort{
 			Name: b.namer.NamedPort(p.NodePort),
 			Port: p.NodePort,
 		}
-		glog.V(2).Infof("Creating backend service for port %v named port %v", p.NodePort, namedPort)
+
+		if p.NEGEnabled {
+			beName = b.namer.NEG(p.SvcName.Namespace, p.SvcName.Name, p.SvcTargetPort)
+		}
```
nicksardo: This seems to be calculated a few times in `ensureBackendService`. Move to top and use it everywhere?

pkg/backends/backends.go
```diff
@@ -262,10 +254,22 @@ func (b *Backends) Init(pp ProbeProvider) {
 }
 
 // Get returns a single backend.
-func (b *Backends) Get(port int64, isAlpha bool) (*BackendService, error) {
-	beGa, err := b.cloud.GetGlobalBackendService(b.namer.Backend(port))
-	if err != nil {
-		return nil, err
+func (b *Backends) Get(port ServicePort, isAlpha bool) (*BackendService, error) {
```
nicksardo: Maybe this function just takes the backend name?

pkg/utils/namer.go
```diff
@@ -223,6 +222,16 @@ func (n *Namer) ParseName(name string) *NameComponents {
 	}
 }
 
+// ParseNEGName parses the name of a resource generated by the namer.
+// This is only valid for NEGs and Backends on NEG.
+func (n *Namer) ParseNEGName(name string) *NameComponents {
+	l := strings.Split(name, "-")
+	return &NameComponents{
+		ClusterName: l[1],
```
nicksardo: ParseNegName should gracefully handle non-neg named things.

pkg/backends/backends_test.go
```diff
@@ -468,6 +468,109 @@ func TestBackendPoolSyncNEG(t *testing.T) {
...
+		}
+
+		if err := pool.Ensure(tc.oldPorts, nil); err != nil {
+			t.Errorf("Expected backend pool to add node ports, err: %v", err)
+		}
+
+		// Ensuring these ports again without first Garbage Collecting goes over
+		// the set quota. Expect an error here, until GC is called.
+		err := pool.Ensure(tc.newPorts, nil)
+		if tc.expectSyncErr && err == nil {
```
nicksardo: Doesn't capture the case of getting an error and not expecting an error.

```
if tc.expectSyncErr != (err != nil) {
```

pkg/backends/backends_test.go
```diff
@@ -468,6 +468,109 @@ func TestBackendPoolSyncNEG(t *testing.T) {
...
+			t.Errorf("Expect initial sync to go over quota, but received no error")
```
nicksardo: Nit: use standardized error messages
t.Errorf("Ensure(%v, nil) = %v, want err? %v", tc.newPorts, err, tc.expectSyncErr)

pkg/backends/backends_test.go
```diff
@@ -468,6 +468,109 @@ func TestBackendPoolSyncNEG(t *testing.T) {
...
+
+			if op == utils.Delete {
+				bsCreated -= 1
+			}
+
+			return nil
+		}
+
+		if err := pool.Ensure(tc.oldPorts, nil); err != nil {
+			t.Errorf("Expected backend pool to add node ports, err: %v", err)
```
nicksardo: nit: start using standard pattern
```
t.Errorf("Ensure(%v, nil) = %v, want nil", tc.oldPorts, err)
```

pkg/healthchecks/healthchecks.go
```diff
@@ -189,8 +189,7 @@ func (h *HealthChecks) getHealthCheckLink(port int64, alpha bool) (string, error
 }
 
 // Delete deletes the health check by port.
-func (h *HealthChecks) Delete(port int64) error {
-	name := h.namer.Backend(port)
+func (h *HealthChecks) Delete(name string) error {
```
nicksardo: Doesn't there need to be more work for the healthchecker to not use nodeport?

PR #243, opened 2018-04-27
Update gce.md

AdamDang:
secrete->secret
"secrete" and "secret" are very different meaning.


AdamDang:
failed to add cla, recreate this pr.

PR #247, opened 2018-05-03
Add util functions for backendConfig annotation

MrHohn:
/assign @nicksardo @bowei

### DIFF
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -40,6 +40,15 @@ const (
 	// 3. Adding this annotation on ingress.
 	NetworkEndpointGroupAlphaAnnotation = "alpha.cloud.google.com/load-balancer-neg"
 
+	// BackendConfigKey is a stringified JSON with two fields:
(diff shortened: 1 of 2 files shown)


nicksardo:
LGTM, will let bowei@ verify annotation value structure is as desired.

MrHohn:
Revised and added `ports` as a top level field per discussion with @bowei. PTAL, thanks!

pkg/annotations/service.go
```diff
@@ -40,6 +40,15 @@ const (
 	// 3. Adding this annotation on ingress.
 	NetworkEndpointGroupAlphaAnnotation = "alpha.cloud.google.com/load-balancer-neg"
 
+	// BackendConfigKey is a stringified JSON with two fields:
+	// - "ports": a map of port names to backendConfig names
```
nicksardo: port names or port numbers? Or just port names?

MrHohn: Thanks, changed to "port names or port numbers".

PR #253, opened 2018-05-08
Add tests for gceurlmap and extends GetDefaultBackendName slightly

rramkumar1:
For future use cases, we will need to perform non-destructive reads on the gceurlmap. This PR adds that capability and also adds tests for all functionality. 

/assign @nicksardo


rramkumar1:
/close
Discussed offline and will aim to express our internal UrlMap using a struct to make it more extensible.

PR #254, opened 2018-05-08
Refactor gceurlmap to a struct representation

rramkumar1:
Will fix tests once approach looks good.

/assign @nicksardo

### DIFF
--- a/pkg/controller/translator/translator.go
+++ b/pkg/controller/translator/translator.go
@@ -71,15 +71,14 @@ type GCE struct {
 	negEnabled     bool
 }
 
-// ToURLMap converts an ingress to a map of subdomain: url-regex: gce backend.
-func (t *GCE) ToURLMap(ing *extensions.Ingress) (utils.GCEURLMap, error) {
-	hostPathBackend := utils.GCEURLMap{}
+// ToURLMap converts an ingress to our internal UrlMap representation.
+func (t *GCE) ToURLMap(ing *extensions.Ingress) (*utils.GCEURLMap, error) {
+	urlMap := utils.NewGCEURLMap()
 	for _, rule := range ing.Spec.Rules {
 		if rule.HTTP == nil {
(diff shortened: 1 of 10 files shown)


rramkumar1:
The tests in controller_test.go need a really big refactor so I am commenting out the portions which cause the unit tests to fail for now. I would like to really dig into fixing those tests in a separate PR.

rramkumar1:
Decided to modify tests and fakes accordingly to make all tests work. Still want to dig deeper into controller_test.go later and fix it up.

pkg/controller/translator/translator.go
```diff
@@ -71,15 +71,17 @@ type GCE struct {
...
+// ToURLMap converts an ingress to our internal UrlMap representation.
+func (t *GCE) ToURLMap(ing *extensions.Ingress) (*utils.GCEURLMap, error) {
+	urlMap := utils.NewGCEURLMap()
 	for _, rule := range ing.Spec.Rules {
 		if rule.HTTP == nil {
-			glog.Errorf("Ignoring non http Ingress rule")
 			continue
 		}
-		pathToBackend := map[string]string{}
+		// Using a mapping of regex -> backend here ensures that we don't
```
nicksardo: Let's not use the word `regex` as this confuses users. Stick with path?

pkg/utils/gceurlmap.go
```diff
@@ -18,45 +15,93 @@ package utils
...
+type GCEURLMap struct {
+	// DefaultBackendName is the k8s name given to the default backend.
+	DefaultBackendName string
+	// hostRules is a mapping from hostnames to path rules for that host.
+	hostRules map[string][]PathRule
+}
+
+// PathRule encapsulates the information for a single path -> backend mapping.
+type PathRule struct {
+	// Path is a regex.
```
nicksardo: sans regex

PR #257, opened 2018-05-09
 Merge the logic of ToUrlMap() and IngressToNodePorts() [WIP]

rramkumar1:
This PR attempts to merge the logic of two function in the translator. Right now, ToUrlMap() and IngressToNodePorts() both iterate over an Ingress spec in very similar ways. In order to reduce complexity of having two similar code paths being executed at different times, we can merge these two functions together to ensure easier testing and maintainability. 

This also gives us the opportunity to store all config in one place (GCEURLMap). Now, this struct uses a ServicePort to denote a backend rather than the backend's name. (credit to @nicksardo  for this idea)

If this approach looks reasonable, I will add tests for it (there are no tests for any of this currently). 

This PR is a prerequisite for condensing backend pool's (#242).

/assign @nicksardo

### DIFF
--- a/cmd/glbc/app/init.go
+++ b/cmd/glbc/app/init.go
@@ -30,11 +30,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"k8s.io/ingress-gce/pkg/annotations"
-	"k8s.io/ingress-gce/pkg/backends"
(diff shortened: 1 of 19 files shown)


rramkumar1:
Note: A bulk of the changes are to test code.

nicksardo:
Thanks for making the changes.
/lgtm

PR #258, opened 2018-05-09
Use leader election to prevent multiple controllers running

nicksardo:
Prior to starting the ingress or neg controllers, the binary will now wait to acquire a resource lock.


cmd/glbc/app/handlers.go
```diff
@@ -33,22 +33,19 @@ import (
 	"k8s.io/ingress-gce/pkg/version"
 )
 
-func RunHTTPServer(lbc *controller.LoadBalancerController) {
+func RunHTTPServer(healthcheck func() error) {
```
bowei: Can you document what the contract is for healthcheck?

cmd/glbc/main.go
```diff
@@ -89,40 +94,98 @@ func main() {
...
 
-	if clusterManager.ClusterNamer.UID() != "" {
-		glog.V(0).Infof("Cluster name is %+v", clusterManager.ClusterNamer.UID())
-	}
-	clusterManager.Init(lbc.Translator, lbc.Translator)
-	glog.V(0).Infof("clusterManager initialized")
+	healthCheck := func() error { return nil }
+	go app.RunHTTPServer(healthCheck)
+
+	run := func() {
```
bowei: make this a real method, unless you are heavily using variable capture

cmd/glbc/main.go
```diff
@@ -89,40 +94,98 @@ func main() {
...
+		for {
+			glog.Infof("Handled quit, awaiting pod deletion.")
+			time.Sleep(30 * time.Second)
+		}
 	}
 
-	go app.RunHTTPServer(lbc)
-	go app.RunSIGTERMHandler(lbc, flags.F.DeleteAllOnQuit)
+	if !flags.F.LeaderElection.Enabled {
+		run()
```
bowei: return

cmd/glbc/main.go
```diff
@@ -89,40 +94,98 @@ func main() {
...
-		time.Sleep(30 * time.Second)
+// makeLeaderElectionConfig builds a leader election configuration. It will
+// create a new resource lock associated with the configuration.
+func makeLeaderElectionConfig(client clientset.Interface, recorder record.EventRecorder, run func()) (*leaderelection.LeaderElectionConfig, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get hostname: %v", err)
 	}
+	// add a uniquifier so that two processes on the same host don't accidentally both become active
+	id := fmt.Sprintf("%v_%x", hostname, rand.Intn(1e6))
```
bowei: why not just math.rand.Int64()? Also, why do we have to use the k8s library?

pkg/flags/flags.go
```diff
@@ -33,6 +34,19 @@ const (
...
+	// DefaultRenewDeadline is a 10 second renewal deadline.
+	DefaultRenewDeadline = 10 * time.Second
+	// DefaultRetryPeriod is a 2 second retry period.
+	DefaultRetryPeriod = 2 * time.Second
+
+	// DefaultLockObjectNamespace is the namespace which owns the lock object.
+	DefaultLockObjectNamespace string = "kube-system"
+
+	// DefaultLockObjectName is the object name of the lock object.
+	DefaultLockObjectName = "ingress-gce"
```
bowei: ingress-gce-lock? it may be a possibility to use kube-system:ingress-gce for configuration

pkg/flags/flags.go
```diff
@@ -56,13 +70,58 @@ var (
...
+
+		LeaderElection LeaderElectionConfiguration
 	}{}
 )
 
+type LeaderElectionConfiguration struct {
+	// Enabled enables a leader election client to gain leadership
+	// before executing the main loop. Enable this when running replicated
+	// components for high availability.
+	Enabled bool
```
bowei: Isn't this already a command line flag?

pkg/flags/flags.go
```diff
@@ -56,13 +70,58 @@ var (
...
+	// RenewDeadline is the interval between attempts by the acting master to
+	// renew a leadership slot before it stops leading. This must be less
+	// than or equal to the lease duration. This is only applicable if leader
+	// election is enabled.
+	RenewDeadline time.Duration
+	// RetryPeriod is the duration the clients should wait between attempting
+	// acquisition and renewal of a leadership. This is only applicable if
+	// leader election is enabled.
+	RetryPeriod time.Duration
+	// ResourceLock indicates the resource object type that will be used to lock
```
bowei: ResourceLock => ResourceType?

pkg/flags/flags.go
bowei: i see it, ignore comment

cmd/glbc/main.go
nicksardo: Done

cmd/glbc/app/handlers.go
```diff
@@ -72,6 +64,36 @@ func RunSIGTERMHandler(lbc *controller.LoadBalancerController, deleteAll bool) {
...
 
+func healthCheckHandler(checker func() context.HealthCheckResults) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var hasErr bool
+		var s strings.Builder
+		for component, result := range checker() {
+			status := "OK"
+			if result != nil {
+				hasErr = true
+				status = fmt.Sprintf("err: %v", result)
```
bowei: would be easier for the consumer if this was JSON, but let's take it up in a follow up PR

cmd/glbc/main.go
```diff
@@ -93,22 +100,78 @@ func main() {
...
+				glog.Fatalf("lost master")
+			},
+		},
+	}, nil
+}
+
+func runControllers(ctx *context.ControllerContext, cloud *gce.GCECloud) {
+	namer, err := app.NewNamer(ctx.KubeClient, flags.F.ClusterName, controller.DefaultFirewallName)
+	if err != nil {
+		glog.Fatalf("%v", err)
```
bowei: probably want to put more explication here

glog.Fatalf("app.NewNamer(ctx.KubeClient, %q, %q) = %v", ...)

pkg/flags/flags.go
```diff
@@ -58,13 +77,40 @@ var (
...
+}
+
+func defaultLeaderElectionConfiguration() LeaderElectionConfiguration {
+	return LeaderElectionConfiguration{
+		LeaderElectionConfiguration: componentconfig.LeaderElectionConfiguration{
+			LeaderElect:   true,
+			LeaseDuration: metav1.Duration{Duration: DefaultLeaseDuration},
+			RenewDeadline: metav1.Duration{Duration: DefaultRenewDeadline},
+			RetryPeriod:   metav1.Duration{Duration: DefaultRetryPeriod},
+			ResourceLock:  resourcelock.ConfigMapsResourceLock,
```
nicksardo: Making this default a constant

Issue #241, opened 2018-04-26
Failing to pick up health check from readiness probe


sonu27:
When I create a GCE ingress, Google Load Balancer does not set the health check from the readiness probe. According to the docs ([Ingress GCE health checks][1]) it should pick it up.

> Expose an arbitrary URL as a readiness probe on the pods backing the Service.

Any ideas why?

Deployment:

    apiVersion: extensions/v1beta1
    kind: Deployment
    metadata:
      name: frontend-prod
      labels:
        app: frontend-prod
    spec:
      selector:
        matchLabels:
          app: frontend-prod
      replicas: 3
      strategy:
        rollingUpdate:
          maxSurge: 1
          maxUnavailable: 1
        type: RollingUpdate
      template:
        metadata:
          labels:
            app: frontend-prod
        spec:
          imagePullSecrets:
            - name: regcred
          containers:
          - image: app:latest
            readinessProbe:
              httpGet:
                path: /healthcheck
                port: 3000
              initialDelaySeconds: 15
              periodSeconds: 5
            name: frontend-prod-app
          - env:
            - name: PASSWORD_PROTECT
              value: "1"
            image: nginx:latest
            readinessProbe:
              httpGet:
                path: /health
                port: 80
              initialDelaySeconds: 5
              periodSeconds: 5
            name: frontend-prod-nginx

---

Service:

    apiVersion: v1
    kind: Service
    metadata:
      name: frontend-prod
      labels:
        app: frontend-prod
    spec:
      type: NodePort
      ports:
      - port: 80
        targetPort: 80
        protocol: TCP
        name: http
      selector:
        app: frontend-prod

---

Ingress:

    apiVersion: extensions/v1beta1
    kind: Ingress
    metadata:
      name: frontend-prod-ingress
      annotations:
        kubernetes.io/ingress.global-static-ip-name: frontend-prod-ip
    spec:
      tls:
        - secretName: testsecret
      backend:
        serviceName: frontend-prod
        servicePort: 80

  [1]: https://github.com/kubernetes/ingress-gce#health-checks

nicksardo:
There are several caveats. The health check should not already exist as it won't overwrite settings. Furthermore, the pods need to exist at the time of ingress creation.

sonu27:
@nicksardo Yes I know that it will not overwrite settings of the ingress.

[...]

PR #260, opened 2018-05-10
Update healthcheck docs regarding containerPort

sonu27:
Fixes issue #241

### DIFF
--- a/examples/health-checks/README.md
+++ b/examples/health-checks/README.md
@@ -66,6 +66,8 @@ You can confirm the health check endpoint point it's using one of 2 ways:
 ## Limitations
 
 A few points to note:
+* The pod's `containerPort` field must be defined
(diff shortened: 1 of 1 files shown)


sonu27:
@nicksardo Let me know if this is clear enough :)

nicksardo:
Thanks for creating this PR @sonu27.  However, it's not matching the servicePort, it's matching the service's targetPort.  It's also complicated because names can be used...

```
The pod's `containerPort` field must be defined.
The service's `targetPort` field must point to the pod port's `containerPort` value or `name`. Note that the `targetPort` defaults to the `port` value if not defined.
```
Feel free to re-phrase any of that. 

This might make more sense
Ingress `servicePort` ▶ Service port.`port` || port.`name`
Service `targetPort` ▶ Pod port.`containerPort` || port.`name`
ReadinessProbe `port` ▶ Pod port.`containerPort` ||  port.`name`

sonu27:
Your wording is good.

I was thinking of added the bottom bit, as it explains it cleanly but possibly too much :)

PR #262, opened 2018-05-10
BackendService naming for NEG backend services & healthchecks

nicksardo:
Continuation of https://github.com/kubernetes/ingress-gce/pull/239

More changes:
 - Neg suffix hash uses the UID truncated to 8 chars.
 - Neg suffix hash input fields are separated by a delimiter
 - ServicePort now has an extension function to return the correct name. 
 - `namer.Backend` is renamed to `namer.IGBackend` to prevent confusion between IG/NEG.

TODO in another PR:
Rewrite unit tests to not care what kind of serviceport their given.

PR #264, opened 2018-05-11
Condense health checkers into one health checker for all backends.

rramkumar1:
This PR should wait on #262

TODOs before merge:

1. Manually test case of ingress using system default backend w/ other backends and verify health checks.
2. Verify that users can still modify health check settings for all backends and that the controller will not reconcile. 

/assign @nicksardo

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -90,7 +90,7 @@ func main() {
 
 	cloud := app.NewGCEClient()
 	defaultBackendServicePort := app.DefaultBackendServicePort(kubeClient)
-	clusterManager, err := controller.NewClusterManager(cloud, namer, flags.F.HealthCheckPath)
+	clusterManager, err := controller.NewClusterManager(cloud, namer, *defaultBackendServicePort, flags.F.HealthCheckPath, flags.F.DefaultSvcHealthCheckPath)
 	if err != nil {
 		glog.Fatalf("Error creating cluster manager: %v", err)
 	}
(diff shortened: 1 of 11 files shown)


cmd/glbc/main.go
```diff
@@ -90,15 +90,15 @@ func main() {
 
 	cloud := app.NewGCEClient()
 	defaultBackendServicePort := app.DefaultBackendServicePort(kubeClient)
-	clusterManager, err := controller.NewClusterManager(cloud, namer, flags.F.HealthCheckPath)
+	clusterManager, err := controller.NewClusterManager(cloud, namer, *defaultBackendServicePort, flags.F.HealthCheckPath)
```
nicksardo: I suggest we get rid of the `app.DefaultBackendServicePort` call. We don't really care that the service exists until an ingress sync. Using the client to fetch it now seems racey for new clusters. 

Maybe create a utility function that converts strings of "namespace/name" to types.NamespacedName and use that to convert the flag value. Then pass that value to the LBC?

pkg/healthchecks/healthchecks.go
```diff
@@ -32,6 +33,9 @@ import (
 )
 
 const (
+	// DefaultBackendHealthCheckPath defines the path that the system default
+	// backend should serve 200's on.
+	DefaultBackendHealthCheckPath = "/healthz"
```
nicksardo: Something tells me that we should make this a flag that defaults to '/healthz'.  Since the default backend is modifiable, this should probably be too.

cmd/glbc/main.go
rramkumar1: Good point, can I do this in a follow up PR? It's slightly more involved than just removing that call so I want to spend more time thinking about it.

pkg/healthchecks/healthchecks.go
rramkumar1: Done

PR #270, opened 2018-05-15
Pass service/port tuple separate from ServicePort

nicksardo:
Changes:
- Users can now configure the default backend service **port** by flag - defaults to "http".
- Controller checks that the default backend service exists and has the specified port name.
- Now only the ServicePortID is passed to the LBC - the nodeport is no longer cached.
- TranslateIngress may now return an error. This will occur if the default backend service does not exist at sync-time.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -612,11 +612,11 @@ func (b *Backends) Status(name string) string {
 	return hs.HealthStatus[0].HealthState
 }
 
-func (b *Backends) Link(port utils.ServicePort, zones []string) error {
-	if !port.NEGEnabled {
+func (b *Backends) Link(sp utils.ServicePort, zones []string) error {
+	if !sp.NEGEnabled {
 		return nil
 	}
-	negName := b.namer.NEG(port.SvcName.Namespace, port.SvcName.Name, port.SvcTargetPort)
+	negName := b.namer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.SvcTargetPort)
 	var negs []*computealpha.NetworkEndpointGroup
 	var err error
 	for _, zone := range zones {
@@ -627,7 +627,8 @@ func (b *Backends) Link(port utils.ServicePort, zones []string) error {
 		negs = append(negs, neg)
 	}
 
-	backendService, err := b.cloud.GetAlphaGlobalBackendService(negName)
+	beName := sp.BackendName(b.namer)
+	backendService, err := b.cloud.GetAlphaGlobalBackendService(beName)
 	if err != nil {
 		return err
 	}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -283,7 +283,11 @@ func (lbc *LoadBalancerController) sync(key string) (retErr error) {
 }
 
 func (lbc *LoadBalancerController) ensureIngress(key string, ing *extensions.Ingress, nodeNames []string, gceSvcPorts []utils.ServicePort) error {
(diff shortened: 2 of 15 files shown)


pkg/backends/backends.go
```diff
@@ -612,11 +612,11 @@ func (b *Backends) Status(name string) string {
...
 }
 
-func (b *Backends) Link(port utils.ServicePort, zones []string) error {
-	if !port.NEGEnabled {
+func (b *Backends) Link(sp utils.ServicePort, zones []string) error {
+	if !sp.NEGEnabled {
 		return nil
 	}
-	negName := b.namer.NEG(port.SvcName.Namespace, port.SvcName.Name, port.SvcTargetPort)
+	negName := b.namer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.SvcTargetPort)
```
rramkumar1: Can we use the new BackendName function here?

pkg/controller/controller.go
```diff
@@ -283,7 +283,11 @@ func (lbc *LoadBalancerController) sync(key string) (retErr error) {
 }
 
 func (lbc *LoadBalancerController) ensureIngress(key string, ing *extensions.Ingress, nodeNames []string, gceSvcPorts []utils.ServicePort) error {
-	urlMap := lbc.Translator.TranslateIngress(ing, lbc.CloudClusterManager.defaultBackendSvcPort)
+	urlMap, err := lbc.Translator.TranslateIngress(ing, lbc.CloudClusterManager.defaultBackendSvcPortID)
+	if err != nil {
+		return fmt.Errorf("error while evaluating the ingress spec: %v", err)
```
rramkumar1: Nice!

pkg/flags/flags.go
```diff
@@ -103,8 +104,10 @@ controller only works with default zones.`)
 must serve a 200 page on this path.`)
```
rramkumar1: I would maybe add one line that says that default-backend-service and default-backend-service-port are mandatory.

pkg/utils/serviceport.go
```diff
@@ -54,5 +61,16 @@ func (sp ServicePort) BackendName(namer *Namer) string {
 		return namer.IGBackend(sp.NodePort)
 	}
 
-	return namer.NEG(sp.SvcName.Namespace, sp.SvcName.Name, sp.SvcTargetPort)
+	return namer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.SvcTargetPort)
+}
+
+// BackendToServicePortID creates a ServicePortID struct from an IngressBackend struct and namespace.
```
rramkumar1: Nit: Remove the word struct from the comment.

pkg/utils/utils.go
```diff
@@ -62,6 +63,18 @@ func IsHTTPErrorCode(err error, code int) bool {
 	return ok && apiErr.Code == code
 }
 
+// ToNamespacedName returns a types.NamespacedName struct parsed from namespace/name.
+func ToNamespacedName(s string) (r types.NamespacedName, err error) {
```
rramkumar1: Tiny test for this?

pkg/backends/backends.go
nicksardo: So this name is used for both the NEG and the BackendService. I'll be redundantly clear and use `BackendName` for when we get the backendservice; however, I think this line is still appropriate for the NEG.

pkg/utils/utils.go
nicksardo: Done

nicksardo:
Fixed a small typo in the test.

PR #275, opened 2018-05-18
TranslateIngress changes

nicksardo:
**Changes**
 - Stop gracefully falling back on system default backend if specified backend is not found - no longer need EventRecorder in translater
 - Return list of errors from TranslateIngress
 - ensureIngress will error if TranslateIngress hits any errors. `ToSvcPorts` iterates all ingresses (for now) and will continue to collect all known service ports
 - Deleted a lot of tests in controller_test.go. Some of them should be re-created in different packages (tests that check GCE resource config), and new tests should be made for the controller that assert controller logic such as garbage collection, k8s object lifecycle edge cases.
 - Re-wrote tests for controller and created tests for TranslateIngress.

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2015 The Kubernetes Authors.
+Copyright 2018 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


pkg/utils/utils.go
```diff
@@ -150,6 +152,19 @@ func trimFieldsEvenly(max int, fields ...string) []string {
 	return ret
 }
 
+// PrettyJson marshals an object in a human-friendly format.
+func PrettyJson(data interface{}) (string, error) {
```
rramkumar1: Maybe I'm missing something, but this doesn't seem to be used anywhere.

pkg/utils/gceurlmap.go
```diff
@@ -39,7 +39,41 @@ type PathRule struct {
 
 // NewGCEURLMap returns an empty GCEURLMap
 func NewGCEURLMap() *GCEURLMap {
-	return &GCEURLMap{hostRules: make(map[string][]PathRule)}
+	return &GCEURLMap{HostRules: make(map[string][]PathRule)}
+}
+
+func EqualStructure(a, b *GCEURLMap) bool {
```
rramkumar1: Can we call this Compare() instead?

pkg/test/utils.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Backend returns an IngressBackend with the given service name/port.
+func Backend(name string, port intstr.IntOrString) *extensions.IngressBackend {
+	return &extensions.IngressBackend{
+		ServiceName: name,
+		ServicePort: port,
+	}
+}
+
+// DecodeIngress returns deserializes an Ingress from yaml.
```
rramkumar1: remove "returns" from comment.

pkg/utils/gceurlmap.go
```diff
@@ -26,9 +26,9 @@ import (
 //       2. All paths for a specific host are unique.
 //       3. Adding paths for a hostname replaces existing for that host.
 type GCEURLMap struct {
-	DefaultBackend ServicePort
+	DefaultBackend *ServicePort
 	// hostRules is a mapping from hostnames to path rules for that host.
-	hostRules map[string][]PathRule
+	HostRules map[string][]PathRule
```
rramkumar1: Any reason for exposing the host rules? I left them package private so that we never access it without using the public getter

pkg/utils/gceurlmap.go
```diff
@@ -49,54 +83,71 @@ func NewGCEURLMap() *GCEURLMap {
 func (g *GCEURLMap) PutPathRulesForHost(hostname string, pathRules []PathRule) {
 	// Convert the path rules to a map to filter out two equal paths
 	// Note that the if two paths are equal, the one later in the list is the winner.
-	uniquePaths := make(map[string]PathRule)
-	for _, pathRule := range pathRules {
-		if _, ok := uniquePaths[pathRule.Path]; ok {
-			glog.V(4).Infof("Equal paths (%v) for host %v. Using backend %+v", pathRule.Path, hostname, pathRule.Backend)
+	seen := make(map[string]struct{})
```
rramkumar1: can the value for this just be a bool? Might read a little better.

pkg/controller/translator/translator_test.go
```diff
@@ -68,6 +63,127 @@ func fakeTranslator(negEnabled bool) *Translator {
 	return gce
 }
 
+func ingressFromFile(filename string) *extensions.Ingress {
```
rramkumar1: Nit: Can we move these helpers to the bottom of the file? I don't like them being the first thing I see when I open the file :)

pkg/controller/translator/translator_test.go
```diff
@@ -68,6 +63,127 @@ func fakeTranslator(negEnabled bool) *Translator {
...
+		log.Fatal(err)
+	}
+	v := &utils.GCEURLMap{}
+	if err := json.Unmarshal(data, v); err != nil {
+		log.Fatal(err)
+	}
+	return v
+}
+
+func TestTranslateIngress(t *testing.T) {
```
rramkumar1: Nice job on this test, looks awesome. One thing I would like to see is maybe a case that results in 2 or more errors.

pkg/controller/controller_test.go
```diff
@@ -163,345 +100,99 @@ func getKey(ing *extensions.Ingress, t *testing.T) string {
...
-		}
-	}
-	lbc.client.Extensions().Ingresses(ing.Namespace).Create(ing)
-	lbc.ctx.IngressInformer.GetIndexer().Add(ing)
-}
-
-func TestLbCreateDelete(t *testing.T) {
-	testFirewallName := "quux"
-	cm := NewFakeClusterManager(flags.DefaultClusterUID, testFirewallName)
+func TestIngressSyncError(t *testing.T) {
```
rramkumar1: Can we add a small comment to this test that is a little more descriptive than the function name itself? Same for other tests as well.

pkg/utils/utils.go
nicksardo: Yea, but is a useful func to exist when we need to add more tests.

pkg/utils/gceurlmap.go
nicksardo: It's only checking ServiceProjectID and `Compare` makes it sound like it's a full ServicePort check.

pkg/test/utils.go
nicksardo: Thanks

pkg/utils/gceurlmap.go
nicksardo: Didn't have an immediate need. Was a little confused why `AllRules()` is used over `HostRules`. I have no strong opinion on this... will make it private

nicksardo: Ah, actually, we need exported for serializing/deserializing into JSON.

rramkumar1: Damn, okay. Let's get rid of AllRules() then too and just replace that with direct access.

pkg/controller/translator/translator_test.go
nicksardo: SG

pkg/utils/gceurlmap.go
rramkumar1: Damn, okay. Let's get rid of AllRules() then too and just replace that with direct access.

nicksardo: Done

PR #276, opened 2018-05-18
fake backendservices save alpha objects by default

agau4779:
- Have FakeBackendServices save Alpha BackendServices by default. GA methods convert the object to alpha version. This is to avoid converting an Alpha object to GA, and potentially losing data present on fields that are only in Alpha but not GA.
- Use `alphaErrFunc` to throw error if not alpha whitelisted (currently only used for HTTP/2)

### DIFF
--- a/pkg/backends/backends_test.go
+++ b/pkg/backends/backends_test.go
@@ -56,7 +56,13 @@ var (
 			},
 		},
 	}
-	noOpErrFunc = func(op int, be *compute.BackendService) error { return nil }
+	noOpErrFunc  = func(op int, be *computealpha.BackendService) error { return nil }
+	alphaErrFunc = func(op int, be *computealpha.BackendService) error {
+		if be.Protocol == string(annotations.ProtocolHTTP2) {
+			return utils.FakeGoogleAPIForbiddenErr()
+		}
+		return nil
+	}
 )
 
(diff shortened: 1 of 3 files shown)


pkg/backends/backends_test.go
```diff
@@ -891,13 +897,13 @@ func TestComparableGroupPath(t *testing.T) {
...
 func TestEnsureBackendServiceProtocol(t *testing.T) {
-	f := NewFakeBackendServices(noOpErrFunc, true)
+	f := NewFakeBackendServices(noOpErrFunc)
 	fakeIGs := instances.NewFakeInstanceGroups(sets.NewString(), defaultNamer)
 	pool, _ := newTestJig(f, fakeIGs, false)
 
 	svcPorts := []utils.ServicePort{
-		{NodePort: 80, Protocol: annotations.ProtocolHTTP, ID: utils.ServicePortID{Port: intstr.FromInt(1)}},
-		{NodePort: 443, Protocol: annotations.ProtocolHTTPS, ID: utils.ServicePortID{Port: intstr.FromInt(2)}},
+		// {NodePort: 80, Protocol: annotations.ProtocolHTTP, ID: utils.ServicePortID{Port: intstr.FromInt(1)}},
```
nicksardo: Ping

agau4779: re-enabled them :zap:

PR #277, opened 2018-05-19
Update gce provider in vendor

MrHohn:
Update kubernetes/kubernetes to https://github.com/kubernetes/kubernetes/commit/77a08ee2d774d1785670677d4d4b7d1a1739c12c.

Also updated the signature for `SetSslCertificateForTargetHttpsProxy()` (https://github.com/kubernetes/kubernetes/commit/f9d1f7eb75c3696c75e7e42848de34d79b3253b8).

/assign @nicksardo


nicksardo:
When was the last time you updated the dep binary? The PR is adding back lots of markdown and some shell files. I was under impression those would be pruned out during ensure...

MrHohn:
> The PR is adding back lots of markdown and some shell files.

Good catch, I never update the dep binary :) Will fix.

MrHohn:
Seem cleaner now :)

PR #280, opened 2018-05-21
Refactor pkg/backends to use new BackendService composite type 

rramkumar1:
This PR introduces a new mechanism for operating on the compute BackendService API's. Previously, we maintained an internal "BackendService" type which contained the two actual compute types (alpha and GA) we needed. This resulted in numerous "if" statements and duplicated code just for one or the other compute type. 

With this PR, we now have a composite "BackendService" which encapsulates all the fields in the compute alpha type. Now, any field mutations can go directly to this composite type and helper methods will deal with converting the composite into the appropriate type before sending the API call off the GCE.

As a longer term TODO, all the code in compute.go and compute_test.go can be generated .


pkg/backends/backends.go
```diff
@@ -342,17 +267,17 @@ func (b *Backends) ensureBackendService(sp utils.ServicePort, igLinks []string)
 		}
 	}
 
-	needUpdate := be.ensureProtocol(sp)
-	needUpdate = needUpdate || be.ensureHealthCheckLink(hcLink)
-	needUpdate = needUpdate || be.ensureDescription(sp.Description())
+	needUpdate := ensureProtocol(be, sp)
+	needUpdate = needUpdate || ensureHealthCheckLink(be, hcLink)
+	needUpdate = needUpdate || ensureDescription(be, sp.Description())
```
nicksardo: While we're here, can we condense these checks down into a single call.

pkg/backends/compute.go
```diff
@@ -0,0 +1,224 @@
...
+// TODO(rramkumar): All code in this file should ideally be generated.
+
+// CompositeBackendServiceHandler provides methods to:
+//	1. persist a compute BackendService given its composite representation
+//	2. get a composite type given a compute BackendService name.
+type CompositeBackendServiceHandler struct {
+	cloud BackendServices
+}
+
+func (h *CompositeBackendServiceHandler) create(be *BackendService, isAlpha bool) error {
```
nicksardo: Noting what we discussed. Pass a version instead of isAlpha bool. Maybe use meta.Version
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/cloud/meta/meta.go#L28

pkg/backends/backends.go
rramkumar1: Done

MrHohn:
cc @bowei

pkg/backends/backends.go
```diff
@@ -328,31 +256,34 @@ func (b *Backends) ensureBackendService(sp utils.ServicePort, igLinks []string)
...
 
 	// Verify existance of a backend service for the proper port, but do not specify any backends/igs
-	be, _ = b.Get(beName, sp.IsAlpha())
-	if be == nil {
-		namedPort := &compute.NamedPort{
-			Name: b.namer.NamedPort(sp.NodePort),
-			Port: sp.NodePort,
-		}
+	be, getErr := b.Get(beName, sp.Version())
+	if getErr != nil {
```
bowei: the nested if's here is kind of hard to understand, lets do (put the create logic in its own func):
```go
be, err := ...
if err != nil {
  if !utils.IsNotFoundErr(err) {
    return err
  }
  if err := createBackendService(...); err != nil {
    return err
  }
}
```

pkg/backends/backends.go
```diff
@@ -379,23 +310,12 @@ func (b *Backends) edgeHop(be *BackendService, igLinks []string) error {
...
-			// Backend service is not able to point to NEG and IG at the same time.
-			// Filter IG backends here.
-			if strings.Contains(backend.Group, "instanceGroups") {
-				originalGaIGBackends = append(originalGaIGBackends, backend)
-			}
+	originalIGBackends := []*Backend{}
+	for _, backend := range be.Backends {
+		// Backend service is not able to point to NEG and IG at the same time.
+		// Filter IG backends here.
+		if strings.Contains(backend.Group, "instanceGroups") {
```
bowei: what is this string? shouldn't we be looking for "/instanceGroups/" or something?

pkg/backends/compute.go
```diff
@@ -0,0 +1,238 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
bowei: 2018

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
bowei: 2018

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+		if err := compareFields(compositeType.Field(i), alphaType.Field(i-1)); err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestBackend(t *testing.T) {
+	compositeType := reflect.TypeOf(Backend{})
+	alphaType := reflect.TypeOf(computealpha.Backend{})
+	typeEquality(compositeType, alphaType)
```
bowei: don't you need to check the return value from typeEquality?

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func TestToAlpha(t *testing.T) {
+	composite := BackendService{
+		Version:  meta.VersionAlpha,
+		Protocol: "HTTP2",
+	}
+	alpha, err := composite.toAlpha()
+	if err != nil {
+		t.Fatalf("Did not expect error when calling toAlpha(): %v", err)
```
bowei: t.Fatalf("composite.toAlpha() =_, %v; want _, nil", err)

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+		},
+		{
+			compute.BackendService{Protocol: "HTTP"},
+			&BackendService{Protocol: "HTTP"},
+		},
+	}
+	for _, testCase := range testCases {
+		result, _ := toBackendService(testCase.input)
+		if !reflect.DeepEqual(result, testCase.expected) {
+			t.Fatalf("Error during conversion to BackendService. Expected %+v but got %+v", testCase.expected, result)
```
bowei: t.Fatalf("toBackendService(input) = \ninput = %s\n%s\nwant = \n%s", pretty.Sprint(testCase.input), pretty.Sprint(result), pretty.Sprint(testCase.expected))

use pretty (https://github.com/kr/pretty)

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+		Version:  meta.VersionAlpha,
+		Protocol: "HTTP2",
+	}
+	alpha, err := composite.toAlpha()
+	if err != nil {
+		t.Fatalf("Did not expect error when calling toAlpha(): %v", err)
+	}
+	// Verify that a known alpha field value is preserved.
+	if alpha.Protocol != "HTTP2" {
+		t.Fatalf("Expected protocol to be HTTP2. Got %v", alpha.Protocol)
```
bowei: t.Errorf("alpha.Protocol = %q, want HTTP2", alpha.Protocol, ...)

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func TestToGA(t *testing.T) {
+	composite := BackendService{
+		Version:  meta.VersionGA,
+		Protocol: "HTTP",
+	}
+	ga, err := composite.toGA()
+	if err != nil {
+		t.Fatalf("Did not expect error when calling toGA(): %v", err)
```
bowei: see above

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if ga.Protocol != "HTTP" {
+		t.Fatalf("Expected protocol to be HTTP. Got %v", ga.Protocol)
```
bowei: same

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+// typeEquality is a generic function that checks type equality.
+func typeEquality(t1 reflect.Type, t2 reflect.Type) error {
+	if t1.NumField() != t2.NumField() {
+		return fmt.Errorf("%v contained %v fields, %v contained %v fields", t1.Name(), t1.NumField(), t2.Name(), t2.NumField())
```
bowei: use %q

I would use a map...

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	for i := 0; i < t1.NumField(); i++ {
+		if err := compareFields(t1.Field(i), t2.Field(i)); err != nil {
```
bowei: don't you need to create a map on the field name? is the order guaranteed to match up?

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// compareFields verifies that two fields in a struct have the same relevant metadata.
+// Note: This comparison ignores field offset, index, and pkg path, all of which don't matter.
+func compareFields(s1 reflect.StructField, s2 reflect.StructField) error {
```
bowei: I would pass in the current field path that is being checked and add that to the error, e.g. for 
```go
A : {
  B: {
    C: {
      D: ...
    }
  }
}
```

when you are checking `D` above, the field path would be "A.B.C.D".

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if s1.Name != s2.Name {
+		return fmt.Errorf("Expected field to be named %v. Got %v", s2.Name, s1.Name)
```
bowei: fmt.Errorf("field %s name = %q, want %q", ...)

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if s1.Tag != s2.Tag {
+		return fmt.Errorf("Expected field tag to be %v. Got %v", s2.Tag, s1.Tag)
```
bowei: is this required to be the same?

pkg/backends/compute_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if s1.Type.Name() != s2.Type.Name() {
+		return fmt.Errorf("Expected field type to be %v. Got %v", s2.Type.Name(), s1.Type.Name())
```
bowei: fmt.Errorf("field %s, type = %q, want %q|", path)

pkg/backends/backends.go
bowei: ack

PR #281, opened 2018-05-22
Define security policy API in BackendConfig

MrHohn:
@rramkumar1 @bowei


MrHohn:
@rramkumar1 @bowei  The API is now approved :)

rramkumar1:
This looks fine to me. IMO, I think we have should have a CloudArmorConfig struct just to be consistent with the other features and then specify the security policy struct inside of that.

PR #289, opened 2018-05-30
Add custom validation for BackendConfig + hook validation into Translator

rramkumar1:
This PR adds custom validation for the BackendConfig, particularly IAP. It also hooks this validation into the existing logic in the translator.

This is the first of a couple upcoming PR's that will provide the plumbing for IAP + CDN support.

/assign @MrHohn @nicksardo

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/289)


pkg/backendconfig/validation.go
```diff
@@ -0,0 +1,69 @@
...
+	OAuthClientSecretKey = "client_secret"
+)
+
+func Validate(kubeClient kubernetes.Interface, beConfig *backendconfigv1beta1.BackendConfig) error {
+	if beConfig != nil {
+		if err := validateIAP(kubeClient, beConfig); err != nil {
+			return err
+		}
+	}
+	return nil
```
MrHohn: What about early return the nil case?
```golang
	if beConfig == nil {
		return nil
	}
	return validateIAP(kubeClient, beConfig)
```

pkg/backendconfig/validation.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func validateIAP(kubeClient kubernetes.Interface, beConfig *backendconfigv1beta1.BackendConfig) error {
+	// If IAP settings are not found or IAP is not enabled then don't bother continuing.
+	if beConfig.Spec.Iap == nil || beConfig.Spec.Iap.Enabled == false {
+		return nil
+	}
+	// If necessary, get the OAuth credentials stored in the K8s secret.
+	if beConfig.Spec.Iap.OAuthClientCredentials.SecretName != "" {
```
MrHohn: Can `beConfig.Spec.Iap.OAuthClientCredentials` be nil?

MrHohn: Ha, I guess it can not because this field doesn't have `omitempty` on it.

pkg/backendconfig/validation.go
```diff
@@ -0,0 +1,69 @@
...
+		secretName := beConfig.Spec.Iap.OAuthClientCredentials.SecretName
+		secret, err := kubeClient.Core().Secrets(beConfig.Namespace).Get(secretName, meta_v1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("error retrieving secret %v: %v", secretName, err)
+		}
+		clientID, ok := secret.Data[OAuthClientIDKey]
+		if !ok {
+			return fmt.Errorf("secret %v has no 'client_id'", secretName)
```
MrHohn: Replace `client_id` with OAuthClientIDKey?

pkg/backendconfig/validation.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		clientSecret, ok := secret.Data[OAuthClientSecretKey]
+		if !ok {
+			return fmt.Errorf("secret %v has no 'client_secret'", secretName)
```
MrHohn: Replace `client_secret` with OAuthClientSecretKey?

pkg/backendconfig/validation.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		beConfig.Spec.Iap.OAuthClientCredentials.ClientID = string(clientID)
+		beConfig.Spec.Iap.OAuthClientCredentials.ClientSecret = string(clientSecret)
```
MrHohn: Can user set `ClientID` and `ClientSecret` directly in BackendConfig?

pkg/backendconfig/validation_test.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func TestValidateIAP(t *testing.T) {
+	testCases := []struct {
+		desc        string
+		init        func(kubeClient kubernetes.Interface)
+		expectError bool
+	}{
+		{
+			"secret does not exist",
```
MrHohn: Should include field name on assignment.
```golang
{
    desc: "secret does not exist",
    ....
```

pkg/backendconfig/validation_test.go
```diff
@@ -0,0 +1,146 @@
...
+					},
+				}
+				kubeClient.Core().Secrets("default").Create(secret)
+			},
+			false,
+		},
+		{
+			"iap and cdn enabled at the same time",
+			func(kubeClient kubernetes.Interface) {
+				beConfig.Spec.Cdn = &backendconfigv1beta1.CDNConfig{
```
MrHohn: Have a shared `beConfig` but modify it in-flight seems error prone...Probably just define a diifrent `beConfig`?

pkg/controller/translator/translator.go
```diff
@@ -112,16 +112,17 @@ PortLoop:
...
-			return nil, err
+			return nil, errors.ErrSvcBackendConfig{ServicePortID: id, Err: err}
 		}
-		if backendConfigInStore != nil {
-			// Object in cache could be changed in-flight. Deepcopy to
-			// reduce race conditions.
-			backendConfig = backendConfigInStore.DeepCopy()
+		// Object in cache could be changed in-flight. Deepcopy to
+		// reduce race conditions.
+		beConfig = beConfig.DeepCopy()
```
MrHohn: This might panic if `beConfig` is nil?

pkg/controller/translator/translator.go
```diff
@@ -112,16 +112,17 @@ PortLoop:
...
+		if err = backendconfig.Validate(t.ctx.KubeClient, beConfig); err != nil {
+			return nil, errors.ErrBackendConfigValidation{BackendConfig: *beConfig, Err: err}
```
MrHohn: Should check nil before `*beConfig` as well?

pkg/controller/errors/errors.go
```diff
@@ -62,3 +63,24 @@ type ErrSvcAppProtosParsing struct {
...
+}
+
+// ErrBackendConfigValidation is returned when there was an error validating a BackendConfig.
+type ErrBackendConfigValidation struct {
+	backendconfigv1beta1.BackendConfig
+	Err error
+}
+
+func (e ErrBackendConfigValidation) Error() string {
+	return fmt.Sprintf("BackendConfig %v is not valid: %v", e.BackendConfig.Name, e.Err)
```
MrHohn: Include namespace in error message?

pkg/controller/translator/translator.go
```diff
@@ -112,16 +112,17 @@ PortLoop:
...
+			return nil, errors.ErrSvcBackendConfig{ServicePortID: id, Err: err}
 		}
-		if backendConfigInStore != nil {
-			// Object in cache could be changed in-flight. Deepcopy to
-			// reduce race conditions.
-			backendConfig = backendConfigInStore.DeepCopy()
+		// Object in cache could be changed in-flight. Deepcopy to
+		// reduce race conditions.
+		beConfig = beConfig.DeepCopy()
+		if err = backendconfig.Validate(t.ctx.KubeClient, beConfig); err != nil {
```
MrHohn: Is it possible to do this validation in an earlier stage (e.g. on BackendConfig creation/update)?

pkg/backendconfig/validation.go
rramkumar1: Done

pkg/backendconfig/validation.go
rramkumar1: Done

pkg/backendconfig/validation.go
rramkumar1: Yes, if they set it directly and also provide a secret, the data in the secret will overwrite the existing one. If they set it directly but don't provide a secret, we will use what they provided directly.

pkg/controller/translator/translator.go
MrHohn: Humm..Though ingress can still be put into queue triggered by service/ingress update so we need it here.

pkg/controller/translator/translator.go
rramkumar1: I checked the DeepCopy() code and it returns nil if the parameter is nil

pkg/controller/translator/translator.go
rramkumar1: No, without a webhook, we cannot do it on creation/update.

pkg/controller/translator/translator.go
rramkumar1: If its nil, then Validate() would not return an error so I dont think we would need that check.

pkg/controller/errors/errors.go
rramkumar1: Done

pkg/backendconfig/validation_test.go
rramkumar1: Done.

pkg/backendconfig/validation_test.go
rramkumar1: I'm going to tackle this in a separate PR. Left myself a TODO.

pkg/controller/translator/translator.go
MrHohn: Ah good point, yeah it should work.

pkg/backendconfig/validation.go
MrHohn: minor: Thinking a nil check would still be good to have --- what if there is a bug in CRD validation that a nil OAuthClientCredentials gets passed in.

rramkumar1: Added check.

MrHohn: Thanks!

PR #284, opened 2018-05-23
Add annotation for exposing NEGs

agau4779:
This PR adds a new annotation for exposing standalone NEGs on Services without requiring an Ingress.

Additionally:
- Refactors NEG naming to use ServicePort Ports instead of TargetPorts, because it is possible to have multiple ServicePorts point to the same TargetPort. Note that ServicePort Ports are always int32, while TargetPorts may be an int or a string.
- Pass ServicePort into via a map mapping Port:TargetPort within `/pkg/neg` (PortNameMap type)
- feature flag around exposed NEGs

To do in later PRs:
- merge the expose NEG and ingress NEG annotations

PR #296, opened 2018-06-01
name NEGs with ServicePort instead of TargetPort

agau4779:
Refactor that makes https://github.com/kubernetes/ingress-gce/pull/284 easier

### DIFF
--- a/pkg/backends/backends_test.go
+++ b/pkg/backends/backends_test.go
@@ -555,34 +555,34 @@ func TestBackendPoolSyncQuota(t *testing.T) {
 		{
 			[]utils.ServicePort{{NodePort: 8080}, {NodePort: 443}},
 			[]utils.ServicePort{
-				{NodePort: 8080, SvcTargetPort: "testport8080", NEGEnabled: true},
-				{NodePort: 443, SvcTargetPort: "testport443", NEGEnabled: true},
+				{SvcPort: 8080, NodePort: 8080, NEGEnabled: true},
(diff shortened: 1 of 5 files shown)


pkg/utils/serviceport.go
```diff
@@ -74,7 +75,7 @@ func (sp ServicePort) BackendName(namer *Namer) string {
 		return namer.IGBackend(sp.NodePort)
 	}
 
-	return namer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.SvcTargetPort)
+	return namer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, fmt.Sprintf("%v", sp.SvcPort))
```
bowei: does it make sense to pass it as a string?

pkg/backends/backends_test.go
```diff
@@ -857,7 +857,7 @@ func TestLinkBackendServiceToNEG(t *testing.T) {
 
 	for _, zone := range zones {
 		err := fakeNEG.CreateNetworkEndpointGroup(&computealpha.NetworkEndpointGroup{
-			Name: defaultNamer.NEG(namespace, name, port),
+			Name: defaultNamer.NEG(namespace, name, fmt.Sprintf("%v", svcPort.SvcPort)),
```
bowei: see comment below

agau4779:
Incorporated into https://github.com/kubernetes/ingress-gce/pull/284 , addressed comments

PR #297, opened 2018-06-02
Typo fix in gce-ingress-controller.yaml

AdamDang:
permissable->permissible


bowei:
It would be best if you submitted this with an actual code fix.

PR #300, opened 2018-06-04
Add IngressValidator and supporting utilities

bowei:
This adds a reusability Ingress spec testing utility. The vanilla
features of Ingress are tested with IngressValidator. Additional addon
features are supported by extending the validator with the Feature
modules in the features/ package.


rramkumar1:
@bowei The unit test is failing on the vendor. I think you should use github.com/kr/pretty instead since now that is vendored in.

PR #301, opened 2018-06-04
IAP + CDN 

rramkumar1:
This is the last PR for IAP + CDN integration.

TODO's in separate PR's:
    1. Need to write an e2e test using new test framework.
    2. Need to add more unit tests to verify behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/301)

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce"
 	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce/cloud/meta"
 
+	"k8s.io/ingress-gce/pkg/backends/features"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/healthchecks"
 	"k8s.io/ingress-gce/pkg/instances"
@@ -76,13 +77,14 @@ const maxRPS = 1
 
 // Backends implements BackendPool.
 type Backends struct {
-	cloud         *gce.GCECloud
-	negGetter     NEGGetter
-	nodePool      instances.NodePool
-	healthChecker healthchecks.HealthChecker
-	snapshotter   storage.Snapshotter
-	prober        ProbeProvider
-	namer         *utils.Namer
+	cloud                *gce.GCECloud
+	negGetter            NEGGetter
+	nodePool             instances.NodePool
+	healthChecker        healthchecks.HealthChecker
+	snapshotter          storage.Snapshotter
+	prober               ProbeProvider
+	namer                *utils.Namer
+	backendConfigEnabled bool
 }
 
 // Backends is a BackendPool.
@@ -101,14 +103,16 @@ func NewBackendPool(
 	healthChecker healthchecks.HealthChecker,
 	nodePool instances.NodePool,
 	namer *utils.Namer,
+	backendConfigEnabled,
 	resyncWithCloud bool) *Backends {
 
 	backendPool := &Backends{
-		cloud:         cloud,
-		negGetter:     negGetter,
-		nodePool:      nodePool,
-		healthChecker: healthChecker,
-		namer:         namer,
+		cloud:                cloud,
+		negGetter:            negGetter,
+		nodePool:             nodePool,
+		healthChecker:        healthChecker,
+		namer:                namer,
+		backendConfigEnabled: backendConfigEnabled,
 	}
 	if !resyncWithCloud {
 		backendPool.snapshotter = storage.NewInMemoryPool()
@@ -289,6 +293,11 @@ func (b *Backends) ensureBackendService(sp utils.ServicePort, igLinks []string) 
 	needUpdate := ensureProtocol(be, sp)
 	needUpdate = ensureHealthCheckLink(be, hcLink) || needUpdate
 	needUpdate = ensureDescription(be, sp.Description()) || needUpdate
+	if b.backendConfigEnabled && sp.BackendConfig != nil {
+		needUpdate = features.EnsureCDN(sp, be) || needUpdate
+		needUpdate = features.EnsureIAP(sp, be) || needUpdate
+	}
+
 	if needUpdate {
 		if err = composite.UpdateBackendService(be, b.cloud); err != nil {
 			return err
--- a/pkg/backends/features/cdn.go
+++ b/pkg/backends/features/cdn.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+import (
+	"reflect"
+
+	"github.com/golang/glog"
+	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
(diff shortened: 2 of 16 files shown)


ConradIrwin:
Thanks @rramkumar1 looking forward to this :D.

pkg/backends/features/cdn.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
MrHohn: 2018

pkg/backends/features/cdn.go
```diff
@@ -0,0 +1,71 @@
...
+// to the passed in compute.BackendService. A GCE API call still needs to be
+// made to actually persist the changes.
+func applyCDNSettings(sp utils.ServicePort, be *composite.BackendService) {
+	beConfig := sp.BackendConfig
+	setCDNDefaults(beConfig)
+	// Apply the boolean switch
+	be.EnableCDN = beConfig.Spec.Cdn.Enabled
+	// Apply the cache key policies
+	cacheKeyPolicy := beConfig.Spec.Cdn.CachePolicy
+	be.CdnPolicy = &composite.BackendServiceCdnPolicy{CacheKeyPolicy: &composite.CacheKeyPolicy{}}
```
MrHohn: Curious is it valid to have `be.EnableCDN=false` but `be.CdnPolicy=non-nil`?

pkg/backends/features/cdn.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return false
+}
+
+// applyCDNSettings applies the CDN settings specified in the BackendConfig
+// to the passed in compute.BackendService. A GCE API call still needs to be
+// made to actually persist the changes.
+func applyCDNSettings(sp utils.ServicePort, be *composite.BackendService) {
+	beConfig := sp.BackendConfig
+	setCDNDefaults(beConfig)
```
MrHohn: So this is applying defaults to backendConfig but not backendService, I was confused :)

pkg/composite/composite.go
```diff
@@ -240,6 +240,13 @@ func (be *BackendService) toAlpha() (*computealpha.BackendService, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error unmarshalling BackendService JSON to compute alpha type: %v", err)
 	}
+	// Set force send fields. This is a temporary hack.
+	if alpha.CdnPolicy != nil && alpha.CdnPolicy.CacheKeyPolicy != nil {
+		alpha.CdnPolicy.CacheKeyPolicy.ForceSendFields = []string{"IncludeHost", "IncludeProtocol", "IncludeQueryString", "QueryStringBlaclist", "QueryStringWhitelist"}
```
MrHohn: QueryStringBlaclist -> QueryStringBlacklist, same for below.

pkg/backends/features/iap.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	beTemp := &composite.BackendService{}
+	applyIAPSettings(sp, beTemp)
+	// We need to compare the SHA256 of the client secret instead of the client secret itself
+	// since that field is redacted when getting a BackendService.
+	beTemp.Iap.Oauth2ClientSecretSha256 = fmt.Sprintf("%x", sha256.Sum256([]byte(beTemp.Iap.Oauth2ClientSecret)))
+	if be.Iap == nil || beTemp.Iap.Enabled != be.Iap.Enabled || beTemp.Iap.Oauth2ClientId != be.Iap.Oauth2ClientId || beTemp.Iap.Oauth2ClientSecretSha256 != be.Iap.Oauth2ClientSecretSha256 {
+		glog.V(3).Infof("BackendService is %+v", be)
+		applyIAPSettings(sp, be)
+		glog.V(3).Infof("Updated IAP settings for service %v/%v.", sp.ID.Service.Namespace, sp.ID.Service.Name)
```
MrHohn: Maybe V(2)? Seems worth to log it if an update is needed.

Also `Updated` might not be accurate? As the actual update will happen after this.

pkg/backends/features/iap.go
```diff
@@ -0,0 +1,71 @@
...
+	if sp.BackendConfig == nil {
+		return false
+	}
+	beTemp := &composite.BackendService{}
+	applyIAPSettings(sp, beTemp)
+	// We need to compare the SHA256 of the client secret instead of the client secret itself
+	// since that field is redacted when getting a BackendService.
+	beTemp.Iap.Oauth2ClientSecretSha256 = fmt.Sprintf("%x", sha256.Sum256([]byte(beTemp.Iap.Oauth2ClientSecret)))
+	if be.Iap == nil || beTemp.Iap.Enabled != be.Iap.Enabled || beTemp.Iap.Oauth2ClientId != be.Iap.Oauth2ClientId || beTemp.Iap.Oauth2ClientSecretSha256 != be.Iap.Oauth2ClientSecretSha256 {
+		glog.V(3).Infof("BackendService is %+v", be)
```
MrHohn: Is this necessary?

pkg/backends/features/iap.go
```diff
@@ -0,0 +1,71 @@
...
+	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
+	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/utils"
+)
+
+// EnsureIAP reads the IAP configuration specified in the BackendConfig
+// and applies it to the BackendService if it is stale. It returns true
+// if there were existing settings on the BackendService that were overwritten.
+func EnsureIAP(sp utils.ServicePort, be *composite.BackendService) bool {
+	if sp.BackendConfig == nil {
```
MrHohn: I might be missing something...How is the disable case handled? Similar question for CDN as well.

pkg/backends/features/cdn.go
rramkumar1: Done

pkg/backends/features/cdn.go
rramkumar1: Yep, we apply defaults to the BackendConfig so that in the apply[IAP,CDN]Settings functions we don't have to have ugly nil checks.

pkg/backends/features/cdn.go
rramkumar1: Yep. If the user disabled CDN but provided cache policies, then we will act on it accordingly.

pkg/backends/features/iap.go
rramkumar1: Oops, was debugging something and forgot to delete.

pkg/backends/features/iap.go
rramkumar1: Done

pkg/composite/composite.go
rramkumar1: Nice catch :)

pkg/backends/features/iap.go
rramkumar1: Discussed this offline but we should probably document that user should remove the annotation from the service first?

pkg/backends/backends.go
```diff
@@ -101,14 +103,16 @@ func NewBackendPool(
...
-		negGetter:     negGetter,
-		nodePool:      nodePool,
-		healthChecker: healthChecker,
-		namer:         namer,
+		cloud:                cloud,
+		negGetter:            negGetter,
+		nodePool:             nodePool,
+		healthChecker:        healthChecker,
+		namer:                namer,
+		backendConfigEnabled: backendConfigEnabled,
```
MrHohn: Might be worth to separate this out to another PR if we can't get this in soon..

ConradIrwin:
w00w00! thanks all :D

rramkumar1:
@ConradIrwin Note that this is not officially available yet. It will only be released when we cut a new ingress-gce release, which will be 1.2.0

ConradIrwin:
Noted, thank you. Do you have an estimate on that timeline?

I'm just excited to see you're working through the backlog of feature requests :). (and I'd cheekily put another nudge in for #33, but don't block shipping this on implementing that :D)

rramkumar1:
@ConradIrwin A very conservative estimate would be early July.

PR #303, opened 2018-06-05
Use generated mocks to implement unit tests for pkg/backends 

rramkumar1:
This PR integrates the gce generated mocks into pkg/backends. 

This PR vendors k/pkg/cloudprovider/providers/gce/cloud/mock and there are two raw patches to the vendored code which will need to be upstreamed at some point:
   1. Add a new file called gce_fake.go which exposes a generic fake GCECloud object.
   2. Add new hooks functions to existing mock.go file

This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/303)


pkg/backends/backends_test.go
```diff
@@ -1041,3 +1000,59 @@ func TestEnsureBackendServiceHealthCheckLink(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		if comparableGroupPath(tc.igPath) != tc.expected {
+			t.Errorf("expected %s, but got %s", tc.expected, comparableGroupPath(tc.igPath))
+		}
+	}
+}
+
+func TestRetrieveObjectName(t *testing.T) {
```
nicksardo: Can you move these funcs back to there original position so my PR won't conflict as much?

rramkumar1: Done.

PR #314, opened 2018-06-06
Add server version to echo

nicksardo:


nicksardo:
Should we bother with including the commit hash?

rramkumar1:
Not worth it right now, we can always add it if needed.
/lgtm

PR #318, opened 2018-06-07
Break out some helper functions for better testing + reuse

rramkumar1:
The motivation behind this change is to be able to reuse these new utility functions in our new testing framework. Also, breaking these functions out make them easier to test.

/assign @nicksardo 
/assign @MrHohn

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/318)

### DIFF
--- a/pkg/backendconfig/utils_test.go
+++ b/pkg/backendconfig/utils_test.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


pkg/backendconfig/utils_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"testing"
+
+	apiv1 "k8s.io/api/core/v1"
+
+	"k8s.io/ingress-gce/pkg/annotations"
+)
+
+func TestBackendConfigName(t *testing.T) {
```
MrHohn: Might be worth to have a test case where both `default` and `ports` are given and check the `ports` one takes precedence?

rramkumar1: Resolved offline

PR #319, opened 2018-06-07
IAP + CDN e2e testing implementation

rramkumar1:
Note that this PR also contains some small changes to the validator:
    1.  In order to prevent the http client in the validator from following redirects, I implemented the 
         CheckRedirect hook. 
    2.  Move the pathToDefaultBackend variable up higher so that other files can use it.

Ref: #301

This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/319)

### DIFF
--- a/pkg/fuzz/features/cdn.go
+++ b/pkg/fuzz/features/cdn.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+import (
+	"fmt"
+	"net/http"
+
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/fuzz"
+)
+
+// CDN is a feature in BackendConfig that supports using GCP CDN.
+var CDN = &CDNFeature{}
+
+// CDNFeature implements the associated feature.
+type CDNFeature struct{}
+
+// NewValidator implements fuzz.Feature.
+func (CDNFeature) NewValidator() fuzz.FeatureValidator {
+	return &cdnValidator{}
+}
+
+// Name implements fuzz.Feature.
+func (*CDNFeature) Name() string {
+	return "CDN"
+}
+
+// cdnValidator is a validator for CDNFeature.
+type cdnValidator struct {
+	fuzz.NullValidator
+
+	env fuzz.ValidatorEnv
+	ing *v1beta1.Ingress
+}
+
+// Name implements fuzz.FeatureValidator.
+func (*cdnValidator) Name() string {
+	return "CDN"
+}
+
+// ConfigureAttributes implements fuzz.FeatureValidator.
+func (v *cdnValidator) ConfigureAttributes(env fuzz.ValidatorEnv, ing *v1beta1.Ingress, a *fuzz.IngressValidatorAttributes) error {
+	// Capture the env for use later in CheckResponse.
+	v.ing = ing
+	v.env = env
+	return nil
+}
+
+// ModifyRequest implements fuzz.FeatureValidator.
+func (v *cdnValidator) ModifyRequest(host, path string, req *http.Request) {
+	// Warm up the cache, regardless of whether this host + path has CDN enabled.
+	values := req.URL.Query()
+	// Note: cache=true tells the echo server to add caching headers.
+	values.Add("cache", "true")
+	req.URL.RawQuery = values.Encode()
+}
+
+// CheckResponse implements fuzz.FeatureValidator.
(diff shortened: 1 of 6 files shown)


pkg/fuzz/features/cdn.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/ingress-gce/pkg/fuzz"
+)
+
+// Cdn is a feature in BackendConfig that supports using GCP CDN.
+var Cdn = &CDNFeature{}
```
bowei: cdn? (no need to export?)

pkg/fuzz/features/cdn.go
```diff
@@ -0,0 +1,93 @@
...
+func (v *cdnValidator) ConfigureAttributes(env fuzz.ValidatorEnv, ing *v1beta1.Ingress, a *fuzz.IngressValidatorAttributes) error {
+	// Capture the env for use later in CheckResponse.
+	v.ing = ing
+	v.env = env
+	return nil
+}
+
+// ModifyRequest implements fuzz.FeatureValidator.
+func (v *cdnValidator) ModifyRequest(host, path string, req *http.Request) {
+	// Warm up the cache, regardless of whether this host + path has CDN enabled.
```
bowei: should probably comment that // cache=true makes the echo server add caching headers

pkg/fuzz/features/features.go
```diff
@@ -25,4 +25,6 @@ import "k8s.io/ingress-gce/pkg/fuzz"
 var All = []fuzz.Feature{
 	AllowHTTP,
 	PresharedCert,
+	Cdn,
```
bowei: CDN

pkg/fuzz/features/features.go
```diff
@@ -25,4 +25,6 @@ import "k8s.io/ingress-gce/pkg/fuzz"
...
+	Iap,
```
bowei: IAP (golang code style)

pkg/fuzz/features/iap.go
```diff
@@ -0,0 +1,92 @@
...
+// CheckResponse implements fuzz.FeatureValidator.
+func (v *iapValidator) CheckResponse(host, path string, resp *http.Response, body []byte) (fuzz.CheckResponseAction, error) {
+	backendConfig, err := fuzz.BackendConfigForPath(host, path, v.ing, v.env)
+	if err != nil {
+		return fuzz.CheckResponseContinue, err
+	}
+	var iapEnabled bool
+	if backendConfig.Spec.Iap != nil && backendConfig.Spec.Iap.Enabled == true {
+		iapEnabled = true
+	}
```
bowei: do you need iapEnabled? why not return if false?

pkg/fuzz/features/iap.go
```diff
@@ -0,0 +1,92 @@
...
+			glog.V(2).Infof("The response was %v", resp)
+			return fuzz.CheckResponseContinue, fmt.Errorf("IAP is turned on but response %v did not return a 302", resp)
+		}
+		if resp.Header.Get("x-goog-iap-generated-response") == "" {
+			return fuzz.CheckResponseContinue, fmt.Errorf("IAP is turned on but response w/ header %v did not contain IAP header", resp.Header)
+		}
+		return fuzz.CheckResponseSkip, nil
+	}
+	// If IAP is turned off, verify that the response code was not 302.
+	if !iapEnabled && resp.StatusCode == http.StatusFound {
```
bowei: don't need this?, standard check will be for 200

pkg/fuzz/features/cdn.go
rramkumar1: Since I made this fully capitalized now, it will technically be exported.

pkg/fuzz/features/cdn.go
rramkumar1: Done..

pkg/fuzz/features/features.go
rramkumar1: Done.

pkg/fuzz/features/features.go
rramkumar1: Done.

pkg/fuzz/features/iap.go
rramkumar1: See comment below.

pkg/fuzz/features/iap.go
rramkumar1: The main reason was to return a custom error.

bowei:
We need to tag the image used for guitar tests. Right now as it targets master, the tests will start failing as it is testing the old ingress controller. Let's talk tomorrow.

/lgtm

pkg/fuzz/helpers.go
```diff
@@ -17,13 +17,59 @@ limitations under the License.
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/annotations"
 	backendconfig "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
+	backendconfigutil "k8s.io/ingress-gce/pkg/backendconfig"
+	translatorutil "k8s.io/ingress-gce/pkg/controller/translator"
 )
 
+// BackendConfigForPath returns the BackendConfig associated with the given path.
+// Note: This function returns an empty object (not nil pointer) if a BackendConfig
```
bowei: we probably want to distinguish these two cases for testing BackendConfig does not exist vs BackendConfig exists but is empty.

rramkumar1: You mean have separate function for each case? I can take care of it in a separate PR.

PR #333, opened 2018-06-12
Update Dockerfile for the e2e test

bowei:


cmd/e2e-test/e2e-test.yaml
```diff
@@ -65,7 +59,7 @@ metadata:
 spec:
   containers:
   - name: ingress-e2e
-    image: gcr.io/k8s-ingress-image-push/ingress-gce-e2e-test-amd64:v1.1.0-123-g7aaa9456
+    image: gcr.io/k8s-ingress-image-push/ingress-gce-e2e-test-amd64:master
```
rramkumar1: This should be ingress-gce-e2e-e2e-test-amd64

bowei: fix pusher to not add the -e2e prefix

PR #348, opened 2018-06-19
Make sure we get a BackendService after updating it to populate object fingerprint

rramkumar1:
Turns out that we had an update call without a corresponding get to update the fingerprint. 

/assign @nicksardo

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -311,6 +311,12 @@ func (b *Backends) ensureBackendService(sp utils.ServicePort, igLinks []string) 
 		if err = composite.UpdateBackendService(be, b.cloud); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


nicksardo:
We can do this for now. It's easy.
/lgtm

pkg/backends/backends.go
```diff
@@ -311,6 +311,12 @@ func (b *Backends) ensureBackendService(sp utils.ServicePort, igLinks []string)
 		if err = composite.UpdateBackendService(be, b.cloud); err != nil {
 			return err
 		}
+		// Get the BackendService after an update so that the Fingerprint
```
bowei: does the update call itself return the new backendservice fingerprint? (may not need to call get against explicitly)

nicksardo: It does not.

PR #352, opened 2018-06-20
Trigger ingress sync on system default backend update

MrHohn:
Given that setting backend config with system default backend is valid, update on system default backend should trigger ingress sync as well.

/assign @rramkumar1


MrHohn:
@nicksardo Removed switch case, PTAL thanks!

MrHohn:
Updated the code a bit..

MrHohn:
Sorry for the noise, unexported `defaultBackendSvcPortID`.

PR #354, opened 2018-06-20
Fix hasAlphaResource and hasBetaResource

MrHohn:
Seems like resourceType was mistakenly hardcoded.
/assign @rramkumar1


MrHohn:
Thanks!

PR #356, opened 2018-06-21
URLMap sync

nicksardo:
Changes
 - `UpdateUrlMap()` mostly turned into static func `toComputeURLMap()`. It's sync logic moved to `ensureComputeURLMap()`.
 - `L7s.Sync()` calls `ensureComputeURLMap()` to assert the URLMap's existence and correctness instead of making a separate, later call to `UpdateUrlMap()` 
 - `utils.GCEURLMap` used a map for hostname->path rules which made the generated URLMap's order non-deterministic (presumably this caused unnecessary API calls). It now uses a slice and maintains uniqueness.

Bulk of this PR is in tests or related datafiles.

### DIFF
--- a/pkg/controller/translator/translator_test.go
+++ b/pkg/controller/translator/translator_test.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"testing"
 	"time"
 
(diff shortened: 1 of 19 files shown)


pkg/controller/translator/translator_test.go
```diff
@@ -544,23 +544,23 @@ func newDefaultEndpoint(name string) *apiv1.Endpoints {
 func ingressFromFile(filename string) *extensions.Ingress {
 	data, err := ioutil.ReadFile("testdata/" + filename)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalln(filename, err)
```
rramkumar1: Just curious, why the need for Fatalln vs Fatal?

nicksardo: Absolutely no reason. Thanks for pointing this out. Decided to use `t.Fatalf` anyway...

nicksardo:
Copied the retrieved URLMap's fingerprint to the expected URLMap.

PR #359, opened 2018-06-21
Make WaitForIngressDeletion() a util func and add a knob for omitting default backend

MrHohn:
Thinking a knob for omitting default backend service might be useful given it can be shared among LBs.

/assign @rramkumar1 @bowei


MrHohn:
Included in https://github.com/kubernetes/ingress-gce/pull/360.

PR #365, opened 2018-06-25
Fix the issue where Shutdown doesn't shutdown taskqueue

anfernee:
Originally, PeriodicTaskQueue has 2 loops:
- wait.Until managed by stopCh
- for loop in worker() managed by workqueue

Shutdown only exits the second loop. Failing to close stopCh before
calling Shutdown will cause panic because of double close workDone
channel.

### DIFF
--- a/pkg/utils/taskqueue_test.go
+++ b/pkg/utils/taskqueue_test.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"reflect"
 	"testing"
-	"time"
 
 	"k8s.io/client-go/tools/cache"
 )
@@ -28,7 +27,6 @@ import (
 func TestPeriodicTaskQueue(t *testing.T) {
 	t.Parallel()
 	synced := map[string]bool{}
-	stopCh := make(chan struct{})
 	doneCh := make(chan struct{}, 1)
 
 	var tq TaskQueue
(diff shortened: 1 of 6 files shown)


bowei:
Is it possible to add a test for the misbehavior?

pkg/utils/taskqueue_test.go
```diff
@@ -44,14 +42,13 @@ func TestPeriodicTaskQueue(t *testing.T) {
...
 
-	go tq.Run(time.Microsecond, stopCh)
+	go tq.Run()
 	tq.Enqueue(cache.ExplicitKey("a"))
 	tq.Enqueue(cache.ExplicitKey("b"))
 	tq.Enqueue(cache.ExplicitKey("err"))
 	tq.Enqueue(cache.ExplicitKey("stop"))
 
 	<-doneCh
-	close(stopCh)
```
anfernee: @bowei, the main change is to remove the `stopCh` from `tq`, so in order to shutdown the `tq`, we need to close the channel and then call `tq.Shutdown()`. There is no misbehavior with current implementation, only complexity. e.g. `close` must happen before shutdown, or it's possible that it won't be shutdown. 

In terms of test, I think I can add another shutdown, to make sure no panic.

anfernee: I added L56 to test shutdown actually works.

bowei:
can you rebase this?

PR #366, opened 2018-06-26
Swtich to use beta HealthCheck for NEG

freehan:
cc @agau4779

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -17,19 +17,20 @@ limitations under the License.
 package healthchecks
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
 	computealpha "google.golang.org/api/compute/v0.alpha"
+	computebeta "google.golang.org/api/compute/v0.beta"
 	compute "google.golang.org/api/compute/v1"
-	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/golang/glog"
-
-	"encoding/json"
-
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce/cloud/meta"
 )
 
 const (
@@ -102,9 +103,7 @@ func (h *HealthChecks) New(sp utils.ServicePort) *HealthCheck {
 // Sync retrieves a health check based on port, checks type and settings and updates/creates if necessary.
 // Sync is only called by the backends.Add func - it's not a pool like other resources.
 func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
-	// Use alpha API when PORT_SPECIFICATION field is specified or when Type
-	// is HTTP2
-	existingHC, err := h.Get(hc.Name, hc.isHttp2() || hc.ForNEG)
+	existingHC, err := h.Get(hc.Name, hc.Version())
 	if err != nil {
 		if !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 			return "", err
@@ -114,7 +113,7 @@ func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
 			return "", err
 		}
 
-		return h.getHealthCheckLink(hc.Name, hc.isHttp2())
+		return h.getHealthCheckLink(hc.Name, hc.Version())
 	}
 
 	if needToUpdate(existingHC, hc) {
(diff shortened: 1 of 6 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -135,30 +134,51 @@ func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
...
+		return h.cloud.UpdateBetaHealthCheck(betaHC)
+	case meta.VersionGA:
 		glog.V(2).Infof("Updating health check for port %v with protocol %v", newHC.Port, newHC.Type)
 		v1hc, err := newHC.ToComputeHealthCheck()
 		if err != nil {
 			return err
 		}
 		return h.cloud.UpdateHealthCheck(v1hc)
+	default:
+		return fmt.Errorf("Unkonwn Version: %q", newHC.Version())
```
agau4779: nit: `Unknown` . I think it also is a typo earlier in this file

pkg/healthchecks/healthchecks.go
```diff
@@ -197,18 +217,23 @@ func (h *HealthChecks) Delete(name string) error {
 }
 
 // Get returns the health check by port
-func (h *HealthChecks) Get(name string, alpha bool) (*HealthCheck, error) {
+func (h *HealthChecks) Get(name string, version meta.Version) (*HealthCheck, error) {
 	var hc *computealpha.HealthCheck
 	var err error
-	if alpha {
+	if version == meta.VersionAlpha {
 		hc, err = h.cloud.GetAlphaHealthCheck(name)
```
agau4779: This part of the if clause is missing the:
```
if err != nil {
    return nil, err
}
````
Alternatively, just do this check once at the end of this function, and otherwise `return NewHealthCheck(hc), err`

File: pkg/healthchecks/healthchecks.go [hunk 3]
```diff
@@ -321,6 +346,16 @@ func (hc *HealthCheck) ToComputeHealthCheck() (*compute.HealthCheck, error) {
...
+func (hc *HealthCheck) ToBetaComputeHealthCheck() (*computebeta.HealthCheck, error) {
+	// Cannot specify both portSpecification and port field.
+	if len(hc.PortSpecification) > 0 {
+		hc.Port = 0
+	}
+	hc.merge()
+	return toBetaHealthCheck(&hc.HealthCheck)
+}
+
 // ToComputeHealthCheck returns a valid compute.HealthCheck object
```
agau4779: nit, this should also be `ToAlphaComputeHealthCheck`

pkg/healthchecks/healthchecks.go
```diff
@@ -376,9 +424,24 @@ func toV1HealthCheck(hc *computealpha.HealthCheck) (*compute.HealthCheck, error)
...
+
+// toV1HealthCheck converts v1 health check to alpha health check.
+// There should be no information lost after conversion.
+func v1ToAlphaHealthCheck(hc *compute.HealthCheck) (*computealpha.HealthCheck, error) {
+	ret := &computealpha.HealthCheck{}
+	err := copyViaJSON(ret, hc)
+	return ret, err
+}
+
 // toV1HealthCheck converts v1 health check to alpha health check.
```
agau4779: nit, `// betaToAlphaHealthCheck converts beta health check to alpha health check`

pkg/healthchecks/healthchecks.go
```diff
@@ -197,18 +217,23 @@ func (h *HealthChecks) Delete(name string) error {
...
 		}
-		hc, err = toAlphaHealthCheck(v1hc)
+		hc, err = v1ToAlphaHealthCheck(v1hc)
+	} else if version == meta.VersionBeta {
+		betaHC, err := h.cloud.GetBetaHealthCheck(name)
+		if err != nil {
+			return nil, err
+		}
+		hc, err = betaToAlphaHealthCheck(betaHC)
 	}
```
agau4779: Should also check for invalid version here  - `return fmt.Errorf("Unknown Version: %q", hc.Version())`. Since this is used in a few places, it could be a const too.

pkg/healthchecks/healthchecks.go
```diff
@@ -183,7 +203,7 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
 }
 
 func (h *HealthChecks) getHealthCheckLink(name string, alpha bool) (string, error) {
```
agau4779: We need to change the method signature of this function to `version meta.Version` also. Below line should be `h.Get(name, version)`.

pkg/healthchecks/healthchecks.go
anfernee: irrelevant, just out of curiosity, why not `switch .. case ..`  here?

freehan: oh yeah. I had `switch .. case .. ` at other places. Will make the change

PR #375, opened 2018-06-28
Add plumbing to support upgrade testing. 

rramkumar1:
Adds the ability to run a test inside a sandbox continuously. This allows us to test functionality while the ingress is being updated.

### DIFF
--- a/cmd/e2e-test/basic_upgrade_test.go
+++ b/cmd/e2e-test/basic_upgrade_test.go
@@ -28,7 +28,7 @@ import (
 	"k8s.io/ingress-gce/pkg/fuzz/features"
 )
 
-func TestBasic(t *testing.T) {
+func TestBasicUpgrade(t *testing.T) {
 	t.Parallel()
 
 	port80 := intstr.FromInt(80)
@@ -41,14 +41,6 @@ func TestBasic(t *testing.T) {
 		numBackendServices int
 	}{
 		{
-			desc: "http default backend",
-			ing: fuzz.NewIngressBuilder("", "ingress-1", "").
-				DefaultBackend("service-1", port80).
-				Build(),
-			numForwardingRules: 1,
-			numBackendServices: 1,
-		},
-		{
 			desc: "http one path",
 			ing: fuzz.NewIngressBuilder("", "ingress-1", "").
 				AddPath("test.com", "/", "service-1", port80).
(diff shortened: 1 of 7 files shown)


cmd/e2e-test/basic_upgrade_test.go
```diff
@@ -0,0 +1,99 @@
...
+				AddPath("test.com", "/", "service-1", port80).
+				Build(),
+			numForwardingRules: 1,
+			numBackendServices: 2,
+		},
+	} {
+		expectUnreachable := true
+		tc := tc // Capture tc as we are running this in parallel.
+		Framework.RunContinuouslyWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			if s.IsFresh {
```
rramkumar1: Not sure if I want to put the notion of "fresh" inside each sandbox or have the framework keep track of each sandbox's freshness

pkg/e2e/helpers.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 // WaitForIngress to stabilize.
-func WaitForIngress(s *Sandbox, ing *v1beta1.Ingress) (*v1beta1.Ingress, error) {
+func WaitForIngress(s *Sandbox, ing *v1beta1.Ingress, expectUnreachable bool) (*v1beta1.Ingress, error) {
```
MrHohn: Probably create a `WaitOptions` struct?

pkg/e2e/helpers.go
```diff
@@ -53,8 +53,13 @@ func WaitForIngress(s *Sandbox, ing *v1beta1.Ingress) (*v1beta1.Ingress, error)
 		result := validator.Check(context.Background())
 		if result.Err == nil {
 			return true, nil
+		} else {
+			if expectUnreachable {
+				return false, nil
+			} else {
```
MrHohn: Unnecessary else block?

pkg/e2e/helpers.go
```diff
@@ -53,8 +53,13 @@ func WaitForIngress(s *Sandbox, ing *v1beta1.Ingress) (*v1beta1.Ingress, error)
...
+				return true, fmt.Errorf("Unexpected error from validation: %v", result.Err)
```
MrHohn: return false, fmt.Errorf("Unexpected error from validation: %v", result.Err)? `Poll()` will return once there is an error.

cmd/e2e-test/basic_upgrade_test.go
```diff
@@ -0,0 +1,99 @@
...
+		{
+			desc: "http one path",
+			ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+				AddPath("test.com", "/", "service-1", port80).
+				Build(),
+			numForwardingRules: 1,
+			numBackendServices: 2,
+		},
+	} {
+		expectUnreachable := true
```
MrHohn: Why do we need `expectUnreachable=true` at the first run?

cmd/e2e-test/security_policy_test.go
```diff
@@ -111,7 +111,7 @@ func TestSecurityPolicyEnable(t *testing.T) {
 
 		t.Logf("Checking on relevant backend service whether security policy is properly attached")
 
-		testIng, err = e2e.WaitForIngress(s, testIng)
+		testIng, err = e2e.WaitForIngress(s, testIng, true)
```
MrHohn: Humm..Why should it expect unreachable?

cmd/e2e-test/basic_upgrade_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/kr/pretty"
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestBasicUpgrade(t *testing.T) {
```
MrHohn: Will this upgrade test run with the other tests or will we have a explicit upgrade job? How would we trigger the upgrade action?

rramkumar1: If you look in run.sh, I added a couple line to trigger just the upgrade tests if we want.

cmd/e2e-test/basic_upgrade_test.go
rramkumar1: In the first run, we will experience the programming latency for the LB so we expect that we will get back some 404's/502's.

cmd/e2e-test/security_policy_test.go
rramkumar1: See above comment.

pkg/e2e/helpers.go
rramkumar1: I want it to return in this case. If expectUnreachable is false, meaning we expect no 404's/502's, then I want to return the error if there was one.

pkg/e2e/helpers.go
rramkumar1: Done

pkg/e2e/helpers.go
rramkumar1: Done

cmd/e2e-test/basic_upgrade_test.go
MrHohn: Thanks, should have seen that.

pkg/e2e/helpers.go
MrHohn: I see, that makes sense. I was mistakenly thinking expectUnreachable means we expect it to be always unreachable.

rramkumar1:
Discussed offline and will go with a different approach.

/close.

PR #382, opened 2018-06-29
promote http2 to beta

agau4779:


pkg/healthchecks/healthchecks.go
```diff
@@ -399,10 +399,7 @@ func (hc *HealthCheck) isHttp2() bool {
 // Use Alpha API for HTTP2
```
bowei: fix references to Alpha

PR #383, opened 2018-06-29
Add more negative test cases for backend config

MrHohn:
/assign @rramkumar1 @bowei

### DIFF
--- a/cmd/e2e-test/backend_config_test.go
+++ b/cmd/e2e-test/backend_config_test.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/ingress-gce/pkg/annotations"
+	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
 	"k8s.io/ingress-gce/pkg/e2e"
 	"k8s.io/ingress-gce/pkg/fuzz"
 )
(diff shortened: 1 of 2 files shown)


cmd/e2e-test/backend_config_test.go
```diff
@@ -37,56 +38,96 @@ var (
...
+			expectedMsg: fmt.Sprintf("%v", annotations.ErrBackendConfigInvalidJSON),
+		},
+		{
+			desc: "enable both IAP and CDN in backend config",
+			svcAnnotations: map[string]string{
+				annotations.BackendConfigKey: `{"default":"backendconfig-1"}`,
+			},
+			backendConfig: fuzz.NewBackendConfigBuilder("", "backendconfig-1").
+				EnableCDN(true).
+				SetIAPConfig(true, "bar").
```
rramkumar1: Given how the code is written, you will actually have to create the secret "bar". Otherwise, it will return an error saying the secret does not exist, rather than the error we actually are testing for,

pkg/backendconfig/backendconfig.go
```diff
@@ -95,7 +94,7 @@ func GetBackendConfigForServicePort(backendConfigLister cache.Store, svc *apiv1.
 		if err == annotations.ErrBackendConfigAnnotationMissing {
 			return nil, nil
 		}
-		return nil, ErrBackendConfigsFromAnnotation
+		return nil, err
```
rramkumar1: What was the reason for the change here?

cmd/e2e-test/backend_config_test.go
MrHohn: Good catch, I'm seeing this:
```
Error during sync: error while evaluating the ingress spec: BackendConfig test-sandbox-48912bf33631f0fe/backendconfig-1 is not valid: error retrieving secret bar: secrets \"bar\" not found
```
Added the logic to actually create the secret "bar".

pkg/backendconfig/backendconfig.go
MrHohn: It was meant to make the warning event more useful. Without this change, when annotation is in invalid format, user will see something like "error getting backend config" instead of "the annotation is invalid".

PR #387, opened 2018-07-03
Modify security policy e2e test to create unique GCP resources.

rramkumar1:
If this test runs in multiple clusters, then we will see failures since the security policy created will not be unique to that clusters. Adding the sandbox namespace to the end of the name ensures we don't see these errors.

/assign @bowei 
/cc @MrHohn


rramkumar1:
@MrHohn Can you merge. Thanks!

MrHohn:
Thanks for the fix!

PR #392, opened 2018-07-10
Add plumbing for upgrade testing

rramkumar1:
Based on the design we discussed offline (used a ConfigMap to manage the stability statuses of Ingresses in test sandboxes).

Open to thoughts on how to structure the type which manages the ConfigMap (StatusManager)

/assign @bowei


rramkumar1:
#546 is taking this over

/close

PR #393, opened 2018-07-11
Implement security policy validator for real

MrHohn:
This PR teaches security policy validator to actually check for response's status code.
/assign @rramkumar1 @bowei

### DIFF
--- a/pkg/fuzz/features/security_policy.go
+++ b/pkg/fuzz/features/security_policy.go
@@ -17,9 +17,14 @@ limitations under the License.
 package features
 
 import (
+	"context"
+	"fmt"
 	"net/http"
 
 	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce/cloud/meta"
+
+	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/fuzz"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/fuzz/features/security_policy.go
```diff
@@ -63,7 +68,50 @@ func (v *securityPolicyValidator) ConfigureAttributes(env fuzz.ValidatorEnv, ing
...
+	case "allow":
+		expectedCode = 200
+	case "deny(403)":
+		expectedCode = 403
+	case "deny(404)":
+		expectedCode = 404
+	case "deny(502)":
+		expectedCode = 502
+	default:
+		return fuzz.CheckResponseContinue, fmt.Errorf("unregonized rule %q", policy.Rules[0].Action)
```
rramkumar1: Nit: "unrecognized"

pkg/fuzz/helpers.go
```diff
@@ -112,7 +112,10 @@ func ServiceMapFromIngress(ing *v1beta1.Ingress) ServiceMap {
 		for _, path := range rule.HTTP.Paths {
 			hp := HostPath{Host: rule.Host, Path: path.Path}
 			if _, ok := ret[hp]; !ok {
-				ret[hp] = &path.Backend
+				// Copy the value over to a new struct so that we won't be
```
rramkumar1: What was the purpose for this change?

MrHohn: During the test I found we are saving the same pointer for `path.Backend` in this for loop because path is defined in the init statement.

Ref https://golang.org/ref/spec#For_statements: "Variables declared by the init statement are re-used in each iteration."

pkg/fuzz/features/security_policy.go
MrHohn: Thanks, fixed.

PR #402, opened 2018-07-17
uniq function should compare more than NodePort difference

freehan:
cc: @agau4779


pkg/controller/utils.go
```diff
@@ -193,16 +193,16 @@ func setInstanceGroupsAnnotation(existing map[string]string, igs []*compute.Inst
...
 
 // uniq returns an array of unique service ports from the given array.
-func uniq(nodePorts []utils.ServicePort) []utils.ServicePort {
-	portMap := map[int64]utils.ServicePort{}
-	for _, p := range nodePorts {
-		portMap[p.NodePort] = p
+func uniq(svcPorts []utils.ServicePort) []utils.ServicePort {
+	portMap := map[string]utils.ServicePort{}
+	for _, p := range svcPorts {
+		portMap[fmt.Sprintf("%q-%q-%q", p.ID.Service.Namespace, p.ID.Service.Name, p.ID.Port.String())] = p
```
nicksardo: We should probably define a `String()` for ServicePortID while we're at it.  
```
func (id ServicePortID) String() string {
   return fmt.Sprintf("%v/%v", id.Service.String(), id.Port.String())
}
```
yielding `kube-system/default-backend/http`

freehan: Good point. Fixed.

PR #403, opened 2018-07-17
Extract firewall management into separate controller

rramkumar1:
This PR attempts to move the management of the L7 firewall rule into a separate controller. 

Summary of changes:

1. Add controller.go to pkg/firewalls. This new controller also creates and interacts with the firewall pool, rather than the ClusterManager. Currently, the controller can be run in two "modes" depending on the passed in config.

2. Remove all firewall related stuff from ClusterManager and pkg/controller/controller.go

3. Move IsHealthy from ClusterManager to the controller the health check is meant for. It really has not place in ClusterManager.

Open to discussion on some of the design decisions in the PR. /assign @nicksardo


rramkumar1:
Note: This does not compile because of an import cycle. Trying to fix that now, but should not affect review.

cmd/glbc/main.go
```diff
@@ -193,6 +200,9 @@ func runControllers(ctx *context.ControllerContext) {
 	ctx.Start(stopCh)
 	lbc.Run()
```
nicksardo: I believe this blocks...

rramkumar1: Oops..moved it start of fwc above lbc.Run()

pkg/utils/join.go
```diff
@@ -0,0 +1,72 @@
...
+type Joiner struct {
+	ingLister               StoreToIngressLister
+	svcInformer             cache.SharedIndexInformer
+	defaultBackendSvcPortID ServicePortID
+}
+
+// NewJoiner returns a Joiner.
+func NewJoiner(
+	ingressInformer cache.SharedIndexInformer,
+	svcInformer cache.SharedIndexInformer,
```
nicksardo: Suggest passing in the listers instead of the full informers.

pkg/utils/join.go
```diff
@@ -0,0 +1,72 @@
...
+	defaultBackendSvcPortID ServicePortID) *Joiner {
+	ingLister := StoreToIngressLister{Store: ingressInformer.GetStore()}
+	return &Joiner{ingLister, svcInformer, defaultBackendSvcPortID}
```
nicksardo: Actually, I'm not seeing a strong reason why `Joiner` needs to exist as a struct. Seems more flexible to make the last two funcs static and pass in the appropriate lister. Then a consuming controller isn't forced to provide all listers even if it only uses one.

pkg/utils/taskqueue.go
```diff
@@ -29,6 +29,7 @@ import (
 type TaskQueue interface {
 	Run(period time.Duration, stopCh <-chan struct{})
 	Enqueue(obj interface{})
+	EnqueueAll(objs []interface{})
```
nicksardo: Variadic func?
```
Enqueue(obj ...interface{})
```

pkg/firewalls/controller.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+			fwc.queue.Enqueue(cur)
+		},
+	})
+
+	// Service event handlers.
+	ctx.ServiceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			ings := joiner.IngressesForObject(obj)
+			lbc.ingQueue.EnqueueAll(ings)
```
nicksardo: fwc.queue

pkg/utils/taskqueue.go
rramkumar1: Done

pkg/firewalls/controller.go
rramkumar1: Done

pkg/utils/join.go
rramkumar1: Done

pkg/utils/join.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// IngressesForObject gets Ingresses that are associated with the
+// passed in object. It is a wrapper around functions which do the actual
+// work of getting Ingresses based on a typed object.
+func (j *Joiner) IngressesForObject(obj interface{}) []*extensions.Ingress {
```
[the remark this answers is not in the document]
rramkumar1: Ditched the wrapper

pkg/utils/join.go
rramkumar1: So after trying the alternative, I think its much easier to read the code when Joiner exists as a struct. The reason is because the initalization of the struct will take care of storing all necessary things needed (listers, default backend service port ID) and then the function calls become nice and compact. The alternative is that each function call becomes really bloated with all the necessary arguments.

rramkumar1: To alleviate the problem of sometimes passing in stuff you don't need, maybe the Joiner can just take a ControllerContext? That contains everything it needs.

rramkumar1:
Realized there are some things I haven't thought about that I need to add to my implementation. Marking as "WIP" for now.

pkg/firewalls/controller.go
```diff
@@ -0,0 +1,190 @@
...
+		urlMap, _ := fwc.translator.TranslateIngress(&ing, fwc.ctx.DefaultBackendSvcPortID)
+		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+	}
+	return knownPorts
+}
+
+func (fwc *FirewallController) Run(stopCh chan struct{}) {
+	defer fwc.shutdown()
+	go fwc.queue.Run(time.Second, stopCh)
+	<-stopCh
```
nicksardo: If `Run(...)` is supposed to block, is there any need to start the `fwc.queue.Run(...)` in a goroutine? Also, `fwc.shutdown()` calls `firewallPool.Shutdown()` which deletes the firewall. That should only be called when we have no ingresses/NEGs.

```
func (fwc *FirewallController) Run(stopCh chan struct{}) {
	fwc.queue.Run(time.Second, stopCh)
```

pkg/firewalls/controller.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (fwc *FirewallController) shutdown() error {
```
nicksardo: Add a comment that this should only be called when we no longer need a firewall rule.

pkg/utils/join.go
```diff
@@ -0,0 +1,50 @@
...
+		if !IsGCEIngress(&ing) {
+			continue
+		}
+		ingList = append(ingList, &ing)
+	}
+	return
+}
+
+// IngressesForBackendConfig gets all Ingresses that reference (indirectly) a BackendConfig.
+func (j *Joiner) IngressesForBackendConfig(beConfig *backendconfigv1beta1.BackendConfig) (ingList []*extensions.Ingress) {
```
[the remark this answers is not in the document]
rramkumar1: Done

pkg/firewalls/controller.go
rramkumar1: So actually the shutdown() on the controller should only consist of the queue being shutdown. Also, we should only call shutdown() on the controller when the overall process is being terminated right?

pkg/firewalls/controller.go
rramkumar1: Removed firewallPool.Shutdown from the fwc.shutdown() call.

rramkumar1: Now I think the behavior for all these shutdowns is correct. We only call shutdown on the firewall pool if there are no ingresses and we only shutdown the queue when the process itself is terminated

nicksardo:
/lgtm
Needs testing and then someone else can merge.

rramkumar1:
Made some small changes: 

1. Added a util function to convert a list of ingresses into a list of interfaces{}. This ensures that we can use the now variadic Enqueue function
2. Made queueKey a pointer. 

I tested this end-to-end and it works so it should be good to merge.

PR #413, opened 2018-07-25
Rename context's hcLock to lock

rramkumar1:
Makes it easier for other stuff to use the lock :)

/assign @MrHohn

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -54,7 +54,8 @@ type ControllerContext struct {
 	EndpointInformer      cache.SharedIndexInformer
 
 	healthChecks map[string]func() error
-	hcLock       sync.Mutex
+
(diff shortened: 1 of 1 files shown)


pkg/context/context.go
```diff
@@ -152,8 +153,8 @@ type HealthCheckResults map[string]error
 
 // HealthCheck runs all registered healthcheck functions.
 func (ctx *ControllerContext) HealthCheck() HealthCheckResults {
-	ctx.hcLock.Lock()
-	defer ctx.hcLock.Unlock()
+	ctx.locl.Lock()
```
MrHohn: locl -> lock

rramkumar1: Fixed.

PR #417, opened 2018-07-31
Introduce cloud.google.com/app-protocols to eventually replace existing annotation

rramkumar1:
/assign @bowei

### DIFF
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -26,11 +26,14 @@ import (
 )
 
 const (
-	// ServiceApplicationProtocolKey is a stringified JSON map of port names to
-	// protocol strings. Possible values are HTTP, HTTPS
+	// ServiceApplicationProtocolKey and GoogleServiceApplicationProtocolKey
+	// is a stringified JSON map of port names to protocol strings.
+	// Possible values are HTTP, HTTPS and HTTP2.
 	// Example:
 	// '{"my-https-port":"HTTPS","my-http-port":"HTTP"}'
-	ServiceApplicationProtocolKey = "service.alpha.kubernetes.io/app-protocols"
+	// Note: ServiceApplicationProtocolKey will be deprecated.
+	ServiceApplicationProtocolKey       = "service.alpha.kubernetes.io/app-protocols"
+	GoogleServiceApplicationProtocolKey = "cloud.google.com/app-protocols"
 
 	// NEGAnnotationKey is the annotation key to enable GCE NEG.
 	// The value of the annotation must be a valid JSON string in the format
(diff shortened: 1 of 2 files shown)


pkg/annotations/service.go
```diff
@@ -105,9 +108,15 @@ func FromService(obj *v1.Service) *Service {
 // ApplicationProtocols returns a map of port (name or number) to the protocol
 // on the port.
 func (svc *Service) ApplicationProtocols() (map[string]AppProtocol, error) {
-	val, ok := svc.v[ServiceApplicationProtocolKey]
+	var val string
+	var ok bool
+	// First check the new annotation, then fall back to the old one.
```
thockin: Actually, you should check old first and if it exists, use it.  That way an old client can "win" in the face of a rollback.

rramkumar1: Maybe I'm missing something, but in a rollback situation, wouldn't the old annotation win always? My thinking is that if we rollback, the controller no longer "knows" about the new annotation and therefore proceeds as it did before, which is to look for the "old" annotation only.

rramkumar1:
@thockin Even if there was a client rollback, I think this code still holds. For example, if a user replaces the alpha annotation w/ the Google-specific one and then rolled back their service, the old annotation would win given that the new one is not present anymore. 

I think in any case, if both annotations exists and the controller knows that both of them exist, we want the new one to win.

thockin:
Look at it this way.

User has YAML that uses the old annotation.  They `apply` that to the apiserver.

User changes their YAML and also upgrades to the new annotation, removing the old one, and patch it. If any sort of merge patch can add the new but leave the old, we now have both.  This is true for 2 fields (annotations -> real field), but maybe not for annotations.

User changes the value of the new annotation in their YAML and applies it.  Now we have old and new disagreeing.  It doesn't matter how we end up in this situation - you have old and new values that disagree...

If we troll back the controller, we have changed the behavior underneath the user.  If they roll back their YAML and edit the old annotation, it becomes a silent no-op (again, not sure that can happen for 2 annotations).

The general rule is that in such cases, the older has to take precedence.  If the user wants the newer, they have to remove the older explicitly.  I am not sure we can get into this corner with 2 annotations, but consistent philosophy should apply.

rramkumar1:
@thockin Ack. Modified the PR, PTAL.

thockin:
For posterity, 2 annotations CAN get into trouble:


```
$ k get svc hostnames -o go-template='{{.metadata.annotations}}{{"\n"}}'
map[old:value1]

$ k patch svc hostnames -p '{"metadata":{"annotations":{"new":"value"}}}'
service/hostnames patched

$ k get svc hostnames -o go-template='{{.metadata.annotations}}{{"\n"}}'
map[new:value old:value1]

$ k patch svc hostnames -p '{"metadata":{"annotations":{"old":"value2"}}}'
service/hostnames patched

$ k get svc hostnames -o go-template='{{.metadata.annotations}}{{"\n"}}'
map[new:value old:value2]
```

Through a series of unfortunate patches, the controller would see both
values, the user can update old and it would never work.

PR #425, opened 2018-08-07
Push dependency on the GCECloud up out of the neg controller

bowei:


pkg/neg/cloudprovideradapter.go
```diff
@@ -0,0 +1,141 @@
+/*
```
rramkumar1: Were your planning on using this adapter in a follow up?

bowei: it's used...

rramkumar1: I don't see NewAdapter() called anywhere. Maybe I am missing it...

bowei: good catch, added, it got lost in a merge

PR #424, opened 2018-08-07
Extract BackendPool interface into three 3 separate interfaces

rramkumar1:
This PR extracts the BackendPool interface into 3 separate interfaces:

    1. BackendPool - Simple CRUD for backends
    2. Syncer - Sync and GC backends
    3. Linker - Link backends to their associated groups. 

This allows us to break the current BackendPool implementation up into more manageable chunks and each chunk is much more testable. Furthermore, some implementations, namely BackendPool and Syncer, can be easily reused. 

/assign @bowei

PR #428, opened 2018-08-08
Introduce a new interface to encapsulate Ingress sync and controller implementation of the sync

rramkumar1:
This PR introduces one high level sync interface and an interface for controllers to implement the sync. 
The benefit of this PR is that it enforces a simple contract for controllers to follow and makes creating controller mocks much easier. For example, the existing controller in pkg/controller/controller.go would be modified to implement the Controller interface. 

Note that this PR works at a layer above #424. A controller would potentially implement SyncBackends using the new interfaces defined in pkg/backends.

Also note that this abstraction will make it easier to parallelize syncs although the current implementation would need to be rewired a bit.

The second commit attempts to use the new interfaces in pkg/controller the way the code is currently structured.

/assign @bowei

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	compute "google.golang.org/api/compute/v1"
 
 	apiv1 "k8s.io/api/core/v1"
 	extensions "k8s.io/api/extensions/v1beta1"
@@ -44,9 +43,17 @@ import (
 	"k8s.io/ingress-gce/pkg/context"
 	"k8s.io/ingress-gce/pkg/controller/translator"
 	"k8s.io/ingress-gce/pkg/loadbalancers"
+	ingsync "k8s.io/ingress-gce/pkg/sync"
 	"k8s.io/ingress-gce/pkg/tls"
 	"k8s.io/ingress-gce/pkg/utils"
 )
+
+// GarbageCollectionState is created by this controller for the purpose
+// of garbage collecting GCLB resources during the sync of an Ingress.
+type GarbageCollectionState struct {
+	lbNames  []string
+	svcPorts []utils.ServicePort
+}
 
 // LoadBalancerController watches the kubernetes api and adds/removes services
 // from the loadbalancer, via loadBalancerConfig.
@@ -81,6 +88,9 @@ type LoadBalancerController struct {
 	// linker implementations for backends
 	negLinker backends.Linker
 	igLinker  backends.Linker
+
+	// Ingress sync + GC implementation
+	ingSyncer ingsync.Syncer
 }
 
 // NewLoadBalancerController creates a controller for gce loadbalancers.
@@ -111,6 +121,7 @@ func NewLoadBalancerController(
 		negLinker:     backends.NewNEGLinker(backendPool, ctx.Cloud, ctx.ClusterNamer),
 		igLinker:      backends.NewInstanceGroupLinker(instancePool, backendPool, ctx.ClusterNamer),
 	}
+	lbc.ingSyncer = ingsync.NewIngressSyncer(&lbc)
 
 	lbc.ingQueue = utils.NewPeriodicTaskQueue("ingresses", lbc.sync)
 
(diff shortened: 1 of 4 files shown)


pkg/controller/controller.go
```diff
@@ -44,10 +43,18 @@ import (
...
+	ingsync "k8s.io/ingress-gce/pkg/sync"
 	"k8s.io/ingress-gce/pkg/tls"
 	"k8s.io/ingress-gce/pkg/utils"
 )
 
+// GarbageCollectionState is created by this controller for the purpose
+// of garbage collecting GCLB resources during the sync of an Ingress.
+type GarbageCollectionState struct {
+	lbNames     []string
+	allSvcPorts []utils.ServicePort
```
bowei: just call this svcPorts?

pkg/controller/controller.go
```diff
@@ -257,69 +268,45 @@ func (lbc *LoadBalancerController) Stop(deleteAll bool) error {
...
-	lbNames := lbc.ingLister.Store.ListKeys()
-	obj, ingExists, err := lbc.ingLister.Store.GetByKey(key)
-	if err != nil {
-		return err
-	}
-	if !ingExists {
-		glog.V(2).Infof("Ingress %q no longer exists, triggering GC", key)
-		// GC will find GCE resources that were used for this ingress and delete them.
-		return lbc.gc(lbNames, gceSvcPorts)
+// Implements Controller.
```
bowei: fix these to be the right format (all other instances as well)

pkg/controller/controller.go
```diff
@@ -257,69 +268,45 @@ func (lbc *LoadBalancerController) Stop(deleteAll bool) error {
...
+	if errs != nil {
+		return "", fmt.Errorf("error while evaluating the ingress spec: %v", joinErrs(errs))
 	}
+	return urlMap, nil
+}
 
-	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
-	ing, ok := obj.(*extensions.Ingress)
+// Implements Controller.
+func (lbc *LoadBalancerController) SyncBackends(ing *extensions.Ingress, state interface{}) error {
```
bowei: Can we make this take in a utils.GCEURLMap? Do the conversion as close to the place where we get the interface{}

pkg/controller/controller.go
```diff
@@ -257,69 +268,45 @@ func (lbc *LoadBalancerController) Stop(deleteAll bool) error {
...
+		return fmt.Errorf("Expected state type to be GCEURLMap, type was %T", state)
 	}
+	ingSvcPorts := urlMap.AllServicePorts()
 
-	// Garbage collection will occur regardless of an error occurring. If an error occurred,
-	// it could have been caused by quota issues; therefore, garbage collecting now may
-	// free up enough quota for the next sync to pass.
-	if gcErr := lbc.gc(lbNames, gceSvcPorts); gcErr != nil {
-		retErr = fmt.Errorf("error during sync %v, error during GC %v", retErr, gcErr)
+	ports := []int64{}
```
bowei: can we make this a function by itself

pkg/sync/sync.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"errors"
+	"fmt"
+
+	extensions "k8s.io/api/extensions/v1beta1"
+)
+
+// ShortCircuit is an error that can be returned by a Controller to
+// short-circuit the remaining sync processes.
+var ShortCircuit = errors.New("")
```
bowei: ErrSkipBackendSync = errors.New("ingress skip backend sync")

If someone accidentally uses, we should know when it gets printed out.

pkg/controller/controller.go
```diff
@@ -332,12 +319,8 @@ func (lbc *LoadBalancerController) ensureIngress(ing *extensions.Ingress, nodeNa
...
 			return err
 		}
-		return nil
-	}
-
-	// Sync the backends
-	if lbc.backendSyncer.Sync(ingSvcPorts); err != nil {
-		return err
+		// This short-circuit will stop the syncer from moving to next step.
+		return ingsync.ShortCircuit
```
bowei: would prefer to have return tuple of (skipBackendSync bool, err error) instead of special error...

pkg/controller/controller.go
```diff
@@ -358,27 +341,130 @@ func (lbc *LoadBalancerController) ensureIngress(ing *extensions.Ingress, nodeNa
 		}
 	}
 
+	return nil
+}
+
+// Implements Controller.
+func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
```
bowei: again, would like to not have to take in interface{}

pkg/controller/controller.go
```diff
@@ -358,27 +341,130 @@ func (lbc *LoadBalancerController) ensureIngress(ing *extensions.Ingress, nodeNa
...
+	// We expect state to be a GarbageCollectionState
+	gcState, ok := state.(*GarbageCollectionState)
+	if !ok {
+		return fmt.Errorf("Expected state type to be GarbageCollectionState, type was %T", state)
```
bowei: don't capitalize errors

pkg/controller/controller.go
```diff
@@ -358,27 +341,130 @@ func (lbc *LoadBalancerController) ensureIngress(ing *extensions.Ingress, nodeNa
...
+	}
+	if err := lbc.l7Pool.GC(gcState.lbNames); err != nil {
+		return err
+	}
+
+	return nil
+}
 
+// Implements Controller.
+func (lbc *LoadBalancerController) PostProcess(ing *extensions.Ingress) error {
```
bowei: call this UpdateStatus?

pkg/controller/controller.go
```diff
@@ -358,27 +341,130 @@ func (lbc *LoadBalancerController) ensureIngress(ing *extensions.Ingress, nodeNa
...
 	// Get the loadbalancer and update the ingress status.
-	l7, err := lbc.l7Pool.Get(lb.Name)
+	k, err := utils.KeyFunc(ing)
+	if err != nil {
+		return fmt.Errorf("cannot get key for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
```
bowei: ... Ingress %v/%v: %q

pkg/sync/interfaces.go
```diff
@@ -0,0 +1,38 @@
+package sync
```
bowei: I think it's kind of an anti-pattern to put interfaces into their own .go file. We can keep it this way for now.

pkg/controller/controller.go
rramkumar1: Done

pkg/controller/controller.go
rramkumar1: Done

pkg/controller/controller.go
rramkumar1: Done

pkg/controller/controller.go
rramkumar1: I feel like PostProcess is a little more generic. We don't have to necessarily just update the status in the post process routine, we could potentially do other things.

pkg/controller/controller.go
rramkumar1: Done

pkg/controller/controller.go
rramkumar1: My thinking was that if its an interface, we aren't constrained to use one specific type to hold pre-processing data.

pkg/controller/controller.go
rramkumar1: %q for an err?

pkg/controller/controller.go
rramkumar1: So this function would return a (bool, error) and if bool is true, the syncer implementation skips everything else?

pkg/controller/controller.go
bowei: ah misread ignore

pkg/sync/sync.go
bowei: Errors should be named Errxxx

PR #431, opened 2018-08-14
Fix bug in backend syncer where backend service was being created without health check

rramkumar1:
Introduced by #424. Looks like we can't create a BackendService without a health check. Fixed the BackendPool.Create implementation to now take in a health check link.

/assign @bowei


bowei:
Did this get detected in our e2e tests?

Issue #433, opened 2018-08-14
When using NEG services only, the controller still creates instance groups


lbernail:
Hello

We have been testing the NEG feature for a little while and I noticed that even when we only have services using NEGs, the controller still creates the instance groups (they are not attached to the backend-services which only point to NEGs)

Is this expected?

Laurent

freehan:
Based on the current implementation (https://github.com/kubernetes/ingress-gce/blob/master/pkg/instances/instances.go#L86), even if the name ports are empty. Instance groups are still created. 

The `default-http-backend` service under kube-system namespace is still not default to use NEG. Ingress controller will attach this backend (Instance Group) to the LB if the ingress does not include default backend.

lbernail:
What about not calling `ensureInstanceGroupsAndPorts` in `ensureIngress` if the service has the NEG annotation?
We'd still need to handle the default-http-backend case. I'm not very familiar with the code yet but I'm happy to work on a PR if you have ideas on how to handle this

freehan:
Yeah. That is what I plan to do. But need to watch out GC logic for this though.

lbernail:
I just built a controller from master and tested it
It worked great. Thanks!

PR #440, opened 2018-08-16
do not create instance group with NEG backends

freehan:
ref: https://github.com/kubernetes/ingress-gce/issues/433

### DIFF
--- a/pkg/controller/utils.go
+++ b/pkg/controller/utils.go
@@ -86,3 +86,15 @@ func convert(ings []*extensions.Ingress) (retVal []interface{}) {
 	}
 	return
 }
+
+// nodePorts returns the list of uniq NodePort from the input ServicePorts.
+// Only NonNEG service backend need NodePort.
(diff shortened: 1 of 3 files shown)


pkg/controller/utils.go
```diff
@@ -86,3 +86,15 @@ func convert(ings []*extensions.Ingress) (retVal []interface{}) {
 	}
 	return
 }
+
+// getNodePortsUsedByIngress returns the list of uniq NodePort from the input ServicePorts.
+// Only NonNEG service backend need NodePort.
+func getNodePortsUsedByIngress(svcPorts []utils.ServicePort) []int64 {
```
rramkumar1: Can we call this nodePorts()?

lbernail:
Thanks! I can test it on our clusters as soon as this is merged to give you feedback

freehan:
> Aside from the PR, I'm thinking it might be worthwhile to add an e2e test for this in the main k8s repo. Maybe even bake the check into an existing test you have.

That requires those tests to be serialized though.

PR #442, opened 2018-08-20
Remove 'Description' and 'Required' from OpenAPI schema root layer

MrHohn:
This is for fixing an incompatibility issue with https://github.com/kubernetes/features/issues/571.

/assign @rramkumar1 
cc @Liujingfang1

<s>Note: The `Description` entry is manually removed, as adding `omitempty` doesn't seem to stop generating it.</s> Fixed by removing struct comment.

### DIFF
--- a/pkg/apis/backendconfig/v1beta1/zz_generated.openapi.go
+++ b/pkg/apis/backendconfig/v1beta1/zz_generated.openapi.go
@@ -32,7 +32,6 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 		"k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1.BackendConfig": {
 			Schema: spec.Schema{
 				SchemaProps: spec.SchemaProps{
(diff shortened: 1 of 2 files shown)


pkg/apis/backendconfig/v1beta1/zz_generated.openapi.go
```diff
@@ -32,7 +32,6 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 		"k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1.BackendConfig": {
 			Schema: spec.Schema{
 				SchemaProps: spec.SchemaProps{
-					Description: "BackendConfig is a specification for a BackendConfig resource",
```
Liujingfang1: Looks like the description is from line 27 in types.go

MrHohn: Good catch, let me see if removing that works.

MrHohn: Yes it works :) Update PR, thanks!

PR #446, opened 2018-08-22
Add some utility functions to support finalizers

rramkumar1:
Adding some functions needed to support finalizers. This code can easily be reused by other modules.

Much of this code was adapted from https://github.com/kubernetes/kubernetes/pull/54569

/assign @bowei @MrHohn

### DIFF
--- a/pkg/utils/finalizer.go
+++ b/pkg/utils/finalizer.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


pkg/utils/finalizer.go
```diff
@@ -0,0 +1,49 @@
+package utils
```
MrHohn: Need copyright boilerplate here.

pkg/utils/finalizer_test.go
```diff
@@ -0,0 +1,37 @@
+package utils
```
MrHohn: Need copyright boilerplate here.

rramkumar1: Done

pkg/utils/finalizer.go
rramkumar1: Done

PR #452, opened 2018-08-28
fix and refactor on NEG annotation handling

freehan:
Refactor NEG annotation handling
Fix a problem where NEG annotation mal format errors are eaten. 

cc: @agau4779

### DIFF
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -92,6 +92,22 @@ type NegAttributes struct {
 	Name string `json:"name,omitempty"`
 }
 
+// NEGEnabledForIngress returns true if the annotation is to be applied on
+// Ingress-referenced ports
+func (n *NegAnnotation) NEGEnabledForIngress() bool {
+	return n.Ingress
+}
+
+// NEGExposed is true if the service exposes NEGs
+func (n *NegAnnotation) NEGExposed() bool {
+	return len(n.ExposedPorts) > 0
+}
+
+// NEGExposed is true if the service uses NEG
+func (n *NegAnnotation) NEGEnabled() bool {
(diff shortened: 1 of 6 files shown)


pkg/annotations/service.go
```diff
@@ -138,59 +154,28 @@ func (svc *Service) ApplicationProtocols() (map[string]AppProtocol, error) {
...
-		return false
-	}
-	return annotation.Ingress
-}
-
 var (
 	ErrBackendConfigNoneFound         = errors.New("no BackendConfig's found in annotation")
 	ErrBackendConfigInvalidJSON       = errors.New("BackendConfig annotation is invalid json")
 	ErrBackendConfigAnnotationMissing = errors.New("BackendConfig annotation is missing")
+	NEGAnnotationErrorFmt             = "NEG annotation is invalid: %v"
```
rramkumar1: I don't think its worth surfacing the actual formatting error to users. They can easily figure out what's wrong by throwing it into a json parser. That way, in your tests, you don't have to check for a specific error string.

pkg/controller/translator/translator.go
```diff
@@ -79,7 +79,11 @@ func (t *Translator) getServicePort(id utils.ServicePortID) (*utils.ServicePort,
 		return nil, errors.ErrSvcPortNotFound{ServicePortID: id}
 	}
 
-	negEnabled := annotations.FromService(svc).NEGEnabled()
+	var negEnabled bool
+	ok, negAnnotation, err := annotations.FromService(svc).GetNEGAnnotation()
```
rramkumar1: What if there is an error? Where is that handled?

freehan:
Fixed. PTAL

PR #460, opened 2018-08-31
Refactor LoadBalancerPool to use cloud listing snapshotter

rramkumar1:
Currently the L7 pool uses an in memory snapshotter. This is problematic because if the controller crashes, the in memory state is lost and load balancer resources could potentially become orphaned. 

This PR refactors the L7 pool to use the cloud listing snapshotter (similar to pkg/backends). By using the cloud lister, we can fetch all LB names from GCP and perform our GC against those names.

/assign @bowei

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -193,8 +193,8 @@ func (b *Backends) List() ([]interface{}, error) {
 		return nil, err
 	}
 	var ret []interface{}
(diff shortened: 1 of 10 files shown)


bowei:
It feels like we should flag gate the cloud lister vs in-memory so we can disable this if it turns out to be a problem.

rramkumar1:
Closing in favor of #590

PR #468, opened 2018-09-08
Fix typo: permissable -> permissible

mirake:
Signed-off-by: ruicao <ruicao@alauda.io>


rramkumar1:
@mirake It would be great if you could add a real bug fix to this change.

PR #475, opened 2018-09-14
Update README.md

wangxy518:
the url is a 404 link, and there is no "services-firewalls.md" in the   HEAD/docs/user-guide path.

### DIFF
--- a/examples/health-checks/README.md
+++ b/examples/health-checks/README.md
@@ -10,13 +10,11 @@ You have exposed your service on an external port on all nodes in your
 cluster.  If you want to expose this service to the external internet, you may
 need to set up firewall rules for the service port(s) (tcp:31165) to serve traffic.
 
-See http://releases.k8s.io/HEAD/docs/user-guide/services-firewalls.md for more details.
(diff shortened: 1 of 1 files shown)


rramkumar1:
@wangxy518 Can you replace the link you removed with a valid one?

wangxy518:
@rramkumar1  I looked for the new link, but I couldn't find it. I think it was deleted instead of replaced.

rramkumar1:
@wangxy518 Can you add a link to the GCP documentation then? 

I'd prefer not to remove links without replacing them.

idealhack:
The links are part of an output of a command, so we should update that output, no matter whether it contains any links.

rramkumar1:
@idealhack Those links are not output of a command, unless I am missing something.

It might look that way because the test underneath those links is stuff from a command output.

rramkumar1:
I'm going to go ahead and close this PR.

We have alot of docs fixes that will be coming in the next few weeks so this will probably get addressed in those fixes.

/close

PR #482, opened 2018-09-19
add work queue to process endpoint changes

freehan:
This makes the informer processing lockless. It used to depend on SyncerManager internal locks.

cc:  @agau4779

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -64,6 +64,8 @@ type Controller struct {
 
 	// serviceQueue takes service key as work item. Service key with format "namespace/name".
 	serviceQueue workqueue.RateLimitingInterface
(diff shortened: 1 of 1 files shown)


pkg/neg/controller.go
```diff
@@ -170,6 +173,7 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	}()
 
 	go wait.Until(c.serviceWorker, time.Second, stopCh)
+	go wait.Until(c.endpointWorker, time.Second, stopCh)
```
bowei: expose the sync interval as a flag

bowei:
expose your interval constants as flags

I won't block this merge based on that though.

/lgtm

PR #484, opened 2018-09-19
Bump Alpine base image version

awly:
Alpine fixed a RCE vulnerability in apk in 3.8.1 (https://alpinelinux.org/posts/Alpine-3.8.1-released.html)
Update the base image used for ingress-gce.

### DIFF
--- a/Dockerfile.fuzzer
+++ b/Dockerfile.fuzzer
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.6
+FROM alpine:3.8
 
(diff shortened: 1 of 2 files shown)


awly:
Ping @bowei

bowei:
we can probably just use scratch and avoid the vulnerability stuff altogether?

awly:
That would be ideal, although it looks like ca-certificates are needed. I'm not aware of a vetted scratch+ca-certs image.
If it's easy to do, please send a separate PR and I'll close this one. Otherwise, we can do this quick patch and file an issue for switching to scratch.

PR #489, opened 2018-09-21
Restructure NEG controller

freehan:
No functional changes. Only refactoring: 

- Created sub-package for `/metrics` and `/syncer`
- Move common types and interfaces into `/types`
- Rename `syncer` to `batchSyncer`

cc: @agau4779

### DIFF
--- a/pkg/neg/syncer/batch.go
+++ b/pkg/neg/syncer/batch.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package neg
+package syncer
 
 import (
 	"fmt"
@@ -33,6 +33,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/ingress-gce/pkg/neg/metrics"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce"
 )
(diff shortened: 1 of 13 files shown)


pkg/neg/interfaces.go
```diff
@@ -17,36 +17,15 @@ limitations under the License.
 package neg
```
agau4779: Does there need to be an interfaces file at all? Could the below types just live in `neg/types` and `neg/syncer`?

pkg/neg/syncer/batch.go
```diff
@@ -46,25 +48,26 @@ const (
...
+	Name       string
+	Port       int32
+	TargetPort string
 }
 
-// syncer handles synchorizing NEGs for one service port. It handles sync, resync and retry on error.
-type syncer struct {
-	servicePort
+// batchSyncer handles synchorizing NEGs for one service port. It handles sync, resync and retry on error.
+// It syncs NEG in batch and waits for all operation to complete before continue to the next batch.
```
agau4779: This is called `batch` because we're `batch`-processing multiple NetworkEndpoints right? Can we make a note of that?

pkg/neg/syncer/batch.go
```diff
@@ -46,25 +48,26 @@ const (
 	maxRetryDelay = 600 * time.Second
```
agau4779: this should probably be a flag

pkg/neg/syncer/batch.go
```diff
@@ -25,14 +25,16 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	compute "google.golang.org/api/compute/v0.beta"
```
agau4779: if we're removing the `compute ` prefix here, we should do it for all files touched

pkg/neg/interfaces.go
freehan: All the interfaces here are internal interfaces for controller and manager.
The rest are shared.

pkg/neg/syncer/batch.go
freehan: hmmm. Not sure why this is removed accidentally. Will add it back.

pkg/neg/syncer/batch.go
freehan: I could not think of a better name. Calling it batch just because it process the NEG changes in batch and wait for all operation to finish before continuing.

pkg/utils/utils.go
```diff
@@ -438,3 +438,7 @@ func TraverseIngressBackends(ing *extensions.Ingress, process func(id ServicePor
 	}
 	return
 }
+
+func ServiceKeyFunc(namespace, name string) string {
```
bowei: docstring

PR #499, opened 2018-09-27
Add a Backofff Handler utils

freehan:
Extract the backoff retry logic from `batchSyncer` and write 2 util package that can be reused by different syncers, plus unit testing:
- `backoffHandler` calculates exponential backoff delays
- `backoffRetryHandler` triggers `retryFunc` after back off delay.

These utils are building blocks for the new `transactionSyncer`

cc: @agau4779

### DIFF
--- a/pkg/neg/syncers/backoff.go
+++ b/pkg/neg/syncers/backoff.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncers
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var ErrRetriesExceeded = fmt.Errorf("maximum retry exceeded")
+
+// backoffHandler handles delays for back off retry
+type backoffHandler interface {
+	// NextRetryDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextRetryDelay() (time.Duration, error)
+	// ResetRetryDelay resets the retry delay
+	ResetRetryDelay()
+}
+
+// exponentialBackOffHandler is a backoff handler that returns retry delays semi-exponentially with random jitter within boundary.
+// exponentialBackOffHandler returns ErrRetriesExceeded when maximum number of retries has reached.
+type exponentialBackOffHandler struct {
+	lock           sync.Mutex
+	lastRetryDelay time.Duration
+	retryCount     int
+	maxRetries     int
+	minRetryDelay  time.Duration
+	maxRetryDelay  time.Duration
+}
+
+func NewExponentialBackendOffHandler(maxRetries int, minRetryDelay, maxRetryDelay time.Duration) *exponentialBackOffHandler {
+	return &exponentialBackOffHandler{
+		lastRetryDelay: time.Duration(0),
+		retryCount:     0,
+		maxRetries:     maxRetries,
+		minRetryDelay:  minRetryDelay,
+		maxRetryDelay:  maxRetryDelay,
+	}
+}
+
+// NextRetryDelay returns the next back off delay for retry.
+func (handler *exponentialBackOffHandler) NextRetryDelay() (time.Duration, error) {
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+	handler.retryCount += 1
+	if handler.maxRetries > 0 && handler.retryCount > handler.maxRetries {
+		return 0, ErrRetriesExceeded
+	}
+	handler.lastRetryDelay = wait.Jitter(handler.lastRetryDelay, 1.0)
+	if handler.lastRetryDelay < handler.minRetryDelay {
+		handler.lastRetryDelay = handler.minRetryDelay
+	} else if handler.lastRetryDelay > handler.maxRetryDelay {
+		handler.lastRetryDelay = handler.maxRetryDelay
+	}
+	return handler.lastRetryDelay, nil
+}
+
+// ResetRetryDelay resets the retry delay.
+func (handler *exponentialBackOffHandler) ResetRetryDelay() {
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+	handler.retryCount = 0
+	handler.lastRetryDelay = time.Duration(0)
+}
--- a/pkg/neg/syncers/retry_handler.go
+++ b/pkg/neg/syncers/retry_handler.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncers
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/clock"
+)
+
+var ErrHandlerRetrying = fmt.Errorf("retry handler is retrying")
+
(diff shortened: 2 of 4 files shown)


pkg/neg/syncers/backoff.go
```diff
@@ -0,0 +1,78 @@
...
+
+package syncers
+
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+var MaxRetryError = fmt.Errorf("maximum retry exceeded")
```
bowei: ErrRetriesExceeded

pkg/neg/syncers/backoff.go
```diff
@@ -0,0 +1,78 @@
...
+
+// backoffHanlder handles delays for back off retry
+type backoffHanlder interface {
+	// NextRetryDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextRetryDelay() (time.Duration, error)
+	// ResetRetryDelay resets the retry delay
+	ResetRetryDelay()
```
bowei: Alternative: construct a new backoffHandler instead of reset

pkg/neg/syncers/backoff.go
```diff
@@ -0,0 +1,78 @@
...
+		lastRetryDelay: time.Duration(0),
+		retryCount:     0,
+		maxRetries:     maxRetries,
+		minRetryDelay:  minRetryDelay,
+		maxRetryDelay:  maxRetryDelay,
+	}
+}
+
+// NextRetryDelay returns the next back off delay for retry.
+func (handler *boundedExponentialBackOffHandler) NextRetryDelay() (time.Duration, error) {
```
bowei: We should probably add jitter. Does the code vendored in from client-go offer any resuable bits?

freehan:
Fixed the comments. Ready for review

pkg/neg/syncers/backoff.go
freehan: That basically push more load on Golang GC. Imagine I need to create a new backoffHandler object after every successful NEG sync.

pkg/neg/syncers/backoff.go
```diff
@@ -0,0 +1,81 @@
...
+		maxRetryDelay:  maxRetryDelay,
+	}
+}
+
+// NextRetryDelay returns the next back off delay for retry.
+func (handler *exponentialBackOffHandler) NextRetryDelay() (time.Duration, error) {
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+	handler.retryCount += 1
+	if handler.maxRetries > 0 && handler.retryCount > handler.maxRetries {
```
agau4779: nit - don't need the `> 0`

pkg/neg/syncers/retry_handler_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (h *retryHandlerTestHelper) getCount() int {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+	return h.count
+}
+
+func TestStartAndStopBackoffRetryHandler(t *testing.T) {
+	helper := &retryHandlerTestHelper{}
+	handler := NewDelayRetryHandler(helper.incrementCount, NewExponentialBackendOffHandler(15, testMinRetryDelay, testMaxRetryDelay))
```
agau4779: This and line 88 - could we put the max retries in a variable at the top? `const testMaxRetries = 10`?

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// backoff delay handling
+	clock   clock.Clock
+	backoff backoffHandler
+
+	// retryFunc called on retry
+	retryFunc func()
+}
+
+func NewDelayRetryHandler(retryFunc func(), backoff backoffHandler) *backoffRetryHandler {
```
MrHohn: To make sure I understand it correct:
- Each NEG can get one retry handler.
- `handler.Run()` starts before any NE operation for this NEG is made.
- Each NEG can trigger multiple go routines that handle NE operations in parallel. Some of them may fail at different time.
- All failed operations will attempt to call `handler.Retry()`. Conceptually all calls to `handler.Retry()` will be merged into one until the next retry runs.
- Each `retryFunc()` can trigger go routines that execute NE operations necessary for the entire NEG based on its current state.

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// backoffRetryHandler handles retry with back off delay
+// At any time, there is only one ongoing retry flow running
+type backoffRetryHandler struct {
+	// stateLock protects internal states and channels
+	stateLock sync.Mutex
+
+	// internal states
+	stopped  bool
```
MrHohn: It feels like we should use `context.WithCancel()` instead of a bool flag `stopped`. So upon `Stop()` is called we can revoke the on-going retry immediately instead of wait till the next retry delay ends.

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	go func() {
+		nextRetry := h.clock.After(delay)
+		<-nextRetry
+		func() {
+			h.stateLock.Lock()
+			defer h.stateLock.Unlock()
+			if !h.stopped {
+				h.ch <- struct{}{}
+			}
+			h.retrying = false
```
MrHohn: Probably put this after `retryFunc()` is executed?

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+
+	h.retrying = true
+	go func() {
+		nextRetry := h.clock.After(delay)
+		<-nextRetry
+		func() {
+			h.stateLock.Lock()
+			defer h.stateLock.Unlock()
+			if !h.stopped {
+				h.ch <- struct{}{}
```
MrHohn: Giving another thought on how to trigger retry. Given there can only be one on-going retry, why making this asynchronous? How is it different than just calling `retryFunc` here?

pkg/neg/syncers/backoff.go
bowei: This is usually not the case for a modern GC (efficiency issues due to GC), unless you are creating an incredible # of objects.

bowei: Not having reset() usually makes it easier to reason about the object.

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+			h.retrying = false
+		}()
+
+	}()
+	return nil
+}
+
+// Reset resets internal back off delay handler
+func (h *backoffRetryHandler) Reset() {
```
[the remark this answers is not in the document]
freehan: Reset is mainly for resting the exponential backoff delay.

pkg/neg/syncers/backoff.go
freehan: Imagine there are multiple go routines running NEG API calls. And when they finished succesfully, every one will call Reset. Or, they need to create duplicated objects.

pkg/neg/syncers/backoff.go
freehan: `> 0` is for the case where one do not want to have a maxRetries limit

pkg/neg/syncers/retry_handler.go
freehan: I am not sure about that. That makes the behavior of this retry handler unpredictable. What if it is in the middle of the retryFunc?

The client probably want this: 1. call `Stop()`. 2. if there is still ongoing retry, let it finish gracefully

pkg/neg/syncers/retry_handler.go
freehan: Yes. 

One more thing, For successful operations, they will call Reset(). 

If some are successful and some failed, the successful operations will not be retried. Eventually, only the failed ones will be retries plus back off exponentially.

pkg/neg/syncers/retry_handler.go
freehan: Yes. You are right. Let me try to simplify this.

pkg/neg/syncers/retry_handler.go
freehan: Let me see if I can merge this into one go routine.

pkg/neg/syncers/retry_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	h.stateLock.Lock()
+	defer h.stateLock.Unlock()
+	if !h.stopped {
+		close(h.ch)
+		h.stopped = true
+	}
+}
+
+// IsStopped returns if handler is stopped
+func (h *backoffRetryHandler) IsStopped() bool {
```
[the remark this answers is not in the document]
freehan: There is use case for this in NEG syncer

pkg/neg/syncers/backoff.go
freehan: Added jitter.

pkg/neg/syncers/retry_handler.go
MrHohn: Nice, I saw this flag was eliminated and the overall structure is much simpler.

MrHohn:
The retry handler part LGTM.

freehan:
Can I get this merged?

freehan:
Ping

PR #508, opened 2018-10-02
Implement support for ManagedCertificate CRD

krzykwas:
This change adds to Ingress support for ManagedCertificate CRD to integrate with https://github.com/GoogleCloudPlatform/gke-managed-certs


krzykwas:
I signed it

krzykwas:
assign @bowei

cmd/glbc/main.go
```diff
@@ -82,6 +83,11 @@ func main() {
 		glog.Fatalf("Failed to create kubernetes client: %v", err)
 	}
 
+	mcrtClient, err := managedcertificatesclient.NewForConfig(kubeConfig)
```
rramkumar1: Can you add a short comment here stating that the ManagedCerts CRD is created + updated by another component.

pkg/controller/controller_test.go
```diff
@@ -43,6 +44,13 @@ var (
 	clusterUID      = "aaaaa"
 )
 
+type recorderProducerMock struct {
+}
+
```
rramkumar1: Can you move this mock to pkg/events.

pkg/loadbalancers/utils.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"strings"
+)
+
+const (
+	separator = ","
+)
+
+// splitAnnotation splits annotation by separator and trims whitespace
```
rramkumar1: Can you put this in pkg/utils instead?

pkg/loadbalancers/certificates.go
```diff
@@ -82,46 +114,89 @@ func (l *L7) checkSSLCert() error {
...
-		return true, fmt.Errorf("PreSharedCert errors - %s", strings.Join(failedCerts, ","))
+		return result, fmt.Errorf("Errors - %s", strings.Join(failedCerts, ","))
 	}
-	return true, nil
+
+	return result, nil
+}
+
+// getManagedCertificates fetches SslCertificates specified via managed-certificates annotation.
+func (l *L7) getManagedCertificates() ([]*compute.SslCertificate, error) {
```
rramkumar1: Is there a unit test for this?

pkg/controller/controller_test.go
rramkumar1: In the same file as the interface definition is fine.

cmd/glbc/main.go
krzykwas: Done. Actually it is created by a user, not by managed-certificate-controller, but the status is set by that controller.

pkg/controller/controller_test.go
krzykwas: Done.

pkg/loadbalancers/utils.go
krzykwas: Done.

pkg/loadbalancers/certificates.go
krzykwas: Last 3 test cases in loadbalancer_test.go test managed certificates, managed certificates + k8s secrets, managed certificates + pre-shared-cert.

cmd/glbc/main.go
rramkumar1: IIUC, the CRD is created by the controller and the CR (custom resource) is created by the user?

pkg/loadbalancers/certificates.go
rramkumar1: Do you mind adding a test for managed certificates + k8s secrets + pre-shared-cert?

rramkumar1: It would also be great if you could add a transition test (i.e add managed cert but then remove it & verify behavior)

cmd/glbc/main.go
krzykwas: I changed the comment slightly, it's not CRD status but CR status.

My idea was to deploy the CRD via kubectl create -f (or something similar) together with managed-certificate-controller in GKE configure/startup scripts whenever managed-certificate-controller gets included in ingress controller manifest (so in alpha clusters for now). Is it a proper approach?

Users create CRs to say „I want to have ssl for example.com”. managed-certificate-controller creates SslCertificate based on CR and sets its status accordingly. Ingress reads the status and adds SslCertificate to TargetProxy.

pkg/loadbalancers/certificates.go
krzykwas: Done both.

bowei:
can you rebase and squash the commits, then split into:

- vendor/ only changes
- files changed for vendoring
- implementation
- unit tests

krzykwas:
I rebased, squashed and split the commits into vendor/+Gopkg.lock changes, implementation and unit tests. I didn't get what files changed for vendoring means.

pkg/loadbalancers/certificates.go
```diff
@@ -23,23 +23,55 @@ import (
...
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils"
 )
 
+const SslCertificateMissing = "SslCertificateMissing"
+
 func (l *L7) checkSSLCert() error {
-	// Handle Pre-Shared cert and early return if used
-	if used, err := l.usePreSharedCert(); used {
+	// Handle annotation managed-certificates
```
rramkumar1: So the previous behavior was that if the user specified both pre-shared certs and secrets, we would only take the pre-shared certs and ignore the secrets. Correct me if I am wrong, but here it seems like you accept all three (managed, pre-shared and secret). 

I would prefer to stick with the existing semantics. Specifically, Managed certs should take precedence over everything else.

pkg/utils/annotation.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"strings"
+)
+
+const (
+	separator = ","
+)
+
+// SplitAnnotation splits annotation by separator and trims whitespace
```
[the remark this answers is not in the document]
krzykwas: I don't have any strong opinion on this one and can inline this function if you think this would be better here. The arguments to have it separately are:
- this is a piece of logic used in two places - here it is separately tested, it would be harder to do it if it was interleaved with other code,
- to me it is easier to understand what the call sites of SplitAnnotations do if the parsing logic is extracted,
- it wasn't obvious to me what strings.Split("", ";") does. Namely, it returns []string{""} instead of []string{} or nil. In this case it doesn't really matter, because the annotation is being tested for being non-empty before SplitAnnotation is called, but if an empty annotation somehow gets through, the code handling pre-shared-cert in my current implementation would try to fetch SslCertificate called "", and that would generate an Event.

pkg/loadbalancers/certificates.go
krzykwas: I think that the current code either uses managed certificates and pre-shared-cert or managed certificates and k8s secrets. I think it does not support all 3 types at once. This is because of return at line 51.:
```
if used {
	l.sslCerts = append(managedSslCerts, preSharedSslCerts...)
	return err
}
```
Also I added a test case for that to loadbalancer_test.go (one of last two test cases).

pkg/events/events.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
prameshj: nit : Change to 2018?

pkg/loadbalancers/certificates.go
prameshj: Any reason we want to mix the modes? If we support both managed certs and preshared together, user might be confused why preshared and secrets don't work together or why all 3 are not picked up. Supporting just one mode with a priority in case more than one is specified might be clearer?

prameshj: I guess one reason to combine preshared and managed certs is because neither of them is ingress created and hence do not need to be garbage-collected. I am curious to see if you had any other reasons, and also motivation for supporting managed certs and secrets .

pkg/loadbalancers/certificates.go
```diff
@@ -135,28 +210,32 @@ func getMapfromCertList(certs []*compute.SslCertificate) map[string]*compute.Ssl
 	return certMap
 }
 
-func (l *L7) populateSSLCert() error {
-	l.sslCerts = make([]*compute.SslCertificate, 0)
+// getExistingSecretsSslCerts fetches SslCertificate resources created and managed by this load balancer
+// instance. These SslCertificate resources were created based on kubernetes secrets in Ingress
+// configuration.
+func (l *L7) getExistingSecretsSslCerts() ([]*compute.SslCertificate, error) {
```
prameshj: Maybe call this "getIngressManagedSslCerts()" ?

pkg/utils/annotation.go
```diff
@@ -0,0 +1,38 @@
...
+// from each resulting token
+func SplitAnnotation(annotation string) []string {
+	var result []string
+	for _, token := range strings.Split(annotation, separator) {
+		if token != "" {
+			result = append(result, strings.TrimSpace(token))
```
prameshj: Is it possible for token to be " ", in which case we will still append an empty string to result since trimspace will make it "" ?
Maybe just check the result of TrimSpace for empty string?

pkg/utils/annotation_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestSplitAnnotation(t *testing.T) {
+	testCases := []struct {
+		in  string
+		out []string
+	}{
+		{"", nil},
+		{"foo", []string{"foo"}},
+		{" foo ", []string{"foo"}},
+		{"foo,bar", []string{"foo", "bar"}},
+		{" foo , bar ", []string{"foo", "bar"}},
```
prameshj: For the case pointed out in the above comment, maybe we can add a test case - "foo, , bar"

pkg/loadbalancers/l7.go
```diff
@@ -50,6 +54,10 @@ type L7RuntimeInfo struct {
 	TLS []*TLSCerts
 	// TLSName is the name of the preshared cert to use. Multiple certs can be specified as a comma-separated string
 	TLSName string
+	// Ingress is the processed Ingress API object.
+	Ingress *extensions.Ingress
+	// ManagedCertificates is a comma-separated list of managed SSL certificates to use.
+	ManagedCertificates string
```
prameshj: Most likely I am not understanding this fully.  If ManagedCertificate names are porvided as config, are we adding a client to watch for managed certs just to track deletions?

pkg/loadbalancers/certificates.go
krzykwas: Only with modes combined it is possible to support no-downtime migration scenarios. GCLB has its own logic for selecting a certificate and Ingress really should just pass all the certificates down to GCLB. But pre-shared-cert and secrets are already implemented differently and I just wanted to keep this behavior unchanged.

pkg/events/events.go
krzykwas: ACK - please leave it unresolved and once we agree what changes are necessary I'll take a look to fix it.

pkg/loadbalancers/certificates.go
krzykwas: ACK, I can rename later when introducing all required changes.

pkg/utils/annotation.go
krzykwas: Good catch, thanks. I'll also fix it once we agree what changes are necessary.

pkg/utils/annotation_test.go
krzykwas: ACK, will do.

pkg/loadbalancers/l7.go
krzykwas: pre-shared-cert is a collection of names of GCP SslCertificate resources to use for SSL termination.

ManagedCertificates are one level of indirection above. The annotation managed-certificates gives a list of names of ManagedCertificate CR. Each of them in its status provides a name of SslCertificate to use. Client is needed to fetch the objects and get the statuses.

pkg/utils/annotation.go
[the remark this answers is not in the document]
rramkumar1: Fair enough, lets leave it like this.

pkg/loadbalancers/certificates.go
rramkumar1: As discussed offline, I think it would make sense to keep consistency with the existing behavior. Namely, only take one form of cert. In this case, the precedence would be managed cert > pre-shared > secret.

pkg/utils/annotation_test.go
krzykwas: Done.

rramkumar1:
/lgtm
/hold

@prameshj if this looks good to you, we can merge.

pkg/loadbalancers/certificates.go
```diff
@@ -82,46 +108,89 @@ func (l *L7) checkSSLCert() error {
...
+	mcrts, err := l.mcrt.ManagedCertificates(l.runtimeInfo.Ingress.Namespace).List(sel)
+	if err != nil {
+		return nil, true, err
+	}
+
+	var names []string
+	for _, mcrt := range mcrts {
+		if mcrt.Status.CertificateName != "" {
+			names = append(names, mcrt.Status.CertificateName)
+		}
```
prameshj: Is it valid to have managed cert status with empty Certificate name? Should we log an error in this case?

pkg/loadbalancers/certificates.go
```diff
@@ -23,23 +23,49 @@ import (
...
-	l.oldSSLCerts = l.sslCerts
-	l.sslCerts = make([]*compute.SslCertificate, 0)
+	l.oldSSLCerts = existingSecretsSslCerts
+	secretsSslCerts, err := l.createSecretsSslCertificates(existingSecretsSslCerts)
+	l.sslCerts = secretsSslCerts
+	return err
+}
+
+// createSecretsSslCertificates creates SslCertificates based on kubernetes secrets in Ingress configuration.
+func (l *L7) createSecretsSslCertificates(existingCerts []*compute.SslCertificate) ([]*compute.SslCertificate, error) {
```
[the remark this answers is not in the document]
krzykwas: Done.

pkg/loadbalancers/certificates.go
```diff
@@ -82,46 +108,89 @@ func (l *L7) checkSSLCert() error {
...
+	}
+
+	mcrtsNames := utils.SplitAnnotation(l.runtimeInfo.ManagedCertificates)
+	req, err := labels.NewRequirement("metadata.name", selection.In, mcrtsNames)
+	if err != nil {
+		return nil, true, err
+	}
+
+	sel := labels.NewSelector()
+	sel.Add(*req)
```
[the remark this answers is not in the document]
krzykwas: I think it can't based on the implementation of apimachinery labels, and also based on the Go convention for constructors I observed: err != nil or err == nil and an object is instantiated.

pkg/loadbalancers/certificates.go
krzykwas: When a user creates a ManagedCertificate, he wouldn't usually fill in the status field, so the CertificateName field can be empty and it is not an error.

fproulx-eoscanada:
@rramkumar1 when can we expect this to be available in prod ?
Also, how and when are those releases made available to Kubernetes Engine, do I need to wait for an update of my Master for it to be available ?

fproulx-eoscanada:
@krzykwas when can we expect this to be available in prod ?
Also, how and when are those releases made available to Kubernetes Engine, do I need to wait for an update of my Master for it to be available ?

PR #509, opened 2018-10-02
Add Syncer Skeleton

freehan:
The syncer skeleton contains all the mechanism that handles syncer life cycle, including
- Start/Stop
- Sync signal handling, reporting error and retry handling
- The syncer only requires the extension struct to provide the sync function. 

cc @agau4779

### DIFF
--- a/pkg/neg/syncers/syncer.go
+++ b/pkg/neg/syncers/syncer.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncers
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+)
+
+// syncer is a NEG syncer skeleton.
+// It handles state transitions and backoff retry operations.
+type syncer struct {
+	// metadata
+	NegSyncerKey
+	negName string
+
+	// NEG sync function
+	syncFunc func() error
+
+	// event recording
+	serviceLister cache.Indexer
+	recorder      record.EventRecorder
+
+	// syncer states
+	stateLock    sync.Mutex
+	stopped      bool
+	shuttingDown bool
+
+	// sync signal and retry handling
+	syncCh  chan interface{}
+	clock   clock.Clock
+	backoff backoffHandler
+}
+
+func newSyncer(negSyncerKey NegSyncerKey, networkEndpointGroupName string, serviceLister cache.Indexer, recorder record.EventRecorder) *syncer {
+	return &syncer{
+		NegSyncerKey:  negSyncerKey,
+		negName:       networkEndpointGroupName,
+		syncFunc:      func() error { return nil },
+		serviceLister: serviceLister,
+		recorder:      recorder,
+		stopped:       true,
+		shuttingDown:  false,
+		clock:         clock.RealClock{},
+		backoff:       NewExponentialBackendOffHandler(maxRetries, minRetryDelay, maxRetryDelay),
+	}
+}
+
+func (s *syncer) Start() error {
+	if !s.IsStopped() {
+		return fmt.Errorf("NEG syncer for %s is already running.", s.NegSyncerKey.String())
+	}
+	if s.IsShuttingDown() {
+		return fmt.Errorf("NEG syncer for %s is shutting down. ", s.NegSyncerKey.String())
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/golang/glog"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+)
+
+// noopNegSyncer is a NEG syncer template.
+// It handles state transitions and backoff retry operations.
+type noopNegSyncer struct {
```
rramkumar1: Not the biggest fan of calling this noopNegSyncer.

I would just call this Syncer.

pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	stopped      bool
+	shuttingDown bool
+
+	// sync signal and retry handling
+	syncCh  chan interface{}
+	clock   clock.Clock
+	backoff backoffHandler
+}
+
+func newNoopSyncer(negSyncerKey NegSyncerKey, networkEndpointGroupName string, serviceLister cache.Indexer, recorder record.EventRecorder) *noopNegSyncer {
```
rramkumar1: Would it make more sense to pass in a RecordProducer? (see pkg/events)

Right now the ControllerContext implements that interface so in whatever code calls newNoopSyncer, you can just pass a context.

pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+)
+
+// noopNegSyncer is a NEG syncer template.
+// It handles state transitions and backoff retry operations.
+type noopNegSyncer struct {
+	// metadata
+	NegSyncerKey
+	negName string
```
rramkumar1: Looks like NegSyncerKey contains the name so how come you need negName?

pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	// NEG sync function
+	syncFunc func() error
+
+	// event recording
+	serviceLister cache.Indexer
+	recorder      record.EventRecorder
+
+	// syncer states
+	stateLock    sync.Mutex
+	stopped      bool
```
rramkumar1: I would suggest abstracting these states into a type like:

type SyncState int

const (
   StateStopped SyncState = iota
   StateShuttingDown SyncState 
   ...
)

pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	defer s.stateLock.Unlock()
+	if !s.stopped {
+		glog.V(2).Infof("Stopping NEG syncer for service port %s", s.NegSyncerKey.String())
+		s.stopped = true
+		s.shuttingDown = true
+		close(s.syncCh)
+	}
+}
+
+func (s *noopNegSyncer) Sync() bool {
```
rramkumar1: Who would call Sync()?

pkg/neg/syncers/noop.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	select {
+	case s.syncCh <- struct{}{}:
+		return true
+	default:
+		return false
+	}
+}
+
+func (s *noopNegSyncer) IsStopped() bool {
```
rramkumar1: Do these need to be exposed?

pkg/neg/syncers/noop_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func TestRetryOnSyncError(t *testing.T) {
+	syncer := newTestNoopNegSyncer()
+	syncer.SyncError = true
+	if err := syncer.Start(); err != nil {
+		t.Fatalf("Failed to start syncer: %v", err)
+	}
+
+	if err := wait.PollImmediate(time.Second, 30*time.Second, func() (bool, error) {
```
rramkumar1: My suggestion would be to make the retry handler an interface and then implement a fake which retries at a much faster pace. That way, unit test completion won't be blocked on this.

pkg/neg/syncers/noop.go
freehan: `NegSyncerKey` contains the name, namespace, port of the service

pkg/neg/syncers/noop.go
freehan: SyncerManager

pkg/neg/syncers/noop.go
freehan: Yes. Because the SyncerManager calls it.

pkg/neg/syncers/noop.go
freehan: okay. Will change to use ControllerContext

pkg/neg/syncers/noop.go
freehan: But then I only have one variable to store state, right?

when `state == ShuttingDown`. Should `IsStopped` return true?

Otherwise, it would be the same as having 2 booleans.

pkg/neg/syncers/noop.go
freehan: SGTM

pkg/neg/syncers/noop_test.go
freehan: Add a block/unblock mechanism for the tester. 

But it still needs to poll for a few seconds just to make sure the other go routine can execute completely. This is to make sure the test does not flak.

freehan:
Ready for another round.

pkg/neg/syncers/noop.go
freehan: Yes. That sounds good.

PR #512, opened 2018-10-16
WIP feat: https sslpolicy support

Freyert:
I'm trying to add support for specifying an `SSLPolicy` via an annotation: `ingress.gcp.kubernetes.io/ssl-policy`.

Would love a bit of direction if possible, but this seemed like a pretty easy change. There just seem to be a lot of different places in the code to make changes and I'm guaranteed to be missing something.

Also any pointers/tips on automation you all have written so I can build an image and deploy it in my cluster. I'm assuming just basic go build for binaries.

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -22,7 +22,13 @@ configured.
 `ingress.gcp.kubernetes.io/pre-shared-cert`  
 
 Instead of storing certificates and keys in Kubernetes secrets, you can upload them to GCP and
-reference them by name through this annotation.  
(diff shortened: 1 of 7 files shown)


rramkumar1:
@Freyert Thanks for your PR.

We are thinking of adding a new CRD for frontend config like SSL policy (similar to BackendConfig). Once we have this frontend config CRD defined, then we will start accepting implementations for each feature. 

At the moment, we do not want to add annotations for these features since this method is not usable or scalable.

PR #529, opened 2018-10-30
Managed Certs integration: Fix code so that informers are not instantiated if managed certs is not enabled

rramkumar1:
This should fix the current failures in the OSS CI:

Ref: https://gubernator.k8s.io/build/kubernetes-jenkins/logs/ci-ingress-gce-e2e/3251


rramkumar1:
cc @krzykwas

prameshj:
good catch, thanks for fixing. looks good to me.

krzykwas:
Sorry for this bug and many thanks for fixing that, rramkumar1@

PR #530, opened 2018-10-31
Replace existing CHANGELOG.md with one generated via a script

rramkumar1:
/assign @MrHohn 
/assign @freehan 

Once we become better with our issue tracking in terms of labels, the output of this script will be even better.


MrHohn:
Nice!

PR #546, opened 2018-11-14
Ingress Upgrade Testing

agau4779:
Continuation of https://github.com/kubernetes/ingress-gce/pull/392 .

### DIFF
--- a/cmd/e2e-test/basic_upgrade_test.go
+++ b/cmd/e2e-test/basic_upgrade_test.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kr/pretty"
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestBasicUpgrade(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	for _, tc := range []struct {
+		desc string
+		ing  *v1beta1.Ingress
+
+		numForwardingRules int
+		numBackendServices int
+	}{
+		{
+			desc: "http default backend",
+			ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+				DefaultBackend("service-1", port80).
+				Build(),
+			numForwardingRules: 1,
+			numBackendServices: 1,
+		},
+	} {
+		tc := tc // Capture tc as we are running this in parallel.
+		Framework.RunWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			t.Parallel()
+
+			_, _, err := e2e.CreateEchoService(s, "service-1", nil)
+			if err != nil {
+				t.Fatalf("error creating echo service: %v", err)
+			}
+			t.Logf("Echo service created (%s/%s)", s.Namespace, "service-1")
+
+			if _, err := Framework.Clientset.Extensions().Ingresses(s.Namespace).Create(tc.ing); err != nil {
+				t.Fatalf("error creating Ingress spec: %v", err)
+			}
+			t.Logf("Ingress created (%s/%s)", s.Namespace, tc.ing.Name)
+			s.PutStatus(e2e.Unstable)
+
+			runs := 0
+			for {
+				options := &e2e.WaitForIngressOptions{
+					ExpectUnreachable: runs == 0,
+				}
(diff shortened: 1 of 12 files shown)


pkg/e2e/status.go
```diff
@@ -0,0 +1,130 @@
...
+		return true
+	}
+	return false
+}
+
+func (sm *StatusManager) flush() {
+	sm.f.lock.Lock()
+	defer sm.f.lock.Unlock()
+	var err error
+	sm.cm, err = sm.f.Clientset.Core().ConfigMaps("default").Update(sm.cm)
```
bowei: what happens if we get a race here towards updating the configmap?

we should handle it:
```
while true {
  attempt update
  succeeds break
  merge
}
```

pkg/e2e/status.go
```diff
@@ -0,0 +1,130 @@
...
+
+var (
+	// Stable indicates an Ingress is stable (i.e consistently serving 200's)
+	Stable IngressStability = "Stable"
+	// Unstable indicates an Ingress is unstable (i.e serving 404/502's).
+	Unstable IngressStability = "Unstable"
+)
+
+const (
+	configMapName = "status-cm"
```
bowei: should we create unique config maps for each run (per framework instance) to avoid potential clashes if we have a messed up run?

pkg/e2e/status.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	newIndexer := func() cache.Indexers {
+		return cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}
+	}
+	cmInformer := informerv1.NewConfigMapInformer(sm.f.Clientset, "default", 30*time.Second, newIndexer())
+	cmInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(old, cur interface{}) {
+			curCm := cur.(*v1.ConfigMap)
+			if curCm.Data[exitKey] == "yes" {
```
bowei: can we put a timestamp here instead of "yes" (existence indicates yes) so debugging issues is easier

pkg/e2e/status.go
```diff
@@ -0,0 +1,130 @@
...
+		for _ = range time.NewTicker(30 * time.Second).C {
+			sm.flush()
+		}
+	}()
+
+	return nil
+}
+
+func (sm *StatusManager) shutdown() {
+	glog.V(2).Infof("Shutting down status manager.")
```
bowei: log the contents of the configmap before nuking it (for debugging)

pkg/e2e/status.go
bowei: let's log the updated contents of the cm

pkg/e2e/status.go
```diff
@@ -0,0 +1,130 @@
...
+			curCm := cur.(*v1.ConfigMap)
+			if curCm.Data[exitKey] == "yes" {
+				glog.V(2).Infof("ConfigMap was updated with exit switch.")
+				sm.f.shutdown(0)
+			}
+		},
+	})
+
+	go func() {
+		for _ = range time.NewTicker(30 * time.Second).C {
```
bowei: 30 should be global const

pkg/e2e/status.go
agau4779: hm, on the guitar test side we need to know the name of the configmap in order to check the conents. @rramkumar1 - thoughts? What does a messed-up run look like in this case?

rramkumar1: @bowei What do you mean by messed up run?

bowei: ignore my comment, after discussion, I don't think it makes sense.

pkg/e2e/status.go
```diff
@@ -0,0 +1,150 @@
...
+	// Value associated with it is a timestamp string.
+	exitKey = "exit"
+	// masterUpgraded is the key used to indicate to the status manager that
+	// the k8s master has successfully finished upgrading.
+	// Value associated with it is a timestamp string.
+	masterUpgraded = "master-upgraded"
+)
+
+// StatusManager manages the status of sandboxed Ingresses via a ConfigMap.
+// Upon initialization, it creates a ConfigMap object which the guitar
```
bowei: can we have this in a 

1. x happens
2. y happens
format (easier to follow...)

pkg/e2e/status.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (sm *StatusManager) flush() {
+	sm.f.lock.Lock()
+	defer sm.f.lock.Unlock()
+
+	// Loop until we successfully update the config map
+	for {
+		var err error
+		sm.cm, err = sm.f.Clientset.Core().ConfigMaps("default").Update(sm.cm)
```
bowei: could it a race if we do update rather than patch?

cmd/e2e-test/basic_upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+					} else {
+						s.PutStatus(e2e.Unstable)
+						needUpdate = false
+					}
+				}
+
+				options := &e2e.WaitForIngressOptions{
+					ExpectUnreachable: runs == 0,
+				}
+				ing, err := e2e.WaitForIngress(s, tc.ing, options)
```
bowei: I forget -- does this have some delay in the check to avoid excessively spamming when the LB is down?

cmd/e2e-test/basic_upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			t.Logf("Ingress created (%s/%s)", s.Namespace, tc.ing.Name)
+			s.PutStatus(e2e.Unstable)
+
+			runs := 0
+			// needUpdate indicates that the Ingress sync has NOT yet been triggered
+			needUpdate := true
+			// Framework.shutdown() kills this loop.
+			for {
+				if s.MasterUpgraded() && needUpdate {
```
bowei: do this in a follow on--

is it possible to separate out the wait for upgrade logic so it's easier to reuse?

PR #547, opened 2018-11-19
do not leak LB when ingress class is changed

### LINKED ISSUE #481: Ingress-GCE does not GC LB resources when ingress class changed and no other ingress on GCE exists
Repro steps:
1. Make sure there is no ingress exists in the cluster.
2. Create a test ingress with no ingress class annotation
3. Wait for ingress-gce controller to provision LB
4. Add a non-GCE ingress class on the test ingress
5. The LB resources will be leaked. 

Root cause:
- Ingress-gce controller does not act on non-relevant ingresses.  https://github.com/kubernetes/ingress-gce/blob/release-1.3/pkg/controller/controller.go#L127
- Ingress-gce controller only GC resources in the ingress sync loop.
https://github.com/kubernetes/ingress-gce/blob/release-1.3/pkg/controller/controller.go#L287

pondohva:
This PR fixes #481

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -48,6 +48,7 @@ import (
 	ingsync "k8s.io/ingress-gce/pkg/sync"
 	"k8s.io/ingress-gce/pkg/tls"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/kubernetes/pkg/util/slice"
 )
 
 // LoadBalancerController watches the kubernetes api and adds/removes services
@@ -153,6 +154,14 @@ func NewLoadBalancerController(
 		UpdateFunc: func(old, cur interface{}) {
 			curIng := cur.(*extensions.Ingress)
 			if !utils.IsGLBCIngress(curIng) {
+				oldIng := old.(*extensions.Ingress)
+				// If ingress was GLBC Ingress, we need to track ingress class change
+				// and run GC to delete LB resources.
+				if utils.IsGLBCIngress(oldIng) {
+					glog.V(4).Infof("Ingress %v class was changed, enqueuing", utils.IngressKeyFunc(curIng))
+					lbc.ingQueue.Enqueue(cur)
+					return
+				}
 				return
 			}
 			if reflect.DeepEqual(old, cur) {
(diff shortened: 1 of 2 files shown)


pondohva:
I signed it

pkg/controller/controller.go
```diff
@@ -456,6 +462,17 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	}
 	ing = ing.DeepCopy()
 
+	if !utils.IsGLBCIngress(ing) {
+		glog.V(2).Infof("Ingress %q class was changed, triggering GC", key)
```
rramkumar1: Can you explain this? 

We shouldn't delete the Ingress even though the class changed.

pondohva: We need to delete it from gcState.lbNames to allow GC delete load balancer.
I can just fix gcState and not delete it from Store.

rramkumar1: Yea, removing it from the gcState.lbNames sounds better.

We for sure do not want to delete it outright because this will affect other controllers that expect this ingress to exist.

pondohva: Fixed

rramkumar1:
@pondohva Sorry for the delay. I will try to have another review by EOD

rramkumar1:
@pondohva This looks good to me. Can you please add a unit test?

pkg/controller/controller.go
```diff
@@ -456,6 +463,13 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
+		// Remove lb from state for GC
+		gcState.lbNames = slice.RemoveString(gcState.lbNames, key, nil)
+		return lbc.ingSyncer.GC(gcState)
```
rramkumar1: Actually one comment here. I don't like that we have to call ingSyncer.GC() twice in this function.

Can you wrap the GC() call in line 491 into a separate function and then call that function here and in line 491?

pondohva: Check please

pondohva:
@rramkumar1 could you please check test and wrapper function?

pkg/controller/controller.go
```diff
@@ -153,6 +154,12 @@ func NewLoadBalancerController(
 		UpdateFunc: func(old, cur interface{}) {
 			curIng := cur.(*extensions.Ingress)
 			if !utils.IsGLBCIngress(curIng) {
+				oldIng := old.(*extensions.Ingress)
```
rramkumar1: Nit: Add a comment indicating the purpose of why we are enqueueing here.

pkg/controller/controller.go
```diff
@@ -456,6 +463,13 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	}
 	ing = ing.DeepCopy()
 
+	if !utils.IsGLBCIngress(ing) {
```
rramkumar1: Nit: Add a small comment here to please.

pkg/controller/controller.go
```diff
@@ -586,3 +600,7 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings *extensions.IngressList) []ut
 	}
 	return knownPorts
 }
+
+func (lbc *LoadBalancerController) runIngressGC(state *gcState) error {
```
rramkumar1: Now that I look at this, I realize we probably don't need a separate function for this. Just inline the call to GC()

PR #573, opened 2018-12-10
Fixed misspell (annotation name) i README.md

jaceq:
Fixed a misspell in annotation name. 
Now readme is correct and matching .yml files.


jaceq:
CLA signed

Issue #50, opened 2017-10-11
Zonelister logic prevents instance removal from instance group & instance group GC


bowei:
_From @nikhiljindal on July 25, 2017 1:38_

Looking at the code (https://github.com/kubernetes/ingress/blob/a58b80017170eecbe8b2d6573b66192cafe0d32a/controllers/gce/controller/controller.go#L177), it looks like we are not doing anything when nodes are added and deleted. We should be updating the instance group when that happens.

Am I missing something?

_Copied from original issue: kubernetes/ingress-nginx#1012_

bowei:
_From @nikhiljindal on July 25, 2017 1:39_

@nicksardo pointed out that the logic was removed in https://github.com/kubernetes/ingress/commit/c7c2a564a994ec28de9fe5fb0b4a1847be94e0ad#diff-9141c651905f3492033cf255f8e12fd7L176.

@aledbf Was that intentional?

bowei:
_From @nicksardo on July 25, 2017 18:18_

Quick FYI: the instance groups are eventually synced when the resync period occurs for any ingress. This may explain why nobody has noticed this in practice.

bowei:
_From @nicksardo on July 31, 2017 23:47_

This still occurs despite having the fix.  I'm guessing that the handler is called when the node is added; however, it's NotReady state prevents the instance group from being updated. I don't know why the handler isn't called again when the node becomes Ready.

bowei:
_From @aledbf on August 1, 2017 0:0_

@nicksardo maybe because the handler does not have an `UpdateFunc`?

bowei:
_From @nicksardo on August 1, 2017 0:32_

Ah, that would certainly do it. According to the comment, we don't want to have an `UpdateFunc` due to the frequency.  @nikhiljindal and I need to look more closely at the instance sync logic.

bowei:
_From @nikhiljindal on August 1, 2017 17:8_

If there was a way to find out the update, we could add an UpdateFunc that adds the node to queue only if there has been a "relevant change" in node. Node's heartbeat being updated, for example, is not a "relevant change".

bowei:
_From @nicksardo on August 1, 2017 17:14_

Right, not a quick change though.

bowei:
_From @nicksardo on August 3, 2017 0:54_

Looks like removing nodes from instances groups has never worked. 
```
[...]

PR #576, opened 2018-12-11
do not use condition predicate while getting zone for node

pondohva:
This is small part of fix for #50 (fixes this log messages https://github.com/kubernetes/ingress-gce/issues/50#issuecomment-335960715)

GetNodeConditionPredicate filtered nodes with conditions different from "Ready", and we're getting this messages when node is deleting from instance group.


freehan:
It would be great if you could add a unit test for that function.

freehan:
/lgtm

I will add a unit test later.

pondohva:
@freehan thank you. Sorry, I was on vacation.

PR #580, opened 2018-12-13
remove loop from flush(). start/stop informers depending on k8s master

agau4779:


cmd/e2e-test/upgrade_test.go
```diff
@@ -88,12 +88,6 @@ func TestUpgrade(t *testing.T) {
 					}
 				}
 
-				// While k8s master is upgrading, it will return a connection refused
```
rramkumar1: We still want this.

pkg/e2e/sandbox.go
```diff
@@ -87,9 +87,3 @@ func (s *Sandbox) PutStatus(status IngressStability) {
 func (s *Sandbox) MasterUpgraded() bool {
 	return s.f.statusManager.masterUpgraded()
 }
-
```
rramkumar1: We still want this.

pkg/e2e/status.go
```diff
@@ -217,6 +174,20 @@ func (sm *StatusManager) flush() {
 		updatedCm.Data = make(map[string]string)
 	}
 
+	// If master is in the process of upgrading, we exit early and turn off the
```
rramkumar1: How come this was moved back down? Where it was originally was fine.

pkg/e2e/status.go
```diff
@@ -152,66 +161,77 @@ func (sm *StatusManager) putStatus(key string, status IngressStability) {
 	sm.cm.Data[key] = string(status)
 }
 
+func (sm *StatusManager) setMasterUpgrading(ts string) {
```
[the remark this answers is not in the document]
rramkumar1: Disregard comment that we don't need this. We do. 

Can we call this updateLocalConfigMap() though? And when you use it, just update the whole config map rather than one key.

PR #585, opened 2018-12-17
Revert "Remove unused named ports from instance group's"

rramkumar1:
This reverts commit fbe88aa88b961ab683af8486da6e040cbd75584c.

/assign @bowei


qi-beep:
Why was this reverted? It was a legitimate and issue and still exists. After deploying/redeploying services on a cluster, the named-port limit is reached, since they are never properly cleaned up, and the cluster becomes unusable. The only way to fix this is to redeploy or manually use gcloud commands to find and remove unused named-ports.

PR #590, opened 2018-12-20
Replace all uses of Snapshotter with CloudLister

agau4779:
Follow-up to https://github.com/kubernetes/ingress-gce/pull/514 .

Replace the use of /pkg/storage/pools.go (Snapshotter) with a CloudLister interface, to be called as-needed during Garbage Collection. This should reduce the number of calls we make to the GCE API, as well as make GC more reliable.

We will use this for L7 loadbalancers, InstanceGroups, and Backends.

Verified that this properly garbage collects IGs, Backends, and Loadbalancers (TargetProxies, Forwarding Rules, URL Maps) via manual testing


pkg/cloudlist/lister.go
```diff
@@ -0,0 +1,78 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
rramkumar1: Nit: 2019 :)

pkg/controller/controller.go
```diff
@@ -417,12 +416,12 @@ func (lbc *LoadBalancerController) PostProcess(state interface{}) error {
...
 	ing := syncState.ing
-	k, err := utils.KeyFunc(ing)
+	lb, err := lbc.toRuntimeInfo(ing, syncState.urlMap)
 	if err != nil {
-		return fmt.Errorf("cannot get key for Ingress %s/%s: %v", ing.Namespace, ing.Name, err)
+		return err
 	}
 
-	l7, err := lbc.l7Pool.Get(k)
+	l7, err := lbc.l7Pool.Ensure(lb)
```
rramkumar1: Instead of calling Ensure() again, what I would do is add the LB VIP to the sync state in SyncLoadBalancer (this would require adding the VIP as a field in the syncState struct).

Then in updateIngressStatus() you simply pass the VIP rather than the whole l7 object

pkg/instances/instances.go
```diff
@@ -170,8 +176,8 @@ func (i *Instances) DeleteInstanceGroup(name string) error {
 	return fmt.Errorf("%v", errs)
 }
 
-// list lists all instances in all zones.
-func (i *Instances) list(name string) (sets.String, error) {
+// listInstances lists all instances in all zones.
```
rramkumar1: I think you can just keep this as "list" like before unless there was an important reason for it.

pkg/backends/syncer.go
```diff
@@ -140,7 +141,12 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
 		name := sp.BackendName(s.namer)
 		knownPorts.Insert(name)
 	}
-	backendNames := s.backendPool.GetLocalSnapshot()
+
+	backendNames, err := s.backendPool.GetManagedBackends()
+	if err != nil {
+		return fmt.Errorf("error getting the names of managed backends: %v", err)
```
rramkumar1: Nit: controller-managed

pkg/loadbalancers/l7s.go
```diff
@@ -48,83 +46,134 @@ func (l *L7s) Namer() *utils.Namer {
...
-	l.snapshotter.Delete(name)
 	return nil
 }
 
+// List returns a list of loadbalancers with the Name field populated.
+// The loadbalancer resource doesn't actually exist in GCP - so we list all the
+// component resources (forwarding rules, URL Maps, TargetProxies) and parse
+// the loadbalancer name from these resources' names, and use a set in order
+// to de-dupe the loadbalancer names.
+func (l *L7s) List() ([]interface{}, error) {
```
rramkumar1: So you shouldn't have to list all resources. Listing UrlMap's should be sufficient since you can derive everything else. 

Once you list all UrlMap's, you can implement a function to scrub out the LB name and use that as the filter in the key func in line 50.

rramkumar1: Example implementation of a function to scrub out the LB name from the UrlMap name:



func (n *Namer) ScrubUrlMapPrefix(name string) string {
    prefix := fmt.Sprintf("%s-%s-", n.prefix, urlMapPrefix)
    parts := strings.Split(name, prefix)
    if len(parts) < 2 {
        return ""
    }
   return parts[1]
}

pkg/controller/controller.go
agau4779: Resolved offline. We need the whole L7 object in `updateIngressStatus` because we use it to crawl over all the L7 resources to determine the expected annotations.

pkg/cloudlist/lister.go
```diff
@@ -0,0 +1,78 @@
...
+	List() ([]string, error)
+}
+
+type keyFunc func(interface{}) (string, error)
+
+type objectLister interface {
+	List() ([]interface{}, error)
+}
+
+// GCECLister lists the names of a single GCE object.
```
bowei: Is this level of indirection necessary? 

Why not just implement the adapter directly as an adapter class and then use an interface:

type KeyLister interface {
  List() ([]string, error)
}

pkg/cloudlist/lister.go
```diff
@@ -0,0 +1,78 @@
...
+// List implements Lister.
+func (c *GCELister) List() ([]string, error) {
+	glog.V(4).Infof("GCELister %q is pulling objects", c.name)
+
+	items, err := c.objectLister.List()
+	if err != nil {
+		return []string{}, fmt.Errorf("Failed to list %q: %v", c.name, err)
+	}
+
+	objectNames := make([]string, 0)
```
bowei: gostyle `var objectNames []string`

pkg/cloudlist/lister.go
```diff
@@ -0,0 +1,78 @@
...
+	keyGetter keyFunc
+}
+
+// List implements Lister.
+func (c *GCELister) List() ([]string, error) {
+	glog.V(4).Infof("GCELister %q is pulling objects", c.name)
+
+	items, err := c.objectLister.List()
+	if err != nil {
+		return []string{}, fmt.Errorf("Failed to list %q: %v", c.name, err)
```
bowei: is it worth wrapping errors?
if you are going to wrap errors, use https://github.com/pkg/errors/blob/master/errors.go#L181

pkg/cloudlist/lister.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// NewGCELister lists the names of a specific object from GCE.
+func NewGCELister(name string, k keyFunc, objectLister objectLister) Lister {
+	gceLister := &GCELister{
+		name:         name,
+		objectLister: objectLister,
+		keyGetter:    k,
+	}
+	glog.V(4).Infof("Starting GCELister %q", gceLister.name)
```
bowei: this does not actually "start" anything?

pkg/cloudlist/lister.go
agau4779: Deleted cloudlister and opted to use `List() ([]string, error)` on Instances, L7s, and Backeends- let me know if you still think we should explicitly use the adapter class. I opted not to because the resulting file would be very short (<10 lines) and the method signature is simple.

pkg/backends/backends.go
```diff
@@ -174,27 +147,38 @@ func (b *Backends) Health(name string) string {
...
 
-// GetLocalSnapshot implements Pool.
-func (b *Backends) GetLocalSnapshot() []string {
-	pool := b.snapshotter.Snapshot()
-	var keys []string
-	for name := range pool {
-		keys = append(keys, name)
+// GetManagedBackends implements Pool.
+func (b *Backends) GetManagedBackends() ([]string, error) {
+	backends, err := b.List()
```
rramkumar1: You can delete this method. I don't think you call it anywhere anymore.

pkg/controller/controller.go
```diff
@@ -473,14 +465,14 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
 	urlMap, errs := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPortID)
-	syncState := &syncState{urlMap, ing}
 	if errs != nil {
 		msg := fmt.Errorf("error while evaluating the ingress spec: %v", utils.JoinErrs(errs))
 		lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, "Translate", msg.Error())
 		return msg
 	}
 
 	// Sync GCP resources.
+	syncState := &syncState{urlMap, ing, &loadbalancers.L7{}}
```
rramkumar1: I would prefer we don't initialize the L7 pointer to something empty here.

Just do &syncState{urlMap, ing}

pkg/instances/instances.go
```diff
@@ -201,10 +194,40 @@ func (i *Instances) Get(name, zone string) (*compute.InstanceGroup, error) {
 	if err != nil {
 		return nil, err
 	}
-	i.snapshotter.Add(name, struct{}{})
+
 	return ig, nil
 }
 
+// List lists the names of all InstanceGroups belonging to this cluster
```
rramkumar1: Nit: Add period at end of sentence.

pkg/loadbalancers/l7.go
```diff
@@ -193,34 +193,38 @@ func (l *L7) GetIP() string {
 // forwarding rule -> target proxy -> url map
 // This leaves backends and health checks, which are shared across loadbalancers.
 func (l *L7) Cleanup() error {
```
rramkumar1: Can you make this a free function and just pass in the name, cloud and namer? That way, you don't have to instantiate a half-baked L7 just to call Cleanup()

pkg/loadbalancers/l7s.go
```diff
@@ -48,83 +44,86 @@ func (l *L7s) Namer() *utils.Namer {
...
-			namer:       l.namer,
-			mcrt:        l.mcrt,
-			recorder:    l.recorderProducer.Recorder(ri.Ingress.Namespace),
-		}
-	} else {
-		if !reflect.DeepEqual(lb.runtimeInfo, ri) {
-			glog.V(3).Infof("LB %v runtime info changed, old %+v new %+v", lb.Name, lb.runtimeInfo, ri)
-			lb.runtimeInfo = ri
-		}
+	if ri.Ingress != nil {
```
rramkumar1: Why would ri.Ingress ever be nil?

pkg/loadbalancers/l7s.go
```diff
@@ -48,83 +44,86 @@ func (l *L7s) Namer() *utils.Namer {
...
+
+	urlMaps, err := l.cloud.ListUrlMaps()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, um := range urlMaps {
+		if l.namer.NameBelongsToCluster(um.Name) {
+			nameParts := l.namer.ParseName(um.Name)
+			names = append(names, nameParts.LbName)
```
rramkumar1: I would suggest splitting this logic into a function in the namer package so we can easily unit test it.

pkg/utils/namer.go
```diff
@@ -206,21 +206,29 @@ func (n *Namer) decorateName(name string) string {
...
 	if len(c) >= 2 {
 		resource = c[1]
 	}
 	if resource == sslCertPrefix {
-		// For ssl certs, the cluster uid is followed by a hypen and the cert hash, so one more string split needed.
+		// For ssl certs, the cluster uid is followed by a hyphen and the cert hash, so one more string split needed.
 		uid = strings.Split(uid, "-")[0]
 	}
+
+	if resource == targetHTTPProxyPrefix || resource == targetHTTPSProxyPrefix ||
```
rramkumar1: Can remove everything but urlMapPrefix

pkg/loadbalancers/l7.go
agau4779: Resolved offline - doing this would mean propagating the change to loadbalancers/certificates.go. This makes more sense as a refactor/follow-up PR of its own.

rramkumar1:
Also can you squash commits?

pkg/controller/controller.go
agau4779: Go compiler returns an error when given the above - says there aren't enough arguments to instantiate. Opted for `&syncState{urlMap, ing, nil}`

pkg/loadbalancers/l7s.go
agau4779: We already have tests for `NameBelongsToCluster` and `ParseName` in Namer. Opted to add a test in L7s for filtering URLMaps.

pkg/utils/namer.go
```diff
@@ -117,7 +117,7 @@ func NewNamerWithPrefix(prefix, clusterName, firewallName string) *Namer {
 // resource name constructed by the namer. The format of such a name
```
rramkumar1: Can you add a comment to the LbName field that it could potentially be empty is the resource is a backend service.

pkg/utils/namer.go
```diff
@@ -206,21 +206,27 @@ func (n *Namer) decorateName(name string) string {
 // UrlMap.
```
rramkumar1: Remove target proxy, and forwarding rule from the comment since we aren't actually processing those here

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -1216,3 +1206,44 @@ func TestCreateHTTPSLoadBalancerAllCerts(t *testing.T) {
...
+		TLS:       []*TLSCerts{{Key: "key", Cert: "cert"}},
+		UrlMap:    gceUrlMap,
+		Ingress:   newIngress(),
+	}
+	f := NewFakeLoadBalancers(lbInfo.Name, namer)
+	names := []string{
+		"invalid-url-map-name1",
+		"invalid-url-map-name2",
+		"wrongprefix-um-test--uid1",
+		"k8s-um-old-l7--uid1", // Expect List() to catch old URL maps
```
rramkumar1: Can you add a case here where the name gets truncated? If you can find multiple cases when that happens, that would be awesome.

pkg/utils/namer.go
```diff
@@ -202,25 +203,30 @@ func (n *Namer) decorateName(name string) string {
...
+// Backend, InstanceGroup, UrlMap.
 func (n *Namer) ParseName(name string) *NameComponents {
 	l := strings.Split(name, clusterNameDelimiter)
-	var uid, resource string
+	var uid, resource, lbName string
 	if len(l) >= 2 {
 		uid = l[len(l)-1]
 	}
-	c := strings.Split(name, "-")
+	c := strings.Split(l[0], "-")
```
rramkumar1: Why was there a change here?

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -1216,3 +1206,53 @@ func TestCreateHTTPSLoadBalancerAllCerts(t *testing.T) {
...
+	lbName := "test"
+	lbInfo := &L7RuntimeInfo{
+		Name:      namer.LoadBalancer(lbName),
+		AllowHTTP: true,
+		TLS:       []*TLSCerts{{Key: "key", Cert: "cert"}},
+		UrlMap:    gceUrlMap,
+		Ingress:   newIngress(),
+	}
+
+	longname := "this-is-an-exceedingly-long-loadbalancer-name-that-exceeds-the-62-char-limit"
```
rramkumar1: Now that I think about it, we don't need test cases for truncation. It turns out when truncation occurs, it will truncate from the end of the name so the cases you have here are not valid. Given that this sort of truncation is a known bug, we can just remove these test cases.

pkg/utils/namer.go
agau4779: Added some clarifying comments in the code.

rramkumar1: Got it. Now that I look at the previous code, I'm surprised it even worked lol.

PR #611, opened 2019-01-15
[BackendConfig] CDN default cache key policy should be true instead of false.

### LINKED ISSUE #599: BackendConfig `cdn: enable: true` uses scary defaults
I created a backend config like this:

```
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: mediaproxy-backend-config
spec:
  timeoutSec: 120
  cdn:
    enabled: true
```

Expecting it to be equivalent to the default CDN policy, which is more explicitly written like this:

```
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: mediaproxy-backend-config
spec:
  timeoutSec: 120
  cdn:
    enabled: true
    cachePolicy:
      includeHost: true
      includeProtocol: true
      includeQueryString: true
```

Unfortunately the default if `cachePolicy` is omitted is actually equivalent to:

```
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: mediaproxy-backend-config
spec:
  timeoutSec: 120
  cdn:
    enabled: true
    cachePolicy:
      includeHost: false
      includeProtocol: false
      includeQueryString: false
```

This is a very unsafe default, in our case query parameters are essential to prevent data leakage between users, and I would like to ask that the default be the same as the Google Cloud CDN default.

For now I can work around the problem by explicitly setting the policy, but it was a nasty surprise.

rramkumar1:
Fixes #599

Currently verifying whether this works and other existing state transitions still hold.

### DIFF
--- a/pkg/backends/features/cdn.go
+++ b/pkg/backends/features/cdn.go
@@ -20,7 +20,6 @@ import (
 	"reflect"
 
 	"github.com/golang/glog"
-	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/utils"
 )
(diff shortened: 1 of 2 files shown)


rramkumar1:
Verified this works. 

/hold cancel

PR #613, opened 2019-01-16
Add Finalizers for Ingress

agau4779:
The finalizer `ingress.finalizer.cloud.google.com` will be applied to the Ingress upon update if it does not have a finalizer yet. It will be removed only after the resources related to the Ingress have been deleted first. Absence of a finalizer means the Ingress itself can be safely deleted.

If a resource used for this ingress fails to be deleted because it’s ‘in use’ by another resource, then the resource will be skipped. If it cannot be deleted for another reason, the controller will attempt deletion at a later attempt.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -368,6 +368,17 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
 			return err
 		}
 	}
+
+	for _, ing := range gcState.ingresses {
+		if utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey()) {
+			ingClient := lbc.ctx.KubeClient.Extensions().Ingresses(ing.Namespace)
+			if err := utils.RemoveFinalizer(ing, ingClient); err != nil {
+				glog.Errorf("Failed to remove Finalizer from Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
+				return err
+			}
+		}
+	}
+
 	return nil
 }
 
@@ -418,10 +429,7 @@ func (lbc *LoadBalancerController) PostProcess(state interface{}) error {
 	}
 
 	// Update the ingress status.
-	if err := lbc.updateIngressStatus(syncState.l7, syncState.ing); err != nil {
-		return fmt.Errorf("update ingress status error: %v", err)
-	}
-	return nil
+	return lbc.updateIngressStatus(syncState.l7, syncState.ing)
 }
 
 // sync manages Ingress create/updates/deletes events from queue.
@@ -440,12 +448,12 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	// gceSvcPorts contains the ServicePorts used by only single-cluster ingress.
 	gceSvcPorts := lbc.ToSvcPorts(gceIngresses)
 	lbNames := lbc.ctx.Ingresses().ListKeys()
-	gcState := &gcState{lbNames, gceSvcPorts}
 
 	ing, ingExists, err := lbc.ctx.Ingresses().GetByKey(key)
 	if err != nil {
 		return fmt.Errorf("error getting Ingress for key %s: %v", key, err)
 	}
+	gcState := &gcState{lbc.ctx.Ingresses().List(), lbNames, gceSvcPorts}
 	if !ingExists {
 		glog.V(2).Infof("Ingress %q no longer exists, triggering GC", key)
 		// GC will find GCE resources that were used for this ingress and delete them.
@@ -454,13 +462,22 @@ func (lbc *LoadBalancerController) sync(key string) error {
 
 	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
 	ing = ing.DeepCopy()
+	ingClient := lbc.ctx.KubeClient.Extensions().Ingresses(ing.Namespace)
+	if err := utils.AddFinalizer(ing, ingClient); err != nil {
+		glog.Errorf("Failed to add Finalizer to Ingress %q: %v", key, err)
+		return err
+	}
 
 	// Check if ingress class was changed to non-GLBC to remove ingress LB from state and trigger GC
 	if !utils.IsGLBCIngress(ing) {
 		glog.V(2).Infof("Ingress %q class was changed, triggering GC", key)
 		// Remove lb from state for GC
 		gcState.lbNames = slice.RemoveString(gcState.lbNames, key, nil)
-		return lbc.ingSyncer.GC(gcState)
+		if gcErr := lbc.ingSyncer.GC(gcState); gcErr != nil {
+			return gcErr
+		}
+
+		return nil
 	}
 
 	// Bootstrap state for GCP sync.
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/context"
 	"k8s.io/ingress-gce/pkg/controller/translator"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/loadbalancers"
 	ingsync "k8s.io/ingress-gce/pkg/sync"
 	"k8s.io/ingress-gce/pkg/tls"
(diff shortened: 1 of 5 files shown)


pkg/utils/finalizer.go
```diff
@@ -36,3 +41,40 @@ func NeedToAddFinalizer(m meta_v1.ObjectMeta, key string) bool {
...
 }
+
+// IngressFinalizerKey generates the finalizer string for Ingress
+func IngressFinalizerKey() string {
+	return fmt.Sprintf("%s.%s", "ingress", FinalizerKeySuffix)
+}
+
+// AddIngressFinalizer tries to add a finalizer to an Ingress. If a finalizer
+// already exists, it does nothing.
+func AddIngressFinalizer(ing *extensions.Ingress, ingClient client.IngressInterface) error {
```
rramkumar1: Can just call this AddFinalizer. Function signature implies its for Ingress. Same for Remove

pkg/utils/finalizer.go
```diff
@@ -36,3 +41,40 @@ func NeedToAddFinalizer(m meta_v1.ObjectMeta, key string) bool {
 func HasFinalizer(m meta_v1.ObjectMeta, key string) bool {
 	return slice.ContainsString(m.Finalizers, key, nil)
 }
+
+// IngressFinalizerKey generates the finalizer string for Ingress
+func IngressFinalizerKey() string {
```
rramkumar1: Can just call this FinalizerKey.

pkg/controller/controller_test.go
```diff
@@ -181,11 +181,17 @@ func TestIngressCreateDelete(t *testing.T) {
 	if len(updatedIng.Status.LoadBalancer.Ingress) != 1 || updatedIng.Status.LoadBalancer.Ingress[0].IP == "" {
 		t.Errorf("Get(%q) = status %+v, want non-empty", updatedIng.Name, updatedIng.Status.LoadBalancer.Ingress)
 	}
-
-	deleteIngress(lbc, ing)
-	if err := lbc.sync(ingStoreKey); err != nil {
-		t.Fatalf("lbc.sync(%v) = err %v", ingStoreKey, err)
-	}
+	//
```
rramkumar1: How come this is commented out?

pkg/controller/controller.go
```diff
@@ -454,13 +464,21 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
 	if !utils.IsGLBCIngress(ing) {
 		glog.V(2).Infof("Ingress %q class was changed, triggering GC", key)
 		// Remove lb from state for GC
 		gcState.lbNames = slice.RemoveString(gcState.lbNames, key, nil)
-		return lbc.ingSyncer.GC(gcState)
+		if gcErr := lbc.ingSyncer.GC(gcState); gcErr != nil {
+			return gcErr
+		}
+
+		if utils.IsDeletionCandidate(ing.ObjectMeta, utils.IngressFinalizerKey()) {
```
rramkumar1: You should call this in GCBackends because if we GC the backends for an Ingress that means it is fully cleaned up.

pkg/controller/controller.go
```diff
@@ -421,6 +421,16 @@ func (lbc *LoadBalancerController) PostProcess(state interface{}) error {
...
 		return fmt.Errorf("update ingress status error: %v", err)
 	}
+
+	ingClient := lbc.ctx.KubeClient.Extensions().Ingresses(syncState.ing.Namespace)
+	currIng, err := ingClient.Get(syncState.ing.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if err := utils.AddIngressFinalizer(currIng, ingClient); err != nil {
```
rramkumar1: How come you add the finalizer here? We should add it before any syncing has taken place

pkg/controller/controller.go
agau4779: Hmm, ok but that means adding a field for Ingress onto gcState. Done.

pkg/controller/controller_test.go
agau4779: Oops - re-added.

bowei:
Can we add copious glog logging in each of the places where the finalizer is being added/removed?

We can put those actions are V(3), errors at Errorf level...

agau4779:
@bowei added errorf logs for when the finalizer fails to be added/removed. Successful adds/removes are glog.V(3) - logged.

pkg/controller/types.go
```diff
@@ -25,6 +25,7 @@ import (
 
 // gcState is used by the controller to maintain state for garbage collection routines.
 type gcState struct {
+	ing      *extensions.Ingress
```
rramkumar1: You will need to store a list of ingresses here.

pkg/controller/controller.go
```diff
@@ -368,6 +368,15 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
 			return err
 		}
 	}
+
+	if gcState.ing != nil && utils.IsDeletionCandidate(gcState.ing.ObjectMeta, utils.FinalizerKey()) {
```
rramkumar1: You need to do this for every ingress in existence, not just one. 

Therefore, you need to store the list of all ingresses in the gcState. See my comment below.

rramkumar1:
General comment: We need to be very careful on the upgrade story here. Theoretically a user could upgrade to a version where we support finalizers, the controller adds a finalizer and then the user downgrades. 

Once they downgrade, the controller software no longer has the ability to remove finalizers, so if the user tries to delete it, the deletion will never succeed. We may need to stagger the rollout of this.

bowei:
Hmm -- that is a good point regarding rollout. It seems like the most we can do now is to enable the code to remove the finalizer. We can flag gate the rest of the code so it is tested but not active in production.

pkg/flags/flags.go
```diff
@@ -131,13 +131,16 @@ type Features struct {
 	NEGExposed bool
 	// ManagedCertificates enables using ManagedCertificate CRD
 	ManagedCertificates bool
+	// Finalizer enables adding a finalizer on Ingress
```
rramkumar1: I think we will need two flags. One to enable the removal logic and one to enable to adding of the finalizers.

bowei:
How is this tested? (esp. with multiple Ingress objects to make sure we don't accidentally GC other Ingresses)...

pkg/flags/flags.go
agau4779: Added a flag for Ingress removal.

agau4779:
@bowei added tests for creating and deleting multiple ingresses, with and without finalizers enabled

pkg/flags/flags.go
```diff
@@ -131,13 +131,19 @@ type Features struct {
...
+	// FinalizerRemove enables removing a finalizer on Ingress.
+	FinalizerRemove bool
 }
 
 var DefaultFeatures = &Features{
 	Http2:               true,
 	NEG:                 true,
 	NEGExposed:          true,
 	ManagedCertificates: false,
+	FinalizerAdd:        false,
```
rramkumar1: Don't you need corresponding flags for these?

pkg/controller/controller_test.go
```diff
@@ -182,10 +193,134 @@ func TestIngressCreateDelete(t *testing.T) {
...
+	// Check Ingress has been deleted
+	updatedIng, _ = lbc.ctx.KubeClient.Extensions().Ingresses(ing.Namespace).Get(ing.Name, meta_v1.GetOptions{})
+	if updatedIng != nil {
+		t.Fatalf("Ingress was not deleted, got: %+v", updatedIng)
+	}
+}
+
+// TestIngressCreateDeleteMultiple asserts that `sync` will properly garbage collect
+// Ingresses that need to be deleted, and keep the ones that don't.
+func TestIngressCreateDeleteMultiple(t *testing.T) {
```
rramkumar1: I would call this TestCreateDeleteFinalizer

pkg/controller/controller_test.go
```diff
@@ -156,6 +165,8 @@ func TestIngressSyncError(t *testing.T) {
 // TestIngressCreateDelete asserts that `sync` will not return an error for a good ingress config
 // and will not return an error when the ingress is deleted.
 func TestIngressCreateDelete(t *testing.T) {
+	flags.F.Features.FinalizerAdd = true
```
rramkumar1: I think whatever you have in this test can be combined into TestCreateDeleteFinalizer.

rramkumar1:
Also as per the new naming policy, we should rename the finalizer to "networking.gke.io/ingress-finalizer"

pkg/controller/controller_test.go
agau4779: Merged TestIngressCreateDelete and TestCreateDeleteFinalizer

PR #620, opened 2019-01-30
Simplify upgrade_test - only loop during k8s master changes.

agau4779:


cmd/e2e-test/upgrade_test.go
```diff
@@ -84,74 +99,64 @@ func TestUpgrade(t *testing.T) {
...
-				s.PutStatus(e2e.Stable)
+			}
 
-				if runs == 0 {
-					t.Logf("GCLB resources created (%s/%s)", s.Namespace, tc.ing.Name)
-				} else {
-					t.Logf("GCLB is stable (%s/%s)", s.Namespace, tc.ing.Name)
-				}
+			// Verify the Ingress has stabilized after the master upgrade and we
+			// trigger an Ingress update
```
rramkumar1: This logic can go into a func as well since its used twice.

pkg/e2e/helpers.go
```diff
@@ -51,8 +52,16 @@ func WaitForIngress(s *Sandbox, ing *v1beta1.Ingress, options *WaitForIngressOpt
 		var err error
 		ing, err = s.f.Clientset.Extensions().Ingresses(s.Namespace).Get(ing.Name, metav1.GetOptions{})
 		if err != nil {
+			if strings.Contains(err.Error(), "connection refused") {
```
rramkumar1: IMO, we should leave this connection refused check outside of the call to WaitForIngress as it was before. It should be the callers responsibility to check for certain errors.

agau4779: I think status.go#flush() handles this case - see these lines:
https://github.com/kubernetes/ingress-gce/blob/07bf1f8fb519cddf5b5a9002822270cb0d13741c/pkg/e2e/status.go#L210-L215

It looks like I accidentally omitted a return statement there prior to this PR - this should suppress the extra error messages.

Issue #32, opened 2017-10-11
[GLBC] Changing front-end configuration does not remove unnecessary target proxies/ssl-certs


bowei:
_From @tonglil on March 20, 2017 17:59_

Porting this issue from contrib over: https://github.com/kubernetes/contrib/issues/1517.

There is no enforcement of the annotation `kubernetes.io/ingress.allow-http: "false"` when it is set to false, after previously being set to true or unset.

See during the edge hop:
- Checks annotation: https://github.com/kubernetes/ingress/blob/37bdb3952e090e50f44207ad0e54b1f2a4ef1055/controllers/gce/loadbalancers/loadbalancers.go#L598-L602
- Creates: https://github.com/kubernetes/ingress/blob/37bdb3952e090e50f44207ad0e54b1f2a4ef1055/controllers/gce/loadbalancers/loadbalancers.go#L619-L627

No deletion/cleanup enforcement happens if it is set to false.

_Copied from original issue: kubernetes/ingress-nginx#468_

bowei:
_From @nicksardo on April 21, 2017 19:7_

No deletion/cleanup happens when removing the TLS section or pre-shared cert annotation.  Will rename this to be more general.

bowei:
_From @nicksardo on April 24, 2017 18:41_

Furthermore, if you disable TLS on an active ingress, then delete the ingress resource, the controller will fail to destroy resources. It doesn't try deleting the https forwarding rule or https target proxy, which means the URL map is still used.

bowei:
_From @montanaflynn on August 7, 2017 23:3_

Any updates on this? We have a production ingress and are looking for a workaround.

bowei:
_From @nicksardo on August 7, 2017 23:19_

@montanaflynn There's no fix yet.

You should be able to delete the HTTP forwarding rule and target proxy manually in the GCP Console.  Highly recommend duplicating the issue in a test cluster and trying it out first.

bluecmd:
It is still relevant, I just hit this today. /remove-lifecycle stale

MichielDeMey:
Same issue here, changed to HTTPS only and the HTTP frontend was not deleted.
I deleted the HTTP frontend manually from the LB and that did not seem to have any side-effects on the traffic.

metral:
Hitting this on v1.1.1

/remove-lifecycle stale

raelga:
[...]

PR #629, opened 2019-02-06
clean up HTTP(S) resources when changing protocols

agau4779:
Addresses #32

### DIFF
--- a/pkg/loadbalancers/l7.go
+++ b/pkg/loadbalancers/l7.go
@@ -138,6 +138,12 @@ func (l *L7) edgeHop() error {
 		if err := l.edgeHopHttp(); err != nil {
 			return err
 		}
+	} else {
+		// if AllowHTTP is disabled, clean up the HTTP resources that would have
+		// been derived from this L7.
+		if err := l.cleanupResourcesWithProtocol(utils.HTTPProtocol); err != nil {
+			return err
+		}
 	}
 	// Defer promoting an ephemeral to a static IP until it's really needed.
 	sslConfigured := l.runtimeInfo.TLS != nil || l.runtimeInfo.TLSName != "" || l.runtimeInfo.ManagedCertificates != ""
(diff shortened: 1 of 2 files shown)


pkg/loadbalancers/l7.go
```diff
@@ -146,13 +152,25 @@ func (l *L7) edgeHop() error {
...
 	if sslConfigured {
 		glog.V(3).Infof("validating https for %v", l.Name)
 		if err := l.edgeHopHttps(); err != nil {
 			return err
 		}
 	}
+
+	// If SSL is disabled, clean up the HTTPS resources that would have been
+	// derived from this L7.
+	if !sslConfigured {
```
rramkumar1: I don't think you need this if condition. If execution get's here, we already know ssl is not configured based on the check in line 160.

pkg/loadbalancers/l7.go
```diff
@@ -146,13 +152,25 @@ func (l *L7) edgeHop() error {
 		if err := l.checkStaticIP(); err != nil {
 			return err
 		}
+	} else {
```
rramkumar1: What's the reason for deleting the IP here?

pkg/loadbalancers/l7.go
```diff
@@ -213,18 +242,10 @@ func (l *L7) Cleanup() error {
...
-
-	tpsName := l.namer.TargetProxy(l.Name, utils.HTTPSProtocol)
-	glog.V(2).Infof("Deleting target https proxy %v", tpsName)
-	if err := utils.IgnoreHTTPNotFound(l.cloud.DeleteTargetHttpsProxy(tpsName)); err != nil {
-		return err
-	}
+	return nil
+}
 
+func (l *L7) deleteSslCerts() error {
```
rramkumar1: can we call this deleteIngressManagedSslCerts()

pkg/loadbalancers/l7.go
agau4779: If we don't need an IP, was hoping to delete it to avoid abandoning it. Do we not need it?

agau4779: Removed this line, added todo - in the future we need to consider whether or not to demote the IP to ephemeral

PR #639, opened 2019-02-13
Import limitations and add that pods must exist

FrankPetrilli:
Import limitations from examples/health-checks/README.md and add new limitation mentioned in #241.

This caveat that all pods must exist is unclear to GCP customers and should be documented as a known limitation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The table below describes what version of Ingress-GCE is running on GKE. Note th
 A few points to note:
 * The pod's `containerPort` field must be defined
 * The service's `targetPort` field must point to the pod port's `containerPort` value or `name`. Note that the `targetPort` defaults to the `port` value if not defined
-* The pods must exist at the time of ingress creation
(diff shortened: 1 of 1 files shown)


FrankPetrilli:
Went through corporate CLA process - recheck requested. :)

rramkumar1:
@FrankPetrilli We already have a public documentation on GCP [1] for this kind of stuff. Do you mind if I add it there?

[1] https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#limitations

FrankPetrilli:
I'll be editing these docs - thanks!

iftachsc:
its not working none the less. i have readiness probe on the pods that show as an healthy endpoint.
it had tcpSocket healtcheck on 80. when i then create ingress (after pod exists) GLB has healtcheck with HTTP / to the nodeport instead of tcp socket. if i try http healtchecks to other paths and ports - also not working. this is really killing us.

PR #641, opened 2019-02-14
Support secret-based and pre-shared certs at the same time

michallowicki:
This should allow for a no-downtime migration from secret-based to pre-shared certs, as the secret-based cert will be still used until the secret is removed.

### DIFF
--- a/pkg/loadbalancers/certificates.go
+++ b/pkg/loadbalancers/certificates.go
@@ -30,23 +30,35 @@ import (
 const SslCertificateMissing = "SslCertificateMissing"
 
 func (l *L7) checkSSLCert() error {
+	// Use both pre-shared and secret-based certs if available,
+	// combining encountered errors.
+	errs := []error{}
+
 	// Handle annotation pre-shared-cert
-	used, preSharedSslCerts, err := l.getPreSharedCertificates()
-	if used {
-		l.sslCerts = preSharedSslCerts
-		return err
+	preSharedSslCerts, err := l.getPreSharedCertificates()
+	if err != nil {
+		errs = append(errs, err)
 	}
+	l.sslCerts = preSharedSslCerts
 
 	// Get updated value of certificate for comparison
 	existingSecretsSslCerts, err := l.getIngressManagedSslCerts()
 	if err != nil {
-		return err
+		errs = append(errs, err)
+		// Do not continue if getIngressManagedSslCerts() failed.
+		return utils.JoinErrs(errs)
 	}
 
 	l.oldSSLCerts = existingSecretsSslCerts
 	secretsSslCerts, err := l.createSslCertificates(existingSecretsSslCerts)
-	l.sslCerts = secretsSslCerts
-	return err
+	if err != nil {
+		errs = append(errs, err)
+	}
+	l.sslCerts = append(l.sslCerts, secretsSslCerts...)
+	if len(errs) > 0 {
+		return utils.JoinErrs(errs)
+	}
+	return nil
 }
 
 // createSslCertificates creates SslCertificates based on kubernetes secrets in Ingress configuration.
@@ -134,17 +146,17 @@ func (l *L7) getSslCertificates(names []string) ([]*compute.SslCertificate, erro
 }
 
 // getPreSharedCertificates fetches SslCertificates specified via pre-shared-cert annotation.
-func (l *L7) getPreSharedCertificates() (bool, []*compute.SslCertificate, error) {
+func (l *L7) getPreSharedCertificates() ([]*compute.SslCertificate, error) {
 	if l.runtimeInfo.TLSName == "" {
-		return false, nil, nil
+		return nil, nil
 	}
 	sslCerts, err := l.getSslCertificates(utils.SplitAnnotation(l.runtimeInfo.TLSName))
 
 	if err != nil {
-		return true, sslCerts, fmt.Errorf("pre-shared-cert errors: %s", err.Error())
+		return sslCerts, fmt.Errorf("pre-shared-cert errors: %s", err.Error())
 	}
 
-	return true, sslCerts, nil
+	return sslCerts, nil
 }
 
 func getMapfromCertList(certs []*compute.SslCertificate) map[string]*compute.SslCertificate {
--- a/pkg/loadbalancers/loadbalancers_test.go
+++ b/pkg/loadbalancers/loadbalancers_test.go
@@ -384,7 +384,7 @@ func TestMaxCertsUpload(t *testing.T) {
 	namer := utils.NewNamer("uid1", "fw1")
 	lbName := namer.LoadBalancer("test")
 
-	for ix := 0; ix < FakeCertLimit; ix++ {
+	for ix := 0; ix < FakeCertQuota; ix++ {
 		str := strconv.Itoa(ix)
 		tlsCerts = append(tlsCerts, createCert("key-"+str, "cert-"+str, "name-"+str))
 		certName := namer.SSLCertName(lbName, GetCertHash("cert-"+str))
@@ -409,7 +409,7 @@ func TestMaxCertsUpload(t *testing.T) {
 	lbInfo.TLS = append(lbInfo.TLS, failCert)
 	_, err := pool.Ensure(lbInfo)
 	if err == nil {
-		t.Fatalf("Creating more than %d certs should have errored out", FakeCertLimit)
+		t.Fatalf("Creating more than %d certs should have errored out", FakeCertQuota)
 	}
 	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
 	// Set cert count less than cert creation limit but more than target proxy limit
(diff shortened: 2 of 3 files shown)


pkg/loadbalancers/certificates.go
```diff
@@ -30,23 +30,39 @@ import (
 const SslCertificateMissing = "SslCertificateMissing"
 
 func (l *L7) checkSSLCert() error {
+	// Use both pre-shared and secret-based certs if available,
+	// combining encountered errors.
+	joinErrors := func(errs []string) error {
+		return fmt.Errorf("checkSSLCert() errors:\n%s", strings.Join(errs, "\n"))
+	}
+	errs := []string{}
```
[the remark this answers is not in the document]
michallowicki: Done.

pkg/loadbalancers/certificates.go
```diff
@@ -30,23 +30,39 @@ import (
 const SslCertificateMissing = "SslCertificateMissing"
 
 func (l *L7) checkSSLCert() error {
+	// Use both pre-shared and secret-based certs if available,
+	// combining encountered errors.
+	joinErrors := func(errs []string) error {
+		return fmt.Errorf("checkSSLCert() errors:\n%s", strings.Join(errs, "\n"))
```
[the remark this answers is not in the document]
michallowicki: Done.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,180 @@ func TestList(t *testing.T) {
...
+
+	// Delete the secret.
+	lbInfo.TLS = []*TLSCerts{}
+	if _, err := pool.Ensure(lbInfo); err != nil {
+		t.Fatalf("pool.Ensure() = err %v", err)
+	}
+	delete(expectCerts, certName1)
+	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
+
+	// Add the secret again.
```
rramkumar1: What are you trying to test by adding the secret again? Seems unnecessary to me but I could be missing something.

pkg/loadbalancers/certificates.go
```diff
@@ -30,23 +30,42 @@ import (
 const SslCertificateMissing = "SslCertificateMissing"
 
 func (l *L7) checkSSLCert() error {
+	// Use both pre-shared and secret-based certs if available,
```
rramkumar1: So, I just remembered that we have a function in pkg/utils/utils.go to join errors. Can you use that instead? 

Also the caller of this function can take care of wrapping the error with "Errors ensuring SSL certs"

pkg/loadbalancers/loadbalancers_test.go
michallowicki: Good point, removed this part of the test.
The idea was to also test reverting the update, but it's already tested by TestPreSharedToSecretBasedCertUpdate.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,180 @@ func TestList(t *testing.T) {
...
+	}
+	expectCerts[certName1] = lbInfo.TLS[0].Cert
+	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
+
+	// Remove the pre-shared cert.
+	lbInfo.TLSName = ""
+	if _, err := pool.Ensure(lbInfo); err != nil {
+		t.Fatalf("pool.Ensure() = err %v", err)
+	}
+	// Pre-shared cert is retained in the fake loadbalancer.
```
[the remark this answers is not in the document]
michallowicki: Removed this part of the test.
Pre-shared cert can be removed from target proxy but retained in the loadbalancer, verifyCertAndProxyLink() verifies that.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,180 @@ func TestList(t *testing.T) {
 		t.Fatalf("pool.List() returned %+v, want %+v", lbNames, expected)
 	}
 }
+
+// TestSecretBasedAndPreSharedCerts creates both pre-shared and
+// secret-based certs that should all be used.
```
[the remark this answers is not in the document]
michallowicki: Done.

pkg/loadbalancers/certificates.go
michallowicki: Done.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,163 @@ func TestList(t *testing.T) {
...
+	// pre-shared cert is not used by the target proxy.
+	verifyCertAndProxyLink(expectCerts, expectCertsProxy, f, t)
+
+	// Fix the typo.
+	lbInfo.TLSName = preSharedCert1.Name
+	if _, err := pool.Ensure(lbInfo); err != nil {
+		t.Fatalf("pool.Ensure() = err %v", err)
+	}
+	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
+}
```
prameshj: Can we add a test for a case where 10 ingress secrets/certs are specified and lb is created. Then as an update, 10 pre-shared certs are specified. From the code, looks like we will prefer the pre-shared certs first correct? So all the 10 certs that were initially used will be replaced by the 10 new pre-shared certs that were specified. Correct me if i misunderstood this. I chose 10 because we use that as a fake limit in some other tests.

It would be good to have a test validating this so we can document the behavior.

michallowicki: Good point, added a test for this case:
10 ingress secrets/certs are specified and lb is created.
Then 5 pre-shared are created (reaching FakeCertLimit of 15).
Trying to use all 15 certs fails, and old setup with 10 secret-base certs is not overwritten. This is treated the same as trying to use 15 certs of the same type - if there are more certs than the TargetProxyCertLimit of 10, the update fails.
After removing 5 secret-based certs, the remaining 5 secret-based and 5 pre-shared certs can be used.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,163 @@ func TestList(t *testing.T) {
...
+	// Prepare pre-shared certs.
+	preSharedCert1, _ := f.CreateSslCertificate(&compute.SslCertificate{
+		Name:        "test-pre-shared-cert",
+		Certificate: "abc",
+		SelfLink:    "existing",
+	})
+	preSharedCert2, _ := f.CreateSslCertificate(&compute.SslCertificate{
+		Name:        "test-pre-shared-cert2",
+		Certificate: "xyz",
+		SelfLink:    "existing",
```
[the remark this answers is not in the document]
michallowicki: Done.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,253 @@ func TestList(t *testing.T) {
...
+		certName1:           lbInfo.TLS[0].Cert,
+		certName2:           lbInfo.TLS[1].Cert,
+	}
+
+	// Verify that both secret-based and pre-shared certs are used.
+	pool.Ensure(lbInfo)
+	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
+}
+
+// TestMaxSecretBasedAndPreSharedCerts uploads 10 secret-based certs, which is the limit for the fake loadbalancer.
```
prameshj: Can you reword this as "which is the limit for the fake target proxy"?

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,253 @@ func TestList(t *testing.T) {
...
+// Then creates 5 pre-shared certs, reaching the limit of 15 certs for the fake loadbalancer.
```
prameshj: reword this as "limit of 15 certs which is the fake certs quota"

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -966,3 +966,253 @@ func TestList(t *testing.T) {
...
+
+	if _, err := pool.Ensure(lbInfo); err != nil {
+		t.Fatalf("pool.Ensure() = err %v", err)
+	}
+	verifyCertAndProxyLink(expectCerts, expectCerts, f, t)
+
+	// Create pre-shared certs up to FakeCertLimit.
+	preSharedCerts := []*compute.SslCertificate{}
+	tlsNames := []string{}
+	for ix := TargetProxyCertLimit; ix < FakeCertLimit; ix++ {
```
prameshj: nit: This isn't your change, but if you wouldn't mind, can you rename "FakeCertLimit" to "FakeCertQuota"? We can do it in a follow up change too, if that's easier.

pkg/loadbalancers/loadbalancers_test.go
prameshj: Thanks for verifying the behavior and adding the new tests!

 Since the update fails due to targetProxy limit exceeding, the certs used by ingress will not be modified, I missed that in my initial comment.

pkg/loadbalancers/loadbalancers_test.go
michallowicki: Done.

pkg/loadbalancers/loadbalancers_test.go
michallowicki: Done.

pkg/loadbalancers/loadbalancers_test.go
michallowicki: Done.

PR #642, opened 2019-02-15
Add backendconfig v1

agau4779:
This PR adds v1 version of the BackendConfig CustomResourceDefinition. It is feature gated behind a flag, `EnableBackendConfigV1`. Note that to enable BackendConfig v1, ingress requires both `EnableBackendConfigV1` and `EnableBackendConfig` to be true.

We also introduce a composite type, `backendconfig.BackendConfig`, to house the shared attributes between v1 and v1beta1.

### DIFF
--- a/cmd/e2e-test/backend_config_test.go
+++ b/cmd/e2e-test/backend_config_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
@@ -28,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/ingress-gce/pkg/annotations"
+	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
 	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
 	"k8s.io/ingress-gce/pkg/e2e"
 	"k8s.io/ingress-gce/pkg/fuzz"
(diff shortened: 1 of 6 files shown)


rramkumar1:
@agau4779 Let's discuss this more offline. My feeling is that we need to be careful with this and somehow follow the docs in https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/

pkg/backendconfig/backendconfig.go
```diff
@@ -33,17 +34,31 @@ var (
 	ErrBackendConfigDoesNotExist = errors.New("no BackendConfig for service port exists.")
 	ErrBackendConfigFailedToGet  = errors.New("client had error getting BackendConfig for service port.")
 	ErrNoBackendConfigForPort    = errors.New("no BackendConfig name found for service port.")
+	CRDVersion                   = []apiextensionsv1beta1.CustomResourceDefinitionVersion{
```
rramkumar1: Can you add a space after the err vars and before this one.

pkg/backendconfig/backendconfig.go
```diff
@@ -33,17 +34,31 @@ var (
...
+		CRDVersion,
 		apisbackendconfig.GroupName,
-		"v1beta1",
 		"BackendConfig",
 		"BackendConfigList",
 		"backendconfig",
 		"backendconfigs",
 	)
+
+	// TODO: when we deprecate v1beta1, use v1.BackendConfig, v1.GetOpenAPIDefinitions
```
rramkumar1: Can you add to this TODO that we need to use v1 client and watchers as well.

pkg/crd/meta.go
```diff
@@ -33,12 +34,18 @@ type CRDMeta struct {
 	fn         common.GetOpenAPIDefinitions
 }
 
+type CRDVersion struct {
```
rramkumar1: I think this can be deleted. It doesn't seem to be used.

rramkumar1:
Can we also add an e2e test for this? You can either put it in the existing backend_config_test.go file or start a new file called backend_config_migration_test.go

pkg/e2e/fixtures.go
```diff
@@ -74,6 +74,23 @@ func CreateEchoService(s *Sandbox, name string, annotations map[string]string) (
 	return pod, service, nil
 }
 
+// UpdateEchoService updates the annotations of the  service serving echoheaders.
+func UpdateEchoService(s *Sandbox, name string, annotations map[string]string) (*v1.Service, error) {
```
rramkumar1: I would just make this a generic Update function so instead of passing in the annotations, pass in a new service

cmd/e2e-test/backend_config_test.go
```diff
@@ -143,3 +145,113 @@ func TestBackendConfigNegatives(t *testing.T) {
...
+				Build(),
+		},
+		{
+			desc: "enable CDN in backend config",
+			backendConfig: fuzz.NewBackendConfigBuilder("", "backendconfig-v1beta1").
+				EnableCDN(true).
+				Build(),
+		},
+		{
+			desc: "enable IAP in backend config",
```
rramkumar1: Currently, enabling IAP is not supported in the e2e test since we don't have a Oauth2 token to use. 

I would just remove this test case for now. The other two should be sufficient

rramkumar1:
/lgtm
/hold

Holding this since this will go in after the 1.5 cut.

Issue #537, opened 2018-11-05
Long namespace/ingress names cause collisions with auto-created resources on GKE


NickLavrov:
I noticed this after creating one ingress that worked fine, then creating another ingress which caused the first ingress to behave incorrectly (default backend 404, etc).

As far as I can tell, at least these resources and annotations are added to the ingress by GKE:
- ingress.kubernetes.io/forwarding-rule
- ingress.kubernetes.io/target-proxy
- ingress.kubernetes.io/url-map

If my namespace was called `xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx` (32 chars) and my ingress was named `xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`, I'd see resources with names like:
`k8s-um-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-xxxxxxxxxxxxx0`

This is fine if there's only one ingress, but since my other ingress also started with the same set of characters, there would be a collision. I've looked at the resources/annotations it creates when using shorter names, and I notice that it means to append some random characters at the end. Here, it looks like I'm hitting the 64 character limit before it gets to that random string.

Maybe removing the 64 character limit or adding documentation about how these names are generated would help with this. It took me a while to look into this annotation as the root of my issues.

NOTE: This is the file that contains the naming logic https://github.com/kubernetes/ingress-gce/blob/master/pkg/utils/namer.go

freehan:
Thanks for the bug report. We will try to fix this problem. The key is to maintain backward compatibility.

freehan:
Add toleration of long name to certain degree: ##537

NickLavrov:
Closing the issue since it was addressed in the PR above!

karlkfi:
Changes in #650 do not actually solve this issue completely.

It is still possible for GCLB name and backend name (Network Endpoint Group) to collide if the namespace name + ingress name are too long, especially between two clusters with similar names, because the cluster name is at the end and gets truncated first.

LB: https://github.com/kubernetes/ingress-gce/blob/master/pkg/utils/namer.go#L340
NEG: https://github.com/kubernetes/ingress-gce/blob/master/pkg/utils/namer.go#L430

The UID also commonly gets truncated, which increases the chance of UID collisions.

I like that the cluster/namespace/ingress names are in these GCLB names, because it makes them easier to identify when debugging, but the whole name-generation scheme needs to be redesigned to better guarantee uniqueness.

[...]

PR #650, opened 2019-02-26
Modify NameBelongToCluter to tolerate truncated cluster name suffix

freehan:
ref: #537

### DIFF
--- a/pkg/utils/namer.go
+++ b/pkg/utils/namer.go
@@ -69,6 +69,12 @@ const (
 	// avoid terminating on an invalid character ('-').
 	nameLenLimit = 62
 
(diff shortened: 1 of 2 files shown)


pkg/utils/namer.go
```diff
@@ -226,7 +232,7 @@ func (n *Namer) ParseName(name string) *NameComponents {
 	if resource == urlMapPrefix {
 		// It is possible for the loadbalancer name to have dashes in it - so we
 		// join the remaining name parts.
-		lbName = strings.Join(c[2:len(c)], "-")
+		lbName = strings.Join(c[2:], "-")
```
freehan: gofmt removed this

PR #655, opened 2019-02-28
Rebase base images from scratch and alpine to distroless

yuwenma:
[Kubernetes issue/70249
](https://github.com/kubernetes/kubernetes/issues/70249)

Internal Doc: go/rebase-gke-image-to-distroless

Test: Tested the Dockerfile files validation via "make build bin".

### DIFF
--- a/Dockerfile.404-server
+++ b/Dockerfile.404-server
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
+FROM gcr.io/distroless/static:latest
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
 
--- a/Dockerfile.fuzzer
+++ b/Dockerfile.fuzzer
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.8
-
-RUN apk add --no-cache ca-certificates
+# This image requires ca-certificate, which is pre-installed in distroless.
+FROM gcr.io/distroless/static:latest
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
 ENTRYPOINT ["/ARG_BIN"]
--- a/Dockerfile.glbc
+++ b/Dockerfile.glbc
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.8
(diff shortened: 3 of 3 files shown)


Dockerfile.404-server
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
+FROM gcr.io/distroless/static:latest
```
MrHohn: Curious, why is using `latest` better than a fixed version, which would be reproducible?

yuwenma: Good question! I guess we are less worried that the static distroless image is the root cause of some k8s problems. So considering the maintenance cost of updating fixed distroless versions periodically versus reproducing an error potentially caused by a bad distroless version, the former reason is escalated.

Actually, @timstclair is the better person to answer the question.

rramkumar1:
/lgtm
/hold

Will merge once discussion around latest image is resolved.

Dockerfile.404-server
tallclair: Yeah, it's a tradeoff and I don't have a good answer. We don't have any way of ensuring that images are based on an up-to-date version of distroless/static right now, so using latest reduces the maintenance burden on us. Note that you can still reproduce the build by pulling the previous version, and using the same sha for the base.

MrHohn: Thanks both! Good point that we can use the same sha. That answered my question.

rramkumar1: Thanks, that sounds fine. I'll unblock this now.

bowei:
We should change to Debian. Going distroless will require more testing.

tallclair:
@bowei basing on debian has a much higher maintenance cost (rebasing for CVE fixes). I'd much rather we do the testing work now, and avoid that maintenance overhead.

bowei:
Hmm -- is the only dep /bin/sh in the YAML? Maybe we can get rid of that?

tallclair:
Yeah, looks like it's just doing a log redirect. I suggest switching from glog to klog, and using the --log-file argument.

rramkumar1:
I'm testing out removal of the sh dependency now.

yuwenma:
Rohit and I just synced up offline. After Rohit changes the cmd to remove log redirect, we can revert back [the "revert" PR](https://github.com/kubernetes/ingress-gce/pull/664). Thanks Rohit!

PR #664, opened 2019-03-05
Revert "Rebase base images from scratch and alpine to distroless"

rramkumar1:
Reverts kubernetes/ingress-gce#655.

Since this was merged, our CI has been failing [1] with the following error:

oci runtime error: container_linux.go:247: starting container process caused "exec: \"sh\": executable file not found in $PATH"

[1] https://k8s-testgrid.appspot.com/sig-network-ingress-gce-e2e

PR #665, opened 2019-03-05
Rebase from scratch images to distroless

yuwenma:
Only rebase the empty base image to distroless. Previous PR 655 rebased both empty and alpine and was reverted in [PR/664](https://github.com/kubernetes/ingress-gce/pull/664/files) due to the fact that the alpine images are run in kubelet with bash commands (which distroless doesn't have).

### DIFF
--- a/Dockerfile.404-server
+++ b/Dockerfile.404-server
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM scratch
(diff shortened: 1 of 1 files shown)


yuwenma:
Close the PR. Due to the offline discussion with Rohit, Rohit will change the cmd to remove log redirect (the only place shell is required). After that is done, we are save to revert the "revert" PR. This PR is not longer needed.

PR #671, opened 2019-03-07
Add tests for HTTPS 

rramkumar1:
Porting from k/k

This PR is already pretty big so I'll add transition tests in a followup

### DIFF
--- a/cmd/e2e-test/basic_https_test.go
+++ b/cmd/e2e-test/basic_https_test.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestBasicHTTPS(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	for _, tc := range []struct {
+		desc       string
+		ingBuilder *fuzz.IngressBuilder
+		hosts      []string
+		certType   e2e.CertType
+
+		numForwardingRules int
+		numBackendServices int
+	}{
+		{
+			desc: "http(s) one path via pre-shared cert",
+			ingBuilder: fuzz.NewIngressBuilder("", "ingress-1", "").
+				DefaultBackend("service-1", port80).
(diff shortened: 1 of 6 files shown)


rramkumar1:
@bowei @MrHohn This is rebased and ready for a full review.

cmd/e2e-test/basic_https_test.go
```diff
@@ -0,0 +1,145 @@
...
+				t.Fatalf("error creating echo service: %v", err)
+			}
+			t.Logf("Echo service created (%s/%s)", s.Namespace, "service-1")
+
+			if _, err := Framework.Clientset.Extensions().Ingresses(s.Namespace).Create(ing); err != nil {
+				t.Fatalf("error creating Ingress spec: %v", err)
+			}
+			t.Logf("Ingress created (%s/%s)", s.Namespace, ing.Name)
+
+			ing, err = e2e.WaitForIngress(s, ing, nil)
```
MrHohn: I don't recall we have logic implemented to check if server responses with the correct cert based on request hostname. Similar to this https://github.com/kubernetes/kubernetes/blob/118e33dfcd8bf052f34d9a922145ab5f97e6a354/test/e2e/framework/ingress/ingress_utils.go#L638-L670.

Is that still necessary? (Maybe as followup if needed.)

rramkumar1: Maybe I'm missing something but I don't know if that check is worth it. Is there a case when we could potentially configure GCLB to serve the wrong cert? My thinking was that this could only happen if the user made a config error.

pkg/e2e/fixtures.go
```diff
@@ -153,3 +156,68 @@ func EnsureIngress(s *Sandbox, ing *v1beta1.Ingress) (*v1beta1.Ingress, error) {
 
 	return currentIng, nil
 }
+
+// DeleteSecret deletes a secret.
+func DeleteSecret(s *Sandbox, name string) error {
```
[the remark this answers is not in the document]
rramkumar1: Good point, done.

cmd/e2e-test/basic_https_test.go
```diff
@@ -0,0 +1,145 @@
...
+		Framework.RunWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			t.Parallel()
+
+			ctx := context.Background()
+
+			for i, h := range tc.hosts {
+				name := fmt.Sprintf("cert%d--%s", i, s.Namespace)
+				cert, err := e2e.NewCert(name, h, tc.certType)
+				if err != nil {
+					t.Fatalf("error initializing cert: %v", err)
```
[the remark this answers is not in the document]
rramkumar1: Done

cmd/e2e-test/basic_https_test.go
MrHohn: One case that comes to my mind is in case of using multiple certs, if there is a bug that ingress controller attaches only one cert to GCLB, it will still pass this test. Should we at least check that all certs are attached?

bowei: Can we put a TODO / issue to fix this? We do need to check this behavior.

rramkumar1: Good point, I think checking that all certs are attached may be more worthwhile than checking if the "correct" cert is served. Will tackle that in the followup.

pkg/e2e/fixtures.go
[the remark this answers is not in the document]
MrHohn: Oh..seems like this func is forgotten?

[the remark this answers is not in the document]
rramkumar1: I purposefully left just that one since there may be a use case later to delete a secret not related to a cert.

ok09:
pull request

Issue #675, opened 2019-03-11
Changing app-protocol from HTTP2 to other protocols will cause glbc to hit nil pointer 


freehan:
**Symptom:** Create an ingress with HTTP2 app protocol. For instance: https://github.com/kubernetes/kubernetes/tree/master/test/e2e/testing-manifests/ingress/http2
Manually change the app protocol on the spec of service from HTTP2 to something like HTTPS,HTTP. This will cause glbc to hit nil pointer and crash with following stacktrace in ingress-gce 1.2:

```
/go/src/k8s.io/ingress-gce/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
/go/src/k8s.io/ingress-gce/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/go/src/k8s.io/ingress-gce/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/local/go/src/runtime/asm_amd64.s:573
/usr/local/go/src/runtime/panic.go:502
/usr/local/go/src/runtime/panic.go:63
/usr/local/go/src/runtime/signal_unix.go:388
/go/src/k8s.io/ingress-gce/pkg/healthchecks/healthchecks.go:332
/go/src/k8s.io/ingress-gce/pkg/healthchecks/healthchecks.go:244
/go/src/k8s.io/ingress-gce/pkg/healthchecks/healthchecks.go:106
/go/src/k8s.io/ingress-gce/pkg/backends/backends.go:220
/go/src/k8s.io/ingress-gce/pkg/backends/backends.go:283
/go/src/k8s.io/ingress-gce/pkg/backends/backends.go:252
/go/src/k8s.io/ingress-gce/pkg/controller/cluster_manager.go:98
/go/src/k8s.io/ingress-gce/pkg/controller/controller.go:349
/go/src/k8s.io/ingress-gce/pkg/controller/controller.go:299
/go/src/k8s.io/ingress-gce/pkg/controller/controller.go:111
/go/src/k8s.io/ingress-gce/pkg/utils/taskqueue.go:84
/go/src/k8s.io/ingress-gce/pkg/utils/taskqueue.go:54
/go/src/k8s.io/ingress-gce/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
/go/src/k8s.io/ingress-gce/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
[...]

PR #678, opened 2019-03-12
examine nil pointer before applying backend protocol configurations

freehan:
Mitigates #675

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -62,6 +62,11 @@ const (
 	// backends, the port or named port specified in the Backend Service is
 	// used for health checking.
 	UseServingPortSpecification = "USE_SERVING_PORT"
+
+	// TODO: revendor the GCE API go client so that this error will not be hit.
+	newHealthCheckErrorMessageTemplate = "the %v health check configuration on the existing health check %v is nil. " +
+		"This is usually caused by an application protocol change on the k8s service spec. " +
+		"Please revert the change on application protocol to avoid this error message."
 )
 
 // HealthChecks manages health checks.
@@ -241,7 +246,10 @@ func (h *HealthChecks) Get(name string, version meta.Version) (*HealthCheck, err
 	default:
 		return nil, fmt.Errorf("unknown version %v", version)
 	}
-	return NewHealthCheck(hc), err
+	if err != nil {
+		return nil, err
+	}
+	return NewHealthCheck(hc)
 }
 
 // DefaultHealthCheck simply returns the default health check.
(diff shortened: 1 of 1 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -62,6 +62,10 @@ const (
 	// backends, the port or named port specified in the Backend Service is
 	// used for health checking.
 	UseServingPortSpecification = "USE_SERVING_PORT"
+
+	newHealthCheckErrorMessageTemplate = "the %v health check configuration on the exiting health check %v is nil. " +
+		"This is usually caused by application protocol changed. " +
```
rramkumar1: what do you mean by exiting?

rramkumar1: "..caused by an application..."

pkg/healthchecks/healthchecks.go
```diff
@@ -62,6 +62,10 @@ const (
...
+		"Please revert the change on application protocol to avoid this error message."
```
rramkumar1: "...revert the change on the application..."

pkg/healthchecks/healthchecks.go
cmluciano:
```suggestion
		"This is usually caused by a change of the application's protocol." +
```

pkg/healthchecks/healthchecks.go
cmluciano:
```suggestion
		"Please revert the change on the application's protocol to avoid this error message."
```

I don't have a lot of insight into how the ingress-gce works internally, but I think a TODO on handling this type of change would be helpful. I think progressing from HTTP-1.1 to HTTP-2 (soon HTTP-3) or introducing changes on the protocol to adopt a service mesh will be a common occurrence.

freehan: yeah. Agreed. In the long run we want to support this. But currently, due to GCE api changes, it was broken. So we need to put this short term fix in place so that it can be cherry picked.

freehan:
fixed.

mariantalla:
Hi, is there anything missing for this to be `/lgtm`ed?
cc @MrHohn @thockin

mariantalla:
Assigning for `lgtm` (please reassign if I've done this wrong):
/assign @MrHohn

MrHohn:
This LGTM, but will defer to @rramkumar1 to make the final call.
/assign @rramkumar1

pkg/healthchecks/healthchecks.go
```diff
@@ -62,6 +62,10 @@ const (
 	// backends, the port or named port specified in the Backend Service is
 	// used for health checking.
 	UseServingPortSpecification = "USE_SERVING_PORT"
+
+	newHealthCheckErrorMessageTemplate = "the %v health check configuration on the existing health check %v is nil. " +
+		"This is usually caused by an application protocol changed. " +
```
bowei: can we give the name of the actual field involved, this error is a bit vague sounding. are we talking about the GCP object, the protocol used by the app itself, etc

pkg/healthchecks/healthchecks.go
```diff
@@ -304,18 +311,30 @@ type HealthCheck struct {
...
 		v.HTTPHealthCheck = *hc.HttpHealthCheck
 	case annotations.ProtocolHTTPS:
+		if hc.HttpsHealthCheck == nil {
+			err = fmt.Errorf(newHealthCheckErrorMessageTemplate, annotations.ProtocolHTTPS, hc.Name)
+			break
+		}
 		v.HTTPHealthCheck = computealpha.HTTPHealthCheck(*hc.HttpsHealthCheck)
 	case annotations.ProtocolHTTP2:
+		if hc.Http2HealthCheck == nil {
+			err = fmt.Errorf(newHealthCheckErrorMessageTemplate, annotations.ProtocolHTTP2, hc.Name)
```
bowei: can we not return immediately (same in all case above)

pkg/healthchecks/healthchecks.go
freehan: HealthCheck object contains HTTPHealthCheck configurations. It has bunch of fields. 

In this case, the controller thinks that HTTP2HealthCheck configuration is nil because it was not vendored.

freehan:
Ready for next round

pkg/healthchecks/healthchecks.go
```diff
@@ -62,6 +62,10 @@ const (
 	// backends, the port or named port specified in the Backend Service is
 	// used for health checking.
 	UseServingPortSpecification = "USE_SERVING_PORT"
+
+	newHealthCheckErrorMessageTemplate = "the %v health check configurations on the existing health check %v is nil. " +
```
rramkumar1: "the %v health check configuration..."

pkg/healthchecks/healthchecks.go
rramkumar1: Can we add the TODO just so we remember to do it?

freehan:
Done

Issue #667, opened 2019-03-06
Move e2e tests from k/k into this repository


rramkumar1:
Many of the Ingress-GCE e2e tests defined in k/k [1] can also be defined in this repo and will be much easier to maintain in this repo. At then end of this migration, the only tests in k/k that should remain are the conformance tests.

Using this issue to track the required changes. 

[1] https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/ingress.go

rramkumar1:
- Upgrade tests
 [x] Removal of scaffolding - https://github.com/kubernetes/kubernetes/pull/75002
 [x]  Removal of prow jobs - https://github.com/kubernetes/test-infra/pull/11638

PR #668, opened 2019-03-06
Move tests for app protocols from k/k to Ingress-GCE repo

rramkumar1:
The PR is split into two commits:

1. Extract echoheaders response body into public struct so it can be used elsewhere
2. Porting of tests

Note that the PR to remove the tests from k/k is https://github.com/kubernetes/kubernetes/pull/75143

PR #679, opened 2019-03-12
Port basic NEG test from k/k

spencerhance:
Related Issue: https://github.com/kubernetes/ingress-gce/issues/667

Adds a basic NEG test "TestNEGBasic" that tests the functionality of enabling/disabling NEG for ingress, and verifying that the corresponding actions/GC are correct.

Adds NEG as a component to the fuzz validator framework

Adds helper functions:
- pkg/annotations/service.go: 
  - NegAnnotation.String()
  - Service.NEGAnnotation()
  - Service.NEGStatus()
- pkgg/e2e/helpers.go
  - WaitForNEGDeletion()
  - WaitForNEGConfiguration()
- pkg/e2e/fixtures.go
  - EnsureEchoService()
  - EnsureIngress()
- pkg/fuzz/gcp.go
  - GCLB.CheckNEGDeletion()
- pkg/neg/types/types.go
  - ParseNegStatus()
- pkg/fuzz/features/neg.go
  - verifyNegBackend()
- pkg/fuzz/helpers.go
  - ServiceForPath()

### DIFF
--- a/cmd/e2e-test/cdn_test.go
+++ b/cmd/e2e-test/cdn_test.go
@@ -84,7 +84,7 @@ func TestCDN(t *testing.T) {
 			}
 			t.Logf("BackendConfig created (%s/%s) ", s.Namespace, tc.beConfig.Name)
 
-			_, _, err := e2e.CreateEchoService(s, "service-1", backendConfigAnnotation)
+			_, err := e2e.CreateEchoService(s, "service-1", backendConfigAnnotation)
 			if err != nil {
 				t.Fatalf("error creating echo service: %v", err)
 			}
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -1,0 +1,232 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"testing"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestNEG(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	ctx := context.Background()
+
+	for _, tc := range []struct {
+		desc               string
+		annotations        annotations.NegAnnotation
+		negExpected        bool
+		numForwardingRules int
+		numBackendServices int
+	}{
+		{
+			desc:               "Create a basic NEG",
+			annotations:        annotations.NegAnnotation{Ingress: true},
+			negExpected:        true,
+			numForwardingRules: 1,
+			numBackendServices: 1,
+		},
+		{
+			desc:               "Annotation with no NEG",
+			annotations:        annotations.NegAnnotation{Ingress: false},
+			negExpected:        false,
+			numForwardingRules: 1,
+			numBackendServices: 1,
+		},
+	} {
+		tc := tc // Capture tc as we are running this in parallel.
+		Framework.RunWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			_, err := e2e.EnsureEchoService(s, "service-1", map[string]string{
+				annotations.NEGAnnotationKey: tc.annotations.String()}, v1.ProtocolTCP, 80, v1.ServiceTypeNodePort)
+			if err != nil {
+				t.Fatalf("error ensuring echo service: %v", err)
+			}
+			t.Logf("Echo service ensured (%s/%s)", s.Namespace, "service-1")
+
+			// Create the ingress
+			ing := fuzz.NewIngressBuilder("", "ingress-1", "").
+				DefaultBackend("service-1", port80).
+				Build()
+			ing, err = e2e.EnsureIngress(s, ing)
+			if err != nil {
+				t.Fatalf("error ensuring Ingress spec: %v", err)
+			}
+			t.Logf("Ingress ensured (%s/%s)", s.Namespace, ing.Name)
+
+			ing, err = e2e.WaitForIngress(s, ing, nil)
+			if err != nil {
+				t.Fatalf("error waiting for Ingress to stabilize: %v", err)
+			}
+			t.Logf("GCLB resources created (%s/%s)", s.Namespace, ing.Name)
+
+			// Perform whitebox testing.
+			if len(ing.Status.LoadBalancer.Ingress) < 1 {
+				t.Fatalf("Ingress does not have an IP: %+v", ing.Status)
+			}
+
+			vip := ing.Status.LoadBalancer.Ingress[0].IP
+			t.Logf("Ingress %s/%s VIP = %s", s.Namespace, ing.Name, vip)
+			gclb, err := fuzz.GCLBForVIP(context.Background(), Framework.Cloud, vip,
+				fuzz.FeatureValidators(features.All))
+			if err != nil {
+				t.Fatalf("Error getting GCP resources for LB with IP = %q: %v", vip, err)
+			}
+
+			// Do some cursory checks on the GCP objects.
+			if len(gclb.ForwardingRule) != tc.numForwardingRules {
+				t.Errorf("got %d fowarding rules, want %d", len(gclb.ForwardingRule), tc.numForwardingRules)
+			}
+			if len(gclb.BackendService) != tc.numBackendServices {
+				t.Errorf("got %d backend services, want %d", len(gclb.BackendService), tc.numBackendServices)
+			}
+
+			if (len(gclb.NetworkEndpointGroup) > 0) != tc.negExpected {
+				t.Errorf("Error: negExpected = %v, %d negs found for gclb %v", tc.negExpected, len(gclb.NetworkEndpointGroup), gclb)
+			}
+
+			if err := e2e.WaitForIngressDeletion(ctx, gclb, s, ing, &fuzz.GCLBDeleteOptions{}); err != nil {
+				t.Errorf("e2e.WaitForIngressDeletion(..., %q, nil) = %v, want nil", ing.Name, err)
+			}
+		})
+	}
+}
+
+func TestNEGTransition(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	ctx := context.Background()
+
+	Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
+
+		ing := fuzz.NewIngressBuilder("", "ingress-1", "").
+			DefaultBackend("service-1", port80).
+			Build()
+
+		var previousGCLBState *fuzz.GCLB
+
+		for _, tc := range []struct {
+			desc        string
+			annotations annotations.NegAnnotation
+			// negGC is true if a NEG should be garbage collected after applying the annotations
+			negGC              bool
+			numForwardingRules int
+			numBackendServices int
+		}{
+			{
+				desc:               "Using ingress only",
+				annotations:        annotations.NegAnnotation{Ingress: true},
+				negGC:              false,
+				numForwardingRules: 1,
+				numBackendServices: 1,
+			},
+			{
+				desc:               "Disable NEG for ingress",
+				annotations:        annotations.NegAnnotation{Ingress: false},
+				negGC:              true,
+				numForwardingRules: 1,
+				numBackendServices: 1,
+			},
+			{
+				desc:               "Re-enable NEG for ingress",
+				annotations:        annotations.NegAnnotation{Ingress: true},
+				negGC:              false,
+				numForwardingRules: 1,
+				numBackendServices: 1,
+			},
+			{
+				desc:               "No annotations",
+				annotations:        annotations.NegAnnotation{},
+				negGC:              true,
+				numForwardingRules: 1,
+				numBackendServices: 1,
+			},
+		} {
+			// First create the echo service, we will be adapting it throughout the basic tests
+			_, err := e2e.EnsureEchoService(s, "service-1", map[string]string{
+				annotations.NEGAnnotationKey: tc.annotations.String()}, v1.ProtocolTCP, 80, v1.ServiceTypeNodePort)
+			if err != nil {
+				t.Fatalf("error ensuring echo service: %v", err)
+			}
+			t.Logf("Echo service ensured (%s/%s)", s.Namespace, "service-1")
+
+			// Create the ingress
+			ing, err = e2e.EnsureIngress(s, ing)
+			if err != nil {
+				t.Fatalf("error ensuring Ingress spec: %v", err)
+			}
+			t.Logf("Ingress ensured (%s/%s)", s.Namespace, ing.Name)
+
+			ing, err = e2e.WaitForIngress(s, ing, nil)
+			if err != nil {
+				t.Fatalf("error waiting for Ingress to stabilize: %v", err)
+			}
+			t.Logf("GCLB resources created (%s/%s)", s.Namespace, ing.Name)
+
+			// Perform whitebox testing.
+			if len(ing.Status.LoadBalancer.Ingress) < 1 {
+				t.Fatalf("Ingress does not have an IP: %+v", ing.Status)
+			}
+
+			vip := ing.Status.LoadBalancer.Ingress[0].IP
+			t.Logf("Ingress %s/%s VIP = %s", s.Namespace, ing.Name, vip)
+			gclb, err := fuzz.GCLBForVIP(context.Background(), Framework.Cloud, vip,
+				fuzz.FeatureValidators(features.All))
+			if err != nil {
+				t.Fatalf("Error getting GCP resources for LB with IP = %q: %v", vip, err)
+			}
+
+			// Do some cursory checks on the GCP objects.
+			if len(gclb.ForwardingRule) != tc.numForwardingRules {
+				t.Errorf("got %d fowarding rules, want %d", len(gclb.ForwardingRule), tc.numForwardingRules)
+			}
+			if len(gclb.BackendService) != tc.numBackendServices {
+				t.Errorf("got %d backend services, want %d", len(gclb.BackendService), tc.numBackendServices)
+			}
+
+			if tc.negGC {
+				if len(gclb.NetworkEndpointGroup) != 0 {
+					t.Errorf("NegGC = true, expected 0 negs for gclb %v, got %d", gclb, len(gclb.NetworkEndpointGroup))
+				}
+				if err = e2e.WaitForNEGDeletion(ctx, s.ValidatorEnv.Cloud(), previousGCLBState, nil); err != nil {
+					t.Errorf("Error waiting for NEGDeletion: %v", err)
+				}
+			} else {
+				if len(gclb.NetworkEndpointGroup) < 1 {
+					t.Errorf("Error, no NEGS associated with gclb %v, expected at least one", gclb)
+				}
+			}
+			previousGCLBState = gclb
+		}
+
+		if ing != nil && previousGCLBState != nil {
+			if err := e2e.WaitForIngressDeletion(ctx, previousGCLBState, s, ing, &fuzz.GCLBDeleteOptions{}); err != nil {
+				t.Errorf("e2e.WaitForIngressDeletion(..., %q, nil) = %v, want nil", ing.Name, err)
+			}
+		}
+	})
+}
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"k8s.io/ingress-gce/pkg/neg/types"
 
 	"k8s.io/api/core/v1"
 )
@@ -107,6 +108,11 @@ func (n *NegAnnotation) NEGEnabled() bool {
 	return n.NEGEnabledForIngress() || n.NEGExposed()
 }
 
+func (n *NegAnnotation) String() string {
+	bytes, _ := json.Marshal(n)
+	return string(bytes)
+}
+
 // AppProtocol describes the service protocol.
 type AppProtocol string
 
@@ -159,19 +165,35 @@ var (
 
 // NEGAnnotation returns true if NEG annotation is found.
 // If found, it also returns NEG annotation struct.
-func (svc *Service) NEGAnnotation() (bool, *NegAnnotation, error) {
+func (svc *Service) NEGAnnotation() (*NegAnnotation, bool, error) {
 	var res NegAnnotation
 	annotation, ok := svc.v[NEGAnnotationKey]
 	if !ok {
-		return false, nil, nil
+		return nil, false, nil
 	}
 
 	// TODO: add link to Expose NEG documentation when complete
 	if err := json.Unmarshal([]byte(annotation), &res); err != nil {
-		return true, nil, ErrNEGAnnotationInvalid
+		return nil, true, ErrNEGAnnotationInvalid
 	}
 
-	return true, &res, nil
+	return &res, true, nil
+}
+
+func (svc *Service) NEGStatus() (*types.NegStatus, bool, error) {
+	var res types.NegStatus
+	var err error
+
+	annotation, ok := svc.v[NEGStatusKey]
+	if !ok {
+		return nil, false, nil
+	}
+
+	if res, err = types.ParseNegStatus(annotation); err != nil {
+		return nil, true, fmt.Errorf("Error parsing neg status: %v", err)
+	}
+
+	return &res, true, nil
 }
 
 type BackendConfigs struct {
--- a/pkg/annotations/service_test.go
+++ b/pkg/annotations/service_test.go
@@ -18,6 +18,7 @@ package annotations
 
 import (
 	"fmt"
+	"k8s.io/ingress-gce/pkg/neg/types"
 	"reflect"
 	"testing"
 
@@ -136,7 +137,7 @@ func TestNEGAnnotation(t *testing.T) {
 			exposed:    true,
 		},
 	} {
-		found, negAnnotation, err := FromService(tc.svc).NEGAnnotation()
+		negAnnotation, found, err := FromService(tc.svc).NEGAnnotation()
 		if fmt.Sprintf("%q", err) != fmt.Sprintf("%q", tc.expectError) {
 			t.Errorf("Test case %q: Expect error to be %q, but got: %q", tc.desc, tc.expectError, err)
 		}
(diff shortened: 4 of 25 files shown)


cmd/e2e-test/cdn_test.go
```diff
@@ -69,7 +69,7 @@ func TestCDN(t *testing.T) {
 				Build(),
 		},
 	} {
-		tc := tc // Capture tc as we are running this in parallel.
+		tc := tc // Capture tc as we are running this in parallel
```
freehan: revert this one?

pkg/e2e/helpers.go
```diff
@@ -19,6 +19,9 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/neg/types"
 	"time"
```
freehan: usually we seperate the k8s.io imports and golang ones with an extra line.

pkg/e2e/helpers.go
```diff
@@ -28,6 +31,7 @@ import (
 	"k8s.io/ingress-gce/pkg/fuzz/features"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce/cloud"
+	//"k8s.io/ingress-gce/pkg/utils"
```
freehan: remove this

pkg/utils/namer.go
```diff
@@ -427,7 +427,7 @@ func (n *Namer) NEG(namespace, name string, port int32) string {
 	truncNamespace := truncFields[0]
 	truncName := truncFields[1]
 	truncPort := truncFields[2]
-	return fmt.Sprintf("%s-%s-%s-%s-%s", n.negPrefix(), truncNamespace, truncName, truncPort, negSuffix(n.shortUID(), namespace, name, portStr))
+	return fmt.Sprintf("%s-%s-%s-%s-%s", n.negPrefix(), truncNamespace, truncName, truncPort, NegSuffix(n.shortUID(), namespace, name, portStr))
```
freehan: revert this?

pkg/annotations/service.go
```diff
@@ -174,6 +180,22 @@ func (svc *Service) NEGAnnotation() (bool, *NegAnnotation, error) {
 	return true, &res, nil
 }
 
+func (svc *Service) NEGStatus() (bool, *types.NegStatus, error) {
```
freehan: usually `ok` or `exists`  is returned as the 2nd value. 

 (bool, *types.NegStatus, error)

pkg/backends/features/securitypolicy_test.go
```diff
@@ -50,7 +50,7 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 		expectSetCall         bool
 	}{
 		{
-			desc: "attach-policy",
+			desc:                  "attach-policy",
```
freehan: revert?

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	port80 := intstr.FromInt(80)
+
+	numForwardingRules := 1
+	numBackendServices := 1
+
+	ctx := context.Background()
+
+	Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
+		//ctx := context.Background()
```
freehan: delete

rramkumar1:
@spencerhance Can you squash the commits?

pkg/e2e/fixtures.go
```diff
@@ -29,6 +29,27 @@ const (
 	echoheadersImage = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
 )
 
+func CreateService(s *Sandbox, name string, ports []v1.ServicePort, annotations map[string]string) (*v1.Service, error) {
```
freehan: Let us leave the previous implementation here. We can evolve this later. 

There are 2 directions I am thinking: 
1. have a  complex enough service template with some parameters to tweak. (annotation and service type)
2. allow user to fully generate the service spec.

pkg/fuzz/validator.go
```diff
@@ -283,17 +284,18 @@ func (v *IngressValidator) CheckPaths(ctx context.Context, vr *IngressResult) er
 }
 
 // checkPath performs a check for scheme://host/path.
-func (v *IngressValidator) checkPath(ctx context.Context, scheme, host, path string) error {
+// Returns HTTP status code and error
+func (v *IngressValidator) checkPath(ctx context.Context, scheme, host, path string) (int, error) {
```
freehan: I do not think you need to return response code per each checkPath, right?

freehan: checkPath should return an aggregated result after checking everything (including status code )

spencerhance: This was added as further debug information but I can remove it

pkg/fuzz/validator.go
```diff
@@ -140,10 +140,11 @@ type IngressResult struct {
...
 type PathResult struct {
-	Scheme string
-	Host   string
-	Path   string
-	Err    error
+	Scheme     string
+	Host       string
+	Path       string
+	Err        error
+	StatusCode int
```
freehan: PathResult is the result. So it should not care of the detail reponse status code

pkg/neg/types/types.go
```diff
@@ -52,3 +54,10 @@ type NegStatus struct {
 	NetworkEndpointGroups PortNameMap `json:"network_endpoint_groups,omitempty"`
 	Zones                 []string    `json:"zones,omitempty"`
 }
+
+// ParseNegStatus parses the given annotation into NEG status struct
+func ParseNegStatus(annotation string) (NegStatus, error) {
```
freehan: add a unit test

pkg/e2e/helpers.go
```diff
@@ -95,3 +104,34 @@ func WaitForGCLBDeletion(ctx context.Context, c cloud.Cloud, g *fuzz.GCLB, optio
...
+		return true, nil
+	})
+}
+
+func WaitForNEGConfiguration(svc *v1.Service, f *Framework, s *Sandbox) error {
+	return wait.Poll(negPollInterval, negPollTimeout, func() (bool, error) {
+		// Get Annotation
+		svc, _ = f.Clientset.CoreV1().Services(s.Namespace).Get(svc.Name, metav1.GetOptions{})
+
+		if negStatusString, ok := svc.Annotations[annotations.NEGStatusKey]; ok {
```
freehan: Use NegStatus function instead? it is up to you.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+		// Do some cursory checks on the GCP objects.
+		if len(gclb.ForwardingRule) != numForwardingRules {
+			t.Errorf("got %d fowarding rules, want %d", len(gclb.ForwardingRule), numForwardingRules)
+		}
+		if len(gclb.BackendService) != numBackendServices {
+			t.Errorf("got %d backend services, want %d", len(gclb.BackendService), numBackendServices)
+		}
+
+		for _, tc := range []struct {
```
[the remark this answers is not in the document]
spencerhance: test case 1: 
ingress + NEG

test caes 2: 
INgress + IG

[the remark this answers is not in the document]
spencerhance:
```
var gclbPreviousState gclb

for each tc {
      EnsureEchoService
     EnsureIngress    
     WaitForIngress
     gclb := GCLBFromVIP
     whitebox testing
     if negGC {
        previousGclb.WaitForNEGDeletion
     }    

     previousState = gclb      
}


```

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,182 @@
...
+	ctx := context.Background()
+
+	Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
+
+		var previousGCLBState *fuzz.GCLB
+		var ing *v1beta1.Ingress
+
+		for _, tc := range []struct {
+			desc        string
+			annotations map[string]string
```
freehan: Instead of using annotation, we can use NEGAnnotation type directly. This can also get rid of the `newNegAnnotation` function?
```
for _, tc := []struct{
   desc string
   negAnnotation NegAnnotation
   ing *Ingress
   negGC bool
}{
   annotations := map[string]string{annotations.NegAnnotation.Key: tc.negAnnotation.String()}
   EnsureEchoService(........, annotations)
}

```

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,182 @@
...
+				t.Fatalf("Error getting GCP resources for LB with IP = %q: %v", vip, err)
+			}
+
+			// Do some cursory checks on the GCP objects.
+			if len(gclb.ForwardingRule) != numForwardingRules {
+				t.Errorf("got %d fowarding rules, want %d", len(gclb.ForwardingRule), numForwardingRules)
+			}
+			if len(gclb.BackendService) != numBackendServices {
+				t.Errorf("got %d backend services, want %d", len(gclb.BackendService), numBackendServices)
+			}
```
freehan: also check Number of NEGs?
```
If negGC {
  check   len(gclb.NetworkEndpointGroups) == 0
} else {
  check   len(gclb.NetworkEndpointGroups) > 0
}
```

pkg/e2e/helpers.go
```diff
@@ -36,16 +39,22 @@ const (
 
 	gclbDeletionInterval = 30 * time.Second
 	gclbDeletionTimeout  = 15 * time.Minute
+
+	negPollInterval = 30 * time.Second
+	negPollTimeout  = 5 * time.Minute
```
freehan: negPollInterval = 5 * time.Second
negPollTimeout  = 3 * time.Minute

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,150 @@
...
+
+	ctx := context.Background()
+
+	negStatus, found, err := annotationSvc.NEGStatus()
+
+	if err != nil {
+		return fuzz.CheckResponseContinue, fmt.Errorf("Error obtaining Neg status for service: %v, error: %v", svc, err)
+	}
+
+	// Do verification checks
```
freehan: // Do whitebox verifications
// Check if the corresponding backend service is targeting NEGs instead of IGs
// WARNING: assumes the backend service naming scheme is using the NEG naming scheme

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+}
+
+func backendServicesAreNeg(ctx *context.Context, env fuzz.ValidatorEnv, negs *types.PortNameMap) (bool, error) {
+	for _, name := range *negs {
+		nameFilter := filter.F{}
+		nameFilter.AndRegexp("name", name)
+
+		beServices, err := env.Cloud().BackendServices().List(*ctx, &nameFilter)
```
freehan: nit: Use `env.Cloud().BackendServices().Get(name)?`

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,150 @@
...
+			return false, err
+		}
+
+		if len(beServices) == 0 {
+			return false, fmt.Errorf("No backend services returned for name %s", name)
+		}
+
+		for _, beService := range beServices {
+			for _, be := range beService.Backends {
+				if !strings.Contains(be.Group, fuzz.NegResourceType) {
```
freehan: you can also check 
```
if !strings.Contains(be.Group, name) 
```

I think name for both Backend-Service and NEGs are the same

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,150 @@
...
+// Check that the neg is being used for the path if it is configured
+func (v *negValidator) CheckResponse(host, path string, resp *http.Response, body []byte) (fuzz.CheckResponseAction, error) {
+	svcName, _, err := fuzz.ServiceForPath(host, path, v.ing)
+
+	if err != nil {
+		klog.V(2).Infof("Unable to get service for path %v, error: ", path, err)
+		return fuzz.CheckResponseContinue, nil
+	}
+	// Get service
+	services, _ := v.env.Services()
```
freehan: I think servicePort is actually useful for you. 

The NEG status has a key with service port. So `negStatus.NetworkEndpointsGroups[servicePort]` should help you pin point the corresponding NEG wrt to the host && path combination.

So, later you can do this:
```
backendServicesAreNeg(&ctx, v.env, &negStatus.NetworkEndpointsGroups[servicePort])
```

This will also eliminate future problems when you have exposed NEG and ingress NEG both enabled.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,182 @@
...
+			svc, _ = Framework.Clientset.CoreV1().Services(s.Namespace).Get(svc.Name, metav1.GetOptions{})
+			svc.ObjectMeta.Annotations = tc.annotations
+			svc, err = Framework.Clientset.CoreV1().Services(s.Namespace).Update(svc)
+			if err != nil {
+				t.Fatalf("error updating annotation for service: %v, error: %v", svc, err)
+			}
+
+			// Wait for neg configuration
+			if !tc.negGC {
+				err := e2e.WaitForNEGConfiguration(svc, Framework, s)
```
freehan: I do not think you need `WaitForNEGConfiguration`.

If you EnsureEchoService and WaitForIngress, the NEGStatus should already be updated

freehan: Do not remove `WaitForNEGConfiguration` though? I would be useful for other cases. (Standalone NEGs)

cmd/e2e-test/neg_test.go
spencerhance: I think this makes sense for negGC = false, however I'm not sure that checking the number of NEGs with negGC = true is necessary since it will always be 0 (WaitForNEGDeletion() will always run until all NEGs associated with the GCLB are deleted)

pkg/fuzz/features/neg.go
spencerhance: is there a way to do this without knowing the zone and region?

spencerhance: nevermind, it looks like it does work with just the name

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,146 @@
...
+	} else {
+		if found {
+			return fuzz.CheckResponseContinue, fmt.Errorf("Error, path %v for service %v should not be targetting a neg", path, svc)
+		} else {
+			return fuzz.CheckResponseContinue, nil
+		}
+	}
+}
+
+func backendsAreNeg(ctx *context.Context, env fuzz.ValidatorEnv, negName string) (bool, error) {
```
freehan: change the name to `verifyNegBackend`?

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestNEGBasic(t *testing.T) {
```
rramkumar1: Can you call this TestNEG? 

We use regex to filter which tests should run and there is another test called "TestBasic" so I don't want the regex "Basic" to cover both that test and this one.

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,146 @@
...
+		return fuzz.CheckResponseContinue, fmt.Errorf("Unable to get service for path %v, error: %v", path, err)
+
+	}
+	// Get service
+	services, err := v.env.Services()
+	if err != nil {
+		return fuzz.CheckResponseContinue, fmt.Errorf("Unable to get services for env: %v", v.env)
+	}
+
+	svc := services[svcName]
```
freehan: svc, ok := services[svcName]
if !ok {
  return fmt.Errorf(....)
}

pkg/fuzz/helpers.go
```diff
@@ -29,30 +29,40 @@ import (
 	translatorutil "k8s.io/ingress-gce/pkg/controller/translator"
 )
 
-// BackendConfigForPath returns the BackendConfig associated with the given path.
-// Note: This function returns an empty object (not nil pointer) if a BackendConfig
-// did not exist in the given environment.
-func BackendConfigForPath(host, path string, ing *v1beta1.Ingress, env ValidatorEnv) (*backendconfig.BackendConfig, error) {
+func ServiceForPath(host, path string, ing *v1beta1.Ingress) (string, intstr.IntOrString, error) {
```
rramkumar1: Can you return a *v1.Service and *v1.ServicePort instead of a string an IntOrString?

pkg/e2e/helpers.go
```diff
@@ -95,3 +103,35 @@ func WaitForGCLBDeletion(ctx context.Context, c cloud.Cloud, g *fuzz.GCLB, optio
 		return true, nil
 	})
 }
+
+// WaitForNEGDeletion waits for all NEGs associated with a GCLB to be deleted
+// Unlike WaitForIngressDeletion(), this *does not* actually delete anything
```
rramkumar1: Just curious why this does not actually delete anything?

freehan: It is waiting for NEGController to delete NEGs? Similar to WaitForGCLBDeletion.

pkg/e2e/helpers.go
```diff
@@ -36,16 +38,22 @@ const (
...
 type WaitForIngressOptions struct {
 	// ExpectUnreachable is true when we expect the LB to still be
 	// programming itself (i.e 404's / 502's)
 	ExpectUnreachable bool
 }
 
 // WaitForIngress to stabilize.
+// We expect the ingress to be unreachable at first as LB is
+// still programming itself (i.e 404's / 502's)
+// If no options are passed, uses defaultIngressStableWaitTime
```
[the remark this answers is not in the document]
spencerhance: whoops, forgot to remove this comment with the code

pkg/e2e/helpers.go
spencerhance: ^It's really waiting for the NEGs to be GC'd, I could change the name to something like `WaitForNegGC()` instead

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,152 @@
...
+				desc:        "No annotations",
+				annotations: annotations.NegAnnotation{},
+				ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+					DefaultBackend("service-1", port80).
+					Build(),
+				negGC: true,
+			},
+		} {
+			// Capture test case
+			tc := tc
```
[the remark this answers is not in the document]
spencerhance: Ah you're right, since our test cases are serialized here, I guess it doesn't matter

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,146 @@
...
+	// TODO: (shance) update this to support exposed_port as well
+	annotationSvc := annotations.FromService(svc)
+	negAnnotation, found, _ := annotationSvc.NEGAnnotation()
+
+	// Path does not have a NEG annotation, nothing to check here
+	if !found {
+		return fuzz.CheckResponseContinue, nil
+	}
+
+	ctx := context.Background()
```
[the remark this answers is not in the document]
spencerhance: I think I need this since I don't believe you can get the pointer of a return value in go

pkg/e2e/helpers.go
rramkumar1: Maybe I was confused by the comment then. It said "Unlike WaitForIngressDeletion() this does not..." but I don't see WaitForIngressDeletion() delete anything either.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	numForwardingRules := 1
+	numBackendServices := 1
+
+	ctx := context.Background()
+
+	Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
```
rramkumar1: Framework.RunWithSandbox should be executed for every test case. See basic_test.go for an example.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+		var previousGCLBState *fuzz.GCLB
+		var ing *v1beta1.Ingress
+
+		for _, tc := range []struct {
+			desc        string
+			annotations annotations.NegAnnotation
+			ing         *v1beta1.Ingress
+			negGC       bool // Should a NEG be garbage collected after applying the annotations
```
rramkumar1: Nit: Add the comment above the variable name and have it read something like "negGC is true if a NEG should be garbage collected after applying the annotations.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,149 @@
...
+			{
+				desc:        "Disable NEG for ingress",
+				annotations: annotations.NegAnnotation{Ingress: false},
+				ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+					DefaultBackend("service-1", port80).
+					Build(),
+				negGC: true,
+			},
+			{
+				desc:        "Re-enable NEG for ingress",
```
rramkumar1: By re-enable do you mean setting the annotation to true after it has been previously set to false? In that case, I would introduce separate test func called TestNEGTransition which holds all test cases that test such transitions.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,149 @@
...
+				annotations: annotations.NegAnnotation{Ingress: true},
+				ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+					DefaultBackend("service-1", port80).
+					Build(),
+				negGC: false,
+			},
+			{
```
rramkumar1: Is this testing something different from "Disable NEG for Ingress"?

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,149 @@
...
+	port80 := intstr.FromInt(80)
+
+	numForwardingRules := 1
+	numBackendServices := 1
+
+	ctx := context.Background()
+
+	Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
+
+		var previousGCLBState *fuzz.GCLB
```
rramkumar1: I would write another test func to test state transitions. See comment below.

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// CheckResponse implements fuzz.FeatureValidator.
+// Check that the neg is being used for the path if it is configured
+func (v *negValidator) CheckResponse(host, path string, resp *http.Response, body []byte) (fuzz.CheckResponseAction, error) {
+	svcName, svcPort, err := fuzz.ServiceForPath(host, path, v.ing)
+
+	if err != nil {
+		return fuzz.CheckResponseContinue, fmt.Errorf("Unable to get service for path %v, error: %v", path, err)
+
```
rramkumar1: Nit: extra space

cmd/e2e-test/neg_test.go
freehan: "No Annotations" means the neg annotation does not exists. "Disable NEG for Ingress" set it to false

pkg/fuzz/helpers.go
spencerhance: I'm not sure the best way to do this since I'm just using the IngressBackend types.  I guess I could return the backend directly or take in the env and look up the service myself.  Still looking into it. 
 Unless you mean to convert them to the api object types?

pkg/e2e/helpers.go
spencerhance: The first line of WaitForIngressDeletion is a Ingresses().Delete() call so I just added the comment for clarity.  If it actually makes things more confusing though I can remove the comment.

spencerhance: I guess since WaitForGCLBDeletion already uses the same functionality I probably don't need the comment.

pkg/fuzz/helpers.go
rramkumar1: Yeah I was thinking you can take in the env and look it up. See https://github.com/kubernetes/ingress-gce/pull/668/files for an example.

spencerhance: oh I see, will do, didn't realize you were also working on the same functionality

spencerhance:
@freehan @rramkumar1 Mind taking another look?

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,238 @@
...
+			desc:        "Annotation with no NEG",
+			annotations: annotations.NegAnnotation{Ingress: false},
+			ing: fuzz.NewIngressBuilder("", "ingress-1", "").
+				DefaultBackend("service-1", port80).
+				Build(),
+			numNegsExpected: 0,
+		},
+	} {
+		tc := tc // Capture tc as we are running this in parallel.
+
```
rramkumar1: Nit: remove space

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,238 @@
...
+		Framework.RunWithSandbox("Basic NEG Tests", t, func(t *testing.T, s *e2e.Sandbox) {
```
rramkumar1: Pass in tc.desc as the test name rather than "Basic NEG Tests".

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestNEG(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	numForwardingRules := 1
```
rramkumar1: This is optional but I would suggest making these two fields in the test struct. That way, if you ever add a test that requires more than 1 backend service, you can set it as such just for that test.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,238 @@
...
+		})
+	}
+}
+
+func TestNEGTransition(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+
+	numForwardingRules := 1
```
rramkumar1: See comment above.

pkg/annotations/service_test.go
```diff
@@ -167,6 +168,67 @@ func TestNEGAnnotation(t *testing.T) {
...
+		{
+			desc: "Basic NEG Status",
+			svc: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						NEGStatusKey: `{"network_endpoint_groups":{"80":"neg-name"},"zones":["us-central1-a"]}`,
+					},
+				},
+			},
+			expectNegStatus: &types.NegStatus{NetworkEndpointGroups: types.PortNameMap{80: "neg-name"}, Zones: []string{"us-central1-a"}},
```
rramkumar1: Missing desc

pkg/annotations/service_test.go
```diff
@@ -167,6 +168,67 @@ func TestNEGAnnotation(t *testing.T) {
 	}
 }
 
+func TestNEGStatus(t *testing.T) {
+	for _, tc := range []struct {
```
rramkumar1: Can you do a t.Run() for each test case? That way, if one of the test cases fails, it much easier to see which one it was.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,238 @@
...
+			// First create the echo service, we will be adapting it throughout the basic tests
+			_, err := e2e.EnsureEchoService(s, "service-1", map[string]string{
+				annotations.NEGAnnotationKey: tc.annotations.String()}, v1.ProtocolTCP, 80, v1.ServiceTypeNodePort)
+			if err != nil {
+				t.Fatalf("error ensuring echo service: %v", err)
+			}
+			t.Logf("Echo service ensured (%s/%s)", s.Namespace, "service-1")
+
+			// Create the ingress
+			ing, err = e2e.EnsureIngress(s, tc.ing)
```
rramkumar1: Since you never actually change the Ingress, you can just call EnsureIngress at the beginning (but after Framework.RunWithSandbox)

pkg/fuzz/features/neg.go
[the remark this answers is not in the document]
freehan: I think you can do this ? `&context.Background() `

pkg/annotations/service_test.go
spencerhance: Not sure I understand, it looks like all the tests have a desc

rramkumar1: Oops my bad, my eyesight is bad

pkg/fuzz/features/neg.go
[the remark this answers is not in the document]
spencerhance: https://stackoverflow.com/questions/30744965/how-to-get-the-pointer-of-return-value-from-function-call

pkg/annotations/service_test.go
spencerhance: no worries! :smile:

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,232 @@
...
+		desc               string
+		annotations        annotations.NegAnnotation
+		numNegsExpected    int
+		numForwardingRules int
+		numBackendServices int
+	}{
+		{
+			desc:               "Create a basic NEG",
+			annotations:        annotations.NegAnnotation{Ingress: true},
+			numNegsExpected:    1,
```
freehan: Please do not use number here. Use boolean instead

For a cluster across multiple zones. The number of NEGs will equal to the number of zone.

cmd/e2e-test/neg_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+			vip := ing.Status.LoadBalancer.Ingress[0].IP
+			t.Logf("Ingress %s/%s VIP = %s", s.Namespace, ing.Name, vip)
+			gclb, err := fuzz.GCLBForVIP(context.Background(), Framework.Cloud, vip,
+				fuzz.FeatureValidators(features.All))
+			if err != nil {
+				t.Fatalf("Error getting GCP resources for LB with IP = %q: %v", vip, err)
+			}
+
+			// Do some cursory checks on the GCP objects.
```
bowei: do this in a follow-on: let's put this in a helper function, I see it's being cut-and-paste a bunch of times.

pkg/e2e/helpers.go
```diff
@@ -95,3 +102,36 @@ func WaitForGCLBDeletion(ctx context.Context, c cloud.Cloud, g *fuzz.GCLB, optio
 		return true, nil
 	})
 }
+
+// WaitForNEGDeletion waits for all NEGs associated with a GCLB to be deleted via GC
+func WaitForNEGDeletion(ctx context.Context, c cloud.Cloud, g *fuzz.GCLB, options *fuzz.GCLBDeleteOptions) error {
+	return wait.Poll(negPollInterval, negPollTimeout, func() (bool, error) {
+		if err := g.CheckNEGDeletion(ctx, c, options); err != nil {
+			klog.Infof("format: WaitForNegDeletion(%q) = %v", g.VIP, err)
```
bowei: what does format mean here? Also, log lines are Capitalized.

pkg/fuzz/features/neg.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// NEG is a feature in GCP to support pod as Loadbalancer backends
+var NEG = &NegFeature{}
+
+// NegFeature implements the associated feature.
+type NegFeature struct{}
+
+// NewValidator implements fuzz.Feature.
+func (NegFeature) NewValidator() fuzz.FeatureValidator {
```
bowei: mixing pointer and non-pointer receiver defs is valid? (you should it all pointer type for consistency)

spencerhance:
@bowei Sounds good, will address in an upcoming PR

cmd/e2e-test/neg_test.go
spencerhance: https://github.com/kubernetes/ingress-gce/pull/701

pkg/e2e/helpers.go
spencerhance: https://github.com/kubernetes/ingress-gce/pull/701

pkg/fuzz/features/neg.go
spencerhance: https://github.com/kubernetes/ingress-gce/pull/701

PR #687, opened 2019-03-14
Fix supporting secret-based and pre-shared certs at the same time

michallowicki:
Secret-based certs were skipped during loadbalancer runtime info update when pre-shared certs were specified.

Changed to always include secret-based certs and added a test to document this change.

This is a followup to #641

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -556,19 +556,15 @@ func (lbc *LoadBalancerController) toRuntimeInfo(ing *extensions.Ingress, urlMap
 	var tls []*loadbalancers.TLSCerts
 
 	annotations := annotations.FromIngress(ing)
-	// Load the TLS cert from the API Spec if it is not specified in the annotation.
-	// TODO: enforce this with validation.
-	if annotations.UseNamedTLS() == "" {
-		tls, err = lbc.tlsLoader.Load(ing)
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				// TODO: this path should be removed when external certificate managers migrate to a better solution.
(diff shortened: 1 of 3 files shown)


pkg/controller/controller_test.go
```diff
@@ -350,3 +352,36 @@ func TestEnsureMCIngress(t *testing.T) {
...
 }
+
+// TestToRuntimeInfoCerts asserts that both pre-shared and secret-based certs
+// are included in the RuntimeInfo.
+func TestToRuntimeInfoCerts(t *testing.T) {
+	lbc := newLoadBalancerController()
+	tlsCerts := []*loadbalancers.TLSCerts{&loadbalancers.TLSCerts{Key: "key", Cert: "cert", Name: "tlsCert"}}
+	fakeLoader := &tls.FakeTLSSecretLoader{FakeCerts: map[string]*loadbalancers.TLSCerts{"tlsCert": tlsCerts[0]}}
+	lbc.tlsLoader = fakeLoader
+	tlsName := "preSharedCert"
```
rramkumar1: Nit: call this presharedCertName

pkg/controller/controller.go
```diff
@@ -556,19 +556,16 @@ func (lbc *LoadBalancerController) toRuntimeInfo(ing *extensions.Ingress, urlMap
 	var tls []*loadbalancers.TLSCerts
 
 	annotations := annotations.FromIngress(ing)
-	// Load the TLS cert from the API Spec if it is not specified in the annotation.
 	// TODO: enforce this with validation.
```
rramkumar1: Remove this TODO. I don't think its relevant anymore.

pkg/controller/controller_test.go
michallowicki: Done.

pkg/controller/controller.go
michallowicki: Done.

PR #688, opened 2019-03-14
fix a bug where ingress GC will remove wanted LB if the namespace+nam…

freehan:
fix a bug where ingress GC will remove wanted LB if the namespace+name length is between 44-47 chars Introduced by this commit: https://github.com/kubernetes/ingress-gce/commit/c74db67f12f55c1f9a16d9cc63fcca2a486fb353#diff-b1f4c09e8d55ceb3fe65799b21e57d43

### DIFF
--- a/pkg/loadbalancers/l7s_test.go
+++ b/pkg/loadbalancers/l7s_test.go
@@ -1,0 +1,350 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package loadbalancers
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/context"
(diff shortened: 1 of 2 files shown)


pkg/loadbalancers/l7s_test.go
```diff
@@ -0,0 +1,278 @@
...
+		generateKey(longName[:10], longName[:10]),
+		generateKey(longName[:20], longName[:20]),
+	}
+
+	for _, key := range otherKeys {
+		createFakeLoadbalancer(cloud, otherNamer, key)
+	}
+
+	for _, tc := range testCases {
+		for _, key := range tc.existingLBs {
```
bowei:
```
tc := tc
t.Run(tc.desc, func(t*testing.T){
 ...
})
```

pkg/loadbalancers/l7s_test.go
```diff
@@ -0,0 +1,278 @@
...
+			t.Errorf("For case %q, do not expect err: %v", tc.desc, err)
+		}
+		urlMaps, _ := l7sPool.cloud.ListURLMaps()
+		if len(urlMaps) != 1+numOfExtraUrlMap {
+			t.Errorf("For case %q, expect %d urlmaps, but got %d.", tc.desc, 1+numOfExtraUrlMap, len(urlMaps))
+		}
+		removeFakeLoadBalancer(l7sPool.cloud, namer, tc.key)
+	}
+}
+
```
bowei: can we add a test like this

const leakLimit = ...
for i := 2; i<200; i++ {
  ns, name := "n", key(i)
  if 1+i <= leakLimit {
    // check LB still exists
  } else {
    // check that the LB gets leaked.
  }

 ns, name =  key(i), n
  if 1+i <= leakLimit {
    // check LB still exists
  } else {
    // check that the LB gets leaked.
  }
}

basically go through every length to make sure there are no more weird edge cases

freehan:
Ready for another round

pkg/loadbalancers/l7s_test.go
```diff
@@ -0,0 +1,339 @@
...
+func TestGCToLeakLB(t *testing.T) {
+	t.Parallel()
```
bowei: put a comment here saying this should not leak at all, but verfies existing behavior

freehan:
Pushed

PR #696, opened 2019-03-19
e2e StatusManager init() handle existing config map

spencerhance:
If the status manager does not shut down correctly, the config map will still exist the next time a test starts and prevent the test from running until you delete it via kubectl. This deletes and replaces the config map if it already exists.

### DIFF
--- a/pkg/e2e/status.go
+++ b/pkg/e2e/status.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -91,6 +92,12 @@ func NewStatusManager(f *Framework) *StatusManager {
 
 func (sm *StatusManager) init() error {
 	var err error
+
+	// Optimistically delete a leftover config map
+	if err := sm.f.Clientset.Core().ConfigMaps("default").Delete(sm.cm.Name, &metav1.DeleteOptions{}); !errors.IsNotFound(err) {
+		klog.V(3).Infof("Error deleting config map %s: %v", sm.cm.Name, err)
+	}
+
 	sm.cm, err = sm.f.Clientset.Core().ConfigMaps("default").Create(sm.cm)
 	if err != nil {
 		return fmt.Errorf("error creating ConfigMap: %v", err)


pkg/e2e/status.go
```diff
@@ -91,9 +91,22 @@ func NewStatusManager(f *Framework) *StatusManager {
 
 func (sm *StatusManager) init() error {
 	var err error
-	sm.cm, err = sm.f.Clientset.Core().ConfigMaps("default").Create(sm.cm)
-	if err != nil {
-		return fmt.Errorf("error creating ConfigMap: %v", err)
+
+	// Check for existing config map with the same name.  If it exists, we update it
```
rramkumar1: I think it would actually be better to just delete and recreate. I worry that updating it would cause some issues if there is some bad state that was left over.

spencerhance: Sounds good, I just updated it

pkg/e2e/status.go
```diff
@@ -91,6 +91,18 @@ func NewStatusManager(f *Framework) *StatusManager {
 
 func (sm *StatusManager) init() error {
 	var err error
+
+	// Check for existing config map with the same name.  If it exists, we delete it and replace it
+	cm, err := sm.f.Clientset.Core().ConfigMaps("default").Get(sm.cm.Name, metav1.GetOptions{})
+	if cm != nil && err == nil {
```
rramkumar1: Can we log if this Get() call failed?

spencerhance: Sure! do you mean just log if there's any error?  Get() returns an error if it can't find the cm, so it might be noisy, but it shouldn't be a big deal.

rramkumar1: Yea, I think putting it at V(3) is fine

rramkumar1: I would structure this like:

cm, err := sm.f.Clientset.Core()...
if err != nil {
  klog.V(2).Infof("Error looking up config map %s: ", sm.cm.Name, err)
}
// Proceed with deletion. 


I think it's safe to assume that if err == nil, then cm is != nil so we don't need to have a check for it.

spencerhance: ah okay.  Doesn't that mean that the deletion code will always be hit?  So init() would return an error anytime the cm didn't already exist.

rramkumar1: Good point but now that I think about it more, we don't even need to do a Get(). We can just to an optimistic Delete()  meaning that if we don't get an error, that means the CM existed previously and deletion was successful. If Delete() returned an error and that error was something other than a 404, we just log it. If it was a 404, we know it does not exist so we just ignore. The much simpler logic would then look like:

if err := sm.f.Clientset.Core().ConfigMaps("default").Delete(cm.Name, &metav1.DeleteOptions{}); !errors.IsNotFound(err) {
  // Log the error at V3
}

spencerhance: Ooh good call

rramkumar1:
@spencerhance Can you squash commits? Will merge this once that is done.

PR #698, opened 2019-03-20
In E2E runner script, exit with error code 2 if encountered error with bootstrapping

rramkumar1:
This makes it easier for an external framework to distinguish between an error with the bootstrapping or an error due to the test failing (test failures return error code 1)


ok09:
pull request....👤

PR #702, opened 2019-03-26
Add ClusterIP test case to NEG e2e tests

spencerhance:
Also:
- Removes the Protocol and ServicePort fields from EnsureEchoService() since using non-default values only makes the service invalid.
- Fix EnsureEchoService() to also update deployment scale

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -38,6 +38,7 @@ func TestNEG(t *testing.T) {
 	for _, tc := range []struct {
 		desc               string
 		annotations        annotations.NegAnnotation
+		svcType						 v1.ServiceType
 		negExpected        bool
 		numForwardingRules int
 		numBackendServices int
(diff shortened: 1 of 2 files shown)


cmd/e2e-test/neg_test.go
```diff
@@ -53,14 +55,23 @@ func TestNEG(t *testing.T) {
 			desc:               "Annotation with no NEG",
 			annotations:        annotations.NegAnnotation{Ingress: false},
 			negExpected:        false,
+			svcType:            v1.ServiceTypeNodePort,
+			numForwardingRules: 1,
+			numBackendServices: 1,
+		},
+		{
+			desc:               "No NEG created with ClusterIP Service",
```
freehan: Ingress can use ClusterIP service NEG enabled

cmd/e2e-test/neg_test.go
```diff
@@ -168,7 +179,7 @@ func TestNEGTransition(t *testing.T) {
 		} {
 			// First create the echo service, we will be adapting it throughout the basic tests
 			_, err := e2e.EnsureEchoService(s, "service-1", map[string]string{
-				annotations.NEGAnnotationKey: tc.annotations.String()}, v1.ProtocolTCP, 80, v1.ServiceTypeNodePort)
+				annotations.NEGAnnotationKey: tc.annotations.String()}, v1.ServiceTypeNodePort)
```
freehan: Just use the ClusterIP type.

freehan:
please rebase. LGTM

PR #705, opened 2019-03-29
Static IP E2E test

rramkumar1:
Porting from k/k. Note the PR to remove the tests in k/k is https://github.com/kubernetes/kubernetes/pull/75892

Ref: #667

### DIFF
--- a/cmd/e2e-test/basic_test.go
+++ b/cmd/e2e-test/basic_test.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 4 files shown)


cmd/e2e-test/basic_test.go
```diff
@@ -114,6 +115,54 @@ func TestBasic(t *testing.T) {
...
+
+		testIng, err = e2e.WaitForIngress(s, testIng, nil)
+		if err != nil {
+			t.Fatalf("e2e.WaitForIngress(s, %q) = _, %v; want _, nil", testIng.Name, err)
+		}
+		if len(testIng.Status.LoadBalancer.Ingress) < 1 {
+			t.Fatalf("Ingress does not have an IP: %+v", testIng.Status)
+		}
+
+		vip := testIng.Status.LoadBalancer.Ingress[0].IP
```
MrHohn: Maybe verify this IP is the same as the reserved one?

rramkumar1: Is it necessary? I think if somehow the vip was different the check for 200 would fail right since we use that vip to make the requests?

MrHohn: I was thinking it could be that ingress controller mistakenly creates a new IP (like the default behavior without static-ip annotation) instead of using the one given.

rramkumar1: Ah good point! Since the annotation is technically a "feature", I will add that validation logic to pkg/fuzz/features...Will ping here once done.

rramkumar1: @MrHohn Done, PTAL

MrHohn: Looks good, thanks!

PR #707, opened 2019-03-29
Make sure static-ip validator is not run if annotation is empty + clean up addresses after test

rramkumar1:
Followup to #705 

/hold until I verify the test also fully passes.


rramkumar1:
/hold cancel

Test passes.

PR #710, opened 2019-04-03
make NEGTransition test unflakey under load

freehan:
In NEGTransition test, for every test case, after applying the new setup, it waits for ingress to stablize by calling https://github.com/kubernetes/ingress-gce/blob/master/cmd/e2e-test/neg_test.go#L90

However, if ingress-gce controller is under load, it takes more time to configure the LB based on the new config. Then the old LB setup will stays the same. And `WaitForIngress` will not block because the old setup is already working.

This PR makes the NEG feature validator to validate the reverse case (IG backends). So that `WaitForIngress` will block when NEG is disabled and wait for the ingress-gce to adjust backend service to use IG

### DIFF
--- a/pkg/fuzz/features/neg.go
+++ b/pkg/fuzz/features/neg.go
@@ -23,6 +23,7 @@ package features
 import (
 	"context"
 	"fmt"
+	"k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/fuzz"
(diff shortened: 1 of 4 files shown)


rramkumar1:
@freehan Do you mind adding in the description the exact cause of the issue?

pkg/fuzz/features/neg.go
```diff
@@ -76,65 +77,74 @@ func (v *negValidator) CheckResponse(host, path string, resp *http.Response, bod
...
+}
 
-	if beService == nil {
-		return false, fmt.Errorf("No backend service returned for name %s", negName)
-	}
+func verifyIgBackend(ctx *context.Context, env fuzz.ValidatorEnv, bsName string) error {
+	return verifyBackend(ctx, env, bsName, "instanceGroup")
+}
 
+func verifyBackend(ctx *context.Context, env fuzz.ValidatorEnv, bsName string, backendKeyword string) error {
```
rramkumar1: No need to pass context in since we are just using context.Background everywhere. Just instantiate it inline in this function.

bowei: its' fine to pass in context, but it's strange to pass it as a pointer

freehan: Done

freehan:
Added description

PR #712, opened 2019-04-04
bump ingress setup timeout to 25 min

freehan:


rramkumar1:
@freehan Can you please add a short description?

PR #719, opened 2019-04-09
Updates to kubernetes 1.14.1 levels

liggitt:
Preparation for adding support for ingress in the networking.k8s.io API group

* updates to go1.12.3
* updates to kube 1.14.1
* switches to using github.com/GoogleCloudPlatform/k8s-cloud-provider


a-robinson:
@liggitt I assume you're capable of finding another reviewer if needed, so removing myself.

liggitt:
@a-robinson definitely, no problem

you might want to drop yourself from the OWNERS files to avoid getting auto-tagged in the futre

PR #722, opened 2019-04-12
check if at least one URLmap is targetting the backend service in NEG feature validator

freehan:
Modify the neg feature validator to check if at least one URLmap is targetting the NEG/IG backend service. This is to guarantee that the correct backend service is fully configured with URLMap. This prevents flaky test when new Backend service is created but the old backend service is still hooked with existing URLmap. Hence making test cases validating white box configurations to fail every now and then. 

@spencerhance

### DIFF
--- a/pkg/fuzz/features/neg.go
+++ b/pkg/fuzz/features/neg.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/api/extensions/v1beta1"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/utils"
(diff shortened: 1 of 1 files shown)


pkg/fuzz/features/neg.go
```diff
@@ -149,5 +150,25 @@ func verifyBackend(env fuzz.ValidatorEnv, bsName string, backendKeyword string)
 			return fmt.Errorf("backend group %q of backend service %q does not contain keyword %q", be.Group, bsName, backendKeyword)
 		}
 	}
-	return nil
+
+	// Examine if any url map is targeting the backend service
```
rramkumar1: You plumbed in the namer right? Can't you just do a Get() rather than a list?

freehan: hmmmm yeah.

freehan: Done

PR #726, opened 2019-04-15
move NegStatus from pkg/neg to pkg/annotations

freehan:
This aims to make it easier to vendor pkg/annotations from a different repo

### DIFF
--- a/pkg/neg/types/types.go
+++ b/pkg/neg/types/types.go
@@ -18,7 +18,7 @@ package types
 
 import (
 	"fmt"
-	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/ingress-gce/pkg/annotations"
 	"reflect"
 	"strconv"
 )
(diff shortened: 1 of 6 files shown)


freehan:
cc @spencerhance @cadmuxe

pkg/neg/types/types.go
```diff
@@ -80,39 +79,10 @@ func (p1 PortInfoMap) Difference(p2 PortInfoMap) PortInfoMap {
 	return result
 }
 
-func (p1 PortInfoMap) ToPortNegMap() PortNegMap {
-	ret := PortNegMap{}
+func (p1 PortInfoMap) ToPortNegMap() map[string]string {
```
MrHohn: Why not using `annotations.PortNegMap` here?

freehan: I was trying to trim dependency of pkg/neg. But seems not possible. will revert it back

freehan:
Updated!

PR #727, opened 2019-04-16
Increase number of max goroutines for e2e test from 10 to 100

rramkumar1:
There is no reason not to increase this since goroutines are cheap. Plus, as we add more test cases, we don't want the corresponding goroutine to be completely blocked from starting. This might cause the overall test to take longer than it really should. 

/assign @freehan


freehan:
/lgtm

Not sure if this will blow up GCE. But let us see.

PR #729, opened 2019-04-16
Generate composite types

spencerhance:
This PR adds generated composite types for many of the services from k8s-cloud-provider using the API spec json.  This auto-generates the existing BackendService composite type as well.

The next step will to be to remove the first cloud-provider layer so that all types can be easily generated.  That will also depend on: https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/9


pkg/backends/backends.go
```diff
@@ -74,7 +74,7 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 		HealthChecks: []string{hcLink},
 	}
 	ensureDescription(be, &sp)
-	if err := composite.CreateBackendService(be, b.cloud); err != nil {
+	if err := composite.CreateBackendService(be, b.cloud, meta.GlobalKey(be.Name)); err != nil {
```
rramkumar1: Because we wan't to silo your changes for L7 ILB as much as possible to keep the code base stable, I would suggest not introducing the key parameter in this PR. That way these function calls do not change and there is no real chance for a bug to be introduced.

I would leave a TODO in the composite generator code that states this should be added in a follow up.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Generator for GCE compute wrapper code. You must regenerate the code after
+// modifying this file:
+//
+//   $ go run gen/main.go
```
rramkumar1: It will obviously be really simple wrapper, but can you add a script to hack/ that generates the composites? Something like "gen_composite.sh" or something like that.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+	"text/template"
+	"time"
+	"unicode"
+)
+
+const (
+	gofmt = "gofmt"
+
+	// This assumes that alpha contains a superset of all struct fields
+	apiFilePath = "../../vendor/google.golang.org/api/compute/v0.alpha/compute-api.json"
```
rramkumar1: This should be okay for now I think. Once we migrate to go modules, its possible we may have to do something different. My understanding of go modules is that they remove the need for vendor/ but you can still use it if you want. Anyways, we can cross that bridge when we get there.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+	"ForwardingRule":   "ForwardingRules",
+	"HttpHealthCheck":  "HttpHealthChecks",
+	"HttpsHealthCheck": "HttpsHealthChecks",
+	"UrlMap":           "UrlMaps",
+	"TargetHttpProxy":  "TargetHttpProxies",
+	"TargetHttpsProxy": "TargetHttpsProxies",
+}
+
+// TODO: (shance) Replace this with data gathered from meta.AllServices
+var NoUpdate = sets.NewString(
```
rramkumar1: This means types that do not have an Update() method right? Can you make that a bit clearer in the comment?

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+	if field["type"] == "string" && propType != "string" {
+		override = true
+	}
+
+	return propType, typesQueue, override, err
+}
+
+func genHeader(wr io.Writer) {
+	const text = `/*
+Copyright {{.Year}} Google LLC
```
rramkumar1: Can you preserve the spacing in the header? For example, there is typically a space after the first line. Same comment for the test file header as well.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+	for len(typesQueue) > 0 {
+		typeName := typesQueue[0]
+		typesQueue = typesQueue[1:]
+
+		if completed.Has(typeName) {
+			continue
+		}
+		completed.Insert(typeName)
+
+		fields, ok := result["schemas"].(map[string]interface{})[typeName].(map[string]interface{})["properties"].(map[string]interface{})
```
rramkumar1: Break this into multiple lines for readability.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+
+	byteValue, err := ioutil.ReadAll(apiFile)
+	if err != nil {
+		panic(err)
+	}
+
+	var result map[string]interface{}
+	json.Unmarshal([]byte(byteValue), &result)
+
+	typesQueue := []string{}
```
rramkumar1: Comment for both typesQueue and completed.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,669 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling {{$type.Name}} to JSON: %v", err)
+	}
+	{{$version | ToLower}} := &compute{{$extension}}.{{$type.Name}}{}
+	err = json.Unmarshal(bytes, {{$lower}})
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling {{$type.Name}} JSON to compute {{$lower}} type: %v", err)
+	}
+
+	{{- if eq $type.Name "BackendService"}}
```
rramkumar1: We will have to think of a better solution for this. It may be fine to just add all fields to ForceSendFields but for now I think this is fine.

pkg/composite/gen/main.go
rramkumar1: Actually, you can even just add a line to update_codegen.sh

pkg/backends/backends.go
spencerhance: Done.  Also good call, the only major change this PR is introducing is bypassing the first cloud-provider layer and calling the CRUD functions directly 
(e.g `cloud.Compute().AlphaBackendServices().Insert(meta.GlobalKey())` instead of `cloud.CreateAlphaGlobalBackendService()`.)  

With this change, we would be losing metrics until some replacement is implemented.

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: hmm good point.  Will look into modules.

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Agreed, do you remember why those fields for BackendService were added originally?

pkg/backends/backends.go
spencerhance: The reason for that was to try to make it simpler to use global and regional resources with less switches (and easier code generation logic), but I'm definitely open to suggestions.

spencerhance: The only other change I see in the diff is making the TestToX() functions simpler since I don't use known Alpha/Beta fields anymore.  It's possible to add those in by doing a diff in the generate code but it would definitely add a bunch more complexity.

spencerhance:
@rramkumar1 Thanks! There's also an additional approach which uses the parsing code found in k8s-cloud-provider, but that would have to be heavily modified or copy-pasted into ingress-gce which seemed worse than not having any dependencies.

Also ready for another round of review

pkg/composite/gen/main.go
rramkumar1: They were needed because of some subtle behavior with how the GCE API's work. Specifically, if you there is a boolean field (like "IncludeHost") and you are trying to set the field to false, GCE will not recognize that false setting unless you add the field to the ForceSendFields list. 

From GCE docs: "By default, fields with empty values are omitted from API requests." (https://godoc.org/google.golang.org/api/compute/v1).

I'm trying to remember why though I needed to add this here rather than have callers set those fields...

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,681 @@
...
+	}
+
+	tmpl := template.Must(template.New("funcs").Funcs(funcMap).Parse(text))
+	if err := tmpl.Execute(wr, data); err != nil {
+		panic(err)
+	}
+}
+
+// genTests() generates all of the tests
+// TODO: (shance) figure out a better way to test the toGA(), toAlpha(), and toBeta() functions
```
rramkumar1: I honestly think testing toGA(), toAlpha() and toBeta() can be done similar to how toBackendService() is done. I think all we are really testing is that we got the types right in the implementation. It will be hard to actually test individual fields since those will be very fluid.

Is that doable?

pkg/composite/composite.go
```diff
@@ -30,100 +30,30 @@ import (
...
-
-func CreateBackendService(be *BackendService, cloud *gce.Cloud) error {
-	switch be.Version {
-	case meta.VersionAlpha:
-		alpha, err := be.toAlpha()
-		if err != nil {
-			return err
-		}
-		klog.V(3).Infof("Creating alpha backend service %v", alpha.Name)
-		return cloud.CreateAlphaGlobalBackendService(alpha)
```
rramkumar1: My feeling is that we should not remove the usage of the wrapper until we have a solution for keeping the metrics around.

spencerhance: So the two main reasons for removing the wrapper are that it doesn't follow a consistent formula like the generated wrapper, and it's in k/k so it would be potentially harder for me to change it for the PR.  https://github.com/kubernetes/kubernetes/tree/master/pkg/cloudprovider/providers/gce.

That being said, I could just limit this PR to doing a drop-in replacement of BackendService and leave the other composite types for another PR once I have a replacement for metrics.

pkg/composite/composite_test.go
```diff
@@ -21,52 +20,109 @@ import (
...
+	if err := typeEquality(compositeType, alphaType); err != nil {
+		t.Fatal(err)
+	}
+}
 func TestBackendService(t *testing.T) {
 	// Use reflection to verify that our composite type contains all the
 	// same fields as the alpha type.
-	compositeType := reflect.TypeOf(BackendService{})
-	alphaType := reflect.TypeOf(computealpha.BackendService{})
+	compositeType := reflect.TypeOf(ForwardingRule{})
```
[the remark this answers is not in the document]
spencerhance: Good catch!  I left that in the template by mistake

pkg/composite/composite.go
rramkumar1: I don't mind that idea. That way, we don't have to block this necessarily on the metrics stuff.

pkg/composite/composite_test.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,681 @@
+/*
+Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/composite_test.go
```diff
@@ -1,18 +1,17 @@
 /*
-Copyright 2018 The Kubernetes Authors.
-
+Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done, not sure if it's exactly what you meant.

pkg/composite/composite.go
spencerhance: Done

spencerhance: Done

spencerhance:
@rramkumar1 Ready for another round

pkg/composite/composite.go
spencerhance: side note, do you know off hand if anything uses those metrics?

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,692 @@
...
+)
+
+var Versions = map[string]string{
+	"Alpha": "alpha",
+	"Beta":  "beta",
+	"GA":    "",
+}
+
+// ApiService is a struct to hold all of the relevant data for generating a composite service
+type ApiService struct {
```
rramkumar1: Can you add comments for each field?

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,692 @@
...
+
+	// This assumes that alpha contains a superset of all struct fields
+	apiFilePath = "./vendor/google.golang.org/api/compute/v0.alpha/compute-api.json"
+)
+
+// MainServices describes all of the API types that we want to define all the helper functions for
+// The other types that are discovered as dependencies will simply be wrapped with a composite struct
+// The format of the map is ServiceName -> k8s-cloud-provider wrapper name
+// TODO: (shance) Add the commented services and remove dependency on first cloud-provider layer
+var MainServices = map[string]string{
```
rramkumar1: Can you extract this and all the other types and methods below into a separate package?

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,692 @@
...
+		"ToLower": strings.ToLower,
+	}
+
+	tmpl := template.Must(template.New("tests").Funcs(funcMap).Parse(text))
+	if err := tmpl.Execute(wr, data); err != nil {
+		panic(err)
+	}
+}
+
+func init() {
```
rramkumar1: This init can also be moved to a separate package when you move the other stuff.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,692 @@
...
+	if err := tmpl.Execute(wr, values); err != nil {
+		panic(err)
+	}
+
+	fmt.Fprintf(wr, "\n\n")
+}
+
+func genTestHeader(wr io.Writer) {
+	const text = `/*
+Copyright {{.Year}} Google LLC
```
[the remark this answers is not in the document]
spencerhance: shoot, good catch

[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: I'm starting to wonder if it would make sense for all of this stuff to be moved into k8s-cloud-provider :upside_down_face: .  Maybe not at first though.

rramkumar1: Yeah I've thought about that. Reason I hesitate right now is that its pretty specific to us. If someone else finds a need for this, then definitely it should go there.

spencerhance:
@rramkumar1 Ready for another review

pkg/composite/gen/main.go
spencerhance: makes sense.  It would allow us to refactor the parsing code a bit too which would be nice eventually.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
...
+		All []compositemeta.ApiService
+	}{compositemeta.AllApiServices}
+
+	tmpl := template.Must(template.New("types").Parse(text))
+	if err := tmpl.Execute(wr, data); err != nil {
+		panic(err)
+	}
+}
+
+// genFuncs() generates all of the struct methods
```
rramkumar1: Nit: genFuncs() generates helper methods attached to composite structs.

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	tmpl := template.Must(template.New("funcs").Funcs(funcMap).Parse(text))
+	if err := tmpl.Execute(wr, data); err != nil {
+		panic(err)
+	}
+}
+
+// genTests() generates all of the tests
+// TODO: (shance) figure out a better way to test the toGA(), toAlpha(), and toBeta() functions
```
rramkumar1: Can remove TODO since you addressed it I think.

pkg/composite/meta/meta.go
```diff
@@ -0,0 +1,247 @@
...
+	"TargetHttpsProxy",
+)
+
+var Versions = map[string]string{
+	"Alpha": "alpha",
+	"Beta":  "beta",
+	"GA":    "",
+}
+
+// ApiService is a struct to hold all of the relevant data for generating a composite service
```
rramkumar1: Nit: ApiService holds relevant data for generating a composite type + helper methods for a single API service.

pkg/composite/meta/meta.go
```diff
@@ -0,0 +1,247 @@
...
+	byteValue, err := ioutil.ReadAll(apiFile)
+	if err != nil {
+		panic(err)
+	}
+
+	var result map[string]interface{}
+	json.Unmarshal([]byte(byteValue), &result)
+
+	// Queue of ApiService names for BFS
+	typesQueue := []string{}
```
rramkumar1: Remove the extra spaces between here and line 136.

spencerhance: Done

pkg/composite/meta/meta.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
...
+		"Year": fmt.Sprintf("%v", time.Now().Year()),
+	}
+	if err := tmpl.Execute(wr, values); err != nil {
+		panic(err)
+	}
+
+	fmt.Fprintf(wr, "\n\n")
+}
+
+// genTypes() generates all of the struct definitions
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
...
+	"text/template"
+	"time"
+)
+
+const (
+	gofmt = "gofmt"
+)
+
+// gofmtContent runs "gofmt" on the given contents.
+// Duplicate of the function in k8s-cloud-provider
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
...
+//   $ ./hack/update_codegen.sh
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	compositemeta "k8s.io/ingress-gce/pkg/composite/meta"
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -0,0 +1,476 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
spencerhance: Done

Issue #493, opened 2018-09-25
unable to activate gzip


scaccogatto:
With my system:

```
| (Ingress) 
| Google Cloud Load Balancer 
|--------> Website Backend
|--------> Website Backend
|--------> Website Backend
|--------> Website Backend
```
Ingress config: https://gist.github.com/scaccogatto/8796908c22cc6437fc645f3fba587edd
Backend nginx config: https://github.com/scaccogatto/vue-nginx/blob/master/docker/nginx.conf

Following:
https://cloud.google.com/load-balancing/docs/https/setting-up-https#compression-not-working

As you can see from my configs gzip directives are here, also **I found that my Request Headers contains accept-encoding: gzip, deflate, br**.

Unfortunately when I run Chrome's audit it says that I should enable text compression even if most of online gzip checkers say that gzip is active (but I don't trust them)

Also I found that my **Response Headers does not contain**: `content-encoding: gzip` so I assume gzip is not working correctly.

Searching on the net I found this: https://stackoverflow.com/a/48084669/10412176 (that actually says the same things that's on the docs) and I am starting to think than I am changing the wrong configuration and the docs reefer to the Ingress configuration (that is not reachable as fas as I know)

I don't really know if it is a bug or not so I'm asking here.

rramkumar1:
@scaccogatto This looks like an Nginx specific issue. Going to close this.

Feel free to re-open if you think this is an Ingress-GCE issue.

/close

scaccogatto:
Sorry but this is not a Nginx issue. The fact that you cannot enable gzip with Ingress-GCE is a Ingress-GCE issue not an nginx one.
I just switched to Ingress-Nginx for my production env and it was super easy to enable gzip but the problem remains, gzip is a widely used feature and I feel like not implementing it into Ingress-GCE makes it unusable for web oriented projects.

So this confirm this is actually not a bug but a missing feature I suppose.

rramkumar1:
@scaccogatto 

You said: "Response Headers does not contain: content-encoding: gzip"

Unless I am misunderstanding your setup, doesn't this mean your application needs to return the proper headers?

bowei:
Can you post your HTTP request and response headers?

scaccogatto:
> @scaccogatto
> 
> You said: "Response Headers does not contain: content-encoding: gzip"
> 
> Unless I am misunderstanding your setup, doesn't this mean your application needs to return the proper headers?

[...]

PR #730, opened 2019-04-16
Add support for response compression in echoserver

rramkumar1:
This is nice to have but will also help in debugging #493 

Verified with following request:

```
$ curl -i -H "Accept-Encoding: gzip" localhost:8080

HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Type: application/json
Date: Tue, 16 Apr 2019 21:33:33 GMT
Content-Length: 10

Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.
```

### DIFF
--- a/cmd/echo/app/handlers.go
+++ b/cmd/echo/app/handlers.go
@@ -17,10 +17,13 @@ limitations under the License.
 package app
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"k8s.io/ingress-gce/pkg/version"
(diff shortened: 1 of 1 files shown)


cmd/echo/app/handlers.go
```diff
@@ -114,11 +117,32 @@ func echo(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	w.Write(processedData)
 	klog.V(3).Infof("echo: %v, %v, %v", time.Now(), r.UserAgent(), r.RemoteAddr)
 }
 
+// process does additional processing on response data if necessary.
+func process(w http.ResponseWriter, r *http.Request, b []byte) ([]byte, error) {
+	if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
```
freehan: Just curious if there is any chance it will be double compressed?

rramkumar1: What do you mean by double compressed?

freehan: If the incoming data is already compressed? Will this takes the data and compress it again?

PR #735, opened 2019-04-20
Switch to go modules

spencerhance:
- Updates vendor
- Pins all dependencies to their exact version


rramkumar1:
@spencerhance My understanding was that for us to transition to modules, all our dependencies need to be on modules as well? Is that true?

spencerhance:
@rramkumar1 That i'm not sure.  It seemed to me that it would handle non-module dependencies as before but I'm still investigating.

rramkumar1:
@spencerhance Do you mind adding some docs under docs/dev/ for how to update dependencies?

bowei:
can we try limiting this update to just moving to go modules? this change looks like it pulls a bunch of other stuff in, which makes it harder to review

spencerhance:
@bowei I ran into some complications without running a `dep ensure -update` but I'll see if I can remove it.

spencerhance:
@bowei I got rid of the composite changes and associated api version update so all changes should be localized to modules and vendor/ now

PR #734, opened 2019-04-19
Refactor Transaction NEG syncer to use strong type instead of string

freehan:
- Introduce a NetworkEndpoint Type to represent the meta information contained in NE.
- Introduce NetworkEndpointSet Type to support fast Get, Insert and DIfference calculation
- Adjust Transaction table to use NetworkEndpoint Type as key
- Adjust Transaction Syncer to use NetworkEndpoint and NetworkEndpointSet Type for internal plumbing. 
- Completely remove use of sets.String type for endpoint calculation.

PR #736, opened 2019-04-20
Use transactionTable via Interface

freehan:
Follow up of: https://github.com/kubernetes/ingress-gce/pull/734

Only the last commit is more stuff than #734

### DIFF
--- a/pkg/neg/syncers/transaction_table.go
+++ b/pkg/neg/syncers/transaction_table.go
@@ -16,7 +16,17 @@ limitations under the License.
 
 package syncers
 
-import negtypes "k8s.io/ingress-gce/pkg/neg/types"
+import (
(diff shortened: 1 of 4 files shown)


pkg/neg/syncers/transaction_table.go
```diff
@@ -16,7 +16,17 @@ limitations under the License.
 
 package syncers
 
-import negtypes "k8s.io/ingress-gce/pkg/neg/types"
+import (
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"sync"
+)
+
+type networkEndpointTransactionTable interface {
```
MrHohn: What is the consideration of creating an interface v.s. passing the pointer directly?

freehan: I do not want to handle nil pointers. Just a personal preference.

PR #745, opened 2019-05-02
Adding docker configuration file for the 404-server-with-metrics and …

vbannai:
…also changing the Makefile to push by default to the gcr.io staging repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,8 @@ CONTAINER_BINARIES ?= \
 # Latest commit hash for current branch.
 GIT_COMMIT := $(shell git rev-parse HEAD)
 
-# Registry to push to.
-REGISTRY ?= gcr.io/google_containers
+# Push to the staging registry.
+REGISTRY ?= gcr.io/k8s-image-staging
 
 ARCH ?= amd64
 ALL_ARCH := amd64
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,8 +33,8 @@ CONTAINER_BINARIES ?= \
 # Latest commit hash for current branch.
 GIT_COMMIT := $(shell git rev-parse HEAD)
 
-# Registry to push to.
-REGISTRY ?= gcr.io/google_containers
+# Push to the staging registry.
+REGISTRY ?= gcr.io/k8s-image-staging
```
MrHohn: Should we use `staging-k8s.gcr.io`? Ref https://github.com/kubernetes/dns/blob/master/Makefile.

vbannai: I am all for consistency. There is no difference between gcr.io/k8s-image-staging and staging-k8s.gcr.io as **staging-k8s.gcr.io** points to **gcr.io/k8s-image-staging**. 
I can change it to staging-k8s.gcr.io if you insist. Let me know.

MrHohn: Yeah don't have strong opinion --- I don't recall the reason for using alias either.

PR #748, opened 2019-05-03
readiness reflector

freehan:
- initial commit of readiness reflector
  - the readiness reflector will check if the pods has NEG readiness gate.
  - if so, then will poll NEG health status and patch the pod NEG readiness condition when the endpoint became ready. 
- Refactoring: 
  - modify common NEG types to accomendate readiness gate
  - move NegSyncerKey to pkg/neg/types
- Adapt NEG controller for readiness reflector:
  - include non-terminating pods to add in NEG
    - This will make sure if a pod is not ready due to the NEG readiness gate will be added into NEG
  - adapt NEG controller and Syncer Manager to take readiness gate into account. 
  - adapt transaction NEG syncer to feedback into readiness reflector

### DIFF
--- a/pkg/neg/readiness/interface.go
+++ b/pkg/neg/readiness/interface.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package readiness
+
+import (
+	"k8s.io/api/core/v1"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+)
+
+// Reflector defines the interaction between readiness reflector and other NEG controller components
+type Reflector interface {
+	// Run starts the reflector.
+	// Closing stopCh will signal the reflector to stop running.
+	Run(stopCh <-chan struct{})
+	// SyncPod signals the reflector to evaluate pod and patch pod status if needed.
+	SyncPod(pod *v1.Pod)
+	// CommitPods signals the reflector that pods has been added to and NEG it is time to poll the NEG health status
+	// syncerKey is the key to uniquely identify the NEG syncer
+	// negName is the name of the network endpoint group (NEG) in the zone (e.g. k8s1-1234567-namespace-name-80-1234567)
+	// zone is the corresponding zone of the NEG resource (e.g. us-central1-b)
+	// endpointMap contains mapping from all network endpoints to pods which have been added into the NEG
+	CommitPods(syncerKey negtypes.NegSyncerKey, negName string, zone string, endpointMap negtypes.EndpointPodMap)
+}
+
+// NegLookup defines an interface for looking up pod membership.
+type NegLookup interface {
+	// ReadinessGateEnabledNegs returns a list of NEGs which has readiness gate enabled for the input pod's namespace and labels.
+	ReadinessGateEnabledNegs(namespace string, labels map[string]string) []string
+	// ReadinessGateEnabled returns true if the NEG requires readiness feedback
+	ReadinessGateEnabled(syncerKey negtypes.NegSyncerKey) bool
+}
+
+type NoopReflector struct{}
+
+func (*NoopReflector) Run(<-chan struct{}) {}
+
+func (*NoopReflector) SyncPod(*v1.Pod) {}
+
+func (*NoopReflector) CommitPods(negtypes.NegSyncerKey, string, string, negtypes.EndpointPodMap) {}
--- a/pkg/neg/readiness/poller.go
+++ b/pkg/neg/readiness/poller.go
@@ -1,0 +1,236 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package readiness
+
+import (
+	"fmt"
+	computebeta "google.golang.org/api/compute/v0.beta"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/cache"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/klog"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+const (
+	healthyState = "HEALTHY"
+)
+
+// negMeta references a GCE NEG resource
+type negMeta struct {
+	SyncerKey negtypes.NegSyncerKey
+	// Name is the name of the NEG
+	Name string
+	// Zone is the zone of the NEG resource
+	Zone string
+}
+
+func (n negMeta) String() string {
+	return fmt.Sprintf("%s-%s-%s", n.SyncerKey.String(), n.Name, n.Zone)
+}
+
+// podStatusPatcher interface allows patching pod status
+type podStatusPatcher interface {
+	// syncPod patches the neg condition in the pod status to be True.
+	// key is the key to the pod. It is the namespaced name in the format of "namespace/name"
+	// negName is the name of the NEG resource
+	syncPod(key, negName string) error
+}
+
+// pollTarget is the target for polling
+type pollTarget struct {
+	// endpointMap maps network endpoint to namespaced name of pod
+	endpointMap negtypes.EndpointPodMap
+	// polling indicates if the NEG is being polled
+	polling bool
+}
+
+// poller tracks the negs and corresponding targets needed to be polled.
+type poller struct {
+	lock sync.Mutex
+	// pollMap contains negs and corresponding targets needed to be polled.
+	// all operations(read, write) to the pollMap are lock protected.
+	pollMap map[negMeta]*pollTarget
+
+	podLister cache.Indexer
+	lookup    NegLookup
+	patcher   podStatusPatcher
+	negCloud  negtypes.NetworkEndpointGroupCloud
+}
+
+func NewPoller(podLister cache.Indexer, lookup NegLookup, patcher podStatusPatcher, negCloud negtypes.NetworkEndpointGroupCloud) *poller {
+	return &poller{
+		pollMap:   make(map[negMeta]*pollTarget),
+		podLister: podLister,
+		lookup:    lookup,
+		patcher:   patcher,
+		negCloud:  negCloud,
+	}
+}
+
+// RegisterNegEndpoints registered the endpoints that needed to be poll for the NEG with lock
+func (p *poller) RegisterNegEndpoints(key negMeta, endpointMap negtypes.EndpointPodMap) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	p.registerNegEndpoints(key, endpointMap)
+}
+
+// registerNegEndpoints registered the endpoints that needed to be poll for the NEG
+// It returns false if there is no endpoints needed to be polled, returns true if otherwise.
+// Assumes p.lock is held when calling this method.
+func (p *poller) registerNegEndpoints(key negMeta, endpointMap negtypes.EndpointPodMap) bool {
+	endpointsToPoll := needToPoll(key.SyncerKey, endpointMap, p.lookup, p.podLister)
+	if len(endpointsToPoll) == 0 {
+		delete(p.pollMap, key)
+		return false
+	}
+
+	if v, ok := p.pollMap[key]; ok {
+		v.endpointMap = endpointsToPoll
+	} else {
+		p.pollMap[key] = &pollTarget{endpointMap: endpointsToPoll}
+	}
+	return true
+}
+
+// ScanForWork returns the list of NEGs that should be polled
+func (p *poller) ScanForWork() []negMeta {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	var ret []negMeta
+	for key, target := range p.pollMap {
+		if target.polling {
+			continue
+		}
+		if p.registerNegEndpoints(key, target.endpointMap) {
+			ret = append(ret, key)
+		}
+	}
+	return ret
+}
+
+// Poll polls a NEG and returns error plus whether retry is needed
+// This function is threadsafe.
+func (p *poller) Poll(key negMeta) (retry bool, err error) {
+	if !p.markPolling(key) {
+		klog.V(4).Infof("NEG %q in zone %q as is already being polled or no longer needed to be polled.", key.Name, key.Zone)
+		return true, nil
+	}
+	defer p.unMarkPolling(key)
+
+	// TODO(freehan): refactor errList from pkg/neg/syncers to be reused here
+	var errList []error
+	klog.V(2).Infof("polling NEG %q in zone %q", key.Name, key.Zone)
+	// TODO(freehan): filter the NEs that are in interest once the API supports it
+	res, err := p.negCloud.ListNetworkEndpoints(key.Name, key.Zone /*showHealthStatus*/, true)
+	if err != nil {
+		return true, err
+	}
+
+	// Traverse the response and check if the endpoints in interest are HEALTHY
+	func() {
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		var healthyCount int
+		for _, r := range res {
+			healthy, err := p.processHealthStatus(key, r)
+			if healthy && err == nil {
+				healthyCount++
+			}
+			if err != nil {
+				errList = append(errList, err)
+			}
+		}
+		if healthyCount != len(p.pollMap[key].endpointMap) {
+			retry = true
+		}
+	}()
+	return retry, utilerrors.NewAggregate(errList)
+}
+
+// processHealthStatus evaluates the health status of the input network endpoint.
+// Assumes p.lock is held when calling this method.
+func (p *poller) processHealthStatus(key negMeta, healthStatus *computebeta.NetworkEndpointWithHealthStatus) (healthy bool, err error) {
+	ne := negtypes.NetworkEndpoint{
+		IP:   healthStatus.NetworkEndpoint.IpAddress,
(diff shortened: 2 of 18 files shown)


pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,208 @@
...
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		healthyCount := 0
+		for _, r := range res {
+			podName, ok := p.getPod(key, negtypes.NetworkEndpoint{IP: r.NetworkEndpoint.IpAddress, Port: strconv.FormatInt(r.NetworkEndpoint.Port, 10)})
+			if ok {
+				for _, hs := range r.Healths {
+					if hs.BackendService != nil {
+						// This assumes the ingress backend service uses the NEG naming scheme
+						if strings.Contains(hs.BackendService.BackendService, key.Name) {
```
freehan: Currently I am only checking the health status from the backend service managed by Ingress. 

Other option is to loop thru all of them. If one is Healthy then ready. Or I can say all of them have to be Healthy. But this makes event reporting more complex. I need to say this endpoint in NEG is not ready in a BS.

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+		klog.V(4).Infof("NEG %q in zone %q as is already being polled or no longer needed to be polled.", key.Name, key.Zone)
+		return nil, true
+	}
+	defer p.unMarkPolling(key)
+
+	errList := []error{}
+
+	klog.V(2).Infof("Polling NEG %q in zone %q", key.Name, key.Zone)
+	// TODO(freehan): filter the NEs that are in interest.
+	res, err := p.negCloud.ListNetworkEndpoints(key.Name, key.Zone, true)
```
bowei: `.., /*showHealthStatus*/ true)`

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+		return err, true
+	}
+
+	retry := false
+	// Traverse the response and check if the endpoints in interest are HEALTHY
+	func() {
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		healthyCount := 0
+		for _, r := range res {
```
bowei: can we factor out logic so it's easier to look at
```
var healthy int
for _, neg := range res {
  healthy, err := p.syncPod(neg)
  if err == nil && healthy {
    healthy++
  }
  if err != nil {
    errs = append(errs, err)
  }
}

func (...) syncPod(...) { ...}
```

pkg/neg/readiness/reflector.go
```diff
@@ -0,0 +1,253 @@
...
+	"reflect"
+)
+
+const (
+	maxRetries        = 15
+	negReadyReason    = "LoadBalancerNegReady"
+	negNotReadyReason = "LoadBalancerNegNotReady"
+)
+
+// readinessReflector implements the PodReadinessReflector interface
```
bowei: implements => interface

readinessReflector implements PodReadinessReflector

also probably can delete "Pod" from PodReadinessReflector (redunant)

pkg/neg/readiness/reflector.go
```diff
@@ -0,0 +1,253 @@
...
+	podLister cache.Indexer
+	lookup    NegLookup
+
+	eventBroadcaster record.EventBroadcaster
+	eventRecorder    record.EventRecorder
+
+	queue workqueue.RateLimitingInterface
+}
+
+func NewReadienssReflector(ctx *context.ControllerContext, lookup NegLookup) Reflector {
```
bowei: spelling

bowei: prefer calling ControllerContext variables `cc` to avoid clash with `ctx` which is typically context.Context

pkg/neg/readiness/reflector.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	// This is to prevent if the pod got updated after being added to the queue
+	if !needToProcess(pod) {
+		return nil
+	}
+
+	klog.V(4).Infof("Syncing Pod %q", key)
+	expectedCondition := v1.PodCondition{Type: shared.NegReadinessGate}
+	message := ""
```
bowei: var message, reason string

pkg/neg/readiness/reflector.go
```diff
@@ -0,0 +1,253 @@
...
+		klog.Errorf("obj type is %T, expect *v1.Pod", obj)
+		return
+	}
+
+	if needToProcess(pod) {
+		r.queue.Add(key)
+	}
+}
+
+// CommitPods
```
bowei: comment

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if condition, ok := GetNegReadinessConditionStatus(pod); ok {
+		if condition.Status == v1.ConditionTrue {
+			negReady = true
+		}
+	}
+	return
+}
+
+func keyFunc(namespace, name string) string {
```
bowei: is this not a dupe of function already exists?

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == shared.NegReadinessGate {
+			return condition, true
+		}
+	}
+	return v1.PodCondition{}, false
+}
+
+// evalNegReadinessGate returns if the pod readiness gate includes the NEG readiness condition and the condition status is true
+func evalNegReadinessGate(pod *v1.Pod) (negReady bool, readinessGateExists bool) {
```
bowei: prefer to not use named returns. It's must clearer to just do:
```
var negReady, gateExists bool
...
return negReady, gateExists
```

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/neg/types/shared"
+	"k8s.io/klog"
+)
+
+// GetNEGReadinessStatus retrieves the status of the NEG readiness condition if it exists.
```
bowei: NEGReadinessConditionStatus return (cond, true) if neg condition exists, otherwise (_, false)

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return
+}
+
+func keyFunc(namespace, name string) string {
+	return fmt.Sprintf("%s/%s", namespace, name)
+}
+
+// getPodFromStore retrieves pod from podLister
```
bowei: what is meaning of bool

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		return nil, false, fmt.Errorf("Failed to convert obj type %T to *v1.Pod", obj)
+	}
+	return pod, true, nil
+}
+
+// SetNegReadinessConditionStatus sets the status of the NEG readiness condition
+func SetNegReadinessConditionStatus(pod *v1.Pod, condition v1.PodCondition) {
+	found := false
```
bowei: seems easier to do 
```
for i, c := pod.Status.Conditions {
  if c.Type == shared.NegReadinessGate {
    pod.Status.Conditions[i] = newCondition
    return
  }
}
pod.Status.Conditions = append(pod.Status.Conditions, newCondition)
```

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	} else {
+		pod.Status.Conditions = append(pod.Status.Conditions, condition)
+	}
+}
+
+// patchPodStatus patches pod status with given patchBytes
+func patchPodStatus(c clientset.Interface, namespace, name string, patchBytes []byte) (*v1.Pod, []byte, error) {
+	updatedPod, err := c.CoreV1().Pods(namespace).Patch(name, types.StrategicMergePatchType, patchBytes, "status")
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to patch status %q for pod %q/%q: %v", patchBytes, namespace, name, err)
```
bowei: why wrap the error here?

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return updatedPod, patchBytes, nil
+}
+
+// preparePatchBytesforPodStatus generates patch bytes based on the old and new pod status
+func preparePatchBytesforPodStatus(namespace, name string, oldPodStatus, newPodStatus v1.PodStatus) ([]byte, error) {
```
bowei: this seems super generic, can we make it
```
func prepareObjectPatch(old, new, obj interface{}) ([]byte, error) {
  // jsonMarshal stuff
  ...
}

func preparePatchForPodStatus(old, new v1.PodStatus) ([]byte, error) {
  return prepareObjectPatch(v1.Pod{Status:old}, v1.Pod{Status:new}, v1.Pod{}})
}

```

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return patchBytes, nil
+}
+
+// needToPoll filter out the network endpoint that needs to be polled based on the following conditions:
+// 1. neg syncer has readiness gate enabled
+// 2. the pod exists
+// 3. the pod has neg readiness gate
+// 4. the pod's neg readiness condition is not True
+func needToPoll(syncerKey negtypes.NegSyncerKey, endpointMap map[negtypes.NetworkEndpoint]types.NamespacedName, lookup NegLookup, podLister cache.Indexer) map[negtypes.NetworkEndpoint]types.NamespacedName {
```
bowei: can we make a type for map[negtypes.NetworkEndpoint]types.NamespacedName

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+	if lookup.ReadinessGateEnabled(syncerKey) {
+		filterEndpoint(endpointMap, podLister)
```
bowei: filterEndpoints

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+		return endpointMap
+	}
+	return map[negtypes.NetworkEndpoint]types.NamespacedName{}
```
bowei: return early, put at the top
if !lookup.ReadinessGateEnable() { return }

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// filterEndpoint will filter out the endpoints that does not need health polling from the input endpoint map
```
bowei: you should call this removeIrrelevantEndpoints, as filter usually means no side effects f(set) => newSet

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+func filterEndpoint(endpointMap map[negtypes.NetworkEndpoint]types.NamespacedName, podLister cache.Indexer) {
+	for endpoint, namespacedName := range endpointMap {
+		pod, exists, err := getPodFromStore(podLister, namespacedName.Namespace, namespacedName.Name)
+		if err != nil {
+			klog.Warningf("Failed to retrieve pod %q from store: %v", namespacedName.String(), err)
```
bowei: is this a warning or something that will happen from time to time?

pkg/neg/readiness/utils.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		delete(endpointMap, endpoint)
+	}
+}
+
+// needToProcess check if the pod needs to be processed by readiness reflector
+// If pod has neg readiness gate and its condition is False, then return true.
+func needToProcess(pod *v1.Pod) bool {
+	negConditionReady, readinessGateExists := evalNegReadinessGate(pod)
+	if !readinessGateExists {
```
bowei: return readinessGateExists && negConditionReady

push logging up one layer (don't log in functional predicate if possible, easy way to get tons of log spam)

pkg/neg/syncers/transaction.go
```diff
@@ -286,31 +298,43 @@ func (s *transactionSyncer) commitTransaction(err error, networkEndpointMap map[
 
 	for networkEndpoint := range networkEndpointMap {
 		entry, ok := s.transactions.Get(networkEndpoint)
+		// clear transaction
```
bowei: TODO: review this closer

pkg/neg/syncers/utils.go
```diff
@@ -265,3 +277,31 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet) (map[negtypes.Netw
 	}
 	return endpointBatch, nil
 }
+
+func keyFunc(namespace, name string) string {
```
bowei: duplicated code

pkg/neg/readiness/utils.go
freehan: Yes. But not shared though.

pkg/neg/readiness/utils.go
freehan: This only happen if the cache returns error. I do not think this is a regular error.

pkg/neg/syncers/transaction.go
freehan: I can go over this with you.

pkg/neg/readiness/interface.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"k8s.io/apimachinery/pkg/types"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+)
+
+// Reflector defines
+type Reflector interface {
+	// Run starts up the readiness reflector
+	Run(stopCh <-chan struct{})
+	// SyncPod registers the pod with reflector
```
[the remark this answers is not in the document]
freehan: The reason why I use obj inteface{} mostly is because the informer can throw that into the interface directly. 

I can also pass *v1.Pod, instead

pkg/neg/readiness/interface.go
```diff
@@ -0,0 +1,54 @@
...
+	SyncPod(obj interface{})
+	// CommitPods signals the reflector that pods has been added to one NEG
+	CommitPods(syncerKey negtypes.NegSyncerKey, negName string, zone string, endpointMap map[negtypes.NetworkEndpoint]types.NamespacedName)
+}
+
+// NegLookup defines an interface for looking up pod membership
+type NegLookup interface {
+	// ReadinessGateEnabledNegs returns a list of NEGs which has readiness gate enabled for the input pod's namespace and labels.
+	ReadinessGateEnabledNegs(namespace string, labels map[string]string) []string
```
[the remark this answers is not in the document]
freehan: This interface does not call the compute API to get NEGs.

It inputs a Pod's namespace and labels, and SyncerManager will check if the pod belongs to any NEG services, if so, then return the NEG names of the those services

pkg/neg/readiness/interface.go
```diff
@@ -0,0 +1,54 @@
...
+	// ReadinessGateEnabled returns true if the NEG requires readiness feedback
+	ReadinessGateEnabled(syncerKey negtypes.NegSyncerKey) bool
```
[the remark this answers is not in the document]
freehan: No. This again will only examine the SyncerManager internals.

It checks if a Syncer has ReadinessGate enabled. Only NEGs for Ingress will have ReadinessGate enabled. Because the standalone NEGs may not have health status if they are not connected to Backend Service

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+
+// negMeta references a GCE NEG resource
+type negMeta struct {
+	SyncerKey negtypes.NegSyncerKey
+	// NEG name
+	Name string
+	Zone string
+}
+
+func (n negMeta) String() string {
```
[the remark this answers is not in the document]
freehan: This is used as a key in the map. Using value instead of pointer mostly for avoiding nil pointer bug.

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+// poller trackes
+type poller struct {
+	lock sync.Mutex
+	// pollMap contains the targets needed to be polled
+	pollMap map[negMeta]pollTarget
+
+	podLister cache.Indexer
+	lookup    NegLookup
+	patcher   podStatusPatcher
+	negCloud  negtypes.NetworkEndpointGroupCloud
```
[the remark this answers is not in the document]
freehan: NegLookup is for looking up SyncerManager internals.

negCloud is for interacting with GCE NEG API

pkg/neg/readiness/poller.go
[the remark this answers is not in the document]
bowei: I think it's really complicates the rest of your logic, so I'm not sure it's worth it? Can't we just ensure we don't put a nil into the map?

pkg/neg/readiness/utils.go
freehan: get better logging than the one api server spits?

pkg/neg/readiness/utils.go
rramkumar1: We have a util for this already

https://github.com/kubernetes/ingress-gce/blob/master/pkg/utils/patch.go#L25

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// ScanForWork returns the list of NEGs that should be polled
+func (p *poller) ScanForWork() []negMeta {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	ret := []negMeta{}
+	for key, target := range p.pollMap {
+		if target.Polling {
+			continue
```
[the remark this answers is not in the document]
freehan: added logging outside

pkg/neg/readiness/poller.go
freehan: done

pkg/neg/readiness/poller.go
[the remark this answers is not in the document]
freehan: Actually, I have to use it as value because it is used as key of the map. If I use it as pointer, the map key would not match.

pkg/neg/readiness/utils.go
freehan: Is it used anywhere else? 

Or I can reference it here.

pkg/neg/readiness/poller.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// RegisterNegEndpoints registered the endpoints that needed to be poll for the NEG with lock
+func (p *poller) RegisterNegEndpoints(key negMeta, endpointMap map[negtypes.NetworkEndpoint]types.NamespacedName) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	p.registerNegEndpoints(key, endpointMap)
+}
+
+// registerNegEndpoints registered the endpoints that needed to be poll for the NEG
```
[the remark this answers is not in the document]
freehan: done

pkg/neg/readiness/utils.go
freehan: done

freehan:
I have adjusted the commits. PTAL

PR #751, opened 2019-05-09
Do not run as root by default in docker

reetasingh:
Processes in a container should not run as root, or assume that they are root. Instead, a good practice is to create a user in your Dockerfile with a known UID and GID, and run your process as this user. Images that follow this pattern are easier to run securely by limiting access to resources

### DIFF
--- a/Dockerfile.glbc
+++ b/Dockerfile.glbc
@@ -16,4 +16,8 @@
 FROM gcr.io/distroless/static:latest
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+
+# nobody:nobody
+USER 65534:65534
+
 ENTRYPOINT ["/ARG_BIN"]
--- a/Dockerfile.e2e-test
+++ b/Dockerfile.e2e-test
@@ -20,4 +20,7 @@ COPY cmd/e2e-test/run.sh /run.sh
 RUN apt-get update && apt-get install -y curl python
 RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-206.0.0-linux-x86_64.tar.gz | tar -zxf -
 
+# nobody:nobody
+USER 65534:65534
+
 ENTRYPOINT ["/run.sh"]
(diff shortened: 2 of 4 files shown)


rramkumar1:
@reetasingh Thanks for the PR. Can you please provide a detailed explanation of what the change is doing in the PR description?

reetasingh:
@rramkumar1 updated the description

reetasingh:
@rramkumar1 , @bowei Can you approve this if changes look good?

rramkumar1:
@reetasingh This sounds find to me but I would like @bowei to LGTM since I might be missing something that would be affected by this potentially.

Dockerfile.glbc
```diff
@@ -16,4 +16,8 @@
 FROM gcr.io/distroless/static:latest
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+
+# nobody:nobody
+USER 65534:65534
```
MrHohn: @yuwenma  Would this work on distroless? We knew `nobody` won't (https://github.com/kubernetes/kubernetes/pull/77690#issuecomment-491437440) :)

yuwenma: Thanks Zihong for pointing it out.

@reetasingh have you tried building those images based on distroless images?

reetasingh: @yuwenma yes, I was able to build the images with my changes (USER 65534:65534). Did not see an issue with the user. 
Also, we are already using user 65534 in another Dockerfile in this repo. https://github.com/kubernetes/ingress-gce/blob/master/Dockerfile.404-server

If needed, I can create a new user and use that.

reetasingh: soemthing like ```RUN adduser -D ingress-gce```

reetasingh: @yuwenma does this looks good? or should I change it

yuwenma: Yeah, if it's tested and works well. I have no other concerns. :smile:

Issue #755, opened 2019-05-23
Deploying to GKE self managed has invalid YAML


KatrinaHoffert:
I discovered this while working on another ticket. It's just a silly missing line. Surprised it wasn't caught earlier since it should always prevent the `docs/deploy/gke/gke-self-managed.sh` script from working. I'll fix this and submit a PR.

PR #746, opened 2019-05-02
Update canonical rbac.yaml with latest, minimal bindings

dekkagaijin:
This work came out of https://github.com/kubernetes/kubernetes/pull/77140

/assign rramkumar1

PR #756, opened 2019-05-23
Deploying to GKE self managed has invalid YAML

KatrinaHoffert:
This is for issue #755. Straightforward fix for the invalid syntax. I tested the script with this change.

Given that the invalid syntax should have prevented anyone from using the script, it is rather concerning. Implies no tests for this. But creating a meaningful test for it is quite non-trivial, so not doing that here.

### DIFF
--- a/docs/deploy/resources/rbac.yaml
+++ b/docs/deploy/resources/rbac.yaml
@@ -25,6 +25,7 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
   name: system:controller:glbc
+subjects:
 - kind: ServiceAccount
   name: glbc
   namespace: kube-system


KatrinaHoffert:
Gah, I have dirty commits on that branch, lemme clean it up.

KatrinaHoffert:
Ready for review.

rramkumar1:
@KatrinaHoffert Thanks for the fix. For posterity, looks like the typo was introduced in #746. You are right that we need some unit testing around those files that lints the yaml at the very least. I will open an issue for that.

KatrinaHoffert:
I actually had already made a ticket for that. #760. It's a bit broader than just unit tests, but does mention similar, with points 2 and 3 of simplifying that.

Issue #661, opened 2019-03-04
GCP Description field is blank


thockin:
Service LBs populate the description field with helpful information.   I see GCE ingress does not do that for things like forwarding rules.  Would be helpful in UIs.

KatrinaHoffert:
I assume this is referring to the descriptions on the load balancer forwarding rules like shown below?

![Image of GCP console](https://i.imgur.com/f7TDIzl.png)

This seems to only be visible in the advanced menu, when viewing the forwarding rule directly, or with the likes of `gcloud compute forwarding-rules describe` (noting this since it took me some testing to find what was likely referred here).

Anyway, looking into this one.

Aut0R3V:
@thockin can we close this issue? I think that #757 solves the problem

PR #757, opened 2019-05-24
GCP forwarding rule/target proxy description field populated

KatrinaHoffert:
This is for issue #661.

The description are the [same values as AWS uses in their Ingress](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/#resource-tags).

Unit tests were updated. Can be viewed by creating an ingress and in Cloud Console, go to Network services > Load balancing > advanced view (description column). Note that the cluster name field is probably not the actual cluster name, but does identify the cluster (and instance group, etc).

### DIFF
--- a/pkg/loadbalancers/l7.go
+++ b/pkg/loadbalancers/l7.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 
 	compute "google.golang.org/api/compute/v1"
@@ -307,4 +308,17 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
 	// parsing logic in a single location.
 	resourceName, _ = ingAnnotations[fmt.Sprintf("%v/%v", annotations.StatusPrefix, resourceName)]
 	return resourceName
+}
+
+// description gets a description for the ingress GCP resources.
+func (l *L7) description() (string, error) {
+	if l.runtimeInfo.Ingress == nil {
+		return "", fmt.Errorf("missing Ingress object to construct description for %s", l.Name)
+	}
+
+	namespace := l.runtimeInfo.Ingress.ObjectMeta.Namespace
+	ingressName := l.runtimeInfo.Ingress.ObjectMeta.Name
+	namespacedName := types.NamespacedName{Name: ingressName, Namespace: namespace}
+
+	return fmt.Sprintf(`{"kubernetes.io/ingress-name": %q}`, namespacedName.String()), nil
 }
--- a/pkg/loadbalancers/loadbalancers_test.go
+++ b/pkg/loadbalancers/loadbalancers_test.go
@@ -36,8 +36,10 @@ import (
 )
 
 const (
-	defaultNamespace = "default"
-	defaultZone      = "zone-a"
+	clusterName = "uid1"
+	ingressName = "test"
+	namespace   = "namespace1"
+	defaultZone = "zone-a"
 )
 
 var (
@@ -47,7 +49,8 @@ var (
 func newIngress() *extensions.Ingress {
 	return &extensions.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: defaultNamespace,
+			Name:      ingressName,
+			Namespace: namespace,
 		},
 	}
 }
@@ -66,9 +69,9 @@ func TestCreateHTTPLoadBalancer(t *testing.T) {
 	gceUrlMap := utils.NewGCEURLMap()
 	gceUrlMap.DefaultBackend = &utils.ServicePort{NodePort: 31234}
 	gceUrlMap.PutPathRulesForHost("bar.example.com", []utils.PathRule{utils.PathRule{Path: "/bar", Backend: utils.ServicePort{NodePort: 30000}}})
-	namer := utils.NewNamer("uid1", "fw1")
+	namer := utils.NewNamer(clusterName, "fw1")
 	lbInfo := &L7RuntimeInfo{
-		Name:      namer.LoadBalancer("test"),
+		Name:      namer.LoadBalancer(ingressName),
 		AllowHTTP: true,
 		UrlMap:    gceUrlMap,
 		Ingress:   newIngress(),
@@ -89,9 +92,9 @@ func TestCreateHTTPSLoadBalancer(t *testing.T) {
 	gceUrlMap := utils.NewGCEURLMap()
 	gceUrlMap.DefaultBackend = &utils.ServicePort{NodePort: 31234}
 	gceUrlMap.PutPathRulesForHost("bar.example.com", []utils.PathRule{utils.PathRule{Path: "/bar", Backend: utils.ServicePort{NodePort: 30000}}})
-	namer := utils.NewNamer("uid1", "fw1")
+	namer := utils.NewNamer(clusterName, "fw1")
 	lbInfo := &L7RuntimeInfo{
-		Name:      namer.LoadBalancer("test"),
+		Name:      namer.LoadBalancer(ingressName),
 		AllowHTTP: false,
 		TLS:       []*TLSCerts{createCert("key", "cert", "name")},
 		UrlMap:    gceUrlMap,
@@ -125,6 +128,9 @@ func verifyHTTPSForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers) 
 	if !utils.EqualResourcePaths(fws.Target, tps.SelfLink) {
 		t.Fatalf("fws.Target = %q, want %q", fws.Target, tps.SelfLink)
 	}
+	if fws.Description == "" {
+		t.Errorf("fws.Description not set; expected it to be")
+	}
 }
 
 func verifyHTTPForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers) {
@@ -145,6 +151,9 @@ func verifyHTTPForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers) {
 	if !utils.EqualResourcePaths(fws.Target, tps.SelfLink) {
 		t.Fatalf("fw.Target = %q, want %q", fws.Target, tps.SelfLink)
 	}
+	if fws.Description == "" {
+		t.Errorf("fws.Description not set; expected it to be")
+	}
 }
 
 // Tests that a certificate is created from the provided Key/Cert combo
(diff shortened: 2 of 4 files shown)


KatrinaHoffert:
Oops. That was dumb. 

/retest

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -36,8 +36,10 @@ import (
 )
 
 const (
-	defaultNamespace = "default"
-	defaultZone      = "zone-a"
+	clusterName = "uid1"
+	ingressName = "test"
+	namespace   = "namespace1"
```
rramkumar1: Any particular reason why the namespace was changed from "default" to "namespace1"?

pkg/loadbalancers/l7.go
```diff
@@ -308,3 +308,19 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
 	resourceName, _ = ingAnnotations[fmt.Sprintf("%v/%v", annotations.StatusPrefix, resourceName)]
 	return resourceName
 }
+
+// getDescription gets a description for the L7 forwarding rule. This matches the
+// patterns AWS uses for their tags
```
rramkumar1: I would omit this from the comment. If we ever need to know why the description field has the pattern, we can always go back to this PR's description.

pkg/loadbalancers/l7.go
```diff
@@ -308,3 +308,19 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
...
+	if l.runtimeInfo.Ingress == nil {
+		return "", fmt.Errorf("Missing ingress object for %s", l.Name)
+	}
+
+	// Note: l.Name is not the k8s ingress name, but the GCE internal name.
+	clusterName := l.namer.UID()
+	namespace := l.runtimeInfo.Ingress.ObjectMeta.Namespace
+	ingressName := l.runtimeInfo.Ingress.ObjectMeta.Name
+
+	return fmt.Sprintf(`{"kubernetes.io/cluster/%s": "owned", "kubernetes.io/namespace": "%s", "kubernetes.io/ingress-name":"%s"}`, clusterName, namespace, ingressName), nil
```
rramkumar1: Couple points: 

1. I would substitute "kubernetes.io/cluster/%s" : "owned" with "kubernetes.io/cluster" : "[CLUSTER]"
2. Let me get back to you ASAP on what all information we want in the description. I'm not sure yet if we want more / less / same.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -145,6 +153,14 @@ func verifyHTTPForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers) {
 	if !utils.EqualResourcePaths(fws.Target, tps.SelfLink) {
 		t.Fatalf("fw.Target = %q, want %q", fws.Target, tps.SelfLink)
 	}
+	verifyForwardingRuleDesc(t, fws, clusterName)
+}
+
+func verifyForwardingRuleDesc(t *testing.T, fws *compute.ForwardingRule, clusterName string) {
```
rramkumar1: Does this need to be a function? I would say just inline it inside the two funcs that call it. You can even make it one line less by doing:

if fws.Description != expectedDescription(clusterName, namespace, ingressName) {}

pkg/loadbalancers/loadbalancers_test.go
KatrinaHoffert: I changed it solely so that the tests were obvious in using the specified namespace. I was wary of the possibility of "default" being assumed somewhere (since, well, it's the default when no namespace is specified) and wanted the tests to be able to differentiate from that possibility.

pkg/loadbalancers/l7.go
KatrinaHoffert: Seems reasonable. I was just trying to figure out what kinda metadata others might have. It doesn't seem documented anywhere I could find.

pkg/loadbalancers/loadbalancers_test.go
KatrinaHoffert: I can inline it. I just disliked the repetition (mind you, originally it was longer because I didn't have the `expectedDescription` function).

pkg/loadbalancers/l7.go
KatrinaHoffert: Done. Along with the other comments.

KatrinaHoffert:
@rjnagal, have you seen my updates? Curious if there's anything more needed here.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -60,15 +63,19 @@ func newFakeLoadBalancerPool(f LoadBalancers, t *testing.T, namer *utils.Namer)
 	return NewLoadBalancerPool(f, namer, events.RecorderProducerMock{})
 }
 
+func expectedDescription(clusterName, namespace, ingressName string) string {
```
rramkumar1: Doing this feels a little weird to me. You are basically re-implementing getDescription() to test that it works. 

I'm wondering whether its even worth unit testing that we are putting the right description. To me, it makes more sense to check that description is non-empty.

pkg/loadbalancers/l7.go
```diff
@@ -308,3 +308,17 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
 	resourceName, _ = ingAnnotations[fmt.Sprintf("%v/%v", annotations.StatusPrefix, resourceName)]
 	return resourceName
 }
+
+// getDescription gets a description for the L7 forwarding rule.
+func (l *L7) getDescription() (string, error) {
```
rramkumar1: Nit: omit the get, just call it description()

pkg/loadbalancers/loadbalancers_test.go
KatrinaHoffert: Good point.

pkg/loadbalancers/l7.go
```diff
@@ -308,3 +308,17 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
...
+	if l.runtimeInfo.Ingress == nil {
+		return "", fmt.Errorf("Missing ingress object for %s", l.Name)
+	}
+
+	// Note: l.Name is not the k8s ingress name, but the GCE internal name.
+	clusterName := l.namer.UID()
+	namespace := l.runtimeInfo.Ingress.ObjectMeta.Namespace
+	ingressName := l.runtimeInfo.Ingress.ObjectMeta.Name
+
+	return fmt.Sprintf(`{"kubernetes.io/cluster": "%s", "kubernetes.io/namespace": "%s", "kubernetes.io/ingress-name":"%s"}`, clusterName, namespace, ingressName), nil
```
thockin: It'[s not clear to me what value cluster name will be ?

thockin: A few thoughts:

Service uses "namespace/name" for service-name -- why be different?

The use of "kubernetes.io" is suspect but historically we did it for Service so I guess this is not worse.  The context is not really the domain of Kubernetes, so it's borderline but I won't complain. 

Service doesn't have cluster name, but probably should.  Hint hint.

`%q` is generally safer than `"%s"` in every way.  Should be a no-op for this case, but a good habit.

KatrinaHoffert: The cluster name is probably not actually the cluster's name. It depends on if the GLBC was launched with the cluster name actually set (ie, the `--cluster-uid` flag). Of course, it's usually not and doesn't actually need to match the cluster's name. So that value is actually very likely just a UUID that groups the GLBC resources together. I didn't see a good way to get the actual cluster name. It's surprisingly tricky to get -- the only way I see is to utilize that we know the instance groups and query one of the instances to get the `cluster-name` custom metadata that we set. On GKE, GLBC could be instantiated with the cluster name passed in.

r.e. labels, I was initially looking at how [GCE load balancers use `kubernetes.io/service-name`](https://github.com/kubernetes/kubernetes/blob/8765fa2e48974e005ad16e65cb5c3acf5acff17b/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_naming.go#L88). But since ingresses aren't services, I had to do some research on how the heck that label was chosen without much luck and decided to use the same labels as AWS for their ingresses, assuming they got them from somewhere and that it at least follow precedent.

On services (well, load balancers specifically), they seem to have the cluster name available, so could easily be extended, too (that's something for another issue, though -- they're not even in this repo). I should look into if it's exposed somewhere accessible on the cluster. It's `ctx.ComponentConfig.KubeCloudShared.ClusterName`

KatrinaHoffert: The usage of `%q` has been resolved. I haven't changed the other parts pending the explanation being read.

thockin:
I think consistency between service and ingress on GCE is more useful than
consistency between GCE and AWS, so I would go with
`ingress-name=namespace/name`.

As for cluster name, we ALMOST have a concept of cluster name and we don't
have it on Service yet, so what if we punted that part until we have proper
cluster names?

KatrinaHoffert:
Done @thockin.

pkg/loadbalancers/l7.go
```diff
@@ -308,3 +309,16 @@ func GCEResourceName(ingAnnotations map[string]string, resourceName string) stri
 	resourceName, _ = ingAnnotations[fmt.Sprintf("%v/%v", annotations.StatusPrefix, resourceName)]
 	return resourceName
 }
+
+// description gets a description for the ingress GCP resources.
+func (l *L7) description() (string, error) {
+	if l.runtimeInfo.Ingress == nil {
+		return "", fmt.Errorf("Missing ingress object for %s", l.Name)
```
rramkumar1: Nit: errors should start with lowercase.

rramkumar1: Can you call the error: missing Ingress object to construct description"

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -104,10 +107,10 @@ func TestCreateHTTPSLoadBalancer(t *testing.T) {
 	if err != nil || l7 == nil {
 		t.Fatalf("Expected l7 not created")
 	}
-	verifyHTTPSForwardingRuleAndProxyLinks(t, f)
+	verifyHTTPSForwardingRuleAndProxyLinks(t, f, clusterName)
 }
 
-func verifyHTTPSForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers) {
+func verifyHTTPSForwardingRuleAndProxyLinks(t *testing.T, f *FakeLoadBalancers, clusterName string) {
```
rramkumar1: Don't think you need the clusterName arg anymore

KatrinaHoffert:
@rramkumar1 done

rramkumar1:
@KatrinaHoffert Sorry one last thing, can you squash commits? There needs to be only one commit for this since its pretty small.

KatrinaHoffert:
Squashed @rramkumar1

rramkumar1:
/lgtm

Thanks!

Issue #758, opened 2019-05-24
Update GKE self managed script


KatrinaHoffert:
Kinda a follow up to #755, I noted a few other issues I had with the flow described in `docs/deploy/gke/README.md`. Firstly, the readme doesn't mention the need to update `glbc.yaml` for your custom image location. Secondly, the command flags are outdated and prevent the controller from running.

I want to also look into automating more of this or at least improving the documentation.

* We already get the project name from `gcloud`
* Network and subnetwork name are easily determined from `gcloud container cluster describe` (the cluster name is already provided to the setup script)
* The node-instance-prefix should be determinable from the cluster name and we strangly don't even allow the user to continue if it doesn't make that expectation!
* Node tags are a little harder. Get the instance groups from `describe cluster`, list instances, and describe those instances to get the network tags.

Zone should just be asked by the script. Should take the image URL in, too, so it can be just one command to set everything up.

I'll look into this. I already did the work of figuring out how to get that script to work for me and confirmed its working.

rramkumar1:
Closing since relevant PR has been merged. 

Thanks @KatrinaHoffert !

PR #759, opened 2019-05-24
Updates for usability of GKE self managed setup script

KatrinaHoffert:
For issue #758.

This does several things to make this setup much easier. But in short, it allows a single command to be used to setup for the vast majority of cases (with just `./gke-self-managed.sh -n CLUSTER_NAME -z ZONE`).

1. The CLI flags are updated to actually work (ie, represent those used by the GLBC now).
2. The gce.conf file is automatically populated from querying the cluster. All options can be instead provided as CLI flags if the querying gets them wrong or to speed things up slightly (the time to query is pretty negligible compared to the time to restart the cluster to turn off the default GLBC, though).
3. The `glbc.yaml` file is automatically setup with the image instead of letting users create an invalid deployment that won't be caught till runtime.
4. The script defaults to building and pushing the image, making the script usable for quickly testing changes. The image to use can be specified as a flag instead. The registry to use is customizable the same way `make push` does.
5. We handle progress better. No more winging it by hoping the API server is up  and running.

Opted not to make a test for it this ticket. We really should have some kinda test (since the script got terribly broken in 2 ways), but it's quite difficult to test given that it requires cluster ownership, needs push access to some container registry, requires an existing cluster, and is very slow no matter how you cut it (it needs to restart a cluster and realistically, a test would also need to create an ingress, which takes a few minutes). I'll make a new issue for testing it, but it's likely to be lower priority as the ROI is complicated.

This is a non-breaking change, since the existing `gce.conf` file is used as a template and if it were already filled in, those fields would just be ignored. The files that need modification are now copied to and `.gitignored`, too, so no risks of accidental commits or annoyance of having them show up in the `git status`.

### DIFF
--- a/docs/deploy/gke/README.md
+++ b/docs/deploy/gke/README.md
@@ -30,9 +30,6 @@ permission to do basically anything so all commands the script runs should
 theoretically work. If not, the script will do its best to fail gracefully
 and let you know what might have went wrong.
 
-The second step is to make sure you populate the [gce.conf](../resources/gce.conf) file. The instructions
-for populating the file are in the file itself. You just have to fill it in.
-
 # Important Details
 
 Most likely, you want to know what this script is doing to your cluster in order
@@ -41,11 +38,9 @@ can go ahead and skip this section.
 
 Here is a brief summary of each major thing we do and why:
 
-1. Turn off GLBC and turn on new GLBC in the cluster
-    * To be brief, the maintenance cost of running a new controller on the master
-      is actually pretty high. This is why we chose to move the controller
-      to the cluster.
-1. Create a new k8s RBAC role
+1. (If applicable) figure out information you didn't provide from the cluster and
+   build and push the GLBC image.
+2. Create a new k8s RBAC role
     * On the master, the GLBC has unauthenticated access to the k8s API server.
       Once we move the GLBC to the cluster, that path is gone. Therefore, we need to
       configure a new RBAC role that allows GLBC the same access.
@@ -54,10 +49,13 @@ Here is a brief summary of each major thing we do and why:
       token pulled from a private GKE endpoint. Moving to the cluster will result in
       us not being able to utilize this. Therefore, we need to create a new GCP
       service account and a corresponding key which will grant access.
-4. Start new GLBC (and default backend) in the cluster
-    * As stated before, we need to run the GLBC in the cluster. We also need to
-      startup a new default backend because the mechanism we use to turn off the
-      master GLBC removes both the GLBC and the default backend.
+4. Update the cluster to turn off the default GLBC
+    * This restarts the cluster master. The API server will be temporarily unavailable.
+5. Create our own default backend and custom GLBC on the cluster
+    * We need to startup a new default backend because the mechanism we
+      use to turn off the master GLBC removes both the GLBC and the default backend.
+    * This is dependent on the default GLBC being running initially, as we use
+      the same nodeport.
     * Because we have to recreate the default backend, there will be a small
       segment of time when requests to the default backend will time out.
 
@@ -71,30 +69,46 @@ Here is an explanation of each script dependency.
 1. [gce.conf](../resources/gce.conf)
     * This file normally sits on the GKE master and provides important config for
       the GCP Compute API client within the GLBC. The GLBC is configured to know
-      where to look for this file. In this case, we simply mount the file as a
-      volume and tell GLBC to look for it there.
+      where to look for this file. In this case, we simply mount a customized copy
+      of the file as a volume and tell GLBC to look for it there.
 2. [default-http-backend.yaml](../resources/default-http-backend.yaml)
     * This file contains the specifications for both the default-http-backend
       deployment and service. This is no different than what you are used to
       seeing in your cluster. In this case, we need to recreate the default
-      backend since turning off the GLBC on the master removes it.
+      backend since turning off the GLBC on the master removes it. Note that we
+      modify the file to use the same node port as before before we create the resource.
 3. [rbac.yaml](../resources/rbac.yaml)
     * This file contains specification for an RBAC role which gives the GLBC
       access to the resources it needs from the k8s API server.
 4. [glbc.yaml](../resources/glbc.yaml)
     * This file contains the specification for the GLBC deployment. Notice that in
       this case, we need a deployment because we want to preserve the controller
-      in case of node restarts.
+      in case of node restarts. The resource we create is from a customized copy of
+      this file that uses the specified (or newly built) image.
 
 Take a look at the script to understand where each file is used.
 
 # Running the Script
 
-Run the command below to see the usage:
+The script can take in a number of settings, but only really requires the cluster
+name and zone. You can provide other settings for if we incorrectly intuit any
+values. Usage:
 
-`./gke-self-managed.sh --help`
+```shell
+# Must be in the script directory
+cd docs/deploy/gke
+./gke-self-managed.sh -n CLUSTER_NAME -z ZONE
+```
 
-After that, it should be self-explanatory!
+For other options, see the `--help` output.
+
+Can speed things up if you've already pushed an image (or want to use a specific
+version) with `--image-url PATH` (eg,
+`--image-url k8s.gcr.io/ingress-gce-glbc-amd64:v1.5.2`). Can also set the `REGISTRY`
+env var to provide a custom place to push your image to if it's not the same project
+as your cluster is in. The tags pushed to will be that of the
+`git describe --tags --always --dirty` command. If the `VERSION` env var is set, that
+will be used as the tag.
 
 # Common Issues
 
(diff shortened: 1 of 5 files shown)


KatrinaHoffert:
And yes, my bash skills are not that great. There's probably tons of room for micro-optimizations and all with the tools I use for filtering text and all. I tried to use `gcloud`'s formatting as much as possible. I don't think it's a big area to nitpick, though, since the amount of time to update the cluster massively dominates the runtime.

There's some areas for future improvement that I did *not* do due to time constraints. Future improvements could be:

1. Check the cluster actually exists, as some commands assume it does.
2. Don't require the CWD to be the script's location.
3. Better handling of resources already existing (can often be skipped to allow resuming).
4. Stream the `make` output with `tee`.
5. There's probably some `gcloud --format` magic that could be used to avoid that ugly `cut`  for getting instance group names from the cluster.
6. We use the output of `gcloud container clusters describe` up to three times and it could be cached.
7. We should check the `gcloud config` stuff and permissions before we do anything.
8. We are dependent on the default GCLB HTTP backend already existing. We probably don't need to (just allow a random node port in that case?).
9. The heck is `manual_glbc_provision`?
10. Shebang is not on first line.
11. Most new commands I added don't handle errors.

docs/deploy/gke/gke-self-managed.sh
```diff
@@ -73,6 +82,14 @@ function cleanup() {
   exit 0
 }
 
+# Loops until calls to the API server are succeeding.
+function wait-for-api-server() {
```
rramkumar1: i'm not sure this will work. The reason I previously let the user decide when to proceed was because its possible that kubectl call returns once and then fails the second time.

docs/deploy/gke/gke-self-managed.sh
```diff
@@ -113,17 +159,78 @@ done
...
+# the build could fail and that should be caught early.
+if [[ -z $IMAGE_URL ]]; then
+  echo "image-url is not set; we will build and push ourselves. This may take a few minutes."
+  if [[ -z $REGISTRY ]]; then
+    REGISTRY="gcr.io/$PROJECT_ID"
+    echo "REGISTRY is not set. Defaulting to: $REGISTRY"
+  fi
+  # Extract just the URL (with tag) from the `make push` output. `sed -n` makes it
+  # not print normally while the `/../../p` makes it print just what it matched.
+  MAKE_PUSH_OUTPUT=$(cd ../../../; REGISTRY="$REGISTRY" make push 2>&1)
```
rramkumar1: I'm not a big fan of doing image pushes in this script. Why can't this just be done as a step before running the script?

docs/deploy/resources/glbc.yaml
```diff
@@ -23,7 +23,7 @@ spec:
       terminationGracePeriodSeconds: 600
       hostNetwork: true
       containers:
-      - image: k8s.gcr.io/ingress-gce-glbc-amd64:[SET VERSION HERE]
+      - image: ### IMAGE URL HERE ###
```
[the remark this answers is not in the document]
KatrinaHoffert: Can do. It was a "comment" previously because I wasn't sure I'd generate that file and just wanted it to fail loudly if the user forgot to fill it in (cause I did the first time!).

docs/deploy/gke/gke-self-managed.sh
KatrinaHoffert: What if I make pushing non default? I mostly just wanted a one command solution to getting changes running.

docs/deploy/gke/gke-self-managed.sh
KatrinaHoffert: What condition does that occur under? I can restore the old behavior, but I never encountered the described behavior in my testing.

docs/deploy/gke/gke-self-managed.sh
rramkumar1: Sure, non-default is fine.

docs/deploy/gke/gke-self-managed.sh
rramkumar1: When I first wrote the script, it happened for me quite often. I originally had the approach you have which was to just run some kubectl command. I quickly noticed quite often that the command would succeed but then subsequent commands in the script that talked to the API server would fail.

KatrinaHoffert: Huh. I do wonder if anything else has changed independent of the script (ie, if `cluster update` now blocks until the API server really is ready). But I can add the confirmation back in, too.

rramkumar1: Feel free to add this as a command-line option as well. So a flag dictates whether the script waits for user intervention or is greedy and proceeds after the first 200.

KatrinaHoffert: This was done. I made the `--no-confirm` flag I added skip the confirmation. And in testing that, I realized `--cleanup` could dangerously ignore flags depending on their order, so fixed that.

docs/deploy/gke/gke-self-managed.sh
KatrinaHoffert: This was done with `--build-and-push`. One of either `--image-url` or `--build-and-push` is required (otherwise we'd push the placeholder...). I also added some more protection from things going wrong with the build.

docs/deploy/resources/gce.conf
```diff
@@ -5,4 +5,4 @@ network-name = YOUR CLUSTER'S NETWORK
 subnetwork-name = YOUR CLUSTER'S SUBNETWORK
 node-instance-prefix = gke-YOUR CLUSTER'S NAME
 node-tags = NETWORK TAGS FOR YOUR CLUSTER'S INSTANCE GROUP
-local-zone= YOUR CLUSTER'S ZONE
+local-zone = YOUR CLUSTER'S ZONE
```
rramkumar1: Can you structure the file as follows? This makes it cleaner when substituting via sed.
```
[global]
token-url = nil	
# Your cluster's project
project-id = [PROJECT]
# Your cluster's network
network-name =  NETWORK	
# Your cluster's subnetwork
subnetwork-name = [SUBNETWORK]
# Prefix for your cluster's IG
node-instance-prefix = gke-[CLUSTER_NAME]
# Network tags for your cluster's IG
node-tags = [NETWORK_TAGS]
# Zone the cluster lives in	
local-zone= [ZONE]
```

docs/deploy/gke/README.md
```diff
@@ -71,30 +69,46 @@ Here is an explanation of each script dependency.
...
 
 # Running the Script
 
-Run the command below to see the usage:
-
-`./gke-self-managed.sh --help`
-
-After that, it should be self-explanatory!
+The script can take in a number of settings, but only really requires the cluster
+name and zone. You can provide other settings for if we incorrectly intuit any
```
rramkumar1: s/intuit/deduce

docs/deploy/gke/gke-self-managed.sh
```diff
@@ -211,24 +401,23 @@ done
...
+run_maybe_dry_kubectl kubectl create -f ../resources/default-http-backend.yaml.gen
 if [[ $? -eq 1 ]];
 then
   # Prompt the user to finish the last steps by themselves. We don't want to
   # have to cleanup and start all over again if we are this close to finishing.
-  error_exit "Error-bot: Issue starting default backend. ${PERMISSION_ISSUE}. We are so close to being done so just manually start the default backend with NodePort: ${NODE_PORT} and create glbc.yaml when ready"
+  error_exit "Error-bot: Issue starting default backend. ${PERMISSION_ISSUE}. We are so close to being done so just manually start the default backend with NodePort: ${NODE_PORT} (see default-http-backend.yaml.gen) and create glbc.yaml.gen when ready."
 fi
-rm ../resources/default-http-backend-copy.yaml
+rm ../resources/default-http-backend.yaml.gen # Not useful to keep because the port changes each run.
```
rramkumar1: Please add the comment above the line.

docs/deploy/gke/gke-self-managed.sh
```diff
@@ -15,20 +15,90 @@
...
+  echo "  --build-and-push         If set, we will instead build the code and push the"
+  echo "                           image, which will be used as the image-url. Set the"
+  echo "                           REGISTRY env var or we'll use the project GCR (see"
+  echo "                           the Makefile for defaults and other configuration)"
+  echo "  --help                   Display this help and exit"
+  echo "  --network-name           Override for the network-name gce.conf value"
+  echo "  --subnetwork-name        Override for the subnetwork-name gce.conf value"
+  echo "  --node-instance-prefix   Override for the node-instance-prefix gce.conf value"
+  echo "  --network-tags           Override for the node-tags gce.conf value"
+  echo "  --no-confirm             Don't ask confirmation to reenable GLBC on cleanup"
```
rramkumar1: Can you call this confirm rather than --no-confirm? Make it easier to reason about in the code.

KatrinaHoffert: Done. I assumed you mean the env var should be `CONFIRM`, to avoid the double negative. I didn't change the flag, since confirmation is the default.

KatrinaHoffert:
All done, @rramkumar1

rramkumar1:
/lgtm

Thanks for all these changes!

PR #773, opened 2019-06-14
Add support for L7-ILB

spencerhance:
**DO NOT MERGE**
This PR is being broken up into smaller PRs but I'm keeping this here in the meantime for reference.

Adds a new ingress class for the l7-ILB and integrates it throughout the controller.  This involves switching most things over to composite types, with a few things spared momentarily to improve short-term stability, or spared indefinitely (e.g. health checks, NEGs).  Additionally, includes some handwritten composite functions for more complex things.

Also adds a few unit tests specific to ILB and refactors the mocks, but most of the changes should already be covered by the current tests.

A follow up PR will be added with some e2e tests.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -14,10 +14,11 @@ limitations under the License.
 package backends
 
 import (
+	features2 "k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"net/http"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/compute/v1"
 	"k8s.io/ingress-gce/pkg/backends/features"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/utils"
@@ -27,8 +28,9 @@ import (
 
 // Backends handles CRUD operations for backends.
 type Backends struct {
-	cloud *gce.Cloud
-	namer *utils.Namer
+	cloud          *gce.Cloud
+	namer          *utils.Namer
+	compositeCloud *composite.Cloud
 }
 
 // Backends is a Pool.
@@ -36,11 +38,12 @@ var _ Pool = (*Backends)(nil)
 
 // NewPool returns a new backend pool.
 // - cloud: implements BackendServices
-// - namer: procudes names for backends.
+// - namer: produces names for backends.
 func NewPool(cloud *gce.Cloud, namer *utils.Namer) *Backends {
 	return &Backends{
-		cloud: cloud,
-		namer: namer,
+		cloud:          cloud,
+		namer:          namer,
+		compositeCloud: composite.NewCloud(cloud),
 	}
 }
 
@@ -65,6 +68,7 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 	}
 
 	version := features.VersionFromServicePort(&sp)
+	key := b.compositeCloud.CreateKey(name, sp.ILBEnabled)
 	be := &composite.BackendService{
 		Version:      version,
 		Name:         name,
@@ -72,30 +76,37 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 		Port:         namedPort.Port,
 		PortName:     namedPort.Name,
 		HealthChecks: []string{hcLink},
+		Region:       key.Region,
 	}
+
+	if sp.ILBEnabled {
+		be.LoadBalancingScheme = "INTERNAL"
+	}
+
 	ensureDescription(be, &sp)
-	if err := composite.CreateBackendService(be, b.cloud, meta.GlobalKey(be.Name)); err != nil {
+	if err := b.compositeCloud.CreateBackendService(be, key); err != nil {
 		return nil, err
 	}
 	// Note: We need to perform a GCE call to re-fetch the object we just created
 	// so that the "Fingerprint" field is filled in. This is needed to update the
 	// object without error.
-	return b.Get(name, version)
+	return b.Get(name, version, sp.ILBEnabled)
 }
 
 // Update implements Pool.
 func (b *Backends) Update(be *composite.BackendService) error {
 	// Ensure the backend service has the proper version before updating.
 	be.Version = features.VersionFromDescription(be.Description)
-	if err := composite.UpdateBackendService(be, b.cloud, meta.GlobalKey(be.Name)); err != nil {
+	isRegional := be.ResourceType == meta.Regional
+	if err := b.compositeCloud.UpdateBackendService(be, b.compositeCloud.CreateKey(be.Name, isRegional)); err != nil {
 		return err
 	}
 	return nil
 }
 
 // Get implements Pool.
-func (b *Backends) Get(name string, version meta.Version) (*composite.BackendService, error) {
-	be, err := composite.GetBackendService(version, b.cloud, meta.GlobalKey(name))
+func (b *Backends) Get(name string, version meta.Version, regional bool) (*composite.BackendService, error) {
+	be, err := b.compositeCloud.GetBackendService(version, b.compositeCloud.CreateKey(name, regional))
 	if err != nil {
 		return nil, err
 	}
@@ -103,8 +114,9 @@ func (b *Backends) Get(name string, version meta.Version) (*composite.BackendSer
 	// API version is required so that we don't lose information from
 	// the existing backend service.
 	versionRequired := features.VersionFromDescription(be.Description)
+
 	if features.IsLowerVersion(versionRequired, version) {
-		be, err = composite.GetBackendService(versionRequired, b.cloud, meta.GlobalKey(name))
+		be, err = b.compositeCloud.GetBackendService(versionRequired, b.compositeCloud.CreateKey(name, regional))
 		if err != nil {
 			return nil, err
 		}
@@ -113,7 +125,7 @@ func (b *Backends) Get(name string, version meta.Version) (*composite.BackendSer
 }
 
 // Delete implements Pool.
-func (b *Backends) Delete(name string) (err error) {
+func (b *Backends) Delete(name string, regional bool) (err error) {
 	defer func() {
 		if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 			err = nil
@@ -122,23 +134,37 @@ func (b *Backends) Delete(name string) (err error) {
 
 	klog.V(2).Infof("Deleting backend service %v", name)
 
+	var version meta.Version
+	if regional {
+		version = features2.ILBVersion
+	} else {
+		version = meta.VersionGA
+	}
+
 	// Try deleting health checks even if a backend is not found.
-	if err = b.cloud.DeleteGlobalBackendService(name); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+	// Use GA version since deleting should still work
+	if err = b.compositeCloud.DeleteBackendService(version, b.compositeCloud.CreateKey(name, regional)); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 		return err
 	}
 	return
 }
 
 // Health implements Pool.
-func (b *Backends) Health(name string) string {
-	be, err := b.Get(name, meta.VersionGA)
+func (b *Backends) Health(name string, version meta.Version, regional bool) string {
+	be, err := b.Get(name, version, regional)
 	if err != nil || len(be.Backends) == 0 {
 		return "Unknown"
 	}
 
 	// TODO: Look at more than one backend's status
 	// TODO: Include port, ip in the status, since it's in the health info.
-	hs, err := b.cloud.GetGlobalBackendServiceHealth(name, be.Backends[0].Group)
+	// TODO (shance) convert to composite types
+	var hs *compute.BackendServiceGroupHealth
+	if regional {
+		hs, err = b.cloud.GetRegionalBackendServiceHealth(name, b.cloud.Region(), be.Backends[0].Group)
+	} else {
+		hs, err = b.cloud.GetGlobalBackendServiceHealth(name, be.Backends[0].Group)
+	}
 	if err != nil || len(hs.HealthStatus) == 0 || hs.HealthStatus[0] == nil {
 		return "Unknown"
 	}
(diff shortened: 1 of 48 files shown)


pkg/utils/serviceport.go
```diff
@@ -47,6 +47,7 @@ type ServicePort struct {
 	Protocol      annotations.AppProtocol
 	TargetPort    string
 	NEGEnabled    bool
+	ILBEnabled    bool
```
rramkumar1: I don't think we should put this here. ServicePort was meant to encap backend settings.

pkg/loadbalancers/l7.go
```diff
@@ -96,28 +101,46 @@ type L7 struct {
...
-	oldSSLCerts []*compute.SslCertificate
+	oldSSLCerts []*composite.SslCertificate
 	// namer is used to compute names of the various sub-components of an L7.
 	namer *utils.Namer
 	// recorder is used to generate k8s Events.
 	recorder record.EventRecorder
+	// version stores what version the resources in the loadbalancer should be
+	version meta.Version
+	// resource type stores the KeyType of the resources in the loadbalancer (e.g. Regional)
+	resourceType meta.KeyType
```
rramkumar1: I think this should be exported and passed as part of the instantiation of the L7. That way, you don't have to have the ILB bool in ServicePort. Does that work?

spencerhance: Hm it seems to be possible, but will probably require some small refactoring since most of the backend/health check code relies on just service ports.

rramkumar1: Hmm actually now that I think about it, ILB being enabled actually is kind of a backend setting since it dictates a different type of scoped resource. So, maybe this is fine.

pkg/loadbalancers/features/ilb.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	ILBVersion = meta.VersionAlpha
+)
+
+// Get Subnet source range for ILB from FrontendConfig
+// TODO: (shance) refactor to use filter
+func ILBSubnetSourceRange(cloud cloud.Cloud, region string) (string, error) {
+	subnets, err := cloud.AlphaSubnetworks().List(context.Background(), region, filter.None)
```
rramkumar1: Shouldn't we use the composite cloud instance here?

pkg/firewalls/controller.go
```diff
@@ -165,8 +165,13 @@ func (fwc *FirewallController) sync(key string) error {
...
 	negPorts := fwc.translator.GatherEndpointPorts(gceSvcPorts)
 
+	ILBSrcRange, err := features.ILBSubnetSourceRange(fwc.ctx.Cloud.GceCloud().Compute(), fwc.ctx.Cloud.GceCloud().Region())
+	if err != nil {
+		return fmt.Errorf("error unable to get ILB subnet source ranges: %v", err)
+	}
+
 	// Ensure firewall rule for the cluster and pass any NEG endpoint ports.
-	if err := fwc.firewallPool.Sync(nodeNames, negPorts...); err != nil {
+	if err := fwc.firewallPool.Sync(nodeNames, negPorts, ILBSrcRange); err != nil {
```
rramkumar1: So IIUC this will add the designated subnet range to the firewall rule even if the user does not actually create an Ingress with the ILB class? I'm not sure that's the behavior we want. 

Instead I think you should only add the subnet range if there is an Ingress with the proper class.

pkg/firewalls/firewalls.go
```diff
@@ -63,8 +63,8 @@ func NewFirewallPool(cloud Firewall, namer *utils.Namer, l7SrcRanges []string, n
 	}
 }
 
-// Sync sync firewall rules with the cloud.
-func (fr *FirewallRules) Sync(nodeNames []string, additionalPorts ...string) error {
+// Sync firewall rules with the cloud.
+func (fr *FirewallRules) Sync(nodeNames, additionalPorts []string, ILBSrcRange string) error {
```
rramkumar1: Instead of calling it ILBSrcRange, maybe call it additionalRanges?

pkg/backends/features/features.go
```diff
@@ -33,13 +33,15 @@ const (
...
+	// FeatureILB defines the feature name of the l7 internal load balancer
+	FeatureILB = "ILB"
 )
 
 var (
 	// versionToFeatures stores the mapping from the required API
 	// version to feature names.
 	versionToFeatures = map[meta.Version][]string{
-		meta.VersionAlpha: []string{},
+		meta.VersionAlpha: []string{FeatureILB},
```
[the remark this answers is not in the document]
spencerhance: We definitely should if L7-ILB goes beta before this merges, but otherwise I think we have to use alpha for now.

pkg/backends/backends.go
```diff
@@ -14,10 +14,11 @@ limitations under the License.
 package backends
 
 import (
+	features2 "k8s.io/ingress-gce/pkg/loadbalancers/features"
```
bowei: features2 is non-meaningful -- change to `lbfeatures`

pkg/backends/backends.go
```diff
@@ -27,20 +28,22 @@ import (
 
 // Backends handles CRUD operations for backends.
 type Backends struct {
-	cloud *gce.Cloud
-	namer *utils.Namer
+	cloud          *gce.Cloud
+	namer          *utils.Namer
+	compositeCloud *composite.Cloud
```
bowei: minor: seems better to put cloud, compositeCloud fields next to each other

pkg/composite/cloud.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return &Cloud{}
+}
+
+func (c *Cloud) GceCloud() *gce.Cloud {
+	return c.gceCloud
+}
+
+func (c *Cloud) CreateKey(name string, regional bool) *meta.Key {
```
bowei: This function is going to be an accident waiting to happen.
Is there a reason why `region == ""` and it is regional?

It's better to have `func CreateRegionalKey(name) (*meta.Key, error)` and push the logic up the callstack.

pkg/backends/neg_linker.go
```diff
@@ -58,9 +61,19 @@ func (l *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 		negs = append(negs, neg)
 	}
 
-	cloud := l.backendPool.(*Backends).cloud
+	//cloud := l.backendPool.(*Backends).cloud
```
bowei: you need to remove all commented out code.

pkg/backends/neg_linker.go
```diff
@@ -58,9 +61,19 @@ func (l *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
...
-	backendService, err := cloud.GetGlobalBackendService(beName)
+
+	var version meta.Version
+	if sp.ILBEnabled {
+		version = features.ILBVersion
+	} else {
+		version = meta.VersionGA
+	}
+
+	key := compositeCloud.CreateKey(beName, sp.ILBEnabled)
```
bowei: How does this work if ILB is enabled but the user is not creating an ILB? Doesn't this make that code path fetch the regional construct?
See comment below on CreateKey func.

pkg/backends/neg_linker.go
```diff
@@ -80,15 +93,15 @@ func (l *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 
 	if !oldBackends.Equal(newBackends) {
 		backendService.Backends = targetBackends
-		return cloud.UpdateGlobalBackendService(backendService)
+		return compositeCloud.UpdateBackendService(backendService, key)
```
bowei: Usually the update funcs take the key as the first argument.
This should be UpdateBackendService(key, backendService)

pkg/backends/syncer.go
```diff
@@ -74,23 +76,24 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
 	beName := sp.BackendName(s.namer)
 	version := features.VersionFromServicePort(&sp)
 
-	be, getErr := s.backendPool.Get(beName, version)
+	be, getErr := s.backendPool.Get(beName, version, sp.ILBEnabled)
```
bowei: We should make the Zone a type in gen stuff, e.g.

```
type ZoneType string
 const (
    // Zonal key type.
    Zonal ZoneType = "zonal"
...
 )
``` 

and every place you have regional as a bool, it should be replaced by `ZoneType`. The bool is gone to be a maintenance issue, not to mention it is not very self-documenting.

pkg/backends/syncer.go
```diff
@@ -136,33 +139,41 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
 	knownPorts := sets.NewString()
 	for _, sp := range svcPorts {
 		name := sp.BackendName(s.namer)
-		knownPorts.Insert(name)
+		knownPorts.Insert(name + "-" + strconv.FormatBool(sp.ILBEnabled))
```
bowei: this needs to go into namer.
it needs to use truncation logic if field is length limited
we should NOT have names like "xyz-true" etc, it is non-extensible and quite confusing.

pkg/backends/syncer.go
```diff
@@ -136,33 +139,41 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
...
 
-	for _, name := range backendNames {
-		if knownPorts.Has(name) {
+	for _, be := range backends {
+		name := be.Name
+		regional, err := composite.IsRegionalResource(be.SelfLink)
+		if err != nil {
+			return err
+		}
+		if knownPorts.Has(name + "-" + strconv.FormatBool(regional)) {
```
bowei: this should not be cut and paste -- naming schemes need to be in one place (namer)

pkg/backends/syncer.go
```diff
@@ -173,15 +184,16 @@ func (s *backendSyncer) Shutdown() error {
 	return nil
 }
 
-func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort, hasLegacyHC bool) (string, error) {
+func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort, hasLegacyHC bool, isInternalLB bool) (string, error) {
 	if hasLegacyHC {
 		klog.Errorf("Backend %+v has legacy health check", sp.ID)
 	}
+	//klog.V(3).Infof("ensuring health check: %v, hasLegacyHC=%v, isInternalLB=%v", pretty.Sprint(sp), hasLegacyHC, isInternalLB)
```
bowei: delete

pkg/backends/syncer.go
```diff
@@ -173,15 +184,16 @@ func (s *backendSyncer) Shutdown() error {
 	return nil
 }
 
-func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort, hasLegacyHC bool) (string, error) {
+func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort, hasLegacyHC bool, isInternalLB bool) (string, error) {
```
bowei: is `isInternalLB` used anywhere?

pkg/backends/ig_linker.go
```diff
@@ -93,7 +93,7 @@ func (l *instanceGroupLinker) Link(sp utils.ServicePort, groups []GroupKey) erro
 		igLinks = append(igLinks, ig.SelfLink)
 	}
 
-	be, err := l.backendPool.Get(sp.BackendName(l.namer), meta.VersionGA)
+	be, err := l.backendPool.Get(sp.BackendName(l.namer), meta.VersionGA, false)
```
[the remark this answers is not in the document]
spencerhance: This is because IGs are not supported for L7-ILB

pkg/loadbalancers/features/ilb.go
spencerhance: Will do, I think it still makes sense to call the Compute() API directly though since we don't currently have a composite type for subnets.

pkg/composite/cloud.go
spencerhance: So the reason for this function is to easily be able to pass a key to the composite functions since it only has one return value.  Since the region is grabbed from gceCloud which in turn gets it from gce.conf, there could be a case where the region is empty.  I can get rid of the check though since the result of the function won't actually change either way.  Also I can return an error and just make it a two line call instead of one.

pkg/backends/syncer.go
spencerhance: So this is not a naming scheme in the traditional sense.  This is just a localized convention only in this function to still be able to use a set of strings instead of having to look through all of the service ports.

pkg/backends/syncer.go
spencerhance: Same as above.  I can implement different logic or create a service port namer, but it's really just some helper code for quick comparisons.

pkg/backends/syncer.go
spencerhance: currently i've been using a 
```golang
var resourceType meta.KeyType
``` 
everywhere i'm not using a bool since that covers zonal, regional, and global.  Should I just keep using that?  It's also a string implementation.

pkg/backends/neg_linker.go
spencerhance: This is really a misuse of the word "enabled."  That flag only gets updated if the user specifies an l7-ilb.

pkg/loadbalancers/features/ilb.go
spencerhance: done

pkg/backends/syncer.go

pkg/backends/syncer.go
spencerhance: Must have left that in by mistake. Done.

pkg/backends/neg_linker.go
spencerhance: I can update this along with the resolution of the other boolean flags.

pkg/backends/backends.go
spencerhance: Done

pkg/backends/backends.go
spencerhance: Whoops, Done.

pkg/firewalls/controller.go
spencerhance: Good catch.  Done

pkg/firewalls/firewalls.go
spencerhance: Done

pkg/composite/gen/main.go
```diff
@@ -209,42 +209,46 @@ func genFuncs(wr io.Writer) {
 		if err != nil {
 			return err
 		}
-		klog.V(3).Infof("Creating alpha {{.Name}} %v", alpha.Name)
 		switch key.Type() {
 		case meta.Regional:
-			return mc.Observe(cloud.Compute().Alpha{{.GetCloudProviderName}}().Insert(ctx, key, alpha))
+			klog.V(3).Infof("Creating alpha region {{.Name}} %v", alpha.Name)
```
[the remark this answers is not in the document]
spencerhance: not sure I understand, it looks like all of the api versions have their own klog statement.

pkg/backends/syncer.go
spencerhance: Now that I think about it, I can probably just use meta.Key since it's hashable and more comprehensive

spencerhance:
closing in favor of #789

PR #780, opened 2019-06-19
Update composites and convert to cloud struct

spencerhance:
extracted from: https://github.com/kubernetes/ingress-gce/pull/773/

### DIFF
--- a/pkg/composite/metrics/metrics_test.go
+++ b/pkg/composite/metrics/metrics_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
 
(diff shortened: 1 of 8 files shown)


pkg/composite/metrics/metrics_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
```
spencerhance: Fix this

spencerhance:
The main question here is do we want to keep composite as a static library

spencerhance:
Closing since we removed the LB interface.  I'll send another PR with just the necessary composite updates (e.g. without composite cloud)

PR #340, opened 2018-06-15
Use generated mocks for load balancer unit tests [WIP]

rramkumar1:
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/340)

PR #781, opened 2019-06-21
Remove Loadbalancer interface and use k8s-cloud-provider mocks

spencerhance:
This requires some new common hooks in k8s-cloud-provider, PR: https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/16 and pulling in those changes to vendor

Used https://github.com/kubernetes/ingress-gce/pull/340 as a reference.


pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -46,6 +53,79 @@ var (
...
+	pool    LoadBalancerPool
+	fakeGCE *gce.Cloud
+	namer   *utils.Namer
+	t       *testing.T
+}
+
+func newTestJig(t *testing.T) *testJig {
+	namer := utils.NewNamer(clusterName, "fw1")
+	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
+
```
bowei: mock := fakeGCE.Compute().(*cloud.MockGCE)

makes the code below less verbose

bowei: or better yet, keep a pointer to mock in testJig (e.g. `j.mock`) as it is used all the time

rramkumar1:
@spencerhance do you mind running a code coverage tool before & after just to verify there was no loss in coverage? It's not really a foolproof method to sanity check we are not losing coverage but I can't think of anything else.

pkg/loadbalancers/loadbalancers_test.go
spencerhance: done

spencerhance:
@rramkumar1 `go test -cover` reports pkg/loadbalancers went from 75.9% to 74.1%

bowei:
need to revert until the vendor stuff is fixed

PR #784, opened 2019-06-24
Redo "Remove Loadbalancer interface and use k8s-cloud-provider mocks"

spencerhance:
Redo of #781 now that vendor can be updated


spencerhance:
@bowei @rramkumar1 Fixed the vendor update so this is good to go now

rramkumar1:
/;gtm

Issue #767, opened 2019-06-11
NEG controller should create NEG for default backend when enabled


freehan:
NEG controller currently ignores default backend service even it has NEG enabled. This is because default backend service is not explicitly referenced in the ingress spec.

https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/controller.go#L309

rafi-greenberg:
@freehan  Just curious - does #789  only apply to ILB?

It seems that all our NEG-enabled services create default backends as instance groups.

![image](https://user-images.githubusercontent.com/3811209/90044484-8d239500-dc82-11ea-8188-c6588bbb1882.png)

spencerhance:
@rafi-greenberg right now this only applies to ILB.  @freehan are we planning on supporting this for ELB with neg default?

PR #789, opened 2019-06-29
Add support for L7-ILB (take 2)

spencerhance:
**This PR supersedes https://github.com/kubernetes/ingress-gce/pull/773**

Adds support for the L7 Internal Load Balancer to the controller.  A follow up PR with e2e tests will be added.

In the interest of time, I've added the L7-ILB work on top of https://github.com/kubernetes/ingress-gce/pull/788 so that it can be reviewed in tandem.  Unless you foresee *gigantic* changes in that PR, I would review this as well so it can be merged shortly after.  The first commit on this PR is from that branch.

Currently, the feature is flag-gated with the '--enable-l7-ilb' flag.  Enabling the flag will require resolving https://github.com/kubernetes/ingress-gce/issues/767

### DIFF
--- a/pkg/annotations/ingress.go
+++ b/pkg/annotations/ingress.go
@@ -54,6 +54,7 @@ const (
 	IngressClassKey      = "kubernetes.io/ingress.class"
 	GceIngressClass      = "gce"
 	GceMultiIngressClass = "gce-multi-cluster"
+	GceL7ILBIngressClass = "gce-internal"
 
 	// Label key to denote which GCE zone a Kubernetes node is in.
 	ZoneKey     = "failure-domain.beta.kubernetes.io/zone"
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -14,6 +14,8 @@ limitations under the License.
 package backends
 
 import (
+	"fmt"
+	"k8s.io/ingress-gce/pkg/flags"
 	"net/http"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -73,13 +75,18 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 		PortName:     namedPort.Name,
 		HealthChecks: []string{hcLink},
 	}
-	ensureDescription(be, &sp)
 
+	if sp.L7ILBEnabled {
+		be.LoadBalancingScheme = "INTERNAL"
+	}
+
+	ensureDescription(be, &sp)
 	scope := features.ScopeFromServicePort(&sp)
 	key, err := composite.CreateKey(b.cloud, name, scope)
 	if err != nil {
 		return nil, err
 	}
+
 	if err := composite.CreateBackendService(b.cloud, key, be); err != nil {
 		return nil, err
 	}
@@ -161,10 +168,10 @@ func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType)
 }
 
 // Health implements Pool.
-func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType) string {
+func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType) (string, error) {
 	be, err := b.Get(name, version, scope)
 	if err != nil || len(be.Backends) == 0 {
-		return "Unknown"
+		return "Unknown", fmt.Errorf("error getting health for backend %s: %v", name, err)
 	}
 
 	// TODO: Look at more than one backend's status
@@ -177,21 +184,28 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 	case meta.Regional:
 		hs, err = b.cloud.GetRegionalBackendServiceHealth(name, b.cloud.Region(), be.Backends[0].Group)
 	default:
-		return "Unknown"
+		return "Unknown", fmt.Errorf("invalid scope for Health(): %s", scope)
 	}
 
 	if err != nil || len(hs.HealthStatus) == 0 || hs.HealthStatus[0] == nil {
-		return "Unknown"
+		return "Unknown", fmt.Errorf("error getting health for backend %q: %v", name, err)
 	}
 	// TODO: State transition are important, not just the latest.
-	return hs.HealthStatus[0].HealthState
+	return hs.HealthStatus[0].HealthState, nil
 }
 
 // List lists all backends managed by this controller.
 func (b *Backends) List() ([]*composite.BackendService, error) {
 	// TODO: for consistency with the rest of this sub-package this method
 	// should return a list of backend ports.
-	backends, err := composite.ListAllBackendServices(b.cloud)
+	var backends []*composite.BackendService
+	var err error
+	if flags.F.EnableL7Ilb {
+		backends, err = composite.ListAllBackendServices(b.cloud)
+	} else {
+		// TODO: (shance) this needs to be changed to not take a key
+		backends, err = composite.ListBackendServices(b.cloud, meta.GlobalKey(""), meta.VersionGA)
+	}
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/backends/features/features.go
+++ b/pkg/backends/features/features.go
@@ -33,19 +33,22 @@ const (
 	FeatureSecurityPolicy = "SecurityPolicy"
 	// FeatureNEG defines the feature name of NEG.
 	FeatureNEG = "NEG"
+	// FeatureL7ILB defines the feature name of L7 Internal Load Balancer
+	// L7-ILB Resources are currently alpha and regional
+	FeatureL7ILB = "L7ILB"
 )
 
 var (
 	// versionToFeatures stores the mapping from the required API
 	// version to feature names.
 	versionToFeatures = map[meta.Version][]string{
-		meta.VersionBeta: []string{FeatureSecurityPolicy, FeatureHTTP2},
+		meta.VersionAlpha: []string{FeatureL7ILB},
+		meta.VersionBeta:  []string{FeatureSecurityPolicy, FeatureHTTP2},
 	}
-
 	// TODO: (shance) refactor all scope to be above the serviceport level
-	// scopeToFeatures stores the mapping from the required meta.KeyType
-	// to feature names
-	scopeToFeatures = map[meta.KeyType][]string{}
+	scopeToFeatures = map[meta.KeyType][]string{
+		meta.Regional: []string{FeatureL7ILB},
+	}
 )
 
 // SetDescription sets the XFeatures field for the given Description.
@@ -65,6 +68,9 @@ func featuresFromServicePort(sp *utils.ServicePort) []string {
 	}
 	if sp.NEGEnabled {
 		features = append(features, FeatureNEG)
+	}
+	if sp.L7ILBEnabled {
+		features = append(features, FeatureL7ILB)
 	}
 	// Keep feature names sorted to be consistent.
 	sort.Strings(features)
--- a/pkg/backends/ig_linker.go
+++ b/pkg/backends/ig_linker.go
@@ -94,6 +94,8 @@ func (l *instanceGroupLinker) Link(sp utils.ServicePort, groups []GroupKey) erro
 	}
 
 	// ig_linker only supports L7 HTTP(s) External Load Balancer
+	// Hardcoded here since IGs are not supported for non GA-Global right now
+	// TODO(shance): find a way to remove hardcoded values
 	be, err := l.backendPool.Get(sp.BackendName(l.namer), meta.VersionGA, meta.Global)
 	if err != nil {
 		return err
--- a/pkg/backends/syncer.go
+++ b/pkg/backends/syncer.go
@@ -65,6 +65,7 @@ func (s *backendSyncer) Sync(svcPorts []utils.ServicePort) error {
 		}
 	}
 	return nil
+
 }
 
 // ensureBackendService will update or create a BackendService for the given port.
@@ -81,7 +82,6 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
 	if be != nil {
 		// If the backend already exists, find out if it is using a legacy health check.
 		existingHCLink := getHealthCheckLink(be)
-		klog.V(3).Infof("Got exisiting HCLink: %v", existingHCLink)
 		if strings.Contains(existingHCLink, "/httpHealthChecks/") {
 			hasLegacyHC = true
 		}
@@ -175,7 +175,7 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
 			return err
 		}
 
-		if err := s.healthChecker.Delete(name); err != nil {
+		if err := s.healthChecker.Delete(name, scope); err != nil {
 			return err
 		}
 	}
@@ -199,7 +199,7 @@ func knownPortsFromServicePorts(cloud *gce.Cloud, namer *utils.Namer, svcPorts [
 }
 
 // Status implements Syncer.
-func (s *backendSyncer) Status(name string, version meta.Version, scope meta.KeyType) string {
+func (s *backendSyncer) Status(name string, version meta.Version, scope meta.KeyType) (string, error) {
 	return s.backendPool.Health(name, version, scope)
 }
 
@@ -219,7 +219,7 @@ func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort, hasLegacyHC bool
 	if s.prober != nil {
 		probe, err := s.prober.GetProbe(sp)
 		if err != nil {
-			return "", err
+			return "", fmt.Errorf("Error getting prober: %v", err)
 		}
 		if probe != nil {
 			klog.V(4).Infof("Applying httpGet settings of readinessProbe to health check on port %+v", sp)
--- a/pkg/backends/syncer_test.go
+++ b/pkg/backends/syncer_test.go
@@ -100,7 +100,7 @@ func TestSync(t *testing.T) {
 			beName := sp.BackendName(defaultNamer)
 
 			// Check that the new backend has the right port
-			be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&sp), meta.Global)
+			be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&sp), features.ScopeFromServicePort(&sp))
 			if err != nil {
 				t.Fatalf("Did not find expected backend with port %v", sp.NodePort)
 			}
@@ -108,7 +108,7 @@ func TestSync(t *testing.T) {
 				t.Fatalf("Backend %v has wrong port %v, expected %v", be.Name, be.Port, sp)
 			}
 
-			hc, err := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&sp))
+			hc, err := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&sp), features.ScopeFromServicePort(&sp))
 			if err != nil {
 				t.Fatalf("Unexpected err when querying fake healthchecker: %v", err)
 			}
@@ -132,7 +132,7 @@ func TestSyncUpdateHTTPS(t *testing.T) {
 	syncer.Sync([]utils.ServicePort{p})
 	beName := p.BackendName(defaultNamer)
 
-	be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), meta.Global)
+	be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if err != nil {
 		t.Fatalf("Unexpected err: %v", err)
 	}
@@ -142,7 +142,7 @@ func TestSyncUpdateHTTPS(t *testing.T) {
 	}
 
 	// Assert the proper health check was created
-	hc, _ := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p))
+	hc, _ := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if hc == nil || hc.Protocol() != p.Protocol {
 		t.Fatalf("Expected %s health check, received %v: ", p.Protocol, hc)
 	}
@@ -151,7 +151,7 @@ func TestSyncUpdateHTTPS(t *testing.T) {
 	p.Protocol = annotations.ProtocolHTTPS
 	syncer.Sync([]utils.ServicePort{p})
 
-	be, err = syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), meta.Global)
+	be, err = syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if err != nil {
 		t.Fatalf("Unexpected err retrieving backend service after update: %v", err)
 	}
@@ -162,7 +162,7 @@ func TestSyncUpdateHTTPS(t *testing.T) {
 	}
 
 	// Assert the proper health check was created
-	hc, _ = syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p))
+	hc, _ = syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if hc == nil || hc.Protocol() != p.Protocol {
 		t.Fatalf("Expected %s health check, received %v: ", p.Protocol, hc)
 	}
@@ -176,7 +176,7 @@ func TestSyncUpdateHTTP2(t *testing.T) {
 	syncer.Sync([]utils.ServicePort{p})
 	beName := p.BackendName(defaultNamer)
 
-	be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), meta.Global)
+	be, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if err != nil {
 		t.Fatalf("Unexpected err: %v", err)
 	}
@@ -186,7 +186,7 @@ func TestSyncUpdateHTTP2(t *testing.T) {
 	}
 
 	// Assert the proper health check was created
-	hc, _ := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p))
+	hc, _ := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if hc == nil || hc.Protocol() != p.Protocol {
 		t.Fatalf("Expected %s health check, received %v: ", p.Protocol, hc)
 	}
@@ -195,7 +195,7 @@ func TestSyncUpdateHTTP2(t *testing.T) {
 	p.Protocol = annotations.ProtocolHTTP2
 	syncer.Sync([]utils.ServicePort{p})
 
-	beBeta, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), meta.Global)
+	beBeta, err := syncer.backendPool.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if err != nil {
 		t.Fatalf("Unexpected err retrieving backend service after update: %v", err)
 	}
@@ -206,7 +206,7 @@ func TestSyncUpdateHTTP2(t *testing.T) {
 	}
 
 	// Assert the proper health check was created
-	hc, _ = syncer.healthChecker.Get(beName, meta.VersionAlpha)
+	hc, _ = syncer.healthChecker.Get(beName, features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if hc == nil || hc.Protocol() != p.Protocol {
 		t.Fatalf("Expected %s health check, received %v: ", p.Protocol, hc)
 	}
@@ -418,7 +418,7 @@ func TestSyncNEG(t *testing.T) {
 	// GC should garbage collect the Backend on the old naming schema
 	syncer.GC([]utils.ServicePort{svcPort})
 
-	bs, err := syncer.backendPool.Get(nodePortName, features.VersionFromServicePort(&svcPort), meta.Global)
+	bs, err := syncer.backendPool.Get(nodePortName, features.VersionFromServicePort(&svcPort), features.ScopeFromServicePort(&svcPort))
 	if err == nil {
 		t.Fatalf("Expected not to get BackendService with name %v, got: %+v", nodePortName, bs)
 	}
@@ -491,7 +491,7 @@ func TestEnsureBackendServiceProtocol(t *testing.T) {
 				fmt.Sprintf("Updating Port:%v Protocol:%v to Port:%v Protocol:%v", oldPort.NodePort, oldPort.Protocol, newPort.NodePort, newPort.Protocol),
 				func(t *testing.T) {
 					syncer.Sync([]utils.ServicePort{oldPort})
-					be, err := syncer.backendPool.Get(oldPort.BackendName(defaultNamer), features.VersionFromServicePort(&oldPort), meta.Global)
+					be, err := syncer.backendPool.Get(oldPort.BackendName(defaultNamer), features.VersionFromServicePort(&oldPort), features.ScopeFromServicePort(&oldPort))
 					if err != nil {
 						t.Fatalf("%v", err)
 					}
@@ -535,7 +535,7 @@ func TestEnsureBackendServiceDescription(t *testing.T) {
 				fmt.Sprintf("Updating Port:%v Protocol:%v to Port:%v Protocol:%v", oldPort.NodePort, oldPort.Protocol, newPort.NodePort, newPort.Protocol),
 				func(t *testing.T) {
 					syncer.Sync([]utils.ServicePort{oldPort})
-					be, err := syncer.backendPool.Get(oldPort.BackendName(defaultNamer), features.VersionFromServicePort(&oldPort), meta.Global)
+					be, err := syncer.backendPool.Get(oldPort.BackendName(defaultNamer), features.VersionFromServicePort(&oldPort), features.ScopeFromServicePort(&oldPort))
 					if err != nil {
 						t.Fatalf("%v", err)
 					}
@@ -563,7 +563,7 @@ func TestEnsureBackendServiceHealthCheckLink(t *testing.T) {
 
 	p := utils.ServicePort{NodePort: 80, Protocol: annotations.ProtocolHTTP, ID: utils.ServicePortID{Port: intstr.FromInt(1)}}
 	syncer.Sync([]utils.ServicePort{p})
-	be, err := syncer.backendPool.Get(p.BackendName(defaultNamer), features.VersionFromServicePort(&p), meta.Global)
+	be, err := syncer.backendPool.Get(p.BackendName(defaultNamer), features.VersionFromServicePort(&p), features.ScopeFromServicePort(&p))
 	if err != nil {
 		t.Fatalf("%v", err)
 	}
--- a/pkg/composite/utils.go
+++ b/pkg/composite/utils.go
@@ -47,22 +47,34 @@ func CreateKey(gceCloud *gce.Cloud, name string, scope meta.KeyType) (*meta.Key,
 }
 
 // TODO: (shance) generate this
-// TODO: (shance) add regional and alpha
 // TODO: (shance) populate scope
-// ListAllUrlMaps() merges all configured List() calls into one list of composite UrlMaps
+// TODO: (shance) figure out a better way to get version
+// ListAllUrlMaps() merges all configured List() calls into one list of composite UrlMaps.
+// This function combines both global and regional resources into one slice
+// so users must use ScopeFromSelfLink() to determine the correct scope
+// before issuing an API call.
 func ListAllUrlMaps(gceCloud *gce.Cloud) ([]*UrlMap, error) {
 	resultMap := map[string]*UrlMap{}
-	key1, err := CreateKey(gceCloud, "", meta.Global)
+
+	// TODO(shance): this needs to be replaced with 'scope' and not key
+	globalKey, err := CreateKey(gceCloud, "", meta.Global)
+	if err != nil {
+		return nil, err
+	}
+	regionalKey, err := CreateKey(gceCloud, "", meta.Regional)
 	if err != nil {
 		return nil, err
 	}
 
-	// List ga-global and regional-alpha
-	versions := []meta.Version{meta.VersionGA}
-	keys := []*meta.Key{key1}
+	for _, vk := range []struct {
+		v meta.Version
+		k *meta.Key //TODO(shance): change this to scope
+	}{
+		{meta.VersionGA, globalKey},
+		{meta.VersionAlpha, regionalKey},
+	} {
 
-	for i := range versions {
-		list, err := ListUrlMaps(gceCloud, keys[i], versions[i])
+		list, err := ListUrlMaps(gceCloud, vk.k, vk.v)
 		if err != nil {
 			return nil, fmt.Errorf("error listing all urlmaps: %v", err)
 		}
@@ -81,19 +93,26 @@ func ListAllUrlMaps(gceCloud *gce.Cloud) ([]*UrlMap, error) {
 }
 
 // TODO: (shance) generate this
-// TODO: (shance) add regional and alpha
 // TODO: (shance) populate scope
+// TODO: (shance) figure out a more accurate way to obtain version
 // ListAllUrlMaps() merges all configured List() calls into one list of composite UrlMaps
+// This function combines both global and regional resources into one slice
+// so users must use ScopeFromSelfLink() to determine the correct scope
+// before issuing an API call.
 func ListAllBackendServices(gceCloud *gce.Cloud) ([]*BackendService, error) {
 	resultMap := map[string]*BackendService{}
 	key1, err := CreateKey(gceCloud, "", meta.Global)
 	if err != nil {
 		return nil, err
 	}
+	key2, err := CreateKey(gceCloud, "", meta.Regional)
+	if err != nil {
+		return nil, err
+	}
 
 	// List ga-global and regional-alpha
-	versions := []meta.Version{meta.VersionGA}
-	keys := []*meta.Key{key1}
+	versions := []meta.Version{meta.VersionGA, meta.VersionAlpha}
+	keys := []*meta.Key{key1, key2}
 
 	for i := range versions {
 		list, err := ListBackendServices(gceCloud, keys[i], versions[i])
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -308,6 +308,12 @@ func (lbc *LoadBalancerController) SyncBackends(state interface{}) error {
 	}
 	ingSvcPorts := syncState.urlMap.AllServicePorts()
 
+	if utils.IsGCEL7ILBIngress(syncState.ing) {
+		if err := UpdateServicePortsForILB(ingSvcPorts, syncState.ing); err != nil {
+			return err
+		}
+	}
+
 	// Create instance groups and set named ports.
 	igs, err := lbc.instancePool.EnsureInstanceGroupsAndPorts(lbc.ctx.ClusterNamer.InstanceGroup(), nodePorts(ingSvcPorts))
 	if err != nil {
@@ -639,7 +645,11 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
 		urlMap, _ := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPortID)
-		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		svcPorts := urlMap.AllServicePorts()
+		if utils.IsGCEL7ILBIngress(ing) {
+			UpdateServicePortsForILB(svcPorts, ing)
+		}
+		knownPorts = append(knownPorts, svcPorts...)
 	}
 	return knownPorts
 }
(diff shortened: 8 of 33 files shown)


rramkumar1:
@spencerhance Can you squash commits so that it's easier to see which commits are from the parent PR?

spencerhance:
@rramkumar1 good call, should be good now

pkg/backends/syncer.go
```diff
@@ -170,7 +170,7 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
 		if err := s.backendPool.Delete(name, be.Version, resourceType); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 			return err
 		}
-		if err := s.healthChecker.Delete(name); err != nil {
+		if err := s.healthChecker.Delete(name, resourceType == meta.Regional); err != nil {
```
freehan: Probably better align this with the composite pattern?  What if there is a zonal health check in the feature, although unlikely?

pkg/controller/utils.go
```diff
@@ -88,3 +88,15 @@ func nodePorts(svcPorts []utils.ServicePort) []int64 {
 	}
 	return ports
 }
+
+func UpdateServicePortsForILB(ingSvcPorts []utils.ServicePort, ing *extensions.Ingress) error {
```
freehan: I feel like there is a better way to do this. You should probably fail at validation if there is service backend that does not have NEG enabled.

pkg/firewalls/firewalls.go
```diff
@@ -78,6 +78,7 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts []string) error {
 	sort.Strings(targetTags)
 
 	ports := sets.NewString(additionalPorts...)
+	fr.srcRanges = append(fr.srcRanges, additionalRanges...)
```
freehan: do we want to have a separate firewall rule for ILB?

spencerhance: it's not a separate firewall rule, rather adding the ILB range to the firewall rule since that is a requirement.

pkg/controller/utils.go
spencerhance: True, but I do need to also handle the default backend case

pkg/backends/syncer.go

pkg/firewalls/firewalls.go
spencerhance: Oh I think I misunderstood your question.  It's possible to add another rule but I'm not sure if that would make much of a difference.

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +258,50 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
 	return newHC
 }
 
-func (h *HealthChecks) getHealthCheckLink(name string, version meta.Version) (string, error) {
-	hc, err := h.Get(name, version)
+func (h *HealthChecks) getHealthCheckLink(name string, version meta.Version, regional bool) (string, error) {
```
bowei: scope should not be a bool

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +258,50 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
+	hc, err := h.Get(name, version, regional)
 	if err != nil {
 		return "", err
 	}
 	return hc.SelfLink, nil
 }
 
 // Delete deletes the health check by port.
-func (h *HealthChecks) Delete(name string) error {
+func (h *HealthChecks) Delete(name string, regional bool) error {
```
bowei: scope should not be a bool

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +258,50 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
+	if regional {
+		cloud := h.cloud.(*gce.Cloud)
+		key, err := composite.CreateKey(cloud, name, features.L7ILBResourceType())
+		if err != nil {
+			return err
+		}
+		return composite.DeleteHealthCheck(cloud, key, features.L7ILBVersion())
+	}
+
```
bowei: why does this code have a branch involving composite and one that does not use composite?

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +258,50 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
 }
 
 // Get returns the health check by port
-func (h *HealthChecks) Get(name string, version meta.Version) (*HealthCheck, error) {
+func (h *HealthChecks) Get(name string, version meta.Version, regional bool) (*HealthCheck, error) {
+	klog.V(3).Infof("Getting Health Check, name: %s, version: %v, regional: %v", name, version, regional)
+
+	if regional {
+		cloud := h.cloud.(*gce.Cloud)
+		key, err := composite.CreateKey(cloud, name, features.L7ILBResourceType())
```
bowei: This is going to be very confusing for someone maintaining the code. You are already in a branch for `regional`,
yet you call `feature.L7ILBResourceType()` to get the scope for the feature, which creates a dependency
between these two pieces of code.

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +258,50 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
+		if err != nil {
+			return nil, err
+		}
+		hc, err := composite.GetHealthCheck(cloud, key, features.L7ILBVersion())
```
bowei: Same as comment above --
This branch is for `if regional`, but it seems to assume everything is for ILB (getting `features.L7ILBVersion()`) here.
For example, if we implement network tiers, then non-ILB code will also create regional constructs and the code will need
to be reworked at that point in time.

pkg/healthchecks/healthchecks.go
```diff
@@ -299,6 +381,31 @@ func DefaultNEGHealthCheck(protocol annotations.AppProtocol) *HealthCheck {
 	}
 }
 
+func DefaultILBHealthCheck(protocol annotations.AppProtocol, region string) *HealthCheck {
```
bowei: why is this exported?

bowei: why not take in a key instead of region?
also, I don't think `region` is actually used in this function

pkg/healthchecks/healthchecks.go
```diff
@@ -307,14 +414,19 @@ func DefaultNEGHealthCheck(protocol annotations.AppProtocol) *HealthCheck {
 type HealthCheck struct {
 	computealpha.HTTPHealthCheck
 	computealpha.HealthCheck
+	//compositeHealthCheck composite.HealthCheck
```
bowei: fix

pkg/healthchecks/healthchecks.go
```diff
@@ -307,14 +414,19 @@ func DefaultNEGHealthCheck(protocol annotations.AppProtocol) *HealthCheck {
...
 	ForNEG bool
+	// ForILB designates whether the health check is for an ILB
+	// This means that the HC should be alpha + regional
```
bowei: this comment will quickly be stale -- remove

pkg/healthchecks/healthchecks.go
```diff
@@ -404,8 +516,10 @@ func (hc *HealthCheck) isHttp2() bool {
 // Version returns the appropriate API version to handle the health check
 // Use Beta API for NEG as PORT_SPECIFICATION is required, and HTTP2
 func (hc *HealthCheck) Version() meta.Version {
-	if hc.isHttp2() || hc.ForNEG {
+	if hc.isHttp2() || (hc.ForNEG && !hc.ForILB) {
```
bowei: this is a little weird

this means isHttp2 will force Beta even if used with ILB?

bowei: I think we should have a tiny unit test for this logic

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error obtaining subnets for region: %s", region)
+	}
+
+	for _, subnet := range subnets {
+		if subnet.Role == "ACTIVE" && subnet.Purpose == "INTERNAL_HTTPS_LOAD_BALANCER" {
+			return subnet.IpCidrRange, nil
+		}
+	}
+	return "", nil
```
bowei: I think this should be a not found error, otherwise it will be very easy to end up sending an empty string as an argument everywhere

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// L7ILBVersion is a helper to get the version of L7-ILB
```
bowei: This will likely become an issue as ILB does not necessarily mean every resource is @ this version.

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+func L7ILBVersion() meta.Version {
+	return VersionFromFeatures([]string{FeatureL7ILB})
+}
+
+// L7ILBResourceType is a helper to get the resource type of L7-ILB
```
bowei: scope?

pkg/utils/utils.go
```diff
@@ -278,9 +278,14 @@ func IsGCEMultiClusterIngress(ing *extensions.Ingress) bool {
 	return class == annotations.GceMultiIngressClass
 }
 
+func IsGCEL7ILBIngress(ing *extensions.Ingress) bool {
```
bowei: docstring

pkg/backends/backends.go
```diff
@@ -164,7 +168,11 @@ func (b *Backends) Health(name string, version meta.Version, resourceType meta.K
 	// TODO: Include port, ip in the status, since it's in the health info.
 	// TODO (shance) convert to composite types
 	var hs *compute.BackendServiceGroupHealth
-	hs, err = b.cloud.GetGlobalBackendServiceHealth(name, be.Backends[0].Group)
+	if resourceType == meta.Regional {
```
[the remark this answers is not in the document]
spencerhance: Will do, forgot to update this after the code review feedback from the initial PR

pkg/backends/syncer.go
spencerhance: This was fixed in last commit

pkg/firewalls/firewalls.go
spencerhance: Hmm, good point, I'll look at adding this today.

pkg/backends/syncer_test.go
```diff
@@ -106,7 +106,7 @@ func TestSync(t *testing.T) {
 				t.Fatalf("Backend %v has wrong port %v, expected %v", be.Name, be.Port, sp)
 			}
 
-			hc, err := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&sp))
+			hc, err := syncer.healthChecker.Get(beName, features.VersionFromServicePort(&sp), features.ResourceTypeFromServicePort(&sp) == meta.Regional)
```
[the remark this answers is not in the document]
spencerhance: Fixed in last commit

pkg/controller/controller.go
```diff
@@ -638,7 +642,9 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*extensions.Ingress) []util
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
 		urlMap, _ := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPortID)
-		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		svcPorts := urlMap.AllServicePorts()
+		UpdateServicePortsForILB(svcPorts, ing)
```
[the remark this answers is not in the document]
spencerhance: It's guarded inside the function.  I thought it makes the code more readable but if it's the opposite (which it appears so) I can change it.

pkg/firewalls/controller.go
```diff
@@ -123,7 +125,9 @@ func (fwc *FirewallController) ToSvcPorts(ings []*extensions.Ingress) []utils.Se
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
 		urlMap, _ := fwc.translator.TranslateIngress(ing, fwc.ctx.DefaultBackendSvcPortID)
-		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		svcPorts := urlMap.AllServicePorts()
+		controller.UpdateServicePortsForILB(svcPorts, ing)
```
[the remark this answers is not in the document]
spencerhance: ref above comment

pkg/firewalls/firewalls_test.go
```diff
@@ -38,7 +38,7 @@ func TestFirewallPoolSync(t *testing.T) {
 	fp := NewFirewallPool(fwp, namer, srcRanges, portRanges())
 	nodes := []string{"node-a", "node-b", "node-c"}
 
-	if err := fp.Sync(nodes, nil); err != nil {
+	if err := fp.Sync(nodes, nil, nil); err != nil {
```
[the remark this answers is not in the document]
spencerhance: I need to add a test here

pkg/healthchecks/healthchecks.go
```diff
@@ -163,6 +195,25 @@ func (h *HealthChecks) create(hc *HealthCheck) error {
 }
 
 func (h *HealthChecks) update(oldHC, newHC *HealthCheck) error {
+	// special case ILB to avoid mucking with stable HC code
+	//klog.V(3).Infof("Updating HC, old = %v, new = %v", pretty.Sprint(oldHC), pretty.Sprint(newHC))
+
+	if oldHC.ForILB {
```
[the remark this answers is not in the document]
spencerhance: will do.  Doesn't appear so, i'll unexport it.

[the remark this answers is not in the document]
spencerhance: Makes sense to me, i'll update that throughout.

pkg/healthchecks/healthchecks.go
spencerhance: Fixed in last commit

pkg/healthchecks/healthchecks.go
spencerhance: Fixed in last commit

pkg/healthchecks/healthchecks.go
spencerhance: alpha region and ga global health checks can't be deleted in one API call iirc, so we need a branch here.

pkg/healthchecks/healthchecks.go
spencerhance: Good call, i'll just hard code regional here.

pkg/healthchecks/healthchecks.go
spencerhance: Agreed, I'll look into generalizing this a bit more.

pkg/loadbalancers/features/l7ilb.go
spencerhance: True, but I don't think it makes sense to abstract that now unless they change it.  Otherwise we will need a duplicate feature for every single gce resource.

pkg/firewalls/firewalls.go
spencerhance: This looks like it will require non-trivial refactoring since we currently only support one firewall.  It's still possible, but it might make more sense to just update the src ranges separately to catch two different errors.

pkg/backends/backends.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/controller/controller.go
```diff
@@ -306,6 +306,10 @@ func (lbc *LoadBalancerController) SyncBackends(state interface{}) error {
 		return fmt.Errorf("expected state type to be syncState, type was %T", state)
 	}
 	ingSvcPorts := syncState.urlMap.AllServicePorts()
+	err := UpdateServicePortsForILB(ingSvcPorts, syncState.ing)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/controller/controller.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/controller.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/controller.go
```diff
@@ -164,8 +168,24 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 	negPorts := fwc.translator.GatherEndpointPorts(gceSvcPorts)
 
+	var additionalRanges []string
+	ILBEnabled := false
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/controller.go
```diff
@@ -164,8 +168,24 @@ func (fwc *FirewallController) sync(key string) error {
...
+	for _, sp := range gceSvcPorts {
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/firewalls_test.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/loadbalancers/features/l7ilb.go
spencerhance: done

pkg/utils/utils.go
spencerhance: done

pkg/loadbalancers/features/l7ilb.go
spencerhance: Done

pkg/healthchecks/healthchecks.go
spencerhance: Done

pkg/controller/utils.go
```diff
@@ -88,3 +88,15 @@ func nodePorts(svcPorts []utils.ServicePort) []int64 {
...
+	if !utils.IsGCEL7ILBIngress(ing) {
+		return nil
+	}
+
+	for svcPortIdx, _ := range ingSvcPorts {
+		ingSvcPorts[svcPortIdx].L7ILBEnabled = true
```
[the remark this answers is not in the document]
spencerhance: This might actually make things more complex as new features are added since we would need to plumb both version and scope, when all resources that are based on service ports already use the service port feature functions to determine those.  Basically, we would need to do what happens in Health check where every function has a forX variable for every new type of load balancer.

pkg/backends/features/features.go
```diff
@@ -33,15 +33,21 @@ const (
...
 
 var (
 	// versionToFeatures stores the mapping from the required API
 	// version to feature names.
 	versionToFeatures = map[meta.Version][]string{
-		meta.VersionBeta: []string{FeatureSecurityPolicy, FeatureHTTP2},
+		meta.VersionAlpha: []string{FeatureL7ILB},
+		meta.VersionBeta:  []string{FeatureSecurityPolicy, FeatureHTTP2},
+	}
+	resourceTypeToFeatures = map[meta.KeyType][]string{
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -139,6 +146,31 @@ func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
...
+		}
+
+		compositeType.Version = features.L7ILBVersion()
+		compositeType.Region = key.Region
+		err = composite.CreateHealthCheck(cloud, key, compositeType)
+		if err != nil {
+			return fmt.Errorf("Error creating health check %v: %v", compositeType, err)
+		}
+	}
+
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/interfaces.go
```diff
@@ -22,7 +22,7 @@ import (
 
 // SingleFirewallPool syncs the firewall rule for L7 traffic.
 type SingleFirewallPool interface {
-	Sync(nodeNames, additionalPorts []string) error
+	Sync(nodeNames, additionalPorts, additionalRanges []string) error
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -139,6 +146,31 @@ func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
 }
 
 func (h *HealthChecks) create(hc *HealthCheck) error {
+	// special case ILB to avoid mucking with stable HC code
+	if hc.ForILB {
+		cloud := h.cloud.(*gce.Cloud)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
spencerhance: This was because we can't test composites with the current fakes.  Here's a PR to remove them: https://github.com/kubernetes/ingress-gce/pull/796

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error obtaining subnets for region: %s", region)
+	}
+
+	for _, subnet := range subnets {
+		if subnet.Role == "ACTIVE" && subnet.Purpose == "INTERNAL_HTTPS_LOAD_BALANCER" {
+			return subnet.IpCidrRange, nil
+		}
+	}
+	return "", fmt.Errorf("L7 ILB Subnet not found in region: %s", region)
```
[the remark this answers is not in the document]
spencerhance: What is the standard not found err?  I don't think a 404 makes sense here.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -339,3 +340,62 @@ func TestAlphaHealthCheck(t *testing.T) {
...
+				forILB: true,
+				ForNEG: true,
+			},
+			version: meta.VersionAlpha,
+		},
+	}
+	t.Parallel()
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			result := tc.hc.Version()
```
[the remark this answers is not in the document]
spencerhance: It's just testing the logic of https://github.com/kubernetes/ingress-gce/blob/52c127605fb56dfe43f3433b0aa4f8cbbb946cb2/pkg/healthchecks/healthchecks.go#L539.  I'll try to find the original discussion.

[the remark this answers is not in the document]
spencerhance: Here's the original comment: https://github.com/kubernetes/ingress-gce/pull/789#discussion_r299804373

pkg/loadbalancers/features/l7ilb.go
spencerhance: Gotcha, will resolve

pkg/utils/utils.go
```diff
@@ -269,7 +269,9 @@ func IGLinks(igs []*compute.InstanceGroup) (igLinks []string) {
 func IsGCEIngress(ing *extensions.Ingress) bool {
 	class := annotations.FromIngress(ing).IngressClass()
 	if flags.F.IngressClass == "" {
-		return class == "" || class == annotations.GceIngressClass
+		// TODO: (shance) remove flag check for L7-ILB once fully rolled out
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/utils/utils.go
```diff
@@ -281,6 +283,13 @@ func IsGCEMultiClusterIngress(ing *extensions.Ingress) bool {
 	return class == annotations.GceMultiIngressClass
 }
 
+// IsGCEL7ILBIngress returns true if the given Ingress has
+// ingress.class annotation set to "gce-l7-ilb"
+func IsGCEL7ILBIngress(ing *extensions.Ingress) bool {
```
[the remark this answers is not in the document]
spencerhance: done

pkg/loadbalancers/l7.go
```diff
@@ -315,7 +315,12 @@ func GetLBAnnotations(l7 *L7, existing map[string]string, backendSyncer backends
 	}
 	backendState := map[string]string{}
 	for _, beName := range backends {
-		backendState[beName] = backendSyncer.Status(beName, l7.version, l7.scope)
+		state, err := backendSyncer.Status(beName, l7.version, l7.scope)
+		// Don't return error here since we wan't to keep syncing
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/loadbalancers/l7.go
```diff
@@ -315,7 +315,12 @@ func GetLBAnnotations(l7 *L7, existing map[string]string, backendSyncer backends
...
+		if err != nil {
+			klog.Error(err)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce"
+)
+
+// Get Subnet source range for ILB
+// TODO: (shance) refactor to use filter
+func ILBSubnetSourceRange(cloud *gce.Cloud, region string) (string, error) {
+	subnets, err := cloud.Compute().AlphaSubnetworks().List(context.Background(), region, filter.None)
+	if err != nil {
+		return "", fmt.Errorf("error obtaining subnets for region: %s", region)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	for _, subnet := range subnets {
+		if subnet.Role == "ACTIVE" && subnet.Purpose == "INTERNAL_HTTPS_LOAD_BALANCER" {
+			return subnet.IpCidrRange, nil
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks_test.go
```diff
@@ -339,3 +340,62 @@ func TestAlphaHealthCheck(t *testing.T) {
...
+				HealthCheck: computealpha.HealthCheck{
+					Type: string(annotations.ProtocolHTTP),
+				},
+				forILB: true,
+				ForNEG: true,
+			},
+			version: meta.VersionAlpha,
+		},
+	}
+	t.Parallel()
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +270,59 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
+	}
+	return NewHealthCheck(alphaHC)
+}
+
 // Get returns the health check by port
-func (h *HealthChecks) Get(name string, version meta.Version) (*HealthCheck, error) {
+func (h *HealthChecks) Get(name string, version meta.Version, scope meta.KeyType) (*HealthCheck, error) {
+	klog.V(3).Infof("Getting Health Check, name: %s, version: %v, scope: %v", name, version, scope)
+
+	// L7-ILB is the only use of regional right now
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -206,22 +270,59 @@ func mergeHealthcheck(oldHC, newHC *HealthCheck) *HealthCheck {
...
+		// L7-ILB is the only use of regional right now
+		return composite.DeleteHealthCheck(cloud, key, features.L7ILBVersion())
+	}
+
 	klog.V(2).Infof("Deleting health check %v", name)
+	// Not using composite here since the tests still rely on the fake health check interface
 	return h.cloud.DeleteHealthCheck(name)
 }
 
+func (h *HealthChecks) getILB(name string) (*HealthCheck, error) {
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -162,7 +202,31 @@ func (h *HealthChecks) create(hc *HealthCheck) error {
 	}
 }
 
+func (h *HealthChecks) updateILB(oldHC, newHC *HealthCheck) error {
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/healthchecks/healthchecks.go
```diff
@@ -108,7 +110,14 @@ func (h *HealthChecks) New(sp utils.ServicePort) *HealthCheck {
 // Sync retrieves a health check based on port, checks type and settings and updates/creates if necessary.
 // Sync is only called by the backends.Add func - it's not a pool like other resources.
 func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
-	existingHC, err := h.Get(hc.Name, hc.Version())
+	var scope meta.KeyType
+	if hc.forILB {
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/firewalls/controller.go
```diff
@@ -164,8 +171,26 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 	negPorts := fwc.translator.GatherEndpointPorts(gceSvcPorts)
 
+	var additionalRanges []string
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/backends/syncer_test.go
```diff
@@ -206,7 +206,7 @@ func TestSyncUpdateHTTP2(t *testing.T) {
 	}
 
 	// Assert the proper health check was created
-	hc, _ = syncer.healthChecker.Get(beName, meta.VersionAlpha)
+	hc, _ = syncer.healthChecker.Get(beName, meta.VersionAlpha, meta.Global)
```
bowei: why is this not ScopeFromServicePort?

pkg/composite/utils.go
```diff
@@ -48,17 +48,22 @@ func CreateKey(gceCloud *gce.Cloud, name string, scope meta.KeyType) (*meta.Key,
 // TODO: (shance) generate this
 // TODO: (shance) add regional and alpha
 // TODO: (shance) populate scope
+// TODO: (shance) figure out a better way to get version
```
bowei: We should definitely update the doc for this function call as it now lists both regional and global resources and combines them into a single list. We should be careful that it is not possible for objects created in one scope to be put into an API call for a different scope.

pkg/composite/utils.go
```diff
@@ -82,17 +87,22 @@ func ListAllUrlMaps(gceCloud *gce.Cloud) ([]*UrlMap, error) {
 // TODO: (shance) generate this
 // TODO: (shance) add regional and alpha
 // TODO: (shance) populate scope
+// TODO: (shance) figure out a more accurate way to obtain version
```
bowei: same here

pkg/composite/utils.go
```diff
@@ -82,17 +87,22 @@ func ListAllUrlMaps(gceCloud *gce.Cloud) ([]*UrlMap, error) {
...
+	}
 
 	// List ga-global and regional-alpha
-	versions := []meta.Version{meta.VersionGA}
-	keys := []*meta.Key{key1}
+	versions := []meta.Version{meta.VersionGA, meta.VersionAlpha}
+	keys := []*meta.Key{key1, key2}
 
 	for i := range versions {
 		list, err := ListBackendServices(gceCloud, keys[i], versions[i])
```
bowei: This still seems quite strange to me that the list operation takes a key that it then only looks @ the scope for.

pkg/firewalls/firewalls_test.go
```diff
@@ -110,25 +110,40 @@ func TestFirewallPoolSyncPorts(t *testing.T) {
...
 	}
 	verifyFirewallRule(fwp, ruleName, nodes, srcRanges, append(portRanges(), negTargetports...), t)
 }
 
+func TestFirewallPoolSyncRanges(t *testing.T) {
+	fwp := NewFakeFirewallsProvider(false, false)
+	fp := NewFirewallPool(fwp, namer, srcRanges, portRanges())
+	nodes := []string{"node-a", "node-b", "node-c"}
+
+	additionalRanges := []string{"10.128.0.0/24"}
```
bowei: test cases 

empty list
one item
three items

pkg/firewalls/firewalls_test.go
```diff
@@ -110,25 +110,40 @@ func TestFirewallPoolSyncPorts(t *testing.T) {
...
+
+	if err := fp.Sync(nodes, nil, additionalRanges); err != nil {
+		t.Fatal(err)
```
bowei: Should have message to make this easier to debug:

"fp.Sync(%v, nil, %v) = %v; want nil" or something similar

pkg/loadbalancers/features/l7ilb.go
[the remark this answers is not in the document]
bowei: You could do 

var ErrSubnetNotFound = errors.New(""primary subnet not found")

pkg/composite/utils.go
spencerhance: I agree, there's a TODO in pkg/composite/gen/main.go from when we last discussed it.  I can update here or in another PR depending on your preference

pkg/firewalls/firewalls_test.go
spencerhance: Done

pkg/firewalls/firewalls_test.go
spencerhance: Done

pkg/composite/utils.go
spencerhance: Done

pkg/composite/utils.go
spencerhance: Done

pkg/backends/syncer_test.go
spencerhance: Done

pkg/backends/backends.go
```diff
@@ -177,21 +184,27 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
...
 func (b *Backends) List() ([]*composite.BackendService, error) {
 	// TODO: for consistency with the rest of this sub-package this method
 	// should return a list of backend ports.
-	backends, err := composite.ListAllBackendServices(b.cloud)
+	var backends []*composite.BackendService
+	var err error
+	if flags.F.EnableL7Ilb {
+		backends, err = composite.ListAllBackendServices(b.cloud)
+	} else {
+		backends, err = composite.ListBackendServices(b.cloud, meta.GlobalKey(""), meta.VersionGA)
```
bowei: Add a TODO that this needs to be changed to not take a key

pkg/composite/utils.go
```diff
@@ -47,19 +47,26 @@ func CreateKey(gceCloud *gce.Cloud, name string, scope meta.KeyType) (*meta.Key,
...
+// This function combines both global and regional resources into one slice
+// so users must use ScopeFromSelfLink() to determine the correct scope
+// before issuing an API call.
 func ListAllUrlMaps(gceCloud *gce.Cloud) ([]*UrlMap, error) {
 	resultMap := map[string]*UrlMap{}
 	key1, err := CreateKey(gceCloud, "", meta.Global)
 	if err != nil {
 		return nil, err
 	}
+	key2, err := CreateKey(gceCloud, "", meta.Regional)
```
bowei:
```
// TODO(shance): this all needs to be replaced with `scope` and not key.
globalKey, err := CreateKey(gceCloud, "", meta.Global)
if err != nil {
  return nil, err
}
regionalKey, err := ...

for _, vk := range []struct {
  v meta.Version
  k meta.Key // TODO: change this to scope
} {
  {meta.VersionGA, GlobalKey},
  {meta.VersionAlpha, RegionalKey},
} {
  ...
}
```

pkg/firewalls/controller.go
```diff
@@ -163,9 +170,13 @@ func (fwc *FirewallController) sync(key string) error {
 		return err
 	}
 	negPorts := fwc.translator.GatherEndpointPorts(gceSvcPorts)
+	additionalRanges, err := fwc.ilbFirewallSrcRanges(gceIngresses)
```
bowei: flag gate this?

pkg/firewalls/controller.go
```diff
@@ -123,7 +126,11 @@ func (fwc *FirewallController) ToSvcPorts(ings []*extensions.Ingress) []utils.Se
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
 		urlMap, _ := fwc.translator.TranslateIngress(ing, fwc.ctx.DefaultBackendSvcPortID)
-		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		svcPorts := urlMap.AllServicePorts()
+		if utils.IsGCEL7ILBIngress(ing) {
```
bowei: this should be `if enableILBL7 && utils.IsGCEL7Ingress(ing)` so it's completely disabled

pkg/firewalls/controller.go
```diff
@@ -180,3 +191,25 @@ func (fwc *FirewallController) sync(key string) error {
...
 }
+
+func (fwc *FirewallController) ilbFirewallSrcRanges(gceIngresses []*extensions.Ingress) ([]string, error) {
+	var additionalRanges []string
+	ilbEnabled := false
+	if flags.F.EnableL7Ilb {
+		for _, ing := range gceIngresses {
+			if utils.IsGCEL7ILBIngress(ing) {
+				ilbEnabled = true
+			}
```
bowei: break

pkg/firewalls/controller.go
```diff
@@ -180,3 +191,25 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 	return nil
 }
+
+func (fwc *FirewallController) ilbFirewallSrcRanges(gceIngresses []*extensions.Ingress) ([]string, error) {
+	var additionalRanges []string
+	ilbEnabled := false
+	if flags.F.EnableL7Ilb {
```
bowei: I don't like hiding the enablement inside the function. Then when we unit test this, it has different behavior based on a global variable flag

pkg/firewalls/controller.go
```diff
@@ -180,3 +191,25 @@ func (fwc *FirewallController) sync(key string) error {
...
+			}
+		}
+	}
+
+	if ilbEnabled {
+		L7ILBSrcRange, err := features.ILBSubnetSourceRange(fwc.ctx.Cloud, fwc.ctx.Cloud.Region())
+		if err != nil {
+			return nil, fmt.Errorf("error unable to get ILB subnet source ranges: %v", err)
+		}
+		additionalRanges = append(additionalRanges, L7ILBSrcRange)
```
bowei: no need to append if we are not in a loop

return L7ILBSrcRange, nil

pkg/loadbalancers/features/features_test.go
```diff
@@ -46,10 +48,37 @@ var (
 		meta.Zonal:    {fakeZonalFeature},
 	}
 
-	fakeVersionToFeatures = map[meta.Version][]string{
```
bowei: you can reduce verbosity here:
```
makeEntry := func(ga, alpha, beta meta.Version) {
  return &map[meta.Version]{   meta.VersionGA: {ga}, meta.VersionAlpha:{alpha}, meta.VersionBeta:{beta}   }
}
...
UrlMap: ...,
ForwardingRule: makeEntry(fakeGaFeature, fakeAlphaFeature, fakeBetaFeature),
```

spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

pkg/composite/utils.go
spencerhance: Done

pkg/backends/backends.go
spencerhance: Done

pkg/backends/backends.go
```diff
@@ -177,21 +184,27 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 	case meta.Regional:
 		hs, err = b.cloud.GetRegionalBackendServiceHealth(name, b.cloud.Region(), be.Backends[0].Group)
 	default:
-		return "Unknown"
+		return "Unknown", fmt.Errorf("invalid scope for Health(): %s", scope)
 	}
 
 	if err != nil || len(hs.HealthStatus) == 0 || hs.HealthStatus[0] == nil {
-		return "Unknown"
+		return "Unknown", fmt.Errorf("error getting health for backend %s: %v", name, err)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/annotations/ingress.go
```diff
@@ -54,6 +54,7 @@ const (
 	IngressClassKey      = "kubernetes.io/ingress.class"
 	GceIngressClass      = "gce"
 	GceMultiIngressClass = "gce-multi-cluster"
+	GceL7ILBIngressClass = "gce-l7-ilb"
```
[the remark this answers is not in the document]
spencerhance: Changed it to "gce-internal," which I believe was the original name fwiw

pkg/backends/ig_linker.go
```diff
@@ -94,6 +94,7 @@ func (l *instanceGroupLinker) Link(sp utils.ServicePort, groups []GroupKey) erro
 	}
 
 	// ig_linker only supports L7 HTTP(s) External Load Balancer
+	// Hardcoded here since IGs are not supported for non GA-Global right now
```
bowei: you need to add a TODO here so we can find this later

spencerhance: Done

pkg/utils/utils_test.go
```diff
@@ -518,6 +521,146 @@ func TestGetNodeConditionPredicate(t *testing.T) {
...
+						annotations.IngressClassKey: annotations.GceL7ILBIngressClass},
+				},
+			},
+			expected: true,
+		},
+		{
+			desc: "foo ingress class",
+			ingress: &v1beta1.Ingress{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
```
bowei: either do 

```
{ ... }
```
 or 
```
&blah{\n
  ...\n
},\n
```

don't do
```
&blah{\n
   ...},
```

bowei:
/lgtm
/approve

there are a ton of TODOs, but let's merge this for now and handle things in addons. The github UI is not handling PRs of this size very easily.

hach-que:
Is there a way to get this working on Google Kubernetes Engine?

I added the annotation `kubernetes.io/ingress.class: "gce-internal"` to an ingress on GKE, but it doesn't seem to be doing anything.

Do I have to manually deploy ingress-gce to the cluster, and disable the built-in version?

hach-que:
For anyone trying to get this working on Google Kubernetes Engine, you need to:

1. Make sure your GKE cluster is on `1.14`. `1.13` definitely does not work.
2. Ask your Google support channel or technical account manager to enable Alpha Access on the Compute API for the projects you want to deploy this in. This is because the GA and beta APIs do not yet surface subnets reserved for internal HTTPS load balancing (more on that in a bit).
  - If you're desperate to get around this and you're only using ILBs in one region, you can override the implementation of `ILBSubnetSourceRange` in `i7ilb.go` to just straight up return the CIDR of the subnet you reserve in step 5.
3. Before running the setup script linked below, modify `glbc.yaml` and add the `--enable-l7-ilb=true` argument after `--enable-csm=[ENABLE_CSM]`.
4. Follow the instructions here: https://github.com/kubernetes/ingress-gce/blob/master/docs/deploy/gke/README.md
5. I had to make some fixes to the script e.g. to support multiple node pools and to make API server detection more reliable, which can be found in our fork: https://github.com/networknext/ingress-gce/blob/master/docs/deploy/gke/gke-self-managed.sh
6. You will need to remove the IAM permission from the GLBC service account and re-add Compute Admin permission. For some reason, the controller kept saying access denied until we did this.
7. In the Google Cloud Console, go to add a new HTTPS load balancer, select "Only between my VMs", select a specific region, then you will be prompted to "Reserve subnet". Click that and choose a CIDR mask that doesn't overlap with an existing subnet (e.g. `10.28.0.0/20`). Give it whatever name you like. Finish creating the load balancer (just pick whatever), and then delete it after it's spun up.
8. If you rolled out your ingress before this point, you need to delete it as the GCE ingress controller won't have created it properly.
9. Make sure your ingress is using `networking.k8s.io/v1beta1`, e.g. it should look something like this:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: blah
  annotations:
    kubernetes.io/ingress.allow-http: "false"   
    kubernetes.io/ingress.class: "gce-internal"
spec:
  backend:
    serviceName: blah
    servicePort: grpc
```

spencerhance:
@hach-que I'm glad you were able to to get it working!  As you figured out, the feature hasn't been released to GKE yet but can be used if you deploy the controller yourself.  As for step 3, you can actually avoid enabling CSM by adding the enable NEG annotation to your k8s service.  Here is an example service yaml:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: internal-lb-svc
  namespace: default
  annotations:
spec:
  ports:
  - name: host1
    port: 80
    protocol: TCP
    targetPort: 9376
  selector:
    run: internal-lb-app
  type: ClusterIP
```

Also feel free to file issues if you notice any bugs :smiley:

PR #790, opened 2019-07-02
More composite updates for integration

spencerhance:
/assign bowei

blocking https://github.com/kubernetes/ingress-gce/pull/788

### DIFF
--- a/pkg/composite/composite.go
+++ b/pkg/composite/composite.go
@@ -37,7 +37,7 @@ func SetUrlMapForTargetHttpsProxy(gceCloud *gce.Cloud, key *meta.Key, targetHttp
 
 	// Set name in case it is not present in the key
 	key.Name = targetHttpsProxy.Name
-	klog.V(3).Infof("setting URLMap for TargetHttpsProxy %v", targetHttpsProxy.Name)
+	klog.V(3).Infof("setting URLMap for TargetHttpsProxy %v", key)
 
 	switch targetHttpsProxy.Version {
 	case meta.VersionAlpha:
@@ -65,7 +65,7 @@ func SetSslCertificateForTargetHttpsProxy(gceCloud *gce.Cloud, key *meta.Key, ta
 
 	// Set name in case it is not present in the key
 	key.Name = targetHttpsProxy.Name
-	klog.V(3).Infof("setting URLMap for TargetHttpsProxy %v", targetHttpsProxy.Name)
+	klog.V(3).Infof("setting SslCertificate for TargetHttpsProxy %v", key)
 
 	switch targetHttpsProxy.Version {
 	case meta.VersionAlpha:
@@ -94,7 +94,7 @@ func SetUrlMapForTargetHttpProxy(gceCloud *gce.Cloud, key *meta.Key, targetHttpP
 
 	// Set name in case it is not present in the key
 	key.Name = targetHttpProxy.Name
-	klog.V(3).Infof("setting URLMap for TargetHttpProxy %v", targetHttpProxy.Name)
+	klog.V(3).Infof("setting URLMap for TargetHttpProxy %v", key)
 
 	switch targetHttpProxy.Version {
 	case meta.VersionAlpha:
@@ -122,7 +122,7 @@ func SetProxyForForwardingRule(gceCloud *gce.Cloud, key *meta.Key, forwardingRul
 
 	// Set name in case it is not present in the key
 	key.Name = forwardingRule.Name
-	klog.V(3).Infof("setting proxy for forwarding rule ForwardingRule %v", forwardingRule.Name)
+	klog.V(3).Infof("setting proxy for forwarding rule ForwardingRule %v", key)
 
 	switch forwardingRule.Version {
 	case meta.VersionAlpha:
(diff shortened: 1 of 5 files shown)


pkg/composite/utils.go
```diff
@@ -126,18 +126,20 @@ func IsRegionalUrlMap(um *UrlMap) (bool, error) {
...
 	if err != nil {
-		return false, fmt.Errorf("error parsing self-link for url-map %s: %v", selfLink, err)
+		return false, err
 	}
+	return scope == meta.Regional, nil
+}
 
-	if resourceID.Key.Region != "" {
-		return true, nil
+func ParseScope(selfLink string) (meta.KeyType, error) {
```
bowei: unit test?

pkg/composite/composite.go
```diff
@@ -65,7 +65,7 @@ func SetSslCertificateForTargetHttpsProxy(gceCloud *gce.Cloud, key *meta.Key, ta
 
 	// Set name in case it is not present in the key
 	key.Name = targetHttpsProxy.Name
-	klog.V(3).Infof("setting URLMap for TargetHttpsProxy %v", targetHttpsProxy.Name)
+	klog.V(3).Infof("setting SslCertificate for TargetHttpsProxy %v", targetHttpsProxy.Name)
```
bowei: Can we fix the comment, it's not clear what resource the key refers to.

klog.V(3).Infof("SetSslCertificate for TargetHttpsProxy %v", key)

pkg/composite/utils.go
spencerhance: Done

pkg/composite/composite.go
spencerhance: Done

pkg/composite/utils.go
```diff
@@ -126,18 +126,20 @@ func IsRegionalUrlMap(um *UrlMap) (bool, error) {
...
+	resourceID, err := cloud.ParseResourceURL(selfLink)
+	if err != nil {
+		var empty meta.KeyType
+		return empty, fmt.Errorf("error parsing self-link for url-map %s: %v", selfLink, err)
```
bowei: fix this log message (url-map)

pkg/composite/utils.go
```diff
@@ -126,18 +126,20 @@ func IsRegionalUrlMap(um *UrlMap) (bool, error) {
...
 	}
+	return scope == meta.Regional, nil
+}
 
-	if resourceID.Key.Region != "" {
-		return true, nil
+func ParseScope(selfLink string) (meta.KeyType, error) {
+	resourceID, err := cloud.ParseResourceURL(selfLink)
+	if err != nil {
+		var empty meta.KeyType
```
bowei: this is not necessary

return meta.KeyType{}, fmt.Errorf(...)

pkg/composite/utils_test.go
```diff
@@ -114,3 +114,51 @@ func TestIsRegionalResource(t *testing.T) {
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			result, err := ParseScope(tc.selfLink)
+
+			if tc.wantErr {
+				if err == nil {
+					t.Fatalf("err = nil, want non-nil")
+				} else {
```
bowei: you don't need this in an else {}

pkg/composite/utils.go
bowei: probably better name is ScopeFromSelfLink

We are not really parsing a Scope string

pkg/composite/utils.go
spencerhance: That actually won't compile since KeyType is a string

pkg/composite/utils.go
spencerhance: Done

pkg/composite/utils_test.go
spencerhance: Done

pkg/composite/utils.go
spencerhance: Done

pkg/composite/utils.go
spencerhance: replaced it with meta.KeyType("") and that works

pkg/composite/utils_test.go
```diff
@@ -114,3 +114,50 @@ func TestIsRegionalResource(t *testing.T) {
...
 }
+
+func TestParseScope(t *testing.T) {
+	testCases := []struct {
+		desc     string
+		selfLink string
+		expected meta.KeyType
+		wantErr  bool
+	}{
+		{
```
bowei: add a testcase for empty string

pkg/composite/utils_test.go
```diff
@@ -114,3 +114,50 @@ func TestIsRegionalResource(t *testing.T) {
 		})
 	}
 }
+
+func TestParseScope(t *testing.T) {
+	testCases := []struct {
```
bowei: t.Parallel()

bowei:
/approve

ping when done and I will lgtm

spencerhance:
@bowei should be good to go now

spencerhance:
@bowei friendly ping

Issue #595, opened 2019-01-04
Ingress controller should react to node scale down event from autoscaler


freehan:
Cluster autoscaler adds the [ToBeDeletedByClusterAutoscaler](
https://github.com/kubernetes/autoscaler/blob/e7dd38375165997b057967259b0d577ca253eebb/cluster-autoscaler/utils/deletetaint/delete.go#L34) taint on the candidate node. Then it goes on to evict pods from the nodes. After everything settles, it will delete the node.

Ingress controller should observe the taint and react by removing instance from instance group so that connection draining is triggered. This help avoid new connection to be arrive at the removing nodes which causes 502s.

vinicyusmacedo:
Hi, I would like to grab this issue, if no one is working on it.

vinicyusmacedo:
@freehan I might be looking at it wrong, but I think the problem is not on ingress-gce itself.

Over [here](https://github.com/kubernetes/ingress-gce/blob/5bd3a2a1f50962ec2e005151e96892dac74e701f/pkg/utils/utils.go#L318), on `GetNodeConditionPredicate` (which is used to filter nodes during sync), all Unschedulable nodes are marked out of the filter. 

However, over [here](https://github.com/kubernetes/autoscaler/blob/e7dd38375165997b057967259b0d577ca253eebb/cluster-autoscaler/utils/deletetaint/delete.go#L76) the node is only tainted as NoSchedule.

IMO the autoscaler should also mark the node as Unschedulable. I'm writing a few tests on the autoscaler for that.

freehan:
Thanks for the PR!

freehan:
Yes. As you pointed out, the problem is the handshake between autoscaler and ingress-gce (or any other loadbalancer controller). This required a better design for synchronization between cluster node life cycle and load balancer controller. 

But for now, as a stop gap, we just need to watch autoscaler's taint and react.

philpearl:
Would really like this fix in GKE - can anyone comment on how long it will take before it's available?

freehan:
We will cherry pick this into 1.6 branch. And it will follow the GKE release pipeline and possibly available in newer version of 1.13.8+

PR #792, opened 2019-07-07
Added NoSchedule effect to GetNodeConditionPredicate

vinicyusmacedo:
This PR fixes https://github.com/kubernetes/ingress-gce/issues/595.

It searches for a taint with effect NoSchedule on GetNodeConditionPredicate. This way, the Load Balancer Sync can filter out nodes with taints such as ToBeDeletedByClusterAutoscaler. 

Notice that, on [Kubernetes Autoscaler](https://github.com/kubernetes/autoscaler/blob/e7dd38375165997b057967259b0d577ca253eebb/cluster-autoscaler/utils/deletetaint/delete.go#L76), the node is only tainted, it doesn't look like the `node.Spec.Unschedulable` is set anywhere.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -67,6 +67,9 @@ const (
 	// This label is feature-gated in kubernetes/kubernetes but we do not have feature gates
 	// This will need to be updated after the end of the alpha
 	LabelNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
+	// ToBeDeletedTaint is the taint that the autoscaler adds when a node is scheduled to be deleted
+	// https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-0.5.2/cluster-autoscaler/utils/deletetaint/delete.go#L33
+	ToBeDeletedTaint = "ToBeDeletedByClusterAutoscaler"
 )
 
 // FakeGoogleAPIForbiddenErr creates a Forbidden error with type googleapi.Error
@@ -317,6 +320,13 @@ func GetNodeConditionPredicate() listers.NodeConditionPredicate {
 		// the master.
 		if node.Spec.Unschedulable {
 			return false
+		}
+
+		// Get all nodes that have a taint with NoSchedule effect
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == ToBeDeletedTaint {
+				return false
+			}
 		}
 
 		// As of 1.6, we will taint the master, but not necessarily mark it unschedulable.
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -17,8 +17,11 @@ limitations under the License.
 package utils
 
 import (
+	"fmt"
 	"testing"
+	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -319,6 +319,13 @@ func GetNodeConditionPredicate() listers.NodeConditionPredicate {
 			return false
 		}
 
+		// Get all nodes that have a taint with NoSchedule effect
+		for _, taint := range node.Spec.Taints {
```
freehan: I feel like it should not be this generic. It should be limited to the autoscaler's `ToBeDeletedByClusterAutoscaler` taint. 
Otherwise, if all nodes mark by no schedule by accident, it would disrupt all existing LB traffic.

vinicyusmacedo: Sounds better, I'll change it.

pkg/utils/utils.go
```diff
@@ -319,6 +319,13 @@ func GetNodeConditionPredicate() listers.NodeConditionPredicate {
...
+			if taint.Key == "ToBeDeletedByClusterAutoscaler" {
```
freehan: I am wondering if there is a repo holding this well known taints?
If it can be easily referenced in this repo, then probably reference it directly.
If not, probably worthy putting it into a const variable and add comments and links to its definition.

vinicyusmacedo: @freehan it is exported on the autoscaler, I can try to use it. The only problem I might have is with vendoring.

vinicyusmacedo: @freehan It was kind of hard to vendor it (there's one dependency which go couldn't find) so I kept it as a const.

pkg/utils/utils.go
```diff
@@ -67,6 +67,8 @@ const (
 	// This label is feature-gated in kubernetes/kubernetes but we do not have feature gates
 	// This will need to be updated after the end of the alpha
 	LabelNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
+	// ToBeDeletedTaint is the taint that the autoscaler adds when a node is scheduled to be deleted
+	ToBeDeletedTaint = "ToBeDeletedByClusterAutoscaler"
```
freehan: Is there any public documentation regarding this taint?

If not, probably still need to add a link to the autoscaler repo.
https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-0.5.2/cluster-autoscaler/utils/deletetaint/delete.go#L33

vinicyusmacedo: @freehan I didn't find any docs on it :/

I'll try to vendor it and its dependencies.

freehan: If there is tons of dependencies, just add the link in the comment then. No need for the hustle.

freehan:
Sorry for nit picking. Would you mind squash the commits? 

I will lgtm it afterwards. Thanks!!

vinicyusmacedo:
@freehan not a problem. There ya go!

freehan:
/lgtm
/approve

Thanks!

PR #796, opened 2019-07-09
HealthChecks switch from fakes to fakeGCE

spencerhance:
This is required for the L7-ILB work since we need to remove the fakes in order to add unit tests that use composites.

Requires a new k8s-cloud-provider release


pkg/healthchecks/healthchecks.go
```diff
@@ -71,7 +72,7 @@ const (
 
 // HealthChecks manages health checks.
 type HealthChecks struct {
-	cloud HealthCheckProvider
```
freehan: I am not sure what is the benefit of removing the `HealthCheckProvider` interface?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -202,15 +205,21 @@ func TestHTTP2HealthCheckDelete(t *testing.T) {
...
 	}
 }
 
 func TestHealthCheckUpdate(t *testing.T) {
-	hcp := NewFakeHealthCheckProvider()
-	healthChecks := NewHealthChecker(hcp, "/", "/healthz", namer, defaultBackendSvc)
+	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
+
+	// Add Hooks
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockHealthChecks.UpdateHook = mock.UpdateHealthCheckHook
```
freehan: you can add a helper function to initialize fakeGCE with healthcheck mocks.

freehan: up to you. It look like this is only place it needs the Hook.

pkg/healthchecks/healthchecks.go
spencerhance: The only benefit is just removing stale code since we are using the gce.Cloud interface instead

pkg/healthchecks/healthchecks_test.go
spencerhance: Gotcha, personally I'd rather leave it out unless we need it.  I'll probably end up adding it in an l7-ilb PR anyway.

pkg/healthchecks/healthchecks.go
freehan: I prefer having a clear interface in the HealthChecker. Then it enforces that the Healthchecker  only communicates via the interfaces instead of making arbitrary calls.

spencerhance: Fair enough.  I'll revert it.

spencerhance:
@freehan should be good to go now

PR #802, opened 2019-07-23
dump events in sandbox

freehan:
Dumps events at the end of each test. Make it easier to debug e2e failure.

### DIFF
--- a/pkg/e2e/framework.go
+++ b/pkg/e2e/framework.go
@@ -207,6 +207,8 @@ func (f *Framework) RunWithSandbox(name string, t *testing.T, testFunc func(*tes
 		}
 
 		testFunc(t, sandbox)
+
+		sandbox.DumpSandboxInfo(t)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


pkg/e2e/framework.go
```diff
@@ -207,6 +207,10 @@ func (f *Framework) RunWithSandbox(name string, t *testing.T, testFunc func(*tes
 		}
 
 		testFunc(t, sandbox)
+
+		if t.Failed() {
+			sandbox.DumpSandboxInfo()
```
bowei: why not always dump cluster events?

pkg/e2e/sandbox.go
```diff
@@ -93,3 +94,40 @@ func (s *Sandbox) MasterUpgraded() bool {
 func (s *Sandbox) MasterUpgrading() bool {
 	return s.f.statusManager.masterUpgrading()
 }
+
+// DumpSandboxInfo dumps information about the sandbox into logs
+func (s *Sandbox) DumpSandboxInfo() {
+	s.dumpSandboxEvents()
```
bowei: why have this indirection?

freehan: The indirection is in case in the future we need to dump more category of info. (e.g. GCE objects )

freehan:
Ready fo another round

PR #806, opened 2019-07-29
Change from "extensions.v1beta1" to "networking.v1beta1"

bowei:
Change API group from "extensions" to "networking"
    
    This migrates our code to use the networking.v1beta1.Ingress* API
    types instead of extensions.v1beta1. This also fixes fixture YAMLs
    to use "networking.k8s.io" as the fully-qualified resource kind.

### DIFF
--- a/cmd/e2e-test/affinity_test.go
+++ b/cmd/e2e-test/affinity_test.go
@@ -104,7 +104,7 @@ func TestAffinity(t *testing.T) {
 			}
 			t.Logf("Echo service created (%s/%s)", s.Namespace, "service-1")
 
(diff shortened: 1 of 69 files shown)


bowei:
cc: @liggitt

rramkumar1:
@bowei We would need to update the YAML's in our GCP documentation as well. 

[1] https://cloud.google.com/kubernetes-engine/docs/concepts/ingress

freehan:
Will it still recognize ingress objects in extension api group? 

I assume API machinery should have handled this kind of backward compatibility, right?

bowei:
Yes, asked Jordan, this should work as long as it is used with v1.15+ k8s clusters

Issue #18, opened 2017-10-11
Add example using grpc and http2 


bowei:
_From @aledbf on December 1, 2016 22:39_

_Copied from original issue: kubernetes/ingress-nginx#39_

bowei:
_From @aledbf on December 30, 2016 3:24_

Something like this should work:

https://github.com/caiofilipini/grpc-weather/

- create deployment
```
kubectl run \
  --image=caiofilipini/grpc-weather:master grpc-weather \
  --port=9000 \
  --env="OPEN_WEATHER_MAP_API_KEY=<token from http://openweathermap.org/api>" \
  --env="WEATHER_UNDERGROUND_API_KEY=<token from http://openweathermap.org/api>"
```

- expose service
```
kubectl expose deployment grpc-weather --port=9000 --name=grpc-weather
```

- create ingress
```
echo "
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: grpc-weather
spec:
  rules:
  - host: grpc-weather
    http:
      paths:
      - backend:
          serviceName: grpc-weather
          servicePort: 9000
        path: /
" | kubectl create -f -
```

- use docker image to run the client
```
docker run \
  --rm \
  -it \
  --name weather_service \
  --entrypoint bash \
  --net=host \
  caiofilipini/grpc-weather:master
```

- execute
```
echo "<grpc-weather svc IP> grpc-weather" >>/etc/hosts
curl -v <grpc-weather svc IP>:80 -H 'Host: grpc-weather"
make build-client
./weather_client/client --s grpc-weather --p 80 Santiago
```

bowei:
_From @krancour on April 3, 2017 16:21_

@aledbf I haven't tried, but does this work? I recently tried using a different upstream grpc service and requests kept failing because of connection reset by peer.

bowei:
_From @aledbf on April 3, 2017 16:25_

@krancour works if you use the `ssl-passthrough` annotation, i.e. just use nginx for the TLS hello and then terminate SSL in the pod (this is because nginx does not support http2 in the upstream)

bowei:
_From @krancour on April 3, 2017 16:27_

@aledbf ahhh... thanks! I will try that.

bowei:
_From @philipithomas on June 28, 2017 21:58_

can a grpc server listen on port 80? More specifically - how can ssl-passthrough be configured for port 80?

bowei:
[...]

PR #807, opened 2019-07-29
Use readinessProbe HC spec for platform HC

salrashid123:
Proposal+PR:  Overide GCP healthcheck to allow gRPC protocol services.

This change will override the GCP healthcheck settings with the first container which has an HTTP/HTTPS  `readinessProbe` defined

--

At the moment, customers deploying a gRPC service on GCP that utilizes `ingress-gce` need to either run a mux server or a proxy that handles HTTP and gRPC because the GCP healthcheck uses the _servingPort_ for HTTP healthchecks.  If i run a gRPC service on the serving port, then the GCP healthchecks will never pass.  Users have to implement a workaround like this:
- ref: https://github.com/kubernetes/ingress-gce/issues/18#issuecomment-498669572

With this change if i have a `Deployment` which includes a `readinessProbe` of type HTTP, then the GCP Instance group Healthcheck settings will be overrided and will use the specifications (path/port/protocol) defined in the container.

The attached PR has the changes and i've tested the scenarios  that i've documented here:

- https://github.com/salrashid123/salrashid123.github.io/tree/master/ingress-gce/tmp

(this needs much more exhaustive testing still).

One scenario i did not account for is using NEG+gRPC since the NEG HC will still use HTTP2 against the gRPC service directly.  In that case, customers may have to go back to using the mux.

### DIFF
--- a/pkg/backends/interfaces.go
+++ b/pkg/backends/interfaces.go
@@ -31,11 +31,11 @@ type GroupKey struct {
 	Name string
 }
 
-// ServicePortAndProbe represents the Node and Container redinessProbe
-// specifications for aservice
+// ServicePortAndProbe encapsulates a service port and the associated
+// readiness probe specification.
 type ServicePortAndProbe struct {
-	Probe   *api_v1.Probe
-	Service *api_v1.ServicePort
+	Probe       *api_v1.Probe
+	ServicePort *api_v1.ServicePort
 }
 
 // Pool is an interface to perform CRUD operations on a pool of GCE
@@ -81,7 +81,7 @@ type NEGGetter interface {
 	GetNetworkEndpointGroup(name string, zone string) (*compute.NetworkEndpointGroup, error)
 }
 
-// ProbeProvider retrieves a ServicePortAndProbe struct given a nodePort
+// ProbeProvider retrieves a ServicePortAndProbe given a ServicePort
 type ProbeProvider interface {
 	GetProbe(sp utils.ServicePort) (*ServicePortAndProbe, error)
 }
--- a/pkg/backends/interfaces.go
+++ b/pkg/backends/interfaces.go
@@ -85,4 +85,3 @@ type NEGGetter interface {
 type ProbeProvider interface {
 	GetProbe(sp utils.ServicePort) (*ServicePortAndProbe, error)
 }
-
--- a/pkg/backends/interfaces.go
+++ b/pkg/backends/interfaces.go
@@ -31,6 +31,13 @@ type GroupKey struct {
 	Name string
 }
 
+// ServicePortAndProbe represents the Node and Container redinessProbe
+// specifications for aservice
+type ServicePortAndProbe struct {
+	Probe   *api_v1.Probe
+	Service *api_v1.ServicePort
+}
+
 // Pool is an interface to perform CRUD operations on a pool of GCE
 // Backend Services.
 type Pool interface {
@@ -74,7 +81,8 @@ type NEGGetter interface {
 	GetNetworkEndpointGroup(name string, zone string) (*compute.NetworkEndpointGroup, error)
 }
 
-// ProbeProvider retrieves a probe struct given a nodePort
+// ProbeProvider retrieves a ServicePortAndProbe struct given a nodePort
 type ProbeProvider interface {
-	GetProbe(sp utils.ServicePort) (*api_v1.Probe, error)
+	GetProbe(sp utils.ServicePort) (*ServicePortAndProbe, error)
 }
+
--- a/pkg/backends/syncer_test.go
+++ b/pkg/backends/syncer_test.go
@@ -67,7 +67,11 @@ func newTestSyncer(fakeGCE *gce.Cloud) *backendSyncer {
 		namer:         defaultNamer,
 	}
 
-	probes := map[utils.ServicePort]*api_v1.Probe{{NodePort: 443, Protocol: annotations.ProtocolHTTPS}: existingProbe}
+	sp := &ServicePortAndProbe{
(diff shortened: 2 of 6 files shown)


salrashid123:
signed CLA

salrashid123:
signed it actually

pkg/backends/interfaces.go
```diff
@@ -74,7 +81,8 @@ type NEGGetter interface {
 	GetNetworkEndpointGroup(name string, zone string) (*compute.NetworkEndpointGroup, error)
 }
 
-// ProbeProvider retrieves a probe struct given a nodePort
+// ProbeProvider retrieves a ServicePortAndProbe struct given a nodePort
```
rramkumar1: retrieves a ServicePortAndProbe given a ServicePort

pkg/backends/interfaces.go
```diff
@@ -31,6 +31,13 @@ type GroupKey struct {
 	Name string
 }
 
+// ServicePortAndProbe represents the Node and Container redinessProbe
+// specifications for aservice
+type ServicePortAndProbe struct {
+	Probe   *api_v1.Probe
+	Service *api_v1.ServicePort
```
rramkumar1: Would prefer if this field was called ServicePort

pkg/backends/syncer_test.go
```diff
@@ -67,7 +67,11 @@ func newTestSyncer(fakeGCE *gce.Cloud) *backendSyncer {
 		namer:         defaultNamer,
 	}
 
-	probes := map[utils.ServicePort]*api_v1.Probe{{NodePort: 443, Protocol: annotations.ProtocolHTTPS}: existingProbe}
+	sp := &ServicePortAndProbe{
+		Service: &api_v1.ServicePort{},
+		Probe:   existingProbe,
+	}
```
rramkumar1: Just inline this like sp := &ServicePortAndProbe{Service: ..., Probe: ...}

salrashid123:
are there any other things to address on this?  I ran through the test cases i've got on the live clsuter as mentioned in link on the first comment.  

Is it possible for GKE clusters to specify the version of ingress-gce to use?  that way, customers can safely migrate over (or rollback)

rramkumar1:
@salrashid123 Sorry I have not had time to dedicate to this but rest assured we want to get this out in our next minor release.

User's will be able to upgrade to the new version or rollback via GKE upgrades themselves.

freehan:
I feel like it is quite risky to touch the code modified in the PR. Is it possible to do this surgically without refactoring exiting code? 

Can you do `applyServicePortSettingsToHC` after https://github.com/kubernetes/ingress-gce/blob/v1.6.0/pkg/backends/syncer.go#L188? Since ServicePort abstraction is already there.

salrashid123:
i'm not sure if its possible to do that...`applyProbeSettingsToHC` by itself modifies the healthcheck so you're suggesting another ovewrite of that?  also, i think i need to extract out the `port/protocol/path` stuff out ouf the `readinessProbe` and surface that back.   

@rramkumar1   any thoughts on this (most of these changes were on suggestion earlier).  Also, the comment on ingress-gce using `Ingress` object (annotations?) containing the HC setting we (gcp) would apply.  (pls also cite what that'd look like or a link to th e proposal, etc)

freehan:
> i'm not sure if its possible to do that...applyProbeSettingsToHC by itself modifies the healthcheck so you're suggesting another ovewrite of that? also, i think i need to extract out the port/protocol/path stuff out ouf the readinessProbe and surface that back.

I meant only override in the cases that you are interested (gRPC). Hence the existing logic can be left largely untouched. 

FYI, This piece of code is very error prune in general. (Mostly because we are deriving health check configuration from pod readiness probe. Hence driving the deprecation) Imagine after version upgrade, the new ingress controller thinks the health check needs to be configured differently and traffic suddenly stopped because the health check configuration changed.

salrashid123:
there really isn't a way to know that the underlying service is `gRPC` based though.

re: ` *api_v1.Probe`  object. yeah, thats alrady there but i needed the   `*api_v1.ServicePort` to write out the the system (gcp) healthcheck port
https://github.com/kubernetes/ingress-gce/pull/807/files#diff-7c4fd1c50a56df681d4755e84cbe6e6fR280

---

i do agree these are invasive changes (hence my hesitation on a lot more testing).  Upgrades would also alter how hc's are done (eg, if i hade an existing hc, it'd use the default serving port under `/` and then begin to use the stuff under what `readinesProbe` happens to be deifned.   Its arguable using the readinessProbe is what a HC should be checking but i see the risks here.

Can you send me a link/proposal to how `type:Ingress` will define where the system (gcp) healthchecks will point to (eg, annotations that happens to point to the service port/path where i (preferably) have defined a `readinessProbe`?

freehan:
> re: *api_v1.Probe object. yeah, thats alrady there but i needed the *api_v1.ServicePort to write out the the system (gcp) healthcheck port

`util.ServicePort` contains the information you need, right?
https://github.com/kubernetes/ingress-gce/blob/v1.6.0/pkg/backends/syncer.go#L176

salrashid123:
`util.ServicePort` thats available to syncher.go  includes the port info about the default service and not the derived one that points to one exposed by the first container w/ an http redinessProbe.

for example, the test case [here](https://github.com/salrashid123/salrashid123.github.io/blob/master/ingress-gce/tmp/deployments_grpc/fe-deployment.yaml#L17), if we used the plain `util.ServicePort`, the HC would point to `:50001` and not `:8080` (meaning the HC in that case would still point back to the grpcservice in that case (like what we have now).  THe code in this PR extracts the port value for the first http readinessprobe thats defined as a serivce and uses that.

Issue #797, opened 2019-07-15
Backend syncer should not block syncing if one backend fails GC


spencerhance:
Currently backendSyncer.GC() will break out of the GC loop if one backend fails.  It should collect all of the errors into a slice and return that instead.

https://github.com/kubernetes/ingress-gce/blob/master/pkg/backends/syncer.go#L135

rramkumar1:
@shance Sounds reasonable, you mind fixing it?

spencerhance:
sounds good /assign @spencerhance

rramkumar1:
@spencerhance Can this be closed?

/kind cleanup

PR #810, opened 2019-08-01
Refactor Backend GC

spencerhance:
Refactor backends GC so that we don't need ListAllBackendServices() anymore.  That way, if one of the api calls fails, it won't block the others.

### DIFF
--- a/pkg/backends/syncer.go
+++ b/pkg/backends/syncer.go
@@ -23,7 +23,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/ingress-gce/pkg/backends/features"
 	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/healthchecks"
+	lbfeatures "k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog"
 	"k8s.io/legacy-cloud-providers/gce"
@@ -68,7 +70,6 @@ func (s *backendSyncer) Sync(svcPorts []utils.ServicePort) error {
 		}
 	}
 	return nil
-
 }
 
 // ensureBackendService will update or create a BackendService for the given port.
@@ -143,11 +144,42 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
 		return err
 	}
 
-	backends, err := s.backendPool.List()
-	if err != nil {
-		return fmt.Errorf("error getting the names of controller-managed backends: %v", err)
+	// Only GC L7 ILB backends if it's enabled
+	if flags.F.EnableL7Ilb {
+		// TODO(shance): Refactor out empty key field
+		key, err := composite.CreateKey(s.cloud, "", meta.Regional)
+		if err != nil {
+			return fmt.Errorf("error creating l7 ilb key: %v", err)
+		}
+		ilbBackends, err := s.backendPool.List(key, lbfeatures.L7ILBVersions().BackendService)
+		if err != nil {
+			return fmt.Errorf("error listing regional backends: %v", err)
+		}
+		err = s.gc(ilbBackends, knownPorts)
+		if err != nil {
+			return fmt.Errorf("error GCing regional Backends: %v", err)
+		}
 	}
 
+	// Requires an empty name field until it is refactored out
+	key, err := composite.CreateKey(s.cloud, "", meta.Global)
+	if err != nil {
+		return fmt.Errorf("error creating l7 ilb key: %v", err)
+	}
+	backends, err := s.backendPool.List(key, meta.VersionGA)
+	if err != nil {
+		return fmt.Errorf("error listing backends: %v", err)
+	}
+	err = s.gc(backends, knownPorts)
+	if err != nil {
+		return fmt.Errorf("error GCing Backends: %v", err)
+	}
+
+	return nil
+}
+
+// gc deletes the provided backends
+func (s *backendSyncer) gc(backends []*composite.BackendService, knownPorts sets.String) error {
 	for _, be := range backends {
 		var key *meta.Key
 		name := be.Name
--- a/pkg/backends/syncer_test.go
+++ b/pkg/backends/syncer_test.go
@@ -19,6 +19,7 @@ package backends
 import (
 	"context"
 	"fmt"
+	"k8s.io/ingress-gce/pkg/composite"
 	"net/http"
 	"reflect"
 	"testing"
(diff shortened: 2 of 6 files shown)


pkg/backends/syncer_test.go
```diff
@@ -267,6 +268,82 @@ func TestGC(t *testing.T) {
 	}
 }
 
+// Test GC with both ELB and ILBs
```
bowei: can we make this very generic so we can easily add lots of tests?

```
type portset struct {
  all []utils.ServicePort
  existing []utils.ServicePort
}

func (p *portset) add() error {
}

func (p *portset) del() error {
}

func (p *portset) gc() error {
}

func (p *portset) check() error {
}
```

then the test can be:
```
ps := portset{...}
ps.add()
ps.del()
if err := ps.check(); err != nil { .. }
ps.gc()
ps.check()
if err := ps.check(); err != nil { .. }
```

pkg/backends/syncer.go
```diff
@@ -144,18 +146,51 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
...
+	// Only GC L7 ILB backends if it's enabled
+	if flags.F.EnableL7Ilb {
+		// TODO(shance): Refactor out empty key field
+		key, err := composite.CreateKey(cloud, "", meta.Regional)
+		if err != nil {
+			return fmt.Errorf("error creating l7 ilb key: %v", err)
+		}
+		ilbBackends, err := s.backendPool.List(key, lbfeatures.L7ILBVersion(lbfeatures.BackendService))
+		if err != nil {
+			return fmt.Errorf("error listing L7-ILB backends: %v", err)
```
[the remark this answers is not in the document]

pkg/backends/syncer_test.go
spencerhance: Done, I'm not sure if it was exactly what you were looking for though

pkg/backends/syncer.go
```diff
@@ -143,11 +144,42 @@ func (s *backendSyncer) GC(svcPorts []utils.ServicePort) error {
...
+		}
+		ilbBackends, err := s.backendPool.List(key, lbfeatures.L7ILBVersions().BackendService)
+		if err != nil {
+			return fmt.Errorf("error listing regional backends: %v", err)
+		}
+		err = s.gc(ilbBackends, knownPorts)
+		if err != nil {
+			return fmt.Errorf("error GCing regional Backends: %v", err)
+		}
+	}
```
bowei: this is a TODO, but we may want to consider continuing on error so a single error does not block all GC operations

pkg/backends/syncer_test.go
```diff
@@ -221,49 +276,84 @@ func TestGC(t *testing.T) {
 		{NodePort: 82, Protocol: annotations.ProtocolHTTPS},
 		{NodePort: 83, Protocol: annotations.ProtocolHTTP},
 	}
+	ps := portset{all: svcNodePorts}
 
-	if err := syncer.Sync(svcNodePorts); err != nil {
-		t.Fatalf("Expected syncer to add backends with error, err: %v", err)
+	if err := syncer.Sync(ps.all); err != nil {
+		t.Fatalf("error syncing backends: %v", err)
```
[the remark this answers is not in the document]
bowei: You will want to follow this format for most of the unit test checks

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,57 @@ import (
...
+	for _, sp := range p.existing {
+		beName := sp.BackendName(defaultNamer)
+		key, err := composite.CreateKey(fakeGCE, beName, features.ScopeFromServicePort(&sp))
+		if err != nil {
+			return fmt.Errorf("Error creating key for backend service %s: %v", beName, err)
+		}
+		if _, err := composite.GetBackendService(fakeGCE, key, features.VersionFromServicePort(&sp)); err != nil {
+			return fmt.Errorf("Expected to find backend for port %v, err: %v", sp.NodePort, err)
+		}
+	}
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,57 @@ import (
...
+		for i, sp := range p.all {
+			if dp == sp {
+				// Delete it
+				p.all = append(p.all[:i], p.all[i:]...)
+				break
+			}
+		}
+		// Make sure it was deleted
+		if len(p.all) == beginLen {
+			return fmt.Errorf("Error: %v does not exist in p.all", dp)
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,57 @@ import (
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
+// portset helps keep track of service ports during GC tests
+type portset struct {
+	all      []utils.ServicePort
+	existing []utils.ServicePort
```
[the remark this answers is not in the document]
spencerhance: Done

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,57 @@ import (
...
+		if len(p.all) == beginLen {
+			return fmt.Errorf("Error: %v does not exist in p.all", dp)
+		}
+	}
+	return nil
+}
+
+// Delete ports from existing that are not in 'all'
+func (p *portset) gc() {
+	p.existing = p.all
```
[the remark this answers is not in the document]
spencerhance: Ended up removing gc() since I think it's less confusing to just use add() and delete()

pkg/backends/syncer_test.go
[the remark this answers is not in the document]
spencerhance: Done

pkg/backends/syncer.go
spencerhance: tracking here: https://github.com/kubernetes/ingress-gce/issues/797

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,80 @@ import (
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
+// portset helps keep track of service ports during GC tests
+type portset struct {
+	// all represents the set all of service ports in the test
+	all map[utils.ServicePort]struct{}
```
bowei: usually sets are represented by map[T]bool

then map[X] == false is natural

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,80 @@ import (
...
+func (p *portset) check(fakeGCE *gce.Cloud) error {
+	for sp, _ := range p.all {
+		_, found := p.existing[sp]
+		beName := sp.BackendName(defaultNamer)
+		key, err := composite.CreateKey(fakeGCE, beName, features.ScopeFromServicePort(&sp))
+		if err != nil {
+			return fmt.Errorf("Error creating key for backend service %s: %v", beName, err)
+		}
+
+		// Make sure it exists
```
bowei: comment is probably redundant, but if you are going to put this, put it inside the brace so that the comment for the different cases are at the same indent level.

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,80 @@ import (
...
+		if found {
+			if _, err := composite.GetBackendService(fakeGCE, key, features.VersionFromServicePort(&sp)); err != nil {
+				return fmt.Errorf("Expected to find backend for port %+v, err: %v", sp.NodePort, err)
```
bowei: "backend for port %+v should exist, but got %v"
"backend for port %+v should not exist, but get %v"

pkg/backends/syncer_test.go
```diff
@@ -39,6 +40,80 @@ import (
...
+
+func newPortset(ports []utils.ServicePort) *portset {
+	ps := portset{all: map[utils.ServicePort]struct{}{}, existing: map[utils.ServicePort]struct{}{}}
+	for _, sp := range ports {
+		ps.all[sp] = struct{}{}
+	}
+	return &ps
+}
+
+func (p *portset) existingSlice() []utils.ServicePort {
```
bowei: existingPorts

pkg/healthchecks/healthchecks.go
```diff
@@ -208,17 +208,16 @@ func (h *HealthChecks) updateILB(oldHC, newHC *HealthCheck) error {
...
 	cloud := h.cloud.(*gce.Cloud)
 
-	compositeType, err := composite.ToHealthCheck(newHC)
+	mergedHC := mergeHealthcheck(oldHC, newHC).ToAlphaComputeHealthCheck()
+	compositeType, err := composite.ToHealthCheck(mergedHC)
 	if err != nil {
 		return fmt.Errorf("Error converting newHC to composite: %v", err)
 	}
-	key, err := composite.CreateKey(cloud, newHC.Name, features.L7ILBScope())
+	key, err := composite.CreateKey(cloud, mergedHC.Name, meta.Regional)
```
bowei: is there a reason why we went back to meta.Regional?

pkg/healthchecks/healthchecks.go
```diff
@@ -314,7 +313,17 @@ func (h *HealthChecks) getILB(name string) (*HealthCheck, error) {
 	if err != nil {
 		return nil, err
 	}
-	return NewHealthCheck(alphaHC)
+
+	newHC, err := NewHealthCheck(alphaHC)
```
bowei: call alphaHC gceHC and you won't have to change the name again when the version changes.

pkg/backends/syncer_test.go
spencerhance: done.  Also i'm a fan, I didn't realize how go handled keys that aren't present.

pkg/backends/syncer_test.go
spencerhance: done

pkg/backends/syncer_test.go
spencerhance: done

pkg/backends/syncer_test.go
spencerhance: done

pkg/healthchecks/healthchecks.go
spencerhance: I thought it seemed redundant but I reverted it back

pkg/healthchecks/healthchecks.go
spencerhance: done

bowei:
looks like the CI had issues.

PR #811, opened 2019-08-01
Add cloud pointer to neg linker and backend syncer

spencerhance:
Remove BackendPool downcast by adding a cloud pointer to the neg linker and backend syncer

### DIFF
--- a/pkg/backends/neg_linker.go
+++ b/pkg/backends/neg_linker.go
@@ -19,6 +19,7 @@ import (
 	befeatures "k8s.io/ingress-gce/pkg/backends/features"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/legacy-cloud-providers/gce"
 )
 
 // negLinker handles linking backends to NEG's.
@@ -26,6 +27,7 @@ type negLinker struct {
 	backendPool Pool
 	negGetter   NEGGetter
 	namer       *utils.Namer
+	cloud       *gce.Cloud
 }
 
 // negLinker is a Linker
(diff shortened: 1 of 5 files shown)


rramkumar1:
/ok-to-test

@spencerhance looks like this needs a rebase

spencerhance:
@rramkumar1 thanks, just fixed it now

pkg/backends/syncer.go
```diff
@@ -43,11 +44,13 @@ var _ Syncer = (*backendSyncer)(nil)
 func NewBackendSyncer(
 	backendPool Pool,
```
rramkumar1: Maybe I'm missing something but doesn't seem like backendPool is used anymore? Can we remove it if that is the case?

pkg/backends/neg_linker.go
```diff
@@ -34,11 +36,13 @@ var _ Linker = (*negLinker)(nil)
 func NewNEGLinker(
 	backendPool Pool,
```
rramkumar1: Same comment as below regarding backendPool not being used any longer.

pkg/backends/syncer.go
spencerhance: We definitely still use it in the syncer

pkg/backends/neg_linker.go
spencerhance: We only use it in the neg linker test

pkg/backends/syncer.go
```diff
@@ -43,11 +44,13 @@ var _ Syncer = (*backendSyncer)(nil)
...
 	healthChecker healthchecks.HealthChecker,
-	namer *utils.Namer) Syncer {
+	namer *utils.Namer,
+	cloud *gce.Cloud) Syncer {
```
bowei: [go style] we don't break line for args like this

PR #814, opened 2019-08-02
Revert e2e test changes from #806

rramkumar1:
This broke e2e tests since we test all version of k8s from HEAD for the e2e test binary.  The new code only works for 1.15 clusters and above

Reverting first to get tests green until we can figure out a permanent mitigation.

/assign @bowei


rramkumar1:
It's going to be too hard to revert cleanly.

/close

PR #816, opened 2019-08-04
Get gke-self-managed.sh working on macOS

yfuruyama:
As using sed for appending a line doesn't work on macOS, changed to use awk.
```sh
# on macOS
$ sed "/name: http/a \ \ \ \ nodePort: ${NODE_PORT}" ../resources/default-http-backend.yaml
sed: 1: "/name: http/a \ \ \ \ n ...": extra characters after \ at the end of a command
```


yfuruyama:
I signed it

yfuruyama:
I signed it

PR #822, opened 2019-08-11
Legacy api e2e

bowei:
Make the e2e framework use an adapter so it can use both the old (extensions) and new (networking) APIs

### DIFF
--- a/cmd/e2e-test/draining_test.go
+++ b/cmd/e2e-test/draining_test.go
@@ -42,10 +42,6 @@ const (
 func TestDraining(t *testing.T) {
 	t.Parallel()
 
-	ing := fuzz.NewIngressBuilder("", "ingress-1", "").
-		AddPath("test.com", "/", "service-1", intstr.FromInt(80)).
-		Build()
-
 	for _, tc := range []struct {
 		desc         string
 		beConfig     *backendconfig.BackendConfig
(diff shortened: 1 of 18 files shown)


pkg/e2e/legacy_test.go
```diff
@@ -0,0 +1,50 @@
+package e2e
+
+import (
+	"reflect"
+	"testing"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
rramkumar1: Don't need this extra space I think.

pkg/e2e/legacy.go
```diff
@@ -0,0 +1,127 @@
...
+	klog.Infof("Delete %s/%s", ns, name)
+	if new {
+		return crud.C.NetworkingV1beta1().Ingresses(ns).Delete(name, &metav1.DeleteOptions{})
+	}
+	klog.Warning("Using legacy API")
+	return crud.C.ExtensionsV1beta1().Ingresses(ns).Delete(name, &metav1.DeleteOptions{})
+}
+
+func (crud *IngressCRUD) supportsNewAPI() (bool, error) {
+	apiList, err := crud.C.Discovery().ServerResourcesForGroupVersion("networking/v1beta1")
```
rramkumar1: Do you need to return error? I don't see anything but nil returned.

cmd/e2e-test/draining_test.go
```diff
@@ -86,12 +82,18 @@ func TestDraining(t *testing.T) {
 			}
 			t.Logf("Echo service created (%s/%s)", s.Namespace, "service-1")
 
-			if _, err := Framework.Clientset.NetworkingV1beta1().Ingresses(s.Namespace).Create(ing); err != nil {
+			ing := fuzz.NewIngressBuilder("", "ingress-1", "").
+				AddPath("test.com", "/", "service-1", intstr.FromInt(80)).
+				Build()
+			ing.Namespace = s.Namespace
```
rramkumar1: Could you just set namespace in the builder? Same for other files too.

pkg/e2e/legacy.go
bowei: done

cmd/e2e-test/draining_test.go
bowei: done

PR #825, opened 2019-08-13
Remove finalizer on changing Ingress class to non-GLBC

skmatti:
This fixes a bug on Ingress resource deletion when Ingress class changes from a GLBC resource to a non-GLBC one. Finalizer of a non-GLBC Ingress resource is removed when GC is invoked on it.


pkg/controller/controller.go
```diff
@@ -390,7 +389,11 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
 	}
 
 	for _, ing := range gcState.ingresses {
-		if utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey) {
+		// Remove finalizer if the ingress is
+		// 1. deleted or
+		// 2. a non-GLBC resource with a finalizer
+		if utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey) ||
+				(!utils.IsGLBCIngress(ing) && utils.HasFinalizer(ing.ObjectMeta, utils.FinalizerKey)) {
```
MrHohn: Is it possible to put these into a func, so that we can better unit test the condition?

MrHohn: e.g. https://github.com/kubernetes/kubernetes/blob/ecad65a3f8622fdf1be556a6c1cd8e19007f3949/pkg/controller/service/service_controller.go#L448-L464

pkg/controller/controller_test.go
```diff
@@ -281,6 +282,50 @@ func TestIngressCreateDeleteFinalizer(t *testing.T) {
...
+		})
+	ing.ObjectMeta.Annotations = map[string]string{"kubernetes.io/ingress.class": "gce"}
+	addIngress(lbc, ing)
+
+	ingStoreKey := getKey(ing, t)
+	if err := lbc.sync(ingStoreKey); err != nil {
+		t.Fatalf("lbc.sync(%v) = err %v", ingStoreKey, err)
+	}
+
+	// Check if finalizer is added when an Ingress resource is created with finalizer enabled
```
MrHohn: nit: comment should end with a period ".".

pkg/controller/controller_test.go
```diff
@@ -281,6 +282,50 @@ func TestIngressCreateDeleteFinalizer(t *testing.T) {
...
+	}
+
+	updatedIng.ObjectMeta.Annotations = map[string]string{"kubernetes.io/ingress.class": "new-class"}
+	updateIngress(lbc, updatedIng)
+
+	if err := lbc.sync(ingStoreKey); err != nil {
+		t.Fatalf("lbc.sync(%v) = err %v", ingStoreKey, err)
+	}
+
+	// Check if finalizer is removed after class changes
```
MrHohn: nit: ditto

pkg/controller/controller_test.go
```diff
@@ -281,6 +282,50 @@ func TestIngressCreateDeleteFinalizer(t *testing.T) {
 	}
 }
 
+// TestIngressClassChangeWithFinalizer asserts that `sync` will not return an error for
+// a good ingress config status is updated and LB is deleted after class change.
+func TestIngressClassChangeWithFinalizer(t *testing.T) {
```
MrHohn: Thanks, good to have this behavior tested.

pkg/controller/controller.go
```diff
@@ -94,8 +93,8 @@ type LoadBalancerController struct {
 
 // NewLoadBalancerController creates a controller for gce loadbalancers.
 func NewLoadBalancerController(
-	ctx *context.ControllerContext,
-	stopCh chan struct{}) *LoadBalancerController {
+		ctx *context.ControllerContext,
+		stopCh chan struct{}) *LoadBalancerController {
```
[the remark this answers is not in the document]
skmatti: Removed

pkg/controller/controller.go

pkg/controller/controller_test.go
skmatti: Done

pkg/controller/controller_test.go
skmatti: Done

pkg/controller/controller_test.go
skmatti: ACKED

bowei:
/hold

If the Ingress has a finalizer, that means we have potentially allocated resources associated with the previous state of the Ingress. Which means we need to keep the finalizer...

rramkumar1:
@bowei If you look at line 496 of controller.go, you can see that if the class was changed from glbc to non-glbc, then we add the Ingress for garbage collection to ensure we clean stuff up. This was to fix some bug not related to finalizer support. 

Adding on that, I think we should add all non-glbc Ingresses for garbage collection always (rather than just when the class is changed) as a sanity check that stuff is getting cleaned up.

bowei:
Discussed offline -- looks like the finalizer handling is currently incorrect and needs to be fixed.

skmatti:
To summarize whats been added in the latest commits,
1. Handle resource deletions in update event handler and ignore queuing delete events when finalizer is enabled.
2. Do not re-queue backend resource delete requests if the backend resource is being utilized by another Ingress resource.

pkg/backends/backends.go
```diff
@@ -123,7 +123,9 @@ func (b *Backends) Delete(name string) (err error) {
 	klog.V(2).Infof("Deleting backend service %v", name)
 
 	// Try deleting health checks even if a backend is not found.
-	if err = b.cloud.DeleteGlobalBackendService(name); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+	// Ignore error if the deletion candidate is being used by another resource.
+	if err = b.cloud.DeleteGlobalBackendService(name); err != nil &&
+		!utils.IgnoreResourceBeingUsedError(err) && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 		return err
 	}
 	return
```
MrHohn: Probably explicitly return nil here? The defer func above seems error prone (it doesn't rewrite err to nil in case of in-used-by error)...

pkg/backends/backends.go
```diff
@@ -123,7 +123,9 @@ func (b *Backends) Delete(name string) (err error) {
 	klog.V(2).Infof("Deleting backend service %v", name)
 
 	// Try deleting health checks even if a backend is not found.
-	if err = b.cloud.DeleteGlobalBackendService(name); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+	// Ignore error if the deletion candidate is being used by another resource.
+	if err = b.cloud.DeleteGlobalBackendService(name); err != nil &&
+		!utils.IgnoreResourceBeingUsedError(err) && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
```
MrHohn: Is it worth to at least log the error (at a higher verbose level, e.g. V(4)) if we gonna eat it?

pkg/controller/controller.go
```diff
@@ -145,6 +144,12 @@ func NewLoadBalancerController(
 				return
 			}
 
+			// Update event takes of deletion when finalizer is enabled. Do not need to invoke delete
```
MrHohn: nit: takes care of?

pkg/controller/controller.go
```diff
@@ -145,6 +144,12 @@ func NewLoadBalancerController(
...
+			// explicitly.
+			if flags.F.FinalizerRemove {
```
MrHohn: We can do this even before the ingress class check?

pkg/controller/controller.go
```diff
@@ -145,6 +144,12 @@ func NewLoadBalancerController(
...
+				return
```
MrHohn: To confirm - we are attaching finalizer to the existing ingresses via the `AddFunc` handler upon controller upgrade, so we are sure their deletion can be handled via the `UpdateFunc` handler instead of here, right? (I haven't look carefully :) )

pkg/utils/utils.go
```diff
@@ -396,3 +399,42 @@ func TraverseIngressBackends(ing *extensions.Ingress, process func(id ServicePor
...
+		return []ServicePort{}
+	}
+
+	return knownPorts
+}
+
+// IgnoreResourceBeingUsedError returns true if the error is due to resource being
+// used by another resource.
+func IgnoreResourceBeingUsedError(err error) bool {
+	return strings.Contains(err.Error(), ResourceBeingUsedKey)
```
MrHohn: We can possibly check also on the error code to be sure?

pkg/controller/controller.go
skmatti: Yeah, but I think this the above log message would be appropriate for non-GLBC resources with finalizer enabled.

pkg/controller/controller.go
skmatti: Yes, that's correct.

pkg/controller/controller.go
skmatti: Thanks

pkg/backends/backends.go
skmatti: Added

pkg/utils/utils.go
skmatti: Done

pkg/backends/backends.go
```diff
@@ -123,10 +123,18 @@ func (b *Backends) Delete(name string) (err error) {
...
+	// Ignore error if the deletion candidate is being used by another resource.
+	if utils.IgnoreResourceBeingUsedError(err) {
+		klog.V(4).Infof("Ignoring delete, backend service %v is being used by another resource", name)
+		return nil
+	}
+
+	if err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 		return err
 	}
-	return
```
MrHohn: Can we somehow get rid of that defer func? Like:
```golang
func (b *Backends) Delete(name string) error {
	klog.V(2).Infof("Deleting backend service %v", name)

	// Try deleting health checks even if a backend is not found.
	if err := b.cloud.DeleteGlobalBackendService(name); err != nil {
		if !utils.IsHTTPErrorCode(err, http.StatusNotFound) &&
			!utils.IgnoreResourceBeingUsedError(err) {
			return err
		}
		klog.V(4).Infof("Ignoring deletion error for backend service %q: %v", name, err)

	}
	return nil
}
```

pkg/controller/controller.go
MrHohn: My impression is that ingress controller won't care about any ingress deletion (gce or non-gce) when finalizer is enabled. Eventually the entire `DeleteFunc` handler can be removed?

pkg/healthchecks/healthchecks.go
```diff
@@ -218,7 +218,14 @@ func (h *HealthChecks) getHealthCheckLink(name string, version meta.Version) (st
...
 	klog.V(2).Infof("Deleting health check %v", name)
-	return h.cloud.DeleteHealthCheck(name)
+	err := h.cloud.DeleteHealthCheck(name)
+
+	// Ignore error if the deletion candidate is being used by another resource.
+	if utils.IgnoreResourceBeingUsedError(err) {
+		klog.V(4).Infof("Ignoring delete, health check %v is being used by another resource", name)
+		return nil
+	}
+	return err
```
MrHohn: It is probably more intuitive to return nil at the end?
```golang
func (h *HealthChecks) Delete(name string) error {
	klog.V(2).Infof("Deleting health check %v", name)

	if err := h.cloud.DeleteHealthCheck(name); err != nil {
	 	// Ignore error if the deletion candidate is being used by another resource.
		if !utils.IgnoreResourceBeingUsedError(err) {
			return err
		}
		klog.V(4).Infof("Ignoring deletion error for health check %q: %v", name, err)
	}
	return nil

}
```

pkg/utils/utils.go
```diff
@@ -396,3 +399,43 @@ func TraverseIngressBackends(ing *extensions.Ingress, process func(id ServicePor
...
+	if len(knownPorts) == 0 {
+		return []ServicePort{}
+	}
+
+	return knownPorts
+}
+
+// IgnoreResourceBeingUsedError returns true if the error is due to resource being
+// used by another resource.
+func IgnoreResourceBeingUsedError(err error) bool {
```
MrHohn: Seems like we already have a similar function `IsInUsedByError()` above?

pkg/controller/controller.go
```diff
@@ -478,6 +484,15 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	gcState := &gcState{lbc.ctx.Ingresses().List(), lbNames, gceSvcPorts}
 	if !ingExists || utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey) {
 		klog.V(2).Infof("Ingress %q no longer exists, triggering GC", key)
+		// Force clean up only if Ingress exists.
```
MrHohn: What is the reason behind this? Does it mean the cleanup for ingress without finalizer will take place else where? Do we need to gate such behavior?

skmatti: The resources are excluded from gcState already if ingress does not exist or when requested a delete event. So, we don't need to remove them explicitly.

pkg/utils/utils.go
skmatti: Good find :)

pkg/controller/controller.go
skmatti: > My impression is that ingress controller won't care about any ingress deletion (gce or non-gce) when finalizer is enabled. Eventually the entire DeleteFunc handler can be removed?

I think we still have to handle GLBC resource deletions without finalizer.

pkg/backends/backends.go
skmatti: Done

pkg/controller/controller.go
```diff
@@ -390,7 +396,7 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
 	}
 
 	for _, ing := range gcState.ingresses {
-		if utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey) {
+		if utils.NeedsCleanup(ing) {
 			ingClient := lbc.ctx.KubeClient.ExtensionsV1beta1().Ingresses(ing.Namespace)
 			if flags.F.FinalizerRemove {
```
MrHohn: Seems like this check could be done outside of the for loop?

MrHohn: Also, took a deeper look, I think it'd be better to move all these finalizer logic (probably wrapped in another function) outside of `GCBackends()` to `GC()`:
https://github.com/kubernetes/ingress-gce/blob/e906f93959ede5a745ead5c7330a27e5f29a5e2c/pkg/sync/sync.go#L59-L70

We should only remove finalizer if both `GCLoadBalancers` and `GCBackends` return no error.

pkg/controller/controller.go
MrHohn: Quickly relay from the discussion offline, the idea is that once the finalizer logic is enabled, all gce ingresses will be on board to use the finalizer model and handled via the updateFunc. However this is fine as is as it ain't not much difference.

pkg/controller/controller.go
```diff
@@ -495,8 +510,12 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	// Check if ingress class was changed to non-GLBC to remove ingress LB from state and trigger GC
 	if !utils.IsGLBCIngress(ing) {
 		klog.V(2).Infof("Ingress %q class was changed, triggering GC", key)
-		// Remove lb from state for GC
-		gcState.lbNames = slice.RemoveString(gcState.lbNames, key, nil)
+		// Remove lb from state for GC.
+		gcState.lbNames = slice.RemoveString(lbNames, key, nil)
+		// Remove current Ingress's SVC ports from state for GC.
+		ingSvcPorts := lbc.ToSvcPorts([]*extensions.Ingress{ing})
+		gcState.svcPorts = utils.RemoveSvcPorts(gcState.svcPorts, ingSvcPorts)
```
MrHohn: Humm, can we be sure that these SvcPorts are no longer needed? It may possibly be in used by other ingresses?

pkg/controller/controller.go
skmatti: Because, there is one more GC at line 544 towards the end of the `sync` function. Its probably a good idea to create two separate gcStates so we dont have to exclude resources from the current Ingress?.

pkg/controller/controller.go
skmatti: The deletes will be skipped if there are being used by other Ingress resources.

pkg/controller/controller.go
```diff
@@ -636,8 +635,30 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
+// createGCState constructs GC State for GC.
+func (lbc *LoadBalancerController) createGCState(needsCleanup bool, ing *extensions.Ingress) *gcState {
+	var glbcIngresses []*extensions.Ingress
+	if needsCleanup {
+		glbcIngresses = operator.Ingresses(lbc.ctx.Ingresses().List()).Filter(func(ingress *extensions.Ingress) bool {
+			return utils.IsGLBCIngress(ingress) && !utils.IsDeletionCandidate(ingress.ObjectMeta, utils.FinalizerKey)
+		}).AsList()
+	} else {
+		glbcIngresses = operator.Ingresses(lbc.ctx.Ingresses().List()).Filter(func(ingress *extensions.Ingress) bool {
+			return utils.IsGLBCIngress(ingress)
```
MrHohn: Just trying to shorten this, what about:
```golang
filterFunc := func(ingress *extensions.Ingress) bool {
	if needsCleanup {
		return utils.IsGLBCIngress(ingress) && !utils.IsDeletionCandidate(ingress.ObjectMeta, utils.FinalizerKey)
	}
	return utils.IsGLBCIngress(ingress)
}
glbcIngresses := operator.Ingresses(lbc.ctx.Ingresses().List()).Filter(filterFunc).AsList()
```

pkg/controller/controller.go
```diff
@@ -388,19 +392,28 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
...
+	gcState, ok := state.(*gcState)
+	if !ok {
+		return fmt.Errorf("expected state type to be gcState, type was %T", state)
+	}
+
+	if flags.F.FinalizerRemove {
+		for _, ing := range gcState.ingresses {
+			if utils.NeedsToCleanupFinalizer(ing) {
+				ingClient := lbc.ctx.KubeClient.ExtensionsV1beta1().Ingresses(ing.Namespace)
 				if err := utils.RemoveFinalizer(ing, ingClient); err != nil {
```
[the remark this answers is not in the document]
skmatti: It's already logged in `RemoveFinalizer` method.

pkg/controller/controller.go
```diff
@@ -388,19 +392,28 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
...
 					klog.Errorf("Failed to remove Finalizer from Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
```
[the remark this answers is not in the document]
skmatti: This is the only place where we are logging finalizer removal error. A level above, this is logged as GC failure.

pkg/utils/utils.go
```diff
@@ -396,3 +396,26 @@ func TraverseIngressBackends(ing *extensions.Ingress, process func(id ServicePor
 func ServiceKeyFunc(namespace, name string) string {
 	return fmt.Sprintf("%s/%s", namespace, name)
 }
+
+// NeedsToCleanupFinalizer returns true if finalizer should be removed.
+func NeedsToCleanupFinalizer(ingress *extensions.Ingress) bool {
```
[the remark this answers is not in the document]
skmatti: I can modify this to use single function using the following implementation:
```
func NeedsCleanup(ing *extensions.Ingress) bool {
	return IsDeletionCandidate(ing.ObjectMeta, utils.Finalizer) ||
                  !IsGLBCIngress(ing)
}
```
To check if the ingress needs to GCed.
`!ingExists || NeedsCleanup(ing)`

To check if finalizer needs to be removed:
`NeedsCleanup(ing) && utils.HasFinalizer(ing.ObjectMeta, utils.Finalizer)`

Not sure if this makes cleaner than before.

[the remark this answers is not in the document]
skmatti: Discussed offline to modify `NeedsToCleanupFinalizer` to use `NeedsCleanup` as they cannot be merged.

pkg/utils/utils.go
```diff
@@ -396,3 +396,25 @@ func TraverseIngressBackends(ing *extensions.Ingress, process func(id ServicePor
...
+	// Remove finalizer if the ingress is
+	// 1. deleted or
```
MrHohn: nit (to match the actual condition): deleted with a finalizer?

pkg/controller/controller.go
```diff
@@ -636,8 +631,28 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
+		}
 	}
 	return knownPorts
 }
+
+// createGCState constructs GC State for GC.
+func (lbc *LoadBalancerController) createGCState(needsCleanup bool, ing *extensions.Ingress) *gcState {
+	filterFunc := func(ing *extensions.Ingress) bool {
+		if needsCleanup {
+			return utils.IsGLBCIngress(ing) && !utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey)
```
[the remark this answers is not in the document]
MrHohn: Also, does `return !NeedsCleanup(ing)` work here? :)

pkg/utils/utils_test.go
```diff
@@ -455,3 +457,171 @@ func TestTraverseIngressBackends(t *testing.T) {
...
+			expectNeedsCleanup:  true,
+		},
+		{
+			desc:                "GCE Ingress without timestamp",
+			isGLBCIngress:       true,
+			isDeletionCandidate: false,
+			expectNeedsCleanup:  false,
+		},
+		{
+			desc:                "GCE Ingress with ingExists with timestamp",
```
MrHohn: nit: "GCE Ingress with timestamp"?

pkg/controller/controller_test.go
```diff
@@ -281,6 +283,154 @@ func TestIngressCreateDeleteFinalizer(t *testing.T) {
...
+	}
+
+	// Ensure that no finalizer is added.
+	updatedIng, _ := lbc.ctx.KubeClient.ExtensionsV1beta1().Ingresses(ing.Namespace).Get(ing.Name, meta_v1.GetOptions{})
+	if len(updatedIng.GetFinalizers()) != 0 {
+		t.Fatalf("Unexpected Finalizer for Ingress: %v", ingStoreKey)
+	}
+
+	// enable finalizer
+	flags.F.FinalizerAdd = true
```
MrHohn: Should we defer a `flags.F.FinalizerAdd = false` to reset this config to avoid potentially affecting tests that run after this?

pkg/controller/controller.go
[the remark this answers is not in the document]
skmatti: Good find :)

MrHohn:
Would be great to have @nikhiljindal's feedback around the GC logic w.r.t kubemci usage :)

pkg/controller/controller.go
```diff
@@ -636,8 +635,30 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
 func (lbc *LoadBalancerController) ToSvcPorts(ings []*extensions.Ingress) []utils.ServicePort {
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
-		urlMap, _ := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPortID)
-		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		if utils.IsGCEIngress(ing) {
```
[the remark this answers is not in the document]
MrHohn: Discussed offline with Nikhil and Satish, it turned out the only resource kubemci needs from gce ingress controller is the instance group (shared among all the ingresses). As long as the multi-cluster class ingress is included in the GC state we are good. So this looks good now :)

MrHohn:
/lgtm

Would be great to squash the fix-up commits, thanks!

pkg/controller/controller.go
```diff
@@ -147,6 +143,13 @@ func NewLoadBalancerController(
 				return
 			}
 
+			// Update takes care of deletion when finalizer is enabled.
+			// Do not need to invoke delete explicitly.
+			if flags.F.FinalizerRemove {
```
bowei: [important] I'm not sure this will work interoperate with Ingress objects that have already exist that have not had a chance to add a finalizer to the object. We probably need to also trigger on the deletion event as well to be safe.

pkg/controller/controller.go
```diff
@@ -390,19 +393,28 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
...
-			if flags.F.FinalizerRemove {
+// RemoveFinalizers cleans up Finalizers.
+func (lbc *LoadBalancerController) RemoveFinalizers(state interface{}) error {
+	// We expect state to be a gcState
+	gcState, ok := state.(*gcState)
+	if !ok {
+		return fmt.Errorf("expected state type to be gcState, type was %T", state)
+	}
+
+	if flags.F.FinalizerRemove {
```
bowei: shouldn't this be at the top
`if ! flags.F.FinalizerRemove { return nil }`

pkg/controller/controller.go
```diff
@@ -390,19 +393,28 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
...
+		for _, ing := range gcState.ingresses {
+			if utils.NeedsToCleanupFinalizer(ing) {
```
bowei:
```
if ! utils.NeedsToCleanup...() {
  continue
}
```

pkg/controller/controller.go
```diff
@@ -639,8 +632,37 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
+
+	return &gcState{lbc.ctx.Ingresses().List(), lbNames, gceSvcPorts}
+}
+
+// toLbNames translates a list of ingresses into their corresponding load balancer names.
+func toLbNames(ings []*v1beta1.Ingress) []string {
+	lbNames := make([]string, 0, len(ings))
+	for _, ing := range ings {
+		// Translate only Gce Ingresses as multi-cluster Ingress lbs are not owned by LoadBalancerController.
+		if utils.IsGCEIngress(ing) {
```
bowei: Didn't you already filter in the caller? this seems to be redundant.

pkg/controller/controller.go
```diff
@@ -639,8 +632,37 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
+	// Filter Ingresses for GC state.
+	glbcIngresses := operator.Ingresses(lbc.ctx.Ingresses().List()).Filter(func(ing *v1beta1.Ingress) bool {
+		return !utils.NeedsCleanup(ing)
+	}).AsList()
+
+	// gvcPorts contains the ServicePorts used by only single-cluster ingress.
+	gceSvcPorts := lbc.ToSvcPorts(glbcIngresses)
+	lbNames := toLbNames(glbcIngresses)
+
+	return &gcState{lbc.ctx.Ingresses().List(), lbNames, gceSvcPorts}
```
bowei: why is this not glbcIngresses

pkg/controller/controller_test.go
```diff
@@ -282,6 +284,162 @@ func TestIngressCreateDeleteFinalizer(t *testing.T) {
 	}
 }
 
+// TestIngressClassChangeWithFinalizer asserts that `sync` will not return an error for
+// a good ingress config status is updated and LB is deleted after class change.
+func TestIngressClassChangeWithFinalizer(t *testing.T) {
+	flags.F.FinalizerAdd = true
```
bowei: // This test cannot be run in parallel as it stubs global flags.`

pkg/healthchecks/healthchecks.go
```diff
@@ -293,7 +297,14 @@ func (h *HealthChecks) Delete(name string, scope meta.KeyType) error {
 
 	klog.V(2).Infof("Deleting health check %v", name)
 	// Not using composite here since the tests still rely on the fake health check interface
-	return h.cloud.DeleteHealthCheck(name)
+	if err := h.cloud.DeleteHealthCheck(name); err != nil {
+		// Ignore error if the deletion candidate is being used by another resource.
+		if !utils.IsInUsedByError(err) {
+			return err
+		}
+		klog.V(4).Infof("Ignoring deletion error for health check %v: %v", name, err)
```
bowei: log looks to be in the wrong place

pkg/utils/utils.go
```diff
@@ -424,3 +424,16 @@ func TraverseIngressBackends(ing *v1beta1.Ingress, process func(id ServicePortID
 func ServiceKeyFunc(namespace, name string) string {
 	return fmt.Sprintf("%s/%s", namespace, name)
 }
+
+// NeedsToCleanupFinalizer returns true if finalizer should be removed.
```
bowei: you should be consistent in language
either cleanup or remove, but don't mix them

`NeedsToRemoveFinalizer`

pkg/utils/utils.go
```diff
@@ -424,3 +424,16 @@ func TraverseIngressBackends(ing *v1beta1.Ingress, process func(id ServicePortID
...
+func NeedsToCleanupFinalizer(ingress *v1beta1.Ingress) bool {
+	// Remove finalizer if the ingress is
+	// 1. deleted with a finalizer or
+	// 2. a non-GLBC resource with a finalizer
+	return NeedsCleanup(ingress) && HasFinalizer(ingress.ObjectMeta, FinalizerKey)
+}
+
+// NeedsCleanup returns true if the ingress needs to be excluded from GC State.
```
bowei: This comment is very hard to understand without knowing about the rest of the code.
It looks like this returns true if the object needs to have its associated resources deleted, is that accurate?

pkg/utils/utils_test.go
```diff
@@ -664,3 +665,171 @@ func TestIsGCEL7ILBIngress(t *testing.T) {
...
+func TestNeedsToCleanUpFinalizer(t *testing.T) {
+	testCases := []struct {
+		desc                string
+		ingressClass        string
+		withFinalizer       bool
+		withDeleteTimestamp bool
+		expectNeedsCleanup  bool
+	}{
+		{
+			desc:                "GCE Ingress without finalizer without timestamp",
```
bowei: you can make this much easier to maintain by generating desc from the testcase

`desc := fmt.Sprintf("ingress class %s withFinalizer %b ...", ...)`

Also, probably just write this out short format to make it easier to see test case coverage
```
testCases := []struct {
...
} {
  {"gce", false, true, false"},
  ...
}
```

pkg/utils/utils_test.go
```diff
@@ -664,3 +665,171 @@ func TestIsGCEL7ILBIngress(t *testing.T) {
...
+}
+
+func TestNeedsCleanup(t *testing.T) {
+	testCases := []struct {
+		desc                string
+		isGLBCIngress       bool
+		isDeletionCandidate bool
+		expectNeedsCleanup  bool
+	}{
+		{
```
bowei: see above

pkg/backends/backends.go
```diff
@@ -114,19 +114,17 @@ func (b *Backends) Get(name string, version meta.Version) (*composite.BackendSer
...
-			err = nil
-		}
-	}()
-
 	klog.V(2).Infof("Deleting backend service %v", name)
 
 	// Try deleting health checks even if a backend is not found.
-	if err = b.cloud.DeleteGlobalBackendService(name); err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
-		return err
+	if err = b.cloud.DeleteGlobalBackendService(name); err != nil {
```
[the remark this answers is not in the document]
skmatti: Done

pkg/backends/backends.go
```diff
@@ -1,9 +1,12 @@
 /*
 Copyright 2018 The Kubernetes Authors.
+
```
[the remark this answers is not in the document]
skmatti: Done

pkg/backends/backends.go
```diff
@@ -140,12 +143,6 @@ func (b *Backends) Get(name string, version meta.Version, scope meta.KeyType) (*
 
 // Delete implements Pool.
 func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType) (err error) {
-	defer func() {
```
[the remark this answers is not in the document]
skmatti: Done

pkg/backends/backends.go
```diff
@@ -158,12 +155,15 @@ func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType)
 		if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
 			klog.Infof("DeleteBackendService(_, %v, %v) = %v; backend service does not exists, ignoring", key, version, err)
 			return nil
+		} else if utils.IsInUsedByError(err) {
```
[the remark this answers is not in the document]
skmatti: Done

pkg/controller/controller.go
```diff
@@ -37,23 +33,24 @@ import (
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+
```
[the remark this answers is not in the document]

pkg/controller/controller.go
```diff
@@ -147,6 +143,13 @@ func NewLoadBalancerController(
 				return
 			}
 
+			// Update takes care of deletion when finalizer is enabled.
```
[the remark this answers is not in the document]

pkg/controller/controller.go
```diff
@@ -147,6 +143,13 @@ func NewLoadBalancerController(
...
+			// Do not need to invoke delete explicitly.
+			if flags.F.FinalizerRemove {
+				klog.V(4).Infof("Ignoring delete, Resources are already deleted via an Update")
```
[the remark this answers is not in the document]

pkg/controller/controller.go
skmatti: Discussed offline with Bowei and Zhihong about this. We can ignore delete event only if the resource has a deletion timestamp in which case deletion is handled by the finalizer.

pkg/controller/controller.go

pkg/controller/controller.go

pkg/controller/controller.go
skmatti: Filtered only GLBC resources in the caller. I think this function should take care of excluding multi-cluster Ingress resources.

pkg/controller/controller.go
skmatti: This list is used to remove Finalizers after the resources are cleaned up. This ensures removing finalizers of non-glbc resources also.

pkg/controller/controller_test.go
skmatti: Done

pkg/healthchecks/healthchecks.go
skmatti: It's in the right place. but I think its confusing, rewritten this.

pkg/utils/utils.go
skmatti: Fixed

pkg/utils/utils.go
skmatti: Yes, that's right. Fixed the comment.

pkg/utils/utils_test.go
skmatti: Done

pkg/utils/utils_test.go
skmatti: Done

pkg/controller/controller.go
```diff
@@ -142,6 +137,11 @@ func NewLoadBalancerController(
 		},
 		DeleteFunc: func(obj interface{}) {
 			delIng := obj.(*v1beta1.Ingress)
+			if delIng.ObjectMeta.DeletionTimestamp != nil {
```
bowei:
```
if delIng == nil {
  klog.Errorf("Invalid object type: %T", obj)
  return
}
```

pkg/controller/controller.go
```diff
@@ -464,22 +475,15 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
-
 	ing, ingExists, err := lbc.ctx.Ingresses().GetByKey(key)
 	if err != nil {
 		return fmt.Errorf("error getting Ingress for key %s: %v", key, err)
 	}
-	gcState := &gcState{lbc.ctx.Ingresses().List(), lbNames, gceSvcPorts}
-	if !ingExists || utils.IsDeletionCandidate(ing.ObjectMeta, utils.FinalizerKey) {
-		klog.V(2).Infof("Ingress %q no longer exists, triggering GC", key)
+
+	// Create GC State needed for GC.
```
bowei: delete comment, redundant

pkg/controller/controller.go
```diff
@@ -639,8 +631,36 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
+			urlMap, _ := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPort.ID)
+			knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
+		}
 	}
 	return knownPorts
 }
+
+// createGCState constructs GC State for GC.
+func (lbc *LoadBalancerController) createGCState() *gcState {
+	// Filter Ingresses for GC state.
```
bowei:
```
all := lbc.ctx.Ingresses().List()
toKeep := all.Filter(func(ing *v1beta1.Ingress) bool {￼ return !utils.NeedsCleanup(ing)￼ }).AsList()
// gcState needs to have all three fields be a snapshot of the current state to avoid inconsistency.
return &gcState{
  ingresses: all.List(),
  ...: toLbNames(toKeep),
  ...: lbc.ToSvcPorts(toKeep),
}
```

pkg/backends/backends.go
```diff
@@ -155,15 +149,15 @@ func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType)
 	}
 	err = composite.DeleteBackendService(b.cloud, key, version)
 	if err != nil {
-		if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
-			klog.Infof("DeleteBackendService(_, %v, %v) = %v; backend service does not exists, ignoring", key, version, err)
+		if utils.IsHTTPErrorCode(err, http.StatusNotFound) || utils.IsInUsedByError(err) {
```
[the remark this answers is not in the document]
skmatti: Removed it

pkg/controller/controller.go
```diff
@@ -142,6 +137,11 @@ func NewLoadBalancerController(
 		},
 		DeleteFunc: func(obj interface{}) {
 			delIng := obj.(*v1beta1.Ingress)
+			if delIng.ObjectMeta.DeletionTimestamp != nil {
+				klog.V(4).Infof("Ignoring delete event as resource deletion will be handled via the finalizer")
```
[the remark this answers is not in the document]

pkg/controller/controller.go
```diff
@@ -390,19 +390,30 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
...
 	}
+	return nil
+}
+
+// RemoveFinalizers cleans up Finalizers.
+func (lbc *LoadBalancerController) RemoveFinalizers(state interface{}) error {
+	if !flags.F.FinalizerRemove {
+		return nil
+	}
+	// We expect state to be a gcState
```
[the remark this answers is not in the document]

pkg/controller/controller.go
```diff
@@ -390,19 +390,30 @@ func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
 			return err
 		}
 	}
+	return nil
+}
+
+// RemoveFinalizers cleans up Finalizers.
+func (lbc *LoadBalancerController) RemoveFinalizers(state interface{}) error {
+	if !flags.F.FinalizerRemove {
+		return nil
```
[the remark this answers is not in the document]

pkg/controller/controller.go
skmatti: Added

pkg/controller/controller.go

pkg/controller/controller.go

pkg/controller/controller.go
```diff
@@ -644,3 +636,31 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+
+	// gcState needs to have all three fields be a snapshot of the current state to avoid inconsistency.
+	return &gcState{toCleanup, toKeepLbNames, toKeepSvcPorts}
+}
+
+// toLbNames translates a list of ingresses into their corresponding load balancer names.
+func toLbNames(ings []*v1beta1.Ingress) []string {
+	lbNames := make([]string, 0, len(ings))
+	for _, ing := range ings {
+		lbNames = append(lbNames, utils.IngressKeyFunc(ing))
```
skmatti: IngressKeyFunc used here is incorrect. It should be replaced with KeyFunc defined [here](https://github.com/skmatti/ingress-gce/blob/e23564d807c2c6d85e0e5745cef4ee0b9d540b7b/pkg/utils/taskqueue.go#L26).

PR #830, opened 2019-08-21
Add NEG keyword to function name for Readiness Reflector e2e test

rramkumar1:
This is so that an external test framework can automatically pick up the test using the regex "NEG"


freehan:
No. This one is sort of new. Previous glbc version with NEG but not readiness reflect. Hence I did not add the key word in it.

Issue #836, opened 2019-08-28
Fix or verify the GC logic for multi-cluster ingress


MrHohn:
Ref https://github.com/kubernetes/ingress-gce/pull/825#discussion_r318355014, we may have broke the GC logic for multi-cluster ingress --- we need to keep the instance group if we have at least one multi-cluster ingress. The current logic may delete the instance group if there isn't any gce ingresses (regardless of multi-cluster ingress).

(By multi-cluster ingress I meant below)
https://github.com/kubernetes/ingress-gce/blob/78801b3ab962310b55c6565a7164752044391edf/pkg/utils/utils.go#L288-L293

/assign @skmatti

PR #838, opened 2019-08-29
Bugfix: don't GC instance groups while multi-cluster ingresses exist

skmatti:
Fixes https://github.com/kubernetes/ingress-gce/issues/836

Existing GC workflow deletes instance groups as soon all GCE ingresses are deleted. But, instance groups should be retained as they are shared by multi-cluster ingresses as well.

### DIFF
--- a/pkg/common/operator/ingress.go
+++ b/pkg/common/operator/ingress.go
@@ -39,6 +39,20 @@ func (op *IngressesOperator) Filter(f func(*v1beta1.Ingress) bool) *IngressesOpe
 	return Ingresses(i)
 }
 
+// Partition partitions the list of Ingresses into two: Ingresses that match given predicate and those that don't.
+func (op *IngressesOperator) Partition(f func(*v1beta1.Ingress) bool) (*IngressesOperator, *IngressesOperator) {
+	// A list of ingresses that satisfy the predicate condition and it's complement
+	var i, ci []*v1beta1.Ingress
+	for _, ing := range op.i {
+		if f(ing) {
+			i = append(i, ing)
+		} else {
+			ci = append(ci, ing)
+		}
+	}
+	return Ingresses(i), Ingresses(ci)
+}
+
 // ReferencesService returns the Ingresses that references the given Service.
 func (op *IngressesOperator) ReferencesService(svc *api_v1.Service) *IngressesOperator {
 	dupes := map[string]bool{}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -375,43 +375,16 @@ func (lbc *LoadBalancerController) SyncBackends(state interface{}) error {
 }
 
 // GCBackends implements Controller.
-func (lbc *LoadBalancerController) GCBackends(state interface{}) error {
-	// We expect state to be a gcState
-	gcState, ok := state.(*gcState)
-	if !ok {
-		return fmt.Errorf("expected state type to be gcState, type was %T", state)
-	}
-
-	if err := lbc.backendSyncer.GC(gcState.svcPortsToKeep); err != nil {
+func (lbc *LoadBalancerController) GCBackends(toKeep []*v1beta1.Ingress) error {
+	svcPortsToKeep := lbc.ToSvcPorts(toKeep)
+	if err := lbc.backendSyncer.GC(svcPortsToKeep); err != nil {
 		return err
 	}
-
 	// TODO(ingress#120): Move this to the backend pool so it mirrors creation
-	if len(gcState.lbNamesToKeep) == 0 {
+	if len(toKeep) == 0 {
 		igName := lbc.ctx.ClusterNamer.InstanceGroup()
 		klog.Infof("Deleting instance group %v", igName)
 		if err := lbc.instancePool.DeleteInstanceGroup(igName); err != err {
-			return err
-		}
-	}
-	return nil
-}
-
-// MaybeRemoveFinalizers cleans up Finalizers if needed.
-func (lbc *LoadBalancerController) MaybeRemoveFinalizers(state interface{}) error {
-	if !flags.F.FinalizerRemove {
-		klog.V(4).Infof("Removing finalizers not enabled")
-		return nil
-	}
-	gcState, ok := state.(*gcState)
-	if !ok {
-		return fmt.Errorf("expected state type to be gcState, type was %T", state)
-	}
-
-	for _, ing := range gcState.ingressesToCleanup {
-		ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
-		if err := utils.RemoveFinalizer(ing, ingClient); err != nil {
-			klog.Errorf("Failed to remove Finalizer from Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
 			return err
 		}
 	}
@@ -442,17 +415,23 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
 }
 
 // GCLoadBalancers implements Controller.
-func (lbc *LoadBalancerController) GCLoadBalancers(state interface{}) error {
-	// We expect state to be a gcState
-	gcState, ok := state.(*gcState)
-	if !ok {
-		return fmt.Errorf("expected state type to be gcState, type was %T", state)
-	}
+func (lbc *LoadBalancerController) GCLoadBalancers(toKeep []*v1beta1.Ingress) error {
+	return lbc.l7Pool.GC(toLbNames(toKeep))
+}
 
-	if err := lbc.l7Pool.GC(gcState.lbNamesToKeep); err != nil {
-		return err
+// MaybeRemoveFinalizers cleans up Finalizers if needed.
+func (lbc *LoadBalancerController) MaybeRemoveFinalizers(toCleanup []*v1beta1.Ingress) error {
+	if !flags.F.FinalizerRemove {
+		klog.V(4).Infof("Removing finalizers not enabled")
+		return nil
 	}
-
+	for _, ing := range toCleanup {
+		ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+		if err := utils.RemoveFinalizer(ing, ingClient); err != nil {
+			klog.Errorf("Failed to remove Finalizer from Ingress %s/%s: %v", ing.Namespace, ing.Name, err)
+			return err
+		}
+	}
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -442,17 +415,43 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
 }
 
 // GCLoadBalancers implements Controller.
-func (lbc *LoadBalancerController) GCLoadBalancers(state interface{}) error {
+func (lbc *LoadBalancerController) GCLoadBalancers(toKeep []*v1beta1.Ingress) error {
+	lbNamesToKeep := toLbNames(toKeep)
+	if err := lbc.l7Pool.GC(lbNamesToKeep); err != nil {
```
MrHohn: nit:
```golang
return lbc.l7Pool.GC(lbNamesToKeep)
```

pkg/sync/sync.go
```diff
@@ -58,14 +58,19 @@ func (s *IngressSyncer) Sync(state interface{}) error {
 
 // GC implements Syncer.
 func (s *IngressSyncer) GC(state interface{}) error {
```
MrHohn: I wonder is it still necessary to take state as a `interface{}` here? It seems like all other GC funcs below already take specific type, why bother keep the extra type casting here?

MrHohn: It is also confusing because this `state interface{}` is a different one than what `Sync()` takes.

pkg/common/operator/ingress.go
```diff
@@ -39,6 +39,20 @@ func (op *IngressesOperator) Filter(f func(*v1beta1.Ingress) bool) *IngressesOpe
 	return Ingresses(i)
 }
 
+// Partition partitions the list of Ingresses based on a predicate.
+func (op *IngressesOperator) Partition(f func(*v1beta1.Ingress) bool) ([]*v1beta1.Ingress, []*v1beta1.Ingress) {
+	// a list of ingresses that satisfy the predicate condition and it's complement
```
MrHohn: nit: A list ... and its complement.

pkg/sync/interfaces.go
```diff
@@ -32,14 +36,16 @@ type Syncer interface {
...
 	SyncBackends(state interface{}) error
-	// GCBackends garbage collects backends for all Ingresses given some existing state.
-	GCBackends(state interface{}) error
+	// GCBackends garbage collects backends for all ingresses given a list of ingresses to exclude from GC.
+	GCBackends(toKeep []*v1beta1.Ingress) error
 	// SyncLoadBalancer syncs the front-end load balancer resources for a GCLB given some existing state.
 	SyncLoadBalancer(state interface{}) error
-	// GCLoadBalancers garbage collects front-end load balancer resources for all Ingresses given some existing state.
-	GCLoadBalancers(state interface{}) error
+	// GCLoadBalancers garbage collects front-end load balancer resources for all ingresses given a list of ingresses exclude from GC.
```
MrHohn: nit: ... a list for ingresses to exclude from GC.

pkg/sync/interfaces.go
```diff
@@ -32,14 +36,16 @@ type Syncer interface {
...
+	GCLoadBalancers(toKeep []*v1beta1.Ingress) error
+	// ProcessGCState parses given GC state and splits GC state ingresses into two sets.
+	ProcessGCState(state interface{}) ([]*v1beta1.Ingress, []*v1beta1.Ingress, error)
 	// PostProcess allows for doing some post-processing after an Ingress is synced to a GCLB.
 	PostProcess(state interface{}) error
-	// MaybeRemoveFinalizers removes Finalizers after all the resources used by Ingresses are deleted.
-	MaybeRemoveFinalizers(state interface{}) error
+	// MaybeRemoveFinalizers removes finalizers from a list of ingresses one by one after all of it's associated resources are deleted.
```
MrHohn: nit: it's -> its

pkg/controller/controller.go
```diff
@@ -442,17 +415,43 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// ProcessGCState partitions a list of ingresses into a set of ingresses those need to be cleaned up
+// during GC and a set of those don't.
+func (lbc *LoadBalancerController) ProcessGCState(state interface{}) ([]*v1beta1.Ingress, []*v1beta1.Ingress, error) {
```
MrHohn: Seems like this doesn't have to be a method of `LoadBalancerController`?

MrHohn:
BTW should we fix the GC logic for L7ILB here as well? It may be as simple as tweaking these:
https://github.com/kubernetes/ingress-gce/blob/78801b3ab962310b55c6565a7164752044391edf/pkg/utils/utils.go#L288-L305

cc @spencerhance

pkg/sync/sync.go
skmatti: I fixed this so `GC` does not take an interface as input anymore. Not sure if we should be doing the same for `Sync` as well.

pkg/controller/controller.go
skmatti: Does not need this function anymore.

pkg/sync/sync.go
MrHohn: Thanks! I like that better.

MrHohn:
> BTW should we fix the GC logic for L7ILB here as well?

Discussed offline, seems like that works properly as is and doesn't need any fix :)

MrHohn:
Will let @bowei also take a look.
/approve

pkg/controller/controller.go
```diff
@@ -442,17 +415,24 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
...
 
 // GCLoadBalancers implements Controller.
-func (lbc *LoadBalancerController) GCLoadBalancers(state interface{}) error {
-	// We expect state to be a gcState
-	gcState, ok := state.(*gcState)
-	if !ok {
-		return fmt.Errorf("expected state type to be gcState, type was %T", state)
-	}
+func (lbc *LoadBalancerController) GCLoadBalancers(toKeep []*v1beta1.Ingress) error {
+	lbNamesToKeep := toLbNames(toKeep)
```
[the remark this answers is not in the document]
bowei: unless you wanted to log

pkg/controller/controller.go
```diff
@@ -442,17 +415,24 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
...
+// MaybeRemoveFinalizers cleans up Finalizers if needed.
+func (lbc *LoadBalancerController) MaybeRemoveFinalizers(toCleanup []*v1beta1.Ingress) error {
+	if !flags.F.FinalizerRemove {
+		klog.V(4).Infof("Removing finalizers not enabled")
+		return nil
+	}
+	for _, ing := range toCleanup {
+		ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+		if err := utils.RemoveFinalizer(ing, ingClient); err != nil {
+			klog.Errorf("Failed to remove Finalizer from Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
```
bowei: use %s/%s to match type

pkg/controller/controller.go
```diff
@@ -481,7 +461,8 @@ func (lbc *LoadBalancerController) sync(key string) error {
 		return fmt.Errorf("error getting Ingress for key %s: %v", key, err)
 	}
 
-	gcState := lbc.createGCState()
+	// Snapshot of list of ingresses.
+	gcState := lbc.ctx.Ingresses().List()
```
bowei: you should rename the variable as it is no longer gcState

pkg/controller/controller.go
```diff
@@ -631,35 +612,24 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
 func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.ServicePort {
 	var knownPorts []utils.ServicePort
 	for _, ing := range ings {
+		// Only resources associated with GCE Ingress are managed by this controller.
+		if !utils.IsGCEIngress(ing) {
```
bowei: Do we want to do the filter inside or outside this function?

pkg/controller/controller.go
```diff
@@ -631,35 +612,24 @@ func updateAnnotations(client kubernetes.Interface, name, namespace string, anno
...
-	// gcState needs to have all three fields be a snapshot of the current state to avoid inconsistency.
-	return &gcState{toCleanup, toKeepLbNames, toKeepSvcPorts}
-}
-
 // toLbNames translates a list of ingresses into their corresponding load balancer names.
 func toLbNames(ings []*v1beta1.Ingress) []string {
 	lbNames := make([]string, 0, len(ings))
 	for _, ing := range ings {
+		// Only resources associated with GCE Ingress are managed by this controller.
+		if !utils.IsGCEIngress(ing) {
```
bowei: see comment above

pkg/sync/interfaces.go
```diff
@@ -24,22 +28,22 @@ type Syncer interface {
...
 
 // Controller is an interface for ingress controllers and declares methods
 // on how to sync the various portions of the GCLB for an Ingress.
 type Controller interface {
 	// SyncBackends syncs the backends for a GCLB given some existing state.
 	SyncBackends(state interface{}) error
-	// GCBackends garbage collects backends for all Ingresses given some existing state.
-	GCBackends(state interface{}) error
+	// GCBackends garbage collects backends for all ingresses given a list of ingresses to exclude from GC.
+	GCBackends(toKeep []*v1beta1.Ingress) error
```
bowei: you might want to sync up with rohit about why this was interface

pkg/common/operator/ingress.go
```diff
@@ -39,6 +39,20 @@ func (op *IngressesOperator) Filter(f func(*v1beta1.Ingress) bool) *IngressesOpe
 	return Ingresses(i)
 }
 
+// Partition partitions the list of Ingresses based on a predicate.
+func (op *IngressesOperator) Partition(f func(*v1beta1.Ingress) bool) ([]*v1beta1.Ingress, []*v1beta1.Ingress) {
```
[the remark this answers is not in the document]
skmatti: Done

pkg/common/operator/ingress.go
```diff
@@ -39,6 +39,20 @@ func (op *IngressesOperator) Filter(f func(*v1beta1.Ingress) bool) *IngressesOpe
 	return Ingresses(i)
 }
 
+// Partition partitions the list of Ingresses based on a predicate.
```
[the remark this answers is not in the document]
skmatti: Done

pkg/controller/controller.go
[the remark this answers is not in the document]

pkg/controller/controller.go
skmatti: We either needed to have `toKeepGceIngresses` and `toKeepIngresses` both or handle this filter in the downstream. I think second one is straightforward. I improved the doc of both helper methods to specify that it returns a list of svc ports owned by this controller.

pkg/controller/controller.go

pkg/sync/interfaces.go
skmatti: Discussed offline with Rohit and interface is not required anymore.

bowei:
/lgtm
/approve

It seems strange that toCleanup is not related at all with the GC action (e.g. it is only after GC deletes the resources for a given Ingress that the finalizer is removed). It seems worthwhile to have a follow on change that makes it more explicit connection, as this will lead to bugs.

PR #839, opened 2019-08-29
Adding the ability of switch container apiendpoint to

cadmuxe:
docs/deploy/gke/gke-self-managed.sh.

### DIFF
--- a/docs/deploy/gke/gke-self-managed.sh
+++ b/docs/deploy/gke/gke-self-managed.sh
@@ -32,6 +32,7 @@ function usage() {
   echo "  --help                   Display this help and exit"
   echo "  --network-name           Override for the network-name gce.conf value"
   echo "  --enable-csm             If set, enable CSM mode which will create NEGs for Istio objects."
+  echo "  --gke-api-endpoint       If set, using the provided api-endpoint instead of prod."
   echo "  --subnetwork-name        Override for the subnetwork-name gce.conf value"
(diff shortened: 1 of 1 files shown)


docs/deploy/gke/gke-self-managed.sh
```diff
@@ -207,6 +208,11 @@ case $key in
   ENABLE_CSM=true
   shift
   ;;
+  --gke-api-endpoint)
+  export CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER=$2
```
freehan: if you export here, it will stick to the shell session right?

Then the person may get confused due to the override.

docs/deploy/gke/gke-self-managed.sh
```diff
@@ -32,6 +32,7 @@ function usage() {
   echo "  --help                   Display this help and exit"
   echo "  --network-name           Override for the network-name gce.conf value"
   echo "  --enable-csm             If set, enable CSM mode which will create NEGs for Istio objects."
+  echo "  --gke-api-endpoint       If set, using the provided api-endpoint instead of prod."
```
freehan: instead of the environment setting.

docs/deploy/gke/gke-self-managed.sh
cadmuxe: I think it only last in the script but not the shell session.

PR #841, opened 2019-09-04
try to spread the pods across zone to avoid test timeout

freehan:
Trying to fix flaky tests.

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -331,7 +331,16 @@ func TestNEGSyncEndpoints(t *testing.T) {
 
 			scaleAndValidate := func(replicas int32) {
 				t.Logf("Scaling echo deployment to %v replicas", replicas)
-				if err := e2e.EnsureEchoDeployment(s, svcName, replicas, e2e.NoopModify); err != nil {
+				// The deployment is created with pod anti affinity rules trying to spread the pods across zones.
+				// GCLB only creates the underlying infrastructure in each zone when there is at least one backend.
+				// Since this test tries to validate by sending traffic, it is essential that the LB backends are fully
+				// instantiated in all zones so that the new endpoints can show up faster before test timeout occur.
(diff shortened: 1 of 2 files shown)


cmd/e2e-test/neg_test.go
```diff
@@ -331,7 +331,16 @@ func TestNEGSyncEndpoints(t *testing.T) {
...
-				if err := e2e.EnsureEchoDeployment(s, svcName, replicas, e2e.NoopModify); err != nil {
+				// The deployment is created with pod anti affinity rules trying to spread the pods across zones.
+				// GCLB only creates the underlying infrastructure in each zone when there is at least one backend.
+				// Since this test tries to validate by sending traffic, it is essential that the LB backends are fully
+				// instantiated in all zones so that the new endpoints can show up faster before test timeout occur.
+				// If the LB backend need to be freshly setup when a new pod is scheduled to the zone, this may lead to
+				// test timeout as it takes more time for the pod to respond to traffic
+				// However, the anti affinity rule may not fully solve this problem in the case where there
+				// is no capacity left in all nodes in a zone. Hence, it may still cause all pods to be scheduled into
+				// other zones. A pod started later may get scheduled to a zone when capacity freed up.
```
MrHohn: Thanks for the detailed comments, just a nit --- seems like each line is a bit longer and shows up as two lines, maybe wrap them a bit further?

pkg/e2e/fixtures.go
```diff
@@ -29,18 +29,49 @@ import (
...
 	"k8s.io/ingress-gce/cmd/echo/app"
 	"k8s.io/klog"
 )
 
 const (
 	echoheadersImage = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
+
+	// TODO: remove beta topology key once it is GAed
+	zoneBetaTopologyKey = "failure-domain.beta.kubernetes.io/zone"
+	zoneGATopologyKey   = "failure-domain.beta.kubernetes.io/zone"
```
MrHohn: Why is this the same as the beta one?

PR #842, opened 2019-09-04
fix a typo

freehan:
follow up of #841


MrHohn:
Hold for using the vendored constants.

PR #824, opened 2019-08-13
Update e2e framework for ILB and add basic test

spencerhance:
Adds HasAlphaRegionResource() to handle L7-ILB and adds it as a feature.  Ideally we should switch e2e to use composites and ResourceVersions.

The actual tests will be added in a seperate PR.

PR #843, opened 2019-09-05
Update e2e framework for ILB - Part 1

spencerhance:
Replaces #824 

Updates the e2e test framework to have basic support for ILB (e.g. Supports path checking but doesn't support GCE whitebox or GC testing)

A follow up PR will be added to support those things

/assign @rramkumar1


spencerhance:
@rramkumar1 can you re-approve? I just had to run gofmt

PR #844, opened 2019-09-06
Add supplementary e2e tests for Finalizer

skmatti:
Adds the following:
1. Finalizer behavior with multiple ingresses sharing back-ends among themselves. 
2. Upgrade test: When master upgrades from a version without finalizer to a version with finalizer.


cmd/e2e-test/finalizer_test.go
```diff
@@ -39,122 +44,262 @@ func TestFinalizer(t *testing.T) {
...
+}
 
-		vip := ing.Status.LoadBalancer.Ingress[0].IP
-		t.Logf("Ingress %s/%s VIP = %s", s.Namespace, ing.Name, vip)
-		gclb, err := fuzz.GCLBForVIP(context.Background(), Framework.Cloud, vip, fuzz.FeatureValidators(features.All))
+// TestUpgradeTo1dot7 asserts that finalizer is added to an ingress when upgraded from a version
+// without finalizer to version 1.7. Also, verifies that ingress is deleted with finalizer enabled.
+// Note: The test is named in such a way that it does run as a normal test or an upgrade test for
+// other versions.
+func TestUpgradeTo1dot7(t *testing.T) {
```
MrHohn: Probably better to have a more meaningful name? Maybe `TestUpgradeToEnableFinalizer` :/

cmd/e2e-test/finalizer_test.go
```diff
@@ -39,122 +44,262 @@ func TestFinalizer(t *testing.T) {
...
+		for {
+			// While k8s master is upgrading, it will return a connection refused
+			// error for any k8s resource we try to hit. We loop until the
+			// master upgrade has finished.
+			if s.MasterUpgrading() {
+				continue
+			}
+
+			if s.MasterUpgraded() {
+				t.Logf("Detected master upgrade, adding a path to Ingress to force Ingress update")
```
MrHohn: Why do we need to force an update? When ingress controller starts up it would sync all existing ingresses once.

cmd/e2e-test/finalizer_test.go
skmatti: Cannot use `Finalizer` word in the test as it gets picked as a normal test.

cmd/e2e-test/finalizer_test.go
skmatti: I don't think we are force upgrading here. We run this loop forever until we see an upgrade and continue the test.

MrHohn: Right, the behavior I saw is that:
1. Wait forever until we see an upgrade
2. Adding a path to Ingress to force update
3. Wait for Ingress (and LB) to stabilized
4. Check if finalizer presents on Ingress

IMO step 2 seems weird, a real user may not touch their Ingress at all after the upgrade, but we are expecting finalizer to be attached in that case. What about simplify that to:
1. Wait forever until we see an upgrade
2. Wait for finalizer to be attached to Ingress

cmd/e2e-test/finalizer_test.go
MrHohn: Ah, I was hoping we have better filtering here.. But this makes sense now :)

cmd/e2e-test/finalizer_test.go
skmatti: Removed Ingress force update as suggested.

cmd/e2e-test/finalizer_test.go
```diff
@@ -39,122 +44,242 @@ func TestFinalizer(t *testing.T) {
...
+			}
+
+			if s.MasterUpgraded() {
+				t.Logf("Detected master upgrade")
+				break
+			}
+		}
+
+		// Verify that finalizer is added to the ingress after the upgrade.
+		ingFinalizers = ing.GetFinalizers()
```
MrHohn: We should poll on this --- the initial sync may not finish yet when master is upgraded?

cmd/e2e-test/finalizer_test.go
```diff
@@ -39,122 +44,242 @@ func TestFinalizer(t *testing.T) {
...
+		deleteOptions = &fuzz.GCLBDeleteOptions{
+			SkipDefaultBackend: true,
+		}
+		if err := e2e.WaitForIngressDeletion(ctx, otherGclb, s, otherIng, deleteOptions); err != nil {
+			t.Errorf("e2e.WaitForIngressDeletion(..., %q, nil) = %v, want nil", otherIng.Name, err)
 		}
+	})
+}
+
+// TestUpgradeTo1dot7 asserts that finalizer is added to an ingress when upgraded from a version
```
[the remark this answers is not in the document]
MrHohn: Or maybe `TestUpdateTo1dot7`.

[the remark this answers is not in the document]
skmatti: Done

cmd/e2e-test/finalizer_test.go
skmatti: Done

MrHohn: Sorry, I meant polling on the finalizer to be attached, instead of polling on the ingress to be stable :) Master upgrade usually won't have any impact on the data path so it will already be stable before the initial sync.

PR #846, opened 2019-09-10
e2e test framework add support for ilb subnet

spencerhance:
Adds support for creating and deleting ilb subnets to the e2e test framwork
Also adds optional "region" CLI flag to e2e binary which is needed for the subnet
Region will also be necessary for part 2 of the e2e test framework changes

### DIFF
--- a/pkg/e2e/fixtures.go
+++ b/pkg/e2e/fixtures.go
@@ -21,9 +21,13 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"k8s.io/ingress-gce/pkg/utils"
+	"net/http"
 	"reflect"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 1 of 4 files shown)


spencerhance:
@bowei Following up on this.  We really should be using this in the e2e tests.

pkg/e2e/fixtures.go
```diff
@@ -254,3 +260,34 @@ func DeleteGCPAddress(s *Sandbox, name string) error {
...
+			return ErrSubnetExists
+		} else {
+			return fmt.Errorf("Error creating ILB subnet: %v", err)
+		}
+	}
+	return nil
+}
+
+// DeleteIlbSubnet deletes the ILB subnet
+func DeleteIlbSubnet(s *Sandbox, name string) error {
```
bowei: `DeleteILBSubnet`

bowei:
Small nit, otherwise lgtm

pkg/e2e/fixtures.go
spencerhance: Done, I also changed the CreateIlbSubnet to CreateILBSubnet

PR #851, opened 2019-09-11
De-dupe additional source ranges for firewall

spencerhance:
- Fixes a bug where ILB subnet range is added multiple times
- Fixes a bug where empty ilb subnet is added to firewall

/assign @rramkumar1

### DIFF
--- a/pkg/firewalls/controller.go
+++ b/pkg/firewalls/controller.go
@@ -17,9 +17,8 @@ limitations under the License.
 package firewalls
 
 import (
+	"errors"
 	"fmt"
-	"k8s.io/ingress-gce/pkg/flags"
-	"k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/firewalls/firewalls.go
```diff
@@ -77,13 +77,18 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	}
 	sort.Strings(targetTags)
 
+	// De-dupe ports
```
bowei: It's better to have the permanent ones first and the additional adds after. 

```
ports := sets.NewString(fr.portRanges...)
ports.Insert(additionalPorts...)

ranges := sets.NewString(fr.srcRanges...)
ranges.Insert(additionalRanges)
```

pkg/firewalls/controller.go
```diff
@@ -172,7 +172,9 @@ func (fwc *FirewallController) sync(key string) error {
 		if err != nil {
 			return err
 		}
-		additionalRanges = append(additionalRanges, ilbRange)
+		if ilbRange != "" {
```
bowei: return a precise err

`errNoILBSubnetFound`

pkg/firewalls/firewalls.go
spencerhance: Done

pkg/firewalls/controller.go
spencerhance: Done

spencerhance:
@bowei Ready for another review

Issue #858, opened 2019-09-19
V2 Frontend Naming Scheme for Ingress-GCE L7 


skmatti:
This tracks the PRs related to V2 frontend naming scheme. This migration is primarily done in the following major PRs.

- [x] Refactor namer into a separate package: https://github.com/kubernetes/ingress-gce/pull/859
- [x] Cleanup backend namer workflow: https://github.com/kubernetes/ingress-gce/pull/861
- [x] Migrate existing namer to Legacy Frontend Namer: https://github.com/kubernetes/ingress-gce/pull/860
- [x] Add V2 Frontend Namer: https://github.com/kubernetes/ingress-gce/pull/892
  - New naming scheme to resolve naming collision issue.
  - Use kube-system namespace UID as the Cluster UID instead of Config Map.

PR #860, opened 2019-09-20
Migrating existing front-end namer logic to Legacy

skmatti:
This is part 3 of V2 Namer: https://github.com/kubernetes/ingress-gce/issues/858.

There is no change in naming policy with this change. But this provides an abstracted front-end namer which internally calls existing naming scheme. This will allow us to extend the frontend namer to migrate to V2 namer seamlessly.

This does not change GC workflow for frontend resources.


pkg/utils/utils.go
```diff
@@ -384,3 +385,12 @@ func ServiceKeyFunc(namespace, name string) string {
 func NeedsCleanup(ing *v1beta1.Ingress) bool {
 	return IsDeletionCandidate(ing.ObjectMeta, FinalizerKey) || !IsGLBCIngress(ing)
 }
+
+// ToLbNames returns a list of load balancers given a list of ingresses.
+func ToLbNames(ings []*v1beta1.Ingress) []string {
+	var lbNames []string
```
freehan: I think this is incorrect. You need to copy the logic from the original `toLbName` function

pkg/loadbalancers/l7.go
```diff
@@ -118,14 +116,19 @@ type L7 struct {
...
+	// feNamer is used to compute names of the various sub-components of an L7.
+	feNamer namer.IngressFrontendNamer
 	// recorder is used to generate k8s Events.
 	recorder record.EventRecorder
 	// resource type stores the KeyType of the resources in the loadbalancer (e.g. Regional)
 	scope meta.KeyType
 }
 
+// GetName() returns the name of the loadbalancer.
+func (l *L7) GetName() string {
```
freehan: I would add WARNING comment that this should only used in logging. 
It should not be used by actual naming.

pkg/loadbalancers/l7s.go
```diff
@@ -20,68 +20,90 @@ import (
...
+	namer_util "k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
 // L7s implements LoadBalancerPool.
 type L7s struct {
 	cloud            *gce.Cloud
-	namer            *namer.Namer
+	namer            *namer_util.Namer
```
freehan: I would highly recommend avoid passing the namer directly here. Use a interface to mask the actual namer. So that you have a clear set of functions used by `L7s`.

pkg/utils/namer/interfaces.go
```diff
@@ -0,0 +1,35 @@
...
+// IngressFrontendNamer is an interface to name GCE frontend resources.
+type IngressFrontendNamer interface {
+	ForwardingRule(protocol NamerProtocol) string
+	TargetProxy(protocol NamerProtocol) string
+	UrlMap() string
+	SSLCertName(secretHash string) string
+	IsCertUsedForLB(resourceName string) bool
+	IsLegacySSLCert(resourceName string) bool
+	GetLbName() string
+	GetBENamer() *Namer
```
freehan: Remove `GetBENamer()` interface. This interface breaks the abstraction.

Add the backendName per ServicePort on ServicePort creation.

pkg/utils/namer/interfaces.go
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+package namer
+
+import (
+	"k8s.io/api/networking/v1beta1"
+)
+
+// IngressFrontendNamer is an interface to name GCE frontend resources.
+type IngressFrontendNamer interface {
```
freehan: Add comments for the functions

pkg/utils/namer/fenamer_test.go
```diff
@@ -0,0 +1,106 @@
...
+func newIngress(namespace, name string) *v1beta1.Ingress {
+	return &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
+
+func TestLegacyIngressFrontendNamer(t *testing.T) {
```
freehan: Add a separate Unit test or change the existing unit tests to do this:

Validate LegacyIngressFrontendNamer and raw namer behave exactly the same:
1. Short names
2. Long names

pkg/loadbalancers/interfaces.go
```diff
@@ -26,8 +27,14 @@ import (
...
-	Shutdown() error
+	Delete(ing *v1beta1.Ingress, versions *features.ResourceVersions, scope meta.KeyType) error
+	// GC deletes gce loadbalancers given a list of ingresses to delete.
+	// This is used only with finalizers enabled.
+	GC(ingsToCleanup []*v1beta1.Ingress) error
+	// LegacyGC deletes gce loadbalancers not in the given list.
+	LegacyGC(lbNames []string) error
+	// Shutdown shutdowns loadbalancer pool by deleting the given list of ingresses.
+	Shutdown(ings []*v1beta1.Ingress) error
+	LegacyShutdown() error
```
freehan: Add comment?

freehan: How about not exposing Shutdown as a separate interface? Since all it does is GC.

pkg/utils/utils.go
```diff
@@ -384,3 +385,12 @@ func ServiceKeyFunc(namespace, name string) string {
 func NeedsCleanup(ing *v1beta1.Ingress) bool {
 	return IsDeletionCandidate(ing.ObjectMeta, FinalizerKey) || !IsGLBCIngress(ing)
 }
+
+// ToLbNames returns a list of load balancers given a list of ingresses.
+func ToLbNames(ings []*v1beta1.Ingress) []string {
```
freehan: consider having a special type for LbName instead of string. String is pretty error prune.

pkg/utils/utils.go
skmatti: I think the naming is confusing, this is meant to just return ingress keys which will be converted to GCE load balancer names.

pkg/loadbalancers/l7.go
skmatti: Done

pkg/loadbalancers/l7s.go
skmatti: Need to remove with `GetBENamer`. Add a TODO to clean this up.

pkg/utils/namer/interfaces.go
skmatti: This will be removed in subsequent PR. Added a TODO to clean this up.

pkg/utils/namer/interfaces.go
skmatti: Done

pkg/utils/namer/fenamer_test.go
skmatti: Done

pkg/loadbalancers/interfaces.go
skmatti: Removed shutdown.

pkg/utils/utils.go
skmatti: This is a helper function for older naming scheme and making changes would require changes to some of those methods belonging to that naming scheme. Besides, I refactored the name of this function which is meant to return string values.

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/go-openapi/spec v0.19.0
 	github.com/go-openapi/swag v0.19.0 // indirect
 	github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef // indirect
+	github.com/google/go-cmp v0.3.0
```
bowei: put module changes in a separate commit

pkg/utils/namer/utils.go
```diff
@@ -49,3 +55,13 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
...
 	return ret
 }
+
+// frontendNamingScheme returns naming scheme for given ingress.
+func frontendNamingScheme(ing *v1beta1.Ingress) Scheme {
+	// TODO(smatti): return V2NamingScheme if ingress has V2 finalizer
+	if common.HasFinalizer(ing.ObjectMeta, common.FinalizerKey) {
+		return LegacyNamingScheme
+	}
+	klog.V(2).Infof("No naming scheme found for Ingress %v, using Legacy naming scheme", ing)
```
bowei: V(3)

pkg/utils/namer/utils.go
```diff
@@ -49,3 +55,13 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
 
 	return ret
 }
+
+// frontendNamingScheme returns naming scheme for given ingress.
+func frontendNamingScheme(ing *v1beta1.Ingress) Scheme {
+	// TODO(smatti): return V2NamingScheme if ingress has V2 finalizer
```
bowei: you can fix this in the final commit

use switch:
```
switch {
  case common.HasFinalizer(...):
    return ...
  default:
    return ..
}
```

pkg/utils/namer/interfaces.go
```diff
@@ -29,3 +63,17 @@ type BackendNamer interface {
 	// this cluster's UID.
 	NameBelongsToCluster(resourceName string) bool
 }
+
+// LegacyFrontendNamer wraps frontend naming policy helper functions of namer.Namer.
```
bowei: Helper is redundant, use LegacyFrontendNamer
Comment does not match
Prefer to not have "legacy", "next", use v1,  v2, ... if there is every a v3, then legacy doesn't mean the right thing.

pkg/utils/namer/interfaces.go
```diff
@@ -13,6 +13,40 @@ limitations under the License.
...
+	IsLegacySSLCert(certName string) bool
+	// GetLbName returns load-balancer name for the ingress.
+	GetLbName() string
+}
+
+// IngressFrontendNamerFactory is an interface to create a front namer for an Ingress
+// a load balancer.
+type IngressFrontendNamerFactory interface {
+	// CreateIngressFrontendNamer returns IngressFrontendNamer for given ingress.
+	CreateIngressFrontendNamer(ing *v1beta1.Ingress) IngressFrontendNamer
```
bowei: Namer()
NamerForLBName()

pkg/utils/namer/interfaces.go
```diff
@@ -13,6 +13,40 @@ limitations under the License.
...
+	// SSLCertName returns the SSL certificate name given secret hash.
+	SSLCertName(secretHash string) string
+	// IsCertUsedForLB returns true if certName belongs to this ingress.
+	IsCertUsedForLB(certName string) bool
+	// IsLegacySSLCert returns true if certName follows the older naming convention
+	// and cert is managed by this ingress.
+	// old naming convention is of the form k8s-ssl-<lbName> or k8s-ssl-1-<lbName>.
+	IsLegacySSLCert(certName string) bool
+	// GetLbName returns load-balancer name for the ingress.
+	GetLbName() string
```
bowei: Get is redundant

pkg/utils/namer/interfaces.go
```diff
@@ -13,6 +13,40 @@ limitations under the License.
...
+	// UrlMap returns the name of the URL Map.
+	UrlMap() string
+	// SSLCertName returns the SSL certificate name given secret hash.
+	SSLCertName(secretHash string) string
+	// IsCertUsedForLB returns true if certName belongs to this ingress.
+	IsCertUsedForLB(certName string) bool
+	// IsLegacySSLCert returns true if certName follows the older naming convention
+	// and cert is managed by this ingress.
+	// old naming convention is of the form k8s-ssl-<lbName> or k8s-ssl-1-<lbName>.
+	IsLegacySSLCert(certName string) bool
```
bowei: IsV1CertName

pkg/utils/namer/interfaces.go
```diff
@@ -13,6 +13,40 @@ limitations under the License.
...
+	// ForwardingRule returns the name of the gce forwarding rule for given protocol.
+	ForwardingRule(protocol NamerProtocol) string
+	// TargetProxy returns the name of the gce target proxy for given protocol.
+	TargetProxy(protocol NamerProtocol) string
+	// UrlMap returns the name of the URL Map.
+	UrlMap() string
+	// SSLCertName returns the SSL certificate name given secret hash.
+	SSLCertName(secretHash string) string
+	// IsCertUsedForLB returns true if certName belongs to this ingress.
+	IsCertUsedForLB(certName string) bool
```
bowei: IsCertNameForLB

pkg/utils/namer/fenamer_test.go
```diff
@@ -0,0 +1,223 @@
...
+			"%s-fw-01234567890123456789012345678901234567890123456789-01230",
+			"%s-fws-01234567890123456789012345678901234567890123456789-0120",
+			"%s-um-01234567890123456789012345678901234567890123456789-01230",
+		},
+	}
+	for _, prefix := range []string{"k8s", "mci"} {
+		oldNamer := NewNamerWithPrefix(prefix, clusterUID, "")
+		secretHash := fmt.Sprintf("%x", sha256.Sum256([]byte("test123")))[:16]
+		for _, tc := range testCases {
+			tc.desc = fmt.Sprintf("%s namespaceLength %d nameLength %d prefix %s", tc.desc, len(tc.namespace), len(tc.name), prefix)
```
bowei: this seems redundant for desc as all the test cases have unique strings. use t.Log to print out specific details

pkg/utils/namer/fenamer_test.go
```diff
@@ -0,0 +1,223 @@
...
+			t.Run(tc.desc, func(t *testing.T) {
+				key := fmt.Sprintf("%s/%s", tc.namespace, tc.name)
+				feNamerFromLbName := NewLegacyIngressFrontendNamerFromLBName(oldNamer.LoadBalancer(key), oldNamer)
```
bowei: feNamerFromLbName => namer

pkg/utils/namer/fenamer_test.go
```diff
@@ -0,0 +1,223 @@
...
+			"%s-tp-namespace-name--uid1",
+			"%s-tps-namespace-name--uid1",
+			"%s-ssl-9a60a5272f6eee97-%s--uid1",
+			"%s-fw-namespace-name--uid1",
+			"%s-fws-namespace-name--uid1",
+			"%s-um-namespace-name--uid1",
+		},
+		{
+			"62 characters",
+			longString[:23],
```
bowei: put a comment here as to how these are computed.

pkg/utils/namer/fenamer_test.go
```diff
@@ -0,0 +1,223 @@
...
+		// Expected values.
+		lbName              string
+		targetHTTPProxy     string
+		targetHTTPSProxy    string
+		sslCert             string
+		forwardingRuleHTTP  string
+		forwardingRuleHTTPS string
+		urlMap              string
+	}{
+		{
```
bowei: we need a test case with

ns = "a", name = longstring
ns = longstring, name = "a"

pkg/utils/namer/fenamer.go
```diff
@@ -0,0 +1,108 @@
...
+
+// NewIngressFrontendNamerFactoryImpl returns IngressFrontendNamerFactory given a legacy namer.
+func NewIngressFrontendNamerFactoryImpl(namer *Namer) IngressFrontendNamerFactory {
+	return &IngressFrontendNamerFactoryImpl{namer: namer}
+}
+
+// CreateIngressFrontendNamer implements IngressFrontendNamerFactory.
+func (rn *IngressFrontendNamerFactoryImpl) CreateIngressFrontendNamer(ing *v1beta1.Ingress) IngressFrontendNamer {
+	switch frontendNamingScheme(ing) {
+	case LegacyNamingScheme:
```
bowei: LegacyNamingScheme => V1

pkg/utils/namer/fenamer.go
```diff
@@ -0,0 +1,108 @@
...
+
+package namer
+
+import (
+	"k8s.io/api/networking/v1beta1"
+	"k8s.io/ingress-gce/pkg/utils/common"
+)
+
+const (
+	LegacyNamingScheme = Scheme("v1")
```
bowei: usually people write
```
V1NamingScheme Scheme = "v1"
V2NamingScheme Scheme = "v2"
```

pkg/loadbalancers/url_maps.go
```diff
@@ -218,12 +218,12 @@ func mapsEqual(a, b *composite.UrlMap) bool {
 // and remove the mapping. When a new path is added to a host (happens
 // more frequently than service deletion) we just need to lookup the 1
 // pathmatcher of the host.
-func toCompositeURLMap(lbName string, g *utils.GCEURLMap, namer *namer.Namer, key *meta.Key) *composite.UrlMap {
+func toCompositeURLMap(g *utils.GCEURLMap, feNamer namer.IngressFrontendNamer, key *meta.Key) *composite.UrlMap {
```
bowei: feNamer => namer

pkg/loadbalancers/url_maps.go
```diff
@@ -68,12 +68,12 @@ func (l *L7) ensureComputeURLMap() error {
...
 
 	if mapsEqual(currentMap, expectedMap) {
-		klog.V(4).Infof("URLMap for %q is unchanged", l.Name)
+		klog.V(4).Infof("URLMap for %q is unchanged", l.ToString())
 		l.um = currentMap
 		return nil
 	}
 
-	klog.V(3).Infof("Updating URLMap for %q", l.Name)
+	klog.V(3).Infof("Updating URLMap for %q", l.ToString())
```
bowei: You should call the toString() method String(). toString is a Java-ism
I also think %q automatically calls String() if it exists.

pkg/loadbalancers/l7s_test.go
```diff
@@ -297,46 +307,47 @@ func newTestLoadBalancerPool() LoadBalancerPool {
 	namer := namer_util.NewNamer(testClusterName, "fw1")
 	fakeGCECloud := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
 	ctx := &context.ControllerContext{}
-	return NewLoadBalancerPool(fakeGCECloud, namer, ctx)
+	return NewLoadBalancerPool(fakeGCECloud, namer, ctx, namer_util.NewIngressFrontendNamerFactoryImpl(namer))
 }
 
-func createFakeLoadbalancer(cloud *gce.Cloud, namer *namer_util.Namer, lbKey string, versions *features.ResourceVersions, scope meta.KeyType) {
-	lbName := namer.LoadBalancer(lbKey)
+func getLegacyFrontendNamer(namerHelper namer_util.LegacyFrontendNamerHelper, feNamerFactory namer_util.IngressFrontendNamerFactory, lbKey string) namer_util.IngressFrontendNamer {
```
bowei: makeV1FrontendNamer

pkg/loadbalancers/l7s.go
```diff
@@ -25,63 +25,60 @@ import (
...
 
 // NewLoadBalancerPool returns a new loadbalancer pool.
 // - cloud: implements LoadBalancers. Used to sync L7 loadbalancer resources
 //	 with the cloud.
-func NewLoadBalancerPool(cloud *gce.Cloud, namer *namer.Namer, recorderProducer events.RecorderProducer) LoadBalancerPool {
+func NewLoadBalancerPool(cloud *gce.Cloud, namerHelper namer_util.LegacyFrontendNamerHelper, recorderProducer events.RecorderProducer, feNamerFactory namer_util.IngressFrontendNamerFactory) LoadBalancerPool {
 	return &L7s{
 		cloud:            cloud,
-		namer:            namer,
+		namerHelper:      namerHelper,
```
bowei: namerHelper is kind of a non-descript name. Can we have a more descriptive name?

pkg/loadbalancers/l7.go
```diff
@@ -118,14 +110,21 @@ type L7 struct {
...
 	// recorder is used to generate k8s Events.
 	recorder record.EventRecorder
 	// resource type stores the KeyType of the resources in the loadbalancer (e.g. Regional)
 	scope meta.KeyType
 }
 
+// ToString() returns the name of the loadbalancer.
+// Warning: This should be used only for logging and should not be used to
+// retrieve/ delete gce resource names.
+func (l *L7) ToString() string {
```
bowei: String()

pkg/controller/controller_test.go
```diff
@@ -86,7 +86,7 @@ func newLoadBalancerController() *LoadBalancerController {
 	lbc := NewLoadBalancerController(ctx, stopCh)
 	// TODO(rramkumar): Fix this so we don't have to override with our fake
 	lbc.instancePool = instances.NewNodePool(instances.NewFakeInstanceGroups(sets.NewString(), namer), namer)
-	lbc.l7Pool = loadbalancers.NewLoadBalancerPool(fakeGCE, namer, events.RecorderProducerMock{})
+	lbc.l7Pool = loadbalancers.NewLoadBalancerPool(fakeGCE, namer, events.RecorderProducerMock{}, namer_util.NewIngressFrontendNamerFactoryImpl(namer))
```
bowei: NewIngressFrontendNamerFactoryImpl => NewFrontendNamerFactory

pkg/controller/controller.go
```diff
@@ -553,15 +554,10 @@ func (lbc *LoadBalancerController) updateIngressStatus(l7 *loadbalancers.L7, ing
 
 // toRuntimeInfo returns L7RuntimeInfo for the given ingress.
 func (lbc *LoadBalancerController) toRuntimeInfo(ing *v1beta1.Ingress, urlMap *utils.GCEURLMap) (*loadbalancers.L7RuntimeInfo, error) {
-	k, err := common.KeyFunc(ing)
-	if err != nil {
-		return nil, fmt.Errorf("cannot get key for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
-	}
-
 	var tls []*loadbalancers.TLSCerts
```
bowei: delete this

pkg/utils/namer/fenamer_test.go
skmatti: Done

pkg/utils/namer/fenamer_test.go
skmatti: Done

pkg/loadbalancers/l7s_test.go
skmatti: Removed this function.

pkg/loadbalancers/l7s.go
skmatti: Done

pkg/utils/namer/interfaces.go
skmatti: There was an earlier migration of certs. This function refers to that old naming scheme that was used just for ssl certificates.

pkg/loadbalancers/l7.go
```diff
@@ -119,13 +111,20 @@ type L7 struct {
...
 	// namer is used to compute names of the various sub-components of an L7.
-	namer *namer.Namer
+	namer namer.IngressFrontendNamer
 	// recorder is used to generate k8s Events.
 	recorder record.EventRecorder
 	// resource type stores the KeyType of the resources in the loadbalancer (e.g. Regional)
 	scope meta.KeyType
 }
 
+// String() returns the name of the loadbalancer.
```
bowei: // String returns

pkg/loadbalancers/l7s.go
```diff
@@ -25,63 +25,60 @@ import (
...
 	"k8s.io/klog"
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
 // L7s implements LoadBalancerPool.
 type L7s struct {
-	cloud            *gce.Cloud
-	namer            *namer.Namer
+	cloud *gce.Cloud
+	// v1FeNamerHelper is an interface for helper functions for v1 frontend naming scheme.
```
bowei: v1NamerHelper, is there also a backend namer in this struct, otherwise there is no need to disambiguate.

pkg/utils/namer/fenamer.go
```diff
@@ -0,0 +1,108 @@
+/*
```
bowei: frontendnamer.go

PR #870, opened 2019-09-25
Add CreateILBSubnet() call to all ilb e2e tests

spencerhance:
This will ensure that all the tests have an ILB subnet in their region.  Later, we can add tests which delete or migrate the subnet and test that behavior.

### DIFF
--- a/cmd/e2e-test/ilb_test.go
+++ b/cmd/e2e-test/ilb_test.go
@@ -27,11 +27,14 @@ import (
 	"testing"
 )
 
-var negVal = annotations.NegAnnotation{
-	Ingress: true,
-}
-var negAnnotation = map[string]string{
-	annotations.NEGAnnotationKey: negVal.String()}
+var (
+	negVal        = annotations.NegAnnotation{Ingress: true}
+	negAnnotation = map[string]string{
+		annotations.NEGAnnotationKey: negVal.String()}
+
+	// subnetCidr is the default CIDR for ILB subnets for these tests
+	subnetCidr = "10.126.0.0/22"
+)
 
 func TestILB(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/ilb_test.go
```diff
@@ -27,11 +27,14 @@ import (
...
 
-var negVal = annotations.NegAnnotation{
-	Ingress: true,
-}
-var negAnnotation = map[string]string{
-	annotations.NEGAnnotationKey: negVal.String()}
+var (
+	negVal        = annotations.NegAnnotation{Ingress: true}
+	negAnnotation = map[string]string{
+		annotations.NEGAnnotationKey: negVal.String()}
```
bowei: why not
```
anootations.NEGAnnotationKey: annotations.NegAnnotation{Ingress:true}.String()
```

cmd/e2e-test/ilb_test.go
```diff
@@ -80,6 +83,11 @@ func TestILB(t *testing.T) {
 
 			t.Logf("Ingress = %s", tc.ing.String())
 
+			// Create Subnet if it doesn't already exist
+			if err := e2e.CreateILBSubnet(s, "ilb-subnet-ingress-e2e", subnetCidr); err != nil && err != e2e.ErrSubnetExists {
```
bowei: might want to put a random suffix on the subnet name?

cmd/e2e-test/ilb_test.go
spencerhance:
```golang
annotations.NegAnnotation{Ingress:true}.String()
```
throws a compiler error since String() wants a pointer

spencerhance: ```cannot call pointer method on annotations.NegAnnotation literal```

cmd/e2e-test/ilb_test.go
spencerhance: I can but it won't change the behavior.  We can only ever create one subnet throughout all the tests running in parallel.

PR #871, opened 2019-09-27
Update ILB e2e test resource names

spencerhance:
Change the names of the resources in the ilb e2e tests to make debugging easier

### DIFF
--- a/cmd/e2e-test/ilb_test.go
+++ b/cmd/e2e-test/ilb_test.go
@@ -39,6 +39,11 @@ var (
 func TestILB(t *testing.T) {
 	t.Parallel()
 
+	// These names are useful when reading the debug logs
+	testName := "test-ilb-basic"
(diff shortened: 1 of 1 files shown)


spencerhance:
@MrHohn fixed the test failure

cmd/e2e-test/ilb_test.go
```diff
@@ -39,6 +39,11 @@ var (
 func TestILB(t *testing.T) {
 	t.Parallel()
 
+	// These names are useful when reading the debug logs
+	testName := "test-ilb-basic"
```
bowei: probably want to match testName with "TestILB"

spencerhance: Only reason I didn't do that here was that every other testName starts with "test-ilb," so this test wouldn't be easily searchable.

PR #873, opened 2019-09-27
Update CreateILBSubnet() to catch 400s as well

spencerhance:
We should also expect to see 400s when attempting to create the subnet after the first time.


spencerhance:
@MrHohn fixed the error, can you re-lgtm?

PR #875, opened 2019-09-27
Introduce a flag for creating non-GCP NEGs.

cxhiano:
With '--enable-non-gcp-mode', NEG controller will create NEGs of type NON_GCP_PRIVATE_IP_PORT instead of GCE_VM_IP_PORT.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -87,6 +87,7 @@ var (
 		EnableL7Ilb                 bool
 		EnableCSM                   bool
 		CSMServiceNEGSkipNamespaces []string
+		CreateHybridNeg             bool
 
 		LeaderElection LeaderElectionConfiguration
 	}{}
@@ -204,6 +205,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableCSM, "enable-csm", false, "Enable CSM(Istio) support")
 	flag.StringSliceVar(&F.CSMServiceNEGSkipNamespaces, "csm-service-skip-namespaces", []string{}, "Only for CSM mode, skip the NEG creation for Services in the given namespaces.")
+	flag.BoolVar(&F.CreateHybridNeg, "create-hybrid-neg", false, "Create Negs of type NON_GCP_PRIVATE_IP_PORT instead of GCE_VM_IP_PORT.")
 }
 
 type RateLimitSpecs struct {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -87,7 +87,7 @@ var (
 		EnableL7Ilb                 bool
 		EnableCSM                   bool
 		CSMServiceNEGSkipNamespaces []string
-		EnableHybrid                bool
+		EnableNonGCPMode            bool
 
 		LeaderElection LeaderElectionConfiguration
 	}{}
@@ -205,7 +205,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableCSM, "enable-csm", false, "Enable CSM(Istio) support")
 	flag.StringSliceVar(&F.CSMServiceNEGSkipNamespaces, "csm-service-skip-namespaces", []string{}, "Only for CSM mode, skip the NEG creation for Services in the given namespaces.")
-	flag.BoolVar(&F.EnableHybrid, "enable-hybrid", false, "Enable hybrid support.")
+	flag.BoolVar(&F.EnableNonGCPMode, "enable-non-gcp-mode", false, "Set to true when running on a non-GCP cluster.")
 }
 
 type RateLimitSpecs struct {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -87,7 +87,7 @@ var (
 		EnableL7Ilb                 bool
 		EnableCSM                   bool
 		CSMServiceNEGSkipNamespaces []string
-		CreateHybridNeg             bool
+		EnableHybrid                bool
 
 		LeaderElection LeaderElectionConfiguration
 	}{}
@@ -205,7 +205,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableCSM, "enable-csm", false, "Enable CSM(Istio) support")
 	flag.StringSliceVar(&F.CSMServiceNEGSkipNamespaces, "csm-service-skip-namespaces", []string{}, "Only for CSM mode, skip the NEG creation for Services in the given namespaces.")
-	flag.BoolVar(&F.CreateHybridNeg, "create-hybrid-neg", false, "Create Negs of type NON_GCP_PRIVATE_IP_PORT instead of GCE_VM_IP_PORT.")
+	flag.BoolVar(&F.EnableHybrid, "enable-hybrid", false, "Enable hybrid support.")
 }
 
 type RateLimitSpecs struct {
--- a/pkg/neg/manager_test.go
+++ b/pkg/neg/manager_test.go
@@ -315,19 +315,22 @@ func TestGarbageCollectionNEG(t *testing.T) {
 		t.Fatalf("Failed to ensure syncer: %v", err)
 	}
 
-	negName := manager.namer.NEG("test", "test", 80)
-	manager.cloud.CreateNetworkEndpointGroup(&compute.NetworkEndpointGroup{
-		Name: negName,
-	}, negtypes.TestZone1)
+	for _, networkEndpointType := range []string{"GCE_VM_IP_PORT", "NON_GCP_PRIVATE_IP_PORT"} {
+		negName := manager.namer.NEG("test", "test", 80)
+		manager.cloud.CreateNetworkEndpointGroup(&compute.NetworkEndpointGroup{
+			Name:                negName,
+			NetworkEndpointType: networkEndpointType,
+		}, negtypes.TestZone1)
 
-	if err := manager.GC(); err != nil {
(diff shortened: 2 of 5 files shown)


cxhiano:
I have signed

pkg/flags/flags.go
```diff
@@ -204,6 +205,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableCSM, "enable-csm", false, "Enable CSM(Istio) support")
 	flag.StringSliceVar(&F.CSMServiceNEGSkipNamespaces, "csm-service-skip-namespaces", []string{}, "Only for CSM mode, skip the NEG creation for Services in the given namespaces.")
+	flag.BoolVar(&F.CreateHybridNeg, "create-hybrid-neg", false, "Create Negs of type NON_GCP_PRIVATE_IP_PORT instead of GCE_VM_IP_PORT.")
```
freehan: create-hybrid-neg  ->  enable-hybrid

This is the first hybrid use case. We may want to keep the flag generic to prevent growing number of flags.

pkg/neg/syncers/utils.go
```diff
@@ -141,11 +143,17 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 
 	if needToCreate {
 		klog.V(2).Infof("Creating NEG %q for %s in %q.", negName, negServicePortName, zone)
+		networkEndpointType := negIPPortNetworkEndpointType
+		subnetwork := cloud.SubnetworkURL()
+		if flags.F.CreateHybridNeg {
```
freehan: add a comment regarding why the override is needed?
something like:

NEG type is different.
Subnetwork is relevant

pkg/neg/syncers/utils.go
```diff
@@ -282,10 +290,13 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet) (map[negtypes.Netw
 		}
 
 		endpointBatch[networkEndpoint] = &compute.NetworkEndpoint{
-			Instance:  networkEndpoint.Node,
 			IpAddress: networkEndpoint.IP,
 			Port:      int64(portNum),
 		}
+
+		if !flags.F.CreateHybridNeg {
```
freehan: I would do a override instead. Add comments please

```
if flags.F.CreateHybridNeg {
     endpointBatch[networkEndpoint].Instance = ""
}

File: pkg/flags/flags.go [hunk 4]
```diff
@@ -87,6 +87,7 @@ var (
 		EnableL7Ilb                 bool
 		EnableCSM                   bool
 		CSMServiceNEGSkipNamespaces []string
+		EnableHybrid                bool
```
bowei: Needs to be commented

pkg/flags/flags.go
```diff
@@ -204,6 +205,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableCSM, "enable-csm", false, "Enable CSM(Istio) support")
 	flag.StringSliceVar(&F.CSMServiceNEGSkipNamespaces, "csm-service-skip-namespaces", []string{}, "Only for CSM mode, skip the NEG creation for Services in the given namespaces.")
+	flag.BoolVar(&F.EnableHybrid, "enable-hybrid", false, "Enable hybrid support.")
```
bowei: Hybrid what?

pkg/neg/syncers/utils.go
```diff
@@ -39,11 +40,12 @@ const (
...
-	maxRetries                   = 15
-	minRetryDelay                = 5 * time.Second
-	maxRetryDelay                = 600 * time.Second
-	separator                    = "||"
-	negIPPortNetworkEndpointType = "GCE_VM_IP_PORT"
+	maxRetries                          = 15
+	minRetryDelay                       = 5 * time.Second
+	maxRetryDelay                       = 600 * time.Second
+	separator                           = "||"
+	negIPPortNetworkEndpointType        = "GCE_VM_IP_PORT"
```
bowei: put a newline between the duration consts and the string consts

pkg/neg/syncers/utils.go
```diff
@@ -39,11 +40,12 @@ const (
...
+	negPrivateIPPortNetworkEndpointType = "NON_GCP_PRIVATE_IP_PORT"
```
bowei: These names are kind of redundant  neg = "network endpoint group" and you have NewtorkEndpointType at the end of the word.
Let's change this to 
```
vmNetworkEndpointType
nonGCPPrivateEndpointType
```

pkg/neg/syncers/utils_test.go
```diff
@@ -292,6 +293,61 @@ func TestNetworkEndpointCalculateDifference(t *testing.T) {
 	}
 }
 
+func TestEnsureNetworkEndpointGroup(t *testing.T) {
+	testNegName := "test-neg"
```
bowei:
```
const (
  testNegName = "..."
  ...
)
```

pkg/neg/syncers/utils.go
```diff
@@ -141,11 +143,18 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 
 	if needToCreate {
 		klog.V(2).Infof("Creating NEG %q for %s in %q.", negName, negServicePortName, zone)
+		networkEndpointType := negIPPortNetworkEndpointType
```
bowei: can we do  (makes the flow more obvious)
```
var neType, subnetwork string
switch {
  case flags.F.EnableHybrd:
   neType = ...
  default:
    neType = negIP...
    subnetwork = cloud.SubnetworkURL()
}
```

pkg/flags/flags.go
cxhiano: Updated the flag description below to make it more clear. Let me know if you think comment is still needed here. (I will just copy-paste the description as comment here)

pkg/neg/syncers/utils_test.go
```diff
@@ -292,6 +293,63 @@ func TestNetworkEndpointCalculateDifference(t *testing.T) {
...
+		testServiceName      = "test-svc"
+		testServiceNameSpace = "test-ns"
+		testSubnetwork       = "test-subnetwork"
+		testNetwork          = "test-network"
+	)
+
+	fakeCloud := negtypes.NewFakeNetworkEndpointGroupCloud(testSubnetwork, testNetwork)
+
+	testCases := []struct {
+		enableNonGCPMode            bool
```
freehan: add a description field with string type in the test case.

pkg/neg/syncers/utils_test.go
```diff
@@ -292,6 +293,63 @@ func TestNetworkEndpointCalculateDifference(t *testing.T) {
...
+		{
+			enableNonGCPMode:            false,
+			expectedNetworkEndpointType: vmNetworkEndpointType,
+			expectedSubnetwork:          testSubnetwork,
+		},
+		{
+			enableNonGCPMode:            true,
+			expectedNetworkEndpointType: nonGCPPrivateEndpointType,
+			expectedSubnetwork:          "",
+		},
```
freehan: add one more test case with duplicate NEG and make sure that the `ensureNetworkEndpointGroup` function does not return error.

pkg/neg/syncers/utils_test.go
```diff
@@ -700,7 +761,7 @@ func TestShouldPodBeInNeg(t *testing.T) {
 
 }
 
-func genTestEndpoints(num int) (negtypes.NetworkEndpointSet, map[negtypes.NetworkEndpoint]*compute.NetworkEndpoint) {
+func genTestEndpoints(num int, enableNonGCPMode bool) (negtypes.NetworkEndpointSet, map[negtypes.NetworkEndpoint]*compute.NetworkEndpoint) {
```
freehan: you do not need to change this, right?

since the `ensureNetworkEndpointGroup` function itself will wipe out the instace field.

pkg/neg/syncers/utils_test.go
```diff
@@ -292,6 +295,84 @@ func TestNetworkEndpointCalculateDifference(t *testing.T) {
...
+			testZone,
+			testNamedPort,
+			fakeCloud,
+			nil,
+			nil,
+		)
+
+		if err != nil {
+			t.Errorf("Unexpected error when called with duplicated NEG: %v", err)
+		}
```
cxhiano: Ideally I'd want to test the NEG is not deleted but could not find a good way to do that.

freehan: That is a good point. 

You probably need a unit test for GC Hybrid NEG.  
Take a look at unit test for GC. You probably need a different test for that.
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/manager_test.go#L303

freehan: I take that back. Testing GC is also needed. But would not prevent the bug you fixed.

There is also a way to intercept GCE API call in the cloud provider layer. But it is a bit hacky, but it can work. 
This is essentially how this works. 
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction_test.go#L58

pkg/neg/manager_test.go
```diff
@@ -315,19 +315,22 @@ func TestGarbageCollectionNEG(t *testing.T) {
...
 
-	negName := manager.namer.NEG("test", "test", 80)
-	manager.cloud.CreateNetworkEndpointGroup(&compute.NetworkEndpointGroup{
-		Name: negName,
-	}, negtypes.TestZone1)
-
-	if err := manager.GC(); err != nil {
-		t.Fatalf("Failed to GC: %v", err)
-	}
+	for _, networkEndpointType := range []string{"GCE_VM_IP_PORT", "NON_GCP_PRIVATE_IP_PORT"} {
```
freehan: add comment here to describe the intention: test GC in different modes. 

Use the static variable instead of  the raw string

cxhiano:
Closing in favor of https://github.com/kubernetes/ingress-gce/pull/920

Issue #879, opened 2019-10-01
Bug: Incorrect ingress key function is being used


skmatti:
Refer: https://github.com/kubernetes/ingress-gce/pull/825#discussion_r330277262. The ingress key generated by IngressKeyFunc does not match the key generated in Ingress store in some special cases (when namespace is empty). IngressKeyFunc should be replaced by [this](https://github.com/skmatti/ingress-gce/blob/e23564d807c2c6d85e0e5745cef4ee0b9d540b7b/pkg/utils/taskqueue.go#L26). This function is used for GC'ing loadbalancers, so this may cause some loadbalancers to be deleted incorrectly.

skmatti:
In general this should not cause a bug if ingress has a valid object meta defined [here](https://github.com/kubernetes/ingress-gce/blob/be3cc807e8fc9255a50b7c6cedec562b066094e2/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/meta.go#L33) and a non-empty namespace. So, this would not cause any problems with valid ingresses in production.

PR #881, opened 2019-10-01
BugFix: Update ingress key function used for GC

skmatti:
Fixes https://github.com/kubernetes/ingress-gce/issues/879

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -49,7 +49,6 @@ import (
 	ingsync "k8s.io/ingress-gce/pkg/sync"
 	"k8s.io/ingress-gce/pkg/tls"
 	"k8s.io/ingress-gce/pkg/utils"
-	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/utils/namer/utils.go
```diff
@@ -15,14 +15,19 @@ package namer
 
 import (
 	"k8s.io/api/networking/v1beta1"
-	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/ingress-gce/pkg/utils/common"
+	"k8s.io/klog"
 )
 
+// IngressKeyFunc returns ingress key used in Ingress Cache Store.
 func IngressKeyFunc(ing *v1beta1.Ingress) string {
```
MrHohn: Seems like no one is calling this function anymore?

skmatti: https://github.com/kubernetes/ingress-gce/blob/7b67130e0064f3af8957f01cdb2c11b76633c292/pkg/controller/controller.go#L634

This line is not modified in this change. So does not show this in diff.

MrHohn: Can we do this explicitly in that `toLbNames()` func? It seems confusing to have `IngressKeyFunc`, `KeyFunc` and `ToIngressKey`..

skmatti: Done. I will move this logic to separate function with v2 namer changes. Also refactored original IngressKeyFunc to utils as it will not be used by namer any more.

pkg/controller/controller.go
```diff
@@ -630,7 +629,13 @@ func toLbNames(ings []*v1beta1.Ingress) []string {
 		if !utils.IsGCEIngress(ing) {
 			continue
 		}
-		lbNames = append(lbNames, namer.IngressKeyFunc(ing))
+		ingKey, err := utils.KeyFunc(ing)
+		// An error is returned only if ingress object does not have a valid meta.
+		// So, logging the error would be sufficient here.
+		if err != nil {
+			klog.Errorf("Cannot get key for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
+		}
```
MrHohn: Can we do:
```golang
		if err != nil {
			klog.Errorf("Cannot get key for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
			continue
		}
```

PR #883, opened 2019-10-02
Refactor ingress key function and finalizer into separate package

skmatti:
Refactors some of ingress key utility functions and finalizer logic to a separate package so these can be accessed by namer package without import cycle. A subsequent PR https://github.com/kubernetes/ingress-gce/pull/860 with utilization of these refactored functions inside namer package.


pkg/utils/common/common.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+var (
+	KeyFunc = cache.DeletionHandlingMetaNamespaceKeyFunc
```
freehan: How about we organize it this way: 

KeyFunc is used only used to handle items in workqueue 


After the initial process of ingress, everything is handled by IngressKeyFunc.

So you do not need the ToIngressKey function as middleman?

skmatti: The idea is to generate same key as that of Ingress Store and preserve backward compatibility. In theory, both `KeyFunc` and `IngressKeyFunc` should return same value with a valid Ingress. Some of the edge cases where they output different values is when namespace is empty. `KeyFunc` returns `<name>` where as `IngressKeyFunc` returns `/<name>`.

I changed the names a bit so they are less confusing and also added some description about why we need two separate functions as explained here.

pkg/utils/common/common.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// ToString returns ingress key for given ingress.
```
freehan: ToString returns namespaced name string of a given ingress

skmatti: Done

PR #885, opened 2019-10-04
Fix CreateILBSubnet() logic in e2e tests

spencerhance:
Update subnet to /24 instead of /22
Add retry logic for 10 different possible subnets

/assign @MrHohn

### DIFF
--- a/pkg/e2e/fixtures.go
+++ b/pkg/e2e/fixtures.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"k8s.io/ingress-gce/pkg/utils"
+	"math/rand"
 	"net/http"
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


pkg/e2e/fixtures.go
```diff
@@ -262,8 +262,28 @@ func DeleteGCPAddress(s *Sandbox, name string) error {
...
-func CreateILBSubnet(s *Sandbox, name string, ipCidrRange string) error {
+func CreateILBSubnet(s *Sandbox) error {
+	klog.V(3).Info("CreateILBSubnet()")
+	name := "ilb-subnet-ingress-e2e"
+
+	// Try up to 10 different subnets since we can't conflict with anything in the test project
+	// TODO(shance): find a more reliable way to pick the subnet
+	var err error
+	for i := 0; i < 10; i++ {
+		ipCidrRange := fmt.Sprintf("10.126.%d.0/24", i)
```
bowei: I think if you use 192.168.x.x, almost no one uses that range in the tests.

bowei: Also, I would start at a random place in the range
```
start := // random value
...
fmt.Sprintf("192.168.%d.0/24", i+start%256)
```

spencerhance: Good call, updated

PR #886, opened 2019-10-04
remove "addonmanager.kubernetes.io/mode: Reconcile" annotation from

cadmuxe:
user space default backend service.

With this annotation, the backend service will be deleted if created with kubectl apply -f


bowei:
can we instead comment out and put a comment saying:
```
# If this YAML is used with the addon-manager, the following line must be
# set to signal that this configuration is managed by the addon-manager.
```

PR #892, opened 2019-10-15
Add V2 frontend namer

skmatti:
This part4 of V2 Namer Migration: https://github.com/kubernetes/ingress-gce/issues/858. 
Fixes: https://github.com/kubernetes/ingress-gce/issues/537

This PR introduces a new naming scheme for for the following frontend resources:
- Forwarding Rule
- Target Proxy
- URL Map
- SSL Certificates
- Ingress managed Static IP

This should resolve frontend resource naming collision issues. 

The changes made in v2 naming scheme:
- New naming scheme:
`k8s{version-id}-{resource-prefix}-{kube-system-uid}-{namespace}-{name}-{8-char-hash}`
Example http target proxy: `k8s2-tp-ksuid123-namespace-name-wddys49o`
- Resource names will include an 8 character kube-system uid instead of cluster uid. This makes the new naming policy immune to config map(which tracks cluster uid) changes.

Frontend  GC workflow changes:
- New GC workflow for v2 naming scheme
  - GC will be invoked only on the ingress being synced.
  - GC workflow will not leak any GCE resources.

**GC workflow**
 - If finalizer is disabled
   - GC workflow is untouched i.e, run GC for all ingresses
- If finalizer is enabled
  - If ingress being synced needs cleanup
    - Frontend GC
      - if V1 naming scheme - run frontend GC for all v1 ingresses.
      - if V2 naming scheme - run frontend GC for ingress being synced. 
    - Backend GC - run for all ingresses.
    - Ensure delete finalizers
      - if V1 naming scheme - remove finalizers for all v1 ingresses.
      - if V2 naming scheme - remove finalizer for ingress being synced.
   - If  ingress being synced does not need cleanup
     - Backend GC - run for all ingresses.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	flag "github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/ingress-gce/pkg/frontendconfig"
 	"k8s.io/klog"
 
@@ -123,6 +125,16 @@ func main() {
 		klog.V(0).Infof("Cluster name: %+v", namer.UID())
 	}
 
+	var kubeSystemUID types.UID
+	if flags.F.EnableV2FrontendNamer {
+		// Get kube-system UID that will be used for v2 frontend naming scheme.
+		kubeSystemNS, err := kubeClient.CoreV1().Namespaces().Get("kube-system", metav1.GetOptions{})
+		if err != nil {
+			klog.Fatalf("Error getting kube-system namespace: %v", err)
+		}
+		kubeSystemUID = kubeSystemNS.GetUID()
+	}
+
 	cloud := app.NewGCEClient()
 	defaultBackendServicePort := app.DefaultBackendServicePort(kubeClient)
 	ctxConfig := ingctx.ControllerContextConfig{
@@ -136,7 +148,7 @@ func main() {
 		ASMConfigMapNamespace:         flags.F.ASMConfigMapBasedConfigNamespace,
 		ASMConfigMapName:              flags.F.ASMConfigMapBasedConfigCMName,
 	}
-	ctx := ingctx.NewControllerContext(kubeConfig, kubeClient, backendConfigClient, frontendConfigClient, cloud, namer, ctxConfig)
+	ctx := ingctx.NewControllerContext(kubeConfig, kubeClient, backendConfigClient, frontendConfigClient, cloud, namer, kubeSystemUID, ctxConfig)
 	go app.RunHTTPServer(ctx.HealthCheck)
 
 	if !flags.F.LeaderElection.LeaderElect {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -115,7 +115,7 @@ func NewLoadBalancerController(
 		hasSynced:     ctx.HasSynced,
 		nodes:         NewNodeController(ctx, instancePool),
 		instancePool:  instancePool,
-		l7Pool:        loadbalancers.NewLoadBalancerPool(ctx.Cloud, ctx.ClusterNamer, ctx, namer.NewFrontendNamerFactory(ctx.ClusterNamer)),
+		l7Pool:        loadbalancers.NewLoadBalancerPool(ctx.Cloud, ctx.ClusterNamer, ctx, namer.NewFrontendNamerFactory(ctx.ClusterNamer, ctx.KubeSystemUID)),
 		backendSyncer: backends.NewBackendSyncer(backendPool, healthChecker, ctx.Cloud),
 		negLinker:     backends.NewNEGLinker(backendPool, negtypes.NewAdapter(ctx.Cloud), ctx.Cloud),
 		igLinker:      backends.NewInstanceGroupLinker(instancePool, backendPool),
@@ -330,9 +330,10 @@ func (lbc *LoadBalancerController) Stop(deleteAll bool) error {
 	// TODO(rramkumar): Do we need deleteAll? Can we get rid of its' flag?
 	if deleteAll {
 		klog.Infof("Shutting down cluster manager.")
-		if err := lbc.l7Pool.Shutdown(); err != nil {
+		if err := lbc.l7Pool.Shutdown(lbc.ctx.Ingresses().List()); err != nil {
 			return err
 		}
+
 		// The backend pool will also delete instance groups.
 		return lbc.backendSyncer.Shutdown()
 	}
@@ -453,25 +454,42 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
 	return nil
 }
 
-// GCLoadBalancers implements Controller.
-func (lbc *LoadBalancerController) GCLoadBalancers(toKeep []*v1beta1.Ingress) error {
-	// Only GCE ingress associated resources are managed by this controller.
-	GCEIngresses := operator.Ingresses(toKeep).Filter(utils.IsGCEIngress).AsList()
-	return lbc.l7Pool.GC(common.ToIngressKeys(GCEIngresses))
+// GCv1LoadBalancers implements Controller.
+func (lbc *LoadBalancerController) GCv1LoadBalancers(toKeep []*v1beta1.Ingress) error {
+	return lbc.l7Pool.GCv1(common.ToIngressKeys(toKeep))
 }
 
-// MaybeRemoveFinalizers cleans up Finalizers if needed.
-func (lbc *LoadBalancerController) MaybeRemoveFinalizers(toCleanup []*v1beta1.Ingress) error {
+// GCv2LoadBalancer implements Controller.
+func (lbc *LoadBalancerController) GCv2LoadBalancer(ing *v1beta1.Ingress) error {
+	return lbc.l7Pool.GCv2(ing)
+}
+
+// EnsureDeleteV1Finalizers implements Controller.
+func (lbc *LoadBalancerController) EnsureDeleteV1Finalizers(toCleanup []*v1beta1.Ingress) error {
 	if !flags.F.FinalizerRemove {
 		klog.V(4).Infof("Removing finalizers not enabled")
 		return nil
 	}
 	for _, ing := range toCleanup {
 		ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
-		if err := common.RemoveFinalizer(ing, ingClient); err != nil {
-			klog.Errorf("Failed to remove Finalizer from Ingress %s/%s: %v", ing.Namespace, ing.Name, err)
+		if err := common.EnsureDeleteFinalizer(ing, ingClient, common.FinalizerKey); err != nil {
+			klog.Errorf("Failed to ensure delete finalizer %s for ingress %s: %v", common.FinalizerKey, common.NamespacedName(ing), err)
 			return err
 		}
+	}
+	return nil
+}
+
+// EnsureDeleteV2Finalizer implements Controller.
+func (lbc *LoadBalancerController) EnsureDeleteV2Finalizer(ing *v1beta1.Ingress) error {
+	if !flags.F.FinalizerRemove {
+		klog.V(4).Infof("Removing finalizers not enabled")
+		return nil
+	}
+	ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+	if err := common.EnsureDeleteFinalizer(ing, ingClient, common.FinalizerKeyV2); err != nil {
+		klog.Errorf("Failed to ensure delete finalizer %s for ingress %s: %v", common.FinalizerKeyV2, common.NamespacedName(ing), err)
+		return err
 	}
 	return nil
 }
@@ -501,20 +519,24 @@ func (lbc *LoadBalancerController) sync(key string) error {
 		return fmt.Errorf("error getting Ingress for key %s: %v", key, err)
 	}
 
-	// Snapshot of list of ingresses.
+	// Capture GC state for ingress.
 	allIngresses := lbc.ctx.Ingresses().List()
+
 	// Determine if the ingress needs to be GCed.
 	if !ingExists || utils.NeedsCleanup(ing) {
+		frontendGCAlgorithm := frontendGCAlgorithm(ingExists, ing)
+		klog.V(3).Infof("Using algorithm %v to GC ingress %v", frontendGCAlgorithm, ing)
 		// GC will find GCE resources that were used for this ingress and delete them.
-		return lbc.ingSyncer.GC(allIngresses)
+		err := lbc.ingSyncer.GC(allIngresses, ing, frontendGCAlgorithm)
+		if err != nil {
+			lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, "GC", fmt.Sprintf("Error during GC: %v", err))
+		}
+		return err
 	}
 
-	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
-	ing = ing.DeepCopy()
-	ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+	// Ensure that a finalizer is attached.
 	if flags.F.FinalizerAdd {
-		if err := common.AddFinalizer(ing, ingClient); err != nil {
-			klog.Errorf("Failed to add Finalizer to Ingress %q: %v", key, err)
+		if err = lbc.ensureFinalizer(ing); err != nil {
 			return err
 		}
 	}
@@ -538,7 +560,10 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	// Garbage collection will occur regardless of an error occurring. If an error occurred,
 	// it could have been caused by quota issues; therefore, garbage collecting now may
 	// free up enough quota for the next sync to pass.
-	if gcErr := lbc.ingSyncer.GC(allIngresses); gcErr != nil {
+	frontendGCAlgorithm := frontendGCAlgorithm(ingExists, ing)
+	klog.V(3).Infof("Using algorithm %v to GC ingress %v", frontendGCAlgorithm, ing)
+	if gcErr := lbc.ingSyncer.GC(allIngresses, ing, frontendGCAlgorithm); gcErr != nil {
+		lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, "GC", fmt.Sprintf("Error during GC: %v", gcErr))
 		return fmt.Errorf("error during sync %v, error during GC %v", syncErr, gcErr)
 	}
 
@@ -648,4 +673,92 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
 		knownPorts = append(knownPorts, urlMap.AllServicePorts()...)
 	}
 	return knownPorts
+}
+
+// defaultFrontendNamingScheme returns frontend naming scheme for an ingress without finalizer.
+// This is used for adding an appropriate finalizer on the ingress.
+func (lbc *LoadBalancerController) defaultFrontendNamingScheme(ing *v1beta1.Ingress) (namer.Scheme, error) {
+	// Ingress frontend naming scheme is determined based on the following logic,
+	// V2 frontend namer is disabled         : v1 frontend naming scheme
+	// V2 frontend namer is enabled
+	//     - VIP does not exists             : v2 frontend naming scheme
+	//     - VIP exists
+	//         - GCE URL Map exists          : v1 frontend naming scheme
+	//         - GCE URL Map does not exists : v2 frontend naming scheme
+	if !flags.F.EnableV2FrontendNamer {
+		return namer.V1NamingScheme, nil
+	}
+	if !utils.HasVIP(ing) {
+		return namer.V2NamingScheme, nil
+	}
+	urlMapExists, err := lbc.l7Pool.HasUrlMap(ing)
+	if err != nil {
+		return "", err
+	}
+	if urlMapExists {
+		return namer.V1NamingScheme, nil
+	}
+	return namer.V2NamingScheme, nil
+}
+
+// ensureFinalizer ensures that a finalizer is attached.
+func (lbc *LoadBalancerController) ensureFinalizer(ing *v1beta1.Ingress) error {
+	ingKey := common.NamespacedName(ing)
+	if common.HasFinalizer(ing.ObjectMeta) {
+		klog.V(4).Infof("Finalizer exists for ingress %s", ingKey)
+		return nil
+	}
+	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
+	ing = ing.DeepCopy()
+	ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+	namingScheme, err := lbc.defaultFrontendNamingScheme(ing)
+	if err != nil {
+		return err
+	}
+	finalizerKey, err := namer.FinalizerForNamingScheme(namingScheme)
+	if err != nil {
+		return err
+	}
+	if err := common.EnsureFinalizer(ing, ingClient, finalizerKey); err != nil {
+		klog.Errorf("Failed to ensure finalizer %s for ingress %s: %v", finalizerKey, ingKey, err)
+		return err
+	}
+	return nil
+}
+
+// frontendGCAlgorithm returns the naming scheme using which frontend resources needs to be cleanedup.
+// This also returns a boolean to specify if we need to delete frontend resources.
+// GC path is
+// If ingress does not exist :   v1 frontends and all backends
+// If ingress exists
+//    - Needs cleanup
+//      - If v1 naming scheme  :    v1 frontends and all backends
+//      - If v2 naming scheme  :    v2 frontends and all backends
+//    - Does not need cleanup
+//      - Finalizer enabled    :    all backends
+//      - Finalizer disabled   :    v1 frontends and all backends
+func frontendGCAlgorithm(ingExists bool, ing *v1beta1.Ingress) utils.FrontendGCAlgorithm {
+	// If ingress does not exist, that means its pre-finalizer era.
+	// Run GC via v1 naming scheme.
+	if !ingExists {
+		return utils.CleanupV1FrontendResources
+	}
+	// Determine if we do not need to delete current ingress.
+	if !utils.NeedsCleanup(ing) {
+		// GC backends only if current ingress does not need cleanup and finalizers is enabled.
+		if flags.F.FinalizerAdd {
+			return utils.NoCleanUpNeeded
+		}
+		return utils.CleanupV1FrontendResources
+	}
+	namingScheme := namer.FrontendNamingScheme(ing)
+	switch namingScheme {
+	case namer.V2NamingScheme:
+		return utils.CleanupV2FrontendResources
+	case namer.V1NamingScheme:
+		return utils.CleanupV1FrontendResources
+	default:
+		klog.Errorf("Unexpected naming scheme %v", namingScheme)
+		return utils.NoCleanUpNeeded
+	}
 }
--- a/pkg/controller/translator/translator_test.go
+++ b/pkg/controller/translator/translator_test.go
@@ -65,7 +65,7 @@ func fakeTranslator() *Translator {
 		HealthCheckPath:               "/",
 		DefaultBackendHealthCheckPath: "/healthz",
 	}
-	ctx := context.NewControllerContext(nil, client, backendConfigClient, nil, nil, defaultNamer, ctxConfig)
+	ctx := context.NewControllerContext(nil, client, backendConfigClient, nil, nil, defaultNamer, "" /*kubeSystemUID*/, ctxConfig)
 	gce := &Translator{
 		ctx: ctx,
 	}
--- a/pkg/firewalls/controller_test.go
+++ b/pkg/firewalls/controller_test.go
@@ -46,7 +46,7 @@ func newFirewallController() *FirewallController {
 		DefaultBackendSvcPort: test.DefaultBeSvcPort,
 	}
 
-	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, fakeGCE, defaultNamer, ctxConfig)
+	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, fakeGCE, defaultNamer, "" /*kubeSystemUID*/, ctxConfig)
 	fwc := NewFirewallController(ctx, []string{"30000-32767"})
 	fwc.hasSynced = func() bool { return true }
 
--- a/pkg/loadbalancers/interfaces.go
+++ b/pkg/loadbalancers/interfaces.go
@@ -17,17 +17,20 @@ limitations under the License.
 package loadbalancers
 
 import (
-	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	"k8s.io/ingress-gce/pkg/composite"
-	"k8s.io/ingress-gce/pkg/loadbalancers/features"
+	"k8s.io/api/networking/v1beta1"
 )
 
 // LoadBalancerPool is an interface to manage the cloud resources associated
 // with a gce loadbalancer.
 type LoadBalancerPool interface {
+	// Ensure ensures a loadbalancer and its resources given the RuntimeInfo.
 	Ensure(ri *L7RuntimeInfo) (*L7, error)
-	Delete(name string, versions *features.ResourceVersions, scope meta.KeyType) error
-	GC(names []string) error
-	Shutdown() error
-	List(key *meta.Key, version meta.Version) ([]*composite.UrlMap, error)
+	// GCv2 garbage collects loadbalancer associated with given ingress using v2 naming scheme.
+	GCv2(ing *v1beta1.Ingress) error
+	// GCv1 garbage collects loadbalancers not in the input list using v1 naming scheme.
+	GCv1(names []string) error
+	// Shutdown deletes all loadbalancers for given list of ingresses.
+	Shutdown(ings []*v1beta1.Ingress) error
+	// HasUrlMap returns true if an URL map exists in GCE for given ingress.
+	HasUrlMap(ing *v1beta1.Ingress) (bool, error)
 }
--- a/pkg/loadbalancers/l7s.go
+++ b/pkg/loadbalancers/l7s.go
@@ -18,13 +18,18 @@ package loadbalancers
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/common/operator"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/events"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/common"
 	namer_util "k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 	"k8s.io/legacy-cloud-providers/gce"
@@ -52,7 +57,7 @@ func NewLoadBalancerPool(cloud *gce.Cloud, v1NamerHelper namer_util.V1FrontendNa
 	}
 }
 
-// Ensure ensures a loadbalancer and its resources given the RuntimeInfo
+// Ensure implements LoadBalancerPool.
 func (l *L7s) Ensure(ri *L7RuntimeInfo) (*L7, error) {
 	lb := &L7{
 		runtimeInfo: ri,
@@ -69,8 +74,8 @@ func (l *L7s) Ensure(ri *L7RuntimeInfo) (*L7, error) {
 	return lb, nil
 }
 
-// Delete deletes a load balancer by name.
-func (l *L7s) Delete(name string, versions *features.ResourceVersions, scope meta.KeyType) error {
+// delete deletes a loadbalancer by name.
+func (l *L7s) delete(name string, versions *features.ResourceVersions, scope meta.KeyType) error {
 	lb := &L7{
 		runtimeInfo: &L7RuntimeInfo{},
 		cloud:       l.cloud,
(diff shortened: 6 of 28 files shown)


pkg/loadbalancers/l7s.go
```diff
@@ -86,6 +90,23 @@ func (l *L7s) Delete(name string, versions *features.ResourceVersions, scope met
 	return nil
 }
 
+// V2Delete deletes a load balancer (using v2 naming scheme) by ingress.
```
freehan: V2Delete deletes a load balancer frontend resources by ingressusing v2 naming scheme.

pkg/loadbalancers/interfaces.go
```diff
@@ -18,16 +18,26 @@ package loadbalancers
...
 	Delete(name string, versions *features.ResourceVersions, scope meta.KeyType) error
+	// V2GC garbage collects loadbalancer associated with given ingress using v2 naming scheme.
+	V2GC(ing *v1beta1.Ingress) error
+	// GC garbage collects loadbalancers not in the input list using v1 naming scheme.
 	GC(names []string) error
-	Shutdown() error
+	// Shutdown deletes all loadbalancers for given list of ingresses.
+	Shutdown(ings []*v1beta1.Ingress) error
+	// List returns a list of urlMaps (the top level LB resource) that belong to the cluster.
 	List(key *meta.Key, version meta.Version) ([]*composite.UrlMap, error)
```
freehan: It looks like this is not used? remove?

pkg/controller/controller.go
```diff
@@ -649,3 +691,25 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+	// Ingress frontend naming scheme is determined based on the following logic,
+	// V2 frontend namer is disabled         : v1 frontend naming scheme
+	// V2 frontend namer is enabled
+	//     - VIP does not exists             : v2 frontend naming scheme
+	//     - VIP exists
+	//         - GCE URL Map exists          : v1 frontend naming scheme
+	//         - GCE URL Map does not exists : v2 frontend naming scheme
+	if !flags.F.EnableV2FrontendNamer {
+		return true, nil
+	} else if utils.HasVIP(ing) {
```
freehan: no need to have a else if here.
```
if !utils.HasVIP(ing) {
    return false, nil
}

urlMapExists, err := lbc.l7Pool.HasUrlMap(ing); 

if err != nil {
 return false, err
}

if urlMapExists {
    return true, nil
}
return false, nil 
```

pkg/controller/controller.go
```diff
@@ -115,7 +115,7 @@ func NewLoadBalancerController(
 		hasSynced:     ctx.HasSynced,
 		nodes:         NewNodeController(ctx, instancePool),
 		instancePool:  instancePool,
-		l7Pool:        loadbalancers.NewLoadBalancerPool(ctx.Cloud, ctx.ClusterNamer, ctx, namer.NewFrontendNamerFactory(ctx.ClusterNamer)),
+		l7Pool:        loadbalancers.NewLoadBalancerPool(ctx.Cloud, ctx.ClusterNamer, ctx, namer.NewFrontendNamerFactory(ctx.ClusterNamer, ctx.KubeSystemUID)),
```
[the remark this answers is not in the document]
skmatti: ctx might too broad to pass it to LoadBalancerPool. Is it desirable to move FrontendNamerFactory to ctx even if we still have to pass in the 4 variables?.

pkg/controller/controller.go
```diff
@@ -505,17 +523,41 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
 	// Determine if the ingress needs to be GCed.
 	if !ingExists || utils.NeedsCleanup(ing) {
-		// GC will find GCE resources that were used for this ingress and delete them.
+		// GC workflow for ingress with v2 naming policy.
+		// 1. GC load balancer for ingress being synced.
+		// 2. GC load balancers for v1 naming policy.
+		// 3. GC backends.
+		// 4. Ensure delete finalizers for v1 naming scheme.
+		// 5. Ensure delete finalizers for ingress being synced.
+		if namer.IsV2NamingScheme(ing) {
```
freehan: Here sort of breaks the abstraction among: controller, syncer, l7pool 

@rramkumar1 I am wondering if we can retire the ingSyncer interface all together. 
If so, I am thinking to refactor IngSyncer and Controller interface.

pkg/controller/controller.go
```diff
@@ -505,17 +523,41 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
 	}
 
 	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
 	ing = ing.DeepCopy()
 	ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
-	if flags.F.FinalizerAdd {
-		if err := common.AddFinalizer(ing, ingClient); err != nil {
-			klog.Errorf("Failed to add Finalizer to Ingress %q: %v", key, err)
+	if flags.F.FinalizerAdd && !common.HasFinalizer(ing.ObjectMeta) {
+		// Determine if we need to use v1 frontend naming scheme.
```
freehan: This section can be enclose into one function. `EnsureFinalizer`

pkg/loadbalancers/interfaces.go
```diff
@@ -18,16 +18,26 @@ package loadbalancers
...
 // with a gce loadbalancer.
 type LoadBalancerPool interface {
+	// Ensure ensures a loadbalancer and its resources given the RuntimeInfo.
 	Ensure(ri *L7RuntimeInfo) (*L7, error)
+	// Delete deletes a loadbalancer by name.
 	Delete(name string, versions *features.ResourceVersions, scope meta.KeyType) error
+	// V2GC garbage collects loadbalancer associated with given ingress using v2 naming scheme.
+	V2GC(ing *v1beta1.Ingress) error
+	// GC garbage collects loadbalancers not in the input list using v1 naming scheme.
 	GC(names []string) error
```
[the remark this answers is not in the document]
skmatti: Done

pkg/sync/interfaces.go
```diff
@@ -40,10 +40,11 @@ type Controller interface {
 	GCBackends(toKeep []*v1beta1.Ingress) error
 	// SyncLoadBalancer syncs the front-end load balancer resources for a GCLB given some existing state.
 	SyncLoadBalancer(state interface{}) error
-	// GCLoadBalancers garbage collects front-end load balancer resources for all ingresses given a list of ingresses to exclude from GC.
-	GCLoadBalancers(toKeep []*v1beta1.Ingress) error
+	// GCV1LoadBalancers garbage collects front-end load balancer resources for all ingresses
+	// given a list of ingresses with v1 naming policy to exclude from GC.
+	GCV1LoadBalancers(toKeep []*v1beta1.Ingress) error
```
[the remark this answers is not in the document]
skmatti: Done

pkg/loadbalancers/interfaces.go
skmatti: It being used in GC.
https://github.com/kubernetes/ingress-gce/blob/66038edb53c9e12fc82329e9d99d9eb7c73b3b23/pkg/loadbalancers/l7s.go#L133

pkg/sync/interfaces.go
```diff
@@ -40,10 +40,11 @@ type Controller interface {
...
-	GCLoadBalancers(toKeep []*v1beta1.Ingress) error
+	// GCv1LoadBalancers garbage collects front-end load balancer resources for all ingresses
+	// given a list of ingresses with v1 naming policy to exclude from GC.
+	GCv1LoadBalancers(toKeep []*v1beta1.Ingress) error
 	// PostProcess allows for doing some post-processing after an Ingress is synced to a GCLB.
 	PostProcess(state interface{}) error
-	// MaybeRemoveFinalizers removes finalizers from a list of ingresses one by one after all of its associated resources are deleted.
-	MaybeRemoveFinalizers(toCleanup []*v1beta1.Ingress) error
+	// EnsureDeleteV1Finalizers ensures that v1 finalizers are removed for given list of ingresses.
+	EnsureDeleteV1Finalizers(toCleanup []*v1beta1.Ingress) error
```
freehan: See if this makes sense @rramkumar1 @skmatti 

```
IngState {
	Ing    *v1beta1.Ingress
	AllIng []*v1beta1.Ingress
}
func NewIngState
func(state IngState) GetV1ToKeep()
func(state IngState) GetAllToKeep()
func(state IngState) GetV1ToCleanup()


SyncContext {
	IngState
	UrlMap *utils.GCEURLMap
	L7     *loadbalancers.L7
}


type Syncer inteface {
   Sync(syncContext)
   GC(ingState)
}

// Controller is an interface for ingress controllers and declares methods
// on how to sync the various portions of the GCLB for an Ingress.
type Controller interface {
	// SyncBackends syncs the backends for a GCLB given some existing state.
	SyncBackends(syncContext SyncContext) error
	// SyncFrontend syncs the front-end load balancer resources for a GCLB given some existing state.
	SyncFrontend(syncContext SyncContext) error
	// PostProcess allows for doing some post-processing after an Ingress is synced to a GCLB.
	PostProcess(syncContext SyncContext) error
	// EnsureDeleteFinalizers removes finalizers from a list of ingresses one by one after all of its associated resources are deleted.
	EnsureDeleteFinalizers(ingState IngState) error
	// GCBackends garbage collects backends for all ingresses given a list of ingresses to exclude from GC.
	GCBackends(ingState IngState) error
	// GCFrontends garbage collects front-end load balancer resources for all ingresses given a list of ingresses to exclude from GC.
	GCFrontends(ingState IngState) error
}
```

pkg/loadbalancers/interfaces.go
skmatti: Discussed offline. This is used in tests, so we keep this.

pkg/utils/namer/utils.go
```diff
@@ -58,12 +58,21 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
...
 		return V1NamingScheme
 	default:
-		klog.V(3).Infof("No finalizer found on Ingress %v, using Legacy naming scheme", ing)
+		klog.V(3).Infof("No finalizer found on Ingress %v, using V1 naming scheme", ing)
 		return V1NamingScheme
 	}
 }
+
+// IsV2NamingScheme returns true if given ingress uses v2 naming scheme.
+func IsV2NamingScheme(ing *v1beta1.Ingress) bool {
```
freehan: I feel like there should be a suite of helper function like this:
```

// retrieve naming scheme from finalizer
func GetNamingSchemeFromFinalizer(ing)  (NamingScheme, error)

// calculates default naming scheme for a ingress without finalizer
func GetDefaultNamingScheme(ing, lbPool)  (NamingScheme, error)
```

pkg/utils/namer/utils.go
```diff
@@ -58,12 +58,21 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
 
 // frontendNamingScheme returns naming scheme for given ingress.
 func frontendNamingScheme(ing *v1beta1.Ingress) Scheme {
```
freehan: probably rename this to GetNamingSchemeFromFinalizer

pkg/controller/controller.go
```diff
@@ -649,3 +662,53 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+
+// ensureFinalizer ensures that a finalizer is attached.
+func (lbc *LoadBalancerController) ensureFinalizer(ing *v1beta1.Ingress) error {
+	// Get ingress and DeepCopy for assurance that we don't pollute other goroutines with changes.
+	ing = ing.DeepCopy()
+	ingKey := common.NamespacedName(ing)
+	ingClient := lbc.ctx.KubeClient.NetworkingV1beta1().Ingresses(ing.Namespace)
+	if flags.F.FinalizerAdd && !common.HasFinalizer(ing.ObjectMeta) {
+		// Determine if we need to use v1 frontend naming scheme.
+		if isV1NamingScheme, err := lbc.needsToUseV1NamingScheme(ing); err != nil {
```
freehan: nit:
namingScheme, err := util.GetDefaultNamingScheme...
common.EnsureFinalzier(ing, ingClient, namingScheme)

pkg/sync/sync.go
```diff
@@ -61,21 +62,38 @@ func (s *IngressSyncer) Sync(state interface{}) error {
 }
 
 // GC implements Syncer.
-func (s *IngressSyncer) GC(ings []*v1beta1.Ingress) error {
+func (s *IngressSyncer) GC(ings []*v1beta1.Ingress, currIng *v1beta1.Ingress) error {
+	isV2NamingScheme := namer.IsV2NamingScheme(currIng)
```
freehan: What if we do not GCv1LB when currIng is v2?

```
namingScheme, err := GetNamingScheme(currIng)

swtich (namingScheme) {
case v2:
     GCv2LB
     defer func() {if err == nil  EnsureDeleteV2Fianlizer }
case v1:
      GCv1LBs
     defer func() {if err == nil  EnsureDeleteV1Fianlizer }
default: 
    return err      
}
err := GCBackends


```

pkg/utils/common/finalizer.go
```diff
@@ -23,52 +23,91 @@ import (
...
-// FinalizerKey is the string representing the Ingress finalizer.
-const FinalizerKey = "networking.gke.io/ingress-finalizer"
+const (
+	// FinalizerKey is the string representing the Ingress finalizer.
+	FinalizerKey = "networking.gke.io/ingress-finalizer"
+	// FinalizerKeyV2 is the string representing the Ingress finalizer version.
+	// Ingress with V2 finalizer uses V2 frontend naming scheme.
+	FinalizerKeyV2 = "networking.gke.io/ingress-finalizer-V2"
+)
+
```
freehan: probably maintain a map between NamingScheme and FinalizerKey


```
map[namer.NamingScheme]string{
   namer.V1NamingScheme: "networking.gke.io/ingress-finalizer",
   namer.V2NamingScheme: "networking.gke.io/ingress-finalizer-v2"
}


GetNamingSchemeFromFinalizer(ing) (namingScheme, error)
EnsureFinalizer(ing, namingScheme)
EnsureDeleteFinalizer(ing, namingScheme)
```

pkg/controller/controller.go
```diff
@@ -538,7 +549,8 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	// Garbage collection will occur regardless of an error occurring. If an error occurred,
 	// it could have been caused by quota issues; therefore, garbage collecting now may
 	// free up enough quota for the next sync to pass.
-	if gcErr := lbc.ingSyncer.GC(allIngresses); gcErr != nil {
+	// Note that we pass `nil` here because the current ingress should not to be GCed.
+	if gcErr := lbc.ingSyncer.GC(allIngresses, nil); gcErr != nil {
```
freehan: try not to trigger GC V1 LBs for V2 Ingress.

pkg/utils/namer/utils.go
```diff
@@ -56,14 +58,27 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
 	return ret
 }
 
-// frontendNamingScheme returns naming scheme for given ingress.
-func frontendNamingScheme(ing *v1beta1.Ingress) Scheme {
-	// TODO(smatti): return V2NamingScheme if ingress has V2 finalizer
+// FrontendNamingScheme returns naming scheme for given ingress based on its finalizer.
+func FrontendNamingScheme(ing *v1beta1.Ingress) Scheme {
```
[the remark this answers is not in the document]
skmatti: Discussed offline that this function should be deterministic. We rely on sync function ensure that a finalizer is attached when it should.

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,24 +108,119 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+// NewFrontendNamerFactory returns IngressFrontendNamerFactory given a v1 namer and uid.
+func NewFrontendNamerFactory(namer *Namer, uid types.UID) IngressFrontendNamerFactory {
+	clusterUID := common.ContentHash(string(uid), clusterUIDLength)
+	return &FrontendNamerFactory{namer: namer, clusterUID: clusterUID}
 }
 
 // Namer implements IngressFrontendNamerFactory.
 func (rn *FrontendNamerFactory) Namer(ing *v1beta1.Ingress) IngressFrontendNamer {
-	switch frontendNamingScheme(ing) {
+	switch FrontendNamingScheme(ing) {
```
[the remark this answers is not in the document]
skmatti: Done

pkg/controller/controller.go
```diff
@@ -649,3 +661,58 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+}
+
+// ensureFinalizer ensures that a finalizer is attached.
+func (lbc *LoadBalancerController) ensureFinalizer(ing *v1beta1.Ingress) error {
+	ingKey := common.NamespacedName(ing)
+	if common.HasFinalizer(ing.ObjectMeta) {
+		klog.V(4).Infof("Finalizer exists for ingress %s", ingKey)
+		return nil
+	}
+	if !flags.F.FinalizerAdd {
```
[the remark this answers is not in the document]

pkg/controller/controller.go
```diff
@@ -503,45 +520,69 @@ func (lbc *LoadBalancerController) sync(key string) error {
 
 	// Snapshot of list of ingresses.
 	allIngresses := lbc.ctx.Ingresses().List()
+
+	var syncErr error
+	// Perform GC as a deferred function.
+	defer func() {
```
freehan: I think this can be simplified to 3 step:
err != nil
determine gcPath
determine call or not to call GC

pkg/sync/sync.go
```diff
@@ -61,21 +62,55 @@ func (s *IngressSyncer) Sync(state interface{}) error {
...
+		switch gcPath {
+		case namer.V2NamingScheme:
+			lbErr = s.controller.GCv2LoadBalancer(currIng)
+
+			defer func() {
+				if lbErr != nil || beErr != nil {
+					return
+				}
+				err = s.controller.EnsureDeleteV2Finalizer(currIng)
+			}()
```
freehan: you need to break here right? Same below.
otherwise, it will continue to V1 GC after V2 is done.

pkg/sync/sync.go
```diff
@@ -61,21 +62,55 @@ func (s *IngressSyncer) Sync(state interface{}) error {
...
+	if gcFrontends {
+		switch gcPath {
+		case namer.V2NamingScheme:
+			lbErr = s.controller.GCv2LoadBalancer(currIng)
+
+			defer func() {
+				if lbErr != nil || beErr != nil {
+					return
+				}
+				err = s.controller.EnsureDeleteV2Finalizer(currIng)
```
freehan: This error may not be the actual return error. Same below.

pkg/sync/sync.go
```diff
@@ -61,21 +62,55 @@ func (s *IngressSyncer) Sync(state interface{}) error {
 }
 
 // GC implements Syncer.
-func (s *IngressSyncer) GC(ings []*v1beta1.Ingress) error {
-	// Partition GC state into ingresses those need cleanup and those don't.
+// TODO(smatti): Change this to take in GC state and GC options.
+func (s *IngressSyncer) GC(ings []*v1beta1.Ingress, currIng *v1beta1.Ingress, gcPath namer.Scheme, gcFrontends bool) error {
```
freehan: consider the following.
```
// step 1: GC frontends
if gcFrontends {
    switch GC  LB
    ...
}

// step 2: GC backends

// step 3: EnsureDeleteFinalizer
if gcFrontends {
   swtich gcPath
   ...
}


```


This makes the flow simpler and error handling easier. No more deferred function

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,24 +108,119 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+
+// newV2IngressFrontendNamer returns a v2 frontend namer for given ingress, cluster uid and prefix.
+func newV2IngressFrontendNamer(ing *v1beta1.Ingress, clusterUID string, prefix string) IngressFrontendNamer {
+	namer := &V2IngressFrontendNamer{ing: ing, prefix: prefix, clusterUID: clusterUID}
+	namer.setLbName()
+	return namer
+}
+
+// ForwardingRule returns the name of forwarding rule based on given protocol.
+func (vn *V2IngressFrontendNamer) ForwardingRule(protocol NamerProtocol) string {
```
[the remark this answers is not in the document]
freehan: ping

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,24 +108,119 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+// lbNameToHash returns hash string of length 16 of lbName.
+func (vn *V2IngressFrontendNamer) lbNameToHash() string {
+	return common.ContentHash(vn.lbName, 16)
+}
+
 // FrontendNamerFactory implements IngressFrontendNamerFactory.
 type FrontendNamerFactory struct {
-	namer *Namer
+	namer      *Namer
+	clusterUID string
```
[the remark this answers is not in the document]
freehan: ping

pkg/sync/sync.go
```diff
@@ -61,21 +62,55 @@ func (s *IngressSyncer) Sync(state interface{}) error {
...
-	toKeepIngresses := toKeep.AsList()
-	lbErr := s.controller.GCLoadBalancers(toKeepIngresses)
-	beErr := s.controller.GCBackends(toKeepIngresses)
+	// 2) It is not a deletion candidate. A deletion candidate is an ingress
+	//    with deletion stamp and a finalizer.
+	toKeep := operator.Ingresses(ings).Filter(func(ing *v1beta1.Ingress) bool {
+		return !utils.NeedsCleanup(ing)
+	}).AsList()
+	beErr = s.controller.GCBackends(toKeep)
 	if lbErr != nil {
```
freehan: []error

return AggregateError
sample:
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/readiness/poller.go#L165

pkg/sync/sync.go
skmatti: Discussed offline. Does not need a break statement.

pkg/sync/sync.go
skmatti: Fixed this to use same error variable.

pkg/sync/sync.go
skmatti: Discussed offline to keep the original structure.

pkg/sync/sync.go
skmatti: Done

pkg/utils/utils.go
```diff
@@ -383,5 +383,16 @@ func ServiceKeyFunc(namespace, name string) string {
 
 // NeedsCleanup returns true if the ingress needs to have its associated resources deleted.
 func NeedsCleanup(ing *v1beta1.Ingress) bool {
-	return common.IsDeletionCandidate(ing.ObjectMeta, common.FinalizerKey) || !IsGLBCIngress(ing)
+	return common.IsDeletionCandidate(ing.ObjectMeta) || !IsGLBCIngress(ing)
+}
+
+// HasVIP returns true if given ingress has a vip.
```
rramkumar1: This should have a unit test

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
 	return ln.lbName
 }
 
+// V2IngressFrontendNamer implements IngresFrontendNamer.
```
rramkumar1: Spelling

pkg/utils/namer/frontendnamer.go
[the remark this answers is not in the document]
rramkumar1: Please address this.

pkg/utils/namer/frontendnamer.go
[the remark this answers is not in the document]
rramkumar1: Please address this

pkg/utils/namer/frontendnamer_test.go
```diff
@@ -237,3 +240,152 @@ func TestV1IngressFrontendNamer(t *testing.T) {
...
+			"ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+			"%s2-tp-ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+			"%s2-ts-ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+			"%s2-cr-ksuid123-7trbunxyryt39g1b-%s",
+			"%s2-fr-ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+			"%s2-fs-ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+			"%s2-um-ksuid123-012345678901234567-012345678901234567-4mwbp6m5",
+		},
+	}
+	for _, prefix := range []string{"k8s", "mci"} {
```
rramkumar1: Why is "mci" here?

cmd/glbc/main.go
```diff
@@ -123,6 +125,16 @@ func main() {
 		klog.V(0).Infof("Cluster name: %+v", namer.UID())
 	}
 
+	var kubeSystemUID types.UID
+	if flags.F.EnableV2FrontendNamer {
+		// Get kube-system UID that will be used for v2 frontend naming scheme.
+		ksNameSpace, err := kubeClient.CoreV1().Namespaces().Get("kube-system", metav1.GetOptions{})
+		if err != nil {
+			klog.Fatalf("kubeClient.CoreV1().Namespaces().Get(%q, _): %v, want nil", "kube-system", err)
```
rramkumar1: This sort of error format is typically only used in tests. Just say "Error getting kube-system namespace"

cmd/glbc/main.go
```diff
@@ -123,6 +125,16 @@ func main() {
 		klog.V(0).Infof("Cluster name: %+v", namer.UID())
 	}
 
+	var kubeSystemUID types.UID
+	if flags.F.EnableV2FrontendNamer {
+		// Get kube-system UID that will be used for v2 frontend naming scheme.
+		ksNameSpace, err := kubeClient.CoreV1().Namespaces().Get("kube-system", metav1.GetOptions{})
```
rramkumar1: kubeSystemNS

pkg/test/utils.go
```diff
@@ -80,3 +86,27 @@ func DecodeIngress(data []byte) (*v1beta1.Ingress, error) {
 
 	return obj.(*v1beta1.Ingress), nil
 }
+
+// Flag is a type representing controller flag.
+type Flag string
```
rramkumar1: What is this used for? I don't see it used anywhere in this PR.

pkg/utils/namer/frontendnamer.go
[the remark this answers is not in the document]
skmatti: Added in the function definition:
https://github.com/kubernetes/ingress-gce/pull/892/files#diff-e5e95bb1cdbf35e24dc0d1c558762691R125-R134

pkg/utils/namer/frontendnamer.go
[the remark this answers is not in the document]
skmatti: Done

pkg/utils/namer/frontendnamer_test.go
skmatti: Testing this with different prefixes, it could have been any random word.

cmd/glbc/main.go
skmatti: Done

pkg/test/utils.go
skmatti: This is used for saving flags when overriding them in unit tests. This potentially other tests from using the value set by a test. 
Made this variable private as all of its uses are within the file.

pkg/controller/controller.go
```diff
@@ -503,45 +520,53 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
```
bowei: I very much dislike doing complex actions like this in a defer call as it is very hard to understand the code flow (almost like exception handling).

Can we make this a helper function and call it on the appropriate error paths?

pkg/controller/controller.go
```diff
@@ -503,45 +520,53 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
 
 	// Bootstrap state for GCP sync.
 	urlMap, errs := lbc.Translator.TranslateIngress(ing, lbc.ctx.DefaultBackendSvcPort.ID, lbc.ctx.ClusterNamer)
 
 	if errs != nil {
-		msg := fmt.Errorf("error while evaluating the ingress spec: %v", utils.JoinErrs(errs))
-		lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, "Translate", msg.Error())
-		return msg
+		err = fmt.Errorf("error while evaluating the ingress spec: %v", utils.JoinErrs(errs))
+		lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, "Translate", err.Error())
```
bowei: Is there a reason why we don't always emit an event on all error paths?

pkg/controller/controller.go
```diff
@@ -649,3 +674,87 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+	if urlMapExists {
+		return namer.V1NamingScheme, nil
+	}
+	return namer.V2NamingScheme, nil
+}
+
+// ensureFinalizer ensures that a finalizer is attached.
+func (lbc *LoadBalancerController) ensureFinalizer(ing *v1beta1.Ingress) error {
+	if !flags.F.FinalizerAdd {
+		klog.V(4).Infof("Adding finalizers not enabled")
```
bowei: Error log

pkg/controller/controller.go
```diff
@@ -649,3 +674,87 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+	namingScheme, err := lbc.defaultFrontendNamingScheme(ing)
+	if err != nil {
+		return err
+	}
+	finalizerKey, err := namer.FinalizerForNamingScheme(namingScheme)
+	if err != nil {
+		return err
+	}
+	if err := common.EnsureFinalizer(ing, ingClient, finalizerKey); err != nil {
+		klog.Errorf("common.EnsureFinalizer(%q, _, %q) = %v, want nil", ingKey, finalizerKey, err)
```
bowei: leave out ", want nil", it's redundant

bowei: Does this use the optimistic concurrency control on write? (see https://github.com/eBay/Kubernetes/blob/master/docs/devel/api-conventions.md#concurrency-control-and-consistency)

bowei: Seems like underlying call should be a "PATCH" -- can fix this later, file a bug.

pkg/controller/controller.go
```diff
@@ -649,3 +674,87 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+// GC path is
+// If ingress does not exist :   v1 frontends and all backends
+// If ingress exists
+//    - Needs cleanup
+//      - If v1 naming scheme  :    v1 frontends and all backends
+//      - If v2 naming scheme  :    v2 frontends and all backends
+//    - Does not need cleanup
+//      - Finalizer enabled    :    all backends
+//      - Finalizer disabled   :    v1 frontends and all backends
+func gcPath(ingExists bool, ing *v1beta1.Ingress) (namer.Scheme, bool) {
```
bowei: shouldn't we call this  gcNamingScheme?

bowei: Can we make the return value less error prone? I think you want to return one of three values, but it's being expressed as a pair `({v1|v2}, bool)`
```
type gcAlgorithm int

const (
  noCleanupNeeded gcAlgorithm = iota
  cleanupV1FrontendResources
  cleanupV2FrontendResources
)

func gcAlgorithmForIngress(ingExists ...) gcAlgorithm {
  ...
}

File: pkg/firewalls/controller_test.go [hunk 37]
```diff
@@ -46,7 +46,7 @@ func newFirewallController() *FirewallController {
 		DefaultBackendSvcPort: test.DefaultBeSvcPort,
 	}
 
-	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, fakeGCE, defaultNamer, ctxConfig)
+	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, fakeGCE, defaultNamer, "", ctxConfig)
```
bowei: for raw consts empty string, we usually put the name of the variable to aid the reader:
`ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, fakeGCE, defaultNamer, /*kubeSystemUID*/"", ctxConfig)`

bowei:
Can we make sure that all major actions (deleting, adding finalizer) will result in a log line at V(2)?

bowei:
Looks pretty good, almost there

pkg/controller/controller.go
skmatti: We wanted to have a single function to handle all GC paths. But, I guess this makes it very hard to understand. Modified this back to the original workflow.

pkg/controller/controller.go
skmatti: Add events for GC errors

pkg/controller/controller.go
skmatti: Do we need to error out if finalizer is not enabled. Moved this check outside so this function does what the name says.

pkg/firewalls/controller_test.go
skmatti: Not sure if this needed for tests. gofmt changes this to,
`"" /*kubeSystemUID*/`
Added this anyway.

pkg/controller/controller.go

pkg/controller/controller.go
skmatti: Added a TODO

pkg/controller/controller.go

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+func (vn *V2IngressFrontendNamer) IsLegacySSLCert(certName string) bool {
+	return false
+}
+
+// LbName returns loadbalancer name.
+func (vn *V2IngressFrontendNamer) LbName() string {
+	return vn.lbName
+}
+
+// setLbName sets loadbalancer name.
```
bowei: I would rather call this `init()`, but usually this code is just part of the constructor, e.g. newV2IngressFrontendNamer()
You might want to push the code there and get rid of this method.

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+	prefix := fmt.Sprintf("%s%s-%s-%s-%s", vn.prefix, schemaVersionV2, sslCertPrefixV2, vn.clusterUID, vn.lbNameToHash())
+	return strings.HasPrefix(certName, prefix)
+}
+
+// IsLegacySSLCert always return false as v2 naming scheme does not use legacy certs.
+func (vn *V2IngressFrontendNamer) IsLegacySSLCert(certName string) bool {
+	return false
+}
+
+// LbName returns loadbalancer name.
```
bowei: You should document what LbName is used for. Is it the name of a GCP resource or something else?

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+func (vn *V2IngressFrontendNamer) setLbName() {
+	truncFields := TrimFieldsEvenly(maximumAllowedCombinedLength, vn.ing.Namespace, vn.ing.Name)
+	truncNamespace := truncFields[0]
+	truncName := truncFields[1]
+	suffix := vn.suffix(vn.clusterUID, vn.ing.Namespace, vn.ing.Name)
+	vn.lbName = fmt.Sprintf("%s-%s-%s-%s", vn.clusterUID, truncNamespace, truncName, suffix)
+}
+
+// suffix returns hash string of length 8 of a concatenated string generated from
+// uid, namespace and name. These fields define an ingress/ load-balancer uniquely.
```
bowei: which is uid coming from? You should say it's from the k8s resource

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
 	case V1NamingScheme:
 		return newV1IngressFrontendNamer(ing, rn.namer)
 	case V2NamingScheme:
-		// TODO(smatti): return V2 Namer.
-		return nil
+		return newV2IngressFrontendNamer(ing, rn.clusterUID, rn.namer.prefix)
 	default:
-		return newV1IngressFrontendNamer(ing, rn.namer)
+		klog.Errorf("Unexpected frontend naming scheme %v", namingScheme)
+		return nil
```
bowei: we should be careful returning nil.
I would prefer logging an error but returning V1 namer.

pkg/utils/namer/frontendnamer.go
```diff
@@ -79,26 +108,136 @@ func (ln *V1IngressFrontendNamer) LbName() string {
...
+	switch namingScheme {
 	case V1NamingScheme:
 		return newV1IngressFrontendNamer(ing, rn.namer)
 	case V2NamingScheme:
-		// TODO(smatti): return V2 Namer.
-		return nil
+		return newV2IngressFrontendNamer(ing, rn.clusterUID, rn.namer.prefix)
 	default:
-		return newV1IngressFrontendNamer(ing, rn.namer)
+		klog.Errorf("Unexpected frontend naming scheme %v", namingScheme)
```
bowei: %v => %s

pkg/utils/namer/utils.go
```diff
@@ -56,14 +58,27 @@ func TrimFieldsEvenly(max int, fields ...string) []string {
...
+
+// FinalizerForNamingScheme returns finalizer corresponding to given frontend naming scheme.
+func FinalizerForNamingScheme(scheme Scheme) (string, error) {
+	switch scheme {
+	case V1NamingScheme:
+		return common.FinalizerKey, nil
+	case V2NamingScheme:
+		return common.FinalizerKeyV2, nil
+	default:
+		return "", fmt.Errorf("unexpected naming scheme: %v", scheme)
```
bowei: %s

pkg/controller/controller.go
```diff
@@ -453,29 +454,45 @@ func (lbc *LoadBalancerController) SyncLoadBalancer(state interface{}) error {
...
+		if err := common.EnsureDeleteFinalizer(ing, ingClient, common.FinalizerKey); err != nil {
+			klog.Errorf("Failed to ensure delete finalizer %s for ingress %s: %v", common.FinalizerKey, common.NamespacedName(ing), err)
 			return err
 		}
 	}
 	return nil
 }
 
+// EnsureDeleteV2Finalizer implements Controller.
+func (lbc *LoadBalancerController) EnsureDeleteV2Finalizer(ing *v1beta1.Ingress) error {
```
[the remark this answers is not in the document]
skmatti: Should return error. Fixed

pkg/controller/controller.go
```diff
@@ -649,3 +670,91 @@ func (lbc *LoadBalancerController) ToSvcPorts(ings []*v1beta1.Ingress) []utils.S
...
+		return err
+	}
+	finalizerKey, err := namer.FinalizerForNamingScheme(namingScheme)
+	if err != nil {
+		return err
+	}
+	if err := common.EnsureFinalizer(ing, ingClient, finalizerKey); err != nil {
+		klog.Errorf("Failed to ensure finalizer %s for ingress %s: %v", finalizerKey, ingKey, err)
+		return err
+	}
```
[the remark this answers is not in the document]
skmatti: Logged in EnsureFinalizer.

pkg/utils/utils.go
```diff
@@ -70,8 +70,21 @@ const (
 	// ToBeDeletedTaint is the taint that the autoscaler adds when a node is scheduled to be deleted
 	// https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-0.5.2/cluster-autoscaler/utils/deletetaint/delete.go#L33
 	ToBeDeletedTaint = "ToBeDeletedByClusterAutoscaler"
+
+	// NoCleanUpNeeded, CleanupV1FrontendResources and CleanupV2FrontendResources
+	// are used to specify the frontend GC algorithm.
```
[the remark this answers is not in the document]
skmatti: Yes, needs to used in both controller and sync packages.

pkg/utils/namer/frontendnamer.go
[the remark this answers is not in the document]
skmatti: Added at the top of newV2IngressFrontendNamer function.

pkg/utils/namer/frontendnamer.go
skmatti: This is used for generating GCP resource names, added a comment.

pkg/controller/translator/translator_test.go
```diff
@@ -65,7 +65,7 @@ func fakeTranslator() *Translator {
 		HealthCheckPath:               "/",
 		DefaultBackendHealthCheckPath: "/healthz",
 	}
-	ctx := context.NewControllerContext(nil, client, backendConfigClient, nil, nil, defaultNamer, ctxConfig)
+	ctx := context.NewControllerContext(nil, client, backendConfigClient, nil, nil, defaultNamer, "" /*kubeSystemUID*/, ctxConfig)
```
bowei: the style is to put the comment in front of the value.
`/*kubeSystemUID*/""`

PR #893, opened 2019-10-15
Add -network flag to e2e test framework for ILB subnet creation

spencerhance:
We need this to create the ILB subnet in the right network

/assign @MrHohn

### DIFF
--- a/cmd/e2e-test/main_test.go
+++ b/cmd/e2e-test/main_test.go
@@ -43,6 +43,7 @@ var (
 		kubeconfig          string
 		project             string
 		region              string
+		network             string
 		seed                int64
 		destroySandboxes    bool
 		handleSIGINT        bool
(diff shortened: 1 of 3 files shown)


cmd/e2e-test/main_test.go
```diff
@@ -43,6 +43,7 @@ var (
 		kubeconfig          string
 		project             string
 		region              string
+		network							string
```
MrHohn: gofmt?

pkg/e2e/fixtures.go
```diff
@@ -285,7 +285,7 @@ func CreateILBSubnet(s *Sandbox) error {
 
 // trySubnetCreate is a helper for CreateILBSubnet
 func trySubnetCreate(s *Sandbox, name, ipCidrRange string) error {
```
MrHohn: Can we make sure to cleanup the created subnet after the test (e.g. `TestILB`)? Or eventually we will be failing on deleting the network because this subnet is still there.

spencerhance:
@MrHohn Updated to make the subnet creation optional

Issue #764, opened 2019-06-01
Cannot Disable HTTP when using managed-certificates. HTTPS + HTTP are both provisioned


jakebolam:
Cannot disable HTTP when using managed-certificates

When specifying the annotation `kubernetes.io/ingress.allow-http: "false"` the ingress still accepts HTTP traffic. 

### Versions
Kubernetes master/node-kube version: `1.12.7-gke.17` (Note: must be equal to or greater than this version, managed certificates had issues prior to this version see https://github.com/GoogleCloudPlatform/gke-managed-certs/issues/2)

### Additional Details:
I also tried (https://github.com/kubernetes/ingress-gce/issues/738#issuecomment-486738549):
1. Creating an ingress without `kubernetes.io/ingress.allow-http: "false"`.
2. Verified the ingress was serving HTTPS
3. Then setting `allow-http to "false"` 

### Configuration
```
apiVersion: networking.gke.io/v1beta1
kind: ManagedCertificate
metadata:
  name: api-certifcate
spec:
  domains:
    - example.com
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress
  annotations:
    kubernetes.io/ingress.allow-http: "false"
    kubernetes.io/ingress.global-static-ip-name: "web-static-ip"
    networking.gke.io/managed-certificates: api-certifcate
spec:
  rules:
    - http:
        paths:
          - path: /my-service/*
            backend:
              serviceName: my-service
              servicePort: 8080
```

jakebolam:
This was previously reported, but the issue was masked by issues in managed-certificates up until recent versions. See https://github.com/kubernetes/ingress-gce/issues/738

jakebolam:
Debugging config from provisioned ingress:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    ingress.gcp.kubernetes.io/pre-shared-cert: xxxxxx, xxxxx, xxxx
    ingress.kubernetes.io/backends: '{"xxxx":"HEALTHY","xxxx":"HEALTHY"}'
    ingress.kubernetes.io/forwarding-rule: xxxx
    ingress.kubernetes.io/https-forwarding-rule: xxxxx
    ingress.kubernetes.io/https-target-proxy: xxxxx
    ingress.kubernetes.io/ssl-cert: xxxxxxxx, xxxxxx, xxxx
    ingress.kubernetes.io/target-proxy: xxxxx
[...]

Issue #465, opened 2018-09-04
Disabling TLS on an active ingress needs to delete resources(target proxy, certs)


prameshj:
Currently these will not be garbage collected. Unused certs will be deleted only when there is a tls config update.

PR #894, opened 2019-10-16
Bugfix: Delete unused GCE load-balancer resources on ingress spec change

skmatti:
Fixes: https://github.com/kubernetes/ingress-gce/issues/32, https://github.com/kubernetes/ingress-gce/issues/465, https://github.com/kubernetes/ingress-gce/issues/764

/assign @bowei @freehan 

The Ensure workflow of Loadbalancer pool is modified in order to delete unused GCE resources.
This looks at user specified ingress options/annotations to determine if we need to delete specific frontend resources. 

- AllowHTTP is set to false:
  if existing ingress annotations has http resources, then those will be deleted.
- SSL is not configured:
  if existing ingress annotations has https resources, then those will be deleted.
- If User specifies an IP:
  We try to delete Ingress managed static IP if exists. This is performed only when both http and https are enabled.

Note that the existing behavior of load-balancer pool does not update frontend resource annotations for the above mentioned cases. This modifies this behavior to update these annotations which will be used for determining whether we need to delete frontend resources.

### DIFF
--- a/pkg/annotations/ingress.go
+++ b/pkg/annotations/ingress.go
@@ -42,7 +42,7 @@ const (
 	StaticIPNameKey = "kubernetes.io/ingress.global-static-ip-name"
 
 	// PreSharedCertKey represents the specific pre-shared SSL
-	// certicate for the Ingress controller to use. The controller *does not*
+	// certificate for the Ingress controller to use. The controller *does not*
 	// manage this certificate, it is the users responsibility to create/delete it.
 	// In GCP, the Ingress controller assigns the SSL certificate with this name
 	// to the target proxies of the Ingress.
@@ -77,6 +77,25 @@ const (
 	// - annotations:
 	//     networking.gke.io/v1beta1.FrontendConfig: 'my-frontendconfig'
 	FrontendConfigKey = "networking.gke.io/v1beta1.FrontendConfig"
+
+	// UrlMapKey is the annotation key used by controller to record GCP URL map.
+	UrlMapKey = StatusPrefix + "/url-map"
+	// HttpForwardingRuleKey is the annotation key used by controller to record
+	// GCP http forwarding rule.
+	HttpForwardingRuleKey = StatusPrefix + "/forwarding-rule"
+	// HttpsForwardingRuleKey is the annotation key used by controller to record
+	// GCP https forwarding rule.
+	HttpsForwardingRuleKey = StatusPrefix + "/https-forwarding-rule"
+	// TargetHttpProxyKey is the annotation key used by controller to record
+	// GCP target http proxy.
+	TargetHttpProxyKey = StatusPrefix + "/target-proxy"
+	// TargetHttpsProxyKey is the annotation key used by controller to record
+	// GCP target https proxy.
+	TargetHttpsProxyKey = StatusPrefix + "/https-target-proxy"
+	// SSLCertKey is the annotation key used by controller to record GCP ssl cert.
+	SSLCertKey = StatusPrefix + "/ssl-cert"
+	// StaticIPKey is the annotation key used by controller to record GCP static ip.
+	StaticIPKey = StatusPrefix + "/static-ip"
 )
 
 // Ingress represents ingress annotations.
--- a/pkg/loadbalancers/addresses.go
+++ b/pkg/loadbalancers/addresses.go
@@ -21,6 +21,8 @@ import (
 	"net/http"
 
 	"google.golang.org/api/compute/v1"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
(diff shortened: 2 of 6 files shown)


pkg/loadbalancers/l7.go
```diff
@@ -268,57 +283,156 @@ func (l *L7) Cleanup(versions *features.ResourceVersions) error {
...
+		existing[annotations.HttpsForwardingRuleKey] = l.fws.Name
+	} else {
+		delete(existing, annotations.HttpsForwardingRuleKey)
+	}
+	if l.tps != nil {
+		existing[annotations.TargetHttpsProxyKey] = l.tps.Name
+	} else {
+		delete(existing, annotations.TargetHttpsProxyKey)
+	}
+	if l.ip != nil {
```
freehan: why not delete it?

skmatti: The IP is retained for use after it's promoted to static. This will be deleted when load-balancer is deleted or when use specifies a different IP. Added a comment to describe that.

pkg/annotations/ingress.go
```diff
@@ -77,6 +77,25 @@ const (
 	// - annotations:
 	//     networking.gke.io/v1beta1.FrontendConfig: 'my-frontendconfig'
 	FrontendConfigKey = "networking.gke.io/v1beta1.FrontendConfig"
+
+	// UrlMapKey is the annotation key used by controller to record GCP URL map.
+	UrlMapKey = StatusPrefix + "/url-map"
```
rramkumar1: does StatusPrefix need to be exported anymore?

pkg/loadbalancers/l7.go
```diff
@@ -216,6 +226,32 @@ func (l *L7) edgeHopHttps() error {
 	return l.checkHttpsForwardingRule()
 }
 
+// needToDeleteFrontendResources returns true if gce loadbalancer resources needs to deleted for given protocol.
+func needToDeleteFrontendResources(ing v1beta1.Ingress, protocol namer.NamerProtocol) bool {
```
rramkumar1: Personal preference but I would call this func requireDeleteFrontend()

pkg/loadbalancers/l7.go
```diff
@@ -216,6 +226,32 @@ func (l *L7) edgeHopHttps() error {
...
+			annotations.HttpsForwardingRuleKey,
+			annotations.TargetHttpsProxyKey,
+		}...)
+	case namer.HTTPProtocol:
+		keys = append(keys, []string{
+			annotations.HttpForwardingRuleKey,
+			annotations.TargetHttpProxyKey,
+		}...)
+	default:
+		klog.Fatalf("Invalid frontend resource protocol %v", protocol)
```
rramkumar1: It would technically be a programming error to ever get to this point so ideally unit tests should catch this but if not, do we want the behavior to be klog.Fatalf? I wonder if we should log this but not do anything since I'm not sure we want the controller to crash because of this?. Worth thinking about alternatives like returning an error.

pkg/loadbalancers/addresses.go
```diff
@@ -31,26 +32,31 @@ func (l *L7) checkStaticIP() (err error) {
...
 	// Don't manage staticIPs if the user has specified an IP.
 	if address, manageStaticIP := l.getEffectiveIP(); !manageStaticIP {
 		klog.V(3).Infof("Not managing user specified static IP %v", address)
+		// Delete ingress controller managed static ip if exists.
+		if ip, ok := l.ingress.Annotations[annotations.StaticIPKey]; ok && ip == managedStaticIPName {
+			return l.deleteStaticIP()
+		}
 		return nil
 	}
-	staticIPName := l.namer.ForwardingRule(namer.HTTPProtocol)
```
rramkumar1: What was the motivation for changing the name of this var?

pkg/loadbalancers/l7.go
```diff
@@ -227,39 +263,18 @@ func (l *L7) GetIP() string {
...
-	if err := utils.IgnoreHTTPNotFound(composite.DeleteForwardingRule(l.cloud, key, versions.ForwardingRule)); err != nil {
-		return err
-	}
-
-	fwsName := l.namer.ForwardingRule(namer.HTTPSProtocol)
-	klog.V(2).Infof("Deleting global forwarding rule %v", fwsName)
-	if key, err = l.CreateKey(fwsName); err != nil {
+// deleteHttp deletes http forwarding rule and target http proxy.
+func (l *L7) deleteHttp(versions *features.ResourceVersions) error {
+	frName := l.namer.ForwardingRule(namer.HTTPProtocol)
```
rramkumar1: The first couple lines of this func and deleteHttps are the same. Maybe consider wrapping that logic in a func with unit tests?

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -1476,3 +1481,135 @@ func TestSecretBasedToPreSharedCertUpdateWithErrors(t *testing.T) {
...
+
+	gceUrlMap := utils.NewGCEURLMap()
+	gceUrlMap.DefaultBackend = &utils.ServicePort{NodePort: 31234, BackendNamer: j.namer}
+	gceUrlMap.PutPathRulesForHost("bar.example.com", []utils.PathRule{{Path: "/bar", Backend: utils.ServicePort{NodePort: 30000, BackendNamer: j.namer}}})
+	ing := newIngress()
+	feNamer := namer_util.NewFrontendNamerFactory(j.namer).Namer(ing)
+	versions := features.GAResourceVersions
+	certName1 := feNamer.SSLCertName(GetCertHash("cert1"))
+
+	testCases := []struct {
```
rramkumar1: I would add a description to the test case struct itself. That is the typical way it is done.

freehan:
Discussed offline, put it here for reference:

The code added here is quite embedded into the workflow. My main concern was that there is no easy way to flag gate this behavior. I would recommend gathering all the logic and move it into a `trim` function and call it after `edgeHop` with flag gating.

pkg/annotations/ingress.go
skmatti: There are couple of other places where this is referenced. https://github.com/kubernetes/ingress-gce/blob/ba81191da5dab2c8787f265f0988a4d1726b1e4e/pkg/loadbalancers/l7.go#L378

pkg/loadbalancers/l7.go
skmatti: Makes sense, modified this to surface an error instead.

pkg/loadbalancers/addresses.go
skmatti: I had to move this line to the top of the loop. At that point, we are still nor sure if this is the static IP name. getEffectiveIP may return a different name for static IP.

pkg/loadbalancers/l7.go
skmatti: Done

pkg/loadbalancers/loadbalancers_test.go
skmatti: Done

rramkumar1:
> Discussed offline, put it here for reference:
> 
> The code added here is quite embedded into the workflow. My main concern was that there is no easy way to flag gate this behavior. I would recommend gathering all the logic and move it into a `trim` function and call it after `edgeHop` with flag gating.

+1

skmatti:
> > Discussed offline, put it here for reference:
> > The code added here is quite embedded into the workflow. My main concern was that there is no easy way to flag gate this behavior. I would recommend gathering all the logic and move it into a `trim` function and call it after `edgeHop` with flag gating.
> 
> +1

Done

pkg/loadbalancers/l7.go
```diff
@@ -171,6 +171,11 @@ func (l *L7) edgeHop() error {
 		if err := l.edgeHopHttp(); err != nil {
 			return err
 		}
+	} else if flags.F.EnableDeleteUnusedFrontends && requireDeleteFrontend(l.ingress, namer.HTTPProtocol) {
+		if err := l.deleteHttp(features.VersionsFromIngress(&l.ingress)); err != nil {
+			return err
+		}
+		klog.V(3).Infof("l.deleteHttp(_) ok for load-balancer %s", l)
```
rramkumar1: Change this to V2 and just say "Successfully deleted unused HTTP frontend resources". Same for other one too.

pkg/loadbalancers/l7.go
```diff
@@ -227,98 +263,189 @@ func (l *L7) GetIP() string {
...
+		klog.V(2).Infof("Deleting sslcert %s", cert.Name)
+		key, err := l.CreateKey(cert.Name)
+		if err != nil {
+			return err
 		}
-		l.sslCerts = nil
-		if certErr != nil {
-			return certErr
+		if err := utils.IgnoreHTTPNotFound(composite.DeleteSslCertificate(l.cloud, key, versions.SslCertificate)); err != nil {
+			klog.Errorf("Old cert delete failed - %v", err)
```
rramkumar1: Is it necessary to log here if the error is returned? Is the error logged higher up in the stack?

skmatti: This is not logged higher up in the stack. Also, this was just refactored from the existing workflow. May be we want to keep this?

rramkumar1:
Looks good to me. Will leave final approval to @freehan

PR #898, opened 2019-10-18
Fix service name for ilb e2e tests

spencerhance:
Remove extra dash that was added accidentally and creates an invalid service name :upside_down_face: 

/assign @MrHohn


spencerhance:
closing in favor of #899 since that also fixes the same issue

PR #905, opened 2019-10-21
Add e2e tests for frontend resource leak fix

skmatti:
This adds tests for https://github.com/kubernetes/ingress-gce/pull/894. This test asserts that frontend resources existence is in agreement with user specified ingress configuration.

### DIFF
--- a/cmd/e2e-test/basic_test.go
+++ b/cmd/e2e-test/basic_test.go
@@ -239,3 +239,92 @@ func whiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, region str
 	}
 	return gclb
 }
+
+// TestFrontendResourceDeletion asserts that unused GCP frontend resources are
+// deleted. This also tests that necessary GCP frontend resources exist.
+func TestFrontendResourceDeletion(t *testing.T) {
+	t.Parallel()
+	port80 := intstr.FromInt(80)
+	svcName := "service-1"
+	host := "foo.com"
+
+	// All the test cases create an ingress with a HTTP + HTTPS load-balancer
+	// at the beginning of the test and turnoff these based on the testcase params.
+	for _, tc := range []struct {
+		desc         string
+		disableHTTP  bool
+		disableHTTPS bool
+	}{
+		// Note that disabling both HTTP & HTTPS would result in a sync error, so excluded.
+		{"http only", false, true},
+		{"https only", true, false},
+	} {
+		tc := tc
+		Framework.RunWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			t.Parallel()
+			ctx := context.Background()
+
+			_, err := e2e.CreateEchoService(s, svcName, nil)
+			if err != nil {
+				t.Fatalf("CreateEchoService(_, %q, nil): %v, want nil", svcName, err)
+			}
+			t.Logf("Echo service created (%s/%s)", s.Namespace, svcName)
+
+			// Create SSL certificate.
+			certName := fmt.Sprintf("cert-1")
+			cert, err := e2e.NewCert(certName, host, e2e.K8sCert, false)
+			if err != nil {
+				t.Fatalf("e2e.NewCert(%q, %q, _, %t) = %v, want nil", certName, host, false, err)
+			}
+			if err := cert.Create(s); err != nil {
+				t.Fatalf("cert.Create(_) = %v, want nil, error creating cert %s", err, cert.Name)
+			}
+			t.Logf("Cert created %s", certName)
+			defer cert.Delete(s)
+
+			ing := fuzz.NewIngressBuilder(s.Namespace, "ing1", "").
+				AddPath(host, "/", svcName, port80).AddTLS([]string{}, cert.Name).Build()
+
+			crud := e2e.IngressCRUD{C: Framework.Clientset}
+			if _, err := crud.Create(ing); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/e2e-test/basic_test.go
```diff
@@ -239,3 +239,108 @@ func whiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, region str
...
+	host := "foo.com"
+
+	for _, tc := range []struct {
+		desc         string
+		disableHTTP  bool
+		disableHTTPS bool
+	}{
+		{"both enabled", false, false},
+		{"http only", false, true},
+		{"http only", true, false},
```
rramkumar1: https only?

cmd/e2e-test/basic_test.go
```diff
@@ -239,3 +239,108 @@ func whiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, region str
...
+
+// TestFrontendResourceDeletion asserts that unused GCP frontend resources are
+// deleted. This also tests that necessary GCP frontend resources exist.
+func TestFrontendResourceDeletion(t *testing.T) {
+	t.Parallel()
+	port80 := intstr.FromInt(80)
+	svcName := "service-1"
+	host := "foo.com"
+
+	for _, tc := range []struct {
```
rramkumar1: I think it would be useful to add a comment somewhere that all test cases start with a HTTP + HTTPS load balancer.

cmd/e2e-test/basic_test.go
```diff
@@ -239,3 +239,108 @@ func whiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, region str
...
+				CheckHttpFrontendResources:  tc.disableHTTP,
+				CheckHttpsFrontendResources: tc.disableHTTPS,
+			}
+			// Wait for unused frontend resources to be deleted.
+			if err := e2e.WaitForFrontendResourceDeletion(ctx, Framework.Cloud, gclb, deleteOptions); err != nil {
+				t.Errorf("e2e.WaitForIngressDeletion(..., %q, _) = %v, want nil", ing.Name, err)
+			}
+
+			// Special case in which we whitebox test just the number of forwarding rules.
+			if tc.disableHTTP && tc.disableHTTPS {
```
rramkumar1: This is sort of out of scope of this PR but I'm worried this is not the right behavior. If someone tries to turn off both HTTP & HTTPS, I feel like we should return an error to them rather than garbage collecting both of their forwarding rules.

skmatti: This may require changes to existing behavior, currently we emit a warning:
https://github.com/kubernetes/ingress-gce/blob/a2e96e0dbedf9c5929c0202e1c8a7fd4ef92deb4/pkg/loadbalancers/l7.go#L200-L202
The fix would return a warning too when user disables both HTTP and HTTPS.

We either switch to a workflow where we return an error on both creation and update or we keep the existing workflow. 

Is it worth changing the existing workflow to return an error if both HTTP & HTTPS are disabled?.

rramkumar1: Let me rephrase what I think the behavior should be. In my mind, disabling both HTTP & HTTPS is a mistake that the user makes since its not valid config. If a user truly wanted both frontends to be deleted when this happens, they should just delete the Ingress outright.

Disabling both protocols ideally should be an error than is propagated all the way up. I don't see any issues with this but please keep me honest.

skmatti: Raised a PR to fix this: https://github.com/kubernetes/ingress-gce/pull/950

The only concern is that we re-queue the ingress repeatedly until user fixes the configuration.

rramkumar1: I think that's fine. This is no different than another situation w/ invalid config such as referencing a service that does not exist.

skmatti: The fix is merged now. Updated this test to exclude the test case.

pkg/fuzz/gcp.go
```diff
@@ -245,7 +256,89 @@ func (g *GCLB) CheckResourceDeletion(ctx context.Context, c cloud.Cloud, options
...
+// getForwardingRuleProtocol returns the protocol for given forwarding rule.
+func getForwardingRuleProtocol(forwardingRule *compute.ForwardingRule) (Protocol, error) {
+	resID, err := cloud.ParseResourceURL(forwardingRule.Target)
+	if err != nil {
+		return "", fmt.Errorf("error parsing Target (%q): %v", forwardingRule.Target, err)
+	}
+	switch resID.Resource {
+	case "targetHttpProxies":
+		return HttpProtocol, nil
+	case "targetHttpsProxies":
```
MrHohn: nit: should this be a constant somewhere?

cmd/e2e-test/basic_test.go
```diff
@@ -239,3 +239,93 @@ func whiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, region str
...
+
+	// All the test cases create an ingress with a HTTP + HTTPS load-balancer
+	// at the beginning of the test and turnoff these based on the testcase params.
+	for _, tc := range []struct {
+		desc         string
+		disableHTTP  bool
+		disableHTTPS bool
+	}{
+		// Note that disabling both HTTP & HTTPS would result in a sync error, so excluded.
+		{"both enabled", false, false},
```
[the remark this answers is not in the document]
skmatti: Yes, removed the test case.

PR #906, opened 2019-10-22
Fix ILB forwarding rule bug

spencerhance:
Adds cluster subnet to l7 ilb forwarding rule when the cluster network is in
custom mode

/assign @MrHohn

### DIFF
--- a/pkg/loadbalancers/features/l7ilb.go
+++ b/pkg/loadbalancers/features/l7ilb.go
@@ -65,6 +65,23 @@ func isSameNetwork(l, r string) (bool, error) {
 	return lID.Equal(rID), nil
 }
 
+// IsCustomModeNetwork is a helper for determining if a network is a custom mode network
+// (as opposed to a auto mode network).  This is used for l7-ilb since custom mode networks
+// require an additional field on the forwarding rule
+func IsCustomModeNetwork(c *gce.Cloud, networkURL string) (bool, error) {
+	netID, err := cloud.ParseResourceURL(networkURL)
+	if err != nil {
+		return false, err
+	}
+
+	network, err := c.Compute().Networks().Get(context.Background(), netID.Key)
+	if err != nil {
+		return false, err
+	}
+
+	return !network.AutoCreateSubnetworks, nil
+}
+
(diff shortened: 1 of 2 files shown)


spencerhance:
@MrHohn Thanks for finding the bug 😄

pkg/loadbalancers/features/l7ilb.go
```diff
@@ -65,6 +65,23 @@ func isSameNetwork(l, r string) (bool, error) {
 	return lID.Equal(rID), nil
 }
 
+// IsCustomModeNetwork is a helper for determining if a network is a custom mode network
+// (as opposed to a auto mode network).  This is used for l7-ilb since custom mode networks
+// require an additional field on the forwarding rule
+func IsCustomModeNetwork(c *gce.Cloud, networkURL string) (bool, error) {
```
bowei: we might want to cache this at some point to avoid doing too many GCE calls

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -91,9 +93,17 @@ func (l *L7) checkForwardingRule(name, proxyLink, ip, portRange string) (fw *com
...
 
 		// Update rule for L7-ILB
-		if utils.IsGCEL7ILBIngress(l.runtimeInfo.Ingress) {
+		if flags.F.EnableL7Ilb && utils.IsGCEL7ILBIngress(l.runtimeInfo.Ingress) {
 			rule.LoadBalancingScheme = "INTERNAL_MANAGED"
+
+			// Custom mode networks require adding the cluster subnetwork
+			if isCustomMode, err := features.IsCustomModeNetwork(l.cloud, l.cloud.NetworkURL()); err != nil {
+				return nil, err
+			} else if isCustomMode {
```
bowei: if you are returning, you don't need the else

pkg/loadbalancers/features/l7ilb.go
spencerhance: Good call.  Originally this was thought to be the proxy subnet which would change, but now that it's the cluster subnet, we do only need to do it once.

PR #910, opened 2019-10-23
use configured service port for default backend

freehan:
cc @spencerhance


freehan:
close in favor of https://github.com/kubernetes/ingress-gce/pull/917/files#diff-e22c7da658ce973a44ca4b6cb0033873R442

Issue #832, opened 2019-08-21
GCE Ingress creates a Network Endpoint Group with 0 configured


giladsh1:
GCE Ingress creates a network endpoint group with 0 configured, when setting targetPort name instead of the actual number.
This config will not work and will essentially create a useless network endpoint group, which doesn't recognise the pods.
```
apiVersion: v1
kind: Service
metadata:
  name: mgmt
  namespace: riscale-test
  annotations:
    cloud.google.com/neg: '{"ingress": true}'
    beta.cloud.google.com/backend-config: '{"ports": {"8080":"mgmt-service-backend"}}'
spec:
  selector:
    app: mgmt
  type: NodePort
  ports:
    - port: 8080
      targetPort: mgmt-port
      protocol: TCP
```

However, when changing the `targetPort` to 8080, the network endpoint group will recognise the running pods.
This was a tough bug to catch :-(

giladsh1:
@rramkumar1 adding the deployment config as requested - 
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mgmt
  namespace: riscale-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mgmt
  template:
    metadata:
      labels:
        app: mgmt
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: cloud.google.com/gke-nodepool
                    operator: In
                    values:
                      - default-pool
      restartPolicy: Always
      containers:
        - name: mgmt
          image: eu.gcr.io/riscale/mgmt
          imagePullPolicy: IfNotPresent
          resources:
            requests:
              cpu: 50m
              memory: 10Mi
            limits:
              cpu: 150m
              memory: 50Mi
          ports:
            - name: mgmt-port
              containerPort: 8080
          readinessProbe:
            httpGet:
              path: /m/health
              port: mgmt-port
            initialDelaySeconds: 2
            periodSeconds: 15
            successThreshold: 2
            failureThreshold: 4
          livenessProbe:
            httpGet:
              path: /m/health
              port: mgmt-port
            periodSeconds: 15
            failureThreshold: 4
          envFrom:
            - configMapRef:
                name: common-config
            - configMapRef:
                name: service-discovery
            - configMapRef:
                name: postgres-config
            - configMapRef:
                name: mongo-config
            - secretRef:
                name: rethinkdb-secrets
[...]

PR #917, opened 2019-10-26
NEG Namedport Fix

freehan:
Fixes: #832 

- Introduce Port Tuple for service port
- Refactor NegSyncerKey to include portTuple
- Use service port name to retrieve wanted endpoints instead of target port. 
- adjust all existing unit test to work with it

### DIFF
--- a/pkg/neg/types/types.go
+++ b/pkg/neg/types/types.go
@@ -36,14 +36,53 @@ const (
 	NonGCPPrivateEndpointType = NetworkEndpointType("NON_GCP_PRIVATE_IP_PORT")
 )
 
-// SvcPortMap is a map of ServicePort:TargetPort
-type SvcPortMap map[int32]string
+// SvcPortTuple is the tuple representing one service port
+type SvcPortTuple struct {
+	// Port is the service port number
+	Port int32
+	// Name is the service port name
+	Name string
+	// TargetPort is the service target port.
+	// This can be a port number or named port
+	TargetPort string
+}
+
+// String returns the string representation of SvcPortTuple
+func (t SvcPortTuple) String() string {
+	return fmt.Sprintf("%s/%v-%s", t.Name, t.Port, t.TargetPort)
+}
+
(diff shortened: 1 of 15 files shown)


pkg/neg/types/types.go
```diff
@@ -36,14 +36,53 @@ const (
...
+	}
+	return SvcPortTuple{}, false
+}
 
 // PortInfo contains information associated with service port
 type PortInfo struct {
-	// TargetPort is the target port of the service port
-	// This can be port number or a named port
-	TargetPort string
+	// Tuple is port tuple of a service.
```
MrHohn: Tuple -> PortTuple

pkg/neg/types/types.go
```diff
@@ -36,14 +36,53 @@ const (
...
+// Insert inserts a SvcPortTuple into SvcPortTupleSet
+func (set SvcPortTupleSet) Insert(tuples ...SvcPortTuple) {
+	for _, tuple := range tuples {
+		set[tuple] = struct{}{}
+	}
+}
+
+// Get returns the SvcPortTuple with matching svc port if found
+func (set SvcPortTupleSet) Get(svcPort int32) (SvcPortTuple, bool) {
+	for tuple := range set {
```
MrHohn: Would there be any concern for this operation being O(N)?

MrHohn: Assuming SvcPortTupleSet's scope is for all neg enabled services - would there be a possibility for conflicting svcPort?

freehan: One `SvcPortTupleSet` is usually extracted from one service. It was not used to store SvcPortTuple for multiple services. 

So I guess it should not be too much a concern for efficiency. Also, Service Spec sort of guarantee uniqueness.

pkg/neg/types/types.go
freehan: Fixed

freehan:
Added more unit test cases.

/unhold

PR #920, opened 2019-10-29
Add a flag for Non-GCP mode NEG controller.

cxhiano:
With '--enable-non-gcp-mode', NEG controller will create NEGs of type NON_GCP_PRIVATE_IP_PORT instead of GCE_VM_IP_PORT.

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/neg/readiness"
 	negsyncer "k8s.io/ingress-gce/pkg/neg/syncers"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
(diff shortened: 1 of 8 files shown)


pkg/neg/manager.go
```diff
@@ -113,9 +113,15 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 		syncerKey := getSyncerKey(namespace, name, svcPort, portInfo)
 		syncer, ok := manager.syncerMap[syncerKey]
 		if !ok {
+			networkEndpointType := negtypes.VMNetworkEndpointType
```
cxhiano: type negtypes.NEGType string

const (
VMNetworkEndpiontType = NEGType(VM)
)

pkg/neg/syncers/transaction.go
```diff
@@ -35,7 +35,8 @@ import (
 type transactionSyncer struct {
 	// metadata
 	negtypes.NegSyncerKey
-	negName string
+	negName             string
+	networkEndpointType string
```
cxhiano: add comment

pkg/neg/syncers/utils.go
```diff
@@ -286,6 +298,11 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet) (map[negtypes.Netw
 			IpAddress: networkEndpoint.IP,
 			Port:      int64(portNum),
 		}
+
+		if networkEndpointType == negtypes.NonGCPPrivateEndpointType {
```
cxhiano: remove this. no need

pkg/neg/syncers/utils.go
```diff
@@ -113,7 +112,7 @@ func getService(serviceLister cache.Indexer, namespace, name string) *apiv1.Serv
 }
 
 // ensureNetworkEndpointGroup ensures corresponding NEG is configured correctly in the specified zone.
-func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negServicePortName string, cloud negtypes.NetworkEndpointGroupCloud, serviceLister cache.Indexer, recorder record.EventRecorder) error {
+func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negServicePortName string, networkEndpointType negtypes.NetworkEndpointType, cloud negtypes.NetworkEndpointGroupCloud, serviceLister cache.Indexer, recorder record.EventRecorder) error {
```
freehan: Consider 2 things:

1. refactor the code to structure like this:
A. construct expected NEG resource
B. Check if the existing NEG is the same as expected
C. Delete/Create/No-Op 

2. Or add a TODO on top and leave it later.

PR #921, opened 2019-10-30
Add e2e tests for v2 frontend namer 

skmatti:


cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+		})
+	}
+}
+
+// TestUpdateTo1dot8 asserts that v1 finalizer is retained/attached and ingress
+// continues to use v1 naming naming scheme when master is upgraded to a version that use v1.8.
+// Note: The test is named in such a way that it does run as a workflow test or an
+// upgrade test for other ingress versions.
+func TestUpdateTo1dot8(t *testing.T) {
```
rramkumar1: @freehan wrote a generic upgrade test framework. See pkg/e2e/interfaces.go.

Can you use that to implement this test?

pkg/fuzz/default_validator_env.go
```diff
@@ -89,3 +98,8 @@ func (e *DefaultValidatorEnv) Cloud() cloud.Cloud {
 func (e *DefaultValidatorEnv) Namer() *namer.Namer {
 	return e.namer
 }
+
+// DefaultValidatorEnv implements ValidatorEnv.
+func (e *DefaultValidatorEnv) FeNamerFactory() namer.IngressFrontendNamerFactory {
```
rramkumar1: FrontendNamerFactory

pkg/e2e/helpers.go
```diff
@@ -499,3 +502,15 @@ func CheckV1Finalizer(ing *v1beta1.Ingress) error {
 	}
 	return nil
 }
+
+// CheckV2Finalizer asserts that v2 finalizer exists on Ingress.
```
MrHohn: nit: asserts that only v2 finalizer exists?

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/utils/common"
+)
+
+// TestV2FrontendNamer tests ingress basic lifecycle with v2 frontend naming scheme.
+// This also tests the lifecycle an ingress with v1 naming scheme with v2 namer enabled.
```
MrHohn: nit: lifecycle of an ingress

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+		{"both v1 and v2, delete v2 first", []*v1beta1.Ingress{v2Ing, v1Ing}},
+	} {
+		tc := tc
+		desc := fmt.Sprintf("v2 frontend namer %s", tc.desc)
+		Framework.RunWithSandbox(desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			t.Parallel()
+			ctx := context.Background()
+
+			_, err := e2e.CreateEchoService(s, svcName, nil)
+			if err != nil {
```
MrHohn: nit: `if _, err := ...`

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+				if ing, err = e2e.WaitForIngress(s, ing, &e2e.WaitForIngressOptions{ExpectUnreachable: true}); err != nil {
+					t.Fatalf("error waiting for Ingress %s to stabilize: %v", ingKey, err)
+				}
+				if isV1Finalizer {
+					// Assert that v1 finalizer is added.
+					if err := e2e.CheckV1Finalizer(ing); err != nil {
+						t.Fatalf("CheckV1Finalizer(%s) = %v, want nil", ingKey, err)
+					}
+				} else {
+					// Assert that v2 finalizer is added.
```
MrHohn: Should we also assert GCE resource is actually using v2 naming?

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			// Determine if we need to skip backends.
+			if len(updatedIngs) > 1 {
+				deleteOptions.SkipBackends = true
+			}
+			if err := e2e.WaitForIngressDeletion(ctx, gclbs[0], s, updatedIngs[0], deleteOptions); err != nil {
+				t.Errorf("e2e.WaitForIngressDeletion(..., %q, nil) = %v, want nil", tc.ings[0].Name, err)
+			}
+
+			if len(updatedIngs) > 1 {
```
MrHohn: Probably return early here so we don't need to indent below?

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+				// Perform whitebox testing. This also tests naming scheme.
+				gclb, err := e2e.WhiteboxTest(ing, s, Framework.Cloud, "")
+				if err != nil {
+					t.Fatalf("e2e.WhiteboxTest(%s, ...) = %v, want nil", ingKey, err)
+				}
+				gclbs = append(gclbs, gclb)
+				updatedIngs = append(updatedIngs, ing)
+			}
+
+			// Delete the first ingress.
```
MrHohn: What is the purpose of deleting them one by one - testing new behavior? Maybe worth putting in the comment.

cmd/e2e-test/v2frontendnamer_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Build()
+
+	for _, tc := range []struct {
+		desc string
+		ings []*v1beta1.Ingress
+	}{
+		{"v2 only", []*v1beta1.Ingress{v2Ing}},
+		{"v1 only", []*v1beta1.Ingress{v1Ing}},
+		{"both v1 and v2, delete v1 first", []*v1beta1.Ingress{v1Ing, v2Ing}},
+		{"both v1 and v2, delete v2 first", []*v1beta1.Ingress{v2Ing, v1Ing}},
```
MrHohn: The implicit behavior of the first ingress in the slice being deleted first is pretty hidden - can we at least put a comment?

cmd/e2e-test/v2frontendnamer_test.go
skmatti: Whitebox below tests the GCE resource naming scheme too. 
https://github.com/kubernetes/ingress-gce/pull/921/commits/d71e11fd9f6681b59902ff92fa9553f5ca428f6a#diff-678e7bb83aca0d344cce2d8ed31d1fadR194. 

The frontend naming scheme test merged in early: https://github.com/kubernetes/ingress-gce/blob/7008288c8e786bb85145f74ed9105bedea5a8ede/pkg/fuzz/validator.go#L249-L268

cmd/e2e-test/v2frontendnamer_test.go
skmatti: With v2 naming scheme, we have two GC paths for frontends, one each for v1 naming scheme and v2 naming scheme. This is to assert that one path does not affect other one.

cmd/e2e-test/v2frontendnamer_test.go
MrHohn: Great that it is being tested :)

cmd/e2e-test/v2frontendnamer_test.go
MrHohn: Make sense, thanks for the explanation.

PR #928, opened 2019-11-05
add configmap based config

cadmuxe:


pkg/cmconfig/controller_test.go
```diff
@@ -0,0 +1,55 @@
...
+		t.Errorf("GetConfig should return the same config as NewConfig, got: %v, want: %v", cmcController.GetConfig(), NewConfig())
+	}
+
+	if !strings.Contains(logBuf.String(), "ConfigMapConfigController mode disabled") {
+		t.Errorf("The conroller log should show the controller is running with ConfigMapBasedConfig=off mode, logs: %s", logBuf.String())
+	}
+
+}
+
+func TestController(t *testing.T) {
```
cadmuxe: I will add more testcases

cmd/glbc/main.go
```diff
@@ -220,7 +231,8 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	lbc.Run()
 
 	for {
-		klog.Infof("Handled quit, awaiting pod deletion.")
+		klog.Warning("Handled quit, awaiting pod deletion.")
 		time.Sleep(30 * time.Second)
+		return
```
cadmuxe: Should we add IsStopped functions to all the controllers and do a check in here?

cmd/glbc/main.go
```diff
@@ -192,20 +199,24 @@ func makeLeaderElectionConfig(client clientset.Interface, recorder record.EventR
...
 			},
 		},
 	}, nil
 }
 
-func runControllers(ctx *ingctx.ControllerContext) {
+func runControllers(ctx *ingctx.ControllerContext, cmConifgController *cmconfig.ConfigMapConfigController) {
 	stopCh := make(chan struct{})
+	cmConfig := cmConifgController.GetConfig()
+	if cmConifgController.IsConfigMapConfigModeOn() {
```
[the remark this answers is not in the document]
cadmuxe: I added the flag gate to the controller. if it's enabled, it will try to load the configmap and watch it.
if disabled, it will just return the default config structure. 
The idea is,  even this feature is disabled, we want to get the default config. and I don't want to put the get default config out from the controller.

cmd/glbc/main.go
```diff
@@ -139,23 +132,29 @@ func main() {
...
 
 	if !flags.F.LeaderElection.LeaderElect {
 		runControllers(ctx)
 		return
 	}
 
+	leaderElectionCtx, leaderElectionCancel := context.WithCancel(context.Background())
 	electionConfig, err := makeLeaderElectionConfig(leaderElectKubeClient, ctx.Recorder(flags.F.LeaderElection.LockObjectNamespace), func() {
 		runControllers(ctx)
+		klog.Info("Shutting down leader election")
```
freehan: Can you test one thing? Without this change, do this:

1. send SIGTERM to the container
2. then the SIGTERM handler should shutdown all controller go routines.
3. I would assume runControllers will exit.
4. Make the graceful termination period super long, like 10 mins
5. Validate what happens after runControllers exited. 
6. I assume, without this change, it will do nothing and keep on getting the master election lock until graceful termination times out.

cmd/glbc/main.go
```diff
@@ -139,23 +132,29 @@ func main() {
...
 	go app.RunHTTPServer(ctx.HealthCheck)
 
 	if !flags.F.LeaderElection.LeaderElect {
 		runControllers(ctx)
 		return
 	}
 
+	leaderElectionCtx, leaderElectionCancel := context.WithCancel(context.Background())
 	electionConfig, err := makeLeaderElectionConfig(leaderElectKubeClient, ctx.Recorder(flags.F.LeaderElection.LockObjectNamespace), func() {
 		runControllers(ctx)
```
freehan: @bowei suggested exiting the program directly. after runController exits. That is more reliable. 

https://gobyexample.com/exit

cadmuxe: yeah, agree.

cmd/glbc/main.go
cadmuxe: Synced offline, this got tested before. assumption is correct.

pkg/cmconfig/controller.go
```diff
@@ -0,0 +1,95 @@
...
+		configMapNamespace: configMapNamespace,
+		configMapName:      configMapName,
+		currentConfig:      &currentConfig,
+		kubeClient:         kubeClient,
+	}
+	return c
+}
+
+// GetConfig returns the internal Config
+func (c *ConfigMapConfigController) GetConfig() Config {
```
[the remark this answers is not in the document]
cadmuxe: It the controller was created with NewConfigMapConfigController, the c.currentConfig would never be null. Which means, if it's null, we probably want the program fail directly. 

Imaging we change this to (Config, boolean), and in the case that it return (config, false).
What should the caller to doe? 
1. fail the program    [ same as now]
2. make a new Config which use the default value. [ I want to put the make default logic in the NewConfigMapConfigController  function but not anywhere else.  ]

pkg/cmconfig/controller.go
```diff
@@ -0,0 +1,95 @@
...
+	return *c.currentConfig
```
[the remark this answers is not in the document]
cadmuxe: c.currentConfig = nil, means the caller is using a controller that not created with NewConfigMapConfigController. In that case, we want it failed ASAP.

cmd/glbc/main.go
```diff
@@ -192,20 +191,27 @@ func makeLeaderElectionConfig(client clientset.Interface, recorder record.EventR
...
 }
 
 func runControllers(ctx *ingctx.ControllerContext) {
 	stopCh := make(chan struct{})
+	ctx.Init()
 	lbc := controller.NewLoadBalancerController(ctx, stopCh)
 
 	fwc := firewalls.NewFirewallController(ctx, flags.F.NodePortRanges.Values())
 
+	if ctx.EnableASMConfigMapConfig {
```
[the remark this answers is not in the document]
cadmuxe: I moved it to the below of lbc := controller.NewLoadBalancerController(ctx, stopCh)
Because the RegisterInformer need to call lbc.Stop.

pkg/cmconfig/config.go
```diff
@@ -0,0 +1,74 @@
...
+	return Config{ASMServiceNEGSkipNamespaces: []string{"kube-system"}}
+}
+
+// Equals returns true if c equals to other.
+func (c *Config) Equals(other *Config) bool {
+	return reflect.DeepEqual(c, other)
+}
+
+// LoadValue loads configs from a map, it will ignore any unknow/unvalid field.
+func (c *Config) LoadValue(m map[string]string, recordWarning func(msg string)) {
```
freehan: I think this function may be too complex. 

Check out this one, it is very simple and straightforward.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/gce/gce_clusterid.go#L234

docs/deploy/resources/configmap-based-config.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ingress-controller-asm-cm-config
+  namespace: kube-system
+data:
+  EnableASM: "false"
```
freehan: consider a few alternatives:

enable-asm: 
asm-skip-namespaces:

pkg/cmconfig/controller.go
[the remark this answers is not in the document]
freehan: Okay.

pkg/cmconfig/config.go
cadmuxe: agree, the initial idea was keeping this generic. 
Updated.

pkg/cmconfig/controller.go
```diff
@@ -0,0 +1,115 @@
...
+	cm, err := c.kubeClient.CoreV1().ConfigMaps(c.configMapNamespace).Get(c.configMapName, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Infof("ConfigMapConfigController: Not found the configmap based config, using default config: %v", config)
+		} else {
+			klog.Warningf("ConfigMapConfigController failed to load config from api server, using the defualt config. Error: %v", err)
+		}
+	} else {
+		c.currentConfigMapObject = cm
+		config.LoadValue(cm.Data, func(msg string) {
```
freehan: can we do this?

```
if err := config.LoadValue(cm.Data); err != nil {
   c.RecordEvent(".....")
}

```

pkg/cmconfig/config.go
```diff
@@ -0,0 +1,56 @@
...
+type Config struct {
+	EnableASM                   bool
+	ASMServiceNEGSkipNamespaces []string
+}
+
+const (
+	sTrue  = "true"
+	sFalse = "false"
+
+	sEnableASM                   = "enable-asm"
```
freehan: nit: enableASM

pkg/cmconfig/config.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Config holds configmap based configurations.
+type Config struct {
+	EnableASM                   bool
+	ASMServiceNEGSkipNamespaces []string
+}
+
+const (
+	sTrue  = "true"
```
freehan: nit: trueValue

pkg/cmconfig/config.go
```diff
@@ -0,0 +1,56 @@
...
+	for k, v := range m {
+		if k == sEnableASM {
+			if v == sTrue {
+				c.EnableASM = true
+			} else if v == sFalse {
+				c.EnableASM = false
+			} else {
+				msg := fmt.Sprintf("The map provided a unvalid value for field: %s, value: %s, valid values are: %s/%s", k, v, sTrue, sFalse)
+				klog.Errorf(msg)
+				recordWarning(msg)
```
freehan: you can maintain a errList := []error
look at this:
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/readiness/poller.go#L165

PR #933, opened 2019-11-07
Add TrafficManagement Spec to BackendConfig v1beta1

spencerhance:
This will be followed up with individual PRs for the implementation of each feature.

I'm adding this to v1beta1 to avoid having to do backendconfig GA at the same time


bowei:
Can we have a very brief API review for what is being added?

PR #934, opened 2019-11-11
Use Alpha API for hybrid NEGs.

cxhiano:


freehan:
May be just a unit test for the sake of regression prevention.

PR #942, opened 2019-11-14
Disabling ASM mode if the Istio CRD is not available.

cadmuxe:


pkg/flags/flags.go
```diff
@@ -204,7 +204,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableL7Ilb, "enable-l7-ilb", false,
 		`Optional, whether or not to enable L7-ILB.`)
 	flag.BoolVar(&F.EnableASMConfigMapBasedConfig, "enable-asm-config-map-config", false, "Enable ASMConfigMapBasedConfig")
-	flag.StringVar(&F.ASMConfigMapBasedConfigNamespace, "asm-configmap-based-config-namespace", "kube-system,istio-system", "ASM Configmap based config: configmap namespace")
```
cadmuxe: This is a bug, it's the configmap namespace, but not the skip namespace.

pkg/context/context.go
```diff
@@ -135,25 +139,63 @@ func (ctx *ControllerContext) Init() {
...
+		destinationRuleGroup      = "networking.istio.io"
+		destinationRuleAPIVersion = "v1alpha3"
+		destinationRulePlural     = "destinationrules"
+		// This must match the spec fields below, and be in the form: <plural>.<group>
+		destinationRuleCRDName = "destinationrules.networking.istio.io"
+	)
+
+	apiextensionClient, err := apiextensionsclientset.NewForConfig(ctx.KubeConfig)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to validate DestinationRule CRD: failed to create apiextensionClient, disabling ASM Mode, error: %s", err)
```
freehan: I am not sure if this is related to DestinationRule CRD at all. This just means that the clientSet cannot be initialized, right?

pkg/context/context.go
```diff
@@ -135,25 +139,63 @@ func (ctx *ControllerContext) Init() {
...
+	if err != nil {
+		msg := fmt.Sprintf("Failed to create kubernetes dynamic client, disabling ASM Mode, please retry. Error: %v", err)
+		klog.Fatalf(msg)
+		ctx.ASMConfigController.RecordEvent("Warning", "FailedCreateDynamicClient", msg)
+		ctx.ASMConfigController.DisableASMMode()
+		return
+	}
+
+	klog.Info("The DestinationRule group version is v1alpha3 in group networking.istio.io. Need to update as istio API graduates.")
+	destrinationGVR := schema.GroupVersionResource{Group: "networking.istio.io", Version: "v1alpha3", Resource: "destinationrules"}
```
freehan: use constant vars above?

pkg/context/context.go
```diff
@@ -135,25 +139,63 @@ func (ctx *ControllerContext) Init() {
...
+	dynamicClient, err := dynamic.NewForConfig(ctx.KubeConfig)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to create kubernetes dynamic client, disabling ASM Mode, please retry. Error: %v", err)
+		klog.Fatalf(msg)
+		ctx.ASMConfigController.RecordEvent("Warning", "FailedCreateDynamicClient", msg)
+		ctx.ASMConfigController.DisableASMMode()
+		return
+	}
+
+	klog.Info("The DestinationRule group version is v1alpha3 in group networking.istio.io. Need to update as istio API graduates.")
```
freehan: V(2)

use constant vars above?

pkg/cmconfig/config.go
```diff
@@ -24,7 +24,7 @@ const (
 
 // NewConfig returns a Conifg instances with default values.
 func NewConfig() Config {
-	return Config{ASMServiceNEGSkipNamespaces: []string{"kube-system"}}
+	return Config{ASMServiceNEGSkipNamespaces: []string{"kube-system", "istio-system"}}
```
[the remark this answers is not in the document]
cadmuxe: istio-system was in the default skip flag, I just forgot to move it here.

pkg/context/context.go
cadmuxe: yes, but it's a required step to validate the CRD

pkg/cmconfig/config.go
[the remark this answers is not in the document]
cadmuxe: yes, explicitly config is good.

pkg/cmconfig/controller.go
```diff
@@ -58,6 +63,23 @@ func (c *ConfigMapConfigController) GetConfig() Config {
 	return *c.currentConfig
 }
 
+// DisableASMMode disables the ASM Mode by updating the ConfigMap and setting the internal flag.
+func (c *ConfigMapConfigController) DisableASMMode() {
```
freehan: I am thinking if it is possible to not patch the original config map but just record events. Patching configmap requires some RBAC changes I assume?

pkg/context/context.go
freehan: I mean the message should be failed to create extension clientset right? Not validating destinationRule CRD.

PR #948, opened 2019-11-19
Fix naming scheme on creation

skmatti:
This fixes the following two things

- Naming scheme on Ingress Creation
 When controller ensures a finalizer for a new ingress being created, the load-balancer pool requires ingress latest finalizer annotation for creating ingress frontend resources based on corresponding naming scheme. This fix makes sure that finalizer annotation is updated before passing the ingress for sync.
- Kube-system UID needs to populated for all cases
 Imagine a case where an ingress is created using v2 naming scheme and ingress controller is restarted with `--enable-v2-frontend-namer` disabled. We expect the ingress controller handle existing ingresses with v2 namer properly. We need kube-system uid to retrieve/delete load-balancer frontend resources in this case.

/assign @MrHohn

### DIFF
--- a/pkg/utils/common/finalizer.go
+++ b/pkg/utils/common/finalizer.go
@@ -52,18 +52,18 @@ func HasGivenFinalizer(m meta_v1.ObjectMeta, key string) bool {
 }
 
 // EnsureFinalizer ensures that the specified finalizer exists on given Ingress.
-func EnsureFinalizer(ing *v1beta1.Ingress, ingClient client.IngressInterface, finalizerKey string) error {
+func EnsureFinalizer(ing *v1beta1.Ingress, ingClient client.IngressInterface, finalizerKey string) (*v1beta1.Ingress, error) {
+	updated := ing.DeepCopy()
 	if needToAddFinalizer(ing.ObjectMeta, finalizerKey) {
-		updated := ing.DeepCopy()
 		updated.ObjectMeta.Finalizers = append(updated.ObjectMeta.Finalizers, finalizerKey)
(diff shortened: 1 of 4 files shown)


pkg/utils/common/finalizer.go
```diff
@@ -52,18 +52,18 @@ func HasGivenFinalizer(m meta_v1.ObjectMeta, key string) bool {
...
 // EnsureFinalizer ensures that the specified finalizer exists on given Ingress.
-func EnsureFinalizer(ing *v1beta1.Ingress, ingClient client.IngressInterface, finalizerKey string) error {
+func EnsureFinalizer(ing *v1beta1.Ingress, ingClient client.IngressInterface, finalizerKey string) (*v1beta1.Ingress, error) {
+	updated := ing.DeepCopy()
 	if needToAddFinalizer(ing.ObjectMeta, finalizerKey) {
-		updated := ing.DeepCopy()
 		updated.ObjectMeta.Finalizers = append(updated.ObjectMeta.Finalizers, finalizerKey)
 		// TODO(smatti): Make this optimistic concurrency control compliant on write.
 		// Refer: https://github.com/eBay/Kubernetes/blob/master/docs/devel/api-conventions.md#concurrency-control-and-consistency
 		if _, err := ingClient.Update(updated); err != nil {
```
MrHohn: Should we capture the "updated" here for return?
```golang
if updated, err = ingClient.Update(updated); err != nil {
...
```

MrHohn: Resolved offline. Seems safer to update just the finalizer but the others.

PR #951, opened 2019-11-20
Migrate upgrade tests to use the upgrade framework

skmatti:
/assign @rramkumar1 @MrHohn 
cc: @freehan

### DIFF
--- a/pkg/e2e/interfaces.go
+++ b/pkg/e2e/interfaces.go
@@ -23,6 +23,8 @@ import (
 // UpgradeTest is an interface for writing generic upgrade test.
 // TODO: add version compatibility into the interface
 type UpgradeTest interface {
+	// Name returns the name/description of the test.
+	Name() string
 	// Init initialized the Upgrade Test.
 	Init(t *testing.T, s *Sandbox, framework *Framework) error
 	// PreUpgrade runs before master upgrade.
--- a/cmd/e2e-test/basic_https_test.go
+++ b/cmd/e2e-test/basic_https_test.go
@@ -111,7 +111,10 @@ func TestBasicHTTPS(t *testing.T) {
 			t.Logf("GCLB resources created (%s/%s)", s.Namespace, ing.Name)
 
 			// Perform whitebox testing.
-			gclb := whiteboxTest(ing, s, t, "")
+			gclb, err := e2e.WhiteboxTest(ing, s, Framework.Cloud, "")
+			if err != nil {
+				t.Fatalf("e2e.WhiteboxTest(%s/%s, ...) = %v, want nil", ing.Namespace, ing.Name, err)
+			}
 
 			deleteOptions := &fuzz.GCLBDeleteOptions{
 				SkipDefaultBackend: true,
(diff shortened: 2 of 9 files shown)


rramkumar1:
@skmatti Can you split up the migration of each test into a single commit each?

skmatti:
> @skmatti Can you split up the migration of each test into a single commit each?

Done

rramkumar1:
> > @skmatti Can you split up the migration of each test into a single commit each?
> 
> Done

Am I missing it? I would expect to see 3 commits.

skmatti:
> > > @skmatti Can you split up the migration of each test into a single commit each?
> > 
> > 
> > Done
> 
> Am I missing it? I would expect to see 3 commits.

Raised as separate PRs: https://github.com/kubernetes/ingress-gce/pull/952, https://github.com/kubernetes/ingress-gce/pull/921

pkg/e2e/interfaces.go
```diff
@@ -23,6 +23,8 @@ import (
 // UpgradeTest is an interface for writing generic upgrade test.
 // TODO: add version compatibility into the interface
 type UpgradeTest interface {
+	// Name returns the name/ description of the test.
```
rramkumar1: extra space

cmd/e2e-test/common/common.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"k8s.io/api/networking/v1beta1"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func WaitForStableIngress(expectUnreachable bool, ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T) *v1beta1.Ingress {
```
rramkumar1: Is WaitForStableIngress used for non-upgrade tests? If so, I wouldn't call s.PutStatus() here since the whole status is only used for upgrade tests.

cmd/e2e-test/common/common.go
```diff
@@ -0,0 +1,65 @@
...
+	ing, err := e2e.WaitForIngress(s, ing, options)
+	if err != nil {
+		t.Fatalf("error waiting for Ingress to stabilize: %v", err)
+	}
+
+	s.PutStatus(e2e.Stable)
+	return ing
+}
+
+func WhiteboxTest(ing *v1beta1.Ingress, s *e2e.Sandbox, t *testing.T, cloud cloud.Cloud, region string) *fuzz.GCLB {
```
rramkumar1: Can you put this in pkg/e2e/ instead?

cmd/e2e-test/common/common.go
rramkumar1: Nvmd I see that its only used for upgrade tests. Can we put this in a shared place other than cmd/? It's weird to have library code in cmd/

skmatti: I guess this function is not needed, dissolved it.

PR #952, opened 2019-11-20
Migrate Finalizer upgrade test to use the upgrade framework

skmatti:


cmd/e2e-test/upgrade/finalizer.go
```diff
@@ -0,0 +1,122 @@
...
+		AddPath("foo.com", "/", svcName, port80).
+		Build()
+	ingKey := common.NamespacedName(ing)
+	fr.crud = e2e.IngressCRUD{C: fr.framework.Clientset}
+	if _, err := fr.crud.Create(ing); err != nil {
+		fr.t.Fatalf("error creating Ingress %s: %v", ingKey, err)
+	}
+	fr.t.Logf("Ingress created (%s/%s)", fr.s.Namespace, ing.Name)
+
+	if ing, err = e2e.WaitForIngress(fr.s, ing, &e2e.WaitForIngressOptions{ExpectUnreachable: true}); err != nil {
```
rramkumar1: Well now it seems like e2e.WaitForIngress() + s.PutStatus() is being used in multiple places. Can we add a wrapper for these two calls in pkg/e2e and maybe call it UpgradeTestWaitForIngress?

cmd/e2e-test/upgrade_test.go
```diff
@@ -31,43 +31,60 @@ func TestGenericUpgrade(t *testing.T) {
...
-				// While k8s master is upgrading, it will return a connection refused
-				// error for any k8s resource we try to hit. We loop until the
-				// master upgrade has finished.
-				if s.MasterUpgrading() {
-					if err := test.DuringUpgrade(); err != nil {
-						t.Fatalf("For upgrade test %v, step DuringUpgrade failed due to %v", desc, err)
-					}
-					continue
-				}
+func genericUpgrade(t *testing.T, test e2e.UpgradeTest) {
```
rramkumar1: runUpgradeTest

cmd/e2e-test/upgrade_test.go
```diff
@@ -31,43 +31,60 @@ func TestGenericUpgrade(t *testing.T) {
...
-			func() {
-				// always mark the test as stable in order to unblock other upgrade tests.
-				defer s.PutStatus(e2e.Stable)
-				if err := test.PreUpgrade(); err != nil {
-					t.Fatalf("For upgrade test %v, step PreUpgrade failed due to %v", desc, err)
-				}
-			}()
+	for _, test := range []e2e.UpgradeTest{
+		upgrade.NewFinalizerUpgradeTest(),
+	} {
```
rramkumar1: Don't need a for loop for now since you only have one case.

pkg/e2e/helpers.go
```diff
@@ -90,6 +90,16 @@ func IsRfc1918Addr(addr string) bool {
 	return false
 }
 
+// UpgradeTestWaitForIngress waits for ingress to stabilize and set sandbox status to stable.
```
rramkumar1: Add a note to the comment that states that this is only used in upgrade tests similar to what you have below.

rramkumar1:
Looks good to me. Will officially approve once that last comment is addressed.

pkg/e2e/helpers.go
skmatti: Done

PR #956, opened 2019-11-22
Update finalizer tests to validate against an ingress finalizer

skmatti:
After adding new finalizer (V2 Finalizer), update finalizer upgrade tests to validate against either of v1 or v2 finalizers.

### DIFF
--- a/pkg/e2e/helpers.go
+++ b/pkg/e2e/helpers.go
@@ -143,7 +143,9 @@ func WaitForFinalizer(s *Sandbox, ing *v1beta1.Ingress) (*v1beta1.Ingress, error
 			return false, nil
 		}
 		ingFinalizers := ing.GetFinalizers()
-		if len(ingFinalizers) != 1 || ingFinalizers[0] != common.FinalizerKey {
+		if l := len(ingFinalizers); l > 1 {
+			return true, fmt.Errorf("unexpected number of finalizers for ingress %v, got %d", ing, l)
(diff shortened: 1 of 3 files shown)


rramkumar1:
@skmatti Can you update the description a bit to describe what exactly you are doing? To me it seems like you just created a couple helper functions and used them throughout the code but I would like to be sure of that.

skmatti:
> @skmatti Can you update the description a bit to describe what exactly you are doing? To me it seems like you just created a couple helper functions and used them throughout the code but I would like to be sure of that.

The v2-frontend-namer introduced a new finalizer (common.FinalizerKeyV2). Now there are two ingress finalizers after this change. But the finalizer tests were setup to validate only against old finalizer. This PR updates the tests so they expect one of the two ingress finalizers for validation.

pkg/e2e/helpers.go
```diff
@@ -473,3 +475,27 @@ func CheckNegStatus(svc *v1.Service, expectSvcPors []string) (annotations.NegSta
 	}
 	return negStatus, nil
 }
+
+// CheckFinalizer asserts that an ingress finalizer exists on Ingress.
+func CheckFinalizer(ing *v1beta1.Ingress) error {
```
rramkumar1: This is checking for any finalizer right? If so I would call this CheckForAnyFinalizer().

skmatti: Done

PR #961, opened 2019-12-03
Fix default backend port bug for ilb neg

spencerhance:
Fixes a bug where we create a neg for the default backend with the wrong port

cc @freehan 

/assign @MrHohn


spencerhance:
We'll need to cherry pick this to 1.8

PR #969, opened 2019-12-07
RegionalGCLBForVIP() Fix

spencerhance:
Fixes a bug where the ilb e2e test helper function may confuse multiple forwarding rules
from different VPCs with the same VIP

Will require another PR first to add the network flag to internal tests
/hold 

Blocked by #971

### DIFF
--- a/pkg/fuzz/gcp.go
+++ b/pkg/fuzz/gcp.go
@@ -387,6 +387,7 @@ func hasBetaResource(resourceType string, validators []FeatureValidator) bool {
 type GCLBForVIPParams struct {
 	VIP        string
 	Region     string
+	Network    string
 	Validators []FeatureValidator
 }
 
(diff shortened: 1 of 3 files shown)


pkg/fuzz/gcp.go
```diff
@@ -398,6 +399,7 @@ func GCLBForVIP(ctx context.Context, c cloud.Cloud, params *GCLBForVIPParams) (*
 		if err := RegionalGCLBForVIP(ctx, c, gclb, params); err != nil {
 			return nil, err
 		}
+		return gclb, nil
```
freehan: How about this?
```
err := RegionalGCLBForVIP(.....
return gclb, err
```

spencerhance: ooh much nicer, fixed

PR #983, opened 2020-01-07
Skip checking delete for default NEG

skmatti:
/assign @freehan

### DIFF
--- a/pkg/fuzz/gcp.go
+++ b/pkg/fuzz/gcp.go
@@ -43,6 +43,8 @@ const (
 	HttpsProtocol            = Protocol("HTTPS")
 	targetHTTPProxyResource  = "targetHttpProxies"
 	targetHTTPSProxyResource = "targetHttpsProxies"
+	kubeSystemNS             = "kube-system"
+	defaultHTTPBackend       = "default-http-backend"
 )
 
 // Protocol specifies GCE loadbalancer protocol.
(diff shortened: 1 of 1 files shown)


pkg/fuzz/gcp.go
```diff
@@ -237,12 +239,17 @@ func (g *GCLB) CheckResourceDeletion(ctx context.Context, c cloud.Cloud, options
...
-		_, err := c.BetaNetworkEndpointGroups().Get(ctx, &k)
+		ns, err := c.BetaNetworkEndpointGroups().Get(ctx, &k)
 		if err != nil {
 			if err.(*googleapi.Error) == nil || err.(*googleapi.Error).Code != http.StatusNotFound {
 				return fmt.Errorf("NetworkEndpointGroup %s is not deleted/error to get: %s", k.Name, err)
 			}
 		} else {
+			// TODO(smatti): Add NEG description to make this less error prone.
+			if options != nil && options.SkipDefaultBackend &&
+				strings.Contains(ns.Name, fmt.Sprintf("%s-%s", kubeSystemNS, defaultHTTPBackend)) {
```
freehan: are you sure the name is not truncated?

skmatti: I think default http name and namespace are constant values mentioned above. The only fields that lead to truncation are these two fields itself for NEG.

freehan: Just add comment about the limitations here

If the default backend service name changed, it may not get recognized here.

freehan: This is to prevent L7 ILB tests (using NEG backend service) to block other tests on WaitForDeletion..

PR #984, opened 2020-01-08
add neg usage metrics

freehan:
cc @skmatti

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/ingress-gce/pkg/context"
 	"k8s.io/ingress-gce/pkg/controller/translator"
 	"k8s.io/ingress-gce/pkg/flags"
+	usage "k8s.io/ingress-gce/pkg/metrics"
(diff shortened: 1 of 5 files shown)


pkg/neg/controller.go
```diff
@@ -325,22 +331,26 @@ func (c *Controller) processService(key string) error {
...
+	negUsage := usage.NegServiceState{}
 	svcPortInfoMap := make(negtypes.PortInfoMap)
 	if err := c.mergeDefaultBackendServicePortInfoMap(key, svcPortInfoMap); err != nil {
 		return err
 	}
+	negUsage.IngressNeg = len(svcPortInfoMap)
 	if err := c.mergeIngressPortInfo(service, types.NamespacedName{Namespace: namespace, Name: name}, svcPortInfoMap); err != nil {
 		return err
 	}
+	negUsage.IngressNeg += len(svcPortInfoMap)
```
MrHohn: QQ: why `+=`? Isn't the previous default backend service already included in `svcPortInfoMap`?

freehan: fixed. 

Good catch. That may cause double count in some very rare situation.

Issue #993, opened 2020-01-16
Ingress with both http and https enabled yields two separate IPs


skmatti:
The following sequence of configuration changes to an ingress leads to split IPs with ingress v1.8 (with enable-delete-unused-frontends set)
1. Create an Ingress with both http and https enabled. (Static IP is used)
2. Disable http (Static IP is retained and used by https forwarding rule)
2. Re-enable http (New ephemeral IP is created for http forwarding rule, but https forwarding rule uses static IP)

Ingress versions older than v1.8 do not run this issue because http forwarding rule is not deleted when disabled. But, manual deletion of http forwarding rule results in separate IPs.

PR #994, opened 2020-01-16
Bugfix: Use Static IP to create forwarding rule when available

skmatti:
Fixes https://github.com/kubernetes/ingress-gce/issues/993

### DIFF
--- a/pkg/loadbalancers/forwarding_rules.go
+++ b/pkg/loadbalancers/forwarding_rules.go
@@ -18,6 +18,8 @@ package loadbalancers
 
 import (
 	"fmt"
+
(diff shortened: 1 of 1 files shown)


bowei:
does this need cherry pick back to previous release?

skmatti:
> does this need cherry pick back to previous release?
Yes, this needs to be be cherry picked.

However, this is not common in old versions or may be not considered a bug?.
The occurrence of following events associated an ingress leads to the bug,
1. Enable both http and https
3. Delete http forwarding rule manually from GCP
4. Make changes to ingress so a sync is triggered

At this point, Both http and https have a VIP each created in GCP. And, only VIP associated with http is shown in ingress status.

PR #995, opened 2020-01-16
Make frontend resource deletion test verify that ingress VIP is unchanged 

skmatti:
Extends existing e2e test to cover bug: https://github.com/kubernetes/ingress-gce/issues/993

### DIFF
--- a/cmd/e2e-test/basic_test.go
+++ b/cmd/e2e-test/basic_test.go
@@ -299,6 +299,42 @@ func TestFrontendResourceDeletion(t *testing.T) {
 			if gclb, err = e2e.WhiteboxTest(ing, s, Framework.Cloud, ""); err != nil {
 				t.Fatalf("e2e.WhiteboxTest(%s, ...) = %v, want nil", ingKey, err)
 			}
+
+			expectedVIP := ing.Status.LoadBalancer.Ingress[0].IP
+			// Re-enable http/https and verify that the ingress uses same VIP.
+			if ing, err = crud.Get(ing.Namespace, ing.Name); err != nil {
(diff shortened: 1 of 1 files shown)


skmatti:
/assign @freehan 
Tested this and confirmed that it works on dev setup.

cmd/e2e-test/basic_test.go
```diff
@@ -299,6 +299,34 @@ func TestFrontendResourceDeletion(t *testing.T) {
...
+			}
+			if tc.disableHTTPS {
+				ingBuilder.AddTLS([]string{}, cert.Name)
+			}
+			ing = ingBuilder.Build()
+			if _, err := crud.Update(ing); err != nil {
+				t.Fatalf("Update(%s) = %v, want nil; ingress: %v", ingKey, err, ing)
+			}
+			t.Logf("Ingress updated (%s)", ingKey)
+			if ing, err = e2e.WaitForIngress(s, ing, &e2e.WaitForIngressOptions{ExpectUnreachable: true}); err != nil {
```
freehan: I am wondering if this would be waiting for LB config to be reconfigured. I think it would not. It will pass immediately after ingress is updated. Can you confirm?

skmatti: It may just work fine in this case. WaitForIngress validates both http and https paths.
https://github.com/kubernetes/ingress-gce/blob/7ac5f47e6cc8c11fdf3d5630272626e82a94b674/pkg/fuzz/validator.go#L120-L126

Here, I am enabling both http and https back again. So, it should wait for both checks to be validated.

freehan: got it.

PR #1003, opened 2020-01-17
Remove Status field from FrontendConfig

spencerhance:
Regen FrontendConfig and update hack/update-codegen.sh script


pkg/apis/frontendconfig/v1beta1/types.go
```diff
@@ -28,9 +28,7 @@ import (
 type FrontendConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-
 	Spec   FrontendConfigSpec   `json:"spec"`
```
bowei: don't need Spec

bowei: Hmm unfortunately BackendConfig already have spec and status. Let's keep it consistent w/ that, sorry about the churn.

spencerhance: I see your point, but i'm worried that differing from BackendConfig will be confusing for users

spencerhance: okay no worries, I'll close this

PR #1006, opened 2020-01-24
added test case TestBasicWindows

yliaog:
added test case TestBasicWindows that tests basic ingress for windows sever

### DIFF
--- a/cmd/e2e-test/basic_test.go
+++ b/cmd/e2e-test/basic_test.go
@@ -29,7 +29,15 @@ import (
 	"k8s.io/ingress-gce/pkg/utils/common"
 )
 
+func TestWindows(t *testing.T) {
+	testBasicOS(t, e2e.Windows)
+}
+
 func TestBasic(t *testing.T) {
+	testBasicOS(t, e2e.Linux)
+}
+
+func testBasicOS(t *testing.T, os e2e.OS) {
 	t.Parallel()
 
 	port80 := intstr.FromInt(80)
@@ -64,7 +72,7 @@ func TestBasic(t *testing.T) {
 
 			ctx := context.Background()
 
-			_, err := e2e.CreateEchoService(s, "service-1", nil)
+			_, err := e2e.CreateEchoServiceWithOS(s, "service-1", nil, os)
 			if err != nil {
 				t.Fatalf("error creating echo service: %v", err)
 			}
--- a/pkg/e2e/fixtures.go
+++ b/pkg/e2e/fixtures.go
@@ -48,8 +48,16 @@ import (
 )
 
 const (
-	echoheadersImage = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
-	porterPort       = 80
+	echoheadersImage        = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
+	echoheadersImageWindows = "gcr.io/gke-windows-testing/ingress-gce-echo-amd64-windows:master"
+	porterPort              = 80
+)
+
+type OS int
+
+const (
+	Linux OS = iota
+	Windows
 )
 
 var ErrSubnetExists = fmt.Errorf("ILB subnet in region already exists")
@@ -81,9 +89,19 @@ func CreateEchoService(s *Sandbox, name string, annotations map[string]string) (
 	return EnsureEchoService(s, name, annotations, v1.ServiceTypeNodePort, 1)
 }
 
+// CreateEchoServiceWithOS creates the pod and service serving echoheaders
+// Todo: (shance) remove this and replace uses with EnsureEchoService()
+func CreateEchoServiceWithOS(s *Sandbox, name string, annotations map[string]string, os OS) (*v1.Service, error) {
+	return ensureEchoService(s, name, annotations, v1.ServiceTypeNodePort, 1, os)
+}
+
 // EnsureEchoService that the Echo service with the given description is set up
 func EnsureEchoService(s *Sandbox, name string, annotations map[string]string, svcType v1.ServiceType, numReplicas int32) (*v1.Service, error) {
-	if err := EnsureEchoDeployment(s, name, numReplicas, NoopModify); err != nil {
+	return ensureEchoService(s, name, annotations, svcType, numReplicas, Linux)
+}
+
+func ensureEchoService(s *Sandbox, name string, annotations map[string]string, svcType v1.ServiceType, numReplicas int32, os OS) (*v1.Service, error) {
+	if err := ensureEchoDeployment(s, name, numReplicas, NoopModify, os); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


bowei:
Wondering if this should be a separate test case that runs a check on a windows node if one exists on the cluster. 
I would like to avoid having to maintain specific OS flags on the test executable.

yliaog:
For windows, the backend service pod container image has to be built to run on windows, and it needs  to be scheduled to run on windows.

the test cases are the same for windows and for linux.

yliaog:
@rramkumar1 ping ...

yliaog:
@bowei PTAL

yliaog:
@bowei 
friendly ping ...

pkg/e2e/fixtures.go
```diff
@@ -48,8 +48,9 @@ import (
 )
 
 const (
-	echoheadersImage = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
-	porterPort       = 80
+	echoheadersImage        = "gcr.io/k8s-ingress-image-push/ingress-gce-echo-amd64:master"
+	echoheadersImageWindows = "gcr.io/gke-windows-testing/ingress-gce-echo-amd64-windows:master"
+	porterPort              = 80
```
rramkumar1: just call this port?

cmd/e2e-test/basic_test.go
```diff
@@ -29,7 +29,15 @@ import (
 	"k8s.io/ingress-gce/pkg/utils/common"
 )
 
+func TestBasicWindows(t *testing.T) {
+	testBasicOS(t, "windows")
```
rramkumar1: Can we use the const iota pattern here rather than passing in a raw string? So just define a type called "OS" and then have two values for it

pkg/e2e/fixtures.go
yliaog: it was called porterPort, not sure why though

cmd/e2e-test/basic_test.go
yliaog: sure, changed, PTAL

pkg/e2e/fixtures.go
```diff
@@ -48,8 +48,16 @@ import (
...
+)
+
+type OS int
+
+const (
+	Linux   OS = 0
```
rramkumar1: = iota. 

See https://github.com/golang/go/wiki/Iota

pkg/e2e/fixtures.go
```diff
@@ -179,7 +205,9 @@ func EnsureEchoDeployment(s *Sandbox, name string, numReplicas int32, modify fun
 			},
 		},
 	}
-
+	if os == Windows {
```
rramkumar1: Can you move this conditional up so that you only check if os == Windows once? 

You can store the node selector in a var similar to image.

yliaog: done

pkg/e2e/fixtures.go
```diff
@@ -81,9 +89,19 @@ func CreateEchoService(s *Sandbox, name string, annotations map[string]string) (
...
+// CreateEchoServiceOS creates the pod and service serving echoheaders
+// Todo: (shance) remove this and replace uses with EnsureEchoService()
+func CreateEchoServiceOS(s *Sandbox, name string, annotations map[string]string, os OS) (*v1.Service, error) {
+	return ensureEchoServiceOS(s, name, annotations, v1.ServiceTypeNodePort, 1, os)
+}
+
 // EnsureEchoService that the Echo service with the given description is set up
 func EnsureEchoService(s *Sandbox, name string, annotations map[string]string, svcType v1.ServiceType, numReplicas int32) (*v1.Service, error) {
-	if err := EnsureEchoDeployment(s, name, numReplicas, NoopModify); err != nil {
+	return ensureEchoServiceOS(s, name, annotations, svcType, numReplicas, Linux)
```
rramkumar1: ensureEchoService

pkg/e2e/fixtures.go
```diff
@@ -135,16 +153,27 @@ func EnsureEchoService(s *Sandbox, name string, annotations map[string]string, s
 
 // EnsureEchoDeployment ensures that the Echo deployment with the given description is set up
 func EnsureEchoDeployment(s *Sandbox, name string, numReplicas int32, modify func(deployment *apps.Deployment)) error {
+	return ensureEchoDeploymentOS(s, name, numReplicas, modify, Linux)
+}
+
+func ensureEchoDeploymentOS(s *Sandbox, name string, numReplicas int32, modify func(deployment *apps.Deployment), os OS) error {
```
rramkumar1: ensureEchoDeployment

pkg/e2e/fixtures.go
```diff
@@ -81,9 +89,19 @@ func CreateEchoService(s *Sandbox, name string, annotations map[string]string) (
 	return EnsureEchoService(s, name, annotations, v1.ServiceTypeNodePort, 1)
 }
 
+// CreateEchoServiceOS creates the pod and service serving echoheaders
+// Todo: (shance) remove this and replace uses with EnsureEchoService()
+func CreateEchoServiceOS(s *Sandbox, name string, annotations map[string]string, os OS) (*v1.Service, error) {
```
rramkumar1: Just call this CreateWindowsEchoService

rramkumar1:
@yliaog Last comments above, then LGTM.

pkg/e2e/fixtures.go
yliaog: In basic_test.go
 75                         _, err := e2e.CreateEchoServiceOS(s, "service-1", nil, os)

if we change the name to CreateWindowsEchoService, it cannot be used for 'Linux'

pkg/e2e/fixtures.go
yliaog: are you suggesting to change ensureEchoServiceOS to ensureEchoService? that won't work.

yliaog: never mind, i guess i know what you mean.

pkg/e2e/fixtures.go
rramkumar1: That's fine, the original CreateEchoService func can be used for Linux. I don't really like appending OS to the name of the func.

yliaog: do you mean we do something like below to replace _, err := e2e.CreateEchoServiceOS(s, "service-1", nil, os) ?

switch os {
    e2e.Linux:  _, err := e2e.CreateEchoService(s, "service-1", nil)
    e2e.Windows: _, err := e2e.CreateWindowsEchoService(s, "service-1", nil)
}

rramkumar1: No, just rename CreateEchoServiceOS to CreateWindowsService

yliaog: got it. done. thanks.

PR #1008, opened 2020-01-28
Ignore 'resource being used by' error

skmatti:
This does the following,
1. Removes the code block that is never reached, because the case is handled in upstream. 
2. Ignore 'resource being used by' error on healthcheck deletion because the associated backend service still uses it.

### DIFF
--- a/pkg/backends/syncer.go
+++ b/pkg/backends/syncer.go
@@ -15,7 +15,6 @@ package backends
 
 import (
 	"fmt"
-	"net/http"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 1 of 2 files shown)


rramkumar1:
> 1. Removes the code block that is never reached, because the case is handled in upstream.

@skmatti Can you point to where its handled upstream? 

> 2. Ignore 'resource being used by' error on healthcheck deletion because the associated backend service still uses it.

Can you explain why we would get this error in the first place? Are we sharing health checks across many backend services? I thought that was not the case.

skmatti:
> > 1. Removes the code block that is never reached, because the case is handled in upstream.
> 
> @skmatti Can you point to where its handled upstream?
https://github.com/kubernetes/ingress-gce/blob/541ea6aa72d15753e9ec0c65bbd0e71339675332/pkg/backends/backends.go#L150-L155
> > 1. Ignore 'resource being used by' error on healthcheck deletion because the associated backend service still uses it.
> 
> Can you explain why we would get this error in the first place? Are we sharing health checks across many backend services? I thought that was not the case.
We are running into this because backend resource deletion failure due to `resource being used by` interpreted as no error. The backend resource is not deleted but we are trying to delete health check associated with it. So, we get 'resource being used it' error.
No, health checks are not being shared. But, the single backend resource referencing this health check is not deleted.

rramkumar1:
/ok-to-test

The removal of the unneeded error checking looks good to me. I would like @spencerhance to take a look at the other change since it involves L7 ILB.

PR #1010, opened 2020-01-30
Add HealthCheckConfig to v1beta1 BackendConfig CRD

AnishShah:


omatsko:
If this is going to be included with the next release, could someone please tell when is it planned?

rramkumar1:
@omatsko We are planning to include this as part of our 1.10 release. We are about to release 1.9 so 1.10 will hopefully drop in a couple weeks after that. Will keep this thread updated.

bryanhuntesl:
> @omatsko We are planning to include this as part of our 1.10 release. We are about to release 1.9 so 1.10 will hopefully drop in a couple weeks after that. Will keep this thread updated.

Any news - id I understand correctly, this prevents exposing grafana because / returns 302 - and the healthcheck configured for the service is ignored by the ingress ?

tmc:
@rramkumar1  It seems missed here to not allow Host header configuration, thoughts?

mad-it:
Any updates on the release schedule for this?

johndoe123441:
Looking for an update on this item

Gatsby-Lee:
This merged seems available on beta. I am on 1.17.6-gke.11.

I don't need change in Ingress.

```yaml
kind: BackendConfig
metadata:
  name: config-default
  namespace: apimars-prod
spec:
  healthCheck:
    checkIntervalSec: 10
    timeoutSec: 3
    requestPath: /healthz
```

```
apiVersion: v1
kind: Service
metadata:
  name: test-healthz
  namespace: apimars-prod
  annotations:
    beta.cloud.google.com/backend-config: '{"default": "config-default"}'
spec:
  ports:
    - port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app: test-healthz
  type: NodePort
```

Gatsby-Lee:
I was wrong. I don't think setting healthcheck through Backendconfig works.
Ingress can use custom healthcheck only if Service exists before Ingress knows about the Service.

And, even if custom healthcheck works by bringing up Ingress after Service,
the custom healthcheck won't work if new Pod is deployed.

This is what @nicksardo explained before in different msg thread.

BTW, ingress doesn't have to be removed to use custom healthcheck.
if Service info is removed from Ingress and added back again, then Ingress uses custom healthcheck again.
(But, we can't do this on prod) lol

PR #1017, opened 2020-02-03
Update k8s-cloud-provider to 1.12.0

spencerhance:
Note that even though k8s-cloud-provider now uses go modules, the tag doesn't show up correctly
in the go.mod file since k8s-cloud-provider versions are not prefixed with a v


spencerhance:
@freehan Can you take a look at this?  It is blocking #1021

PR #1022, opened 2020-02-06
Add deployment guide for GLBC in non-gcp mode.

cxhiano:


docs/deploy/gke/non-gcp/README.md
```diff
@@ -0,0 +1,141 @@
+# Overview
+
+This document will deploy the self managed Ingress-GCE controller in Non-GCP
```
bowei: you mean non-GKE

bowei: or something else...

docs/deploy/gke/non-gcp/README.md
```diff
@@ -0,0 +1,141 @@
...
+ mode.
```
freehan: This document will deploy the self managed Ingress-GCE controller running on GCP in Non-GCP
 mode.

docs/deploy/gke/non-gcp/README.md
cxhiano: Eventually we'd like deploy it on a non-GCP environment like GKE On-Prem. Currently we deploy it on GCP GKE to facilitate testing.

PR #1029, opened 2020-02-17
Set healthcheck from BackendConfig

bowei:
```
    Take healthcheck configuration overrides from the backendconfig

    This enables taking the healthcheck configuration from a backendconfig
    associated with the given Service.

    As part of this PR, we also clean up the override handling to be easier
    to understand where the settings come from.

    For a given ServicePort `sp` and its associated healthcheck `hc`:

    1. `sp` has some default settings depending on its type (e.g. IG, NEG, ILB).
    2. SyncServicePort() will override those settings if the Probe exists.
         Path, Host, TimeoutSec, CheckIntervalSec.
    3. We then sync with what exists in GCE:
        3.1 If the healthcheck is does not exist, it will be CREATED.
            3.1.1 If it has a backendconfig, then the following settings are used:
                    CheckIntervalSec, TimeoutSec, HealthyThreshold,
                    UnhealthyThreshold, Type, RequestPath
                  We have `Port` in the struct, but it does not seem to be a valid
                  field to use so it is currently ignored.
        3.2 Otherwise the healthcheck will need to be UPDATED. The criteria
            for update can be found in `calculateDiff()`:
            3.2.1 Services w/out backendconfig will only update when `Protocol`,
                  `PortSpecification` changes.
            3.2.2 Services w/ backendconfig will also consider any fields
                  covered by the backendconfig (see above)
            3.2.3 We merge in the settings from the existing healthcheck. As
                  users have (in the past) done edits to their HCs directly, we
                  do not want to break any previous behavior. Thus we merge in
                  all of the settings on HTTPHealthCheck as well as any fields on
                  HealthCheck itself, with the exception of: `Port`, `PortSpecification`
            3.2.2 We now consider the backendconfig to be a final override.
                  See list above for settings from the backendconfig.

    Several situations do not seem to be handled currently:

    * There is no way to change HC based on probe after it is created.
      This is because there is no way to tell if the difference is due
      to a user changing the setting directly in GCE or the HC should
      be updated.
    * Changing types to/from NEG, ILB will leak the old healthcheck.
      User healthchecksettings will not propagate to the new healthcheck.
    * backendconfig seems to be missing `Host`
```

```
TestSyncServicePort does the following:

1. Setup any preexisting objects (tc.setup) that should exist in the
   mocked GCE.
2. Run one sync given a ServicePort.
3. Attempt to re-run the sync with the same ServicePort, checking that
   no update will be needed.
```

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -18,8 +18,10 @@ package healthchecks
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -30,6 +32,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/ingress-gce/pkg/annotations"
+	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
@@ -116,12 +119,16 @@ func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (
 		klog.V(4).Infof("Applying httpGet settings of readinessProbe to health check on port %+v", sp)
 		applyProbeSettingsToHC(probe, hc)
 	}
-	return h.sync(hc)
+	var bchcc *backendconfigv1.HealthCheckConfig
+	if sp.BackendConfig != nil {
+		bchcc = sp.BackendConfig.Spec.HealthCheck
+	}
+	return h.sync(hc, bchcc)
 }
 
 // sync retrieves a health check based on port, checks type and settings and updates/creates if necessary.
 // sync is only called by the backends.Add func - it's not a pool like other resources.
-func (h *HealthChecks) sync(hc *HealthCheck) (string, error) {
+func (h *HealthChecks) sync(hc *HealthCheck, bchcc *backendconfigv1.HealthCheckConfig) (string, error) {
 	var scope meta.KeyType
 	// TODO(shance): find a way to remove this
 	if hc.forILB {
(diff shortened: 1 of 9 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -121,47 +137,57 @@ func (h *HealthChecks) Sync(hc *HealthCheck) (string, error) {
...
+		return selfLink, err
+	}
+	if err != nil {
+		return "", err
 	}
 
-	if needToUpdate(existingHC, hc) {
-		err = h.update(existingHC, hc)
-		return existingHC.SelfLink, err
+	// First, merge in the configuration from the existing healthcheck to cover
```
rramkumar1: Is this to ensure a smooth migration path to BackendConfig?

pkg/healthchecks/healthchecks.go
```diff
@@ -575,17 +603,79 @@ func (hc *HealthCheck) Version() meta.Version {
...
+	if old.Protocol() != new.Protocol() {
+		changes.add("Protocol", string(old.Protocol()), string(new.Protocol()))
+	}
 	if old.PortSpecification != new.PortSpecification {
-		klog.V(2).Infof("Updating health check %v because it has port specification %q but need %q", old.Name, old.PortSpecification, new.PortSpecification)
-		return true
+		changes.add("PortSpecification", old.PortSpecification, new.PortSpecification)
 	}
-	return false
+	// TODO: why don't we check Port?
```
rramkumar1: What does this mean?

pkg/healthchecks/healthchecks.go
bowei: This is behavior that we have to preserve (unfortunately)

pkg/healthchecks/healthchecks.go
bowei: We omit quite a few fields when checking for update, which is intentional but not documented (nor reasoning)

bowei:
This should be ready for review.

pkg/healthchecks/healthchecks.go
```diff
@@ -131,47 +138,57 @@ func (h *HealthChecks) sync(hc *HealthCheck) (string, error) {
...
-		if err = h.create(hc); err != nil {
+	if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+		klog.V(2).Infof("Health check %q does not exist, creating (hc=%+v, bchcc=%+v)", hc.Name, hc, bchcc)
+		if err = h.create(hc, bchcc); err != nil {
+			klog.Errorf("Health check %q creation error: %v", hc.Name, err)
 			return "", err
 		}
-
-		return h.getHealthCheckLink(hc.Name, hc.Version(), scope)
+		// TODO(bowei) -- we don't need to fetch the self-link here as it is
```
rramkumar1: I think you can use NewHealthChecksResourceID in k8s_cloud_provider to generate this link.

pkg/healthchecks/healthchecks.go
```diff
@@ -131,47 +138,57 @@ func (h *HealthChecks) sync(hc *HealthCheck) (string, error) {
...
 
-	if needToUpdate(existingHC, hc) {
-		err = h.update(existingHC, hc)
-		return existingHC.SelfLink, err
+	// First, merge in the configuration from the existing healthcheck to cover
+	// the case where the user has changed healthcheck settings outside of
+	// GKE.
+	premergeHC := hc
+	hc = mergeUserSettings(existingHC, hc)
+	klog.V(3).Infof("mergeHealthcheck(%v,%v) = %v", existingHC, premergeHC, hc)
```
rramkumar1: %+v?

pkg/healthchecks/healthchecks.go
```diff
@@ -187,7 +204,12 @@ func (h *HealthChecks) createILB(hc *HealthCheck) error {
 	return nil
 }
 
-func (h *HealthChecks) create(hc *HealthCheck) error {
+func (h *HealthChecks) create(hc *HealthCheck, bchcc *backendconfigv1.HealthCheckConfig) error {
+	if bchcc != nil {
+		klog.V(2).Infof("Health check %q has backendconfig override (%+v)", hc.Name, bchcc)
```
rramkumar1: You log this same thing in line 165 as well. Maybe remove one of them?

pkg/healthchecks/healthchecks.go
```diff
@@ -235,53 +255,68 @@ func (h *HealthChecks) updateILB(oldHC, newHC *HealthCheck) error {
...
+
+	hc.HTTPHealthCheck = existing.HTTPHealthCheck
+	hc.HealthCheck.CheckIntervalSec = existing.HealthCheck.CheckIntervalSec
+	hc.HealthCheck.HealthyThreshold = existing.HealthCheck.HealthyThreshold
+	hc.HealthCheck.TimeoutSec = existing.HealthCheck.TimeoutSec
+	hc.HealthCheck.UnhealthyThreshold = existing.HealthCheck.UnhealthyThreshold
+
+	if existing.HealthCheck.LogConfig != nil {
+		l := *existing.HealthCheck.LogConfig
+		hc.HealthCheck.LogConfig = &l
```
rramkumar1: Is this an Alpha-only field? If that is the case, where do we configure this HC to be Alpha?

pkg/healthchecks/healthchecks.go
```diff
@@ -532,25 +563,23 @@ func (hc *HealthCheck) ToComputeHealthCheck() (*compute.HealthCheck, error) {
...
-	return &hc.HealthCheck
+	x := hc.HealthCheck // Make a copy to ensure no aliasing.
+	return &x
 }
 
 func (hc *HealthCheck) merge() {
+	// Cannot specify both portSpecification and port field.
+	if hc.PortSpecification != "" {
+		hc.Port = 0
+	}
```
rramkumar1: Why was this added here?

pkg/healthchecks/healthchecks.go
```diff
@@ -569,33 +599,89 @@ func (hc *HealthCheck) merge() {
...
+	if c.UnhealthyThreshold != nil {
+		hc.UnhealthyThreshold = *c.UnhealthyThreshold
 	}
+	if c.Type != nil {
+		hc.Type = *c.Type
+	}
+	if c.RequestPath != nil {
+		hc.RequestPath = *c.RequestPath
+	}
+	if c.Port != nil {
```
rramkumar1: I'm a little confused by this? Shouldn't setting port be allowed for IG backends?

pkg/healthchecks/healthchecks.go
bowei: Doing this handles all resource versions due to how version conversion via JSON drops fields when it gets converted.
This is proactive. I don't expect to see this field right now.

pkg/healthchecks/healthchecks.go
bowei: I did this kind of on purpose to make the log line shorter...

pkg/healthchecks/healthchecks.go
bowei: It's not the same.  This is on the create path, that instance is on the update path.

pkg/healthchecks/healthchecks.go
bowei: This was done on all paths that ended up at merge() previously.
I consolidated it here and remove the other instances.
Really, this should exist, but people have been writing imprecise code, so I am keeping it for now.

pkg/healthchecks/healthchecks.go
bowei: Doesn't make sense. Port is set from NodePort.
NEG should take port from the neg itself.

bowei: Would be good to hear about a use case though...

pkg/healthchecks/healthchecks.go
spencerhance: Might make sense to differentiate the log lines

pkg/healthchecks/healthchecks.go
```diff
@@ -569,33 +599,89 @@ func (hc *HealthCheck) merge() {
...
+		klog.Warningf("Setting Port is not supported (healthcheck %q, backendconfig = %+v)", hc.Name, c)
+	}
+}
+
+// fieldDiffs encapsulate which fields are different between health checks.
+type fieldDiffs struct {
```
spencerhance: This might be easier if in a new file

pkg/healthchecks/healthchecks.go
bowei: You can distinguish from the source line in klog.

pkg/healthchecks/healthchecks.go
bowei: I will send a PR to split the file.
The diff gets very hard to read when I did the split...

bowei:
Some of the TODOs can't be done right now (esp the ones around bugs).
I left them so we can file issues around them and track them long-term.

pkg/healthchecks/healthchecks.go
spencerhance: Good point, seems reasonable to leave it as is then

pkg/healthchecks/healthchecks.go
spencerhance: sgtm

pkg/healthchecks/healthchecks.go
bowei: Will do this in a follow-up PR

pkg/healthchecks/healthchecks.go
bowei: actuall, let me fix this.

bowei:
I just tested this by manually running the e2e test and am getting a PR together for the e2e test.
Going to merge this for now...

mark-church:
@bowei based off of how users are desiring to configure health checks for istio-ingressgateway, we'll need to support the host field as well.

PR #1030, opened 2020-02-20
Fix backendconfig annotation key for beta

skmatti:
/assign @freehan 
cc: @spencerhance


freehan:
/lgtm
/approve

Probably need to understand what is the follow up to flip it to GA for some versions.

PR #1031, opened 2020-02-22
Add L4 ILB usage metrics

skmatti:
/assign @freehan @prameshj

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/firewalls"
 	"k8s.io/ingress-gce/pkg/healthchecks"
+	"k8s.io/ingress-gce/pkg/metrics"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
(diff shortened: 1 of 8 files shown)


pkg/metrics/types.go
```diff
@@ -35,6 +35,28 @@ type NegServiceState struct {
...
+
+// VmPrimaryIpNegType contains whether a GCE_VM_PRIMARY_IP NEG is requesting for
+// local traffic (or service external policy set to local).
+type VmPrimaryIpNegType struct {
+	isOnlyLocal bool
+}
+
+// NewVmPrimaryIpNegType returns a new VmPrimaryIpNegType.
+func NewVmPrimaryIpNegType(onlyLocal bool) *VmPrimaryIpNegType {
+	return &VmPrimaryIpNegType{isOnlyLocal: onlyLocal}
```
prameshj: Maybe call this trafficPolicyLocal?

pkg/metrics/metrics.go
```diff
@@ -226,20 +264,63 @@ func (im *ControllerMetrics) computeIngressMetrics() (map[feature]int, map[featu
...
+	counts := map[feature]int{
+		l4ILBService:      0,
+		l4IlbGlobalAccess: 0,
+		l4IlbCustomSubnet: 0,
+	}
+
+	for key, state := range im.l4ILBServiceMap {
+		klog.V(6).Infof("ILB Service %s has EnabledGlobalAccess: %t, EnabledCustomSubnet: %t", key, state.EnabledGlobalAccess, state.EnabledCustomSubnet)
+		counts[l4ILBService] += 1
+		if state.EnabledGlobalAccess {
```
prameshj: A service can be updated to disable/enable global acess anytime. Will this handle the case where a service is created with global access on, then updated to disable it?

pkg/loadbalancers/l4.go
```diff
@@ -143,6 +146,10 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) error {
 		klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s, err %v", l.NamespacedName.String(), err)
 		return err
 	}
+	if retErr == nil {
```
prameshj: We can add this before line 139, since failing delete of a shared healthcheck is ok.  Or we can do it in the caller of EnsureInternalLoadBalancerDeleted

pkg/metrics/types.go
skmatti: Done

pkg/loadbalancers/l4.go
skmatti: Modified this to take those cases into account. Keeping it in EnsureInternalLoadBalancerDeleted makes metrics collector symmetric, so prefer keeping it here.

pkg/metrics/metrics.go
skmatti: Yes, the state of a service gets updated on each of it's sync. Metrics are computed based on the current state of all services. Since the metrics are computed at 10 mins interval, the updates may be delayed by 10 minutes at max.

pkg/loadbalancers/l4.go
```diff
@@ -135,13 +138,17 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) error {
...
-			return retErr
+		// Ignore deletion error due to health check in use by another resource.
+		// This will be hit if this is a shared healthcheck.
+		if !utils.IsInUsedByError(err) {
+			klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s, err %v", l.NamespacedName.String(), err)
+			return err
 		}
-		klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s, err %v", l.NamespacedName.String(), err)
-		return err
+		klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
```
prameshj: nit: Maybe move comments from line 141, 142 here?

PR #1040, opened 2020-02-25
Add feature flag `EnableBackendConfigHealthCheck`

bowei:
-enable-backendconfig-healthcheck will enable the backendconfig healthcheck feature


bowei:
sorted the flags struct to make it more obvious where to put future flags.

PR #1041, opened 2020-02-25
Add support for Access Logs in BackendConfig

skmatti:


freehan:
Is this still needed?

skmatti:
Yes, this will go into v1.9.0. Older versions do not need this as a feature. But, starting v1.9.0, users can configure access logs as they like.

pkg/backends/features/logging.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// expectedBackendServiceLogConfig returns composite.BackendServiceLogConfig for
+// the access log settings specified in the BackendConfig to the passed in
+// composite.BackendService.
+func expectedBackendServiceLogConfig(sp utils.ServicePort) *composite.BackendServiceLogConfig {
+	logConfig := &composite.BackendServiceLogConfig{
+		Enable: sp.BackendConfig.Spec.Logging.Enable,
+	}
+	if !sp.BackendConfig.Spec.Logging.Enable || sp.BackendConfig.Spec.Logging.SampleRate == nil {
```
freehan: you need to check if Logging is nil?

skmatti: This is checked in caller function. Added a comment that this is assumed to be not nil. Would it be cleaner if I explicitly include a check here as well?.

Issue #1042, opened 2020-02-27
fix inaccessible links in the CHANGELOG.md


npu21:
There are two inaccessible links in the CHANGELOG.md
1. do not use condition predicate while getting zone for node [\#576](https://github.com/kubernetes/ingress-gce/pull/576) ([pondohva](https://github.com/pondohva))
2. do not leak LB when ingress class is changed [\#547](https://github.com/kubernetes/ingress-gce/pull/547) ([pondohva](https://github.com/pondohva))

PR #1043, opened 2020-02-27
fix inaccessible links

npu21:
https://github.com/kubernetes/ingress-gce/issues/1042


spencerhance:
I can't think of a good reason why it would be a bad idea, but it seems weird to edit old changelog entries just because someone changed their github username

bowei:
This kind of change doesn't make sense.

/close

PR #1045, opened 2020-03-01
Use backendconfig v1 for e2e tests

skmatti:
- Existing e2e tests are migrated to use backendconfig v1
- one e2e test for backendconfig v1beta1 API is added

### DIFF
--- a/cmd/e2e-test/affinity_beta_test.go
+++ b/cmd/e2e-test/affinity_beta_test.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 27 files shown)


cmd/e2e-test/affinity_beta_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+	testutils "k8s.io/ingress-gce/pkg/test"
+)
+
+func TestAffinityBeta(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
```
freehan: I am wondering if it is worth merging test cases into one transition test together. Meaning that each test case becomes a state/expectation. The test would go thru each test case, trigger a transition and verify if the final state is expected.

Then there is no need to create separate ingress for each test case.

skmatti: Discussed offline that we should keep beta tests in a separate file for simplicity. 

Consolidated affinity transitions to use a single ingress as suggested.

Issue #1049, opened 2020-03-11
Backendconfig and Frontendconfig CRD informers break with protobuf


skmatti:
CRD watch fails with clientset that overrides 
`ContentType = "application/vnd.kubernetes.protobuf"`(Ref: https://github.com/kubernetes/kubernetes/issues/55541)

This affects ingress v1.6 and above (introduced by https://github.com/kubernetes/ingress-gce/pull/847)

PR #1050, opened 2020-03-11
Revert #849[Use protobufs for communication with apiserver] in v1.6

skmatti:
Fixes https://github.com/kubernetes/ingress-gce/issues/1049

### DIFF
--- a/cmd/glbc/app/clients.go
+++ b/cmd/glbc/app/clients.go
@@ -46,16 +46,6 @@ const (
 
 // NewKubeConfig returns a Kubernetes client config given the command line settings.
 func NewKubeConfig() (*rest.Config, error) {
-	config, err := newKubeConfig()
-	if err != nil {
-		return nil, err
-	}
-	// Use protobufs for communication with apiserver
-	config.ContentType = "application/vnd.kubernetes.protobuf"
-	return config, nil
-}
-
-func newKubeConfig() (*rest.Config, error) {
 	if flags.F.InCluster {
 		klog.V(0).Infof("Using in cluster configuration")
 		return rest.InClusterConfig()
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -191,19 +191,40 @@ func NewLoadBalancerController(
 	// BackendConfig event handlers.
 	ctx.BackendConfigInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			beConfig := obj.(*backendconfigv1beta1.BackendConfig)
+			beConfig, ok := obj.(*backendconfigv1beta1.BackendConfig)
+			if !ok {
+				klog.Errorf("Cannot cast obj to BackendConfig (obj type is %T)", obj)
+				return
+			}
 			ings := operator.Ingresses(ctx.Ingresses().List()).ReferencesBackendConfig(beConfig, operator.Services(ctx.Services().List())).AsList()
 			lbc.ingQueue.Enqueue(convert(ings)...)
 		},
 		UpdateFunc: func(old, cur interface{}) {
 			if !reflect.DeepEqual(old, cur) {
-				beConfig := cur.(*backendconfigv1beta1.BackendConfig)
+				beConfig, ok := cur.(*backendconfigv1beta1.BackendConfig)
+				if !ok {
+					klog.Errorf("Cannot cast obj to BackendConfig (obj type is %T)", cur)
+					return
+				}
 				ings := operator.Ingresses(ctx.Ingresses().List()).ReferencesBackendConfig(beConfig, operator.Services(ctx.Services().List())).AsList()
 				lbc.ingQueue.Enqueue(convert(ings)...)
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
-			beConfig := obj.(*backendconfigv1beta1.BackendConfig)
+			beConfig, ok := obj.(*backendconfigv1beta1.BackendConfig)
+			if !ok {
+				klog.V(3).Infof("Cannot cast obj to BackendConfig (obj type is %T), casting into DeletedFinalStateUnknown", obj)
+				var tombstone cache.DeletedFinalStateUnknown
+				tombstone, ok = obj.(cache.DeletedFinalStateUnknown)
+				if !ok {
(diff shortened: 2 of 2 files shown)


cmd/glbc/app/clients.go
```diff
@@ -44,18 +44,17 @@ const (
...
-// NewKubeConfig returns a Kubernetes client config given the command line settings.
-func NewKubeConfig() (*rest.Config, error) {
-	config, err := newKubeConfig()
-	if err != nil {
-		return nil, err
-	}
+// KubeConfigForProtobuf returns a shallow copy of client config after updating
+// content type to protobuf.
+func KubeConfigForProtobuf(kubeConfig *rest.Config) *rest.Config {
+	configShallowCopy := *kubeConfig
```
bowei: why not deep copy?

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
 	ctx.BackendConfigInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			beConfig := obj.(*backendconfigv1beta1.BackendConfig)
+			beConfig, ok := obj.(*backendconfigv1beta1.BackendConfig)
+			if !ok {
+				klog.Errorf("Error decoding object(%T) into BackendConfig", obj)
+				return
+			}
 			ings := operator.Ingresses(ctx.Ingresses().List()).ReferencesBackendConfig(beConfig, operator.Services(ctx.Services().List())).AsList()
+			if len(ings) == 0 {
```
bowei: I think Enqueue already logs?

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
 		UpdateFunc: func(old, cur interface{}) {
 			if !reflect.DeepEqual(old, cur) {
-				beConfig := cur.(*backendconfigv1beta1.BackendConfig)
+				beConfig, ok := cur.(*backendconfigv1beta1.BackendConfig)
+				if !ok {
+					klog.Errorf("Error decoding object(%T) into BackendConfig", cur)
+					return
+				}
 				ings := operator.Ingresses(ctx.Ingresses().List()).ReferencesBackendConfig(beConfig, operator.Services(ctx.Services().List())).AsList()
+				if len(ings) == 0 {
```
bowei: Same here

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
 	// BackendConfig event handlers.
 	ctx.BackendConfigInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			beConfig := obj.(*backendconfigv1beta1.BackendConfig)
+			beConfig, ok := obj.(*backendconfigv1beta1.BackendConfig)
+			if !ok {
+				klog.Errorf("Error decoding object(%T) into BackendConfig", obj)
```
bowei: Let use standard terminology:

"Cannot cast obj to BackendConfig (obj type is %T)"

fix below as well

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
+				}
+				klog.V(3).Infof("BackendConfig %s/%s updated, enqueuing ingresses %v", beConfig.Namespace, beConfig.Name, ings)
 				lbc.ingQueue.Enqueue(convert(ings)...)
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
-			beConfig := obj.(*backendconfigv1beta1.BackendConfig)
+			beConfig, ok := obj.(*backendconfigv1beta1.BackendConfig)
+			if !ok {
+				klog.V(3).Infof("Failed to decode object(%T) as BackendConfig, casting into DeletedFinalStateUnknown", obj)
```
bowei: see above

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
+				var tombstone cache.DeletedFinalStateUnknown
+				tombstone, ok = obj.(cache.DeletedFinalStateUnknown)
+				if !ok {
+					klog.Errorf("Error decoding object(%T) into DeletedFinalStateUnknown", obj)
```
bowei: see above

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
+					return
+				}
+				if beConfig, ok = tombstone.Obj.(*backendconfigv1beta1.BackendConfig); !ok {
+					klog.Errorf("Error decoding deleted state(%T) of object into BackendConfig", tombstone.Obj)
```
bowei: see above

pkg/controller/controller.go
```diff
@@ -191,20 +191,56 @@ func NewLoadBalancerController(
...
+					return
+				}
+			}
 			ings := operator.Ingresses(ctx.Ingresses().List()).ReferencesBackendConfig(beConfig, operator.Services(ctx.Services().List())).AsList()
+			if len(ings) == 0 {
```
bowei: again, check that we log in Enqueue so we don't have to log here as well

cmd/glbc/app/clients.go
skmatti: We are modifying only the top level variable and this is used as read only by kubeclient for most part. I modified this to create separate configurations in each case.

pkg/controller/controller.go
skmatti: Not when ingresses are not enqueued(empty slice). I preserved this log and removed the other.

bowei: can we make Enqueue log for empty slice and delete the extra logs outside the call

freehan:
Discussed offline. @skmatti will rollback the protobuf PR for ingress 1.6 as for now.

pkg/controller/controller.go
skmatti: Removed these logs. Not sure adding this log inside enqueue would be any help as this backendconfig obj is not available inside that method.

skmatti:
> Discussed offline. @skmatti will rollback the protobuf PR for ingress 1.6 as for now.
> 
> Fix will be pushed to HEAD and cherrypick back to 1.6 1.7

Rolled back #849 in v1.6. I think this more cleaner as this change never got released in any GKE versions.

PR #1058, opened 2020-03-26
Handle health-check not found error gracefully

skmatti:
Return after deleting regional health-checks and not try to delete global health checks.

Results in the following error and the ILB gets re-queued,
```
I0325 09:31:12.089142       1 healthchecks.go:310] Deleting health check k8s1-0cf92254-test-sandbox-7ca3ebc1bfe2ffbd-svc-1-80-54141625
E0325 09:31:12.308709       1 taskqueue.go:62] Requeuing "test-sandbox-7ca3ebc1bfe2ffbd/ing1-1" due to error: error running backend garbage collection routine: error GCing regional Backends: googleapi: Error 404: The resource 'projects/gke-test-ingress-11/global/healthChecks/k8s1-0cf92254-test-sandbox-7ca3ebc1bfe2ffbd-svc-1-80-54141625' was not found, notFound (ingresses)
```

Also, ignore health-check not found errors.

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -297,21 +297,26 @@ func (h *HealthChecks) Delete(name string, scope meta.KeyType) error {
 		if err != nil {
 			return err
 		}
+		klog.V(2).Infof("Deleting regional health check %v", name)
 		// L7-ILB is the only use of regional right now
-		err = composite.DeleteHealthCheck(cloud, key, features.L7ILBVersions().HealthCheck)
-		// Ignore error if the deletion candidate is being used by another resource.
-		// In most of the cases, this is the associated backend resource itself.
-		if utils.IsHTTPErrorCode(err, http.StatusNotFound) || utils.IsInUsedByError(err) {
-			klog.V(4).Infof("DeleteRegionalHealthCheck(%s, _): %v, ignorable error", name, err)
-			return nil
+		if err = composite.DeleteHealthCheck(cloud, key, features.L7ILBVersions().HealthCheck); err != nil {
+			// Ignore error if the deletion candidate is being used by another resource.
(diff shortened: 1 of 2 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -297,6 +297,7 @@ func (h *HealthChecks) Delete(name string, scope meta.KeyType) error {
 		if err != nil {
 			return err
 		}
+		klog.V(2).Infof("Deleting regional health check %v", name)
 		// L7-ILB is the only use of regional right now
 		err = composite.DeleteHealthCheck(cloud, key, features.L7ILBVersions().HealthCheck)
```
freehan: how about this?
```
if err = composite.DeleteHealthCheck(cloud, key, features.L7ILBVersions().HealthCheck); err != nil {
    if utils.IsHTTPErrorCode(err, http.StatusNotFound) || utils.IsInUsedByError(err) {
         klog.V(4).Infof("DeleteRegionalHealthCheck(%s, _): %v, ignorable error", name, err)
         return nil
     }
     return err
}

return nil
```

pkg/healthchecks/healthchecks.go
```diff
@@ -305,13 +306,15 @@ func (h *HealthChecks) Delete(name string, scope meta.KeyType) error {
 			klog.V(4).Infof("DeleteRegionalHealthCheck(%s, _): %v, ignorable error", name, err)
 			return nil
 		}
+		return err
```
freehan: @spencerhance can you double check if we want to return here. It should right?

freehan:
LGTM
waiting for one more pair of eyes.

bowei:
Can you add some test cases? 

/lgtm 
/approve

Issue #1059, opened 2020-03-26
Ingress for Internal L7 LB creates backend-service with wrong scheme


Hermi999:
The Ingress for the Internal L7 LB creates a backend-service with the loadBalancingScheme: INTERNAL (see pkg/backends/backends.go).

According to the [docs](https://cloud.google.com/load-balancing/docs/l7-internal/setting-up-traffic-management) the scheme should be **INTERNAL_MANAGED** instead.

While creation of ILB works in this way, I can not update the Urlmap to use routeRules. Instead the following error returns:

_"Invalid value for field 'resource.defaultService':..... Advanced routing rules are not supported for scheme INTERNAL"_

rramkumar1:
@spencerhance

spencerhance:
@Hermi999 Thanks for submitting this.  There's a fix for this in https://github.com/kubernetes/ingress-gce/pull/1062

spencerhance:
This should be closed since it is has been added in v1.9.3

asgoel:
@spencerhance has 1.9.3 been published to the google container registry? I don't see it.

PR #1062, opened 2020-03-27
Update BackendConfig LoadBalancingScheme for L7-ILB to INTERNAL_MANAGED

spencerhance:
Both options work for basic routing, but we should migrate to INTERNAL_MANAGED to support advanced routing rules (https://github.com/kubernetes/ingress-gce/issues/1059)

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -87,7 +87,8 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 	}
 
 	if sp.L7ILBEnabled {
-		be.LoadBalancingScheme = "INTERNAL"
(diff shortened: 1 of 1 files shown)


pkg/backends/backends.go
```diff
@@ -87,7 +87,7 @@ func (b *Backends) Create(sp utils.ServicePort, hcLink string) (*composite.Backe
 	}
 
 	if sp.L7ILBEnabled {
-		be.LoadBalancingScheme = "INTERNAL"
+		be.LoadBalancingScheme = "INTERNAL_MANAGED"
```
rramkumar1, line 90: comment why this is needed

spencerhance: Done

PR #1064, opened 2020-03-30
Add tests for regional health check deletions

skmatti:
/assign @freehan @bowei

### DIFF
--- a/pkg/healthchecks/healthchecks_test.go
+++ b/pkg/healthchecks/healthchecks_test.go
@@ -36,7 +36,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/ingress-gce/pkg/annotations"
 	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
+	"k8s.io/ingress-gce/pkg/composite"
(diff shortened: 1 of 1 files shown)


pkg/healthchecks/healthchecks_test.go
```diff
@@ -269,6 +271,57 @@ func TestHTTP2HealthCheckDelete(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unexpected error creating composite key: %v", err)
+	}
+
+	// Verify that Health-check exists.
+	existingHc, err := composite.GetHealthCheck(fakeGCE, key, features.L7ILBVersions().HealthCheck)
+	if err != nil {
+		t.Fatalf("GetHealthCheck(%q) = %v, want nil", hc.Name, err)
+	}
+	if existingHc == nil {
```
bowei: combine with above:

```
if err != nil || existingHC == nil {
  t.Fatalf("GetHealthCheck(%q) = %v, %v, want nil", hc.Name, err, existingHC)
}
```

skmatti:
> would this test have caught the bug that you fixed?

Yes, the delete would return non-nil error trying to delete global health-check.

freehan (approved):
/lgtm
/approve

PR #1070, opened 2020-04-06
Update go dependencies to k8s v1.18.0

aledbf:
Note: not sure this can be done in several PRs due to the breaking changes in client-go.

- Go 1.13.9 is required because of `errors.Is` in API machinery
- Use [clientgofix](https://github.com/kubernetes-sigs/clientgofix) to update client-go calls
- Sync changes from kubernetes/kubernetes/pkg/controller/service/service_controller.go
- Simplify k8s dependency declaration to use new format (v1.18.0 instead of commits)


pkg/l4/l4controller_test.go
```diff
@@ -166,7 +166,7 @@ func TestProcessCreateOrUpdate(t *testing.T) {
 		t.Errorf("Failed to sync updated service %s, err %v", newSvc.Name, err)
 	}
 	// TODO remove this once https://github.com/kubernetes/client-go/issues/607 has been fixed.
-	validatePatchRequest(l4c.client, resetLBStatus, t)
+	//validatePatchRequest(l4c.client, resetLBStatus, t)
```
aledbf, line 169: With `validatePatchRequest` uncommented, the tests do not pass.

bowei:
@prameshj can you take a look at why the tests don't pass when the validate patch is run (see the last commit in this PR)

/assign @prameshj

pkg/l4/l4controller_test.go
prameshj: Can you share the logs from this particular test failure? The commented out function ensures that a patch request was indeed sent to reset the ingressIP in the service status.

aledbf: @prameshj the validatePatchRequest call fails because

> t.Errorf("Expected patch '%s', got '%s'", patchVal, string(patchAction.GetPatch()))

contains a CreateAction instead of PatchAction

aledbf:
ok, I removed the last commit and now is passing

prameshj:
> ok, I removed the last commit and now is passing

Thanks @aledbf 

/assign @bowei

pkg/l4/l4controller_test.go
```diff
@@ -201,8 +179,7 @@ func TestProcessDeletion(t *testing.T) {
 	if err != nil {
 		t.Errorf("Failed to sync updated service %s, err %v", newSvc.Name, err)
 	}
-	// TODO remove this once https://github.com/kubernetes/client-go/issues/607 has been fixed.
-	validatePatchRequest(l4c.client, resetLBStatus, t)
+
 	// List the service and ensure that it contains the finalizer as well as Status field.
```
prameshj, line 183: Would you mind changing the comment to "does not contain the finalizer or the status field"? Thanks for cleaning this up!

pkg/l4/l4controller_test.go
```diff
@@ -165,8 +144,7 @@ func TestProcessCreateOrUpdate(t *testing.T) {
...
```
prameshj, line 148: Same as below - comment should say "does not contain the finalizer or the status field"? Thanks for cleaning this up!

aledbf: done

pkg/l4/l4controller_test.go
aledbf: done

pkg/l4/l4controller_test.go
```diff
@@ -184,7 +161,7 @@ func TestProcessDeletion(t *testing.T) {
 	if err != nil {
 		t.Errorf("Failed to sync newly added service %s, err %v", newSvc.Name, err)
 	}
-	// List the service and ensure that it contains the finalizer as well as Status field.
+	// does not contain the finalizer or the status field?
```
prameshj, line 164: comment should read "List the service and ensure that it does not contain the finalizer or the status field."
Here and below.

Thanks!

prameshj:
l4/l4_controller_test.go changes look good to me. Since this change moved to clientgofix, we have been able to cleanup some temporary checks that were added. Thanks for cleaning this up @aledbf.

PR #1077, opened 2020-04-16
Allow local zone override for NonGCP NEG

freehan:
This is for NonGCP where node label of fault domains may not be present.


cxhiano:
LGTM

Tested in a GKE cluster

PR #1082, opened 2020-04-21
Update FrontendConfig v1beta1 API so that Spec and Status fields are optional

spencerhance:
This matches the same behavior of the BackendConfig CRD

This will also need to be cherry picked into v1.9

/assign @bowei


bowei:
Is this all of the generated files?

spencerhance:
@bowei Yes that's everything that the hack/ script updated

PR #1092, opened 2020-05-02
Add support for Port to BackendConfig HealthCheckConfig

spencerhance:
Wires in support to override the Port of a HealthCheck via BackendConfig.
This also forces the PortSpecification to be "USE_FIXED_PORT" regardless
of what type the health check is (e.g. neg, ilb etc.)

### DIFF
--- a/cmd/e2e-test/healthcheck_test.go
+++ b/cmd/e2e-test/healthcheck_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
@@ -45,7 +46,7 @@ func TestHealthCheck(t *testing.T) {
 		want     *backendconfig.HealthCheckConfig
 	}{
 		{
-			desc:     "override healthcheck",
+			desc:     "override healthcheck with IG",
 			beConfig: fuzz.NewBackendConfigBuilder("", "backendconfig-1").Build(),
 			want: &backendconfig.HealthCheckConfig{
 				CheckIntervalSec:   pint64(7),
@@ -53,6 +54,14 @@ func TestHealthCheck(t *testing.T) {
 				HealthyThreshold:   pint64(3),
 				UnhealthyThreshold: pint64(5),
 				RequestPath:        pstring("/my-path"),
+			},
+		},
+		{
+			desc:     "override healthcheck and port with NEG",
+			beConfig: fuzz.NewBackendConfigBuilder("", "backendconfig-1").Build(),
+			want: &backendconfig.HealthCheckConfig{
+				RequestPath: pstring("/my-path"),
+				Port:        pint64(8080), // Matches the targetPort
 			},
 		},
 	} {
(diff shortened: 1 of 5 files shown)


bowei:
/ok-to-test

this may not be sufficient -- you will need to change the type if it is a network endpoint (part of NEG)

spencerhance:
@bowei updated with USE_FIXED_PORT.  Tested in a sandbox but adding an e2e-test will take a few changes to the test setup.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1085,6 +1088,8 @@ func TestSyncServicePort(t *testing.T) {
 	wantCHC.HealthyThreshold = 1234
 	wantCHC.UnhealthyThreshold = 1234
 	wantCHC.TimeoutSec = 1234
+	wantCHC.HttpHealthCheck.Port = 1234
+	wantCHC.HttpHealthCheck.PortSpecification = "USE_FIXED_PORT"
```
bowei, line 1092: can you add a couple of test cases specifically around port? You added only to the "all fields test"

bowei:
can you beef up the commit message describing this change and what functionality it is adding (copy paste into github pr desc as well)

pkg/healthchecks/healthchecks_test.go
spencerhance: Done

spencerhance:
@bowei Please take another look when you get a chance

pkg/apis/backendconfig/v1/types.go
```diff
@@ -154,7 +154,10 @@ type HealthCheckConfig struct {
 	// Type is a health check parameter. See
 	// https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
 	Type *string `json:"type,omitempty"`
-	Port *int64  `json:"port,omitempty"`
+	// Port is a health check parameter. See
+	// https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
+	// If Port is used, portSpecification must also be updated
```
bowei, line 159: I don't think this updating portSpecification can be done by the user, so this doesn't make sense here.

pkg/healthchecks/healthcheck.go
```diff
@@ -239,7 +240,9 @@ func (hc *HealthCheck) updateFromBackendConfig(c *backendconfigv1.HealthCheckCon
 		hc.RequestPath = *c.RequestPath
 	}
 	if c.Port != nil {
-		klog.Warningf("Setting Port is not supported (healthcheck %q, backendconfig = %+v)", hc.Name, c)
+		hc.Port = *c.Port
+		// This override is necessary regardless of type
+		hc.PortSpecification = "USE_FIXED_PORT"
```
bowei, line 245: this works for both NEG and non-NEG?

spencerhance: Yes, it's technically not required if no other fields are set, but it makes sense to force it here to avoid accidental bugs

pkg/apis/backendconfig/v1/types.go
spencerhance: I added this to say that the controller needs to update portSpecification, not the user.  I can update this to make this more clear.

spencerhance:
@bowei Ready for another review

rramkumar1:
@spencerhance Is there an e2e test to go along with this?

spencerhance:
@rramkumar1 Added to the e2e test case

cmd/e2e-test/healthcheck_test.go
```diff
@@ -52,6 +52,7 @@ func TestHealthCheck(t *testing.T) {
 				TimeoutSec:         pint64(3),
 				HealthyThreshold:   pint64(3),
 				UnhealthyThreshold: pint64(5),
+				Port:               pint64(100),
```
rramkumar1, line 55: How would this pass? Wouldn't the health checks fail unless the application is serving something on this port?

spencerhance: The test passes when I run it locally, I believe we don't actually check for the health check to work, only that it matches the specified config

rramkumar1: But you call e2e.WaitForIngress() below which verifies the VIP returns a 200 right?

spencerhance: Yes.  I'm totally sure why it's passing.  I updated the test so that there's a second test for port that uses negs and updates the port on the service as well.  That should give a much clearer signal.

spencerhance: @rramkumar1 Updated the test again, should be more clear and is passing locally

rramkumar1:
@spencerhance I'll LGTM this for now, but I also think we need a test case that configures the health check on a different port from the application port. For example, the app is serving on 8080 but we point the health check to 8081. 

We can modify the test code to use the new zoneprinter app (gcr.io/google-samples/gcp-zoneprinter:0.2 ) which allows for customizing the HC port from args.

spencerhance:
@rramkumar1 Thanks!  That sgtm.  I submitted a cherry-pick to release-1.9 here https://github.com/kubernetes/ingress-gce/pull/1108

PR #1094, opened 2020-05-05
Add e2e tests for backendconfig access logging

skmatti:


cmd/e2e-test/logging_test.go
```diff
@@ -0,0 +1,219 @@
...
+	enabled bool
+	// log sampling rate, takes a value between 0.0 and 1.0.
+	sampleRate float64
+}
+
+func TestLogging(t *testing.T) {
+	t.Parallel()
+	const (
+		svcName = "service1"
+	)
```
rramkumar1, line 49: Get rid of the parentheses since you only have 1 variable?

skmatti: Done

PR #1096, opened 2020-05-06
Enable pruning of unknown fields for CRDs

### LINKED ISSUE #1055: OpenAPI v2 schema not published for BackendConfig
BackendConfig CRDs are non-structural because they do not have pruning enabled (note in the CRD that`preserveUnknownFields: true`). See https://v1-17.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2.

How to reproduce:

- curl `<apiserver-url>/openapi/v2`
- Look for the definition "com.google.cloud.v1beta1.BackendConfig". The entry just defines the type as "object".

Pruning can be manually enabled by setting `preserveUnknownFields: false` in the BackendConfig CRD. Once enabled, `/openapi/v2` returns the full schema.

I don't see any reason why pruning needs to be disabled for BackendConfig. It might be worth upgrading to the CRD v1 API at the same time?

skmatti:
/assign @rramkumar1 

fixes #1055

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -27,6 +27,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilpointer "k8s.io/utils/pointer"
 )
 
 const (
@@ -118,6 +119,9 @@ func crd(meta *CRDMeta) *apiextensionsv1beta1.CustomResourceDefinition {
 			klog.Errorf("Error adding simple validation for %v CRD: %v", meta.kind, err)
 		}
 		crd.Spec.Validation = validationSpec
+		// Drop unknown fields of a CRD resource. Note that this needs be specified
+		// in conjunction with validation spec.
+		crd.Spec.PreserveUnknownFields = utilpointer.BoolPtr(false)
 	}
 	return crd
 }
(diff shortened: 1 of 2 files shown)


pkg/crd/crd.go
```diff
@@ -110,6 +111,7 @@ func crd(meta *CRDMeta) *apiextensionsv1beta1.CustomResourceDefinition {
 				Singular:   meta.singular,
 				ShortNames: meta.shortNames,
 			},
+			PreserveUnknownFields: utilpointer.BoolPtr(false),
```
rramkumar1, line 114: Can you add a comment?

pkg/crd/crd.go
```diff
@@ -118,6 +119,9 @@ func crd(meta *CRDMeta) *apiextensionsv1beta1.CustomResourceDefinition {
 			klog.Errorf("Error adding simple validation for %v CRD: %v", meta.kind, err)
 		}
 		crd.Spec.Validation = validationSpec
+		// Drop unknown fields of a CRD resource. Note that this needs be specified
+		// in conjunction with validation spec.
+		crd.Spec.PreserveUnknownFields = utilpointer.BoolPtr(false)
```
bowei, line 124: does the validation spec need to be updated too?

bowei:
/lgtm
unless spec needs to be updated

pkg/crd/crd.go
skmatti: No, it does not

PR #1100, opened 2020-05-12
Fix TestILBError test for NEG defaulting

freehan:
cc @spencerhance


bowei:
can you update the commit message to describe what is being fixed

PR #1102, opened 2020-05-15
Wait for backendconfig CRD to be established

skmatti:
/assign @freehan 
cc: @spencerhance

### DIFF
--- a/pkg/e2e/helpers.go
+++ b/pkg/e2e/helpers.go
@@ -32,6 +32,9 @@ import (
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/networking/v1beta1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/e2e/helpers.go
```diff
@@ -631,3 +638,33 @@ func WaitConfigMapEvents(s *Sandbox, namespace, name string, msgs []string, time
...
+			}
+		}
+		klog.V(3).Infof("CRD %s is not established, retrying", backendConfigCRDName)
+		return false, nil
+	}
+	if waitErr := wait.Poll(backendConfigEnsurePollInterval, backendConfigEnsurePollTimeout, condition); waitErr != nil {
+		if waitErr == wait.ErrWaitTimeout {
+			return fmt.Errorf("error timeout waiting for CRD established, last seen error: %v", err)
+		}
+		return fmt.Errorf("error waiting for CRD established,: %v", waitErr)
```
rramkumar1, line 667: Do you really need to distinguish here? Why not just return the error?

skmatti: Fixed it. This would be helpful when we wait on an error. But, it is not the case here.

PR #1110, opened 2020-05-21
Script to run local e2e test

bowei:
hack/run-local-e2e.sh will setup your environment and
write some scripts to help run e2e tests from a locally built
binary.

Usage

$ ./hack/run-local-e2e.sh my-cluster
...
$ bash /tmp/run-glbc.sh # run proxy and controller

### DIFF
--- a/hack/run-local-glbc.sh
+++ b/hack/run-local-glbc.sh
@@ -1,0 +1,46 @@
+#/bin/bash
+
+# Run glbc. First run `run-local.sh` to set things up.
+#
+# Files touched: /tmp/kubectl-proxy.log /tmp/glbc.log
+
+GOOGLE_APPLICATION_CREDENTIALS="${HOME}/.config/gcloud/application_default_credentials.json"
+
+if [ ! -r ${GOOGLE_APPLICATION_CREDENTIALS} ]; then
+    echo "You must login your application default credentials"
+    echo "$ gcloud auth application-default login"
+    exit 1
+fi
+
+GCECONF=${GCECONF:-/tmp/gce.conf}
+GLBC=${GLBC:-./glbc}
+PORT=${PORT:-7127}
+V=${V:-3}
+
+echo "GCECONF=${GCECONF} GLBC=${GLBC} PORT=${PORT} V=${V}"
+
+if [ ! -x "${GLBC}" ]; then
+    echo "ERROR: No ${GLBC} executable found" >&2
+    exit 1
+fi
+
+echo "$(date) start" >> /tmp/kubectl-proxy.log
+kubectl proxy --port="${PORT}" \
+    >> /tmp/kubectl-proxy.log &
+
+PROXY_PID=$!
+cleanup() {
+    echo "Killing proxy (pid=${PROXY_PID})"
+    kill ${PROXY_PID}
+}
+trap cleanup EXIT
+
+kubectl apply -f docs/deploy/resources/default-http-backend.yaml
+
+sleep 2 # Wait for proxy to start up
+${GLBC} \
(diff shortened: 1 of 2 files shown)


bowei:
@rramkumar1 @spencerhance

hack/run-local-e2e.sh
```diff
@@ -0,0 +1,89 @@
...
+    net=$(echo ${net} | sed 's+.*networks/\([-a-z]*\).*$+\1+')
+    subnet=$(echo ${subnet} | sed 's+.*subnetworks/\([-a-z]*\)$+\1+')
+    project=$(echo ${selfLink} | sed 's+.*/projects/\([-a-z]*\)/.*+\1+')
+}
+
+parseInstance() {
+    local name=$1
+    local zone=$2
+    # Globals.
+    nodeTag=$(gcloud compute instances describe gke-test-default-pool-2b81ea3d-sw3l --zone us-west2-b --format='value(tags.items[0])')
```
rramkumar1, line 23: fix this? Seems to be hardcoded.

hack/run-local-e2e.sh
```diff
@@ -0,0 +1,89 @@
+#/bin/bash
```
rramkumar1, line 1: should this be called run-local.sh? It's just running the controller, not the e2e tests.

hack/run-local-e2e.sh
```diff
@@ -0,0 +1,89 @@
...
+}
+
+clusterName="$1"
+if [ -z ${clusterName} ]; then
+    echo "Usage: $0 CLUSTER_NAME"
+    exit 1
+fi
+
+fmt='value(networkConfig.network,networkConfig.subnetwork,zone,selfLink,name)'
+parseCluster $(gcloud container clusters list --format=${fmt} | grep "${clusterName}\$")
```
rramkumar1, line 33: if there two clusters with the same name in different zones would this work? I think the script should accept zone/region as an arg.

rramkumar1:
When trying this locally I get the following error:

```
deployment.extensions/l7-default-backend configured
service/default-http-backend configured
/tmp/run-glbc.sh: line 24: --apiserver-host=http://localhost:7127: No such file or directory
Killing proxy (pid=125670)
```

hack/run-local-e2e.sh
bowei: good point, fixing

hack/run-local-e2e.sh
bowei: fixed

hack/run-local-e2e.sh
bowei: added optional param

bowei:
fixed issues, got rid of unneeded script generation

hack/run-local-glbc.sh
```diff
@@ -0,0 +1,46 @@
...
+GOOGLE_APPLICATION_CREDENTIALS="${HOME}/.config/gcloud/application_default_credentials.json"
+
+if [ ! -r ${GOOGLE_APPLICATION_CREDENTIALS} ]; then
+    echo "You must login your application default credentials"
+    echo "$ gcloud auth application-default login"
+    exit 1
+fi
+
+GCECONF=${GCECONF:-/tmp/gce.conf}
+GLBC=${GLBC:-./glbc}
```
rramkumar1, line 16: this seems to assume the glbc executable is in this directory?  If you run the script as is, it won't work unless you build first and move the executable here.

bowei: you can do that by

GLBC=mypath/xxx hack/run-local-glbc.sh

PR #1118, opened 2020-05-26
Update CRD client(apiextensions) to v1

skmatti:
`apiextensions.k8s.io/v1beta1` is deprecated as of [k8s-v1.16](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning).

So, this works just fine with all 1.16+ GKE versions.

### DIFF
--- a/pkg/svcneg/svcneg.go
+++ b/pkg/svcneg/svcneg.go
@@ -24,13 +24,14 @@ import (
 func CRDMeta() *crd.CRDMeta {
 	meta := crd.NewCRDMeta(
 		apisneg.GroupName,
-		"v1beta1",
 		"ServiceNetworkEndpointGroup",
 		"ServiceNetworkEndpointGroupList",
 		"servicenetworkendpointgroup",
 		"servicenetworkendpointgroups",
+		[]*crd.Version{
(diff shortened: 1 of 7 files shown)


rramkumar1:
/ok-to-test

@skmatti Can you update the description with details on how this will roll out? Is there a specific cluster version? Will it just work?

skmatti:
> @skmatti Can you update the description with details on how this will roll out? Is there a specific cluster version? Will it just work?

All 1.16 versions are already using apiextensions/v1 on the server. So, when the client requests for v1beta1 resource in Ingress-GCE, it just swaps the version.
So, the rollout is seem less in all 1.16+ versions. But, this fails with any versions less than that.

rramkumar1:
Ok thanks, will take a look.

skmatti:
@rramkumar1 Can you take a look this PR?

pkg/svcneg/svcneg.go
```diff
@@ -28,9 +28,11 @@ func CRDMeta() *crd.CRDMeta {
 		"ServiceNetworkEndpointGroup",
```
rramkumar1, line 28: v1beta1 here needs to be removed?

skmatti: Good catch, done!

PR #1119, opened 2020-05-26
Force send Enable field for LogConfig

skmatti:
/assign @rramkumar1


pkg/composite/gen.go
```diff
@@ -3220,6 +3220,9 @@ func (backendService *BackendService) ToGA() (*compute.BackendService, error) {
 	if ga.Iap != nil {
 		ga.Iap.ForceSendFields = []string{"Enabled", "Oauth2ClientId", "Oauth2ClientSecret"}
 	}
+	if ga.LogConfig != nil {
```
rramkumar1, line 3223: I don't think you can modify gen.go? I think you have to modify the code which generates this file.

skmatti: Got it. Modified the source of the generate file and updated the generated code

PR #1123, opened 2020-06-01
Use patch for updating ingress/service finalizers

skmatti:
/assign @MrHohn @freehan

### DIFF
--- a/pkg/utils/common/finalizer.go
+++ b/pkg/utils/common/finalizer.go
@@ -15,11 +15,14 @@ package common
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/networking/v1beta1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/utils/common/finalizer.go
```diff
@@ -84,14 +85,35 @@ func EnsureDeleteFinalizer(ing *v1beta1.Ingress, ingClient client.IngressInterfa
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to Marshal newData for ingress %s: %v", ingKey, err)
+	}
+
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, v1beta1.Ingress{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create TwoWayMergePatch for ingress %s: %v", ingKey, err)
+	}
+
+	klog.V(2).Infof("Patch bytes for ingress %s: %s", ingKey, patchBytes)
```
MrHohn, line 113: Just double-checking - logging `patchBytes` at V2 won't be too noisy because it is only triggered when attaching/detaching finalizer (not very frequently)?

skmatti: Makes sense, bumped it to v3. Also, added patch logic for service as well

MrHohn:
LGTM retrospectively, thanks!

PR #1127, opened 2020-06-03
Use Patch instead of Update for k8s Service client

skmatti:
This replaces of all instances of update calls of k8s Service resource from this controller.

Also, this uses k8s service helper function to patch Service.

/assign @freehan @prameshj

### DIFF
--- a/pkg/cmconfig/controller.go
+++ b/pkg/cmconfig/controller.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/klog"
 )
 
@@ -65,7 +65,7 @@ func (c *ConfigMapConfigController) GetConfig() Config {
 }
 
 func (c *ConfigMapConfigController) updateASMReady(status string) {
-	patchBytes, err := utils.StrategicMergePatchBytes(v1.ConfigMap{Data: map[string]string{}},
+	patchBytes, err := patch.StrategicMergePatchBytes(v1.ConfigMap{Data: map[string]string{}},
 		v1.ConfigMap{Data: map[string]string{asmReady: status}}, v1.ConfigMap{})
 	if err != nil {
 		c.RecordEvent("Warning", "FailedToUpdateASMStatus", fmt.Sprintf("Failed to update ASM Status, failed to create patch for ASM ConfigMap, error: %s", err))
--- a/pkg/experimental/workload/daemon/daemon.go
+++ b/pkg/experimental/workload/daemon/daemon.go
@@ -31,7 +31,7 @@ import (
 	workloadv1a1 "k8s.io/ingress-gce/pkg/experimental/apis/workload/v1alpha1"
 	workloadclient "k8s.io/ingress-gce/pkg/experimental/workload/client/clientset/versioned"
 	daemonutils "k8s.io/ingress-gce/pkg/experimental/workload/daemon/utils"
-	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/klog"
 )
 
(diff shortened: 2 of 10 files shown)


pkg/utils/common/common.go
```diff
@@ -136,3 +139,21 @@ func patchIngress(ic client.IngressInterface, oldIngress, newIngress *v1beta1.In
...
+	_, err := svchelpers.PatchService(client, svc, newSvc)
+	return err
+}
+
+// PatchServiceLoadBalancerStatus patches the given service's LoadBalancerStatus
+// based on new service's load-balancer status.
+func PatchServiceLoadBalancerStatus(client coreclient.CoreV1Interface, svc *corev1.Service, newStatus corev1.LoadBalancerStatus) error {
+	newSvc := svc.DeepCopy()
+	newSvc.Status.LoadBalancer = newStatus
+	_, err := svchelpers.PatchService(client, svc, newSvc)
```
prameshj, line 157: Thanks for reusing this!

pkg/utils/common/common.go
```diff
@@ -136,3 +139,21 @@ func patchIngress(ic client.IngressInterface, oldIngress, newIngress *v1beta1.In
 	klog.V(4).Infof("Patch bytes for ingress %s: %s", ingKey, patchBytes)
 	return ic.Patch(context.TODO(), oldIngress.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}, "status")
 }
+
+// PatchServiceObjectMetadata patches the given service's metadata based on new
+// service metadata.
+func PatchServiceObjectMetadata(client coreclient.CoreV1Interface, svc *corev1.Service, newObjectMetadata metav1.ObjectMeta) error {
```
bowei, line 145: Should this be in patch.go?

pkg/utils/common/common.go
```diff
@@ -136,3 +139,21 @@ func patchIngress(ic client.IngressInterface, oldIngress, newIngress *v1beta1.In
...
+	newSvc := svc.DeepCopy()
+	newSvc.ObjectMeta = newObjectMetadata
+	_, err := svchelpers.PatchService(client, svc, newSvc)
+	return err
+}
+
+// PatchServiceLoadBalancerStatus patches the given service's LoadBalancerStatus
+// based on new service's load-balancer status.
+func PatchServiceLoadBalancerStatus(client coreclient.CoreV1Interface, svc *corev1.Service, newStatus corev1.LoadBalancerStatus) error {
```
bowei, line 154: patch.go?

pkg/utils/common/common.go
bowei: Is there a way to unit test this?

skmatti: Can't add it in patch.go because of import cycle. I will add unit tests.

skmatti: Maybe I should refactor finalizers into separate package

skmatti: Moved all the patch functions into patch.go and added unit tests

bowei:
Having both utils/common and utils/ doesn't make sense to me.
How about we just collapse common into utils/

skmatti:
> Having both utils/common and utils/ doesn't make sense to me.
> How about we just collapse common into utils/
Its because we will run into an import cycle if not. Both Namer and Finalizer workflows are in utils. I will move those out of utils and fix the import cycle in a followup PR.

skmatti:
> Having both utils/common and utils/ doesn't make sense to me.
> How about we just collapse common into utils/

I have used a different package `patch` for all patch methods. Currently, there are many import cycles if we merge `common` into utils.
Finalizer -> Utils -> Finalizer
Namer -> Utils -> Namer

We created `common` package so that these import cycles are resolved. Maybe we can clean that up later as it exists from longtime now.

bowei:
I think the only thing that really causes the issue is the import of utils/namer from utils/, which is just one file. If we move that out somewhere else, does that resolve the issue/

skmatti:
> I think the only thing that really causes the issue is the import of utils/namer from utils/, which is just one file. If we move that out somewhere else, does that resolve the issue/

This is the line that references namer (https://github.com/kubernetes/ingress-gce/blob/6f08bb6d47ff9325e57f2800ca487133ce3d92df/pkg/utils/serviceport.go#L29).
If I move this to say `serviceport` package, the import cycle would be 
serviceport -> utils/namer -> utils -> serviceport
because of this(https://github.com/kubernetes/ingress-gce/blob/6f08bb6d47ff9325e57f2800ca487133ce3d92df/pkg/utils/utils.go#L434) 

Same with finalizer workflow because of this (https://github.com/kubernetes/ingress-gce/blob/6f08bb6d47ff9325e57f2800ca487133ce3d92df/pkg/utils/utils.go#L465)

skmatti:
@bowei Can you review this again?. Maybe if its better, I can isolate the refactoring into a separate PR so that the change is small.

freehan (commented):
can you rebase?

freehan (approved):
/lgtm
/approve

PR #1128, opened 2020-06-03
Add rbac permissions for networking.gke.io/frontendconfigs

spencerhance:
This will need to be cherry-picked into release-1.9 asap

Tested the rbac config in a sandbox


bowei:
is there an e2e test? How do we catch this next time?

/lgtm
/approve

spencerhance:
Thanks @bowei!  Any e2e test using a new CRD will catch this in a real cluster.  The issue is that you won't see this if you run the binary locally (--running-in-cluster=false).  

As far as catching this earlier, I'm not sure what else to do besides potentially writing a quick guide for adding new CRDs.

PR #1129, opened 2020-06-03
Cherry-pick #1128 to release-1.9 [Add rbac permissions for networking.gke.io/frontendconfigs]

spencerhance:
#1128 

/assign @bowei


spencerhance:
We should tag a new version after this

PR #1131, opened 2020-06-04
Condense logging_test.go to reuse one Ingress by applying transitions in logical order

rramkumar1:
Now, instead of creating three load balancers (one per test case) we can reuse the same load balancer and invoke each test case via transitions. Note that there is no loss in test coverage here since we are still applying the same exact settings.

cc @skmatti

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/1131)

### DIFF
--- a/cmd/e2e-test/logging_test.go
+++ b/cmd/e2e-test/logging_test.go
@@ -25,12 +25,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/ingress-gce/pkg/annotations"
-	backendconfig "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
 	"k8s.io/ingress-gce/pkg/e2e"
 	"k8s.io/ingress-gce/pkg/e2e/adapter"
 	"k8s.io/ingress-gce/pkg/fuzz"
 	"k8s.io/ingress-gce/pkg/fuzz/features"
-	"k8s.io/ingress-gce/pkg/test"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/common"
 )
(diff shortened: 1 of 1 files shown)


skmatti:
This will take longer time to run the test as the transitions are sequential

rramkumar1:
That's a good point though in practice, I don't think the individual test cases run truly parallel either.  The Go runtime can only run a certain number of threads truly in parallel (I think this is equal to the value of GOMAXPROCS). Given the number of test cases we have, there is going to be a lot of context switching happening.

I'll run this though to see how much time it takes vs what we have currently.

bowei:
I think the only thing you need to do is to get rid of the t.Parallel

rramkumar1:
@skmatti Ran this and it took a total of 7.5 minutes. Without this change, took about 1.5-2 minutes less. Seems small enough difference to me?

rramkumar1:
@bowei Why would removing t.Parallel help?

bowei:
What is the impact on the length of the test?

rramkumar1:
@bowei I ran this individual test and it took a couple minutes longer to finish. There might be a negligible difference if we run the whole suite of cases.

skmatti:
LGTM

cmd/e2e-test/logging_test.go
```diff
@@ -44,157 +42,132 @@ type logging struct {
...
 					return false, nil
 				}
-				if err := verifyLogging(t, gclb, s.Namespace, svcName, tc.transition); err != nil {
+				if err := verifyLogging(t, gclb, s.Namespace, svcName, transition); err != nil {
 					return false, nil
 				}
 				return true, nil
 			}); waitErr != nil {
-				t.Errorf("Timeout waiting for BackendConfig logging transition propagation to GCLB, last seen error: %v", err)
+				t.Errorf("Timeout waiting for BackendConfig logging transition propagation to GCLB, last seen error: %v", waitErr)
```
skmatti, line 153: This needs be `err` instead of `waitErr`. `waitErr` does not return useful error here so we log the last seen error

cmd/e2e-test/logging_test.go
```diff
@@ -44,157 +42,132 @@ type logging struct {
...
+			t.Logf("Verifying transition of %+v", transition)
 			// Wait for transition settings to be propagated.
 			if waitErr := wait.Poll(transitionPollInterval, transitionPollTimeout, func() (bool, error) {
 				gclb, err = fuzz.GCLBForVIP(context.Background(), Framework.Cloud, params)
 				if err != nil {
 					t.Logf("Failed to GCP resources for LB with IP = %q: %v", vip, err)
 					return false, nil
 				}
-				if err := verifyLogging(t, gclb, s.Namespace, svcName, tc.transition); err != nil {
+				if err := verifyLogging(t, gclb, s.Namespace, svcName, transition); err != nil {
```
skmatti, line 148: Replace `err :=` with `err =`

PR #1132, opened 2020-06-04
Refactor affinity test for L7 XLB to use transitions for all test cases

rramkumar1:
Now, instead of creating four load balancers (one per test case) we can reuse the same load balancer and invoke each test case via transitions. Note that there is no loss in test coverage here since we are still applying the same exact settings.

---
This change is [Reviewable](https://reviewable.io/reviews/kubernetes/ingress-gce/1132)


rramkumar1:
cc @skmatti

PR #1125, opened 2020-06-02
Replace Ingress resource Update calls with Patch

skmatti:

PR #1134, opened 2020-06-05
Remove unnecessary `update` permissions

skmatti:
This removes `update` permissions for Pod, Service and Ingress resources.
All the `update` calls are removed in Ingress-GCE in
kubernetes/ingress-gce#1125 and kubernetes/ingress-gce#1127.

/assign @rramkumar1

### DIFF
--- a/docs/deploy/resources/rbac.yaml
+++ b/docs/deploy/resources/rbac.yaml
@@ -37,34 +37,25 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["get"]
+  verbs: ["get", "list"]
 - apiGroups: [""]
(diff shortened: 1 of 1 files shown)


docs/deploy/resources/rbac.yaml
```diff
@@ -37,34 +37,25 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["get"]
+  verbs: ["get", "list"]
```
liggitt, line 40: adding list here should not have been required. reverting the List() call added in https://github.com/kubernetes/ingress-gce/pull/1101/files#r456811105 in https://github.com/kubernetes/ingress-gce/pull/1188

PR #1137, opened 2020-06-06
Add events for major lifecycle events

bowei:
{Add,Remove}Nodes from InstanceGroup
SyncIngress, TranslateIngress, IPChanged, GarbageCollection

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -17,7 +17,20 @@ limitations under the License.
 package events
 
 import (
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
+)
+
+const (
+	AddNodes    = "IngressGCE_AddNodes"
+	RemoveNodes = "IngressGCE_RemoveNodes"
+
+	SyncIngress       = "Sync"
+	TranslateIngress  = "Translate"
+	IPChanged         = "IPChanged"
+	GarbageCollection = "GarbageCollection"
+
+	SyncService = "Sync"
 )
 
 type RecorderProducer interface {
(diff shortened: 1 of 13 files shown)


bowei:
/assign @rramkumar1 
cc: @spencerhance

pkg/events/events_test.go
```diff
@@ -0,0 +1,9 @@
+package events
+
+import (
+	"testing"
+)
+
+func TestTruncatedStringList(t *testing.T) {
```
rramkumar1, line 7: address TODO

pkg/events/events.go
```diff
@@ -17,9 +17,22 @@ limitations under the License.
 package events
 
 import (
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
 )
 
+const (
+	AddNodes    = "IngressGCE_AddNodes"
+	RemoveNodes = "IngressGCE_RemoveNodes"
```
rramkumar1, line 26: Why do we need the IngressGCE prefix? Is it to distinguish from other Ingress controllers for the same type of action?

pkg/events/events.go
```diff
@@ -17,9 +17,22 @@ limitations under the License.
...
+
+	SyncIngress       = "Sync"
+	TranslateIngress  = "Translate"
+	IPChanged         = "IPChanged"
+	GarbageCollection = "GarbageCollection"
+
+	SyncService = "Sync"
```
rramkumar1, line 33: what is the difference between this and SyncIngress?

bowei: the value is not the same, but this identifies it so we can easily change it if needed

bowei:
@rramkumar1 PTAL

PR #1144, opened 2020-06-09
Ensure ingress cleanup when finalizer exists

skmatti:


pkg/controller/controller.go
```diff
@@ -173,6 +173,13 @@ func NewLoadBalancerController(
 					lbc.ingQueue.Enqueue(cur)
 					return
 				}
+				// Ingress needs to be enqueued if a ingress finalizer exists.
+				// Note that the GC workflow invoked for this ingress.
+				if common.HasFinalizer(curIng.ObjectMeta) {
+					klog.V(4).Infof("Ingress %v has a finalizer, enqueuing", common.NamespacedName(curIng))
```
bowei, line 179: V(2)

bowei: shouldn't the logic be like this?

if curIng has finalizer => enqueue
if class changed ! new.class == glbc && old.class == glbc => enqueue // Is this condition even needed given the finalizer stuff?

(can you fix the V(4) above as well?)

skmatti: Fixed the log level. 

Yes, we don't need to check if the old ingress was a GLBC ingress.

pkg/controller/controller.go
```diff
@@ -166,20 +166,22 @@ func NewLoadBalancerController(
...
-				// If ingress was GLBC Ingress, we need to track ingress class change
-				// and run GC to delete LB resources.
-				if utils.IsGLBCIngress(oldIng) {
-					klog.V(4).Infof("Ingress %v class was changed, enqueuing", common.NamespacedName(curIng))
+				// Ingress needs to be enqueued if a ingress finalizer exists.
+				// An existing finalizer means that
+				// 1. Ingress update for class change.
+				// 2. Ingress cleanup failed and re-queued.
+				// 3. Finalizer remove failed and re-queued.
+				if common.HasFinalizer(curIng.ObjectMeta) {
```
freehan, line 174: Is it possible that a super old version of ingress controller (without finalizer) gets directly updated to this version?

Would it basically skip all syncs for existing ingress?

skmatti: This will be released to 1.17 only and both 1.15, 1.16 versions have finalizers enabled. So, the above case may be highly unlikely

PR #1145, opened 2020-06-09
Wait for NEGs deletion based on deletion option

freehan:
Fix a e2e test failure when NEG defaulting is enabled.


freehan:
cc @skmatti

PR #1147, opened 2020-06-09
Skip garbage collection for invalid frontend resource names

skmatti:
This fixes a bug where an ingress with name that has dot character in it. Since, dot is an invalid character for GCE resource name, the load balancer will not configured. But the ingress failed to get deleted as we proactively try to delete the GCE resources when they were not created.

This ensures that those ingresses with invalid names are deleted and not re-queued by controller.

### DIFF
--- a/pkg/loadbalancers/l7s.go
+++ b/pkg/loadbalancers/l7s.go
@@ -67,6 +67,12 @@ func (l *L7s) Ensure(ri *L7RuntimeInfo) (*L7, error) {
 		ingress:     *ri.Ingress,
 	}
 
+	if !lb.namer.IsValidLoadBalancer() {
+		err := fmt.Errorf("invalid loadbalancer name %s, the resource name must comply with RFC1035 (https://www.ietf.org/rfc/rfc1035.txt)", lb.namer.LoadBalancer())
+		klog.Error(err)
+		return nil, err
+	}
+
 	if err := lb.edgeHop(); err != nil {
 		return nil, fmt.Errorf("loadbalancer %v does not exist: %v", lb.String(), err)
 	}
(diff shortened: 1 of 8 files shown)


pkg/utils/namer/frontendnamer.go
```diff
@@ -108,6 +108,12 @@ func (ln *V1IngressFrontendNamer) LoadBalancer() LoadBalancerName {
 	return ln.lbName
 }
 
+// IsValidLoadBalancer implements IngressFrontendNamer.
+func (ln *V1IngressFrontendNamer) IsValidLoadBalancer() bool {
+	// Dot is a valid character in k8s resource name but not in GCE.
+	return !strings.Contains(ln.lbName.String(), ".")
```
skmatti, line 114: We could essentially other validations as well. Probably like leading `-` characters or length of the string (<=63). But, these are improbable at the moment. Should we add those other cases as well?.

pkg/loadbalancers/l7s.go
```diff
@@ -67,6 +67,11 @@ func (l *L7s) Ensure(ri *L7RuntimeInfo) (*L7, error) {
 		ingress:     *ri.Ingress,
 	}
 
+	if !lb.namer.IsValidLoadBalancer() {
+		klog.Errorf("Invalid loadbalancer name %s, the resource name must comply with RFC1035 (https://www.ietf.org/rfc/rfc1035.txt)", lb.namer.LoadBalancer())
```
bowei, line 71: err := fmt.Errorf("invalid ...
klog.Error(err)
return nil, err

bowei:
minor fix

skmatti:
> minor fix
Thanks, fixed

skmatti:
@bowei The issue is addressed on this one. Can you revisit this?. Thanks

pkg/utils/namer/utils.go
```diff
@@ -15,12 +15,15 @@ package namer
...
 import (
 	"fmt"
+	"regexp"
 
 	"k8s.io/api/networking/v1beta1"
 	"k8s.io/ingress-gce/pkg/utils/common"
 	"k8s.io/klog"
 )
 
+const gceResourceNamePattern = "(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)"
```
MrHohn, line 25: Is there a public GCP doc regarding what pattern is allowed?

MrHohn (approved):
/lgtm
/approve

pkg/utils/namer/utils.go
skmatti: There is a reference to DNS compatible names: https://cloud.google.com/apis/design/resource_names#full_resource_name

PR #1151, opened 2020-06-10
Reserve Static IP if a static IP name is provided and it doesn't exist

vinicyusmacedo:
This PR creates Static IPs if a static IP name is provided but it wasn't reserved by the user. 

Currently, GLBC will check if the provided static IP name translates to an existing reserved static IP. If it doesn't, then it's up to the user to reserve it. 

However, there might be a case where you expect GLBC to reserve it for you with the name provided (e.g. if you are deploying a new Ingress for an app using Kustomize, you may not want to execute extra steps only to reserve the static IP, and you may also want GLBC to manage it if you delete an ingress).

### DIFF
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -88,6 +88,7 @@ Events:
 
 GCE has a concept of [ephemeral](https://cloud.google.com/compute/docs/instances-and-network#ephemeraladdress) and [static](https://cloud.google.com/compute/docs/instances-and-network#reservedaddress) IPs. A production website would always want a static IP, which ephemeral IPs are cheaper (both in terms of quota and cost), and are therefore better suited for experimentation.
 * Creating a HTTP Ingress (i.e an Ingress without a TLS section) allocates an ephemeral IP, because we don't believe HTTP is the right way to deploy an app.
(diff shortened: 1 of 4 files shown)


spencerhance:
Hm, this was the previous behavior, but we recently reverted that in #1080.  @prameshj

prameshj:
Correct, this was reverted since it would cause an unexpected ingress outage, if the static ip specified eventually got created by the user. The next ingress sync would switch to using that IP without much warning.

vinicyusmacedo:
@spencerhance ops, sorry, I didn't see that.

Can this patch be considered in any way? It would be great if GLBC could manage IPs with a user-provided address name.

prameshj:
Since we just reverted the behavior, adding this patch would be a regression. Can we introduce a new annotation for this? Feel free to send out a PR for that change. It would make the address management code a bit more tricky, we need to consider all the cases to make sure IPs aren't leaked. @rramkumar1 @bowei

rramkumar1:
@vinicyusmacedo Echoing the previous statement, feel free to send a new PR but note that it would have a high bar for submission since this area of the code is tricky.

PR #1152, opened 2020-06-10
Fix ensure logging workflow for nil/empty sampling rate

skmatti:
This fixes an edge case where logging is enabled and sampling rate is set to 0.0

Currently, the value is treated as empty value and not sent to the server in which case the existing sampling rate is retained.

### DIFF
--- a/pkg/backends/features/logging.go
+++ b/pkg/backends/features/logging.go
@@ -32,32 +32,48 @@ func EnsureLogging(sp utils.ServicePort, be *composite.BackendService) bool {
 		return false
 	}
 	svcKey := fmt.Sprintf("%s/%s", sp.ID.Service.Namespace, sp.ID.Service.Name)
-	expectedLogConfig := expectedBackendServiceLogConfig(sp)
-	if !expectedLogConfig.Enable && !be.LogConfig.Enable {
-		klog.V(3).Infof("Logging continues to stay disabled for service %s, skipping update", svcKey)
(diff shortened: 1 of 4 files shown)


pkg/backends/features/logging.go
```diff
@@ -32,32 +32,47 @@ func EnsureLogging(sp utils.ServicePort, be *composite.BackendService) bool {
...
-	if !sp.BackendConfig.Spec.Logging.Enable || sp.BackendConfig.Spec.Logging.SampleRate == nil {
-		return logConfig
+	if !sp.BackendConfig.Spec.Logging.Enable {
+		return
+	}
+	// Existing sample rate is retained if not specified.
+	if sp.BackendConfig.Spec.Logging.SampleRate == nil {
+		// Update sample rate to 1.0 if no existing sample rate or is 0.0
+		if be.LogConfig.SampleRate == 0.0 {
+			klog.V(2).Infof("Sample rate neither specified nor preexists, defaulting to 1.0")
```
MrHohn, line 72: Is it possible to include some identifiers here so it will be clear which ingress hits this?

MrHohn (approved):
Thanks Satish, LGTM with a minor comment on the log.
/approve

skmatti:
> Thanks Satish, LGTM with a minor comment on the log.
> /approve

Thanks Zihong, added service-port info to the log.

MrHohn (approved):
/lgtm

PR #1155, opened 2020-06-12
add logging in NEG syncer

freehan:


pkg/neg/syncers/utils.go
```diff
@@ -244,6 +245,9 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
 			return nil, nil, err
 		}
 	}
+	if !foundMatchingPort {
```
skmatti, line 248: Can we add a log at line 209 as well?

pkg/neg/syncers/transaction.go
```diff
@@ -155,11 +155,15 @@ func (s *transactionSyncer) syncInternal() error {
...
 	}
+	s.logStats(currentMap, "current NEG endpoints")
+
 	// Merge the current state from cloud with the transaction table together
 	// The combined state represents the eventual result when all transactions completed
 	mergeTransactionIntoZoneEndpointMap(currentMap, s.transactions)
+	s.logStats(currentMap, "after in-progress operations have completed, NEG endpoints")
 
 	targetMap, endpointPodMap, err := s.endpointsCalculator.CalculateEndpoints(ep.(*apiv1.Endpoints), currentMap)
+	s.logStats(currentMap, "desired NEG endpoints")
```
skmatti, line 166: targetMap?

pkg/neg/syncers/utils.go
prameshj: +1 to that. Also, we can move this log line to line 194? At that point, we know foundMatchingPort is false right? Or if we want this approach, we need to reset foundMatchingPort to false in line 250.

pkg/neg/syncers/transaction.go
```diff
@@ -185,7 +189,8 @@ func (s *transactionSyncer) syncInternal() error {
 		klog.V(4).Infof("No endpoint change for %s/%s, skip syncing NEG. ", s.Namespace, s.Name)
 		return nil
 	}
-
+	s.logEndpoints(addEndpoints, "adding endpoint")
+	s.logEndpoints(removeEndpoints, "removing endpoint")
```
prameshj, line 193: This would log endpoints twice in case of VM_IP NEGs, see line 172. Maybe we can remove line 172 and log include s.NegType in this function.

pkg/neg/syncers/transaction.go
```diff
@@ -391,3 +396,21 @@ func mergeTransactionIntoZoneEndpointMap(endpointMap map[string]negtypes.Network
...
+		stats = append(stats, fmt.Sprintf("%d endpoints in zone %q", endpointSet.Len(), zone))
+	}
+	klog.V(3).Infof("For NEG %q, %s: %s.", s.negName, desc, strings.Join(stats, ","))
+}
+
+// logEndpoints logs individual endpoint in the input endpointMap
+func (s *transactionSyncer) logEndpoints(endpointMap map[string]negtypes.NetworkEndpointSet, desc string) {
+	for zone, endpointSet := range endpointMap {
+		for _, endpoint := range endpointSet.List() {
+			klog.V(3).Infof("For NEG %q, %s: %+v to zone %q", s.negName, desc, endpoint, zone)
```
prameshj, line 413: Can we log the entire map instead of each endpoint? The map key will have the zone and the value will be the list of endpoints?

pkg/neg/syncers/utils.go
prameshj: Please disregard my comment about moving the log to a different line. It makes sense to keep it the way you have it - we only want this log if no endpoints had the matching port.

freehan: Added another log line when the output is empty.

pkg/neg/syncers/transaction.go
```diff
@@ -391,3 +392,17 @@ func mergeTransactionIntoZoneEndpointMap(endpointMap map[string]negtypes.Network
...
+	stats := []string{}
+	for zone, endpointSet := range endpointMap {
+		stats = append(stats, fmt.Sprintf("%d endpoints in zone %q", endpointSet.Len(), zone))
+	}
+	klog.V(3).Infof("For NEG %q, %s: %s.", s.negName, desc, strings.Join(stats, ","))
+}
+
+// logEndpoints logs individual endpoint in the input endpointMap
+func (s *transactionSyncer) logEndpoints(endpointMap map[string]negtypes.NetworkEndpointSet, desc string) {
+	klog.V(3).Infof("For NEG %q, %s: %+v", endpointMap)
```
prameshj, line 407: Let's include s.negName and s.NegType

freehan: The log line when the syncerInternal starts includes everything. I feel like you can reference it by searching the neg name?

prameshj: true.. this log line needs more params though. Maybe you meant to include "desc" and s.negName? It might be nice to see the negType here too, but we can lookup from name as well.

PR #1160, opened 2020-06-16
Update generated code for frontendconfig CRD

skmatti:
Makes the client methods compatible with k8s v1.18


skmatti:
@spencerhance

PR #1162, opened 2020-06-18
Avoid unnecessary backend service update in NEG linker

skmatti:


pkg/backends/neg_linker.go
```diff
@@ -88,13 +91,14 @@ func (l *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
...
 	if !oldBackends.Equal(newBackends) {
+		klog.V(2).Infof("Backends changed for service port %s, removing: %s and adding: %s", sp.ID, oldBackends.Difference(newBackends), newBackends.Difference(oldBackends))
 		backendService.Backends = targetBackends
 		return composite.UpdateBackendService(l.cloud, key, backendService)
 	}
 	return nil
 }
 
-func getBackendsForNEGs(negs []*composite.NetworkEndpointGroup) []*composite.Backend {
+func getBackendsForNEGs(negs []*composite.NetworkEndpointGroup, version meta.Version) []*composite.Backend {
```
freehan, line 101: version not used?

skmatti: fixed

skmatti:
> I do not think this PR is needed.

The issue still exists as the utility function does not consider beta features
https://github.com/kubernetes/ingress-gce/blob/15f450fd9437b546f4ff3f32a76a582998a4a721/pkg/utils/serviceport.go#L61-L66

freehan:
LGTM from my side.

I will let @prameshj to take a final look

prameshj:
The change looks good to me. But I am not sure how we are avoiding an unnecessary update here. I see version lookup unified and extra logging added.

freehan:
There is an api version section in the self link. 

When you get backend-service with beta API, the selfLink of its backends all contain beta. Same for NEG selfLink. So for example, when you have beta feature enabled on backend-service. NegLinker will use beta API to get backend-service and use alpha or v1 API to get NEG. Then the NEG selfLink is always different because the API version section is different. Hence the negLinker will always try to update the backend service on every sync even though it is a noop.

bowei:
Stuff like this *needs to go into the commit message*. 

Otherwise people reading the code later will have no idea that this is why the change was made.

> There is an api version section in the self link.
> 
> When you get backend-service with beta API, the selfLink of its backends all contain beta. Same for NEG selfLink. So for example, when you have beta feature enabled on backend-service. NegLinker will use beta API to get backend-service and use alpha or v1 API to get NEG. Then the NEG selfLink is always different because the API version section is different. Hence the negLinker will always try to update the backend service on every sync even though it is a noop.

PR #1165, opened 2020-06-22
Add support for GCE path prefix rewrites

### LINKED ISSUE #109: rewrite
Hi, I've been struggling to set a rewrite rule for my ingress (GCP Kubernetes with a Deployment, Service NodePort and Ingress). What I'm trying to get is:
* api.example.com/  →  backend/api/
* www.example.com/ → backend/www/

So that my server can server requests whether they start with /api/ or /www/, without a care about the subdomains.

I'm creating this issue because I have a question regarding your comment in https://github.com/kubernetes/ingress-gce#paths
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: nginxtest-ingress
spec:
  rules:
  - http:
      paths:
      - path: /hostname
        backend:
          serviceName: nginxtest
          servicePort: 80
```

```
Note what just happened, the endpoint exposes /hostname, and the loadbalancer forwarded the entire matching url to the endpoint. This means if you had '/foo' in the Ingress and tried accessing /hostname, your endpoint would've received /foo/hostname and not known how to route it.
```
I wish that was true... for me, loadbalancerip/foo would be the only queryable url, while everything else returns 404 Not Found errors. The Ingress only forward a request if it matches the path, that's it.

Anyways, if you have a suggestion regarding my problem, I'd really appreciate. Couldn't figure out how to get it working with annotations [rewrite-url or app-root](https://github.com/kubernetes/ingress-nginx/blob/master/docs/examples/rewrite/README.md)

sanderploegsma:
This adds support for setting up a GCE path prefix rewrite as described in [Setting up URL rewrite for external HTTP(S) load balancers](https://cloud.google.com/load-balancing/docs/https/setting-up-url-rewrite). It introduces a new annotation `ingress.gcp.kubernetes.io/path-prefix-rewrite` that is set on the underlying URL map for all paths configured through the Ingress.

Usage is similar to the NGINX Controller. For example, the following configuration will route requests from `/foo` to `http://foo:80/` and from `/bar` to `http://bar:80/`:

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: test-ingress
  annotations:
    ingress.gcp.kubernetes.io/path-prefix-rewrite: /
spec:
  rules:
  - http:
      paths:
      - path: /foo
        backend:
          serviceName: foo
          servicePort: 80
      - path: /bar
        backend:
          serviceName: bar
          servicePort: 80
```

This should fix #109

### DIFF
--- a/cmd/e2e-test/path_prefix_rewrite_test.go
+++ b/cmd/e2e-test/path_prefix_rewrite_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/e2e/adapter"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+)
+
+func TestPathPrefixRewrite(t *testing.T) {
+	t.Parallel()
+
+	port80 := intstr.FromInt(80)
+	Framework.RunWithSandbox("with path prefix rewrite", t, func(t *testing.T, s *e2e.Sandbox) {
+		t.Parallel()
+
+		ctx := context.Background()
+
+		_, err := e2e.CreateEchoServiceWithOS(s, "service-1", nil, e2e.Linux)
+		if err != nil {
+			t.Fatalf("error creating echo service: %v", err)
+		}
+		t.Logf("Echo service created (%s/%s)", s.Namespace, "service-1")
+
+		testIng := fuzz.NewIngressBuilder(s.Namespace, "ingress-1", "").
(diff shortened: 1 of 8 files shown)


bowei:
Is your example a use case? I think for Ingress GA, "/foo" and "/foo/" mean the same thing

/ok-to-test

sanderploegsma:
@bowei I’m not sure what you mean, the example shows how the path `/foo` is routed to the `foo` backend and rewritten to `/`. 

Sorry if that wasn’t clear.

rramkumar1:
@bowei Does this make more sense in FrontendConfig?

dassaf1:
Hi @sanderploegsma,

What if I want to rewrite only a prefix and get the remainder of the uri sent to the backend? is this achievable with this PR? if not, could be great if you could include it.

To clarify, please see Ambassador's rewrite example:

image

Image source link:
https://www.getambassador.io/docs/latest/topics/using/rewrites/

sanderploegsma:
> Hi @sanderploegsma,
> 
> 
> 
> What if I want to rewrite only a prefix and get the remainder of the uri sent to the backend? is this achievable with this PR? if not, could be great if you could include it.
> 
> 
> 
> To clarify, please see Ambassador's rewrite example:
> 
> 
> 
> image
> 
> 
> 
> Image source link:
> 
> https://www.getambassador.io/docs/latest/topics/using/rewrites/

Well, this PR only maps an Ingress annotation to the corresponding GCE feature. However, if you follow the GCE documentation link in the PR description they do describe a similar scenario: https://cloud.google.com/load-balancing/docs/https/setting-up-url-rewrite

So yes, this PR should be able to help you achieve your example, too.

mingsterism:
how can we use this in our existing kubernetes cluster. 
according to the docs, there's a default mapping https://github.com/kubernetes/ingress-gce#gke-version-mapping for GKE versions to GLBC versions. 

what if we want to use this latest version after merge? 
is there a way to specify which version of GLBC we are using?

spencerhance:
Hi @mingsterism,

Unfortunately there is no way to specify which glbc version you want to use on GKE.  However, this is possible by running a [self-managed controller](https://github.com/kubernetes/ingress-gce/tree/master/docs/deploy/gke) yourself in your cluster.

PR #1168, opened 2020-06-24
Initial Refactor of TargetProxy to Translator Package

spencerhance:
Refactors TargetHttp/Https Proxy to Translator package, including SSL Policy.

### DIFF
--- a/pkg/translator/translator.go
+++ b/pkg/translator/translator.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/api/networking/v1beta1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	frontendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/frontendconfig/v1beta1"
+	"k8s.io/ingress-gce/pkg/flags"
 
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/utils"
@@ -37,6 +39,9 @@ import (
 type Env struct {
 	// Ing is the Ingress we are translating.
 	Ing *v1beta1.Ingress
+	// TODO(shance): this should be a map, similar to SecretsMap
+	// FrontendConfig is the frontendconfig associated with the Ingress
+	FrontendConfig *frontendconfigv1beta1.FrontendConfig
 	// SecretsMap contains a mapping from Secret name to the actual resource.
 	// It is assumed that the map contains resources from a single namespace.
 	// This is the same namespace as the Ingress namespace.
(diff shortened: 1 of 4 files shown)


spencerhance:
@rramkumar1 I added another commit to handle sslPolicy.  It gets a bit tricky since we the proxy only holds a string, but we use a pointer to hold all three states.  We currently don't update the policy if the user deletes the FC or sets the policy to nil.

I'm not sure if there's a better way to handle this, or if the original commit makes more sense for now until we have a general solution for backendconfig/frontendconfig.  Ideally in the future we should delete the policy if the FC is deleted.

pkg/translator/translator.go
```diff
@@ -37,6 +39,8 @@ import (
 type Env struct {
 	// Ing is the Ingress we are translating.
 	Ing *v1beta1.Ingress
+	// FrontendConfig is the frontendconfig associated with the Ingress
+	FrontendConfig *frontendconfigv1beta1.FrontendConfig
```
rramkumar1, line 43: Add TODO that this should eventually be a map of FrontendConfigs (similar to Secret)

pkg/translator/translator.go
```diff
@@ -222,3 +226,80 @@ func (t *Translator) ToCompositeForwardingRule(env *Env, protocol namer.NamerPro
...
+
+	return proxy, sslPolicySet, nil
+}
+
+// getSslPolicyLink returns the ref to the ssl policy that is described by the
+// frontend config.  Since Ssl Policy is a *string, there are three possible I/O situations
+// 1) policy is nil -> this returns nil
+// 2) policy is an empty string -> this returns an empty string
+// 3) policy is non-empty -> this constructs the resource path and returns it
+func getSslPolicyLink(env *Env) (*string, error) {
```
rramkumar1, line 283: remove get

pkg/translator/translator_test.go
```diff
@@ -383,3 +386,183 @@ func TestToForwardingRule(t *testing.T) {
...
+		},
+		{
+			desc: "frontendconfig with empty string ssl policy",
+			fc:   &frontendconfigv1beta1.FrontendConfig{Spec: frontendconfigv1beta1.FrontendConfigSpec{SslPolicy: utils.NewStringPointer("")}},
+			want: utils.NewStringPointer(""),
+		},
+	}
+
+	for _, tc := range testCases {
+		env := &Env{FrontendConfig: tc.fc}
```
rramkumar1, line 558: use t.Run and pass the description there

pkg/translator/translator.go
spencerhance: Done

pkg/translator/translator.go
spencerhance: Done

pkg/translator/translator_test.go
spencerhance: Done

pkg/translator/translator_test.go
```diff
@@ -383,3 +386,185 @@ func TestToForwardingRule(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			env := &Env{FrontendConfig: tc.fc}
+			result, err := sslPolicyLink(env)
+			if err != nil {
+				t.Errorf("desc: %q, l7.getSslPolicyLink() = %v, want nil", tc.desc, err)
+			}
+
+			if !reflect.DeepEqual(result, tc.want) {
+				t.Errorf("desc: %q, l7.getSslPolicyLink() = %v, want %+v", tc.desc, result, tc.want)
```
rramkumar1, line 566: don't need this anymore

spencerhance: Whoops.  Fixed that and the rest of the error

PR #1172, opened 2020-07-06
Create corresponding Neg CRs when NEG CRD is enabled

swetharepakula:
Controller accepts custom names and manager will create CRs per NEG when NEG CRD is enabled.

This PR is based on PR #1166

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -17,22 +17,29 @@ limitations under the License.
 package neg
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"sync"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	negv1beta1 "k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/neg/readiness"
 	negsyncer "k8s.io/ingress-gce/pkg/neg/syncers"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
 	"k8s.io/klog"
+	utilpointer "k8s.io/utils/pointer"
 )
 
 type serviceKey struct {
@@ -55,6 +62,7 @@ type syncerManager struct {
 	podLister      cache.Indexer
 	serviceLister  cache.Indexer
 	endpointLister cache.Indexer
+	svcNegLister   cache.Indexer
 
 	// TODO: lock per service instead of global lock
 	mu sync.Mutex
@@ -67,9 +75,11 @@ type syncerManager struct {
 	syncerMap map[negtypes.NegSyncerKey]negtypes.NegSyncer
 	// reflector handles NEG readiness gate and conditions for pods in NEG.
 	reflector readiness.Reflector
+	//svcNegClient handles lifecycle operations for NEG CRs
+	svcNegClient svcnegclient.Interface
 }
 
-func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer, recorder record.EventRecorder, cloud negtypes.NetworkEndpointGroupCloud, zoneGetter negtypes.ZoneGetter, podLister, serviceLister, endpointLister, nodeLister cache.Indexer) *syncerManager {
+func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer, recorder record.EventRecorder, cloud negtypes.NetworkEndpointGroupCloud, zoneGetter negtypes.ZoneGetter, svcNegClient svcnegclient.Interface, podLister, serviceLister, endpointLister, nodeLister, svcNegLister cache.Indexer) *syncerManager {
 	return &syncerManager{
 		namer:          namer,
 		recorder:       recorder,
@@ -79,8 +89,10 @@ func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer, recorder record.
 		podLister:      podLister,
 		serviceLister:  serviceLister,
 		endpointLister: endpointLister,
+		svcNegLister:   svcNegLister,
 		svcPortMap:     make(map[serviceKey]negtypes.PortInfoMap),
 		syncerMap:      make(map[negtypes.NegSyncerKey]negtypes.NegSyncer),
+		svcNegClient:   svcNegClient,
 	}
 }
 
@@ -105,14 +117,19 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 	manager.svcPortMap[key] = newPorts
 	klog.V(3).Infof("EnsureSyncer %v/%v: syncing %v ports, removing %v ports, adding %v ports", namespace, name, newPorts, removes, adds)
 
+	errList := []error{}
 	for svcPort, portInfo := range removes {
 		syncer, ok := manager.syncerMap[getSyncerKey(namespace, name, svcPort, portInfo)]
 		if ok {
 			syncer.Stop()
 		}
+
+		err := manager.ensureDeleteSvcNegCR(namespace, portInfo.NegName)
+		if err != nil {
+			errList = append(errList, err)
+		}
 	}
 
-	errList := []error{}
 	// Ensure a syncer is running for each port that is being added.
 	for svcPort, portInfo := range adds {
 		syncerKey := getSyncerKey(namespace, name, svcPort, portInfo)
@@ -121,6 +138,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 			// determine the implementation that calculates NEG endpoints on each sync.
 			epc := negsyncer.GetEndpointsCalculator(manager.nodeLister, manager.podLister, manager.zoneGetter,
 				syncerKey, portInfo.RandomizeEndpoints)
+
 			syncer = negsyncer.NewTransactionSyncer(
 				syncerKey,
 				portInfo.NegName,
@@ -135,6 +153,10 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				epc,
 			)
 			manager.syncerMap[syncerKey] = syncer
+		}
+
+		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
+			errList = append(errList, err)
 		}
 
 		if syncer.IsStopped() {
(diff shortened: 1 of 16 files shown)


pkg/neg/manager.go
```diff
@@ -260,6 +275,32 @@ func (manager *syncerManager) ReadinessGateEnabled(syncerKey negtypes.NegSyncerK
 	return false
 }
 
+func (manager *syncerManager) ensureDeleteNegServiceCRs(namespace, name string) error {
```
freehan, line 278: add comment

pkg/neg/manager.go
```diff
@@ -329,6 +370,129 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
 	return manager.cloud.DeleteNetworkEndpointGroup(name, zone, meta.VersionGA)
 }
 
+func (manager *syncerManager) ensureCreateNegCR(svcKey serviceKey, portInfo negtypes.PortInfo) error {
```
freehan, line 373: add comment

pkg/neg/manager.go
```diff
@@ -329,6 +370,129 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+			},
+		}
+		_, err = manager.negClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(svcKey.namespace).Create(context.Background(), &negCR, metav1.CreateOptions{})
+		return err
+	}
+
+	return nil
+}
+
+func (manager *syncerManager) checkForExistingNegCR(svc *v1.Service, portInfo negtypes.PortInfo) (bool, error) {
```
freehan, line 435: add comment

pkg/neg/manager.go
```diff
@@ -329,6 +370,129 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+		klog.Errorf("Failed to retrieve service %s from store: %v", svcKey.Key(), err)
+	}
+
+	if !exists {
+		return fmt.Errorf("Service not found")
+	}
+
+	service := obj.(*v1.Service)
+
+	createCR, err := manager.checkForExistingNegCR(service, portInfo)
```
freehan, line 389: checkForExistingNegCR -> needToCreate ?

pkg/neg/manager.go
freehan, line 373: ensureSvcNegCR?

pkg/neg/manager.go
```diff
@@ -329,6 +370,129 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+	if err != nil {
+		return fmt.Errorf("Failed checking for existing CR: %s", err)
+	}
+	if createCR {
```
freehan, line 393:
```
if !needToCreate {
  continue
}

```

pkg/neg/manager.go
freehan: I feel like there is some problem with the overall flow of this function. It is not idempotent. 

It should probably 1. generate the expected NEG CR 2. get existing NEG CR 3. compare and see if need to create or update. 4. create/update if needed.     


This is to ensure if v1 of the controller thinks the CR should look like v1 CR, if we ever make changes to our expected NEG CR, then when v2 controller runs, it will convert all NEG CR into v2

pkg/neg/manager.go
freehan, line 278: ensureDeleteSvcNegCRs

pkg/neg/manager.go
```diff
@@ -121,6 +131,11 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 			// determine the implementation that calculates NEG endpoints on each sync.
 			epc := negsyncer.GetEndpointsCalculator(manager.nodeLister, manager.podLister, manager.zoneGetter,
 				syncerKey, portInfo.RandomizeEndpoints)
+
+			if err := manager.ensureCreateNegCR(key, portInfo); err != nil {
```
freehan, line 135: you need to `ensureDeleteSvcNegCR` for `removes`

pkg/neg/manager.go
```diff
@@ -260,6 +285,31 @@ func (manager *syncerManager) ReadinessGateEnabled(syncerKey negtypes.NegSyncerK
...
+	}
+	if !exists {
+		return nil
+	}
+	neg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
+
+	deletionTS := metav1.Now()
+	if neg.GetDeletionTimestamp().IsZero() {
+		neg.SetDeletionTimestamp(&deletionTS)
+		_, err := manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Update(context.Background(), neg, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
freehan: I think you can use the client to delete the object and let API server applying the deletionTimestamp. 


Just need to make sure the API call is non blocking. (I think it is so by default. Otherwise, need to set a deletion option)

pkg/neg/manager.go
```diff
@@ -329,6 +379,99 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+		existingLabels := negCR.GetLabels()
+		for key, value := range labels {
+			if existingVal := existingLabels[key]; existingVal != value {
+				return fmt.Errorf("Neg already exists with name %s but label %s has value %s instead of %s. Delete previous neg before creating this configuration", portInfo.NegName, key, existingVal, value)
+			}
+		}
+		if !reflect.DeepEqual(newCR.ObjectMeta, negCR.ObjectMeta) {
+			// Neg CR exists, has correct labels and names, but Object Meta needs to be updated
+			newCR.Status = negCR.Status
+			_, err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(svcKey.namespace).Update(context.Background(), &newCR, metav1.UpdateOptions{})
```
freehan, line 461: I would recommend organizing the above checks into `needToUpdate`

pkg/neg/manager.go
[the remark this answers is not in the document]
swetharepakula: Verified that the API call is non blocking and switched to directly deleting the negs.

pkg/neg/manager.go
swetharepakula: I was able to clean up the checks a little bit. Since the last check is the only one that doesn't cause a return, I wasn't able to replicate the suggested pattern. Please let me know if the checks still seem difficult to follow.

pkg/neg/manager.go
```diff
@@ -147,19 +169,23 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
...
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
 	key := getServiceKey(namespace, name)
+	var errList []error
 	if ports, ok := manager.svcPortMap[key]; ok {
 		for svcPort, portInfo := range ports {
 			if syncer, ok := manager.syncerMap[getSyncerKey(namespace, name, svcPort, portInfo)]; ok {
 				syncer.Stop()
 			}
+			if err := manager.ensureDeleteSvcNegCR(namespace, portInfo.NegName); err != nil {
```
freehan, line 182: Not sure if it is worth doing this?
- Listing all SvcNegCR using label selector from the cache
- Delete all remaining SvcNegCR. 

If NEG GC can handle this catch all case, then probably no need to handle this case.

pkg/neg/manager.go
```diff
@@ -329,6 +378,99 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+
+	if !exists {
+		return fmt.Errorf("Service not found")
+	}
+
+	service := obj.(*v1.Service)
+
+	gvk := schema.GroupVersionKind{Version: "v1", Kind: "Service"}
+	ownerReference := metav1.NewControllerRef(service, gvk)
+	*ownerReference.BlockOwnerDeletion = false
```
freehan, line 402:
```
ownerReference.BlockOwnerDeletion  = utilpointer.BoolPtr(false)
```

pkg/neg/manager.go
```diff
@@ -329,6 +378,99 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+	now := metav1.Now()
+	//TODO: Add finalizer after Neg CRD Garbage Collection is implemented.
+	newCR := negv1beta1.ServiceNetworkEndpointGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            portInfo.NegName,
+			Namespace:       svcKey.namespace,
+			OwnerReferences: []metav1.OwnerReference{*ownerReference},
+			Labels:          labels,
+		},
+		Status: negv1beta1.ServiceNetworkEndpointGroupStatus{
```
freehan, line 418: leave the Status empty on creation. Let syncer populate it.

pkg/neg/manager.go
```diff
@@ -329,6 +378,99 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+		_, err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(svcKey.namespace).Create(context.Background(), &newCR, metav1.CreateOptions{})
+		return err
+	}
+
+	if !negCR.GetDeletionTimestamp().IsZero() {
+		// Allow GC to complete
+		return fmt.Errorf("Neg CR already exists with this name and is marked for deletion. Allow GC to complete before recreating.")
+	}
+
+	newCR.ResourceVersion = negCR.ResourceVersion
```
freehan, line 454: How about this?

```
needUpdate := false
needUpdate |= ensureNegCRLabels(negCR, expectedLabels)
needUpdate |= ensureNegObjecReference(negCR, expectedObjectReference)

if needUpdate {
   _, err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(svcKey.namespace).Update(context.Background(), &negCR, metav1.UpdateOptions{})
}

return err



func ensureNegCRLabels(negCR, expectedLabels) bool{
  if !reflect.DeepEqual(negCR.Labels, expectedLabels) {
      negCR.Labels = expectedLabels.
      return true
  }
}



```

swetharepakula: Updated using this pattern, but also checked the values of the 3 labels we care about because that ensures that the CR we are updating matches the configuration we expect.

pkg/neg/manager.go
swetharepakula: Removed status

pkg/neg/manager.go
swetharepakula: Removed deletion of neg crs when stoping syncers. Deletion will only occur if service exists and there is a change in ports desired.

pkg/neg/manager.go
```diff
@@ -121,6 +138,11 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 			// determine the implementation that calculates NEG endpoints on each sync.
 			epc := negsyncer.GetEndpointsCalculator(manager.nodeLister, manager.podLister, manager.zoneGetter,
 				syncerKey, portInfo.RandomizeEndpoints)
+
+			if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
```
freehan, line 142: move this to outside of `if !ok`

pkg/neg/manager.go
```diff
@@ -329,6 +373,99 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
...
+	}
+
+	if !reflect.DeepEqual(existingLabels, labels) {
+		negCR.Labels = labels
+		return true, nil
+	}
+	return false, nil
+}
+
+func ensureNegCRObjectMeta(negCR *negv1beta1.ServiceNetworkEndpointGroup, expectedObjMeta metav1.ObjectMeta) bool {
```
freehan, line 460: objectMeta is too big. There is too many fields. It is very likely that some fields are different than expected. That may be due to versioning and defaulting

So just check the objectReference should be good enough

PR #1173, opened 2020-07-06
Update k8s-cloud-provider to 1.13.0

spencerhance:
This is necessary for internal static IP

/assign @freehan


freehan:
I would recommend split the changes into 2 commits: 1. with your manual change 2. auto code vendoring

spencerhance:
@freehan Good call.  I split it into 2 commits.

spencerhance:
@freehan can we merge this?

PR #1174, opened 2020-07-06
Implement support for Internal/Regional Static IP for L7-ILB

spencerhance:
This is implemented via an annotation "kubernetes.io/ingress.regional-static-ip-name".  This uses the same prefix as the existing annotation, "global-static-ip-name."

This PR also depends on the update of k8s-cloud-provider in #1173


spencerhance:
Rebased since #1173 has merged.  The tests should pass now.

pkg/annotations/ingress.go
```diff
@@ -131,8 +140,35 @@ func (ing *Ingress) UseNamedTLS() string {
...
-	val, ok := ing.v[StaticIPNameKey]
+func (ing *Ingress) StaticIPName() (string, error) {
+	if !flags.F.EnableL7Ilb {
+		return ing.GlobalStaticIPName(), nil
+	}
+
+	globalIp := ing.GlobalStaticIPName()
+	regionalIp := ing.RegionalStaticIPName()
+
+	if globalIp != "" && regionalIp != "" {
```
freehan, line 151: I guess you need to validate if this is a L7 ILB then just return regionalStaticIP and ignore GlobalStaticIP?

vice versa for L7 XLB?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -163,10 +163,16 @@ func (l *L7) getEffectiveIP() (string, bool, error) {
...
+			return "", false, err
+		}
+
 		// Existing static IPs allocated to forwarding rules will get orphaned
 		// till the Ingress is torn down.
-		if ip, err := l.cloud.GetGlobalAddress(l.runtimeInfo.StaticIPName); err != nil || ip == nil {
-			return "", false, fmt.Errorf("the given static IP name %v doesn't translate to an existing global static IP.",
+		// TODO(shance): Replace version
+		if ip, err := composite.GetAddress(l.cloud, key, meta.VersionGA); err != nil || ip == nil {
+			return "", false, fmt.Errorf("The given static IP name %v doesn't translate to an existing static IP.",
```
freehan, line 175: lower case for the initial character for error

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -163,10 +163,16 @@ func (l *L7) getEffectiveIP() (string, bool, error) {
 	// TODO: Handle the last case better.
 
 	if l.runtimeInfo.StaticIPName != "" {
+		key, err := l.CreateKey(l.runtimeInfo.StaticIPName)
```
freehan, line 166: don't you need to handle the L7 ILB for this case as well?

freehan: oh I see. `CreateKey` handles it already.

pkg/loadbalancers/addresses.go
```diff
@@ -50,10 +53,21 @@ func (l *L7) checkStaticIP() (err error) {
...
+	key, err := l.CreateKey(managedStaticIPName)
+	if err != nil {
+		return err
+	}
+
+	ip, _ := composite.GetAddress(l.cloud, key, meta.VersionGA)
 	if ip == nil {
 		klog.V(3).Infof("Creating static ip %v", managedStaticIPName)
-		err = l.cloud.ReserveGlobalAddress(&compute.Address{Name: managedStaticIPName, Address: l.fw.IPAddress})
+		address := &composite.Address{Name: managedStaticIPName, Address: l.fw.IPAddress, Version: meta.VersionGA}
```
freehan, line 64: Probably worth having a `l.newStaticAddress()` function to output a composite.Address based on the L7 config.

pkg/annotations/ingress.go
spencerhance: True, but I wanted to have a check for if the user specified both by accident

pkg/loadbalancers/addresses.go
spencerhance: Hm, I'm hesitant to split this out into a function since it's fairly short and we only use it once

freehan: having a separate function makes this a bit more self contained. Which means the special handling for ILB is invisible to the main flow. 
```
if isInternal {
	// Used for L7 ILB
	address.AddressType = "INTERNAL"
}
```
Next time when changes are needed, it can just piggy back to the new function and do not touch the main flow.

spencerhance: Okay done, also added a unit test

pkg/loadbalancers/forwarding_rules.go
spencerhance: done

pkg/loadbalancers/addresses.go
```diff
@@ -20,16 +20,19 @@ import (
...
+	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 )
 
-// checkStaticIP reserves a static IP allocated to the Forwarding Rule.
+// checkStaticIP reserves a regional or global static IP allocated to the Forwarding Rule.
 func (l *L7) checkStaticIP() (err error) {
```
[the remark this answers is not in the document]
spencerhance: https://github.com/kubernetes/ingress-gce/blob/d6080dd6a61a4304baa473c06ef124aede51735f/pkg/loadbalancers/loadbalancers_test.go#L204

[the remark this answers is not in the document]
spencerhance: The tests for checkStaticIP() are sort of included in loadbalancers_test.go right now since it also creates/gets resources from the GCE API, and piggybacks off of those mocks

PR #1176, opened 2020-07-07
Compute number of L4 ILBs in error and success state

skmatti:


pkg/loadbalancers/l4.go
```diff
@@ -165,6 +165,14 @@ func (l *L4) GetFRName() string {
 // EnsureInternalLoadBalancer ensures that all GCE resources for the given loadbalancer service have
 // been created. It returns a LoadBalancerStatus with the updated ForwardingRule IP address.
 func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service) (*corev1.LoadBalancerStatus, error) {
+	var serviceState metrics.L4ILBServiceState
+	// Mark the service InSuccess state as false to begin with.
+	// This will be updated to true if the VIP is configured successfully.
+	serviceState.InSuccess = false
+	defer func() {
```
bowei, line 172: We should be very careful what uses defer, as it's behavior is somewhat non-obvious. Is there a good reason to do this using defer?

skmatti: This is to update service state for metrics if this function returns an error. I added defer for code compactness. 
Would it be better to add this line at each return statement?.

pkg/loadbalancers/l4.go
```diff
@@ -147,7 +147,7 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) error {
 		klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
 	}
 	if retErr == nil {
-		klog.V(6).Infof("Service %s deleted, removing its state from metrics cache", l.NamespacedName.String())
+		klog.V(6).Infof("Internal Loadbalancer for Service %s deleted, removing its state from metrics cache", l.NamespacedName.String())
```
prameshj, line 150: Include "L4" in this log?

pkg/metrics/metrics.go
```diff
@@ -316,18 +316,26 @@ func (im *ControllerMetrics) computeL4ILBMetrics() map[feature]int {
...
+		l4ILBInError:      0,
 	}
 
 	for key, state := range im.l4ILBServiceMap {
-		klog.V(6).Infof("ILB Service %s has EnabledGlobalAccess: %t, EnabledCustomSubnet: %t", key, state.EnabledGlobalAccess, state.EnabledCustomSubnet)
-		counts[l4ILBService] += 1
+		klog.V(6).Infof("ILB Service %s has EnabledGlobalAccess: %t, EnabledCustomSubnet: %t, InSuccess: %t", key, state.EnabledGlobalAccess, state.EnabledCustomSubnet, state.InSuccess)
+		counts[l4ILBService]++
+		if !state.InSuccess {
+			counts[l4ILBInError]++
```
prameshj, line 329: It would be useful to expand this in the future to indicate the reason of the error. So we could have "l4ILBErrorFirewall", "l4ILBErrorBackendService" etc.

skmatti: We can add new metric features and modify service state in order to achieve that.

pkg/loadbalancers/l4.go
```diff
@@ -234,18 +242,17 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 		return nil, err
 	}
 
-	var serviceState metrics.L4ILBServiceState
+	serviceState.InSuccess = true
```
prameshj, line 245: Can we also set the service state to false in https://github.com/kubernetes/ingress-gce/blob/c96b1d8c07002afad5264be4d633e36591620f59/pkg/l4/l4controller.go#L161

It would be worth setting it in each of those error cases. Maybe we pass in a pointer to ServiceState in EnsureInternalLoadBalancer and always set it in the processServiceCreateOrUpdate function?

PR #1187, opened 2020-07-18
syncer writes neg desc on initialization

swetharepakula:
- updates status on NEG CRs when neg crd is enabled
  - errors during sync and initializations are exposed in NEG CR

### DIFF
--- a/pkg/apis/svcneg/v1beta1/types.go
+++ b/pkg/apis/svcneg/v1beta1/types.go
@@ -50,7 +50,7 @@ type ServiceNetworkEndpointGroupStatus struct {
 
 	// Last time the NEG syncer syncs associated NEGs.
 	// +optional
-	LastSyncTime metav1.Time `json:"lastTransitionTime,omitempty"`
+	LastSyncTime metav1.Time `json:"lastSyncTime,omitempty"`
 }
 
 // NegObjectReference is the object reference to the NEG resource in GCE
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -17,6 +17,7 @@ limitations under the License.
 package syncers
 
 import (
+	"context"
 	"sync"
 
 	"fmt"
@@ -25,14 +26,19 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	negv1beta1 "k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/neg/readiness"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/apis/core"
 )
 
 type transactionSyncer struct {
@@ -58,6 +64,7 @@ type transactionSyncer struct {
 	serviceLister       cache.Indexer
 	endpointLister      cache.Indexer
 	nodeLister          cache.Indexer
+	svcNegLister        cache.Indexer
 	recorder            record.EventRecorder
 	cloud               negtypes.NetworkEndpointGroupCloud
 	zoneGetter          negtypes.ZoneGetter
@@ -68,9 +75,15 @@ type transactionSyncer struct {
 
 	// reflector handles NEG readiness gate and conditions for pods in NEG.
 	reflector readiness.Reflector
+
+	//kubeSystemUID used to populate Cluster UID on Neg Description when using NEG CRD
+	kubeSystemUID string
+
+	//svcNegClient used to update status on corresponding NEG CRs when not nil
+	svcNegClient svcnegclient.Interface
 }
 
-func NewTransactionSyncer(negSyncerKey negtypes.NegSyncerKey, networkEndpointGroupName string, recorder record.EventRecorder, cloud negtypes.NetworkEndpointGroupCloud, zoneGetter negtypes.ZoneGetter, podLister cache.Indexer, serviceLister cache.Indexer, endpointLister cache.Indexer, nodeLister cache.Indexer, reflector readiness.Reflector, epc negtypes.NetworkEndpointsCalculator) negtypes.NegSyncer {
+func NewTransactionSyncer(negSyncerKey negtypes.NegSyncerKey, networkEndpointGroupName string, recorder record.EventRecorder, cloud negtypes.NetworkEndpointGroupCloud, zoneGetter negtypes.ZoneGetter, podLister cache.Indexer, serviceLister cache.Indexer, endpointLister cache.Indexer, nodeLister cache.Indexer, svcNegLister cache.Indexer, reflector readiness.Reflector, epc negtypes.NetworkEndpointsCalculator, kubeSystemUID string, svcNegClient svcnegclient.Interface) negtypes.NegSyncer {
 	// TransactionSyncer implements the syncer core
 	ts := &transactionSyncer{
 		NegSyncerKey:        negSyncerKey,
@@ -81,11 +94,14 @@ func NewTransactionSyncer(negSyncerKey negtypes.NegSyncerKey, networkEndpointGro
 		podLister:           podLister,
 		serviceLister:       serviceLister,
 		endpointLister:      endpointLister,
+		svcNegLister:        svcNegLister,
 		recorder:            recorder,
 		cloud:               cloud,
 		zoneGetter:          zoneGetter,
 		endpointsCalculator: epc,
 		reflector:           reflector,
+		kubeSystemUID:       kubeSystemUID,
+		svcNegClient:        svcNegClient,
 	}
 	// Syncer implements life cycle logic
 	syncer := newSyncer(negSyncerKey, networkEndpointGroupName, serviceLister, recorder, ts)
@@ -121,10 +137,16 @@ func (s *transactionSyncer) sync() error {
 func (s *transactionSyncer) syncInternal() error {
 	s.syncLock.Lock()
 	defer s.syncLock.Unlock()
+	// NOTE: Error will be used to update the status on corresponding Neg CR if Neg CRD is enabled
+	// Please reuse and set err before returning
+	var err error
+	defer s.updateStatus(err)
+
 	if s.needInit {
 		if err := s.ensureNetworkEndpointGroups(); err != nil {
 			return err
 		}
+		s.needInit = false
 	}
 
 	if s.syncer.IsStopped() || s.syncer.IsShuttingDown() {
@@ -164,7 +186,8 @@ func (s *transactionSyncer) syncInternal() error {
 
 	targetMap, endpointPodMap, err := s.endpointsCalculator.CalculateEndpoints(ep.(*apiv1.Endpoints), currentMap)
 	if err != nil {
-		return fmt.Errorf("endpoints calculation error in mode %q, err: %v", s.endpointsCalculator.Mode(), err)
+		err = fmt.Errorf("endpoints calculation error in mode %q, err: %v", s.endpointsCalculator.Mode(), err)
+		return err
 	}
 	s.logStats(targetMap, "desired NEG endpoints")
 
@@ -191,7 +214,10 @@ func (s *transactionSyncer) syncInternal() error {
 	}
 	s.logEndpoints(addEndpoints, "adding endpoint")
 	s.logEndpoints(removeEndpoints, "removing endpoint")
-	return s.syncNetworkEndpoints(addEndpoints, removeEndpoints)
+
+	// set err instead of returning directly so that synced condition on neg crd is properly updated in defer
+	err = s.syncNetworkEndpoints(addEndpoints, removeEndpoints)
+	return err
 }
 
 // ensureNetworkEndpointGroups ensures NEGs are created and configured correctly in the corresponding zones.
@@ -203,11 +229,33 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	}
 
 	var errList []error
+	var negObjRefs []negv1beta1.NegObjectReference
 	for _, zone := range zones {
-		if err := ensureNetworkEndpointGroup(s.Namespace, s.Name, s.negName, zone, s.NegSyncerKey.String(), s.NegSyncerKey.NegType, s.cloud, s.serviceLister, s.recorder, s.NegSyncerKey.GetAPIVersion()); err != nil {
+		var negObj negv1beta1.NegObjectReference
+		negObj, err = ensureNetworkEndpointGroup(
+			s.Namespace,
+			s.Name,
+			s.negName,
+			zone,
+			s.NegSyncerKey.String(),
+			s.kubeSystemUID,
+			fmt.Sprint(s.NegSyncerKey.PortTuple.Port),
+			s.NegSyncerKey.NegType,
+			s.cloud,
+			s.serviceLister,
+			s.recorder,
+			s.NegSyncerKey.GetAPIVersion(),
+		)
+		if err != nil {
 			errList = append(errList, err)
 		}
+
+		if s.svcNegClient != nil && err == nil {
+			negObjRefs = append(negObjRefs, negObj)
+		}
 	}
+
+	s.updateInitStatus(negObjRefs, errList)
 	return utilerrors.NewAggregate(errList)
 }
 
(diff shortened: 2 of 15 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -128,17 +130,37 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
 			}
 		}
+	} else {
+		negObj = negv1beta1.NegObjectReference{
+			Id:                  neg.Id,
+			SelfLink:            neg.SelfLink,
+			NetworkEndpointType: negv1beta1.NetworkEndpointType(neg.NetworkEndpointType),
+		}
+
+		if neg.Description != "" {
```
freehan, line 152: add a comment:
skip the following validation if the neg object does not have Description

pkg/neg/syncers/utils.go
```diff
@@ -150,23 +172,32 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
+			return negv1beta1.NegObjectReference{}, err
 		}
 		if recorder != nil && serviceLister != nil {
 			if svc := getService(serviceLister, svcNamespace, svcName); svc != nil {
 				recorder.Eventf(svc, apiv1.EventTypeNormal, "Create", "Created NEG %q for %s in %q.", negName, negServicePortName, zone)
 			}
 		}
 	}
-	return nil
+	return negObj, nil
```
freehan, line 200: probably need to GetNetworkEndpointGroup after Cretion to get the selflink?

pkg/neg/syncers/utils.go
```diff
@@ -150,23 +172,32 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
 			Version:             version,
 			Name:                negName,
 			NetworkEndpointType: string(networkEndpointType),
 			Network:             cloud.NetworkURL(),
 			Subnetwork:          subnetwork,
+			Description:         desc.String(),
 		}, zone)
 		if err != nil {
-			return err
+			return negv1beta1.NegObjectReference{}, err
```
freehan, line 192: return negRef, err

pkg/neg/syncers/utils.go
```diff
@@ -150,23 +172,32 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 		default:
 			subnetwork = cloud.SubnetworkURL()
 		}
+
+		desc := utils.NegDescription{}
```
freehan, line 176:
```
desc := ""
if enableNegCRD {
  negDesc := utils.NegDescription{
      ClusterUID: clusterUID
	Namespace: svcNamespace
	ServiceName: svcName
	Port: port
  }
 desc := negDesc.string()
}
```

pkg/neg/syncers/utils.go
```diff
@@ -150,23 +172,32 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
+			desc.ServiceName = svcName
+			desc.Port = port
+		}
 		err = cloud.CreateNetworkEndpointGroup(&composite.NetworkEndpointGroup{
 			Version:             version,
 			Name:                negName,
 			NetworkEndpointType: string(networkEndpointType),
 			Network:             cloud.NetworkURL(),
 			Subnetwork:          subnetwork,
+			Description:         desc.String(),
```
freehan, line 189: this may generate an empty json? `{}`

pkg/utils/negdescription.go
```diff
@@ -0,0 +1,58 @@
...
+type NegDescription struct {
+	ClusterUID  string `json:"cluster-uid"`
+	Namespace   string `json:"namespace"`
+	ServiceName string `json:"service-name"`
+	Port        string `json:"port"`
+}
+
+// String returns the string representation of a Description.
+func (desc NegDescription) String() string {
+	if desc.ClusterUID == "" || desc.Namespace == "" || desc.ServiceName == "" || desc.Port == "" {
```
freehan, line 35: not sure if this is worth checking

pkg/utils/negdescription.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		klog.Errorf("Failed to generate neg description string: %v, falling back to empty string", err)
+		return ""
+	}
+	return string(descJson)
+}
+
+// DescriptionFromString gets a Description from string,
+func NegDescriptionFromString(descString string) (*NegDescription, error) {
+	if descString == "" {
```
freehan, line 49: not sure if this is needed

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 		return err
 	}
 
+	enableNegCRD := false
```
freehan, line 214: Probably no need to do this. 
You can directly access the flag here
`if flags.F.EnableNegCrd`

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+			zone,
+			s.NegSyncerKey.String(),
+			s.clusterUID,
+			fmt.Sprint(s.NegSyncerKey.PortTuple.Port),
+			s.NegSyncerKey.NegType,
+			s.cloud,
+			s.serviceLister,
+			s.recorder,
+			s.NegSyncerKey.GetAPIVersion(),
+			enableNegCRD,
```
freehan, line 236: i guess no need for this.

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+		if err != nil {
+			errList = append(errList, err)
+		}
+
+		if enableNegCRD && err == nil {
+			negObjRefs = append(negObjRefs, negObj)
+		}
+	}
+
+	if enableNegCRD {
```
freehan, line 247:
```
if flags.F.EnableNegCrd
```

pkg/neg/syncers/transaction.go
```diff
@@ -413,3 +490,43 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+func updateInitializedCondition(condition negv1beta1.Condition, ts metav1.Time, errList []error) negv1beta1.Condition {
+
+	if condition.Status == core.ConditionTrue {
+		return condition
+	}
+
+	condition.LastTransitionTime = ts
+	if len(errList) > 0 {
+		condition.Status = core.ConditionUnknown
+		condition.Reason = "Error"
```
freehan, line 503: Are there common reasons? 

Make them into const if needed. 

Also consider reusing the reason used by creation Event.

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+
+		neg.Status.LastSyncTime = now
+		if len(neg.Status.Conditions) == 0 {
+			neg.Status.Conditions = []negv1beta1.Condition{
+				negv1beta1.Condition{Type: negv1beta1.Initialized},
+				negv1beta1.Condition{Type: negv1beta1.Synced},
+			}
+		}
+
+		for i, condition := range neg.Status.Conditions {
```
freehan, line 267: I would suggest package this into a function. 
```
EnsureCondition(*NegCRD, ConditionType, ConditionStatus, Message, Reason) bool 
```

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+			negObjRefs = append(negObjRefs, negObj)
+		}
+	}
+
+	if enableNegCRD {
+		now := metav1.Now()
+		neg, err := s.negClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(s.Namespace).Get(context.Background(), s.negName, metav1.GetOptions{})
+		if err != nil {
+			errList = append(errList, err)
+			return utilerrors.NewAggregate(errList)
```
freehan, line 252: I am wondering if we should make this a fatal error. If the CRD is somehow missing or cannot be updated. It should not cause the NEG syncer to stop syncing enpdoints. So it might worth packaging this into a function and capture any error output, log or throw events but do not block the NEG endpoint syncing

pkg/neg/syncers/transaction.go
```diff
@@ -199,10 +211,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+				negv1beta1.Condition{Type: negv1beta1.Synced},
+			}
+		}
+
+		for i, condition := range neg.Status.Conditions {
+			if condition.Type == negv1beta1.Initialized {
+				condition = updateInitializedCondition(condition, now, errList)
+			}
+
+			if condition.Type == negv1beta1.Synced {
```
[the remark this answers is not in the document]
swetharepakula: moved synced condition updating to syncInternal

pkg/neg/syncers/utils.go
swetharepakula: originally thought ensureNetworkGroup was called every sync, so thought we could avoid a call and update the objRefs on the next check. Made adjustments now to do a GetNetworkEndpointGroup for the selflink

pkg/utils/negdescription.go
swetharepakula: Removed these checks, as you suggest, not necessary for our purposes.

pkg/utils/negdescription.go
swetharepakula: You are right, this isn't needed, because we do an empty string in the code before parsing for a NegDescription

pkg/neg/syncers/transaction.go
swetharepakula: Updated with some helper functions to make the logic easier.

pkg/neg/syncers/transaction.go
swetharepakula: Changed this to just log an error for now. Should no longer block the neg endpoint syncing.

pkg/neg/syncers/transaction.go
swetharepakula: The event reasons I found seem to be resource specific. Created constants for Negs in `pkg/negs/types/types.go`

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +458,84 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+	negCR, err := s.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(s.Namespace).Get(context.Background(), s.negName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Error updating Neg CR %s : %s", s.negName, err)
+	}
+
+	ts := metav1.Now()
+	negCR.Status.Conditions = ensureCondition(negCR.Status.Conditions, negv1beta1.Synced, types.NegSyncSuccessful, types.NegSyncFailed, ts, syncErr)
+	negCR.Status.LastSyncTime = ts
+
+	_, err = s.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(s.Namespace).Update(context.Background(), negCR, metav1.UpdateOptions{})
```
freehan, line 499: I think it might make sense to use PATCH instead of Update to update the condition. This will save one GET API call. Double check if you can make it work. If not, we can sync up. 

This is an example https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/readiness/reflector.go#L268

pkg/apis/svcneg/v1beta1/types.go
```diff
@@ -50,7 +50,7 @@ type ServiceNetworkEndpointGroupStatus struct {
 
 	// Last time the NEG syncer syncs associated NEGs.
 	// +optional
-	LastSyncTime metav1.Time `json:"lastTransitionTime,omitempty"`
```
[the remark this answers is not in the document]
swetharepakula: Since only the json tag is changing, it doesn't cause any changes in the generated code.

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+	if !exists {
+		return nil, fmt.Errorf("neg %s/%s is not in store", namespace, negName)
+	}
+
+	return n.(*negv1beta1.ServiceNetworkEndpointGroup), nil
+}
+
+// patchNegStatus patches the specified NegCR status with the provided new status
+func patchNegStatus(svcNegClient svcnegclient.Interface, oldStatus, newStatus negv1beta1.ServiceNetworkEndpointGroupStatus, namespace, negName string) (*negv1beta1.ServiceNetworkEndpointGroup, error) {
+	patchBytes, err := utils.StrategicMergePatchBytes(negv1beta1.ServiceNetworkEndpointGroup{Status: oldStatus}, negv1beta1.ServiceNetworkEndpointGroup{Status: newStatus}, negv1beta1.ServiceNetworkEndpointGroup{})
```
freehan, line 568: I am a bit confused. You are calculating StrategicMergePatch but the patch type is MergePatchType. 

Does it work?

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+	negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.negName)
+	if err != nil {
+		klog.Errorf("failed getting neg from store: %s", err)
+		return
+	}
+	oldStatus := negCR.Status.DeepCopy()
+
+	ts := metav1.Now()
+	// Update Initialized Condition only if it does not exist
+	if _, _, _, exists := findCondition(negCR.Status.Conditions, negv1beta1.Initialized); !exists {
```
freehan, line 511: If the initialized condition is missing, I think the easier way is to set needInit to true and let it reinitialize.

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
 func (s *transactionSyncer) logEndpoints(endpointMap map[string]negtypes.NetworkEndpointSet, desc string) {
 	klog.V(3).Infof("For NEG %q, %s: %+v", s.negName, desc, endpointMap)
 }
+
+// updateInitializedAndNegRefsOnCRD queries the k8s api server for the current NEG CR and updates the Initialized condition and neg objects as appropriate.
+// If neg client is nil, will return immediately
+func (s *transactionSyncer) updateInitializedAndNegRefsOnCRD(negObjRefs []negv1beta1.NegObjectReference, errList []error) {
```
freehan, line 471: May be name to `updateInitStatus`?

and then just describe what would be updated.

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+
+	_, err = patchNegStatus(s.svcNegClient, *oldStatus, neg.Status, s.Namespace, s.negName)
+	if err != nil {
+		klog.Errorf("Error updating Neg CR %s : %s", s.negName, err)
+	}
+}
+
+// ensureStatusOnCR ensures that the Status on the Neg CR is completely populated with both conditions, and NegObjectReferences
+// ensureStatusOnCR will always update the `LastSyncTime` and will selectively update the conditions and NegObjectReferences.
+func (s *transactionSyncer) ensureStatusOnCR(syncErr error) {
```
freehan, line 498: `updateStatus`

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+}
+
+// patchNegStatus patches the specified NegCR status with the provided new status
+func patchNegStatus(svcNegClient svcnegclient.Interface, oldStatus, newStatus negv1beta1.ServiceNetworkEndpointGroupStatus, namespace, negName string) (*negv1beta1.ServiceNetworkEndpointGroup, error) {
+	patchBytes, err := utils.StrategicMergePatchBytes(negv1beta1.ServiceNetworkEndpointGroup{Status: oldStatus}, negv1beta1.ServiceNetworkEndpointGroup{Status: newStatus}, negv1beta1.ServiceNetworkEndpointGroup{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare patch bytes: %s", err)
+	}
+
+	return svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Patch(context.Background(), negName, types.MergePatchType, patchBytes, metav1.PatchOptions{})
```
freehan, line 573: I may have mislead you in the previous comment. 

Some background: https://erosb.github.io/post/json-patch-vs-merge-patch/
Strategic Merge patch is different than these two. 


I have some code example for calculating json patch
```
import (
        "encoding/json"
        "fmt"
        "github.com/mattbaird/jsonpatch"
)

// CreatePatch creates a json patch to convert the old object to the new object.
func CreatePatch(old []byte, newObj interface{}) ([]byte, error) {
        newJson, err := json.Marshal(newObj)
        if err != nil {
                return nil, fmt.Errorf("failed to marshal object %v into json: %v", newObj, err)
        }
        patch, err := jsonpatch.CreatePatch(old, newJson)
        if err != nil {
                return nil, fmt.Errorf("failed to creat patch: %v", err)
        }
        return json.Marshal(patch)
}

```

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+	}
+}
+
+// ensureStatusOnCR ensures that the Status on the Neg CR is completely populated with both conditions, and NegObjectReferences
+// ensureStatusOnCR will always update the `LastSyncTime` and will selectively update the conditions and NegObjectReferences.
+func (s *transactionSyncer) ensureStatusOnCR(syncErr error) {
+	if s.svcNegClient == nil {
+		return
+	}
+	negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.negName)
```
freehan, line 502: When you get an object from cache, you should not modify it directly (you modify it later in ensureCondition). 

I would recommend you do the following: 
1. make a copy of the object. 
2. make modification to the object (ensure conditions)
3. calculate json patch for the whole object
4. patch

pkg/neg/syncers/transaction.go
freehan: I have not found handy example for json merge patch. If you find it you might use it as well. No preference for me.

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+	}
+	oldStatus := negCR.Status.DeepCopy()
+
+	ts := metav1.Now()
+	// Update Initialized Condition only if it does not exist
+	if _, _, _, exists := findCondition(negCR.Status.Conditions, negv1beta1.Initialized); !exists {
+		negCR.Status.Conditions = ensureCondition(negCR.Status.Conditions, negv1beta1.Initialized, negtypes.NegInitializationSuccessful, negtypes.NegInitializationFailed, ts, syncErr)
+	}
+
+	negCR.Status.Conditions = ensureCondition(negCR.Status.Conditions, negv1beta1.Synced, negtypes.NegSyncSuccessful, negtypes.NegSyncFailed, ts, syncErr)
```
freehan, line 515: recommend modify `ensureCondition` to directly ensure the conditions on the object directly. 

for example, 
ensureCondition(svcNeg *negv1beta1.ServiceNetworkEndpointGroup, expectedCondition condition)
ensureCondition(svcNeg *negv1beta1.ServiceNetworkEndpointGroup, type, reason, error, message, timestamp)

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +465,148 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+
+	_, err = patchNegStatus(s.svcNegClient, *oldStatus, negCR.Status, s.Namespace, s.negName)
+	if err != nil {
+		klog.Errorf("Error updating Neg CR %s : %s", s.negName, err)
+	}
+}
+
+// ensureNegRefs verfies that the negRefs provided are not empty. If empty, queries the cloud for the neg and returns a
+// list of NegObjectReferences to be set in the Neg CR status
+func ensureNegRefs(cloud negtypes.NetworkEndpointGroupCloud, zones []string, negRefs []negv1beta1.NegObjectReference, negName string, version meta.Version) []negv1beta1.NegObjectReference {
```
freehan, line 532: seems a bit redundant since the `ensureNetworkEndpointGroups ` has done this.

pkg/neg/syncers/utils.go
```diff
@@ -128,17 +131,32 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
 			if svc := getService(serviceLister, svcNamespace, svcName); svc != nil {
 				recorder.Eventf(svc, apiv1.EventTypeNormal, "Delete", "Deleted NEG %q for %s in %q.", negName, negServicePortName, zone)
 			}
 		}
+	} else {
+		// Skip checking description if the neg object has an empty description
+		if neg.Description != "" {
+			description, err := utils.NegDescriptionFromString(neg.Description)
+			if err != nil {
+				klog.V(4).Infof("Error unmarshalling Neg Description %s/%s err:%s", negName, zone, err)
```
freehan, line 151: `klog.Warningf()`

This is sort of unexpected

pkg/neg/syncers/transaction.go
swetharepakula: I did take a look at the MergePatchType. I tested both out and both merge patches worked with the MergePatchType in this situation. I was avoiding the MergePatch just to reduce the amount of code introduced in this PR. But I can switch to using a MergePatch

pkg/neg/syncers/transaction.go
swetharepakula: If needInit is false, the NegRefs never gets updated if something goes wrong the first time in updating the cr or if a neg cr is deleted and recreated before the syncer and neg is not GCed yet. But with the suggestion above in regards to populating the initialization condition, that would cause this function to become unnecessary so I will remove this.

pkg/neg/syncers/transaction.go
swetharepakula: There is a parallel function for calculating the json merge patch in apimachinery (https://github.com/kubernetes/apimachinery/blob/master/pkg/util/jsonmergepatch/patch.go#L31). It will be similar to how the StrategicMergePatch is calculated. I can use that and avoid bringing in another dependency.

pkg/neg/syncers/transaction.go
swetharepakula: I like that suggestion. I made changes accordingly.

pkg/neg/syncers/transaction.go
```diff
@@ -416,3 +464,123 @@ func (s *transactionSyncer) logStats(endpointMap map[string]negtypes.NetworkEndp
...
+
+	condition, index, conditions, _ := findCondition(neg.Status.Conditions, conditionType)
+	if condition.Status == core.ConditionTrue && err == nil {
+		neg.Status.Conditions = conditions
+		return
+	}
+
+	condition.LastTransitionTime = ts
+	if err != nil {
+		condition.Status = core.ConditionUnknown
```
freehan, line 560: ConditionFalse

PR #1188, opened 2020-07-18
Revert to individual secret get

liggitt:
This PR:
* reverts the mechanism for fetching TLS secrets to a Get() (away from listing all secrets in a namespace in https://github.com/kubernetes/ingress-gce/pull/1101/files#r456811105)
* removes the secret list permission added in https://github.com/kubernetes/ingress-gce/pull/1134/files#r456812266
* resolves the test failure reported in https://github.com/kubernetes/kubernetes/pull/93031

/assign @bowei @rramkumar1

### DIFF
--- a/docs/deploy/resources/rbac.yaml
+++ b/docs/deploy/resources/rbac.yaml
@@ -37,7 +37,7 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["get", "list"]
+  verbs: ["get"]
 - apiGroups: [""]
   resources: ["events"]
   verbs: ["get", "list", "watch", "update", "create", "patch"]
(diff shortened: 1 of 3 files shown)


bowei:
Is this a general policy for secrets?

liggitt:
> Is this a general policy for secrets?

it's the recommendation for controllers that are not expected to be highly privileged, see https://kubernetes.io/docs/concepts/configuration/secret/#clients-that-use-the-secret-api

> For these reasons watch and list requests for secrets within a namespace are extremely powerful capabilities and should be avoided, since listing secrets allows the clients to inspect the values of all secrets that are in that namespace. The ability to watch and list all secrets in a cluster should be reserved for only the most privileged, system-level components.
>
> Applications that need to access the Secret API should perform get requests on the secrets they need. This lets administrators restrict access to all secrets while white-listing access to individual instances that the app needs.

This recommendation came out of discussions in sig-auth specifically around ingress controller access patterns:
* https://github.com/kubernetes/community/blob/master/sig-auth/archive/meeting-notes-2017.md#july-12th-2017-11a---noon-pacific-time
* https://github.com/kubernetes/ingress-nginx/issues/816

PR #1197, opened 2020-07-29
Use NEG CRs for NEG Garbage Collection

swetharepakula:
- Add finalizer to NEG CRs to ensure CR exists until NEG is deleted
 during GC
 - When neg crd is enabled, GC will use neg crs, otherwise will fallback
 to regular GC using the naming to determine which NEGs are to be
 deleted

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"sync"
 
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -39,6 +40,8 @@ import (
 	negsyncer "k8s.io/ingress-gce/pkg/neg/syncers"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
 	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/common"
 	"k8s.io/klog"
 	utilpointer "k8s.io/utils/pointer"
 )
@@ -238,10 +241,16 @@ func (manager *syncerManager) GC() error {
 	manager.garbageCollectSyncer()
 
 	// Garbage collect NEGs
-	if err := manager.garbageCollectNEG(); err != nil {
-		return fmt.Errorf("failed to garbage collect negs: %v", err)
+	var err error
+	if manager.svcNegClient != nil {
+		err = manager.garbageCollectNEGWithCRD()
+	} else {
+		err = manager.garbageCollectNEG()
 	}
-	return nil
+	if err != nil {
+		err = fmt.Errorf("failed to garbage collect negs: %v", err)
+	}
+	return err
 }
 
 // ReadinessGateEnabledNegs returns a list of NEGs which has readiness gate enabled for the input pod's namespace and labels.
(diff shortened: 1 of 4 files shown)


pkg/neg/manager.go
```diff
@@ -370,6 +377,62 @@ func (manager *syncerManager) garbageCollectNEG() error {
...
+		}
+		if _, ok := deletionCandidates[neg.Name]; ok {
+			if err := manager.ensureDeleteNetworkEndpointGroup(neg.Name, key.Zone); err != nil {
+				return fmt.Errorf("failed to delete NEG %s in %s: %s", neg.Name, key.Zone, err)
+			}
+		}
+	}
+	for _, cr := range deletionCandidates {
+		cr.Finalizers = []string{}
+		if cr.GetDeletionTimestamp().IsZero() {
```
freehan, line 424: probably need to send a delete call

pkg/neg/manager.go
```diff
@@ -370,6 +377,62 @@ func (manager *syncerManager) garbageCollectNEG() error {
...
+	var errList []error
+	for key, neg := range negList {
+		if key.Type() != meta.Zonal {
+			// covers the case when key.Zone is not populated
+			klog.V(4).Infof("Ignoring key %v as it is not zonal", key)
+			continue
+		}
+		if _, ok := deletionCandidates[neg.Name]; ok {
+			if err := manager.ensureDeleteNetworkEndpointGroup(neg.Name, key.Zone); err != nil {
+				return fmt.Errorf("failed to delete NEG %s in %s: %s", neg.Name, key.Zone, err)
```
freehan, line 418: cannot return here. 
Instead it should record an event to the NEG CR if it is there

pkg/neg/manager.go
```diff
@@ -370,6 +377,62 @@ func (manager *syncerManager) garbageCollectNEG() error {
 	return nil
 }
 
+// garbageCollectNEGWithCRD uses the NEG CRs and the svcPortMap to determine which NEGs
+// need to be garbage collected. NEGs that have a corresponding CR with a deletion
+// timestamp, don't exist in the svcPortMap or don't have a corresponding NEG in the
+// will all be garbage collected
+func (manager *syncerManager) garbageCollectNEGWithCRD() error {
```
freehan, line 384: Assumption is that NEG CR exists then NEG must not be fully GCed.
Try to report event to NEG CR if NEG cannot be GCed. 

Case 1: NEG CR with deletion timestamp + Manager PortInfo Map does not have key 
- Delete NEG 
- Remove Finalizer

Case 2: NEG CR has deletion timestamp + Manager PortInfo Map has the key 
- Remove Finalizer

Case 3: NEG CR does not have deletion timestamp + Manager PortInfoMap does not have key (controller is down when the service or annotation is removed)
- Delete NEG
- Delete NEG CR
- Remove Finalzier

freehan: func A: Delete NEG 
func B: Delete NEG CR (Delete and remove Finalzier)

freehan: For each deletion candidate (bool shouldDeleteNEG):
- if shouldDeleteNEG, then try delete NEG
- Delete NEG CR

freehan: Deletion candidate:

case 1: shouldDeleteNEG = true
case 2: shouldDeleteNEG= false
case 3: shouldDeleteNEG = true

pkg/neg/manager.go
```diff
@@ -370,6 +377,62 @@ func (manager *syncerManager) garbageCollectNEG() error {
...
+		}
+	}()
+
+	negList, err := manager.cloud.AggregatedListNetworkEndpointGroup(meta.VersionGA)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve aggregated NEG list: %v", err)
+	}
+
+	var errList []error
+	for key, neg := range negList {
```
freehan, line 410: one corner to watch out:

cluster A -> try to create NEG foo. (NEG foo belongs cluster B)

when service or annotation got updated, and cluster A wants to GC NEG foo. Will it GC NEG foo?

freehan: check description or initialized condition

pkg/neg/manager.go
```diff
@@ -473,6 +539,29 @@ func ensureNegCROwnerRef(negCR *negv1beta1.ServiceNetworkEndpointGroup, expected
...
+
+	// If CR does not have a deletion timestamp, delete
+	if negCR.GetDeletionTimestamp().IsZero() {
+		return svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(negCR.Namespace).Delete(context.Background(), negCR.Name, metav1.DeleteOptions{})
+	}
+	return nil
+}
+
+// patchNegStatus patches the specified NegCR status with the provided new status
+func patchNegStatus(svcNegClient svcnegclient.Interface, oldNeg, newNeg negv1beta1.ServiceNetworkEndpointGroup) (*negv1beta1.ServiceNetworkEndpointGroup, error) {
```
freehan, line 556: this is not used?

I thought the other PR has added this somewhere

pkg/neg/manager.go
```diff
@@ -370,6 +379,62 @@ func (manager *syncerManager) garbageCollectNEG() error {
...
+		neg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
+		deletionCandidates[neg.Name] = neg
+	}
+
+	func() {
+		manager.mu.Lock()
+		defer manager.mu.Unlock()
+		for _, portInfoMap := range manager.svcPortMap {
+			for _, portInfo := range portInfoMap {
+				// only keep in deletionCandidates if neg does not exist in svcPortMap
```
freehan, line 399: add a comment regarding the 2 corner cases:
1. NEG CR has deletionTimestamp but svcPortMap has the svc
2. NEG CR does not have deletionTImeStamp but svcPortMap does not has the svc.

pkg/neg/manager.go
swetharepakula: This is used above in deleteSvcNegCR

PR #1198, opened 2020-07-29
Refactor tls package into translator

spencerhance:
This is necessary for the eventual refactor of ssl certificates into translator


pkg/translator/translator.go
```diff
@@ -311,3 +325,27 @@ func sslPolicyLink(env *Env) (*string, error) {
 
 	return &resID, nil
 }
+
+func GetCertHash(contents string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(contents)))[:16]
```
rramkumar1, line 330: can be unexported?

spencerhance: ah this is used all over loadbalancers_test.go right now

pkg/controller/controller.go
```diff
@@ -40,7 +40,7 @@ import (
 	"k8s.io/ingress-gce/pkg/backends"
 	"k8s.io/ingress-gce/pkg/common/operator"
 	"k8s.io/ingress-gce/pkg/context"
-	"k8s.io/ingress-gce/pkg/controller/translator"
+	controllertranslator "k8s.io/ingress-gce/pkg/controller/translator"
```
rramkumar1, line 43: can we call this legacytranslator

pkg/translator/translator.go
rramkumar1: Is there a particular reason why loadbalancers_test.go needs it? I feel like that test should use a fake namer. 

Don't have to address it in this CL but we should see if we can leave a TODO for that.

spencerhance: Ah no it can actually be refactored out pretty easily.  There's another call in controller_test.go, so maybe we could add it someone centralized.

pkg/controller/controller.go
spencerhance: done

pkg/translator/translator.go
spencerhance: Okay I added a TODO for now

PR #1202, opened 2020-08-01
Add NegName to syncer key

swetharepakula:
- With custom neg names, multiple negs can exist with the same syncer
 key but different names. To ensure negs are properly managed and GC'd
 add neg name to the syncer key so negs with different names will be
 considered different


freehan:
/assign

can you update the PR and commit message to include why NegName is needed in syncerKey?

freehan:
Also need to remove NegName from the existing Syncer?

PR #1206, opened 2020-08-05
Implement support for HTTPS Redirects

spencerhance:
Adds support for using HTTPS Redirects with L7 ELB via a Redirects Config in FrontendConfig
Also includes a new status annotation for the additional UrlMap that needs to be synced

e2e tests will be added in a follow up


spencerhance:
@freehan can you also take a look?

/assign @freehan

pkg/utils/namer/frontendnamer.go
```diff
@@ -88,6 +90,11 @@ func (ln *V1IngressFrontendNamer) UrlMap() string {
 	return ln.namer.UrlMap(ln.lbName)
 }
 
+// UrlMap implements IngressFrontendNamer.
+func (ln *V1IngressFrontendNamer) RedirectUrlMap() string {
+	return ln.namer.RedirectUrlMap(ln.lbName)
```
freehan, line 95: I strongly recommend NOT to support this feature in V1 Namer.

pkg/utils/namer/interfaces.go
```diff
@@ -25,6 +25,8 @@ type IngressFrontendNamer interface {
 	TargetProxy(protocol NamerProtocol) string
 	// UrlMap returns the name of the URL Map.
 	UrlMap() string
+	// RedirectUrlMap returns the name of the URL Map.
+	RedirectUrlMap() string
```
freehan, line 29:
```
// RedirectUrlMap returns the name of the URL Map and if the namer supports naming redirectUrlMap
RedirectUrlMap() (string, bool)
```

pkg/utils/namer/frontendnamer.go
freehan: Return an error event. Asking user to recreate ingress and it will come with new naming scheme.

spencerhance: Done

pkg/utils/namer/interfaces.go
spencerhance: Done

spencerhance:
@freehan 

>One more thing, there is one corner case: If someone enabled the HTTP redirect and disable later. Will the RedirectURLMap get trimmed? From the code, I do not think so, right?

Good point, I added in code to delete it when disabled

pkg/loadbalancers/url_maps.go
```diff
@@ -80,6 +81,71 @@ func (l *L7) ensureComputeURLMap() error {
...
+
+	name, namerSupported := l.namer.RedirectUrlMap()
+	expectedMap := t.ToRedirectUrlMap(env, l.Versions().UrlMap)
+
+	key, err := l.CreateKey(name)
+	if err != nil {
+		return err
+	}
+
+	currentMap, err := composite.GetUrlMap(l.cloud, key, l.Versions().UrlMap)
```
freehan, line 99: Might worth avoiding this GET call every time an irrelevant ingress was synced. 

One idea is to check if the the ingress status annotation has the redirectUrlMap

pkg/loadbalancers/url_maps.go
```diff
@@ -80,6 +81,71 @@ func (l *L7) ensureComputeURLMap() error {
...
+	if utils.IgnoreHTTPNotFound(err) != nil {
+		return err
+	}
+
+	if expectedMap == nil {
```
freehan, line 104: add a comment here to say "do not expect to have a RedirectUrlMap"

pkg/loadbalancers/url_maps.go
```diff
@@ -80,6 +81,71 @@ func (l *L7) ensureComputeURLMap() error {
...
+		if currentMap == nil {
+			return nil
+		} else {
+			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); err != nil {
+				return err
+			}
+		}
+	} else {
```
freehan, line 112: probably no need for the else block

pkg/loadbalancers/url_maps.go
spencerhance: Done

pkg/loadbalancers/url_maps.go
spencerhance: Done

pkg/loadbalancers/url_maps.go
spencerhance: Done

coderaven:
Good to know this has been merged finally! I'm new to this, when is the expected timeline for this to be available in GKE? And is there any documentation that comes along with it?

peterldowns:
Yes, would really love to use this, when will it be available in GKE?

artvichi:
Is this going to be live at some point in this century ?

spencerhance:
Hi folks, please follow the discussion at https://github.com/kubernetes/ingress-gce/issues/1075

We are doing everything we can to launch this ASAP

pkg/loadbalancers/target_proxies.go
```diff
@@ -34,14 +34,29 @@ const (
 
 // checkProxy ensures the correct TargetHttpProxy for a loadbalancer
 func (l *L7) checkProxy() (err error) {
-	isL7ILB := flags.F.EnableL7Ilb && utils.IsGCEL7ILBIngress(l.runtimeInfo.Ingress)
-	tr := translator.NewTranslator(isL7ILB, l.namer)
+	// Get UrlMap Name, could be the url map or the redirect url map
+	// TODO(shance): move to translator
+	var umName string
```
rhzs, line 39: Why didn't this code written as follow with less branches?

```go
umName := l.um.Name
if flags.F.EnableFrontendConfig {
  // TODO(shance): check for empty name?
    if l.redirectUm != nil && l.runtimeInfo.FrontendConfig.Spec.RedirectToHttps != nil && l.runtimeInfo.FrontendConfig.Spec.RedirectToHttps.Enabled {
      umName = l.redirectUm.Name
    }
}
```

PR #1207, opened 2020-08-05
Add e2e tests for neg naming and neg crd

swetharepakula:


cmd/e2e-test/neg_test.go
```diff
@@ -386,3 +386,241 @@ func TestReadinessReflector(t *testing.T) {
...
+		_, err = e2e.EnsureEchoService(s, svc2, map[string]string{}, v1.ServiceTypeClusterIP, replicas)
+		if err != nil {
+			t.Fatalf("error ensuring echo service: %v", err)
+		}
+
+		e2e.WaitForStandaloneNegDeletion(ctx, Framework.Cloud, s, map[string]string{port80.String(): expectedNEGName}, negStatus.NetworkEndpointGroups, negStatus.Zones, true)
+	})
+}
+
+func TestNegCRDUserActions(t *testing.T) {
```
freehan, line 561: E2e test is quite heavy. I would recommend not adding e2e test for negative cases. 
Unless, it is a very important negative case. 

Let us discuss if this is worth keeping

cmd/e2e-test/neg_test.go
```diff
@@ -386,3 +386,241 @@ func TestReadinessReflector(t *testing.T) {
...
+
+	Framework.RunWithSandbox("NEGs with custom names", t, func(t *testing.T, s *e2e.Sandbox) {
+		var previousZones []string
+		var previousNegs map[string]string
+
+		for _, tc := range []struct {
+			desc             string
+			annotations      annotations.NegAnnotation
+			expectedNegAttrs map[string]string
+			gcNegAttrs       map[string]string
```
freehan, line 407: might be worth adding another parameter: replicas? 

I think `WaitForNegs` should already check the number of endpoints in the NEGs.

pkg/e2e/helpers.go
```diff
@@ -539,6 +545,41 @@ func CheckNegStatus(svc *v1.Service, expectSvcPors []string) (annotations.NegSta
 	return negStatus, nil
 }
 
+// CheckCustomNegNameStatus checks if the NEG Status annotation is present and in the expected state
+func CheckCustomNegNameStatus(svc *v1.Service, expectedNegAttrs map[string]string) (annotations.NegStatus, error) {
```
freehan, line 549: add a comment about `expectedNegAttrs` parameter

freehan: CheckNameInNegStatus?

pkg/e2e/helpers.go
```diff
@@ -664,3 +705,206 @@ func waitForBackendConfigCRDEstablish(crdClient *apiextensionsclient.Clientset)
...
+	if l := len(negFinalizers); l != 1 {
+		return fmt.Errorf("expected 1 neg Finalizer but got %d", l)
+	}
+	if negFinalizers[0] != common.NegFinalizerKey {
+		return fmt.Errorf("expected neg Finalizer %q but got %q", common.NegFinalizerKey, negFinalizers[0])
+	}
+	return nil
+}
+
+// WaitForStandaloneNegDeletion waits for all NEGs associated with a GCLB to be deleted via GC
```
freehan, line 832: // WaitForStandaloneNegDeletion waits for standalone NEGs and corresponding CRDs are deleted via GC.

freehan: add comments about expectedDelteNegsCRs, negConfigs 

consider:
1. pass negConfigs and zones together as pastNegStatus.
2. pass the NEG name which expected to be deleted. So that making the function simpler only validate if one NEG and CRD has been delete .

PR #1215, opened 2020-08-13
Add the daemon for external workload demo

zjkmxy:
- Add `workload` CR proposal to represent an external workload.
- Add `workload-daemon` which does self-registration and heartbeat from the workload side.


cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+	if home != "" {
+		configFile := filepath.Join(home, ".kube", "config")
+		if _, err := os.Stat(configFile); !os.IsNotExist(err) {
+			config, err = clientcmd.BuildConfigFromFlags("", configFile)
+			if err != nil {
+				klog.Fatalf("unable to build config from kubeConfig file: %+v", err)
+			}
+			return
+		}
+	}
```
robscott, line 147: There are a few different ways to do this, but I've lost track of what the _best_ way is. Back in the day I used an [approach that allowed specifying a kube context to use](https://github.com/robscott/kube-capacity/blob/master/pkg/kube/clientset.go#L48-L51). That doesn't seem necessary here, but a simpler version of that might work. Alternatively, this StackOverflow looks a bit more straightforward, I just haven't tried it myself: https://stackoverflow.com/questions/55314152/how-to-get-namespace-from-current-context-set-in-kube-config.

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+
+	return
+}
+
+// NewCloudVM fetches all data needed from the metadata server to create CloudVM
+func NewCloudVM() *CloudVM {
+	var err error
+	vm := CloudVM{}
+
+	// Fetch basic info that every machine on gcloud has
```
robscott, line 175:
```suggestion
	// Fetch basic info that every GCP Instance has
```

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		klog.Fatalf("failed to get Zone from metadata server: %+v", err)
+	}
+
+	// Fetch the cluster name and zone
+	// Not specified if the user want to use ~/.kube/config file
+	vm.clusterName, err = metadata.InstanceAttributeValue("k8s-cluster-name")
+	if err != nil {
+		vm.clusterName = ""
+	}
```
robscott, line 206: This pattern is repeated enough that it may be worth creating a helper function to handle it.

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+	vm.vmLabels = make(map[string]string)
+	attrs, err := metadata.InstanceAttributes()
+	if err != nil {
+		klog.Fatalf("failed to get attribute list from metadata server: %+v", err)
+	}
+	for _, name := range attrs {
+		if strings.HasPrefix(name, "k8s-label-") {
+			val, err := metadata.InstanceAttributeValue(name)
+			if err != nil {
+				klog.V(0).Infof("Faild to fetch label %s: %+v", name, err)
```
robscott, line 232:
```suggestion
				klog.Errorf("failed to fetch label %s: %+v", name, err)
```

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+			}
+			vm.vmLabels[name[10:]] = val
```
robscott, line 234: This could use a comment to explain the 10 character prefix removal.

cmd/workload-daemon/app/kube-config.go
```diff
@@ -0,0 +1,83 @@
...
+    cluster: %[3]s
+    user: %[4]s
+  name: %[3]s
+current-context: %[3]s
+kind: Config
+preferences: {}
+users:
+- name: %[4]s
+  user:
+    token: %[5]s`
```
robscott, line 56: Can you use go templates for this instead?

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,129 @@
...
+		app.OutputCredentials(credentials)
+		return
+	case "start":
+		vm := app.NewCloudVM()
+		var workload app.WorkloadInfo = vm
+		var helper app.ConnectionHelper = vm
+
+		// Generate KubeConfig and connect to it
+		config := helper.KubeConfig()
+		var clientSet workloadclient.Interface
```
robscott, line 58: Nit: this is usually referred to as `clientset`.

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,129 @@
...
+		// Update the heartbeat regularly
+		ticker := time.NewTicker(updateInterval)
+		quit := make(chan interface{})
+		sigs := make(chan os.Signal, 1)
+		go updatingCR(workload, clientSet, ticker, sigs, quit)
+
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		<-quit
+
+		client.Delete(context.TODO(), workload.Name(), metav1.DeleteOptions{})
```
robscott, line 81: Should handle error here

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,129 @@
...
+		klog.V(0).Infof("CR deleted")
+
+		return
+	default:
+		fmt.Printf("Usage: %v [command] \n", os.Args[0])
+		return
+	}
+}
+
+func updatingCR(workload app.WorkloadInfo, clientSet workloadclient.Interface, ticker *time.Ticker,
```
robscott, line 91:
```suggestion
func updateCR(workload app.WorkloadInfo, clientSet workloadclient.Interface, ticker *time.Ticker,
```

cmd/workload-daemon/metadata/metadata.go
```diff
@@ -0,0 +1,513 @@
+// Copyright 2014 Google LLC
```
robscott, line 1:
```suggestion
// Copyright 2020 Google LLC
```

robscott: Or is this just copied from elsewhere?

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -0,0 +1,67 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Workload represents an external workload such as VM
```
robscott, line 23: Nit: I'm used to this following the annotations. Don't think it's actually required, just more of a pattern I've seen commonly used for k8s type definitions.

cmd/workload-daemon/metadata/metadata.go
zjkmxy: Yes. This is [cloud.google.com/go/compute/metadata](https://github.com/googleapis/google-cloud-go/blob/master/compute/metadata/metadata.go). Minhan does not want to have extra dependencies, so I copied it here. Is there any better solution?

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,239 @@
...
+	}{}
+	err = json.Unmarshal([]byte(jsonData), &token)
+	if err != nil {
+		klog.Fatalf("malformed service account access token: %+v", err)
+	}
+
+	expiryTime := time.Now().Add(time.Duration(token.ExpiresIn) * time.Second)
+	return ClusterCredentials{
+		AccessToken: token.AccessToken,
+		TokenExpiry: expiryTime.UTC().Format(time.RFC3339),
```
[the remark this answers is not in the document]
zjkmxy: Expired token is handled by client-go. It automatically executes the command written in KubeConfig to get a new access token.
- For unit test, I don't know.
- For integration test, I leave the access_token field in KubeConfig blank, so client-go has to fetch this token when connecting the first time.

pkg/apis/workload/v1alpha1/types.go
zjkmxy: The existing type definitions use this style:
https://github.com/kubernetes/ingress-gce/blob/2e844caa8201d59962f53e1716082835bbd1aae6/pkg/apis/svcneg/v1beta1/types.go#L24-L31

So I think I'd better make it consistent.

robscott: Oh interesting, good call.

cmd/workload-daemon/metadata/metadata.go
robscott: 👍 I think that's reasonable

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,133 @@
...
+			klog.Fatalf("unable to connect to the cluster: %+v", err)
+		}
+
+		// Create the workload resource
+		client := clientset.NetworkingV1alpha1().Workloads(corev1.NamespaceDefault)
+		_, err = client.Create(context.Background(), getWorkloadCR(workload), metav1.CreateOptions{})
+		if err != nil {
+			klog.Fatalf("unable to create the workload cr: %+v", err)
+		}
+		klog.V(0).Infof("CR created %s", workload.Name())
```
robscott, line 70: Nit: Not familiar with standards here, but in k8s we generally log info at level 2 and debug at level 4. I'd change this to `V(2)`.

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,133 @@
...
+	sigs chan os.Signal, quit chan interface{}) {
+	for {
+		select {
+		case <-ticker.C:
+			patchStr := `[{"op": "replace", "path": "/status/heartbeat", "value": "%s"}]`
+			patch := []byte(fmt.Sprintf(patchStr, time.Now().UTC().Format(time.RFC3339)))
+			vmInstClient := clientset.NetworkingV1alpha1().Workloads(corev1.NamespaceDefault)
+			_, err := vmInstClient.Patch(context.Background(), workload.Name(), types.JSONPatchType, patch, metav1.PatchOptions{})
+			if err != nil {
+				klog.V(0).Infof("CR update failed: %+v", err)
```
robscott, line 105:
```suggestion
				klog.Errorf("Workload resource update failed: %+v", err)
```

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,133 @@
...
+			} else {
+				klog.V(0).Infof("CR updated %s", workload.Name())
```
robscott, line 107: I'd recommend referencing the resource type here, I think it's workload?
```suggestion
				klog.V(2).Infof("Workload resource updated %s", workload.Name())
```

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,259 @@
...
+// Locality associated with the endpoint. A locality corresponds to a failure domain.
+func (vm *CloudVM) Locality() string {
+	return vm.zone
+}
+
+func decodeGsaAccessToken(jsonData string) gsaAccessToken {
+	token := gsaAccessToken{}
+	err := json.Unmarshal([]byte(jsonData), &token)
+	if err != nil {
+		klog.Fatalf("malformed service account access token: %+v", err)
```
robscott, line 103: I think I'd mentioned this elsewhere but not seeing the comment right now. I think it would be good to avoid `klog.Fatalf` and instead return errors up the stack to allow for better handling and processing of things like deferred calls. Although a bit old, this article has some good tips on handle errors in go: https://banzaicloud.com/blog/error-handling-go/. There have been some newer developments since that was written such as `errors.Is` (https://golang.org/src/errors/wrap.go?s=1170:1201#L29), but I think it still generally has good tips.

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,259 @@
...
+	clusterSvc := gkev1.NewProjectsZonesClustersService(gkeSvc)
+	cluster, err = clusterSvc.Get(vm.projectID, vm.clusterZone, vm.clusterName).Do()
+	if err != nil {
+		klog.Errorf("failed to get gke cluster: %+v", err)
+		return
+	}
+	return
+}
+
+// KubeConfig yields the config used to create Kubernetes clientset
```
robscott, line 165: Maybe add a bit more detail to this comment explaining the 3 different ways it can derive kubeconfig.

cmd/workload-daemon/app/kube-config.go
```diff
@@ -0,0 +1,114 @@
+package app
```
robscott, line 1: Looks like this is missing the license header

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,155 @@
...
+	"os/signal"
+	"syscall"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/ingress-gce/cmd/workload-daemon/app"
+	"k8s.io/klog"
+
```
robscott, line 32: Nit: Usually the only empty line in an import statement would be between stdlib and other dependencies like you have on 26, generally the rest should be together.

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,155 @@
...
+	sigs chan os.Signal, quit chan interface{}) {
+	for {
+		select {
+		case <-ticker.C:
+			patchStr := `[{"op": "replace", "path": "/status/heartbeat", "value": "%s"}]`
+			patch := []byte(fmt.Sprintf(patchStr, time.Now().UTC().Format(time.RFC3339)))
+			vmInstClient := clientset.NetworkingV1alpha1().Workloads(corev1.NamespaceDefault)
+			_, err := vmInstClient.Patch(context.Background(), workload.Name(), types.JSONPatchType, patch, metav1.PatchOptions{})
+			if err != nil {
+				klog.Errorf("failed to update the workload resource: %+v", err)
```
robscott, line 127: It seems like we may want some better error handling here, but I'll defer to @freehan for ideas on this. I know you likely don't want to exit for a single error. In Kubernetes we'd usually have a retry mechanism and an exponential backoff but that may be too complex for this use case.

cmd/workload-daemon/app/kube-config.go
```diff
@@ -0,0 +1,114 @@
...
+preferences: {}
+users:
+- name: {{.saName}}
+  user:
+    token: {{.accessToken}}`
+
+// GenKubeConfigForKSA generates a KubeConfig to access the cluster using a Kubernetes service account
+func GenKubeConfigForKSA(clusterCa, clusterIP, clusterName, saName, accessToken string) []byte {
+	var kubeConfig bytes.Buffer
+	t, err := template.New("user").Parse(kubeConfigUserTemp)
```
robscott, line 64: Nit: I'm more used to seeing gotemplates defined as separate files and then using `.ParseFiles`.

pkg/apis/workload/register.go
```diff
@@ -0,0 +1,21 @@
+/*
```
freehan, line 1: recommend moving this api definition and client library into /pkg/experimental/workload/apis

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,157 @@
+/*
```
freehan, line 1: move the core daemon logic in to `pkg/experimental/workload/daemon`

making main.go a simple handler of flags and configs.

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,393 @@
...
+	// TODO: Use selector instaed of name to select endpointslices
+	// esSelector := labels.Set(map[string]string{
+	// 	discovery.LabelServiceName: service.Name,
+	// 	discovery.LabelManagedBy:   controllerName,
+	// })
+
+	// To simplify the complexity, assume there is only one slice managed by this controller.
+	// TODO: fix this
+	// SA: https://github.com/kubernetes/kubernetes/blob/bdb99c8e0954c6b2d4c40233ded94455a343af73/pkg/controller/endpointslice/reconciler.go#L58:22
+	wls := c.workloadLister.List()
```
freehan, line 282: Put list and filter logic into a separate function

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		if !wlSelector.Matches(labels.Set(workload.Labels)) {
+			continue
+		}
+
+		subsets = append(subsets, workloadToEndpoint(workload, service))
+	}
+
+	// Create or update EndpointSlice
+	sliceName := name + "-" + controllerName
```
freehan, line 297: Warning. Do not modify naming scheme

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		Endpoints:   []discovery.Endpoint{},
+	}
+}
+
+func getEndpointPortsFromServicePorts(svcPorts []corev1.ServicePort) []discovery.EndpointPort {
+	ret := []discovery.EndpointPort{}
+	for _, port := range svcPorts {
+		ret = append(ret, discovery.EndpointPort{
+			Name:        &port.Name,
+			Port:        &port.Port,
```
freehan, line 387: you need targetPort as well. coming from Service port spec

cmd/workload-daemon/metadata/metadata.go
```diff
@@ -0,0 +1,513 @@
...
+// limitations under the License.
+
+// Package metadata provides access to Google Compute Engine (GCE)
+// metadata and API service accounts.
+//
+// This package is a wrapper around the GCE metadata service,
+// as documented at https://developers.google.com/compute/docs/metadata.
+package metadata // import "cloud.google.com/go/compute/metadata"
+
+import (
```
freehan, line 22: Add TODO to use GCE metadata client-go when writing up official daemon. 

Add comment: This is copied from ....

cmd/workload-daemon/main.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+const updateInterval time.Duration = 30 * time.Second
+
+func main() {
+	if len(os.Args) < 2 {
+		outputHelp()
+		return
+	}
+	switch os.Args[1] {
```
freehan, line 46: use flags package to handle this.

freehan: authentication mode: KSA, K8s user

freehan: infra mode: 1. normal VM 2. GCE VM.

cmd/workload-daemon/app/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+type WorkloadInfo interface {
+	// Name is the name of the workload
+	Name() string
+	// Hostname is the hostname or DNS address of the workload
+	Hostname() string
+	// IP is the IP used to access this workload from the cluster
+	IP() string
+	// Labels are one or more labels associated with the workload
+	Labels() map[string]string
+	// Locality associated with the endpoint. A locality corresponds to a failure domain.
```
freehan, line 31: Use K8s standardized locality:
Region 
Zone

cmd/workload-daemon/app/interfaces.go
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+package app
+
+import "k8s.io/client-go/rest"
+
+// WorkloadInfo represents basic information of this workload
+type WorkloadInfo interface {
+	// Name is the name of the workload
+	Name() string
```
freehan, line 24: return value might be (string, bool).

Boolean aims to express if this value applies or not.

You need to distinguish 3 cases:  nil, empty, some value.

cmd/workload-daemon/app/gce-vm.go
```diff
@@ -0,0 +1,283 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+	"k8s.io/ingress-gce/cmd/workload-daemon/metadata"
+	"k8s.io/klog"
+)
+
+// CloudVM represents a VM instance running on Google Cloud.
+// It uses the metadata server to fetch all required information.
+type CloudVM struct {
```
freehan, line 39: pkg/experiemental/workload/daemon/provider/gce/vm.go

CloudVM -> VM 

comment: VM implements XXX interface for GCE VMs.

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func (c *Controller) deleteWorkload(obj interface{}) {
+	workload := getWorkloadFromDeleteAction(obj)
+	if workload != nil {
+		c.addWorkload(workload)
+	}
+}
+
+func (c *Controller) processService(key string) error {
```
freehan, line 254: may worth adding unit test for processService.

pkg/experimental/workload/controller.go
zjkmxy: There is no TargetPort in EndpointPort or EndpointSlices.

cmd/workload-controller/main.go
```diff
@@ -0,0 +1,104 @@
...
+	if _, err := crdHandler.EnsureCRD(workloadCRDMeta); err != nil {
+		klog.Fatalf("Failed to ensure Workload CRD: %v", err)
+	}
+	workloadClient, err := workloadclient.NewForConfig(kubeConfig)
+	if err != nil {
+		klog.Fatalf("Failed to create Workload client: %v", err)
+	}
+
+	ctx := workload.NewControllerContext(kubeClient, workloadClient, flags.F.WatchNamespace, flags.F.ResyncPeriod)
+	// TODO: Leader Elect and Health Check?
```
freehan, line 88: Leader election + liveness probe

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,418 @@
...
+	}, stopCh)
+
+	klog.V(2).Infof("Starting workload controller")
+	defer func() {
+		klog.V(2).Infof("Shutting down workload controller")
+		c.stop()
+	}()
+
+	go wait.Until(c.serviceWorker, time.Second, stopCh)
+	<-stopCh
```
freehan, line 139: run a sepearte goroutine periodically to ping workloads

https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/controller.go#L278

```
go wait.Until(c.ExecutePing, flags.PingInterval, stopCh)
```

freehan: Option 1:
```
ExecutePing(){
step1: c.workloadLister.List all workloads
step2: kick off pingWorkload for each
     // https://gobyexample.com/waitgroups
     parallel = 5 (flags.paralellPing, default=5)
     for workload in allWorkloads: {
            go c.pingWorkload(workload)
     }

}


c.pingWorkload(workload) {
      Step1:      Ping workload
      Step2:     patch workload condition (Ready, Ping) when there is a transition
}
```

pkg/experimental/workload/controller.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+}
+
+func (c *Controller) updateWorkload(old, cur interface{}) {
+	oldWorkload := old.(*workloadv1a1.Workload)
+	newWorkload := cur.(*workloadv1a1.Workload)
+	if newWorkload.ResourceVersion == oldWorkload.ResourceVersion {
+		return
+	}
+
```
freehan, line 244: Need to skip processing when all conditions are the same status

pkg/experimental/workload/controller.go
freehan: Option2
```
add pingQueue. // ratelimitingQueue

// Controller Run
go wait.Until(c.TriggerPeriodicPing, flags.PingInterval, stopCh)

parallel = 5 (flags.paralellPing, default=5)
for i in parallel {
     go pingWorker() 
}




TriggerPeriodicPing(){
step1: c.workloadLister.List all workloads
step2: push workload into pingQueue. 
}


c.pingWorker{
    step1: dequeue workload key from pingQueue
    Step2:      Ping workload
    Step3:     patch workload condition (Ready, Ping) when there is a transition
}



On workload Add
add workload key into pingQueue


```

freehan: Need to consider readiness in processService

PR #1084, opened 2020-04-22
Add e2e test for configuring healthchecks via BackendConfig

bowei:

PR #1220, opened 2020-08-18
Cherry-pick  #1209 [Expose custom health check ports in the firewall] into release-1-9

spencerhance:
Also includes the commit from #1084 in order to cherry-pick in the e2e tests as well

### DIFF
--- a/pkg/firewalls/controller.go
+++ b/pkg/firewalls/controller.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


rramkumar1:
@spencerhance Let's keep the cherry pick clean. Can you send a separate PR for the test.

spencerhance:
@rramkumar1 Sure thing.  Should I just include 007bf05 and not include 54885c3 from the original PR?

rramkumar1:
The cherrypick should only contain the same commits from the corresponding PR which was merged to HEAD.

spencerhance:
Hmm okay.  I don't see a reasonable way to cherry-pick both commits from #1209 without including bowei's e2e stuff.  I can just cherry-pick 007bf05 and not include the other commit in this PR.

PR #1221, opened 2020-08-19
Add e2e test for HTTPS Redirects

spencerhance:
/assign @freehan


pkg/fuzz/features/https_redirects.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	v.expectedResponseCode = code
+
+	return nil
+}
+
+// CheckResponse implements fuzz.FeatureValidator.
+func (v *HTTPSRedirectsFeature) CheckResponse(host, path string, resp *http.Response, body []byte) (fuzz.CheckResponseAction, error) {
+	if v.expectedResponseCode != 0 && resp.Request.URL.Scheme == "http" {
+		if resp.StatusCode == v.expectedResponseCode {
```
freehan, line 83: Probably better to return an explicit error? 

Otherwise it goes thru the checking if response is 200

cmd/e2e-test/redirects_test.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+			if err := cert.Create(s); err != nil {
+				t.Fatalf("Error creating cert %s: %v", cert.Name, err)
+			}
+			certs = append(certs, cert)
+
+			defer cert.Delete(s)
+		}
+
+		for _, tc := range []struct {
```
freehan, line 55: I would recommend making this test into a transition tests. All steps can go into the same e2e test case and you would be able to test a more well rounded cases

For instance, 
Step 1. create a ingress without HttpsRedirect
Step 2. modify the ingress with HttpsRedirect
Step 3. modify the ingress without HttpsRedirect

WDYT?

cmd/e2e-test/redirects_test.go
```diff
@@ -0,0 +1,114 @@
...
+			t.Logf("Ingress created (%s/%s)", s.Namespace, ing.Name)
+
+			ing, err = e2e.WaitForIngress(s, ing, nil)
+			if err != nil {
+				t.Fatalf("Error waiting for Ingress to stabilize: %v", err)
+			}
+			t.Logf("GCLB resources created (%s/%s)", s.Namespace, ing.Name)
+
+			// Perform whitebox testing.
+			gclb, err := e2e.WhiteboxTest(ing, s, Framework.Cloud, "")
```
freehan, line 102: did you include the redirect URLmap+forwarding rule validtion in the whiteboxTest?

PR #1223, opened 2020-08-20
Fix internal ingress static IP bug

spencerhance:
We mistakenly provided the cluster subnet for the forwarding rule, which prevents customers
from being able to use a static ip from any subnet in the same VPC/Region as the cluster


spencerhance:
/assign @freehan 

Minhan, I'm assigning to you since you reviewed the original code.  Feel free to reassign.

PR #1224, opened 2020-08-24
Increase NEG Poll Timeout and use subtests in ASM tests

spencerhance:
This should reduce ASM test failures and make them easier to debug

### DIFF
--- a/cmd/e2e-test/asm_test.go
+++ b/cmd/e2e-test/asm_test.go
@@ -13,7 +13,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/ingress-gce/pkg/e2e"
-	"k8s.io/klog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/e2e/helpers.go
```diff
@@ -67,7 +67,7 @@ const (
 	gclbDeletionTimeout = 60 * time.Minute
 
 	negPollInterval  = 5 * time.Second
-	negPollTimeout   = 2 * time.Minute
```
freehan, line 70: do you see instance where the NEG poll timed out?

freehan: can we try to bump it to 3 min instead?

5 mins seems to be too long

cmd/e2e-test/asm_test.go
```diff
@@ -189,46 +188,46 @@ func TestASMServiceAndDestinationRule(t *testing.T) {
 				{desc: "NEG controller should update NEGs for destinationrule", destinationRuleName: "porter-destinationrule", subsetEndpointCountMap: map[string]int{"v1": 1, "v2": 2}, crossNamespace: false},
 				{desc: "NEG controller should create NEGs for cross namespace destinationrule", destinationRuleName: "porter-destinationrule-1", subsetEndpointCountMap: map[string]int{"v1": 1}, crossNamespace: true},
 			} {
-				t.Logf("Running test case: %s", tc.desc)
```
freehan, line 192: This is suppose to be a transition test. I am not sure if we should split it.

PR #1227, opened 2020-08-25
Update ensureRedirectUrlMap() logic

spencerhance:
- Fix bug where v1 namer is used and empty string ends up in the get call
- Remove get call unless feature is enabled

(running e2e test)
/hold

cc: @skmatti 
/assign @freehan

### DIFF
--- a/pkg/loadbalancers/url_maps.go
+++ b/pkg/loadbalancers/url_maps.go
@@ -21,6 +21,7 @@ import (
 
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/events"
 	"k8s.io/ingress-gce/pkg/flags"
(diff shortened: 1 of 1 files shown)


spencerhance:
/unhold

e2e test passed

pkg/loadbalancers/url_maps.go
```diff
@@ -96,15 +97,11 @@ func (l *L7) ensureRedirectURLMap() error {
...
-	if utils.IgnoreHTTPNotFound(err) != nil {
-		return err
-	}
-
 	// Do not expect to have a RedirectUrlMap
 	if expectedMap == nil {
-		if currentMap == nil {
+		// Check if we need to GC
+		status, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
+		if !ok || status == "" || name == "" {
```
skmatti, line 104: Can we add `|| !namerSupported` here?.

compute.DeleteUrlMap will definitely fail if we are using v1 naming scheme. So, it may be okay to skip deletion if the namer is not supported.

spencerhance: Sure, in practice name will always be an empty string when namerSupported = false.  But it's probably best not to rely on that.

spencerhance: done

pkg/loadbalancers/url_maps.go
```diff
@@ -96,15 +97,11 @@ func (l *L7) ensureRedirectURLMap() error {
...
-	if utils.IgnoreHTTPNotFound(err) != nil {
-		return err
-	}
-
 	// Do not expect to have a RedirectUrlMap
 	if expectedMap == nil {
-		if currentMap == nil {
+		// Check if we need to GC
+		status, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
+		if !ok || status == "" || name == "" || !namerSupported {
```
freehan, line 104: why does it matter if your namer is supporting or not?
you should GC if it has the annotation?

pkg/loadbalancers/url_maps.go
```diff
@@ -123,6 +120,11 @@ func (l *L7) ensureRedirectURLMap() error {
 		return fmt.Errorf("error: cannot enable HTTPS Redirects with the V1 Ingress naming scheme.  Please recreate your ingress to use the newest naming scheme.")
 	}
 
+	currentMap, err := composite.GetUrlMap(l.cloud, key, l.Versions().UrlMap)
```
freehan, line 123: namerSupported needs to be considered when you expect a url map right?

pkg/loadbalancers/url_maps.go
spencerhance: If the namer isn't supported, name will be an empty string and the delete will error out anyway

Also since this doesn't exist in prod, I don't see anyway that a v1 namer with a redirect url map would get created.

pkg/loadbalancers/url_maps.go
spencerhance: yes, but it is necessary for the delete as well in this case since it will return an empty string and cause a json error

pkg/loadbalancers/url_maps.go
spencerhance: @skmatti Sure.  I do think that logic is easier to read.  I just pushed the change.

pkg/loadbalancers/url_maps.go
spencerhance: This logic is improved in the above comment thread

pkg/loadbalancers/url_maps.go
```diff
@@ -91,20 +92,29 @@ func (l *L7) ensureRedirectURLMap() error {
...
+	key, err := l.CreateKey(name)
+	if err != nil {
 		return err
 	}
 
 	// Do not expect to have a RedirectUrlMap
 	if expectedMap == nil {
-		if currentMap == nil {
+		// Check if we need to GC
+		status, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
```
freehan, line 116: do you wipe out the annotation if the ingress does not have it in the Annotation sync? 

If not, will it cause a GC loop? where DeleteUrlMap will return error?

spencerhance: Okay fixed

PR #1233, opened 2020-08-26
Add missing features to ingress metrics

spencerhance:
These features have been implemented but are currently not included in the metrics
- SSLPolicy
- Regional Static IP
- Custom Health Checks

This PR adds metrics for them
All of them can be cherry-picked to v1.9 + v1.10

HTTPS Redirects will be added in a follow up since it is currently 1.10 only

### DIFF
--- a/pkg/metrics/features.go
+++ b/pkg/metrics/features.go
@@ -21,6 +21,8 @@ import (
 	"strconv"
 
 	"k8s.io/api/networking/v1beta1"
+	"k8s.io/ingress-gce/pkg/annotations"
+	frontendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/frontendconfig/v1beta1"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog"
 )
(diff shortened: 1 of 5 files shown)


spencerhance:
/assign @prameshj 

Pavithra, any chance you have bandwidth to review this?

prameshj:
Would it be better to split into 2 PRs - one with metrics that can be cherrypicked to 1.9 and one with those that cannot?

pkg/metrics/metrics_test.go
```diff
@@ -484,8 +500,8 @@ var (
 					Namespace: defaultNamespace,
 					Name:      "ingress13",
 					Annotations: map[string]string{
-						StaticIPNameKey: "user-spec-static-ip",
-						staticIPKey:     "user-spec-static-ip",
+						StaticGlobalIPNameKey: "user-spec-static-ip",
+						staticIPKey:           "user-spec-static-ip",
```
prameshj, line 504: nit: should this value be an ip address?

spencerhance: Both of the annotations represent the name of the address, rather than the address itself

spencerhance:
@prameshj Good call!  I removed redirects from this.  Once it is merged I'll submit a follow up PR.

pkg/metrics/features.go
```diff
@@ -191,15 +201,24 @@ func featuresForIngress(ing *v1beta1.Ingress) []feature {
 	if val, ok := ingAnnotations[staticIPKey]; ok && val != "" {
```
prameshj, line 201: Does this mean user has to specify both staticIPKey and StaticGlobalIPNameKey in order to use a given ip address name? Why do we need 2?

spencerhance: oh, no staticIPKey is used as a status annotation

pkg/metrics/features.go
```diff
@@ -191,15 +201,28 @@ func featuresForIngress(ing *v1beta1.Ingress) []feature {
...
 			klog.V(6).Infof("User specified static IP for ingress %s: %s", ingKey, val)
 			features = append(features, specifiedStaticGlobalIP)
 		} else {
 			features = append(features, managedStaticGlobalIP)
 		}
 	}
+
+	// Check for regional static IP
+	// We do this separately from the global static IP because regional static IP
+	// does not support the StaticIPKey annotation or a managed regional IP
```
prameshj, line 215: Can you rephrase this to say - "Controller does not populate StaticIPKey annotation when processing Regional static IP"?

Also, these annotations are also defined in annotations/ingress.go. Do they need to be redefined in features.go?

spencerhance: @skmatti what was the reason for this?  Was there a circular dependency?

spencerhance: Updated the comment

skmatti: This was to enable backporting for the metrics PR to the release branches that do not have the annotations defined. Maybe it does not apply for these metrics.

spencerhance: Okay I removed the annotation

pkg/metrics/features.go
```diff
@@ -191,15 +199,27 @@ func featuresForIngress(ing *v1beta1.Ingress) []feature {
...
 			klog.V(6).Infof("User specified static IP for ingress %s: %s", ingKey, val)
 			features = append(features, specifiedStaticGlobalIP)
 		} else {
 			features = append(features, managedStaticGlobalIP)
 		}
 	}
+
+	// Check for regional static IP
+	// We do this separately from the global static IP because Controller does not
+	// populate StaticIPKey annotation when processing Regional static IP regional static IP
```
prameshj, line 213: remove

spencerhance: done

PR #1234, opened 2020-08-26
fix TestASM test flakes

freehan:
cc @skmatti @spencerhance

### DIFF
--- a/cmd/e2e-test/asm_test.go
+++ b/cmd/e2e-test/asm_test.go
@@ -162,8 +162,7 @@ func TestASMServiceAndDestinationRule(t *testing.T) {
 					}
 				} else {
 					if negName, ok := negStatus.NetworkEndpointGroups[strconv.Itoa(int(porterPort))]; ok {
-						// No backend pod exists, so the NEG has 0 endpoint.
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/asm_test.go
```diff
@@ -162,8 +162,7 @@ func TestASMServiceAndDestinationRule(t *testing.T) {
 					}
 				} else {
 					if negName, ok := negStatus.NetworkEndpointGroups[strconv.Itoa(int(porterPort))]; ok {
-						// No backend pod exists, so the NEG has 0 endpoint.
-						if err := e2e.WaitForNegs(ctx, Framework.Cloud, negName, negStatus.Zones, false, 0); err != nil {
+						if err := e2e.WaitForNegs(ctx, Framework.Cloud, negName, negStatus.Zones, false, 6); err != nil {
 							t.Errorf("Failed to wait Negs, error: %s", err)
```
prameshj, line 166: Can we derive this from number of NEG ports configured? or some other const?

freehan: oh. this is the total number of replicas in the above deployments

freehan: oh sorry. i take that back

PR #1248, opened 2020-09-05
Fix a couple bugs in the redirects e2e test

spencerhance:
During the transition test refactor, a couple bugs were introduced that caused the test to not ensure the Ingress and FrontendConfig correctly.


freehan (approved):
/lgtm
/approve

PR #1249, opened 2020-09-05
Fix GC for RedirectURLMaps

spencerhance:
We currently cannot rely on the annotation as a way to confirm that the
url map has been deleted.  This will cause us to leak Redirect URL Maps

### DIFF
--- a/pkg/loadbalancers/url_maps.go
+++ b/pkg/loadbalancers/url_maps.go
@@ -113,8 +113,8 @@ func (l *L7) ensureRedirectURLMap() error {
 	// Do not expect to have a RedirectUrlMap
 	if expectedMap == nil {
 		// Check if we need to GC
-		status, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
-		if !ok || status == "" {
+		_, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
+		if !ok {
 			return nil
 		} else {
 			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); err != nil {


spencerhance:
Forgot to run gofmt, re-pushed

pkg/loadbalancers/url_maps.go
```diff
@@ -110,14 +110,10 @@ func (l *L7) ensureRedirectURLMap() error {
...
+	// GC Redirect URL Map
 	if expectedMap == nil {
-		// Check if we need to GC
-		status, ok := l.ingress.Annotations[annotations.RedirectUrlMapKey]
-		if !ok || status == "" {
-			return nil
-		} else {
-			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); err != nil {
+		if frontendconfig.HasHTTPSRedirectsAndDisabled(feConfig) {
+			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); !utils.IsNotFoundError(err) {
```
freehan, line 116: does this means that it always trigger DeleteUrlMap if an ingress does not want to have HTTPsRedirect?

freehan: How do you avoid unnecessary deletes

spencerhance: Yes that's correct.  If they have explicitly disabled then we attempt GC.  I don't think we can avoid unnecessary deletes or gets without storing some state.  Previously we were using the annotation, but the annotation gets deleted before we do the GC.

spencerhance: It's possible to refactor it, but I think this code is much easier to read for now at least.

spencerhance:
After some more testing I can't seem to reproduce the original issue I was seeing.  I'm gonna close this for now as I don't actually think there is a bug.

PR #1250, opened 2020-09-05
Refactor transition e2e tests to use t.Run()

spencerhance:
Many serial transition tests loop through a variety of test cases but do not explicitly run them with t.Run().  This makes debugging more difficult since it's hard to tell in which case the test failed.


RoundNutz4 (approved):

freehan (approved):
/lgtm
/approve

spencerhance:
@freehan Rebased changes

Issue #1252, opened 2020-09-06
Changing class from gce-internal to gce leaves orphaned load balancers


carlpett:
Hey,
I created an ingress with `ingress.class: gce-internal`. After testing a bit, I changed the class to `gce` to compare the latency. After a while, a new public load balancer was provisioned, however the old internal one still remains, even after removing the ingress object completely.
(The new public load balancer didn't seem to work either, just returning Google 404:s, but that may be something on my end)

When deleting the load balancer manually, I'm prompted to delete other unused items:
>Regional back-end services:
>- [ ] k8s1-1d6454aa-kube-system-default-http-backend-80-86b603b9
>- [ ] k8s1-1d6454aa-alerting-karma-8080-a8103f1f
>
>Regional health checks:
>- [ ] k8s1-1d6454aa-alerting-karma-8080-a8103f1f
>- [ ] k8s1-1d6454aa-kube-system-default-http-backend-80-86b603b9
>
>Regional SSL certificates: 
>- [ ] k8s2-cr-ocxcn8r1-xoxpd4e1u0q3cjps-1e6575d27cab05e5

Are all of these safe to delete? The "karma" ones relate to my app, so those seem safe, but the default backend? The certificate is not viewable in the console (I can search for it and find it, but get an error when trying to display it), so I can't tell if it is related to the default backend or if it synced the certificate for my ingress into a GCP certificate?

spencerhance:
Hey @carlpett,

Is your cluster still in that state?  If you email me your project/cluster info I can take a look.

carlpett:
Hey!
I had actually cleaned it up, but could reproduce it. I'll mail you the details!

For completeness, here is the repro yamls:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: gke-test-ingress
  annotations:
    cloud.google.com/neg: '{"ingress": true}'
  labels:
    app: foo
spec:
  type: ClusterIP
  selector:
    app: foo
  ports:
  - name: http
    port: 8080
    targetPort: http
    protocol: TCP
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: gke-test-ingress
  annotations:
    kubernetes.io/tls-acme: "true"
    kubernetes.io/ingress.class: gce-internal
  labels:
    app: foo
spec:
  rules:
  - host: ingress-test.my-domain.tld
    http:
      paths:
      - path: /
        backend:
          serviceName: gke-test-ingress
[...]

PR #1259, opened 2020-09-15
Fix GC for switching between LB Scopes on Namer V2

spencerhance:
The V2 GC currently doesn't support GC of duplicate frontend resources from a different scope

cc: @skmatti for V2 namer

/assign @freehan 

Issue: #1252

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -47,6 +47,7 @@ import (
 	"k8s.io/ingress-gce/pkg/healthchecks"
 	"k8s.io/ingress-gce/pkg/instances"
 	"k8s.io/ingress-gce/pkg/loadbalancers"
+	"k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"k8s.io/ingress-gce/pkg/metrics"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
 	ingsync "k8s.io/ingress-gce/pkg/sync"
@@ -303,7 +304,7 @@ func NewLoadBalancerController(
 	return &lbc
 }
 
-// Init the controller.
+// Init the controller
 func (lbc *LoadBalancerController) Init() {
 	// TODO(rramkumar): Try to get rid of this "Init".
 	lbc.instancePool.Init(lbc.Translator)
@@ -489,8 +490,8 @@ func (lbc *LoadBalancerController) GCv1LoadBalancers(toKeep []*v1beta1.Ingress) 
 }
 
 // GCv2LoadBalancer implements Controller.
-func (lbc *LoadBalancerController) GCv2LoadBalancer(ing *v1beta1.Ingress) error {
-	return lbc.l7Pool.GCv2(ing)
+func (lbc *LoadBalancerController) GCv2LoadBalancer(ing *v1beta1.Ingress, scope meta.KeyType) error {
+	return lbc.l7Pool.GCv2(ing, scope)
 }
 
 // EnsureDeleteV1Finalizers implements Controller.
@@ -550,12 +551,13 @@ func (lbc *LoadBalancerController) sync(key string) error {
 
 	// Capture GC state for ingress.
 	allIngresses := lbc.ctx.Ingresses().List()
+	scope := features.ScopeFromIngress(ing)
 
 	// Determine if the ingress needs to be GCed.
 	if !ingExists || utils.NeedsCleanup(ing) {
-		frontendGCAlgorithm := frontendGCAlgorithm(ingExists, ing)
+		frontendGCAlgorithm := frontendGCAlgorithm(ingExists, false, ing)
 		// GC will find GCE resources that were used for this ingress and delete them.
-		err := lbc.ingSyncer.GC(allIngresses, ing, frontendGCAlgorithm)
+		err := lbc.ingSyncer.GC(allIngresses, ing, frontendGCAlgorithm, scope)
 		// Skip emitting an event if ingress does not exist as we cannot retrieve ingress namespace.
 		if err != nil && ingExists {
 			klog.Errorf("Error in GC for %s/%s: %v", ing.Namespace, ing.Name, err)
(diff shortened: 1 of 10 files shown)


pkg/controller/controller.go
```diff
@@ -551,11 +552,28 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
+			scope = features.ScopeFromIngress(ing)
+		}
+	} else {
+		scope = meta.Global
+	}
+
 	// Determine if the ingress needs to be GCed.
-	if !ingExists || utils.NeedsCleanup(ing) {
-		frontendGCAlgorithm := frontendGCAlgorithm(ingExists, ing)
+	if !ingExists || utils.NeedsCleanup(ing) || oldScope != nil {
```
skmatti, line 573: The sync returns w/o ensuring the LB resources for the latest scope after entering the if condition. 

This seems like a special case where ensure is followed by a GC. Or is it okay to perform the GC after ensuring the resources for the current scope?.

pkg/controller/controller.go
```diff
@@ -805,6 +824,9 @@ func frontendGCAlgorithm(ingExists bool, ing *v1beta1.Ingress) utils.FrontendGCA
 	if !utils.NeedsCleanup(ing) {
 		// GC backends only if current ingress does not need cleanup and finalizers is enabled.
 		if flags.F.FinalizerAdd {
+			if scopeChange {
+				return utils.CleanupV2FrontendResources
```
skmatti, line 828: may want to use a separate algorithm `utils.CleanupV2FrontendResourcesScopeChange`

pkg/controller/controller.go
spencerhance: I think in this case it doesn't really matter.  This is an incredibly disruptive operation so I don't think there's any expectation that we will ensure before GC or vice versa.  Either one will happen in the following sync loop anyway.

pkg/controller/controller.go
spencerhance: Done

pkg/controller/controller.go
skmatti: If we don't have preference order for ensure and GC, we might want to do the GC at line 626.

That way we ensure that
1. the LB for new scope before deleting the old resources
2. that create(for new scope) and GC happen in the same sync loop. The current logic seem to rely on a future sync to do that which seems unnecessary.

pkg/sync/sync.go
```diff
@@ -64,13 +65,23 @@ func (s *IngressSyncer) Sync(state interface{}) error {
...
+
+		defer func() {
+			if err != nil {
+				return
+			}
+			err = s.controller.EnsureDeleteV2Finalizer(currIng)
+		}()
+	case utils.CleanupV2FrontendResourcesScopeChange:
+		klog.V(3).Infof("Using algorithm CleanupV2FrontendResourcesScopeChange to GC frontend of ingress %s", common.NamespacedName(currIng))
+		lbErr = s.controller.GCv2LoadBalancer(currIng, scope)
```
skmatti, line 84: We should not delete finalizer for this case.

spencerhance: Done

pkg/controller/controller.go
spencerhance: Makes sense, just pushed the change

freehan (commented):
/ok-to-test

pkg/loadbalancers/l7s.go
```diff
@@ -108,16 +108,48 @@ func (l *L7s) list(key *meta.Key, version meta.Version) ([]*composite.UrlMap, er
...
+	if ing == nil {
+		return nil, nil
+	}
+
+	namer := l.namerFactory.Namer(ing)
+	urlMapName := namer.UrlMap()
+	currentScope := features.ScopeFromIngress(ing)
+
+	for _, scope := range []meta.KeyType{meta.Global, meta.Regional} {
+		key, err := composite.CreateKey(l.cloud, urlMapName, scope)
```
freehan, line 134: Create the key after `scope != currentScope` check?

pkg/loadbalancers/l7s.go
```diff
@@ -108,16 +108,48 @@ func (l *L7s) list(key *meta.Key, version meta.Version) ([]*composite.UrlMap, er
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Look for existing LBs with the same name but of a different scope
+		if scope != currentScope {
+			_, err := composite.GetUrlMap(l.cloud, key, features.VersionsFromIngress(ing).UrlMap)
```
freehan, line 141: I am wondering if there is any other way to check for existing scope without calling GCE API? 

I am okay with this current approach but seems a bit excessive.

freehan: One idea is to record the scope as an annotation.

pkg/controller/controller.go
```diff
@@ -795,7 +815,8 @@ func (lbc *LoadBalancerController) ensureFinalizer(ing *v1beta1.Ingress) (*v1bet
 //    - Does not need cleanup
 //      - Finalizer enabled    :    all backends
 //      - Finalizer disabled   :    v1 frontends and all backends
-func frontendGCAlgorithm(ingExists bool, ing *v1beta1.Ingress) utils.FrontendGCAlgorithm {
+//      - Scope changed        :    v2 frontends and all backends
```
skmatti, line 818: may be replace w/ "v2 frontends for old scope"?

pkg/loadbalancers/l7s.go
```diff
@@ -108,16 +108,48 @@ func (l *L7s) list(key *meta.Key, version meta.Version) ([]*composite.UrlMap, er
...
 		return err
 	}
 	klog.V(2).Infof("GCv2(%v) ok", ingKey)
 	return nil
 }
 
+// FrontendScopeChangeGC returns the scope to GC if the LB has changed scopes
+// (e.g. when a user migrates from ILB to ELB on the same ingress or vice versa.)
+// This only applies to the V2 Naming Scheme
+func (l *L7s) FrontendScopeChangeGC(ing *v1beta1.Ingress) (*meta.KeyType, error) {
```
skmatti, line 124: We need logging in this function. At least when the resources for old scope exists.

pkg/controller/controller.go
```diff
@@ -551,11 +552,18 @@ func (lbc *LoadBalancerController) sync(key string) error {
 	// Capture GC state for ingress.
 	allIngresses := lbc.ctx.Ingresses().List()
 
+	var scope meta.KeyType
+	if ing != nil {
+		scope = features.ScopeFromIngress(ing)
+	} else {
+		scope = meta.Global
+	}
```
skmatti, line 560: nit: can we simplify this ?
```
scope := meta.Global
if ing != nil {
	scope = features.ScopeFromIngress(ing)
}
```

pkg/loadbalancers/l7s.go
spencerhance: Done

pkg/controller/controller.go
spencerhance: Done.  Technically the backends will get deleted as their GC loop is not affected by this bug

pkg/loadbalancers/l7s.go
spencerhance: Done

pkg/controller/controller.go
spencerhance: Done.  Improved the annotation parsing logic downstream.

pkg/loadbalancers/l7s.go
spencerhance: I don't love the GCE API call either, but I like the idea of storing the previous ingress state as an annotation even less.  It seems awkward and confusing to the customer, and if it's out of sync it will potentially create a controller GC deadlock.  

We could adapt our current resource annotations to support a list of resources (e.g. UrlMaps) so that users and the controller know exactly what resources are still alive.  But that seems like a larger refactor.

freehan: Yeah. Then add a TODO to avoid calling GCE API every sync.

spencerhance: Done

freehan (approved):
/lgtm
/approve

PR #1263, opened 2020-09-20
Fix SSL Certs Leak in ILB Tests

spencerhance:
Previously Regional Certs were not getting deleted.

/assign @freehan


spencerhance:
test failed due to gofmt, should be fixed now

@freehan can you re-lgtm?

PR #1264, opened 2020-09-21
Ensure neg crs for all desired ports

swetharepakula:
- if neg crs are incorrectly deleted, next resync on service will
 recreate the neg cr

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -119,6 +119,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 
 	removes := currentPorts.Difference(newPorts)
 	adds := newPorts.Difference(currentPorts)
+	samePorts := newPorts.Difference(adds)
 	// There may be duplicate ports in adds and removes due to difference in readinessGate flag
 	// Service/Ingress config changes can cause readinessGate to be turn on or off for the same service port.
 	// By removing the duplicate ports in removes and adds, this prevents disruption of NEG syncer due to the config changes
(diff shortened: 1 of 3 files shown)


freehan (approved):
just one nit: 
remove TODO here: https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/manager.go#L490

/ok-to-test

pkg/neg/manager.go
```diff
@@ -141,13 +142,24 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 		}
 	}
 
+	for _, portInfo := range samePorts {
+		// To reduce the possibility of NEGs being leaked, ensure a SvcNeg CR exists for every
+		// desired port.
+		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
+			errList = append(errList, err)
+			continue
```
freehan, line 150: you do not need this continue?

freehan (commented):
One more thing to consider, because of the `needInit`, the NegObjectRefs may not be updated when the NegCR gets recreated 
Might worth updating this to include a boolean return or return a sepearte NegCr update error to indicate if the NegCR updates are successful or not, then decide if the syncer should set needInit=false
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction.go#L154

swetharepakula:
There will be some latency between the CR being created and the NegObjectRefs being updated, which could still cause a leak of Negs if the Neg is deleted before those are populated.

We are currently setting `needInit=true` separately when the syncer notices that the NegObjectRefs or if the Initialization conditions is missing : https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction.go#L517-L526.

pkg/neg/manager.go
swetharepakula: The continue is unnecessary, I will remove it. Thanks for catching it!

freehan:
> We are currently setting needInit=true separately when the syncer notices that the NegObjectRefs or if the Initialization conditions is missing : https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction.go#L517-L526. Instead of having the decision there, do you think it would be better to consolidate in one place?

I see. That looks good.

freehan (approved):
/lgtm
/approve

PR #1266, opened 2020-09-22
ensure ingress controller always generate consistent uid and provider uid

freehan:
This would ensure that only one ingress controller can create the `ingress-uid` config map on bootstrap.  This ensures that the provider-uid and uid are consistent.


MrHohn (approved):
Thanks for the fix.
/lgtm
/approve

PR #1267, opened 2020-09-22
skip opening firewall for node port ranges when unnecessary

freehan:


pkg/firewalls/firewalls.go
```diff
@@ -77,7 +77,10 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	sort.Strings(targetTags)
 
 	// De-dupe ports
-	ports := sets.NewString(fr.portRanges...)
+	ports := sets.NewString()
+	if allowNodePort {
+		ports.Insert(fr.portRanges...)
```
rramkumar1, line 82: should we rename this var to nodePortRanges to be more clear?

freehan: yeah. sounds good

rramkumar1 (commented):
/lgtm

PR #1276, opened 2020-09-25
Cherry Pick #1266 [ensure ingress controller always generate consistent uid and provider uid]

freehan:


prameshj:
Same change as https://github.com/kubernetes/ingress-gce/pull/1266

/lgtm

PR #1269, opened 2020-09-23
Fix NEG handling for default backend service when it contains the NEG annotation

freehan:
When the default backend service has the NEG annotation, NEG controller currently does not create NEGs for default backend if a L7 XLB needs it. 

cc: @spencerhance @swetharepakula

PR #1277, opened 2020-09-25
Cherry Pick #1269 [fix NEG handling for default backend service when it contains the NEG annotation] to 1.9

freehan:


pkg/neg/controller.go
```diff
@@ -536,8 +536,28 @@ func (c *Controller) mergeDefaultBackendServicePortInfoMap(key string, portInfoM
...
+	if err != nil {
+		return err
+	}
+	if !foundNEGAnnotation {
+		return nil
+	}
+	if negAnnotation.Ingress == false {
+		return nil
+	}
+	return scanIngress(utils.IsGCEIngress)
```
prameshj, line 560: If the input is an L7 ILB ingress, we would insert to portInfoMapTwice right? Is that expected?

freehan: Good catch. The portInfoMap would dedup it.

PR #1282, opened 2020-10-01
Improve the logic that updates ingress resource annotations

skmatti:
There is a race condition in the controller that causes
the user defined annotations to get overwritten in some
edge cases. The controller retrives the ingress resource at https://github.com/kubernetes/ingress-gce/blob/d9a00eafd11df479ded327c697241971e57b691e/pkg/controller/controller.go#L637
and updates the annotations at
https://github.com/kubernetes/ingress-gce/blob/d9a00eafd11df479ded327c697241971e57b691e/pkg/controller/controller.go#L728.
All the user updated annotations during this time period
will be lost.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -17,7 +17,6 @@ limitations under the License.
 package controller
 
 import (
-	context2 "context"
 	"fmt"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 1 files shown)


freehan (commented):
/lgtm
/approve

PR #1285, opened 2020-10-02
only ignore not found error when getting NEG

freehan:
this is to prevent the small chance of GCE Get NEG API returns a non not found error.


freehan:
cc @swetharepakula

PR #1287, opened 2020-10-06
Improve update annotations workflow for L4 ILB service

skmatti:
This ensures that same service version is used for
populating new annotations and update the resource
in API server. Fixes a race condition where some of
the existing annotations overwritten.

### DIFF
--- a/pkg/l4/l4controller.go
+++ b/pkg/l4/l4controller.go
@@ -186,7 +186,7 @@ func (l4c *L4Controller) processServiceCreateOrUpdate(key string, service *v1.Se
 	}
 	l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeNormal, "SyncLoadBalancerSuccessful",
 		"Successfully ensured load balancer resources")
-	if err = l4c.updateAnnotations(service.Name, service.Namespace, l4.MergeAnnotations(service, annotationsMap)); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/l4/l4controller.go
```diff
@@ -289,20 +289,36 @@ func (l4c *L4Controller) updateServiceMetadata(svc *v1.Service, newObjectMetadat
...
-	return nil
+	klog.V(3).Infof("Updating annotations of service %v/%v", namespace, name)
+	updatedObjectMeta := currSvc.ObjectMeta.DeepCopy()
+	updatedObjectMeta.Annotations = newAnnotations
+	return l4c.updateServiceMetadata(currSvc, *updatedObjectMeta)
 }
+
+// mergeAnnotations merges the new set of ilb resource annotations with the pre-existing service annotations.
+// Existing ILB resource annotation values will be replaced with the values in the new map.
+func mergeAnnotations(svc *v1.Service, ilbAnnotations map[string]string) map[string]string {
```
prameshj, line 310: Do we use the same approach for L7? Can we have a single util function that takes in new annotations as well as a list of annotations to remove and replace?

skmatti: We use a slightly different different approach for l7 https://github.com/kubernetes/ingress-gce/blob/cad3b2c63fab423ce97240a2be481fd5b8505c7a/pkg/loadbalancers/l7.go#L463

Probably difficult to integrate both into a single method. Also, l7 uses patch to update the annotations at the moment but this one does not.

PR #1288, opened 2020-10-06
Check for specific error events

swetharepakula:
- Checking for specific error messages will reduce flaking due to other
 error events being emitted

### DIFF
--- a/pkg/e2e/helpers.go
+++ b/pkg/e2e/helpers.go
@@ -864,8 +864,8 @@ func CheckNegFinalizer(svcNeg negv1beta1.ServiceNetworkEndpointGroup) error {
 	return nil
 }
 
-// WaitForSvcNegErrorEvents waits for  msgs messages present for namespace:name ConfigMap until timeout.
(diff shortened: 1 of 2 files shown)


pkg/e2e/helpers.go
```diff
@@ -876,19 +876,26 @@ func WaitForSvcNegErrorEvents(s *Sandbox, svcName string, numMessages int) error
...
+						break
+					}
+				}
 			}
 		}
 
-		if errEvents != numMessages {
-			klog.Infof("WaitForSvcNegErrorEvents(), expected to find %d error events, found %d", numMessages, errEvents)
+		if len(foundMessages) != len(expectedMessages) {
+			klog.Infof("WaitForSvcNegErrorEvents(), expected to find %d error events, found %d", len(expectedMessages), len(foundMessages))
```
freehan, line 898: I would recommend just print the events so that it is easier to debug.

swetharepakula: That's a good suggestion! Updated the PR with that.

freehan (approved):
/lgtm
/approve

PR #1293, opened 2020-10-10
Fix bug in composite funcs without full regional support

spencerhance:
Both ```SetSslCertificateForTargetHttpsProxy()``` and  ```SetUrlMapForTargetHttpProxy()``` had not been updated once regional beta and ga support was added.  This causes failures when refreshing certs or updating the url map.

This will definitely need to be cherry-picked into release-1.9


freehan (approved):
/lgtm
/approve

PR #1297, opened 2020-10-14
add logging in NEG GC

freehan:
cc @swetharepakula

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -436,6 +436,9 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
 	var errList []error
 	for _, cr := range deletionCandidates {
 		shouldDeleteNegCR := true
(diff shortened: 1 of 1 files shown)


pkg/neg/manager.go
```diff
@@ -436,6 +436,9 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
 	var errList []error
 	for _, cr := range deletionCandidates {
 		shouldDeleteNegCR := true
+		if len(cr.Status.NetworkEndpointGroups) == 0 {
```
prameshj, line 439: is Status guaranteed to be non-empty?

freehan: I think so 
https://github.com/kubernetes/ingress-gce/blob/master/pkg/apis/svcneg/v1beta1/types.go#L35

PR #1303, opened 2020-10-21
Mitigate NEG GC Race and add Logging

swetharepakula:
* Grab lock before deleting NEG CR to ensure that neg is not desired
* Verify description before deleting NEG to ensure NEG is controlled by this cluster
* Delete using name and zone when refs are not populated
* Improve logging and error messages

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -485,9 +485,26 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
 			continue
 		}
 
-		if err := deleteSvcNegCR(manager.svcNegClient, cr); err != nil {
-			errList = append(errList, err)
-		}
+		func() {
+			manager.mu.Lock()
+			defer manager.mu.Unlock()
+
+			// Verify that the NEG is still not wanted before deleting the CR. Mitigates the possibility of the race
+			// condition mentioned above
+			svcKey := getServiceKey(cr.Namespace, cr.GetLabels()[negtypes.NegCRServiceNameKey])
+			portInfoMap := manager.svcPortMap[svcKey]
+			for _, portInfo := range portInfoMap {
+				if portInfo.NegName == cr.Name {
+					klog.V(2).Infof("NEG CR %s/%s is still desired, skipping deletion", cr.Namespace, cr.Name)
+					return
+				}
+			}
+
+			klog.V(2).Infof("Deleting NEG CR %s/%s", cr.Namespace, cr.Name)
+			if err := deleteSvcNegCR(manager.svcNegClient, cr); err != nil {
+				errList = append(errList, err)
+			}
+		}()
 	}
 
 	return utilerrors.NewAggregate(errList)
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -147,7 +147,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 		// To reduce the possibility of NEGs being leaked, ensure a SvcNeg CR exists for every
 		// desired port.
 		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-			errList = append(errList, err)
+			errList = append(errList, fmt.Errorf("failed to ensure svc neg cr %s/%s/%d for existing port: %s", namespace, portInfo.NegName, portInfo.PortTuple.Port, err))
 		}
 	}
 
@@ -161,7 +161,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 			// syncer for the NEG until the NEG CR is successfully created. This will reduce the
 			// possibility of invalid states and reduces complexity of garbage collection
 			if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-				errList = append(errList, err)
+				errList = append(errList, fmt.Errorf("failed to ensure svc neg cr %s/%s/%d for new port: %s ", namespace, portInfo.NegName, svcPort.ServicePort, err))
 				continue
 			}
 
@@ -335,8 +335,10 @@ func (manager *syncerManager) ensureDeleteSvcNegCR(namespace, negName string) er
 	neg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
 
 	if neg.GetDeletionTimestamp().IsZero() {
-		err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Delete(context.Background(), negName, metav1.DeleteOptions{})
-		return err
+		if err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Delete(context.Background(), negName, metav1.DeleteOptions{}); err != nil {
+			return fmt.Errorf("errored while deleting neg cr %s/%s: %s", negName, namespace, err)
+		}
+		klog.V(2).Infof("Deleted neg cr %s/%s", negName, namespace)
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/neg/manager.go
```diff
@@ -147,7 +147,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 		// To reduce the possibility of NEGs being leaked, ensure a SvcNeg CR exists for every
 		// desired port.
 		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-			errList = append(errList, err)
+			errList = append(errList, fmt.Errorf("Failed to ensure svc neg cr %s/%s/%d for existing port: %s", namespace, portInfo.NegName, portInfo.PortTuple.Port, err))
```
freehan, line 150: lower case for initial char in error message

pkg/neg/manager.go
```diff
@@ -161,7 +161,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 			// syncer for the NEG until the NEG CR is successfully created. This will reduce the
 			// possibility of invalid states and reduces complexity of garbage collection
 			if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-				errList = append(errList, err)
+				errList = append(errList, fmt.Errorf("Failed to ensure svc neg cr %s/%s/%d for new port: %s ", namespace, portInfo.NegName, svcPort.ServicePort, err))
```
freehan, line 164: lower case for initial char in error message

pkg/neg/manager.go
```diff
@@ -434,48 +436,90 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
...
 			if err != nil {
 				errList = append(errList, fmt.Errorf("failed to parse selflink for neg cr %s/%s: %s", cr.Namespace, cr.Name, err))
 				continue
 			}
 
-			if err := manager.ensureDeleteNetworkEndpointGroup(resourceID.Key.Name, resourceID.Key.Zone); err != nil {
-				err = fmt.Errorf("failed to delete NEG %s in %s: %s", resourceID.Key.Name, resourceID.Key.Zone, err)
-				manager.recorder.Eventf(cr, v1.EventTypeWarning, negtypes.NegGCError, err.Error())
-				errList = append(errList, err)
+			shouldDeleteNegCR = deleteNegOrReportErr(resourceID.Key.Name, resourceID.Key.Zone, shouldDeleteNegCR, cr)
```
freehan, line 474: how about this?
```
shouldDeleteNegCR = shouldDeleteNegCR && deleteNegOrReportErr(resourceID.Key.Name, resourceID.Key.Zone,  cr)
```

pkg/utils/negdescription.go
```diff
@@ -49,3 +49,21 @@ func NegDescriptionFromString(descString string) (*NegDescription, error) {
...
+// is provided, VerifyDescription will return true.
+func VerifyDescription(descString, clusterUID, namespace, svcName, port, negName string) bool {
+	// Return true if description string is empty
+	if descString != "" {
+		desc, err := NegDescriptionFromString(descString)
+		if err != nil {
+			klog.Warningf("Error unmarshalling Neg Description %s err:%s", negName, err)
+		} else {
+			if desc.ClusterUID != clusterUID || desc.Namespace != namespace || desc.ServiceName != svcName || desc.Port != port {
+				return false
```
freehan, line 64: Should this just throw an detail error out? so that the caller can get why the description does not match. 

for instance, `fmt.Errorf("expect description of NEG object %q/%q to be %v, but got %v", zone, name, expectDesc, desc)`

pkg/neg/manager.go
```diff
@@ -434,48 +436,90 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
...
+			manager.mu.Lock()
+			defer manager.mu.Unlock()
+
+			// Verify that the NEG is still not wanted before deleting the CR. Mitigates the possibility of the race
+			// condition mentioned above
+			svcKey := getServiceKey(cr.Namespace, cr.GetLabels()[negtypes.NegCRServiceNameKey])
+			portInfoMap := manager.svcPortMap[svcKey]
+			for _, portInfo := range portInfoMap {
+				if portInfo.NegName == cr.Name {
+					return
```
freehan, line 498: break?

pkg/neg/manager.go
```diff
@@ -335,8 +335,10 @@ func (manager *syncerManager) ensureDeleteSvcNegCR(namespace, negName string) er
 	neg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
 
 	if neg.GetDeletionTimestamp().IsZero() {
-		err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Delete(context.Background(), negName, metav1.DeleteOptions{})
-		return err
+		if err = manager.svcNegClient.NetworkingV1beta1().ServiceNetworkEndpointGroups(namespace).Delete(context.Background(), negName, metav1.DeleteOptions{}); err != nil {
+			return fmt.Errorf("Errored while deleting neg cr %s/%s: %s", negName, namespace, err)
```
freehan, line 339: lower case for initial char

pkg/neg/manager.go
swetharepakula: I think this should be a return, so it will return out of the function and go to the next deletion candidate.

I was considering making this its own method, but there are already many deleteResource or ensureDeleteResource that I think adding one more will cause the code to be more confusing.

pkg/neg/syncers/utils_test.go
```diff
@@ -1079,6 +1079,169 @@ func TestNegObjectCrd(t *testing.T) {
...
+			subnetwork:     diffSubnetwork,
+			negType:        negtypes.NonGCPPrivateEndpointType,
+			negDescription: "",
+			expectRecreate: false,
+			expectError:    false,
+		},
+	}
+
+	for _, tc := range testCases {
+		fakeCloud := negtypes.NewFakeNetworkEndpointGroupCloud(testSubnetwork, testNetwork)
```
freehan, line 1197: Use the other newer mock gce cloud. 

https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/manager_test.go#L99-L101

freehan (commented):
/ok-to-test

pkg/neg/manager.go
freehan: add a log line before the return and record the decision

pkg/neg/syncers/utils.go
```diff
@@ -132,6 +132,17 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 		(!utils.EqualResourceIDs(neg.Network, cloud.NetworkURL()) ||
 			!utils.EqualResourceIDs(neg.Subnetwork, cloud.SubnetworkURL())) {
 
+		expectedDesc := utils.NegDescription{
```
freehan, line 135: does it make sense to move the description verification all together to the top? 
so that it eliminates the duplicated verification below.

pkg/neg/syncers/utils.go
```diff
@@ -126,34 +126,31 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
...
+			ClusterUID:  kubeSystemUID,
+			Namespace:   svcNamespace,
+			ServiceName: svcName,
+			Port:        port,
+		}
+		if matches, err := utils.VerifyDescription(expectedDesc, neg.Description, negName, zone); !matches {
+			klog.Errorf("Neg Name %s is already in use: %s", negName, err)
+			return negv1beta1.NegObjectReference{}, fmt.Errorf("neg name %s is already in use, found conflicting description: %s", negName, err)
+		} else if networkEndpointType != negtypes.NonGCPPrivateEndpointType &&
```
freehan, line 139: nit: no need to have `else if` here
just do `if`

freehan (commented):
/lgtm
/approve

freehan (commented):
/lgtm

PR #1307, opened 2020-10-28
Update go dependencies to v1.19

aledbf:


liggitt:
/close

in favor of https://github.com/kubernetes/ingress-gce/pull/1400

Issue #1301, opened 2020-10-16
Update to use Ingress and IngressClass networking.k8s.io/v1 APIs before v1.22


liggitt:
Ingress and IngressClass APIs earlier than networking.k8s.io/v1 are deprecated and scheduled to no longer be served in Kubernetes v1.22

The networking.k8s.io/v1 APIs are available starting in v1.19

/cc @robscott

swetharepakula:
This is still in progress.
/remove-lifecycle stale

robscott:
This is still in progress. 

/remove-lifecycle rotten

brondum:
any updates on supporting `spec.ingressClassName`?

visheyra:
I stumbled some time ago on this issue, and one of the google guy I can exchange with shared me this documentation https://cloud.google.com/kubernetes-engine/docs/how-to/custom-ingress-controller#controller_summary TL;DR you can configure ingressClass but you'll have to specify the annotation as well. This is working fine on a 1.21 cluster, I plan to test in 1.22 soon

swetharepakula:
This issue was concerning the APIs that ingress-gce was using to manage ingress resources. Since the controller now only uses the v1 API this issue can be closed.

Please open a new issue for `ingressClassName` support if still desired.

PR #1308, opened 2020-10-28
Ignore Ingresses that have spec.ingressClassName

swetharepakula:
- if annotation is present, controller will manage the ingress

cc/ @spencerhance

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -285,7 +285,10 @@ func IsGCEIngress(ing *v1beta1.Ingress) bool {
 
 	switch class {
 	case "":
-		return true
+		// Ingress controller does not have any ingress classes that can be
+		// specified by spec.IngressClassName. If spec.IngressClassName
+		// is nil, then consider GCEIngress.
+		return ing.Spec.IngressClassName == nil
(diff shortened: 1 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -285,7 +285,7 @@ func IsGCEIngress(ing *v1beta1.Ingress) bool {
 
 	switch class {
 	case "":
-		return true
+		return ing.Spec.IngressClassName == nil
```
aledbf, line 288: How we can start using IngressClassName instead of annotations?

spencerhance (approved):
lgtm.  We may want to add support for gce IngressClass sooner, but can be done separately.

pkg/utils/utils.go
swetharepakula: We will be supporting ingressClassName. #1301 exists to track. This PR is just to make sure not to operate on Ingresses that are meant for other controllers like Istio.

freehan, line 288: Nit: add a comment here to explain the reasoning

freehan:
LGTM overall

freehan (approved):
/lgtm
/approve

PR #1312, opened 2020-11-02
This PR fixes the /metrics from being exposed to the Internet by

vbannai:
having the /mterics and /healthz handlers server from a different server
listening on default port of 8081.

### DIFF
--- a/cmd/404-server-with-metrics/server-with-metrics.go
+++ b/cmd/404-server-with-metrics/server-with-metrics.go
@@ -36,6 +36,7 @@ import (
 
 var (
 	port              = flag.Int("port", 8080, "Port number to serve default backend 404 page.")
+	metricsPort       = flag.Int("metricsPort", 8081, "Port number to serve metrics for the default  backend 404 page.")
 	serverTimeout     = flag.Duration("timeout", 5*time.Second, "Time in seconds to wait before forcefully terminating the server.")
 	readTimeout       = flag.Duration("read_timeout", 10*time.Second, "Time in seconds to read the entire request before timing out.")
 	readHeaderTimeout = flag.Duration("read_header_timeout", 10*time.Second, "Time in seconds to read the request header before timing out.")
@@ -56,10 +57,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	server := newServer(hostName, *port)
+	server := newServer(hostName, *port, *metricsPort)
 	server.registerHandlers()
 	klog.Infof("Default 404 server is running with GOMAXPROCS(%d) on %s:%d\n", runtime.GOMAXPROCS(-1), hostName, *port)
 
+	// The main http server for handling NotFound and healthzrequests
 	go func() {
 		err := server.httpServer.ListenAndServe()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


vbannai:
Looks like the tests are passing. Can I ask for a LGTM?

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -153,17 +183,23 @@ func newServer(hostName string, port int) *server {
...
-		mux.HandleFunc("/shutdown", s.shutdownHandler())
+		httpMux.HandleFunc("/shutdown", s.shutdownHandler())
 	}
-	mux.HandleFunc("/healthz", s.healthzHandler())
-	mux.Handle("/metrics", promhttp.Handler())
 
-	s.mux = mux
-	s.httpServer.Handler = mux
+	// Register the healthz and metrics handlers with the metrics server
+	metricsMux.HandleFunc("/healthz", s.healthzHandler())
```
rramkumar1, line 197: I don't think you can register healthz on a different port. It has to be on port 8080 so LB health checks can pass for it

vbannai: In that case, the /healthz endpoint would also be visible from the internet. Is that okay?

rramkumar1: Yes that's acceptable

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -76,6 +78,23 @@ func main() {
 		}
 	}()
 
+	// The server handling the metrics and healthz requests
```
rramkumar1, line 81: "The server handling metrics"

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -218,10 +254,11 @@ func gracefulShutdown(s *server) {
 
 	select {
 	case interrupt := <-c:
-		klog.Infof("received interrupt, doing a graceful shutdown: %v \n", interrupt)
+		klog.Infof("received interrupt, doing a graceful shutdown of http and metrics servers: %v \n", interrupt)
 	case <-s.ctx.Done():
-		klog.Infof("received /shutdown message, doing a graceful shutdown: \n")
+		klog.Infof("received /shutdown message, doing a graceful shutdow of http and metrics serversn: \n")
```
rramkumar1, line 259: serversn -> servers

rramkumar1 (commented):
LGTM

cmd/404-server-with-metrics/server-with-metrics.go
rramkumar1: This comment doesn't look addressed?

ash2k:
Looks like /metrics has been removed but I can still access /healthz on my ingress. I wonder why?

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -153,17 +183,23 @@ func newServer(hostName string, port int) *server {
...
+	// Register the default notFoundHandler, healthz with the main http server
+	httpMux.HandleFunc("/", s.notFoundHandler())
 	// enable shutdown handler only for non-prod environments
 	if *isProd == false {
-		mux.HandleFunc("/shutdown", s.shutdownHandler())
+		httpMux.HandleFunc("/shutdown", s.shutdownHandler())
 	}
-	mux.HandleFunc("/healthz", s.healthzHandler())
-	mux.Handle("/metrics", promhttp.Handler())
+	httpMux.HandleFunc("/healthz", s.healthzHandler())
```
ash2k, line 195: Shouldn't this be on the `metricsMux` too?

ash2k: Ok, I found the comment above. It's still not clear why the endpoint has to live on the ingress' port. What if an application wants to user this path?

PR #1314, opened 2020-11-06
Delete unsupported spec in backendconfig resources on controller startup

skmatti:
This addresses an edge case where the controller crashes. The steps to
reproduce the crash:

1. Create a BackendConfig resource with malformed spec for a unsupported
   feature. (Ex. custom resource headers in v1.6)
2. Upgrade to a ingress version that supports this feature. The malformed
   backendconfig resource will cause the generated client library to panic
   and crash the controller.


pkg/backendconfig/validation.go
```diff
@@ -122,3 +127,89 @@ func validateLogging(beConfig *backendconfigv1.BackendConfig) error {
...
+// steps that lead to a crash,
+// 1. Create a BackendConfig resource with malformed spec for a unsupported
+// feature. (Ex. custom resource headers in v1.6)
+// 2. Upgrade to a ingress version that supports this feature. The malformed
+// backendconfig resource will cause the generated client library to panic
+// and crash the controller.
+// Note that beginning v1.9 the unsupported fields are automatically dropped so
+// we will not run into this problem.
+func OverrideUnsupportedSpec(dynamicClient dynamic.Interface) error {
+	unstructuredBCList, err := dynamicClient.Resource(CRDMeta().GetGroupVersionResource()).List(context.TODO(), meta_v1.ListOptions{})
```
liggitt, line 156: the logic below is pinning to v1 of the resource, I would hard-code this to v1 as well

pkg/backendconfig/validation.go
```diff
@@ -122,3 +127,89 @@ func validateLogging(beConfig *backendconfigv1.BackendConfig) error {
...
+		for _, specKey := range []overridableSpec{customResourceHeaders, logging, healthCheck} {
+			// Skip if overridable feature spec is not present.
+			oSpec, ok := typedBCSpec[string(specKey)]
+			if !ok {
+				continue
+			}
+
+			marshalledSpec, err := json.Marshal(oSpec)
+			if err != nil {
+				klog.Warningf("Failed to marshall %v: %v", oSpec, err)
```
liggitt, line 184: without logging the name/namespace, this is unlikely to be actionable

pkg/backendconfig/validation.go
```diff
@@ -122,3 +127,89 @@ func validateLogging(beConfig *backendconfigv1.BackendConfig) error {
...
+			case logging:
+				err = json.Unmarshal(marshalledSpec, &backendconfigv1.LogConfig{})
+			case healthCheck:
+				err = json.Unmarshal(marshalledSpec, &backendconfigv1.HealthCheckConfig{})
+			default:
+			}
+
+			if err != nil {
+				klog.V(0).Infof("Deleting malformed %s spec %s for backendconfig %s, unmarshall error: %v", specKey, marshalledSpec, bcKey, err)
+				unstructured.RemoveNestedField(ubc.Object, spec, string(specKey))
```
liggitt, line 200: consider accumulating patch operations that drop the problematic fields and submitting a patch (rather than an update) to avoid conflict errors if something made an unrelated change to the object

pkg/backendconfig/validation.go
```diff
@@ -122,3 +127,89 @@ func validateLogging(beConfig *backendconfigv1.BackendConfig) error {
...
+				needsUpdate = true
+			}
+		}
+
+		if needsUpdate {
+			if _, err := dynamicClient.Resource(CRDMeta().GetGroupVersionResource()).Namespace(ubc.GetNamespace()).Update(context.TODO(), &ubc, meta_v1.UpdateOptions{}); err != nil {
+				errs = append(errs, err)
```
liggitt, line 207: if err is NotFound, you can ignore it

pkg/backendconfig/validation.go
```diff
@@ -122,3 +132,120 @@ func validateLogging(beConfig *backendconfigv1.BackendConfig) error {
...
 }
+
+type overridableSpec string
+
+const (
+	// These features are added in v1.8+ versions which means that a malformed spec
+	// can be added for these in v1.7 and below. Upgrading to 1.8 with these
+	// malformed resource will crash the ingress controller.
+	// These values should match json tags in pkg/backendconfig/apis/v1/types.go
+	customResourceHeaders overridableSpec = "customResourceHeaders"
```
rramkumar1, line 143: I'm not sure how I feel about this. This means we will need to be reactive to bugs rather than proactive. Can we make this more generic?

rramkumar1: Ideally, we can make this more library-like that is agnostic to the exact resource. We can then plug the library into other controllers.

skmatti: Wanted to take the conservative approach and not nuke the config that should not be. 

Also, we are handling an extreme edge case and very hard to reproduce.

skmatti: Maybe it would be ideal to back-port pruning fix for the other controllers where it is possible

pkg/backendconfig/validation.go
skmatti: Done

pkg/backendconfig/validation.go
skmatti: Done

pkg/backendconfig/validation.go
skmatti: Done.

pkg/backendconfig/validation.go
skmatti: Done

pkg/backendconfig/validation.go
rramkumar1: I don't want to incrementally solve this problem though. Tomorrow, it will be some other small edge case which triggers the same issue.

If we have something generic that covers malformed resources then we can be done in one shot.

liggitt: The surgical field-level things this is doing would be hard to do generically. I would not be eager to introduce a controller that deletes objects it cannot decode. Creating an event or labeling the object (so `kubectl get ... -A -l somelabel` could locate them) or something similar might be acceptable.

rramkumar1: > Creating an event or labeling the object (so kubectl get ... -A -l somelabel could locate them) or something similar might be acceptable.

But we can't do this without doing some surgical delete as this PR does right? Otherwise, the informer will crash.

I suppose if there is no way to implement something generic, that's fine. I just wanted to 100% confirm that because we will want to do something similar for other controllers

skmatti: > it will be some other small edge case which triggers the same issue.
I don't think we will run into this for future features. Because those fields will be dropped in v1.9(GKE v1.16) as they are unsupported and pruning is enabled.

Also, I don't think we want to override the spec when we unintentionally release non-backward compatible changes.

bowei:
satish -- are you still working on this?

skmatti:
Closing the PR as we decided internally to not proceed with this approach.

PR #1315, opened 2020-11-06
Support for Path Type

swetharepakula:
- Add support for Exact, Prefix and ImplementationSpecific PathTypes
- Validates and reports errors for incorrect paths based on the specified path type
- empty path type defaults to ImplementationSpecific PathType


pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
 	return urlMap, errs
 }
 
+func validateAndGetPaths(path v1beta1.HTTPIngressPath) ([]string, error) {
```
freehan, line 261: add comment

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+		v1beta1.PathTypeImplementationSpecific: validateImplementationSpecificPathType,
+	}
+
+	if validateFunc, ok := validationFuncs[pathType]; ok {
+		return validateFunc(path)
+	}
+	return nil, fmt.Errorf("failed to process path %s: invalid path type %s", path.Path, pathType)
+}
+
+func validateExactPathType(path v1beta1.HTTPIngressPath) ([]string, error) {
```
freehan, line 279: add comment

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+		return nil, fmt.Errorf("failed to validate exact path type due to empty path")
+	}
+
+	if strings.Contains(path.Path, "*") {
+		return nil, fmt.Errorf("failed to validate exact path %s due to invalid wildcard", path.Path)
+	}
+	return []string{path.Path}, nil
+}
+
+func validateAndModifyPrefixPathType(path v1beta1.HTTPIngressPath) ([]string, error) {
```
freehan, line 290: add comment

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+		pathType = *path.PathType
+	}
+
+	validationFuncs := map[v1beta1.PathType]func(path v1beta1.HTTPIngressPath) ([]string, error){
+		v1beta1.PathTypeExact:                  validateExactPathType,
+		v1beta1.PathTypePrefix:                 validateAndModifyPrefixPathType,
+		v1beta1.PathTypeImplementationSpecific: validateImplementationSpecificPathType,
+	}
+
+	if validateFunc, ok := validationFuncs[pathType]; ok {
```
freehan, line 273: I am wondering if the validation is already done at the API server side?  Do we need to revalidate this?

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+	if path.Path == "/" {
+		return []string{"/*"}, nil
+	}
+
+	if strings.Contains(path.Path, "*") {
+		return nil, fmt.Errorf("failed to validate prefix path %s due to invalid wildcard", path.Path)
+	}
+
+	// Each prefix path should result in two paths for the URLMap, one without a
+	// trailing '/' and one that ends with '/*'
```
freehan, line 304: I am a bit confused on how this maps to URL map. Could use a bit more explanation here.

pkg/controller/translator/translator.go
```diff
@@ -214,13 +215,21 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
 				// The Ingress spec defines empty path as catch-all, so if a user
 				// asks for a single host and multiple empty paths, all traffic is
 				// sent to one of the last backend in the rules list.
-				path := p.Path
-				if path == "" {
```
freehan, line 218: It is possible to have a existing ingress object in this config to be interpreted differently here, right?

We need to be careful in this transition. Basically, this sorts of like a bug or limitation of the previous beta ingress API, but if many people rely on this behavior, it becomes a feature.

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,76 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+	}
+	return []string{path.Path}, nil
+}
+
+func validateAndModifyPrefixPathType(path v1beta1.HTTPIngressPath) ([]string, error) {
+	if path.Path == "" {
+		return nil, fmt.Errorf("failed to validate prefix path type due to empty path")
+	}
+
+	if path.Path == "/" {
```
freehan, line 295: we are translating user's input. Wondering why not just expose this an validation error?

pkg/controller/translator/translator.go
swetharepakula: I updated the description to be clearer. The prefix path for `/foo` should support requests like `/foo`,`/foo/`, and `/foo/bar`. For the URL Map to support that we need to submit two path rules `/foo` to cover `/foo` and `/foo/*` to cover the other requests.

freehan (commented):
I think we might be better off skipping all validation here.

pkg/controller/translator/translator.go
```diff
@@ -249,6 +258,91 @@ func (t *Translator) TranslateIngress(ing *v1beta1.Ingress, systemDefaultBackend
...
+// validateAndGetPaths will validate the path based on the specifed path type and will return the
+// the path rules that should be used. If no path type is provided, the path type will be assumed
+// to be ImplementationSpecific. If a non existent path type is provided, an error will be returned.
+func validateAndGetPaths(path v1beta1.HTTPIngressPath) ([]string, error) {
+	pathType := v1beta1.PathTypeImplementationSpecific
+	if path.PathType != nil {
+		pathType = *path.PathType
+	}
+
+	validationFuncs := map[v1beta1.PathType]func(path v1beta1.HTTPIngressPath) ([]string, error){
```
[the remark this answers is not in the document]
swetharepakula: yea i can change this into a switch statement instead and then we are more specific.

pkg/controller/translator/translator.go
swetharepakula: In general I do agree with that. For ImplementationSpecific I do think we should skip validation so the experience doesn't change. For Exact and Prefix match however the URL Map creation/update won't always fail if the user just puts an invalid path for that path type but still valid for the URL Map. Unfortunately the API server validation doesn't enforce check for invalid paths incorrect wildcard usage.

freehan (approved):
/lgtm
/approve

PR #1316, opened 2020-11-09
Prevent Empty Neg Refs on CRs

swetharepakula:
- Return error when fail to find neg to prevent empty neg ref being set on neg cr
- attempt to delete by zone, if self link is malformed


freehan (commented):
/lgtm
/approve

PR #1318, opened 2020-11-11
GCPIngressParams CRD

swetharepakula:


pkg/apis/ingparams/v1beta1/types.go
```diff
@@ -0,0 +1,54 @@
...
+	Spec   GCPIngressParamsSpec   `json:"spec,omitempty"`
+	Status GCPIngressParamsStatus `json:"status,omitempty"`
+}
+
+// GCPIngressParamsSpec is the spec for a GCPIngressParams resource
+// +k8s:openapi-gen=true
+type GCPIngressParamsSpec struct {
+	// External specifies whether external or internal load balancing is desired
+	// +required
+	External bool `json:"external"`
```
freehan, line 41: you probably want to add a comment and say the default is internal

swetharepakula: That's a good point! I will switch to match current behavior.

freehan (approved):
/lgtm
/approve

PR #1320, opened 2020-11-11
Cherry-pick #1312 [This PR fixes the /metrics from being exposed to the Internet...] into release-1-10

spencerhance:
/assign @rramkumar1 
cc: @vbannai 

Cherry-picking so this can be included in the next tag and pushed to k8s.gcr.io


ash2k:
Looks like `/metrics` has been removed but I can still access `/healthz` on my ingress. I wonder why?

PR #1325, opened 2020-11-16
remove EnableNonGCPMode flag from NEG controller

freehan:
I will push separate PRs for different flags.


pkg/neg/manager.go
```diff
@@ -87,23 +86,38 @@ type syncerManager struct {
 
 	// kubeSystemUID is used to by syncers when NEG CRD is enabled
 	kubeSystemUID types.UID
+
+	// indicates whether nonGcpMode have been enabled
```
rramkumar1, line 90: would add a comment too indicating what this actually implies in terms of how the controller operates

rramkumar1 (commented):
One nit

pkg/neg/manager.go
freehan: Done. Fixed

PR #1326, opened 2020-11-16
remove EnableNegCrd flag

freehan:
cc: @swetharepakula

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -102,9 +102,6 @@ type Controller struct {
 
 	// runL4 indicates whether to run NEG controller that processes L4 ILB services
 	runL4 bool
-
-	// indicates whether neg crd have been enabled
-	enableNegCrd bool
 }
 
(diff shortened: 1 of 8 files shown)


pkg/neg/controller.go
```diff
@@ -131,9 +127,8 @@ func NewController(
 		apiv1.EventSource{Component: "neg-controller"})
 
 	var svcNegInformer cache.Indexer
-	if enableNegCrd {
-		svcNegInformer = ctx.SvcNegInformer.GetIndexer()
-	}
+	svcNegInformer = ctx.SvcNegInformer.GetIndexer()
```
swetharepakula, line 130: nit: This could be moved below to the Controller initialization where the other informers are created.

swetharepakula (approved):
LGTM, one small nit that doesn't need to block merging this

pkg/neg/controller.go
freehan: Done Fixed.

PR #1328, opened 2020-11-17
Ensure ServiceAttachment CRD when PSC is enabled

swetharepakula:
- Add enable-psc flag
 - Add ServiceAttachment CRD Meta


freehan (approved):
/lgtm
/approve

PR #1339, opened 2020-12-09
Refactor SSLPolicy e2e tests

spencerhance:
-Change tests to transition test
-Re-use sslpolicy since we do not need to delete it
-General cleanup


freehan (commented):
/lgtm
/approve

PR #1348, opened 2020-12-21
Add ServiceAttachment naming scheme

swetharepakula:


pkg/utils/namer/serviceattachmentnamer.go
```diff
@@ -0,0 +1,79 @@
...
+// so modifications must be backwards compatible.
+func (n *V1ServiceAttachmentNamer) ServiceAttachment(namespace, name, saUID string) string {
+	clusterUID := common.ContentHash(n.kubeSystemUID, clusterUIDLength)
+	hash := n.suffix(8, n.kubeSystemUID, namespace, name, saUID)
+	truncFields := TrimFieldsEvenly(n.maxDescriptiveLabel, namespace, name)
+	return fmt.Sprintf("%s%s-sa-%s-%s-%s-%s", n.prefix, schemaVersionV1, clusterUID, truncFields[0], truncFields[1], hash)
+}
+
+// hash returns an 8 character hash code of the provided fields
+func (n *V1ServiceAttachmentNamer) suffix(numCharacters int, fields ...string) string {
```
freehan, line 76: There is a slightly better hash function that would has utilize full length of the input field. https://github.com/kubernetes/ingress-gce/blob/master/pkg/utils/namer/frontendnamer.go#L143


The current approach is more likely to get collisions as it only takes subset of the hash code

freehan (commented):
LGTM overall

pkg/utils/namer/serviceattachmentnamer.go
swetharepakula: Switched to the recommended hash function in the next version.

PR #1361, opened 2021-01-29
Crd events

swetharepakula:
Add CRD schemes so that event recorder can use them


freehan (commented):
Let us do a more surgical change to fix the scheme just for svcneg so that it is easier to cherry pick.


I do not think we ever record events on FrontendConfig and BackendConfigs.
The ingParams is okay for HEAD.

swetharepakula:
I will just keep the NEG CRD commit in this PR then. The rest of the changes can be dropped since we don't need the other CRDs don't have any events right now. We can add back as needed.

PR #1368, opened 2021-02-12
Service Attachment Creations and Update

swetharepakula:
ProcessServiceAttachment handles service attachment creations and updates.

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -1,0 +1,314 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package psc
+
+import (
+	context2 "context"
+	"fmt"
+	"net/http"
+	"reflect"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	alpha "google.golang.org/api/compute/v0.alpha"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/ingress-gce/pkg/annotations"
+	sav1alpha1 "k8s.io/ingress-gce/pkg/apis/serviceattachment/v1alpha1"
+	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/context"
+	serviceattachmentclient "k8s.io/ingress-gce/pkg/serviceattachment/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/ingress-gce/pkg/utils/patch"
+	sautils "k8s.io/ingress-gce/pkg/utils/serviceattachment"
+	"k8s.io/klog"
+	"k8s.io/legacy-cloud-providers/gce"
+)
+
+const (
+	svcKind = "service"
+)
+
+// Controller is a private service connect (psc) controller
+// It watches ServiceAttachment resources and creates, deletes, and manages
+// corresponding GCE Service Attachment resources
+type Controller struct {
+	client kubernetes.Interface
+
+	cloud    *gce.Cloud
+	saClient serviceattachmentclient.Interface
+
+	saNamer             namer.ServiceAttachmentNamer
+	svcAttachmentLister cache.Indexer
+	serviceLister       cache.Indexer
+	recorder            func(string) record.EventRecorder
+
+	hasSynced func() bool
+}
+
+func NewController(ctx *context.ControllerContext) *Controller {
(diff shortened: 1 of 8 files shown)


pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+	}
+	return nil
+}
+
+// getConnectionPreference returns the GCE connection preference based on the
+// provided connPreference
+func getConnectionPreference(connPreference string) (string, error) {
+	switch strings.TrimSpace(connPreference) {
+	case "acceptAutomatic":
```
freehan, line 318: let us keep it ver·ba·tim

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+	if !exists {
+		return "", fmt.Errorf("failed to get Service %s/%s: %q", namespace, svcName, err)
+	}
+
+	svc := obj.(*v1.Service)
+
+	frName, ok := svc.Annotations[annotations.TCPForwardingRuleKey]
+	if !ok {
+		if frName, ok = svc.Annotations[annotations.UDPForwardingRuleKey]; !ok {
+			frName = cloudprovider.DefaultLoadBalancerName(svc)
```
swetharepakula, line 191: add a comment about the decisions here

freehan, line 191: add a log line and comment herer

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	fwdRule, err := c.cloud.Compute().ForwardingRules().Get(context2.Background(), meta.RegionalKey(frName, c.cloud.Region()))
+	if err != nil {
+		return "", fmt.Errorf("failed to get Forwarding Rule %s: %q", frName, err)
+	}
+
+	foundMatchingIP := false
+	for _, ing := range svc.Status.LoadBalancer.Ingress {
+		if ing.IP == fwdRule.IPAddress {
+			foundMatchingIP = true
```
freehan, line 202: add a log line here

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+			frName = cloudprovider.DefaultLoadBalancerName(svc)
+		}
+	}
+	fwdRule, err := c.cloud.Compute().ForwardingRules().Get(context2.Background(), meta.RegionalKey(frName, c.cloud.Region()))
+	if err != nil {
+		return "", fmt.Errorf("failed to get Forwarding Rule %s: %q", frName, err)
+	}
+
+	foundMatchingIP := false
+	for _, ing := range svc.Status.LoadBalancer.Ingress {
```
freehan, line 200: double check for nil

swetharepakula, line 200: verify that LoadBalancer is not nil

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return "", fmt.Errorf("forwarding rule does not have matching IPAddr")
+}
+
+// getSubnetURLs will query GCE and gather all the URLs of the provided subnet names
+func (c *Controller) getSubnetURLs(subnets []string) ([]string, error) {
+
+	var subnetURLs []string
+	for _, subnetName := range subnets {
+		subnet, err := c.cloud.Compute().AlphaSubnetworks().Get(context2.Background(), meta.RegionalKey(subnetName, c.cloud.Region()))
```
freehan, line 218: Use GA Subnetwork calls

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
...
+		if err != nil {
+			return subnetURLs, fmt.Errorf("failed to get Subnetwork %s: %q", subnetName, err)
```
swetharepakula, line 220: make error more clear. add region information too

freehan, line 220: add region info in the subnet name

pkg/psc/controller.go
```diff
@@ -0,0 +1,330 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
swetharepakula, line 2: copyright should be 2021

pkg/psc/controller.go
swetharepakula: Verified LoadBalancer will not be empty not nil when not populated: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3991.

pkg/psc/controller.go
swetharepakula: Thanks for catching that!

pkg/psc/controller.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	existingSA, err := c.cloud.Compute().AlphaServiceAttachments().Get(context2.Background(), gceSAKey)
+	if err != nil && !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+		return fmt.Errorf("failed querying for GCE Service Attachment: %q", err)
+	}
+
+	if existingSA != nil {
+		klog.V(4).Infof("Found existing service attachment %s", existingSA.Name)
+		err = validateUpdate(existingSA, gceSvcAttachment)
```
freehan, line 152: I am wondering if it is worth validating update. If we just let it go to call GCE API, GCE will return validation error, correct? What would be the difference?

pkg/psc/controller.go
```diff
@@ -0,0 +1,337 @@
...
+
+	if !exists {
+		// Allow Garbage Collection to Delete Service Attachment
+		klog.V(2).Infof("Service attachment %s/%s does not exist in store. Will be cleaned up by GC", namespace, name)
+		return nil
+	}
+	klog.V(2).Infof("Processing Service attachment %s/%s", namespace, name)
+
+	svcAttachment := obj.(*sav1alpha1.ServiceAttachment)
+	connPref, err := getConnectionPreference(svcAttachment.Spec.ConnectionPreference)
```
freehan, line 110: I am wondering if we need to validate here. Can we just pass it verbatim and let GCE API validates it.

freehan: If a new type is introduced, we do not need to change code and release it to support it.

swetharepakula: that's a good point, i agree, we can remove this validation.

pkg/psc/controller.go
swetharepakula: That is true. I think we can do that in the future when there is an Update API. Currently since there is none in the Alpha API, I think we have to do the validation. Really it is to inform the user whatever they changed will not reflected.

swetharepakula:
Updated the PR so now we don't validate the connection preference. I updated the resource reference validation as well. Since the `Service` is in the `core` api group, the group has to either by empty or nil.

swetharepakula:
One additional update was made to service attachment informer creation in `context.go`.

PR #1370, opened 2021-02-19
Graduate Cloud Armor security policy to GA

skmatti:
Dependency changes in #1371

### DIFF
--- a/pkg/backends/features/features.go
+++ b/pkg/backends/features/features.go
@@ -44,7 +44,7 @@ var (
 	// versionToFeatures stores the mapping from the required API
 	// version to feature names.
 	versionToFeatures = map[meta.Version][]string{
-		meta.VersionBeta: {FeatureSecurityPolicy, FeatureHTTP2, FeatureL7ILB, FeatureVMIPNEG},
+		meta.VersionBeta: {FeatureHTTP2, FeatureL7ILB, FeatureVMIPNEG},
 	}
 	// TODO: (shance) refactor all scope to be above the serviceport level
 	scopeToFeatures = map[meta.KeyType][]string{
--- a/pkg/backends/features/securitypolicy.go
+++ b/pkg/backends/features/securitypolicy.go
@@ -21,9 +21,6 @@ import (
 
 	"k8s.io/klog"
 
-	computebeta "google.golang.org/api/compute/v0.beta"
-
-	gcecloud "github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"k8s.io/legacy-cloud-providers/gce"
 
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -3,19 +3,18 @@ module k8s.io/ingress-gce
 go 1.13
 
 require (
-	github.com/GoogleCloudPlatform/k8s-cloud-provider v0.0.0-20200609233750-ea56f58e187a
+	github.com/GoogleCloudPlatform/k8s-cloud-provider v0.0.0-20201230182714-6a1c9153df9b
```
bowei, line 6: can we tag this in upstream

pkg/backends/features/features.go
```diff
@@ -44,7 +44,7 @@ var (
 	// versionToFeatures stores the mapping from the required API
 	// version to feature names.
 	versionToFeatures = map[meta.Version][]string{
-		meta.VersionBeta: {FeatureSecurityPolicy, FeatureHTTP2, FeatureL7ILB, FeatureVMIPNEG},
+		meta.VersionBeta: {FeatureHTTP2, FeatureL7ILB, FeatureVMIPNEG},
```
bowei, line 47: Are there any issues with this transition that we should be watching for?
For example, I think will be left in the "Description" field, is that ok or not?

pkg/backends/features/securitypolicy.go
```diff
@@ -33,35 +30,28 @@ import (
 
 // EnsureSecurityPolicy ensures the security policy link on backend service.
 // TODO(mrhohn): Emit event when attach/detach security policy to backend service.
-func EnsureSecurityPolicy(cloud *gce.Cloud, sp utils.ServicePort, be *composite.BackendService, beName string) error {
-	if sp.BackendConfig.Spec.SecurityPolicy == nil {
+func EnsureSecurityPolicy(cloud *gce.Cloud, sp utils.ServicePort, be *composite.BackendService) error {
+	if be == nil || sp.BackendConfig.Spec.SecurityPolicy == nil {
```
bowei, line 34: Is this meant to be this way? Why would we call Ensure on a nil `be`?

pkg/backends/features/securitypolicy.go
```diff
@@ -33,35 +30,28 @@ import (
...
-	return nil
-}
 
-// securityPolicyNeedsUpdate checks if security policy needs update and
-// returns the desired policy reference.
-func securityPolicyNeedsUpdate(cloud *gce.Cloud, currentLink, desiredName string) (bool, *computebeta.SecurityPolicyReference) {
-	currentName, _ := utils.KeyName(currentLink)
-	if currentName == desiredName {
-		return false, nil
+	if desiredName == "" {
```
bowei, line 48: Keep this simple:

```
klog...Infof("Set security policy in backend service %s (%s:%s) to %q", ...)
```

I know the logic makes the logs a bit more semantically meaningful, but also is more annoying to maintain.

pkg/composite/composite.go
```diff
@@ -221,3 +221,33 @@ func SetProxyForForwardingRule(gceCloud *gce.Cloud, key *meta.Key, forwardingRul
 		}
 	}
 }
+
+// SetSecurityPolicy sets the cloud armor security policy for a backend service.
```
bowei, line 225: @spencerhance do we need to support regional? does this work for ILB

pkg/backends/features/features.go
skmatti: Nothing I can think of.

Old description fields are okay because we always treat them as the source of truth for versioning,

https://github.com/kubernetes/ingress-gce/blob/b23c5d068dca95a41a1aa5cfb4b6adbbf71e1cf1/pkg/backends/backends.go#L142-L148

And the latest change supports updating sec pol on alpha/ beta backend services.

pkg/backends/features/securitypolicy.go
skmatti: Don't think so. Removed it.

pkg/composite/composite.go
skmatti: From GCP public docs https://cloud.google.com/armor/docs/security-policy-overview, security policies are supported only for global backend services.

skmatti:
bowei@ Can you take another pass at this?. Thanks

pkg/composite/composite.go
```diff
@@ -221,3 +221,33 @@ func SetProxyForForwardingRule(gceCloud *gce.Cloud, key *meta.Key, forwardingRul
...
+func SetSecurityPolicy(gceCloud *gce.Cloud, backendService *BackendService, securityPolicy string) error {
+	key := meta.GlobalKey(backendService.Name)
```
bowei, line 227: We should check that the scope matches

```
if backendService.Scope != meta.GlobalScope { // error
```

pkg/backends/features/securitypolicy.go
```diff
@@ -33,35 +30,24 @@ import (
...
-	if currentName == desiredName {
-		return false, nil
-	}
-	var policyRef *computebeta.SecurityPolicyReference
-	if desiredName != "" {
-		policyRef = &computebeta.SecurityPolicyReference{
-			SecurityPolicy: gcecloud.SelfLink(meta.VersionBeta, cloud.ProjectID(), "securityPolicies", meta.GlobalKey(desiredName)),
-		}
+	klog.V(2).Infof("Set security policy in backend service %s (%s:%s) to %q", be.Name, sp.ID.Service.String(), sp.ID.Port.String(), desiredName)
+	if err := composite.SetSecurityPolicy(cloud, be, desiredName); err != nil {
```
bowei, line 49: Can we add the events and remove the TODO from MrHohn above?

pkg/backends/features/securitypolicy.go
```diff
@@ -33,35 +30,24 @@ import (
...
-	if !needsUpdate {
-		return nil
+	if scope := ScopeFromServicePort(&sp); scope != meta.Global {
+		return fmt.Errorf("cloud armor security policies not supported for %s backend service %s", scope, be.Name)
 	}
 
-	klog.V(2).Infof("Setting security policy %q for backend service %s (%s:%s)", policyRef, beName, sp.ID.Service.String(), sp.ID.Port.String())
-	if err := cloud.SetSecurityPolicyForBetaGlobalBackendService(beName, policyRef); err != nil {
-		return fmt.Errorf("failed to set security policy %q for backend service %s (%s:%s): %v", policyRef, beName, sp.ID.Service.String(), sp.ID.Port.String(), err)
+	existingName, _ := utils.KeyName(be.SecurityPolicy)
```
bowei, line 42: we should check the error

```
existingPolicyName, err := ...
if err != nil { 
}

desiredPolicyName :=
```

pkg/backends/features/securitypolicy.go
skmatti: Do you think we emit these events on the Ingress resource or as global events (not attached to any resources)?.

We update all the backend services in a ingress sync, so it may not be a good idea to emit event on the ingress.

Also, other backend config features or backend service operations do not emit events at the moment. Maybe we can handle this as a separate PR.

PR #1372, opened 2021-02-22
Add disruptive neg e2e tests

swetharepakula:
- test that no errors occur when processing service recreations occur
 before GC

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -555,3 +555,199 @@ func TestNegCRDErrorEvents(t *testing.T) {
 		e2e.WaitForStandaloneNegDeletion(ctx, Framework.Cloud, s, port80.String(), negStatus)
 	})
 }
+
+func TestNegDisruptive(t *testing.T) {
+	t.Parallel()
+	port80 := intstr.FromInt(80)
+	replicas := int32(2)
+	serviceName := "disruptive-neg-service"
+	// gcSvcName is the name of the service used to determine if GC has finished
+	gcSvcName := "gc-service"
+	ctx := context.Background()
+
+	annotation := annotations.NegAnnotation{
+		ExposedPorts: map[int32]annotations.NegAttributes{
+			int32(port80.IntValue()): annotations.NegAttributes{},
+		},
+	}
+
+	ensureGCService := func(s *e2e.Sandbox) annotations.NegStatus {
+		// use gc-service as a way to track if GC has completed or not
+		_, err := e2e.EnsureEchoService(s, "gc-service", map[string]string{
+			annotations.NEGAnnotationKey: annotation.String()}, v1.ServiceTypeClusterIP, replicas)
+		if err != nil {
+			t.Fatalf("error ensuring gc service: %v", err)
+		}
+		t.Logf("GC service ensured (%s/%s)", s.Namespace, gcSvcName)
+
+		expectedNegAttrs := map[string]string{port80.String(): ""}
+		negStatus, err := e2e.WaitForNegCRs(s, gcSvcName, expectedNegAttrs)
+		if err != nil {
+			t.Fatalf("Error: e2e.WaitForNegCRs(%s,%+v) = %s, want nil", gcSvcName, expectedNegAttrs, err)
+		}
+		for port, negName := range negStatus.NetworkEndpointGroups {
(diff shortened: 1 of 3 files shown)


cmd/e2e-test/neg_test.go
```diff
@@ -555,3 +555,79 @@ func TestNegCRDErrorEvents(t *testing.T) {
...
+		}
+
+		for port, negName := range negStatus.NetworkEndpointGroups {
+			err := e2e.WaitForNegs(ctx, Framework.Cloud, negName, negStatus.Zones, false, int(replicas))
+			if err != nil {
+				t.Fatalf("Error: e2e.WaitForNegs service %s/%s neg port/name %s/%s", serviceName, s.Namespace, port, negName)
+			}
+		}
+
+		errChan := make(chan error, 1)
```
freehan, line 593: does it have to be size 1 buffer?

cmd/e2e-test/neg_test.go
```diff
@@ -555,3 +555,79 @@ func TestNegCRDErrorEvents(t *testing.T) {
...
+		go func() {
+			errChan <- e2e.WaitForSvcNegErrorEvents(s, svc.Name, []string{"error processing service"})
+		}()
+
+	SERVICE_RECREATION:
+		for {
```
freehan, line 599: It it possible to add an extra timeout for this for loop?

cmd/e2e-test/neg_test.go
swetharepakula: No, you are right we don't need a buffered channel. I will remove it.

freehan (commented):
testCase struct{
desc string
deleteSvc bool
createSvc bool
waitForNegGC bool
waitForNEG bool
}

0. verify if there is events
1. deleteSvc
2. waitForNegGC
3. createSvc
4. waitForNeg


testCase 1: createSvc + waitForNeg
testCase 2: deleteSvc + createSvc + waitNeg
testCase 3: deleteSvc + createSvc 
testCase 4: deleteSvc + createSvc 
testCase 5: deleteSvc + createSvc + waitNeg
testCase 5: deleteSvc + createSvc + waitNeg
testCase 5: deleteSvc + createSvc + waitNeg
testCase 5: deleteSvc + createSvc + waitNeg
testCase 5: deleteSvc + createSvc + waitNeg
testCase 6: deleteSvc + createSvc 
testCase 7: deleteSvc + createSvc 
testCase 8: deleteSvc + waitNegGC

swetharepakula:
After discussing with @freehan, rewrote the test to be more deterministic and iterative. The test is clearer and should be easier to debug incase of failures.

cmd/e2e-test/neg_test.go
```diff
@@ -555,3 +555,195 @@ func TestNegCRDErrorEvents(t *testing.T) {
...
+			{
+				desc:          "delete service and recreate third time",
+				deleteService: true,
+				createService: true,
+			},
+			{
+				desc:       "check that neg still exists",
+				waitForNeg: true,
+			},
+			//waitForGC is to ensure that GC doesn't occur after deletiong and before the recreation
```
freehan, line 649: nit: deletiong

cmd/e2e-test/neg_test.go
```diff
@@ -555,3 +555,195 @@ func TestNegCRDErrorEvents(t *testing.T) {
...
+				for port, negName := range negStatus.NetworkEndpointGroups {
+					err := e2e.WaitForNegs(ctx, Framework.Cloud, negName, negStatus.Zones, false, int(replicas))
+					if err != nil {
+						t.Fatalf("Error: e2e.WaitForNegs service %s/%s neg port/name %s/%s", serviceName, s.Namespace, port, negName)
+					}
+				}
+				previousNegStatus = negStatus
+			}
+
+			if tc.checkErrorEvents {
```
freehan, line 735: How about list events associated with the service and go through to check if this event showed up?

I am asking because the event might be emitted before or after the check.

swetharepakula: That's a good point. Thanks for catching that!

swetharepakula: So I realized the code was already looking at events that may have been emitted before. But I think just listing once and querying through is better and quicker so still made the change.

PR #1373, opened 2021-02-23
Remove k8s.io/kubernetes dependencies in API packages

abursavich:
I would like to import your API packages without also taking in all of `k8s.io/kubernetes`. Thankfully, all dependencies are limited to the internal `k8s.io/kubernetes/pkg/apis/core` where the public `k8s.io/api/core/v1` is sufficient. This swaps the usage of the internal package with usage of the public package.


freehan:
cc @swetharepakula

swetharepakula:
This PR looks to be doing the same thing as #1366 minus the removal of the short name for Service Attachment

swetharepakula (approved):
Once this is merged I can removing the conflicting changes in #1366 and rebase on this patch. Otherwise LGTM

abursavich:
I went through the exercise to see what it would take to completely remove `k8s.io/kubernetes` and it's not that much... The main caveat is that you should probably wait until you upgrade to k8s v1.19, because the leaderelection flag code was moved to `k8s.io/component-base` in `v0.19.0`.

Broken into a series of small-ish commits: https://github.com/kubernetes/ingress-gce/compare/master...abursavich:remove-kk

PR #1374, opened 2021-02-24
Do not claim custom named negs with empty desc

swetharepakula:
* ensures that controller only manages custom named negs that were
 created by the neg controller
 * only delete custom named negs that have a populated description

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -198,6 +198,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				epc,
 				string(manager.kubeSystemUID),
 				manager.svcNegClient,
+				!manager.namer.IsNEG(portInfo.NegName),
 			)
 			manager.syncerMap[syncerKey] = syncer
 		}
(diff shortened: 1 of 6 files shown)


pkg/neg/manager.go
```diff
@@ -542,6 +543,10 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
 	}
 
 	if expectedDesc != nil {
+		if !manager.namer.IsNEG(name) && neg.Description == "" {
```
freehan, line 546: add a comment to describe the intention here.

freehan (commented):
just nits

pkg/neg/manager.go
swetharepakula: Added a comment to explain the reasoning.

pkg/neg/manager.go
```diff
@@ -542,6 +543,12 @@ func (manager *syncerManager) ensureDeleteNetworkEndpointGroup(name, zone string
 	}
 
 	if expectedDesc != nil {
+		// Controller managed custom named negs will always have a populated description, so do not delete custom named
+		// negs with empty descriptions.
+		if !manager.namer.IsNEG(name) && neg.Description == "" {
```
bowei, line 548: Shouldn't we check that 1) valid JSON 2) Looks like something we created (e.g. has at least the fields that we put in the NEG today).
instead of just empty.

swetharepakula: The next check on ln 552 checks for valid json and that the description has the expected fields. `VerifyDescription` returns true if the description is empty, hence the need to check specifically for custom name and empty description.

bowei: Ahh-- ok, I was reading too much into the title of the PR

freehan (approved):
/lgtm
/approve

PR #1377, opened 2021-03-10
Ignore temporary ProcessServiceFailed error events

swetharepakula:
- syncer shutting down or service not found (due to out of date cache)
 error events are temporary and can safely be ignored

/assign @freehan


freehan (approved):
/lgtm
/approve
/ok-to-test

PR #1381, opened 2021-03-15
Service Attachment Garbage Collection

swetharepakula:
- Add event scheme to event recorders

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -42,11 +42,18 @@ import (
 	"k8s.io/ingress-gce/pkg/utils/patch"
 	sautils "k8s.io/ingress-gce/pkg/utils/serviceattachment"
 	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/util/slice"
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
 const (
 	svcKind = "service"
+
+	// SvcAttachmentGCError is the service attachment GC error event reason
+	SvcAttachmentGCError = "ServiceAttachmentGCError"
+	// ServiceAttachmentFinalizer used by the psc controller to ensure Service Attachment CRs
+	// are deleted after the corresponding Service Attachments are deleted
+	ServiceAttachmentFinalizerKey = "networking.gke.io/service-attachment-finalizer"
 )
 
 // Controller is a private service connect (psc) controller
@@ -103,22 +110,26 @@ func (c *Controller) processServiceAttachment(key string) error {
 	klog.V(2).Infof("Processing Service attachment %s/%s", namespace, name)
 
 	svcAttachment := obj.(*sav1alpha1.ServiceAttachment)
-	if err = validateResourceReference(svcAttachment.Spec.ResourceRef); err != nil {
+	updatedCR, err := c.ensureSAFinalizer(svcAttachment)
+	if err != nil {
+		return fmt.Errorf("Errored adding finalizer on ServiceAttachment CR %s/%s: %s", namespace, name, err)
+	}
+	if err = validateResourceReference(updatedCR.Spec.ResourceRef); err != nil {
 		return err
 	}
 
-	frURL, err := c.getForwardingRule(namespace, svcAttachment.Spec.ResourceRef.Name)
+	frURL, err := c.getForwardingRule(namespace, updatedCR.Spec.ResourceRef.Name)
 	if err != nil {
 		return fmt.Errorf("failed to find forwarding rule: %q", err)
 	}
 
-	subnetURLs, err := c.getSubnetURLs(svcAttachment.Spec.NATSubnets)
+	subnetURLs, err := c.getSubnetURLs(updatedCR.Spec.NATSubnets)
 	if err != nil {
 		return fmt.Errorf("failed to find nat subnets: %q", err)
 	}
 
-	saName := c.saNamer.ServiceAttachment(namespace, name, string(svcAttachment.UID))
-	desc := sautils.ServiceAttachmentDesc{URL: svcAttachment.SelfLink}
+	saName := c.saNamer.ServiceAttachment(namespace, name, string(updatedCR.UID))
+	desc := sautils.ServiceAttachmentDesc{URL: updatedCR.SelfLink}
 	gceSvcAttachment := &alpha.ServiceAttachment{
 		ConnectionPreference:   svcAttachment.Spec.ConnectionPreference,
 		Name:                   saName,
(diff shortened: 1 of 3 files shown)


pkg/psc/controller.go
```diff
@@ -154,11 +163,46 @@ func (c *Controller) processServiceAttachment(key string) error {
...
+			klog.Errorf(eventMsg)
+			c.recorder(sa.Namespace).Eventf(sa, v1.EventTypeWarning, SvcAttachmentGCError, eventMsg)
+			continue
+		}
+		klog.V(2).Infof("Deleted Service Attachment %s", gceName)
+
+		klog.V(2).Infof("Removing finalizer on Service Attachment %s/%s", sa.Namespace, sa.Name)
+		if err := c.ensureSAFinalizerRemoved(sa); err != nil {
+			eventMsg := fmt.Sprintf("Failed to remove finalizer on ServiceAttachment %s/%s: %q", sa.Namespace, sa.Name, err)
+			klog.Errorf("eventMsg")
```
swetharepakula, line 198: change to eventMsg variable, not quoted

freehan, line 198: no quotes

pkg/psc/controller.go
```diff
@@ -154,11 +163,46 @@ func (c *Controller) processServiceAttachment(key string) error {
...
+func (c *Controller) garbageCollectServiceAttachments() {
+	klog.V(2).Infof("Staring Service Attachment Garbage Collection")
+	defer klog.V(2).Infof("Finished Service Attachment Garbage Collection")
+	crs := c.svcAttachmentLister.List()
+	for _, obj := range crs {
+		sa := obj.(*sav1alpha1.ServiceAttachment)
+		if sa.GetDeletionTimestamp().IsZero() {
+			continue
+		}
+		gceName := c.saNamer.ServiceAttachment(sa.Namespace, sa.Name, string(sa.UID))
```
swetharepakula, line 184: pull the name from the CR instead

freehan, line 184: use name from CR

pkg/utils/common/finalizer.go
```diff
@@ -39,6 +39,9 @@ const (
 	ILBFinalizerV2 = "gke.networking.io/l4-ilb-v2"
 	// NegFinalizerKey is the finalizer used by neg controller to ensure NEG CRs are deleted after corresponding negs are deleted
 	NegFinalizerKey = "networking.gke.io/neg-finalizer"
+	// ServiceAttachmentFinalizer used by the psc controller to ensure Service Attachment CRs
+	// are deleted after the corresponding Service Attachments are deleted
+	ServiceAttachmentFinalizerKey = "networking.gke.io/service-attachment-finalizer"
```
freehan, line 44: move this key to your package unless you depend on other utils of this package

pkg/psc/controller.go
```diff
@@ -154,11 +163,46 @@ func (c *Controller) processServiceAttachment(key string) error {
...
+			continue
+		}
+		klog.V(2).Infof("Deleted Service Attachment %s", gceName)
+
+		klog.V(2).Infof("Removing finalizer on Service Attachment %s/%s", sa.Namespace, sa.Name)
+		if err := c.ensureSAFinalizerRemoved(sa); err != nil {
+			eventMsg := fmt.Sprintf("Failed to remove finalizer on ServiceAttachment %s/%s: %q", sa.Namespace, sa.Name, err)
+			klog.Errorf("eventMsg")
+			c.recorder(sa.Namespace).Eventf(sa, v1.EventTypeWarning, SvcAttachmentGCError, eventMsg)
+			return
```
freehan, line 200: if you return here. You might miss GCing other SAs.

pkg/psc/controller.go
```diff
@@ -154,11 +163,46 @@ func (c *Controller) processServiceAttachment(key string) error {
...
+// garbageCollectServiceAttachments queries for all Service Attachments CR that have been marked
+// for deletion and will delete the corresponding GCE Service Attachment resource. If the GCE
+// resource has successfully been deleted, the finalizer is removed from the service attachment
+// cr.
+func (c *Controller) garbageCollectServiceAttachments() {
+	klog.V(2).Infof("Staring Service Attachment Garbage Collection")
+	defer klog.V(2).Infof("Finished Service Attachment Garbage Collection")
+	crs := c.svcAttachmentLister.List()
+	for _, obj := range crs {
+		sa := obj.(*sav1alpha1.ServiceAttachment)
```
freehan, line 180: package whatever executed inside the for loop into a separate function. 

This way, it would be easier to do error handling and avoid having mis handling

pkg/psc/controller.go
```diff
@@ -248,6 +292,65 @@ func (c *Controller) patchServiceAttachment(originalSA, updatedSA *sav1alpha1.Se
...
+	return c.saClient.NetworkingV1alpha1().ServiceAttachments(saCR.Namespace).Patch(context2.Background(), updatedCR.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+}
+
+// ensureSAFinalizerRemoved ensures that the Service Attachment finalizer is removed
+// from the provided CR.
+func (c *Controller) ensureSAFinalizerRemoved(cr *sav1alpha1.ServiceAttachment) error {
+	updatedCR := cr.DeepCopy()
+	updatedCR.Finalizers = slice.RemoveString(updatedCR.Finalizers, common.ServiceAttachmentFinalizerKey, nil)
+
+	patchBytes, err := patch.MergePatchBytes(cr, updatedCR)
```
freehan, line 345: warp up into one helper function for patching ServiceAttachment

freehan (commented):
One more point: think about how to avoid SA and LB leakage upon cluster deletion.

pkg/psc/controller.go
swetharepakula: That's a good point, thanks! Pulled the logic into a helper function.

pkg/psc/controller.go
swetharepakula: I changed the logic to pull the name from the CR, and fall back onto the generated name if there is some error in in the resource URL

pkg/psc/controller.go
swetharepakula: #1368 actually added a helper function to patchServiceAttachment that I forgot about. Current update uses that helper function now.

pkg/utils/common/finalizer.go
swetharepakula: Moved this into the psc package

swetharepakula:
Newest update cleans up the logic like extracting deletion into its own helper function. One notable change is how the service attachment name is found during GC. The name is now found using the service attachment url on the CR. If for some reason a name cannot be found, GC will fallback on generating the name from the CR.

freehan (approved):
/lgtm
/approve

PR #1384, opened 2021-03-17
update to use CRD v1

deads2k:
Fixes https://github.com/kubernetes/kubernetes/issues/100560

CRD v1beta1 is being removed in kubernetes 1.22, which starts development shortly.  This updates the GETs to be CRD v1 and tweaks the version check code to match.  From the godoc on CRD v1beta1 .spec.version: `Must match the name of the first item in the `versions` list if `version` and `versions` are both specified.`.

I haven't found these CRDs are defined.  It doesn't appear to be in this repo, but I'm concerned that they are also v1beta1.  Any hints?

@liggitt

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -200,7 +200,7 @@ func (ctx *ControllerContext) initEnableASM() {
 		ctx.ASMConfigController.DisableASM()
 		return
 	}
-	destinationRuleCRD, err := apiextensionClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context2.TODO(), destinationRuleCRDName, metav1.GetOptions{})
+	destinationRuleCRD, err := apiextensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context2.TODO(), destinationRuleCRDName, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			ctx.ASMConfigController.RecordEvent("Warning", "FailedValidateDestinationRuleCRD", "Cannot find DestinationRule CRD, disabling ASM Mode, please check Istio setup.")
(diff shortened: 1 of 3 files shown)


pkg/context/context.go
```diff
@@ -200,7 +200,7 @@ func (ctx *ControllerContext) initEnableASM() {
 		ctx.ASMConfigController.DisableASM()
 		return
 	}
-	destinationRuleCRD, err := apiextensionClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context2.TODO(), destinationRuleCRDName, metav1.GetOptions{})
+	destinationRuleCRD, err := apiextensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context2.TODO(), destinationRuleCRDName, metav1.GetOptions{})
```
liggitt, line 203: I found this one in https://github.com/istio/istio/blob/master/manifests/charts/base/crds/crd-all.gen.yaml

deads2k: > I found this one in https://github.com/istio/istio/blob/master/manifests/charts/base/crds/crd-all.gen.yaml

already on v1.  yay!

pkg/e2e/helpers.go
```diff
@@ -737,7 +737,7 @@ func WaitConfigMapEvents(s *Sandbox, namespace, name string, msgs []string, time
 // by the ingress controller.
 func waitForBackendConfigCRDEstablish(crdClient *apiextensionsclient.Clientset) error {
 	condition := func() (bool, error) {
-		bcCRD, err := crdClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), backendConfigCRDName, metav1.GetOptions{})
+		bcCRD, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), backendConfigCRDName, metav1.GetOptions{})
```
liggitt, line 740: this one looks defined here:

https://github.com/kubernetes/ingress-gce/blob/051cfde145f34c8721e39df778e262fb4a256c67/pkg/backendconfig/backendconfig.go#L39-L53

and created here:

https://github.com/kubernetes/ingress-gce/blob/051cfde145f34c8721e39df778e262fb4a256c67/pkg/crd/crd.go#L50-L96

liggitt:
this lgtm, needs an ack from the network folks

aojea:
@prameshj @robscott is this something you can take a look ?

pkg/context/context.go
```diff
@@ -210,9 +210,9 @@ func (ctx *ControllerContext) initEnableASM() {
 		ctx.ASMConfigController.DisableASM()
 		return
 	}
-	if destinationRuleCRD.Spec.Version != destinationRuleAPIVersion {
+	if destinationRuleCRD.Spec.Versions[0].Name != destinationRuleAPIVersion {
```
freehan, line 213: I am wondering if it should enumerate the Versions and check if v1alpha3 is available?

liggitt: that seems reasonable to do in a follow-up, this change is a strict translation of the existing checks to the v1 API

bowei: /hold can we resolve this

liggitt: > /hold can we resolve this

I'd recommend keeping this limited to a v1beta1 -> v1 switch, and put functional changes in a follow-up opened by maintainers with more context on this repo

skmatti (approved):

PR #1385, opened 2021-03-19
Run PSC Controller when enabled

swetharepakula:
This PR is dependent on #1381

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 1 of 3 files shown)


pkg/psc/controller.go
```diff
@@ -75,16 +79,96 @@ type Controller struct {
...
+			case <-stopChan:
+				return
+			default:
+				processKey()
+			}
+		}
+	}, time.Second, stopChan)
+
+	go func() {
+		time.Sleep(2 * time.Minute)
```
freehan, line 139: make 2 mins a constant?

pkg/psc/controller.go
```diff
@@ -75,16 +79,96 @@ type Controller struct {
...
+		return c.hasSynced(), nil
+	}, stopChan)
+
+	klog.V(2).Infof("Starting private service connect controller")
+	defer func() {
+		klog.V(2).Infof("Shutting down private service connect controller")
+		c.svcAttachmentQueue.ShutDown()
+	}()
+
+	go wait.Until(func() {
```
freehan, line 117: maybe worth adding this into a separate `sync` function?

and then simply do go wait.Util(c.sync(), time.Second, stopChan)?

pkg/psc/controller.go
```diff
@@ -75,16 +79,96 @@ type Controller struct {
...
 		saClient:            ctx.SAClient,
 		saNamer:             saNamer,
 		svcAttachmentLister: ctx.SAInformer.GetIndexer(),
+		svcAttachmentQueue:  workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
 		serviceLister:       ctx.ServiceInformer.GetIndexer(),
 		hasSynced:           ctx.HasSynced,
 		recorder:            ctx.Recorder,
 	}
+
+	ctx.SAInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
freehan, line 94: no need for DeleteFunc because of finalizer?

freehan (commented):
just nits. mostly LGTM

pkg/psc/controller.go
swetharepakula: Yes, that's correct. We don't do anything on delete, so not processing on those. Deletion will eventually be handled by GC

swetharepakula:
Since #1381 is merged and this pr has been rebased, removing hold.

/hold cancel

Issue #1386, opened 2021-03-21
Cloud CDN is missing multiple features in the BackendConfig


christian-roggia:
According to the official documentation the GCE ingress only supports the basic Cloud CDN features available in GCLB:
- https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#cloud_cdn

Cloud CDN support is currently missing the following features:
- [x] Cache keys https://cloud.google.com/cdn/docs/caching#cache-keys
- [ ] Cache mode https://cloud.google.com/cdn/docs/using-cache-modes#cache-mode
- [ ] TTL https://cloud.google.com/cdn/docs/using-ttl-overrides
- [ ] Signed URLs https://cloud.google.com/cdn/docs/using-signed-urls
- [ ] Signed Cookies https://cloud.google.com/cdn/docs/using-signed-cookies
- [ ] Negative caching (beta) https://cloud.google.com/cdn/docs/using-negative-caching
- [ ] Stale content (beta) https://cloud.google.com/cdn/docs/serving-stale-content
- [ ] Bypass cache header (pre-ga) https://cloud.google.com/cdn/docs/caching#bypassing-cache

The configuration now looks like the following:
```yaml
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: my-backendconfig
spec:
  cdn:
    enabled: true
    cachePolicy:
      includeHost: ...
      includeProtocol: ...
      includeQueryString: ...
      queryStringBlacklist: ...
      queryStringWhitelist: ...
```

An expected future configuration might look like this:
```yaml
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: my-backendconfig
spec:
  cdn:
    enabled: true
    mode: FORCE_CACHE_ALL
    ttl:
      client: 60s
      default: 60s
      max: 3600s
    negativeCaching:
      enabled: true
      policies:
        - code: 403
          ttl: 60s
    requestCoalescing: true
    serveWhileStale: 60s
    signedUrls:
      enabled: true
      maxAge: 60s
      keys:
        - signed-url-key-1
        - signed-url-key-2
    bypassCacheOnRequestHeaders:
      - Bypass-Cache-Header-1
      - Bypass-Cache-Header-2
    cachePolicy:
      includeHost: ...
      includeProtocol: ...
      includeQueryString: ...
      queryStringBlacklist: ...
      queryStringWhitelist: ...
```

[...]

PR #1387, opened 2021-03-23
Add missing CDN config options

caseywebdev:
Re: #1386

My attempt at adding a few of the missing simple scalar config options to the CDN backendconfig.


caseywebdev:
I signed it

christian-roggia:
Any chance to add all missing Cloud CDN configurations in a single PR?
There are also multiple tests that should be updated with the new configuration.

caseywebdev:
@christian-roggia I'd love to but having never used Go it's taken me quite a bit of time already just stumbling to this point. I tried adding new tests based off the existing ones but it's not going well.

spencerhance:
@caseywebdev Thanks for this PR.

fyi you will also have to run `hack/update-codegen.sh` to update the generated client libraries for backendconfig

caseywebdev:
I run `update-codegen.sh` which spit out a `k8s.io` file. I assume that gets merged into `pkg`? The diff when I did that seemed alarmingly big.

caseywebdev:
Ah I see why it's so big, there are a hundred instances of

```diff
-Copyright 2020 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
```

How do you want to handle this?

caseywebdev:
I pushed two commits, one with the 2 related file changes and another with all of the comment changes.

pkg/apis/backendconfig/v1/types.go
```diff
@@ -83,8 +83,11 @@ type OAuthClientCredentials struct {
 // CDNConfig contains configuration for CDN-enabled backends.
 // +k8s:openapi-gen=true
 type CDNConfig struct {
-	Enabled     bool            `json:"enabled"`
-	CachePolicy *CacheKeyPolicy `json:"cachePolicy,omitempty"`
+	Enabled           bool            `json:"enabled"`
+	CachePolicy       *CacheKeyPolicy `json:"cachePolicy,omitempty"`
+	CacheMode         *string         `json:"cacheMode,omitempty"`
+	RequestCoalescing *bool           `json:"requestCoalescing,omitempty"`
+	ServeWhileStale   *int64          `json:"serveWhileStale,omitempty"`
```
caseywebdev, line 90: Not sure what type handles `3m`, `2h`, `30s`, etc, wasn't able to find an example.

spencerhance: I think it makes sense to keep the int64 but rename it to `ServeWhileStaleSeconds`

spencerhance: That way it's descriptive but also matches the behavior from the underlying API

spencerhance:
Thanks for the updates, I think it's okay to drop the copyright updates as long as you keep all the generated files with code changes.

pkg/apis/backendconfig/v1/types.go
caseywebdev: Updated to `ServeWhileStaleSec` so it's consistent with the rest of the names.

spencerhance (approved):
Okay lgtm, have you been able to test these changes out by any chance?

bowei:
Can someone paste the e2e test run. Changes looks fine but needs to be tested

Dev25:
Is it possible to also add in the bool `negativeCaching` to this PR?

caseywebdev:
> Okay lgtm, have you been able to test these changes out by any chance?

I have not. I'm sure I could manage to get everything set up to run the e2e tests at some point but looking at the docs it does seem a bit involved. I'm wondering if someone who's already set up to do so could run them in this branch?

> Is it possible to also add in the bool negativeCaching to this PR?

That wouldn't be too much effort, though adding that flag without adding support for setting the policy config might not be a great solution.

bowei:
@spencerhance can you modify the e2e test and test the changes?

christian-roggia:
You could update the two files with:
```golang
// CDNConfig contains configuration for CDN-enabled backends.
// +k8s:openapi-gen=true
type CDNConfig struct {
	Enabled                     bool                          `json:"enabled"`
	BypassCacheOnRequestHeaders []*BypassCacheOnRequestHeader `json:"bypassCacheOnRequestHeaders,omitempty"`
	CachePolicy                 *CacheKeyPolicy               `json:"cachePolicy,omitempty"`
	CacheMode                   *string                       `json:"cacheMode,omitempty"`
	ClientTtl                   *int64                        `json:"clientTtl,omitempty"`
	DefaultTtl                  *int64                        `json:"defaultTtl,omitempty"`
	MaxTtl                      *int64                        `json:"maxTtl,omitempty"`
	NegativeCaching             *bool                         `json:"negativeCaching,omitempty"`
	NegativeCachingPolicy       []*NegativeCachingPolicy      `json:"negativeCachingPolicy,omitempty"`
	RequestCoalescing           *bool                         `json:"requestCoalescing,omitempty"`
	ServeWhileStaleSec          *int64                        `json:"serveWhileStaleSec,omitempty"`
	SignedUrlCacheMaxAgeSec     *int64                        `json:"signedUrlCacheMaxAgeSec,omitempty,string"`
	SignedUrlKeyNames           []string                      `json:"signedUrlKeyNames,omitempty"`
}

// CDNConfig contains configuration for how requests containing specific request
// headers bypass the cache, even if the content was previously cached.
// +k8s:openapi-gen=true
type BypassCacheOnRequestHeader struct {
	// The header field name to match on when bypassing cache. Values are
	// case-insensitive.
	HeaderName string `json:"headerName,omitempty"`
}

// CDNConfig contains configuration for how negative caching is applied.
// +k8s:openapi-gen=true
type NegativeCachingPolicy struct {
	// The HTTP status code to define a TTL against. Only HTTP status codes
	// 300, 301, 308, 404, 405, 410, 421, 451 and 501 are can be specified
	// as values, and you cannot specify a status code more than once.
	Code int64 `json:"code,omitempty"`
	// The TTL (in seconds) for which to cache responses with the
	// corresponding status code. The maximum allowed value is 1800s (30
	// minutes), noting that infrequently accessed objects may be evicted
	// from the cache before the defined TTL.
	Ttl int64 `json:"ttl,omitempty"`
}
```

and

```golang
// applyCDNSettings applies the CDN settings specified in the BackendConfig
// to the passed in compute.BackendService. A GCE API call still needs to be
// made to actually persist the changes.
func applyCDNSettings(sp utils.ServicePort, be *composite.BackendService) {
	beConfig := sp.BackendConfig
	// Apply the boolean switch
	be.EnableCDN = beConfig.Spec.Cdn.Enabled
	cacheKeyPolicy := beConfig.Spec.Cdn.CachePolicy
	// Apply the cache key policies if the BackendConfig contains them.
	if cacheKeyPolicy != nil {
		be.CdnPolicy = &composite.BackendServiceCdnPolicy{CacheKeyPolicy: &composite.CacheKeyPolicy{}}
		be.CdnPolicy.CacheKeyPolicy.IncludeHost = cacheKeyPolicy.IncludeHost
		be.CdnPolicy.CacheKeyPolicy.IncludeProtocol = cacheKeyPolicy.IncludeProtocol
		be.CdnPolicy.CacheKeyPolicy.IncludeQueryString = cacheKeyPolicy.IncludeQueryString
		be.CdnPolicy.CacheKeyPolicy.QueryStringBlacklist = cacheKeyPolicy.QueryStringBlacklist
		be.CdnPolicy.CacheKeyPolicy.QueryStringWhitelist = cacheKeyPolicy.QueryStringWhitelist
	}
	// Note that upon creation of a BackendServices, the fields 'IncludeHost',
	// 'IncludeProtocol' and 'IncludeQueryString' all default to true if not
	// explicitly specified.
	if beConfig.Spec.Cdn.CacheMode != nil {
		be.CdnPolicy.CacheMode = *beConfig.Spec.Cdn.CacheMode
	}
	if beConfig.Spec.Cdn.RequestCoalescing != nil {
		be.CdnPolicy.RequestCoalescing = *beConfig.Spec.Cdn.RequestCoalescing
	}
	if beConfig.Spec.Cdn.ServeWhileStaleSec != nil {
		be.CdnPolicy.ServeWhileStale = *beConfig.Spec.Cdn.ServeWhileStaleSec
	}
	if beConfig.Spec.Cdn.MaxTtl != nil {
		be.CdnPolicy.MaxTtl = *beConfig.Spec.Cdn.MaxTtl
	}
	if beConfig.Spec.Cdn.ClientTtl != nil {
		be.CdnPolicy.ClientTtl = *beConfig.Spec.Cdn.ClientTtl
	}
	if beConfig.Spec.Cdn.DefaultTtl != nil {
		be.CdnPolicy.DefaultTtl = *beConfig.Spec.Cdn.DefaultTtl
	}
	if beConfig.Spec.Cdn.NegativeCaching != nil {
		be.CdnPolicy.NegativeCaching = *beConfig.Spec.Cdn.NegativeCaching
	}
	if beConfig.Spec.Cdn.SignedUrlCacheMaxAgeSec != nil {
		be.CdnPolicy.SignedUrlCacheMaxAgeSec = *beConfig.Spec.Cdn.SignedUrlCacheMaxAgeSec
	}
	if beConfig.Spec.Cdn.SignedUrlKeyNames != nil {
		be.CdnPolicy.SignedUrlKeyNames = beConfig.Spec.Cdn.SignedUrlKeyNames
	}
	for _, policyRef := range beConfig.Spec.Cdn.NegativeCachingPolicy {
		be.CdnPolicy.NegativeCachingPolicy = append(
			be.CdnPolicy.NegativeCachingPolicy,
			&composite.BackendServiceCdnPolicyNegativeCachingPolicy{
				Code: policyRef.Code,
				Ttl:  policyRef.Ttl,
			},
		)
	}
	for _, policyRef := range beConfig.Spec.Cdn.BypassCacheOnRequestHeaders {
		be.CdnPolicy.BypassCacheOnRequestHeaders = append(
			be.CdnPolicy.BypassCacheOnRequestHeaders,
			&composite.BackendServiceCdnPolicyBypassCacheOnRequestHeader{
				HeaderName: policyRef.HeaderName,
			},
		)
	}
}
```

This will support all the missing features, including negative caching.

caseywebdev:
Thanks @christian-roggia! I've updated the PR and codegen

spencerhance:
@caseywebdev, I just checkout out your code and the TestCDN e2e test is still passing 👍 

Can you just add your additional fields to the `"http one path w/ CDN & cache policies."` test in [TestCDN.go](https://github.com/kubernetes/ingress-gce/blob/master/cmd/e2e-test/cdn_test.go#L51)?

I think you have to update the helper function in the same file as well. 

I'll run it again after to make sure it's passing.

Thanks!

caseywebdev:
I haven't forgotten about this, just haven't had time to finish it up. Help is welcomed! Not being able to control these settings is an ongoing issue for us in GKE.

gabihodoroaga:
Hi,

I need to be able to control the Cloud CDN options int GKE too. This is a blocking issue. 

@caseywebdev I took your branch and I deployed on my cluster and I found out that even if the changes are applied, the current implementation does not correctly detect the changes in backend settings and it triggers an update every time (e.g. CacheMode defaults to CACHE_ALL_STATIC when not specified).
While debugging this I end up with a solution and I published here https://github.com/gabihodoroaga/ingress-gce/blob/backendconfig-cdn-config/pkg/backends/features/cdn.go, I also updated the unit tests but definitively a lot more unit testing is required.
Please take a look and let me know. You mention that you could use some help. I can help.

Thanks!

caseywebdev:
Thanks @gabihodoroaga! I’d happily merge any PRs you want to make against [my branch](https://github.com/caseywebdev/ingress-gce/tree/backendconfig-cdn-config ) to get this PR across the finish line.

caseywebdev:
@gabihodoroaga I think you'll need to sign the CLA since your commits are in here now.

gabihodoroaga:
@caseywebdev I signed the CLA.

gabihodoroaga:
@caseywebdev I misspelled my email address and the commit is not associated with my github account.
The solution is to reset your branch before the merge and to start over again.

caseywebdev:
Hey @spencerhance, after @gabihodoroaga's work on this I think it's ready for another check

nathanielop:
What's the status of this @spencerhance? Looking to use these features in my company's config. Been watching this PR for months now.

gabihodoroaga:
@nathanielop You can deploy your version of ingress-gce on you cluster if you follow the instruction from here ```docs/deploy/gke```. In this [branch](https://github.com/gabihodoroaga/ingress-gce/tree/backendconfig-cdn-config), which is the source of the pull request, I updated also the ```gke-self-managed.sh``` script and the ```glbc.yaml``` and ```rbac.yaml``` file.

nathanielop:
@bowei Are there any other changes to be made before a secondary review is possible?

bowei:
Can you squash your commits into the following:

*  docs/deploy/** changes
* code changes + test

gabihodoroaga:
I will squash the commits and fix the conflicts.

bowei:
Yes, thanks for your patience -- I'm testing it out, hopefully will be able to merge tonight.

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,303 @@ package features
...
 func EnsureCDN(sp utils.ServicePort, be *composite.BackendService) bool {
 	if sp.BackendConfig.Spec.Cdn == nil {
 		return false
 	}
-	beTemp := &composite.BackendService{}
-	applyCDNSettings(sp, beTemp)
-	// Only compare CdnPolicy if it was specified.
-	if (beTemp.CdnPolicy != nil && !reflect.DeepEqual(beTemp.CdnPolicy, be.CdnPolicy)) || beTemp.EnableCDN != be.EnableCDN {
-		applyCDNSettings(sp, be)
+	klog.V(3).Infof("BackendConfig.Spec.Cdn before merge = %s", prettyPrint(sp.BackendConfig.Spec.Cdn))
```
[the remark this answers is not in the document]
gabihodoroaga: Ok. I will make the code changes.

pkg/backends/features/cdn.go
```diff
@@ -17,49 +17,309 @@ limitations under the License.
...
 	"reflect"
+	"strings"
 
+	"github.com/kr/pretty"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog"
 )
 
+// This values are copied from GCP Load balancer
```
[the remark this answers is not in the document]
gabihodoroaga: This was already covered by the e2e tests.

caseywebdev:
Thanks @gabihodoroaga 🙏🏻

mr1name:
@bowei please... This has been going on for so long... What else needs to be done to merge this branch?

pkg/composite/gen.go
```diff
@@ -68,10 +68,8 @@ type Address struct {
 	// after every request to modify or update labels. You must always
```
gabihodoroaga, line 68: The file pkg/composite/gen.go was generated from an updated version of "vendor/google.golang.org/api/compute/v0.alpha/compute-api.json", but this file was not included in the commit as this would require an update to the vendor folder, and there are too many changes.

caseywebdev:
@bowei it looks like @gabihodoroaga has addressed your comments, can we get this merged?

bowei:
Thanks -- taking a look

cmd/e2e-test/cdn_test.go
```diff
@@ -17,146 +17,938 @@ limitations under the License.
...
+		expCdnPolicy := deepCopyCdnPolicy(expected)
+		bsCdnPolicy := deepCopyCdnPolicy(bs.GA.CdnPolicy)
+
+		if sliceEqual(expCdnPolicy.NegativeCachingPolicy, bsCdnPolicy.NegativeCachingPolicy) {
+			expCdnPolicy.NegativeCachingPolicy = nil
+			bsCdnPolicy.NegativeCachingPolicy = nil
+		}
+
+		if !reflect.DeepEqual(expCdnPolicy, bsCdnPolicy) {
+			return fmt.Errorf("expected %s, but got %s",
```
bowei, line 906: use pretty.Sprint

cmd/e2e-test/cdn_test.go
```diff
@@ -17,146 +17,938 @@ limitations under the License.
...
+
+func createInt64(a int64) *int64 {
+	return &a
+}
+
+func createBool(a bool) *bool {
+	return &a
+}
+
+func prettyJSON(data interface{}) string {
```
bowei, line 745: remove this and use pretty.Sprint

cmd/e2e-test/cdn_test.go
```diff
@@ -17,146 +17,938 @@ limitations under the License.
...
+		panic(err)
+	}
+	return buffer.String()
+}
+
+type backendServiceBuilder struct {
+	CdnPolicy *compute.BackendServiceCdnPolicy
+}
+
+func newBSWithDefaults() *backendServiceBuilder {
```
bowei, line 761: I think you can just call this "newBSBuilder"

cmd/e2e-test/cdn_test.go
```diff
@@ -17,146 +17,938 @@ limitations under the License.
...
-			}
+	return ing, nil
+}
 
-			// Wait for GCLB resources to be deleted.
-			if err := crud.Delete(s.Namespace, ing.Name); err != nil {
-				t.Errorf("Delete(%q) = %v, want nil", ing.Name, err)
-			}
+func updateConfigAndValidate(ing *networkingv1.Ingress, namespace, serviceName string,
+	beConfig *backendconfig.BackendConfig, expected *compute.BackendServiceCdnPolicy) error {
```
bowei, line 833: [minor]

If spread across multiple lines, 

prefer 

```
func foo(
  ...
  ) error 
{
  code
```

so the func signature is clear from indentation

cmd/e2e-test/cdn_test.go
```diff
@@ -17,146 +17,938 @@ limitations under the License.
...
-		numBsWithPolicy = numBsWithPolicy + 1
 	}
+
 	if numBsWithPolicy != 1 {
 		return fmt.Errorf("unexpected number of backend service has cache policy attached: got %d, want 1", numBsWithPolicy)
 	}
 	return nil
 }
+
+func sliceEqual(x, y []*compute.BackendServiceCdnPolicyNegativeCachingPolicy) bool {
```
bowei, line 918: Is it valid to have multiple of the same Code with TTL?

is this simpler to do:
```
if len(x) != len(y) { return false }
xMap := map[string]int64{}
yMap := map[string]int64{}
for _, v := range x { xMap[v.Code] = v.Ttl }
for _, v := range y { yMap[v.Code] = v.Ttl }
return reflect.DeepEqual(xMap, yMap)
```

docs/deploy/resources/glbc.yaml
```diff
@@ -63,8 +63,7 @@ spec:
         - --gce-ratelimit=ga.BackendServices.Get,qps,1.8,1
         - --gce-ratelimit=ga.HealthChecks.Get,qps,1.8,1
         - --gce-ratelimit=alpha.HealthChecks.Get,qps,1.8,1
-        - --enable-csm=[ENABLE_CSM]
-        - --csm-service-skip-namespaces=kube-system,istio-system
+        - --enable-backendconfig-healthcheck
```
bowei, line 66: can you add the other enable flags?

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,219 @@ package features
...
+	sort.Sort(ySorted)
+
+	return reflect.DeepEqual(xSorted, ySorted)
+}
+
+func renderConfig(sp utils.ServicePort) *composite.BackendService {
+	cdnConfig := sp.BackendConfig.Spec.Cdn
+	be := &composite.BackendService{}
+	be.EnableCDN = cdnConfig.Enabled
+
```
bowei, line 86: Should we render the rest of the config if  cdnConfig.Enabled = false?

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,219 @@ package features
...
+	be.CdnPolicy = &composite.BackendServiceCdnPolicy{}
+
+	if cdnConfig.CachePolicy != nil {
+		beCacheKeyPolicy := &composite.CacheKeyPolicy{}
+		beCacheKeyPolicy.IncludeHost = cdnConfig.CachePolicy.IncludeHost
+		beCacheKeyPolicy.IncludeProtocol = cdnConfig.CachePolicy.IncludeProtocol
+		beCacheKeyPolicy.IncludeQueryString = cdnConfig.CachePolicy.IncludeQueryString
+		if cdnConfig.CachePolicy.IncludeQueryString {
+			if cdnConfig.CachePolicy.QueryStringBlacklist != nil && len(cdnConfig.CachePolicy.QueryStringBlacklist) > 0 {
```
bowei, line 95: nil list is the same as an empty list -- you should be able to set directly without checking.

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,219 @@ package features
...
+				beCacheKeyPolicy.QueryStringBlacklist = cdnConfig.CachePolicy.QueryStringBlacklist
+			}
+			if cdnConfig.CachePolicy.QueryStringWhitelist != nil && len(cdnConfig.CachePolicy.QueryStringWhitelist) > 0 {
+				beCacheKeyPolicy.QueryStringWhitelist = cdnConfig.CachePolicy.QueryStringWhitelist
```
bowei, line 99: see comment about nil list

pkg/backends/features/cdn.go
bowei: Do we need to handle ForceSendFields for an empty list?

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,219 @@ package features
...
+
+	if current.CdnPolicy == nil {
+		return true
+	}
+
+	// In order to use reflect.DeepEqual we need to handle the special cases
+
+	// SignedUrlKeyNames are not handled by this function and must be preserved as they are
+	if current.CdnPolicy != nil && current.CdnPolicy.SignedUrlKeyNames != nil {
+		new.CdnPolicy.SignedUrlKeyNames = current.CdnPolicy.SignedUrlKeyNames
```
bowei, line 220: Don't we need to compare these inside hasDiff?

One way to handle this is to take these fields out, compare them separately, and nil them for the DeepEqual comparison

```
if hasDiffSignedUrlKeyNames(new.CdnPolicy.SignedUrlKeyNames, current.CdnPolicy.SignedUrlKeyNames) { return true }
new.CdnPolicy.SignedUrlKeyNames = nil
current.CdnPolicy.SignedUrlKeyNames = nil
```

pkg/backends/features/cdn.go
```diff
@@ -18,48 +18,219 @@ package features
...
+	}
+
+	// for NegativeCachingPolicy order does not matter, so if there are equal regardless of order
+	// we take the current one
+	if negativeCachingPolicyEqual(new.CdnPolicy.NegativeCachingPolicy, current.CdnPolicy.NegativeCachingPolicy) {
+		new.CdnPolicy.NegativeCachingPolicy = current.CdnPolicy.NegativeCachingPolicy
```
bowei, line 226: Same here (see above)

pkg/backends/syncer.go
```diff
@@ -117,6 +117,44 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
...
 	}
 
+	// Sync SignedUrlKeys. Why here? This is the only place where I have access to the key values (sp)
+	existingKeyNames := map[string]bool{}
+	if be.CdnPolicy != nil && be.CdnPolicy.SignedUrlKeyNames != nil {
+		for _, key := range be.CdnPolicy.SignedUrlKeyNames {
+			existingKeyNames[key] = false
+		}
+	}
+	// there is a hard limit of maximum 3 keys per bakend, if you add before remove you will hit the limit
```
bowei, line 127: bakend => backend

pkg/backends/syncer.go
```diff
@@ -117,6 +117,44 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
...
+	// delete all removed keys
+	for keyName, found := range existingKeyNames {
+		if !found {
+			klog.V(5).Infof("Removing SignedUrlKey %q from backend %q", keyName, be.Name)
+			if err = s.backendPool.DeleteSignedUrlKey(be, keyName); err != nil {
+				return err
+			}
+		}
+	}
+	// add all apended keys
```
bowei, line 150: apended => appended

pkg/backends/syncer.go
```diff
@@ -117,6 +117,44 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
 		}
 	}
 
+	// Sync SignedUrlKeys. Why here? This is the only place where I have access to the key values (sp)
```
bowei, line 120: Can we factor this out into its own funcs, also can we put inside EnsureCDN, rather than at the toplevel here? What other information is needed to be passed to EnsureCDN

pkg/composite/gen/main.go
```diff
@@ -818,8 +818,11 @@ func ({{$type.VarName}} *{{$type.Name}}) To{{$version}}() (*compute{{$extension}
 
 	{{- if eq $type.Name "BackendService"}}
 	// Set force send fields. This is a temporary hack.
-	if {{$lower}}.CdnPolicy != nil && {{$lower}}.CdnPolicy.CacheKeyPolicy != nil {
+	if {{$lower}}.CdnPolicy != nil { 
+		if {{$lower}}.CdnPolicy.CacheKeyPolicy != nil {
 		{{$lower}}.CdnPolicy.CacheKeyPolicy.ForceSendFields = []string{"IncludeHost", "IncludeProtocol", "IncludeQueryString", "QueryStringBlacklist", "QueryStringWhitelist"}
+		}
+		{{$lower}}.CdnPolicy.ForceSendFields = append({{$type.VarName}}.CdnPolicy.ForceSendFields, []string{"NegativeCaching", "RequestCoalescing","SignedUrlCacheMaxAgeSec","ServeWhileStale"}...)
```
bowei, line 825: Do we need to check that these fields are actually needed to be sent?

pkg/composite/gen/main.go
```diff
@@ -818,8 +818,11 @@ func ({{$type.VarName}} *{{$type.Name}}) To{{$version}}() (*compute{{$extension}
 
 	{{- if eq $type.Name "BackendService"}}
 	// Set force send fields. This is a temporary hack.
-	if {{$lower}}.CdnPolicy != nil && {{$lower}}.CdnPolicy.CacheKeyPolicy != nil {
+	if {{$lower}}.CdnPolicy != nil { 
+		if {{$lower}}.CdnPolicy.CacheKeyPolicy != nil {
 		{{$lower}}.CdnPolicy.CacheKeyPolicy.ForceSendFields = []string{"IncludeHost", "IncludeProtocol", "IncludeQueryString", "QueryStringBlacklist", "QueryStringWhitelist"}
```
bowei, line 823: can you fix the indent here in the codegen?

pkg/e2e/adapter/beconfig.go
```diff
@@ -81,6 +82,24 @@ func (crud *BackendConfigCRUD) Update(bc *v1.BackendConfig) (*v1.BackendConfig,
 	return toV1(legacyBC), err
 }
 
+// Ensure creates or updates the backend config
+func (crud *BackendConfigCRUD) Ensure(bc *v1.BackendConfig) (*v1.BackendConfig, error) {
```
bowei, line 86: If this is only used in tests, can you move this func to test only code

cmd/e2e-test/cdn_test.go
gabihodoroaga: Done.

cmd/e2e-test/cdn_test.go
gabihodoroaga: Done.

cmd/e2e-test/cdn_test.go
gabihodoroaga: Done.

cmd/e2e-test/cdn_test.go
gabihodoroaga: Done, almost. Except that the curly bracket on a new line is not valid in go.

cmd/e2e-test/cdn_test.go
gabihodoroaga: It is simpler. Done.

docs/deploy/resources/glbc.yaml
gabihodoroaga: Added:
"enable-endpoint-slices", "Enable using Endpoint Slices API instead of Endpoints API")
"enable-frontend-config","Optional, whether or not to enable FrontendConfig."

For all other enable-* flags:

"enable-asm-config-map-config", "Enable ASMConfigMapBasedConfig"
"enable-non-gcp-mode", "Set to true when running on a non-GCP cluster."
"enable-delete-unused-frontends", "Enable deleting unused gce frontend resources."
"enable-v2-frontend-namer", "Enable v2 ingress frontend naming policy."
"enable-psc", "Enable PSC controller"
"enable-ingress-ga-fields", "Enable using Ingress Class GA features"
"enable-traffic-scaling", "Enable support for Service {max-rate-per-endpoint, capacity-scaler}"

it looks like they have special purpose.

pkg/backends/features/cdn.go
gabihodoroaga: No. Done.

pkg/backends/features/cdn.go
gabihodoroaga: Yes. We need these checks here to make sure we set nil for an empty list `:[]`, otherwise DeepEqual will fail and it will be considered a change.
These fields "QueryStringBlacklist", "QueryStringWhitelist" are added to the list of ForceSendFields.

pkg/backends/features/cdn.go
gabihodoroaga: Same as above.

pkg/backends/features/cdn.go
gabihodoroaga: This is possible too, but I think is would be better not to alter the "current". Cloning the current it will add unnecessary overhead.

pkg/backends/syncer.go
gabihodoroaga: Done.

pkg/backends/syncer.go
gabihodoroaga: Refactored on its own func. We cannot put this inside EnsureCDN because composite.BackendService.CdnPolicy.SignedUrlKeyNames contains only the keys names ([]string) and we cannot pass the information about the keys values, and what keys needs to be added or removed. Updating the backend inside EnsureCDN does not match the current pattern.

pkg/composite/gen/main.go
gabihodoroaga: Yes. Otherwise empty values are not cleared from the backend.

pkg/e2e/adapter/beconfig.go
gabihodoroaga: It is part of the e2e package.

gabihodoroaga:
@bowei Thank you for the review. I know that by approving this PR you are assuming a big responsibility and I appreciate all your effort.

pkg/backends/features/cdn.go
bowei: Can you add a comment with that explanation -- it is a bit subtle for maintainers that come later.

pkg/backends/features/cdn.go
bowei: Same here, add a comment to explain why the code needs to exist.

pkg/backends/features/cdn.go
bowei: If that is the case, can we have hasDiff() do all of the comparisons.

As currently named, it can be confusing, as it doesn't do comparisons for everything...

pkg/backends/features/cdn.go
gabihodoroaga: Done.

pkg/backends/features/cdn.go
gabihodoroaga: Done.

pkg/backends/features/cdn.go
gabihodoroaga: All the comparisons are done in hasDiff. This part is only sets the value of NegativeCachingPolicy from the current when the values are equal regardless of the order so the reflect.DeepEqual will not fail because of this.

nathanielop:
@bowei Hate to pester after all the work that you've put into merging this PR, but it seems @gabihodoroaga has once again addressed all of the concerns, and the PR is ready for review/merging presuming there is no more outstanding issues to be resolved?

caseywebdev:
@bowei ping, we gotta be close!

caseywebdev:
Thanks! When could we expect to have these new options available in GKE?

bowei:
It may take a while as there are production freezes for Black Friday and the holidays. We will update when we know more in terms of timeline.

nicknotfun:
Just checking in on this one?

turtlesoupy:
Any updates now that the holidays are over? This affects our migration from container-native load balancing to managing it all the k8s.

bowei:
We will update soon -- happy New Year! Note that availability will be first through the latest rapid release. It may take some time to get to regular and stable channels.

nicknotfun:
> We will update soon -- happy New Year! Note that availability will be first through the latest rapid release. It may take some time to get to regular and stable channels.

Thanks for the update! We likely can adopt a more rapid channel when it's available; where's best to be notified as it progresses ?

turtlesoupy:
+1, how should we monitor status here?

swetharepakula:
A new version of Ingress has been cut and is currently in the pipeline to be released to GKE 1.23 clusters. 
It is currently going through testing and validation. The new version will still take about 3-4 weeks before it is available in the RAPID channel. I will update once I have more information.

bakkerpeter:
@swetharepakula any news on this? thanks!

pstreise:
I can verfiy: It's online on GKE on the rapid channel. I've just tested it.

swetharepakula:
Rollout is still underway. It will be available in all zones in the Rapid Channel in the next couple days.

This is available starting in GKE version `1.23.3-gke.900`

nicknotfun:
Moved us to `1.23.3-gke.1100` but surprisingly seems unavailable? Was it disabled?

swetharepakula:
`1.23.3-gke.1100` should have the new CDN options. Can you post any errors that you are seeing?

nicknotfun:
Sure,

Cluster location / version: `us-central1-c` / `1.23.3-gke.1100`
Node version(s): all `1.23.3-gke.1100`

Config:
```
kind: BackendConfig
metadata:
  namespace: default
  name: foo
spec:
  cdn:
    enabled: true
    maxTtl: 31622400 # 1 year
  ``` 
  
  `kubectl apply -f test.yaml` gives me a validation error;
  ```
  error: error validating "test.yaml": error validating data: ValidationError(BackendConfig.spec.cdn): unknown field "maxTTL" in com.google.cloud.v1.BackendConfig.spec.cdn; if you choose to ignore these errors, turn validation off with --validate=false
  ```

**Comment by nicknotfun** [comment]
Just checking if you have any advice @swetharepakula ?

**Comment by swetharepakula** [comment]
@nicknotfun , can you paste the output of `kubectl get crd backendconfigs.cloud.google.com`? Do you happen to have HTTPLoadBalancing disabled on the cluster?

**Comment by nicknotfun** [comment]
We do have HTTPLoadBalancing enabled, as for the command:

```
> kubectl get crd backendconfigs.cloud.google.com
NAME                              CREATED AT
backendconfigs.cloud.google.com   2021-12-14T21:48:34Z
```

Which i'm guessing is the root-cause here. It is worth nothing that we initially created this cluster with the regular release channel; upgrading to rapid (and to the version mentioned above) in order to take advantage of the CDN settings - being in a pre-alpha state we don't care for the SLA stuff just yet. 

I'm guessing the control pane upgrade didn't upgrade the CRDs?

**Comment by alisspers** [comment]
I think I might be running into the same issue with my cluster running 1.23.4-gke.300 in europe-north1-c.
That version should include the options, right?

Getting the following errors when trying to apply the below BackendConfig:
```
error: error validating "frontend-service.yml": error validating data: [
	ValidationError(BackendConfig.spec.cdn): unknown field "clientTtl" in com.google.cloud.v1.BackendConfig.spec.cdn,
	ValidationError(BackendConfig.spec.cdn): unknown field "defaultTtl" in com.google.cloud.v1.BackendConfig.spec.cdn,
	ValidationError(BackendConfig.spec.cdn): unknown field "maxTtl" in com.google.cloud.v1.BackendConfig.spec.cdn
]; if you choose to ignore these errors, turn validation off with --validate=false
```

My backendconfig and service definitions:

```yaml
---
kind: BackendConfig
metadata:
  name: frontend-backendconfig
spec:
  cdn:
    enabled: true
    maxTtl: 31622400 # 1 year
    defaultTtl: 31622400 # 1 year
    clientTtl: 31622400 # 1 year
    cachePolicy:
      includeHost: true
      includeProtocol: true
      includeQueryString: true
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    cloud.google.com/backend-config: '{"default": "frontend-backendconfig"}'
  name: frontend
  labels:
    app: frontend
spec:
  selector:
    app: frontend
  ports:
    - protocol: TCP
      name: frontend-http
      port: 80
      targetPort: 81
  type: NodePort
```

Crd yaml:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: "2019-04-13T15:53:14Z"
  generation: 7
  name: backendconfigs.cloud.google.com
  resourceVersion: "401185756"
  uid: 3f503d61-5e04-11e9-a8df-42010aa60014
spec:
  conversion:
    strategy: None
  group: cloud.google.com
  names:
    kind: BackendConfig
    listKind: BackendConfigList
    plural: backendconfigs
    shortNames:
    - bc
    singular: backendconfig
  scope: Namespaced
  versions:
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            description: 'APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
            type: string
          kind:
            description: 'Kind is a string value representing the REST resource this
              object represents. Servers may infer this from the endpoint the client
              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
            type: string
          metadata:
            type: object
          spec:
            description: BackendConfigSpec is the spec for a BackendConfig resource
            properties:
              cdn:
                description: CDNConfig contains configuration for CDN-enabled backends.
                properties:
                  cachePolicy:
                    description: CacheKeyPolicy contains configuration for how requests
                      to a CDN-enabled backend are cached.
                    properties:
                      includeHost:
                        description: If true, requests to different hosts will be
                          cached separately.
                        type: boolean
                      includeProtocol:
                        description: If true, http and https requests will be cached
                          separately.
                        type: boolean
                      includeQueryString:
                        description: If true, query string parameters are included
                          in the cache key according to QueryStringBlacklist and QueryStringWhitelist.
                          If neither is set, the entire query string is included and
                          if false the entire query string is excluded.
                        type: boolean
                      queryStringBlacklist:
                        description: Names of query strint parameters to exclude from
                          cache keys. All other parameters are included. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                      queryStringWhitelist:
                        description: Names of query string parameters to include in
                          cache keys. All other parameters are excluded. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                    type: object
                  enabled:
                    type: boolean
                required:
                - enabled
                type: object
              connectionDraining:
                description: ConnectionDrainingConfig contains configuration for connection
                  draining. For now the draining timeout. May manage more settings
                  in the future.
                properties:
                  drainingTimeoutSec:
                    description: Draining timeout in seconds.
                    format: int64
                    type: integer
                type: object
              customRequestHeaders:
                description: CustomRequestHeadersConfig contains configuration for
                  custom request headers
                properties:
                  headers:
                    items:
                      type: string
                    type: array
                type: object
              healthCheck:
                description: HealthCheckConfig contains configuration for the health
                  check.
                properties:
                  checkIntervalSec:
                    description: CheckIntervalSec is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  healthyThreshold:
                    description: HealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  port:
                    description: Port is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                      If Port is used, the controller updates portSpecification as
                      well
                    format: int64
                    type: integer
                  requestPath:
                    description: RequestPath is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  timeoutSec:
                    description: TimeoutSec is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  type:
                    description: Type is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  unhealthyThreshold:
                    description: UnhealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                type: object
              iap:
                description: IAPConfig contains configuration for IAP-enabled backends.
                properties:
                  enabled:
                    type: boolean
                  oauthclientCredentials:
                    description: OAuthClientCredentials contains credentials for a
                      single IAP-enabled backend.
                    properties:
                      clientID:
                        description: Direct reference to OAuth client id.
                        type: string
                      clientSecret:
                        description: Direct reference to OAuth client secret.
                        type: string
                      secretName:
                        description: The name of a k8s secret which stores the OAuth
                          client id & secret.
                        type: string
                    required:
                    - secretName
                    type: object
                required:
                - enabled
                - oauthclientCredentials
                type: object
              logging:
                description: LogConfig contains configuration for logging.
                properties:
                  enable:
                    description: This field denotes whether to enable logging for
                      the load balancer traffic served by this backend service.
                    type: boolean
                  sampleRate:
                    description: This field can only be specified if logging is enabled
                      for this backend service. The value of the field must be in
                      [0, 1]. This configures the sampling rate of requests to the
                      load balancer where 1.0 means all logged requests are reported
                      and 0.0 means no logged requests are reported. The default value
                      is 1.0.
                    format: double
                    type: number
                type: object
              securityPolicy:
                description: SecurityPolicyConfig contains configuration for CloudArmor-enabled
                  backends. If not specified, the controller will not reconcile the
                  security policy configuration. In other words, users can make changes
                  in GCE without the controller overwriting them.
                properties:
                  name:
                    description: Name of the security policy that should be associated.
                      If set to empty, the existing security policy on the backend
                      will be removed.
                    type: string
                required:
                - name
                type: object
              sessionAffinity:
                description: SessionAffinityConfig contains configuration for stickyness
                  parameters.
                properties:
                  affinityCookieTtlSec:
                    format: int64
                    type: integer
                  affinityType:
                    type: string
                type: object
              timeoutSec:
                format: int64
                type: integer
            type: object
          status:
            type: object
        type: object
    served: true
    storage: true
  - name: v1beta1
    schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            description: 'APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
            type: string
          kind:
            description: 'Kind is a string value representing the REST resource this
              object represents. Servers may infer this from the endpoint the client
              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
            type: string
          metadata:
            type: object
          spec:
            description: BackendConfigSpec is the spec for a BackendConfig resource
            properties:
              cdn:
                description: CDNConfig contains configuration for CDN-enabled backends.
                properties:
                  cachePolicy:
                    description: CacheKeyPolicy contains configuration for how requests
                      to a CDN-enabled backend are cached.
                    properties:
                      includeHost:
                        description: If true, requests to different hosts will be
                          cached separately.
                        type: boolean
                      includeProtocol:
                        description: If true, http and https requests will be cached
                          separately.
                        type: boolean
                      includeQueryString:
                        description: If true, query string parameters are included
                          in the cache key according to QueryStringBlacklist and QueryStringWhitelist.
                          If neither is set, the entire query string is included and
                          if false the entire query string is excluded.
                        type: boolean
                      queryStringBlacklist:
                        description: Names of query strint parameters to exclude from
                          cache keys. All other parameters are included. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                      queryStringWhitelist:
                        description: Names of query string parameters to include in
                          cache keys. All other parameters are excluded. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                    type: object
                  enabled:
                    type: boolean
                required:
                - enabled
                type: object
              connectionDraining:
                description: ConnectionDrainingConfig contains configuration for connection
                  draining. For now the draining timeout. May manage more settings
                  in the future.
                properties:
                  drainingTimeoutSec:
                    description: Draining timeout in seconds.
                    format: int64
                    type: integer
                type: object
              customRequestHeaders:
                description: CustomRequestHeadersConfig contains configuration for
                  custom request headers
                properties:
                  headers:
                    items:
                      type: string
                    type: array
                type: object
              healthCheck:
                description: HealthCheckConfig contains configuration for the health
                  check.
                properties:
                  checkIntervalSec:
                    description: CheckIntervalSec is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  healthyThreshold:
                    description: HealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  port:
                    format: int64
                    type: integer
                  requestPath:
                    description: RequestPath is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  timeoutSec:
                    description: TimeoutSec is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  type:
                    description: Type is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  unhealthyThreshold:
                    description: UnhealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                type: object
              iap:
                description: IAPConfig contains configuration for IAP-enabled backends.
                properties:
                  enabled:
                    type: boolean
                  oauthclientCredentials:
                    description: OAuthClientCredentials contains credentials for a
                      single IAP-enabled backend.
                    properties:
                      clientID:
                        description: Direct reference to OAuth client id.
                        type: string
                      clientSecret:
                        description: Direct reference to OAuth client secret.
                        type: string
                      secretName:
                        description: The name of a k8s secret which stores the OAuth
                          client id & secret.
                        type: string
                    required:
                    - secretName
                    type: object
                required:
                - enabled
                - oauthclientCredentials
                type: object
              securityPolicy:
                description: SecurityPolicyConfig contains configuration for CloudArmor-enabled
                  backends. If not specified, the controller will not reconcile the
                  security policy configuration. In other words, users can make changes
                  in GCE without the controller overwriting them.
                properties:
                  name:
                    description: Name of the security policy that should be associated.
                      If set to empty, the existing security policy on the backend
                      will be removed.
                    type: string
                required:
                - name
                type: object
              sessionAffinity:
                description: SessionAffinityConfig contains configuration for stickyness
                  parameters.
                properties:
                  affinityCookieTtlSec:
                    format: int64
                    type: integer
                  affinityType:
                    type: string
                type: object
              timeoutSec:
                format: int64
                type: integer
            type: object
          status:
            type: object
        type: object
    served: true
    storage: false
status:
  acceptedNames:
    kind: BackendConfig
    listKind: BackendConfigList
    plural: backendconfigs
    shortNames:
    - bc
    singular: backendconfig
  conditions:
  - lastTransitionTime: "2019-04-13T15:53:14Z"
    message: no conflicts found
    reason: NoConflicts
    status: "True"
    type: NamesAccepted
  - lastTransitionTime: "2020-11-19T13:55:10Z"
    message: the initial names have been accepted
    reason: InitialNamesAccepted
    status: "True"
    type: Established
  storedVersions:
  - v1beta1
  - v1

```

nicknotfun:
@swetharepakula  just checking in? The full output for me is quite similar to @alisspers - I lack the "v1beta1" stored version line.

swetharepakula:
In a 1.23.4-gke.900 cluster in us-central1-a I see the all the proper fields: 
```
kubectl get crd backendconfigs.cloud.google.com -o yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: "2022-03-10T23:03:02Z"
  generation: 1
  managedFields:
  - apiVersion: apiextensions.k8s.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        f:acceptedNames:
          f:kind: {}
          f:listKind: {}
          f:plural: {}
          f:shortNames: {}
          f:singular: {}
        f:conditions:
          k:{"type":"Established"}:
            .: {}
            f:lastTransitionTime: {}
            f:message: {}
            f:reason: {}
            f:status: {}
            f:type: {}
          k:{"type":"NamesAccepted"}:
            .: {}
            f:lastTransitionTime: {}
            f:message: {}
            f:reason: {}
            f:status: {}
            f:type: {}
    manager: Go-http-client
    operation: Update
    subresource: status
    time: "2022-03-10T23:03:02Z"
  - apiVersion: apiextensions.k8s.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:conversion:
          .: {}
          f:strategy: {}
        f:group: {}
        f:names:
          f:kind: {}
          f:listKind: {}
          f:plural: {}
          f:shortNames: {}
          f:singular: {}
        f:scope: {}
        f:versions: {}
    manager: glbc
    operation: Update
    time: "2022-03-10T23:03:02Z"
  name: backendconfigs.cloud.google.com
  resourceVersion: "616"
  uid: c4522422-ba5f-494d-a375-9c476235d3ad
spec:
  conversion:
    strategy: None
  group: cloud.google.com
  names:
    kind: BackendConfig
    listKind: BackendConfigList
    plural: backendconfigs
    shortNames:
    - bc
    singular: backendconfig
  scope: Namespaced
  versions:
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            description: 'APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
            type: string
          kind:
            description: 'Kind is a string value representing the REST resource this
              object represents. Servers may infer this from the endpoint the client
              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
            type: string
          metadata:
            type: object
          spec:
            description: BackendConfigSpec is the spec for a BackendConfig resource
            properties:
              cdn:
                description: CDNConfig contains configuration for CDN-enabled backends.
                properties:
                  bypassCacheOnRequestHeaders:
                    items:
                      description: BypassCacheOnRequestHeader contains configuration
                        for how requests containing specific request headers bypass
                        the cache, even if the content was previously cached.
                      properties:
                        headerName:
                          description: The header field name to match on when bypassing
                            cache. Values are case-insensitive.
                          type: string
                      type: object
                    type: array
                  cacheMode:
                    type: string
                  cachePolicy:
                    description: CacheKeyPolicy contains configuration for how requests
                      to a CDN-enabled backend are cached.
                    properties:
                      includeHost:
                        description: If true, requests to different hosts will be
                          cached separately.
                        type: boolean
                      includeProtocol:
                        description: If true, http and https requests will be cached
                          separately.
                        type: boolean
                      includeQueryString:
                        description: If true, query string parameters are included
                          in the cache key according to QueryStringBlacklist and QueryStringWhitelist.
                          If neither is set, the entire query string is included and
                          if false the entire query string is excluded.
                        type: boolean
                      queryStringBlacklist:
                        description: Names of query strint parameters to exclude from
                          cache keys. All other parameters are included. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                      queryStringWhitelist:
                        description: Names of query string parameters to include in
                          cache keys. All other parameters are excluded. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                    type: object
                  clientTtl:
                    format: int64
                    type: integer
                  defaultTtl:
                    format: int64
                    type: integer
                  enabled:
                    type: boolean
                  maxTtl:
                    format: int64
                    type: integer
                  negativeCaching:
                    type: boolean
                  negativeCachingPolicy:
                    items:
                      description: NegativeCachingPolicy contains configuration for
                        how negative caching is applied.
                      properties:
                        code:
                          description: The HTTP status code to define a TTL against.
                            Only HTTP status codes 300, 301, 308, 404, 405, 410, 421,
                            451 and 501 are can be specified as values, and you cannot
                            specify a status code more than once.
                          format: int64
                          type: integer
                        ttl:
                          description: The TTL (in seconds) for which to cache responses
                            with the corresponding status code. The maximum allowed
                            value is 1800s (30 minutes), noting that infrequently
                            accessed objects may be evicted from the cache before
                            the defined TTL.
                          format: int64
                          type: integer
                      type: object
                    type: array
                  requestCoalescing:
                    type: boolean
                  serveWhileStale:
                    format: int64
                    type: integer
                  signedUrlCacheMaxAgeSec:
                    format: int64
                    type: integer
                  signedUrlKeys:
                    items:
                      description: SignedUrlKey represents a customer-supplied Signing
                        Key used by Cloud CDN Signed URLs
                      properties:
                        keyName:
                          description: 'KeyName: Name of the key. The name must be
                            1-63 characters long, and comply with RFC1035. Specifically,
                            the name must be 1-63 characters long and match the regular
                            expression `[a-z]([-a-z0-9]*[a-z0-9])?` which means the
                            first character must be a lowercase letter, and all following
                            characters must be a dash, lowercase letter, or digit,
                            except the last character, which cannot be a dash.'
                          type: string
                        keyValue:
                          description: 'KeyValue: 128-bit key value used for signing
                            the URL. The key value must be a valid RFC 4648 Section
                            5 base64url encoded string.'
                          type: string
                        secretName:
                          description: The name of a k8s secret which stores the 128-bit
                            key value used for signing the URL. The key value must
                            be a valid RFC 4648 Section 5 base64url encoded string
                          type: string
                      type: object
                    type: array
                required:
                - enabled
                type: object
              connectionDraining:
                description: ConnectionDrainingConfig contains configuration for connection
                  draining. For now the draining timeout. May manage more settings
                  in the future.
                properties:
                  drainingTimeoutSec:
                    description: Draining timeout in seconds.
                    format: int64
                    type: integer
                type: object
              customRequestHeaders:
                description: CustomRequestHeadersConfig contains configuration for
                  custom request headers
                properties:
                  headers:
                    items:
                      type: string
                    type: array
                type: object
              healthCheck:
                description: HealthCheckConfig contains configuration for the health
                  check.
                properties:
                  checkIntervalSec:
                    description: CheckIntervalSec is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  healthyThreshold:
                    description: HealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  port:
                    description: Port is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                      If Port is used, the controller updates portSpecification as
                      well
                    format: int64
                    type: integer
                  requestPath:
                    description: RequestPath is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  timeoutSec:
                    description: TimeoutSec is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  type:
                    description: Type is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  unhealthyThreshold:
                    description: UnhealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                type: object
              iap:
                description: IAPConfig contains configuration for IAP-enabled backends.
                properties:
                  enabled:
                    type: boolean
                  oauthclientCredentials:
                    description: OAuthClientCredentials contains credentials for a
                      single IAP-enabled backend.
                    properties:
                      clientID:
                        description: Direct reference to OAuth client id.
                        type: string
                      clientSecret:
                        description: Direct reference to OAuth client secret.
                        type: string
                      secretName:
                        description: The name of a k8s secret which stores the OAuth
                          client id & secret.
                        type: string
                    required:
                    - secretName
                    type: object
                required:
                - enabled
                - oauthclientCredentials
                type: object
              logging:
                description: LogConfig contains configuration for logging.
                properties:
                  enable:
                    description: This field denotes whether to enable logging for
                      the load balancer traffic served by this backend service.
                    type: boolean
                  sampleRate:
                    description: This field can only be specified if logging is enabled
                      for this backend service. The value of the field must be in
                      [0, 1]. This configures the sampling rate of requests to the
                      load balancer where 1.0 means all logged requests are reported
                      and 0.0 means no logged requests are reported. The default value
                      is 1.0.
                    format: double
                    type: number
                type: object
              securityPolicy:
                description: SecurityPolicyConfig contains configuration for CloudArmor-enabled
                  backends. If not specified, the controller will not reconcile the
                  security policy configuration. In other words, users can make changes
                  in GCE without the controller overwriting them.
                properties:
                  name:
                    description: Name of the security policy that should be associated.
                      If set to empty, the existing security policy on the backend
                      will be removed.
                    type: string
                required:
                - name
                type: object
              sessionAffinity:
                description: SessionAffinityConfig contains configuration for stickyness
                  parameters.
                properties:
                  affinityCookieTtlSec:
                    format: int64
                    type: integer
                  affinityType:
                    type: string
                type: object
              timeoutSec:
                format: int64
                type: integer
            type: object
          status:
            type: object
        type: object
    served: true
    storage: true
  - name: v1beta1
    schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            description: 'APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
            type: string
          kind:
            description: 'Kind is a string value representing the REST resource this
              object represents. Servers may infer this from the endpoint the client
              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
            type: string
          metadata:
            type: object
          spec:
            description: BackendConfigSpec is the spec for a BackendConfig resource
            properties:
              cdn:
                description: CDNConfig contains configuration for CDN-enabled backends.
                properties:
                  cachePolicy:
                    description: CacheKeyPolicy contains configuration for how requests
                      to a CDN-enabled backend are cached.
                    properties:
                      includeHost:
                        description: If true, requests to different hosts will be
                          cached separately.
                        type: boolean
                      includeProtocol:
                        description: If true, http and https requests will be cached
                          separately.
                        type: boolean
                      includeQueryString:
                        description: If true, query string parameters are included
                          in the cache key according to QueryStringBlacklist and QueryStringWhitelist.
                          If neither is set, the entire query string is included and
                          if false the entire query string is excluded.
                        type: boolean
                      queryStringBlacklist:
                        description: Names of query strint parameters to exclude from
                          cache keys. All other parameters are included. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                      queryStringWhitelist:
                        description: Names of query string parameters to include in
                          cache keys. All other parameters are excluded. Either specify
                          QueryStringBlacklist or QueryStringWhitelist, but not both.
                        items:
                          type: string
                        type: array
                    type: object
                  enabled:
                    type: boolean
                required:
                - enabled
                type: object
              connectionDraining:
                description: ConnectionDrainingConfig contains configuration for connection
                  draining. For now the draining timeout. May manage more settings
                  in the future.
                properties:
                  drainingTimeoutSec:
                    description: Draining timeout in seconds.
                    format: int64
                    type: integer
                type: object
              customRequestHeaders:
                description: CustomRequestHeadersConfig contains configuration for
                  custom request headers
                properties:
                  headers:
                    items:
                      type: string
                    type: array
                type: object
              healthCheck:
                description: HealthCheckConfig contains configuration for the health
                  check.
                properties:
                  checkIntervalSec:
                    description: CheckIntervalSec is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  healthyThreshold:
                    description: HealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  port:
                    format: int64
                    type: integer
                  requestPath:
                    description: RequestPath is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  timeoutSec:
                    description: TimeoutSec is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                  type:
                    description: Type is a health check parameter. See https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    type: string
                  unhealthyThreshold:
                    description: UnhealthyThreshold is a health check parameter. See
                      https://cloud.google.com/compute/docs/reference/rest/v1/healthChecks.
                    format: int64
                    type: integer
                type: object
              iap:
                description: IAPConfig contains configuration for IAP-enabled backends.
                properties:
                  enabled:
                    type: boolean
                  oauthclientCredentials:
                    description: OAuthClientCredentials contains credentials for a
                      single IAP-enabled backend.
                    properties:
                      clientID:
                        description: Direct reference to OAuth client id.
                        type: string
                      clientSecret:
                        description: Direct reference to OAuth client secret.
                        type: string
                      secretName:
                        description: The name of a k8s secret which stores the OAuth
                          client id & secret.
                        type: string
                    required:
                    - secretName
                    type: object
                required:
                - enabled
                - oauthclientCredentials
                type: object
              securityPolicy:
                description: SecurityPolicyConfig contains configuration for CloudArmor-enabled
                  backends. If not specified, the controller will not reconcile the
                  security policy configuration. In other words, users can make changes
                  in GCE without the controller overwriting them.
                properties:
                  name:
                    description: Name of the security policy that should be associated.
                      If set to empty, the existing security policy on the backend
                      will be removed.
                    type: string
                required:
                - name
                type: object
              sessionAffinity:
                description: SessionAffinityConfig contains configuration for stickyness
                  parameters.
                properties:
                  affinityCookieTtlSec:
                    format: int64
                    type: integer
                  affinityType:
                    type: string
                type: object
              timeoutSec:
                format: int64
                type: integer
            type: object
          status:
            type: object
        type: object
    served: true
    storage: false
status:
  acceptedNames:
    kind: BackendConfig
    listKind: BackendConfigList
    plural: backendconfigs
    shortNames:
    - bc
    singular: backendconfig
  conditions:
  - lastTransitionTime: "2022-03-10T23:03:02Z"
    message: no conflicts found
    reason: NoConflicts
    status: "True"
    type: NamesAccepted
  - lastTransitionTime: "2022-03-10T23:03:03Z"
    message: the initial names have been accepted
    reason: InitialNamesAccepted
    status: "True"
    type: Established
  storedVersions:
  - v1
 ```
  
  @nicknotfun & @alisspers , can you open GCP support tickets with the same information? Support and our team will be able to take a look at your specific clusters. I am not sure why your clusters do not have the updated CDN config options. Both of the control plane versions should have the newer CRD.

**Comment by alisspers** [comment]
@swetharepakula Yes, have a case open with the support now since a week, thanks. I'm not sure if it has come all the way back to your team yet (case id 29580635) but I hope I get some help there soon.

**Comment by swetharepakula** [comment]
@alisspers , the case has been routed to me. I am taking a look.

**Comment by swetharepakula** [comment]
@alisspers, we have found the root cause of the issue. Your case should be updated shortly.

@nicknotfun, if you have the same issue, you will need to open a ticket to be able to get your cluster fixed. The cluster operations required are not possible for the user to perform.

**Comment by alisspers** [comment]
@swetharepakula that sounds great, thanks! We have another cluster in our project, will that also be fixed or do I need to add that to the support case?

**Comment by swetharepakula** [comment]
@alisspers , we are looking on a solution to fix all clusters in this situation, but it will take some time. Not all clusters are affected, so please mention the other cluster on the support ticket so it will be looked at as well. If that needs to be fixed as well, support will take care of it as well.

**Comment by spencerhance** [comment]
@caseywebdev Can you send me an email with your cluster info?  I can take a look.

**Comment by caseywebdev** [comment]
@spencerhance sent, thanks!

**Comment by Dev25** [comment]
Has anyone actually managed to get this feature working on GKE? I have the CRD available with the new fields on the latest 1.23.x release but changing them does not get applied to the underling CDNs, in my case it's changing the cache mode to USE_ORIGIN_HEADERS and tweaking max TTLs. 

Annoyingly the [max TTL is now being forced to 1 day default](https://cloud.google.com/cdn/docs/using-ttl-overrides) which overrides the 1 year TTL sent by origins, so the default CDN TTL gets used which is 1 hour.

I've asked GCP support who mentioned there is a known problem and pointed me to https://issuetracker.google.com/issues/236035135

**Comment by gabihodoroaga** [comment]
Hi @Dev25  I tested this feature on GCP a few weeks ago and is working as expected.
To solve you problem take in consideration the following:
- Mac TTL is available only for only used when the cache mode is CACHE_ALL_STATIC. Check [documentation]( https://cloud.google.com/cdn/docs/using-ttl-overrides#max-ttl)
- In order for you content to be cashed you need to ensure the requirements for the [Cacheable content](https://cloud.google.com/cdn/docs/caching#cacheability)
For USE_ORIGIN_HEADERS mode the content is cache base solely on the `Cache-Control` header for the response.

**Comment by Dev25** [comment]
re: Origin headers mode, I'm not even getting as far for the request requirements to matter. The cache mode for the CDN resource itself (reported via console/cli) remains stuck on the CACHE_ALL_STATIC default it had before, and any request served ends up using that cache mode (with the 1hr TTL enforced thanks to the mode)

I'll try creating a new backend service to see if that works.

**Comment by gabihodoroaga** [comment]
@Dev25 I just tested on a new cluster and it works as expected. This is the configuration I used:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - name: http
          containerPort: 80
---
kind: BackendConfig
metadata:
  name: nginx-bc
spec:
  cdn:
    enabled: true
    cacheMode: USE_ORIGIN_HEADERS
    cachePolicy:
      includeHost: true
      includeProtocol: true
      includeQueryString: true
    negativeCaching: false
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
  annotations:
    cloud.google.com/backend-config: '{"default": "nginx-bc"}'
spec:
  type: NodePort
  ports:
  - port: 80
    targetPort: 80
  selector:
    app: nginx
```

Make sure you have the proper annotation on the Service:
```yaml
  annotations:
    cloud.google.com/backend-config: '{"default": "nginx-bc"}'
```

Dev25:
So it turns out Ingress supports the new CDN fields just fine but MultiClusterIngress which links to the same v1 BackendConfig does not with no ETA for that support.

spencerhance:
Improving the CDN support for MCI is a feature request for MCI that we are considering.  Unfortunately, this repo contains GKE Ingress (i.e single-cluster ingress).  

The right place to file a public issue: https://cloud.google.com/support/docs/issue-trackers#trackers-list

spencerhance:
@dpkirchner Ah sorry I misconstrued your comment as referencing the one directly above it.  Your issue is that your control plane version is too old.

https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#cloud_cdn

```
The following fields are only supported in GKE versions 1.23.3-gke.900 and later using GKE Ingress. They are not supported using Multi Cluster Ingress:
```

dpkirchner:
Huh, ok. I'm surprised that the schema version wasn't bumped when the fields were added. Thanks for spotting that, I totally missed it.

PR #1390, opened 2021-04-01
Remove empty error from error message

swetharepakula:
Since error is always nil, remove from error message.


MrHohn (approved):
/lgtm
/approve

PR #1391, opened 2021-04-01
PSC Lifecycle E2E Test

swetharepakula:
/assign @freehan

### DIFF
--- a/cmd/e2e-test/psc_test.go
+++ b/cmd/e2e-test/psc_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/legacy-cloud-providers/gce"
+)
+
+func TestPSCLifecycle(t *testing.T) {
+	t.Parallel()
+
+	Framework.RunWithSandbox("PSC Lifecycle", t, func(t *testing.T, s *e2e.Sandbox) {
+		svcName := fmt.Sprintf("ilb-service-e2e-%x", s.RandInt)
+		saName := fmt.Sprintf("service-attachment-e2e-%x", s.RandInt)
+
+		// PSC requires subnet to have purpose PRIVATE_SERVICE_CONNECT
+		err := e2e.CreateSubnet(s, e2e.PSCSubnetName, e2e.PSCSubnetPurpose)
+		defer e2e.DeleteSubnet(s, e2e.PSCSubnetName)
+		if err != nil {
+			// Subnet Creation could fail because of a leaked subnet from a previous run.
+			// Instead of failing the test, attempt to reuse subnet.
+			t.Logf("error creating subnet %s: %q", e2e.PSCSubnetName, err)
+		} else {
+			t.Logf("created subnet for PSC")
+		}
+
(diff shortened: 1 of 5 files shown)


pkg/e2e/fixtures.go
```diff
@@ -339,22 +341,26 @@ func DeleteGCPAddress(s *Sandbox, name string, region string) error {
 // CreateILBSubnet creates the ILB subnet
 func CreateILBSubnet(s *Sandbox) error {
 	klog.V(2).Info("CreateILBSubnet()")
+	return CreateSubnet(s, "ilb-subnet-ingress-e2e", "INTERNAL_HTTPS_LOAD_BALANCER")
```
freehan, line 344: a const for `INTERNAL_HTTPS_LOAD_BALANCER`?

cmd/e2e-test/psc_test.go
```diff
@@ -0,0 +1,108 @@
...
+	t.Parallel()
+
+	testcases := []struct {
+		desc                 string
+		createSubnet         bool
+		createService        bool
+		expectedErrorMessage string
+	}{
+		{
+			desc:                 "subnet was not created",
```
freehan, line 38: I am usually against negative tests. Unless it is a critical one. I do not think these are that critical because if the setup is incorrect, it would not work upfront. 

An alternative is to make this into a transition tests? Basically, 1. missing both 2. create subnet and miss service 3. have both.  But it is more vulnerable to flakes.

swetharepakula: That is a good point. I removed the negative paths and just kept happy path.

pkg/e2e/fixtures.go
swetharepakula: Pulled the purposes into constants in the e2e package.

pkg/e2e/fixtures.go
```diff
@@ -339,22 +343,26 @@ func DeleteGCPAddress(s *Sandbox, name string, region string) error {
 // CreateILBSubnet creates the ILB subnet
 func CreateILBSubnet(s *Sandbox) error {
 	klog.V(2).Info("CreateILBSubnet()")
+	return CreateSubnet(s, "ilb-subnet-ingress-e2e", ILBSubnetPurpose)
```
freehan, line 346: make the ilb subnet name into a const as well.

cmd/e2e-test/psc_test.go
```diff
@@ -0,0 +1,69 @@
...
+	Framework.RunWithSandbox("PSC Lifecycle", t, func(t *testing.T, s *e2e.Sandbox) {
+		var (
+			svcName    = fmt.Sprintf("ilb-service-e2e-%x", s.RandInt)
+			subnetName = fmt.Sprintf("psc-nat-subnet-e2e-%x", s.RandInt)
+			saName     = fmt.Sprintf("service-attachment-e2e-%x", s.RandInt)
+		)
+
+		// PSC requires subnet to have purpose PRIVATE_SERVICE_CONNECT
+		err := e2e.CreateSubnet(s, subnetName, e2e.PSCSubnetPurpose)
+		defer e2e.DeleteILBSubnet(s, subnetName)
```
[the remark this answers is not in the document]
freehan: also probably worth logging the error message. But not fail the test on this.

cmd/e2e-test/psc_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	Framework.RunWithSandbox("PSC Lifecycle", t, func(t *testing.T, s *e2e.Sandbox) {
+		var (
+			svcName    = fmt.Sprintf("ilb-service-e2e-%x", s.RandInt)
+			subnetName = fmt.Sprintf("psc-nat-subnet-e2e-%x", s.RandInt)
+			saName     = fmt.Sprintf("service-attachment-e2e-%x", s.RandInt)
+		)
+
+		// PSC requires subnet to have purpose PRIVATE_SERVICE_CONNECT
+		err := e2e.CreateSubnet(s, subnetName, e2e.PSCSubnetPurpose)
```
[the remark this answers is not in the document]
swetharepakula: Switched to using a static name.

pkg/e2e/fixtures.go
swetharepakula: Added the subnet names as constants.

cmd/e2e-test/psc_test.go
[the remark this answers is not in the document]
swetharepakula: Changed `DeleteILBSubnet` to `DeleteSubnet`.

Changed to logging the message. Added a comment of in what situation the test will still pass even though subnet creation failed.

freehan (approved):
/lgtm
/approve

PR #1392, opened 2021-04-02
Feature Gate Ingress GA features

swetharepakula:
- If disabled, only support ImplementationSpecific path type
 - If disabled, do not ensure GCPIngressParams CRD


swetharepakula:
One option could be supporting the new PathTypes regardless of the feature gate.

freehan (approved):
/lgtm
/approve

PR #1395, opened 2021-04-06
Continue syncing HTTPS if there is at least one valid TLS secret

spencerhance:
This fixes a bug where:
1) User has existing HTTPS LB with TLS secrets
2) User adds a new TLS secret that doesn't exist yet (could be added by another controller)
3) Ingress controller sets up HTTP only, causing their HTTPS resources to be deleted and causes downtime.

What this PR does is continue syncing the HTTPS LB as long as there is at least 1 valid secret.

cc: #388


MrHohn (approved):
Thanks for fixing this up!
/lgtm
/approve

Issue #1394, opened 2021-04-06
The function `Backends.Health()` actually returns a status of the first instance of the first Backend instead of the status of the Backend Service.


sio4:
The function `Backends.Health()` has one meaningful `return` statement [in line 209](https://github.com/kubernetes/ingress-gce/blob/7422259dba059dd3af0017ada779742bf0e5d9a9/pkg/backends/backends.go#L209) but the statement is within `for` loop for the `range` of backends and it returns the first element's value as a result of the function. So the meaning of the function's return value is actually the status of the first instance of the first backend of the given backend service.

https://github.com/kubernetes/ingress-gce/blob/7422259dba059dd3af0017ada779742bf0e5d9a9/pkg/backends/backends.go#L178-L212

It should return a meaningful status of the backend service and we need to declare a criteria when we call the backend service is healthy. Could we improve this?

sio4:
the issue was fixed by #1398. closing it.

PR #1398, opened 2021-04-07
fix the behavior of Backends.Health() to return more reliable value

sio4:
Currently the function Backends.Health() returns a status of the first instance of the first backend of the given backend service. (issue #1394) This PR will fix the issue by make the function to return `Healthy` when at lease one healthy instance exists.

So the possible cases are:

1) return `Healthy` when at least one healthy instance exists
2) return `Unhealthy` when all instances are unhealthy (at least one)
3) return `Unknown` when the hs.HealthStatus is empty or nil or something unknown...

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -186,6 +186,7 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 
 	// TODO: Include port, ip in the status, since it's in the health info.
 	// TODO (shance) convert to composite types
+	ret := "Unknown"
 	for _, backend := range be.Backends {
 		var hs *compute.BackendServiceGroupHealth
 		switch scope {
@@ -205,10 +206,15 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 			continue
 		}
 
-		// TODO: State transition are important, not just the latest.
-		return hs.HealthStatus[0].HealthState, nil
+		for _, instanceStatus := range hs.HealthStatus {
+			ret = instanceStatus.HealthState
+			// return immediately with the value if we found at least one healthy instance
+			if ret == "HEALTHY" {
+				return ret, nil
+			}
+		}
 	}
-	return "Unknown", nil
+	return ret, nil
 }
 
 // List lists all backends managed by this controller.


sio4:
CLA signed

spencerhance (approved):
lgtm

pkg/backends/backends.go
```diff
@@ -186,6 +186,7 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 
 	// TODO: Include port, ip in the status, since it's in the health info.
 	// TODO (shance) convert to composite types
+	ret := "Unknown"
```
freehan, line 189: make `Unkown` into a const value?

pkg/backends/backends.go
```diff
@@ -205,10 +206,15 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
 			continue
 		}
 
-		// TODO: State transition are important, not just the latest.
-		return hs.HealthStatus[0].HealthState, nil
+		for _, instanceStatus := range hs.HealthStatus {
+			ret = instanceStatus.HealthState
+			// return immediately with the value if we found at least one healthy instance
+			if ret == "HEALTHY" {
```
freehan, line 212: make HEALTHY into a const value?

freehan (commented):
With this change, it either returns `Unkown` or `HEALTHY`. 

How about we do this?
Return `UNHEALTHY` when all backends are `UNHEALTHY?
Return `HEALTHY` when at least one is `HEALTHY`
Return `Unknown` otherwise?

pkg/backends/backends.go
sio4: Yeah, I think that is a good idea, but I am very new to this project and not sure which is the best place to add the const. any suggestion?

sio4: Can it be better with this? then I will commit this too.

```diff
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -32,6 +32,9 @@ import (
 
 const (
        DefaultConnectionDrainingTimeoutSeconds = 30
+       Status_Healthy                          = "HEALTHY"
+       Status_Unhealthy                        = "UNHEALTHY"
+       Status_Unknown                          = "Unknown"
 )
 
 // Backends handles CRUD operations for backends.
@@ -178,15 +181,15 @@ func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType)
 func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType) (string, error) {
        be, err := b.Get(name, version, scope)
        if err != nil {
-               return "Unknown", fmt.Errorf("error getting backend service %s: %v", name, err)
+               return Status_Unknown, fmt.Errorf("error getting backend service %s: %v", name, err)
        }
        if len(be.Backends) == 0 {
-               return "Unknown", fmt.Errorf("no backends found for backend service %q", name)
+               return Status_Unknown, fmt.Errorf("no backends found for backend service %q", name)
        }
 
        // TODO: Include port, ip in the status, since it's in the health info.
        // TODO (shance) convert to composite types
-       ret := "Unknown"
+       ret := Status_Unknown
        for _, backend := range be.Backends {
                var hs *compute.BackendServiceGroupHealth
                switch scope {
@@ -195,11 +198,11 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
                case meta.Regional:
                        hs, err = b.cloud.GetRegionalBackendServiceHealth(name, b.cloud.Region(), backend.Group)
                default:
-                       return "Unknown", fmt.Errorf("invalid scope for Health(): %s", scope)
+                       return Status_Unknown, fmt.Errorf("invalid scope for Health(): %s", scope)
                }
 
                if err != nil {
-                       return "Unknown", fmt.Errorf("error getting health for backend %q: %v", name, err)
+                       return Status_Unknown, fmt.Errorf("error getting health for backend %q: %v", name, err)
                }
                if len(hs.HealthStatus) == 0 || hs.HealthStatus[0] == nil {
                        klog.V(3).Infof("backend service %q does not have health status: %v", name, hs.HealthStatus)
@@ -209,7 +212,7 @@ func (b *Backends) Health(name string, version meta.Version, scope meta.KeyType)
                for _, instanceStatus := range hs.HealthStatus {
                        ret = instanceStatus.HealthState
                        // return immediately with the value if we found at least one healthy instance
-                       if ret == "HEALTHY" {
+                       if ret == Status_Healthy {
                                return ret, nil
                        }
                }
```

PR #1404, opened 2021-04-14
Fix log line in deleteForwardingRule

spencerhance:
Removes "Global" so that it's not confusing to debug when deleting a Regional Forwarding Rule.


MrHohn (approved):
/lgtm
/approve

PR #1405, opened 2021-04-16
Use Ingress v1 API

swetharepakula:
* GLBC will use V1 API through a shim client when v1 API is available


pkg/adapterclient/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil
+	}
+	if in.DefaultBackend != nil {
+		out.Backend = &v1beta1.IngressBackend{}
+		if err := Convert_networking_IngressBackend_To_v1beta1_IngressBackend(in.DefaultBackend, out.Backend, s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
liggitt, line 81: I would not recommend generating external<->external conversions... instead, I would do something like this:

```go
import (
	v1 "k8s.io/api/networking/v1"
	v1beta1 "k8s.io/api/networking/v1beta1"
	internalnetworking "k8s.io/kubernetes/pkg/apis/networking"
	v1conversion "k8s.io/kubernetes/pkg/apis/networking/v1"
	v1beta1conversion "k8s.io/kubernetes/pkg/apis/networking/v1beta1"
)

func Convert_v1beta1_Ingress_To_v1_Ingress(in *v1beta1.Ingress, out *v1.Ingress) error {
	intermediate := &internalnetworking.Ingress{}
	if err := v1beta1conversion.Convert_v1beta1_Ingress_To_networking_Ingress(in, intermediate, nil); err != nil {
		return err
	}
	if err := v1conversion.Convert_networking_Ingress_To_v1_Ingress(intermediate, out, nil); err != nil {
		return err
	}
	return nil
}

func Convert_v1_Ingress_To_v1beta1_Ingress(in *v1.Ingress, out *v1beta1.Ingress) error {
	intermediate := &internalnetworking.Ingress{}
	if err := v1conversion.Convert_v1_Ingress_To_networking_Ingress(in, intermediate, nil); err != nil {
		return err
	}
	if err := v1beta1conversion.Convert_networking_Ingress_To_v1beta1_Ingress(intermediate, out, nil); err != nil {
		return err
	}
	return nil
}

func Convert_v1beta1_IngressClass_To_v1_IngressClass(in *v1beta1.IngressClass, out *v1.IngressClass) error {
	intermediate := &internalnetworking.IngressClass{}
	if err := v1beta1conversion.Convert_v1beta1_IngressClass_To_networking_IngressClass(in, intermediate, nil); err != nil {
		return err
	}
	if err := v1conversion.Convert_networking_IngressClass_To_v1_IngressClass(intermediate, out, nil); err != nil {
		return err
	}
	return nil
}

func Convert_v1_IngressClass_To_v1beta1_IngressClass(in *v1.IngressClass, out *v1beta1.IngressClass) error {
	intermediate := &internalnetworking.IngressClass{}
	if err := v1conversion.Convert_v1_IngressClass_To_networking_IngressClass(in, intermediate, nil); err != nil {
		return err
	}
	if err := v1beta1conversion.Convert_networking_IngressClass_To_v1beta1_IngressClass(intermediate, out, nil); err != nil {
		return err
	}
	return nil
}
```

pkg/adapterclient/adapterclient.go
```diff
@@ -0,0 +1,82 @@
...
+			Interface: kubeClient,
+		}, true, nil
+	}
+	return kubeClient, false, nil
+}
+
+// NetworkingV1beta1 returns an custom NetworkingV1beta1 Client
+func (c *AdapterClient) NetworkingV1beta1() networkingv1beta1.NetworkingV1beta1Interface {
+	return &NetworkingClient{
+		NetworkingV1beta1Interface: c.Interface.NetworkingV1beta1(),
```
liggitt, line 50: I would avoid this, it makes it likely v1beta1 calls will be made you didn't intend... this client should be implemented completely using the v1 client, right?

swetharepakula:
This approach adds a layer of risk and complexity into the controller and would still require a second migration that would remove v1beta1 completely from the controller.

Because of that, I will close this PR and open a new one with a different approach in which the complete migration will be done to remove v1beta1.

PR #1406, opened 2021-04-20
Fix cloud armor security policy workflow

skmatti:
1. This fixes an error where backend service scope is empty.
We emit an error that the security policy is not supported
if the scope is empty.
2. Uses full security policy resource link when updating
the backend service.


MrHohn (approved):
/lgtm
/approve

PR #1408, opened 2021-04-20
Cherrypick #1406[Fix cloud armor security policy workflow] into release-1.11

skmatti:
1. This fixes an error where backend service scope is empty.
We emit an error that the security policy is not supported
if the scope is empty.
2. Uses full security policy resource link when updating
the backend service.

/assign @MrHohn


MrHohn (approved):
/approve
/lgtm

PR #1415, opened 2021-04-23
Remove Enable L7ILB Feature Gate 

swetharepakula:
Some neg controller unit tests only pass because when all tests are run and fail when run individually. This is because the EnableL7ILB flag is set in one test and all tests after that have the same flag value. This PR cleans up the test contamination and ensures that the tests will pass when run independently or together.

/assign @freehan

### DIFF
--- a/pkg/annotations/ingress.go
+++ b/pkg/annotations/ingress.go
@@ -21,7 +21,6 @@ import (
 	"strconv"
 
 	"k8s.io/api/networking/v1beta1"
-	"k8s.io/ingress-gce/pkg/flags"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


spencerhance:
Do we still have a use case for keeping this flag around?

We can probably remove it entirely and remove the test cases for it being disabled

swetharepakula:
I have updated this PR to removing the `EnableL7Ilb` flag and updating logic so L7 ILB is always enabled.

@spencerhance

spencerhance (approved):
lgtm, thanks!

@freehan

freehan:
LGTM
/approve

wait for @spencerhance to take a look.

PR #1416, opened 2021-04-23
Convert to use V1 Ingress API

swetharepakula:
This is a large PR and touches many parts of the code. Currently it is based on #1415 , which I will update soon.

While reviewing please keep the following in mind:
**1. E2E tests**
* Tests need to pass for Clusters 1.15 - 1.18 if applicable (where v1 API is unavailable)
* Tests need to pass for Clusters 1.22+  (where v1beta1 API is removed)

**2. Ingress Resource Changes**
`Service.Port` (on the Ingress resource) used to be of `type IntOrString`, but now it is:
```
type ServiceBackendPort struct with {
    Number int32
    Name string
}
```
* Internally the ServiceBackendPort will be used throughout the controller
* Watch out for the `.String()` on that port as before it would return the value in string format but now will return the entire struct

**3. Any accidental change in behavior due to the conversion or the Service.Port change that was made**

**4. Backend services in the Ingress spec can now be nil**


pkg/neg/controller.go
```diff
@@ -902,3 +903,10 @@ func getDestinationRulesFromStore(store cache.Store, svc *apiv1.Service) (drs ma
 	}
 	return
 }
+
+func getServiceBackendPortName(port v1.ServiceBackendPort) string {
```
swetharepakula, line 907: Add a comment

liggitt:
in main.go, I'd suggest replacing this:
```
	ingClassEnabled := flags.F.EnableIngressGAFields && app.IngressClassEnabled(kubeClient)
```

with a check that requires v1 Ingress and IngressClass and exits with an error if those APIs aren't available, rather than acquiring leader election

pkg/common/operator/service.go
```diff
@@ -60,7 +60,7 @@ func (op *ServicesOperator) ReferencedByIngress(ing *v1beta1.Ingress) *ServicesO
 
 // doesIngressReferenceService returns true if the passed in Ingress directly references
 // the passed in Service.
-func doesIngressReferenceService(ing *v1beta1.Ingress, svc *api_v1.Service) bool {
+func doesIngressReferenceService(ing *v1.Ingress, svc *api_v1.Service) bool {
 	if ing.Namespace != svc.Namespace {
 		return false
 	}
```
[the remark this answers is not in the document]
swetharepakula: Oh that is a good catch. I was planning on addressing that in a different PR, but it probably does make more sense to take care of it with the version switch.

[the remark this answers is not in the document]
liggitt: in v1beta1, ServiceName and ServicePort were never nil for a resource-backed Ingress, merely empty

in v1, the Service field is nil for resource-backed Ingress objects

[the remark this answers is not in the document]
swetharepakula: TraverseIngressBackends can now handle when Service is nil.

pkg/e2e/adapter/conversion.go
```diff
@@ -0,0 +1,81 @@
...
+// internal ingress structure
+
+import (
+	networking "k8s.io/api/networking/v1"
+	v1beta1 "k8s.io/api/networking/v1beta1"
+	conversion "k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func Convert_v1beta1_IngressBackend_To_networking_IngressBackend(in *v1beta1.IngressBackend, out *networking.IngressBackend, s conversion.Scope) error {
```
liggitt, line 30: why do we still need non-standard conversion? I expected all the tests and clients to only use the v1 APIs

pkg/e2e/adapter/ingress.go
```diff
@@ -24,55 +23,94 @@ type IngressCRUD struct {
 }
 
 // Get Ingress resource.
-func (crud *IngressCRUD) Get(ns, name string) (*v1beta1.Ingress, error) {
-	new, err := crud.supportsNewAPI()
+func (crud *IngressCRUD) Get(ns, name string) (*v1.Ingress, error) {
+	useV1, err := crud.supportsV1API()
```
liggitt, line 27: if this version will only support the v1 API, we should be able to make a failure here return an error and drop all the v1beta1 code and paths, right?

liggitt: or even just drop the supportsV1API check altogether and use NetworkingV1 directly

pkg/e2e/adapter/ingress.go
```diff
@@ -81,39 +119,39 @@ func (crud *IngressCRUD) Patch(oldIng, newIng *v1beta1.Ingress) (*v1beta1.Ingres
...
 	}
-	klog.Warning("Using legacy API")
-	return crud.C.ExtensionsV1beta1().Ingresses(ns).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	klog.Warning("Using v1beta1 API")
+	return crud.C.NetworkingV1beta1().Ingresses(ns).Delete(context.TODO(), name, metav1.DeleteOptions{})
 }
 
-func (crud *IngressCRUD) supportsNewAPI() (bool, error) {
+func (crud *IngressCRUD) supportsV1API() (bool, error) {
 	if apiList, err := crud.C.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1beta1"); err == nil {
```
liggitt, line 155: I'm not sure we even need supportsV1API, but if we do, this should be checking `networking.k8s.io/v1`

liggitt:
is there a check early in main.go now to ensure that v1 Ingresses are supported? would that make sense to add to ensure we don't get in a state where we're running with v1 informers that won't sync (xref https://github.com/kubernetes/kubernetes/pull/99840#issuecomment-829259136)

also, this appears to be blocking a successful run of the kubernetes/kubernetes PR that tests without any beta APIs enabled (the gce cluster won't start with this controller crashlooping on extensions/v1beta1 being missing)

pkg/e2e/adapter/conversion.go
swetharepakula: We still need the conversions because the same e2e tests are run against different kubernetes minor versions so the tests need to be able to handle v1 and v1beta1.

pkg/e2e/adapter/ingress.go
swetharepakula: Same tests are run against different kubernetes minor versions so the tests need to be able to handle v1 and v1beta1.

pkg/e2e/adapter/ingress.go
swetharepakula: Thanks for catching that!

pkg/e2e/adapter/conversion.go
liggitt: hmm.. ok... what are the implications if these conversions are incorrect?

swetharepakula: It would affect our e2e test runs. If the conversions fail, test runs against 1.18 clusters and older would error. Though relatively this is a a lower risk as we can always fix the tests as needed. The conversion code will not be part of the ingress-controller binary/image .

liggitt: ok, I'm not thrilled with the idea of duplicating conversions here, but if it is completely limited to test code against older servers, that's tolerable

swetharepakula:
There was one more call to ExtensionsV1beta1() that I missed: https://github.com/kubernetes/ingress-gce/blob/master/cmd/glbc/app/namer.go#L164 which was causing the failure in #99840. 

I have fixed it in this PR. Once this is merged and a release is cut I will update the image being used in the glbc addon.

pkg/neg/controller.go
```diff
@@ -898,3 +899,12 @@ func getDestinationRulesFromStore(store cache.Store, svc *apiv1.Service) (drs ma
 	}
 	return
 }
+
+// getServiceBackendPortName returns the name the should be used as the Neg key in the status
+// annotation
+func getServiceBackendPortName(port v1.ServiceBackendPort) string {
```
freehan, line 905: I am wondering if this should just return the port Name no matter if it is empty string or not?
If you return the string of port number. It may not be able to find the port name in the service spec

pkg/neg/controller.go
```diff
@@ -570,13 +571,13 @@ func (c *Controller) mergeDefaultBackendServicePortInfoMap(key string, service *
...
+	scanIngress := func(qualify func(*v1.Ingress) bool) error {
 		for _, m := range c.ingressLister.List() {
-			ing := *m.(*v1beta1.Ingress)
-			if qualify(&ing) && ing.Spec.Backend == nil {
+			ing := *m.(*v1.Ingress)
+			if qualify(&ing) && ing.Spec.DefaultBackend == nil {
 				svcPortTupleSet := make(negtypes.SvcPortTupleSet)
 				svcPortTupleSet.Insert(negtypes.SvcPortTuple{
-					Name:       c.defaultBackendService.ID.Port.String(),
+					Name:       getServiceBackendPortName(c.defaultBackendService.ID.Port),
```
freehan, line 580: I am wondering if it should just get the port Name here. 

The previous solution is a bit hacky as the defaultBackendService port are provided as an flag using port name `http`.

cmd/glbc/app/init.go
```diff
@@ -53,7 +54,7 @@ func DefaultBackendServicePort(kubeClient kubernetes.Interface) utils.ServicePor
 		klog.Fatalf("Failed to verify default backend service: %v", err)
 	}
 
-	backendPort := intstr.FromString(flags.F.DefaultSvcPortName)
+	backendPort := networkingv1.ServiceBackendPort{Name: flags.F.DefaultSvcPortName}
```
freehan, line 57: how about we just pass `flags.F.DefaultSvcPortName` as string into `servicePortForDefaultService`, then construct the `networkingv1.ServiceBackendPort` using the port name directly.

This change essentially limit `flags.F.DefaultSvcPortName` flag to only take port name and not port number

pkg/neg/controller.go
freehan: This function may not be needed if `flags.F.DefaultSvcPortName` only can specify port name

pkg/utils/utils_test.go
```diff
@@ -354,26 +390,89 @@ func TestTraverseIngressBackends(t *testing.T) {
...
+													Name: "foo-service",
+													Port: networkingv1.ServiceBackendPort{
+														Number: 80,
+													},
+												},
+											},
+										},
+										{
+											Path:    "/non-service",
+											Backend: networkingv1.IngressBackend{},
```
robscott, line 459: You may already have this covered in a different test I missed, but it might be helpful to have e2e coverage for more variations of this. For example, an Ingress with only one Backend, and that Backend is a reference to some arbitrary object like a non-existent ConfigMap or something.

pkg/neg/controller.go
swetharepakula: This function is no longer needed once the flag is forced to only provide a name.

pkg/neg/controller.go
swetharepakula: I like this approach. Newest patch is updated to use port Name here.

cmd/glbc/app/init.go
swetharepakula: This makes the logic a little cleaner. Newest patch will assume that the flag only takes port name now.

pkg/utils/utils_test.go
swetharepakula: We don't have an e2e test, but there a few other combinations that are tested in the controller code: https://github.com/swetharepakula/ingress-gce/blob/use-ingress-v1/pkg/controller/translator/translator_test.go#L174-L188

swetharepakula:
Only [unaddressed comment](https://github.com/kubernetes/ingress-gce/pull/1416#issuecomment-827093604) is about the startup checks for IngressClass and the IngressGAFields.

Since this is already a large PR with a lot of subtle changes, I will address the remaining comment in a followup PR.

Otherwise I think this PR is ready for a final set of reviews and is almost ready to be merged.
Thanks everyone for all the reviews and feedback!

pkg/e2e/adapter/ingress.go
```diff
@@ -24,55 +23,94 @@ type IngressCRUD struct {
...
-		return crud.C.NetworkingV1beta1().Ingresses(ing.Namespace).Create(context.TODO(), ing, metav1.CreateOptions{})
+	if useV1 {
+		return crud.C.NetworkingV1().Ingresses(ing.Namespace).Create(context.TODO(), ing, metav1.CreateOptions{})
 	}
-	klog.Warning("Using legacy API")
-	legacyIng := toIngressExtensionsGroup(ing)
-	legacyIng, err = crud.C.ExtensionsV1beta1().Ingresses(ing.Namespace).Create(context.TODO(), legacyIng, metav1.CreateOptions{})
-	return toIngressNetworkingGroup(legacyIng), err
+
+	klog.Warning("Using v1beta1 API")
```
prameshj, line 54: Should we add a TODO to clean this up once it is no longer needed?

pkg/e2e/adapter/ingress.go
```diff
@@ -81,47 +119,47 @@ func (crud *IngressCRUD) Patch(oldIng, newIng *v1beta1.Ingress) (*v1beta1.Ingres
...
+func (crud *IngressCRUD) supportsV1API() (bool, error) {
+	if apiList, err := crud.C.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1"); err == nil {
 		for _, r := range apiList.APIResources {
 			if r.Kind == "Ingress" {
 				return true, nil
 			}
 		}
 	}
-	if apiList, err := crud.C.Discovery().ServerResourcesForGroupVersion("extensions/v1beta1"); err == nil {
+	if apiList, err := crud.C.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1beta1"); err == nil {
```
prameshj, line 162: nit: make "networking.k8s.io/v1beta1" and "networking.k8s.io/v1" as const?

pkg/e2e/adapter/ingress_test.go
```diff
@@ -5,45 +5,53 @@ import (
...
 	} {
-		gotExt := toIngressExtensionsGroup(tc.net)
-		gotNet := toIngressNetworkingGroup(tc.ext)
+		gotV1, err := toIngressV1(tc.netV1beta1)
+		if err != nil {
+			t.Errorf("got error in converstion from v1 to v1beta1: %q", err)
+		}
+		gotV1beta1, err := toIngressV1beta1(tc.netV1)
+		if err != nil {
+			t.Errorf("got error in converstion from v1beta1 to v1: %q", err)
```
prameshj, line 47: typo: "conversion"
here and 3 lines above.

liggitt:
happy to see the land. is the next step a v1.11.2 and a bump in kubernetes/kubernetes?

swetharepakula:
We will be cutting a 1.12 release soon. Just finishing up some testing before we cut the tag so should be ready in about a day.

swetharepakula:
https://github.com/kubernetes/kubernetes/pull/101772 updates the glbc image which should unblock https://github.com/kubernetes/kubernetes/pull/99840

PR #1417, opened 2021-04-27
Neg metrics

swetharepakula:
* Fix the latency metrics buckets.
* Measure only the controller processing time for GC and exclude latency added from the K8s and GCE API calls

/cc @prameshj 
/assign @freehan

### DIFF
--- a/pkg/neg/metrics/metrics.go
+++ b/pkg/neg/metrics/metrics.go
@@ -71,6 +71,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      negOpLatencyKey,
 			Help:      "Latency of a NEG Operation",
+			// custom buckets - [1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048 (~34min), 4096(~68min), +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
 		},
 		negOpLatencyMetricsLabels,
 	)
@@ -89,6 +91,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      syncerLatencyKey,
 			Help:      "Sync latency for NEG Syncer",
+			// custom buckets - [1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048 (~34min), 4096(~68min), +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
 		},
 		syncerMetricsLabels,
 	)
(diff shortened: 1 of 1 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -71,6 +71,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      negOpLatencyKey,
 			Help:      "Latency of a NEG Operation",
+			// custom buckets - [30s, 60s, 120s, 240s(4min), 480s(8min), 960s(16m), +Inf]
+			Buckets: prometheus.ExponentialBuckets(15, 2, 7),
```
freehan, line 75: I am wondering why not start with 1s.  

```
Buckets: prometheus.ExponentialBuckets(1, 2, 10),
```

I am asking because vast majority of the operation will be finish within 15s. Depending on LB config, some operation may take 1 hour. (e.g. DetachNE operation depends on the connection draining timeout config, which can be maximum of 1 hour)

pkg/neg/metrics/metrics.go
```diff
@@ -89,6 +91,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      syncerLatencyKey,
 			Help:      "Sync latency for NEG Syncer",
+			// custom buckets - [30s, 60s, 120s, 240s(4min), 480s(8min), 960s(16m), +Inf]
+			Buckets: prometheus.ExponentialBuckets(15, 2, 7),
```
freehan, line 95: Most of the case, this would finish within 1 second

pkg/neg/metrics/metrics.go
```diff
@@ -98,6 +102,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      managerProcessLatencyKey,
 			Help:      "Process latency for NEG Manager",
+			// custom buckets - [30s, 60s, 120s, 240s(4min), 480s(8min), 960s(16m), +Inf]
+			Buckets: prometheus.ExponentialBuckets(15, 2, 7),
```
freehan, line 106: same here. most cases would finish within 1 s

pkg/neg/metrics/metrics.go
swetharepakula: I didn't think there was too much gain in having more granular buckets and that the real focus should be in ensuring all data points fall into a bucket, so I tried to make sure the last bucket was at least 15 min and that we didn't create too many extra buckets overall.

I don't see a problem starting with 1s. So with the proposed change the buckets would be: 

`1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 512s, 1024s (~17 min)`

I will extend to 12 buckets to ensure we capture operations that could take an hour.

freehan (approved):
/lgtm
/approve

PR #1420, opened 2021-04-29
Add success and error neg counts

swetharepakula:
* successful syncer creations are tracked as part of the feature usage metrics
 
 
 /assign @freehan

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -464,8 +464,10 @@ func (c *Controller) processService(key string) error {
 		if err := svcPortInfoMap.Merge(destinationRulesPortInfoMap); err != nil {
 			return fmt.Errorf("failed to merge service ports referenced by Istio:DestinationRule (%v): %v", destinationRulesPortInfoMap, err)
 		}
+
+		negUsage.SuccessfulNeg, negUsage.ErrorNeg, err = c.manager.EnsureSyncers(namespace, name, svcPortInfoMap)
 		c.collector.SetNegService(key, negUsage)
-		return c.manager.EnsureSyncers(namespace, name, svcPortInfoMap)
+		return err
 	}
 
 	// do not need Neg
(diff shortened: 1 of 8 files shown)


pkg/neg/controller.go
```diff
@@ -464,8 +464,10 @@ func (c *Controller) processService(key string) error {
 		if err := svcPortInfoMap.Merge(destinationRulesPortInfoMap); err != nil {
 			return fmt.Errorf("failed to merge service ports referenced by Istio:DestinationRule (%v): %v", destinationRulesPortInfoMap, err)
 		}
+
+		negUsage.SuccessfulNeg, negUsage.ErrorNeg, err = c.manager.EnsureSyncers(namespace, name, svcPortInfoMap)
```
prameshj, line 468: Does valid syncer for the service mean valid NEG? The count is not 1:1 right? A single Syncer can create multiple NEGs(one per zone).

Also, should we expose the successful/error NEG count after a manager.Sync as well?

freehan: It is not counting the GCE NEG objects. 
It is counting the instances of NEG(s) wanted. Which is expressed by the intention on Service API. Then the intention has a 1:1 mapping to syncer, as each NEG syncer handles one NEG instance. (one per zone)

Each individual syncer has a separate Sync latency metrics which comes with success or failure. So the SLI will be a combination of Syncer Count and SyncLatency.

freehan (commented):
LGTM overall. 
/approve

pkg/neg/controller.go
swetharepakula: That is correct a valid syncer can create multiple NEGs (one per zone). However when we do feature metric counts, we do not count the NEG per zone. It is also 1:1 with what a user would request.

Since `manager.Sync` only kicks off the sync routines and doesn't check for error or success I don't think we would gain much insight other than the number of syncers. We could look into adding more metrics inside of the Syncer, but I think it will be more noisy and not as actionable as we can't correlate to the features requested.

prameshj: Sounds good, thanks.

PR #1427, opened 2021-05-05
Populate namespace when creating ingress

swetharepakula:
Ingress V1 requires namespace to be populated at creation.

@spencerhance 
/assign @freehan


freehan (approved):
/lgtm
/approve

PR #1437, opened 2021-05-11
Generate URL for GCE instances resource

swetharepakula:
- removes dependency on legacy-cloud-provider for generating the
 instance references/URLs

/cc @prameshj @spencerhance 
/assign @freehan

### DIFF
--- a/pkg/instances/instances.go
+++ b/pkg/instances/instances.go
@@ -19,6 +19,7 @@ package instances
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"google.golang.org/api/compute/v1"
@@ -42,8 +43,9 @@ const (
 type Instances struct {
 	cloud InstanceGroups
 	ZoneLister
-	namer    namer.BackendNamer
-	recorder record.EventRecorder
+	namer              namer.BackendNamer
+	recorder           record.EventRecorder
+	instanceLinkFormat string
 }
 
 type recorderSource interface {
@@ -53,11 +55,12 @@ type recorderSource interface {
 // NewNodePool creates a new node pool.
 // - cloud: implements InstanceGroups, used to sync Kubernetes nodes with
 //   members of the cloud InstanceGroup.
-func NewNodePool(cloud InstanceGroups, namer namer.BackendNamer, recorders recorderSource) NodePool {
+func NewNodePool(cloud InstanceGroups, namer namer.BackendNamer, recorders recorderSource, basePath string) NodePool {
 	return &Instances{
-		cloud:    cloud,
-		namer:    namer,
-		recorder: recorders.Recorder(""), // No namespace
+		cloud:              cloud,
+		namer:              namer,
+		recorder:           recorders.Recorder(""), // No namespace
+		instanceLinkFormat: basePath + "zones/%s/instances/%s",
 	}
 }
 
@@ -257,13 +260,22 @@ func (i *Instances) splitNodesByZone(names []string) map[string][]string {
 	return nodesByZone
 }
 
+// getInstanceReferences creates and returns the instance references by generating the
+// expected instance URLs
+func (i *Instances) getInstanceReferences(zone string, nodeNames []string) (refs []*compute.InstanceReference) {
+	for _, nodeName := range nodeNames {
+		refs = append(refs, &compute.InstanceReference{Instance: fmt.Sprintf(i.instanceLinkFormat, zone, canonicalizeInstanceName(nodeName))})
+	}
+	return refs
+}
+
 // Add adds the given instances to the appropriately zoned Instance Group.
 func (i *Instances) Add(groupName string, names []string) error {
 	events.GlobalEventf(i.recorder, core.EventTypeNormal, events.AddNodes, "Adding %s to InstanceGroup %q", events.TruncatedStringList(names), groupName)
 	var errs []error
 	for zone, nodeNames := range i.splitNodesByZone(names) {
 		klog.V(1).Infof("Adding nodes %v to %v in zone %v", nodeNames, groupName, zone)
-		if err := i.cloud.AddInstancesToInstanceGroup(groupName, zone, i.cloud.ToInstanceReferences(zone, nodeNames)); err != nil {
+		if err := i.cloud.AddInstancesToInstanceGroup(groupName, zone, i.getInstanceReferences(zone, nodeNames)); err != nil {
 			errs = append(errs, err)
 		}
 	}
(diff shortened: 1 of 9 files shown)


pkg/controller/utils.go
```diff
@@ -86,3 +88,16 @@ func nodePorts(svcPorts []utils.ServicePort) []int64 {
...
 	return ports
 }
+
+// getBasePath returns the compute API endpoint with the `projects/<project-id>` element
+func getBasePath(cloud *gce.Cloud) string {
+	basePath := cloud.ComputeServices().GA.BasePath
+	elements := strings.Split(basePath, "/")
+
+	// Last object in elements will be an empty string because basePath ends with a
+	// trailing /, so consider the element before as the last elment
```
prameshj, line 98: Can you include context that compute API 0.36 changed basePath to drop the "projects/" suffix?

pkg/controller/utils.go
```diff
@@ -86,3 +88,16 @@ func nodePorts(svcPorts []utils.ServicePort) []int64 {
 	}
 	return ports
 }
+
+// getBasePath returns the compute API endpoint with the `projects/<project-id>` element
+func getBasePath(cloud *gce.Cloud) string {
+	basePath := cloud.ComputeServices().GA.BasePath
+	elements := strings.Split(basePath, "/")
+
+	// Last object in elements will be an empty string because basePath ends with a
```
prameshj, line 97: Can we instead do `strings.Split(strings.TrimSuffix(basePath, "/"), "/")`?

pkg/instances/instances.go
```diff
@@ -53,11 +55,12 @@ type recorderSource interface {
 // NewNodePool creates a new node pool.
 // - cloud: implements InstanceGroups, used to sync Kubernetes nodes with
 //   members of the cloud InstanceGroup.
-func NewNodePool(cloud InstanceGroups, namer namer.BackendNamer, recorders recorderSource) NodePool {
+func NewNodePool(cloud InstanceGroups, namer namer.BackendNamer, recorders recorderSource, basePath string) NodePool {
 	return &Instances{
 		cloud:    cloud,
 		namer:    namer,
 		recorder: recorders.Recorder(""), // No namespace
+		basePath: basePath,
```
prameshj, line 63: how about we call this "instanceLinkPrefix" or something and includes the "zones/instances/" as well?

pkg/utils/utils.go
```diff
@@ -622,3 +623,16 @@ func NewStringPointer(s string) *string {
 func NewInt64Pointer(i int64) *int64 {
 	return &i
 }
+
+// GetBasePath returns the compute API endpoint with the `projects/<project-id>` element
+func GetBasePath(cloud *gce.Cloud) string {
+	basePath := cloud.ComputeServices().GA.BasePath
+	elements := strings.Split(basePath, "/")
+
+	// Last object in elements will be an empty string because basePath ends with a
```
prameshj, line 632: same comment about using Strings.TrimSuffix, as above.

pkg/controller/utils.go
```diff
@@ -86,3 +88,16 @@ func nodePorts(svcPorts []utils.ServicePort) []int64 {
 	}
 	return ports
 }
+
+// getBasePath returns the compute API endpoint with the `projects/<project-id>` element
+func getBasePath(cloud *gce.Cloud) string {
```
prameshj, line 93: I think this needs to be removed?

pkg/instances/instances.go
```diff
@@ -257,13 +260,23 @@ func (i *Instances) splitNodesByZone(names []string) map[string][]string {
 	return nodesByZone
 }
 
+// getInstanceReferences creates and returns the instance references by generating the
+// expected instance URLs
+func (i *Instances) getInstanceReferences(zone string, nodeNames []string) (refs []*compute.InstanceReference) {
+	for _, nodeName := range nodeNames {
+		instanceLink := i.basePath + strings.Join([]string{"zones", zone, "instances", nodeName}, "/")
```
prameshj, line 267: If Instances object has an instancePrefixLink, this function just needs to append the nodename.

prameshj (approved):

pkg/instances/instances.go
swetharepakula: Since zone and instance cannot be determined beforehand, I changed this to be the format string instead. Then when generating the URL we just add in the zone and node name

pkg/instances/instances.go
swetharepakula: See above comment, I switched to `instanceLinkFormat` and this now just adds the zone and nodeName. The function that calls this traverses on zone so couldn't add it to the link as a prefix.

pkg/instances/instances_test.go
```diff
@@ -29,7 +29,7 @@ const defaultZone = "default-zone"
 var defaultNamer = namer.NewNamer("uid1", "fw1")
 
 func newNodePool(f *FakeInstanceGroups, zone string) NodePool {
-	pool := NewNodePool(f, defaultNamer, &test.FakeRecorderSource{})
+	pool := NewNodePool(f, defaultNamer, &test.FakeRecorderSource{}, "https://www.googleapis.com/compute/v1/projects/")
```
prameshj, line 32: Out of curiosity - why are we using hardcoded string in this test?

swetharepakula: This actually could be anything, even an empty string. But that brings up a good point. I will add a test to check that the instance urls come out right.

swetharepakula: Added a test specifically for the instance link creation

pkg/utils/utils_test.go
```diff
@@ -995,3 +995,41 @@ func TestBackendToServicePortID(t *testing.T) {
 		})
 	}
 }
+
+func TestGetBasePath(t *testing.T) {
```
freehan, line 999: add a test case (probably a separate test case) to validate the exact expected path.

This is to prevent future base path changes got in without notice, or make it more obvious.

pkg/instances/instances.go
```diff
@@ -257,13 +259,22 @@ func (i *Instances) splitNodesByZone(names []string) map[string][]string {
 	return nodesByZone
 }
 
+// getInstanceReferences creates and returns the instance references by generating the
+// expected instance URLs
+func (i *Instances) getInstanceReferences(zone string, nodeNames []string) (refs []*compute.InstanceReference) {
```
freehan, line 264: if we are replacing `cloud.ToInstanceReferences`, then we need to copy all the underlying helper functions here.

freehan: I notice this function has less filtering and guardrails?

pkg/utils/utils_test.go
swetharepakula: That's a good idea, added a hardcoded one so we will detect and investigate if basepath changes in the future.

pkg/instances/instances.go
swetharepakula: I missed that aspect. I copied over that function and updated the tests accordingly.

freehan (approved):
/lgtm
/approve

Issue #1447, opened 2021-05-15
Backend service synchronization triggers multiple times for the same backend


gabihodoroaga:
The backend service synchronization triggers multiple times when multiple paths use the same backend service 

For example for this configuration: 

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
spec:
  rules:
  - host: "example.com"
    http:
      paths:
      - path: /alpha
        pathType: ImplementationSpecific
        backend:
          service:
            name: backend-service
            port:
              number: 80
      - path: /beta
        pathType: ImplementationSpecific
        backend:
          service:
            name: backend-service
            port:
              number: 80
      # more paths here
```

the backend synchronization (health checks, cdn, iap, ...) will run for each of the paths, and each sync triggers api call to the GCE, and I assume there is a quota and rate limiting because it takes lot of time just for one backend service.

gabihodoroaga:
The pull request was merged  #1448 into master.
I will close this issue.

PR #1448, opened 2021-05-15
Fix duplicate service ports sync

gabihodoroaga:
The backend service synchronization triggers multiple times when multiple paths use the same backend service 

For example for this configuration  

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
spec:
  rules:
  - host: "example.com"
    http:
      paths:
      - path: /alpha
        pathType: ImplementationSpecific
        backend:
          service:
            name: backend-service
            port:
              number: 80
      - path: /beta
        pathType: ImplementationSpecific
        backend:
          service:
            name: backend-service
            port:
              number: 80
      # more paths here
```

the backend synchronization (health checks, cdn, iap, ...) will run for each of the paths, and each sync triggers api call to the GCE, and I assume there is a quota and rate limiting because it takes lot of time just for one backend service.

Fix for #1447

### DIFF
--- a/pkg/utils/gceurlmap.go
+++ b/pkg/utils/gceurlmap.go
@@ -127,13 +127,19 @@ func (g *GCEURLMap) PutPathRulesForHost(hostname string, pathRules []PathRule) {
 
 // AllServicePorts return a list of all ServicePorts contained in the GCEURLMap.
 func (g *GCEURLMap) AllServicePorts() (svcPorts []ServicePort) {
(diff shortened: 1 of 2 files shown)


freehan (approved):
Thanks!

/lgtm
/approve

freehan:
cc @spencerhance for a second opinion just in case I missed anything

PR #1452, opened 2021-05-19
Do not process status only updates

swetharepakula:
* ignores the controller made updates to service attachments

/assign @freehan

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/apis/serviceattachment/v1alpha1"
 	sav1alpha1 "k8s.io/ingress-gce/pkg/apis/serviceattachment/v1alpha1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/context"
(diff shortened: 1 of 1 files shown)


swetharepakula:
/hold 
This PR may be better after conversion to the beta APIs

pkg/psc/controller.go
```diff
@@ -105,7 +106,14 @@ func NewController(ctx *context.ControllerContext) *Controller {
...
 		UpdateFunc: func(old, cur interface{}) {
-			controller.enqueueServiceAttachment(cur)
+			curSA := cur.(*v1alpha1.ServiceAttachment)
+			oldSA := old.(*v1alpha1.ServiceAttachment)
+
+			if shouldProcess(oldSA, curSA) {
+				controller.enqueueServiceAttachment(cur)
+				return
+			}
+			klog.V(4).Infof("Ignoring status update for ServiceAttachment")
```
bowei, line 116: [minor]: it's simpler to understand if code is structured:

if exception { return }
if exception { return }
regularProcessing()

pkg/psc/controller.go
```diff
@@ -564,6 +572,17 @@ func validateUpdate(existingSA, desiredSA *alpha.ServiceAttachment) error {
 	return nil
 }
 
+// shouldProcess checks if service attachment should be processed or not.
+// It will ignore status only updates
+func shouldProcess(old, cur *v1alpha1.ServiceAttachment) bool {
+	if !reflect.DeepEqual(old.Spec, cur.Spec) ||
```
bowei, line 578: Maybe:

```
specChanged := reflect.DeepEqual(old.Spec, cur.Spec)
metaChanged:= reflect.DeepEqual(old.ObjectMeta, cur.ObjectMeta)
// Ignore changes to TypeMeta and Status.
return specChanged || metaChanged

```

      metav1.TypeMeta   `json:",inline"`
      metav1.ObjectMeta `json:"metadata,omitempty"`

return !reflect.DeepEqual(old.Spec, cur.Spec) || ?

pkg/psc/controller.go
swetharepakula: I have rearranged the structure here for readability.

pkg/psc/controller.go
swetharepakula: We still want to process in situations where nothing has changed because we want to periodically update the ServiceAttachment Status. This is mostly to ignore changes that are only to the status because that change is from the controller as that triggers an never ending cycle of updates.

swetharepakula: I restructured the code to be easier to understand based on your suggestion.

PR #1456, opened 2021-05-20
Process service if old ingress class was glbc

swetharepakula:
* catches cleanup required when ingress class changes

/assign @freehan


freehan (approved):
/lgtm
/approve

Issue #1461, opened 2021-06-05
Firewall is not configured correctly when using named ports


gabihodoroaga:
The firewall rule that allows the traffic from the Load Balancer to the NEGs endpoints does not contain the endpoint port when a named service port is used.

For this configuration:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - name: http
          containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
  annotations:
    cloud.google.com/neg: '{"ingress": true}'
spec:
  type: NodePort
  ports:
  - port: 80
    targetPort: http
  selector:
    app: nginx
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx-ingress
spec:
  rules:
  - host: ""
    http:
      paths:
      - path: /*
        pathType: ImplementationSpecific
        backend:
          service:
            name: nginx-service
            port:
              number: 80
```

will generate a firewall rule which allows traffic only to the node port ranges

```text
allowed:
- IPProtocol: tcp
  ports:
  - 30000-32767
...
sourceRanges:
- 35.191.0.0/16
- 130.211.0.0/22
```

as a result the health checks are failing and the load balancing is not working either.

The second issue with the firewall rule is that when the list of the ports is empty, a firewall rule is generated with protocol tcp and an empty port list.

```text
allowed:
- IPProtocol: tcp
...
sourceRanges:
- 35.191.0.0/16
- 130.211.0.0/22
```

But this means unrestricted access to any tcp port. Not ok, but not extremely bad as allows traffic only from the LB only.

The situation can be reproduced by using and invalid service name for the default backend

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx-ingress
spec:
  defaultBackend:
    service:
      name: nonexisting-service
      port:
        number: 80
```

I will open a pull request for these issues.

gabihodoroaga:
I found in the documentation where it states that:

"If you specify a protocol without any port information, the firewall rule applies to that protocol and all its applicable ports."

https://cloud.google.com/vpc/docs/firewalls#protocols_and_ports

gabihodoroaga:
[...]

PR #1462, opened 2021-06-05
Fix L7 firewall rule sync errors

gabihodoroaga:
Fix issue #1461

Short description of what changed

- add a new field for the name of the service port to ```utils.ServicePort``` struct
- fixed ```GatherEndpointPorts```
- add/fixed unit test for ```GatherEndpointPorts```
- add check for empty port list in ```FirewallRules.Sync```
- add unit tests for empty port list


gabihodoroaga:
The issue with the named service port has been fixed by @alan-kut in this commit 302fdc0851318682effc35ff6102845fb4d05dd0. I removed all my code and kept the master version.

The second issue related to firewall rule with empty port still there.

When the list of the ports is empty (no valid service found), a firewall rule is generated with protocol tcp and an empty port list.

```txt
allowed:
- IPProtocol: tcp
...
sourceRanges:
- 35.191.0.0/16
- 130.211.0.0/22
```
This means UNRESTRICTED access to any port according to the documentation 
"If you specify a protocol without any port information, the firewall rule applies to that protocol and all its applicable ports."
https://cloud.google.com/vpc/docs/firewalls#protocols_and_ports

This can be easily reproduced by creating an empty ingress with non-existing default service.

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx-ingress
spec:
  defaultBackend:
    service:
      name: nonexisting-service
      port:
        number: 80
```

PR #1475, opened 2021-06-17
Only bump LastModifiedTimestamp for Status updates

swetharepakula:
* If status has not changed, modified timestamp is not updated and the
   CR will not be patched

/assign @freehan , @prameshj


freehan (approved):
/lgtm
/approve

PR #1479, opened 2021-06-18
Emit total number of services metrics

swetharepakula:
* ensure err is reused in processServiceAttachment for more accurate
   ServiceAttachment error state metrics

/assign @freehan @prameshj

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -206,7 +206,7 @@ func (c *Controller) enqueueServiceAttachment(obj interface{}) {
 func (c *Controller) addServiceToMetrics(obj interface{}) {
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
-		klog.Errorf("Failed to generate service: %q", err)
+		klog.Errorf("Failed to generate service key for obj %v: %v", obj, err)
 		return
 	}
 	c.collector.SetService(key)
@@ -216,7 +216,7 @@ func (c *Controller) addServiceToMetrics(obj interface{}) {
 func (c *Controller) deleteServiceFromMetrics(obj interface{}) {
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
-		klog.Errorf("Failed to generate service key: %q", err)
+		klog.Errorf("Failed to generate service key for obj %v: %v", obj, err)
 		return
 	}
 	c.collector.DeleteService(key)
@@ -243,7 +243,9 @@ func (c *Controller) processServiceAttachment(key string) error {
 		return err
 	}
 
-	obj, exists, err := c.svcAttachmentLister.GetByKey(key)
+	var obj interface{}
+	var exists bool
+	obj, exists, err = c.svcAttachmentLister.GetByKey(key)
 	if err != nil {
 		return fmt.Errorf("errored getting service from store: %q", err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/psc/controller.go
```diff
@@ -193,6 +202,26 @@ func (c *Controller) enqueueServiceAttachment(obj interface{}) {
 	c.svcAttachmentQueue.Add(key)
 }
 
+// addServiceToMetrics adds the metrics collector
+func (c *Controller) addServiceToMetrics(obj interface{}) {
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		klog.Errorf("Failed to generate service: %q", err)
```
prameshj, line 209: nit: "Failed to generate service key for obj - %v, err - %v"?

pkg/psc/controller.go
```diff
@@ -193,6 +202,26 @@ func (c *Controller) enqueueServiceAttachment(obj interface{}) {
...
+		return
+	}
+	c.collector.SetService(key)
+}
+
+// deleteServiceFromMetrics deletes the metrics collector
+func (c *Controller) deleteServiceFromMetrics(obj interface{}) {
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
prameshj, line 217: Same as above.

pkg/psc/controller.go
```diff
@@ -237,12 +268,14 @@ func (c *Controller) processServiceAttachment(key string) error {
 		return err
 	}
 
-	frURL, err := c.getForwardingRule(namespace, updatedCR.Spec.ResourceRef.Name)
+	var frURL string
+	frURL, err = c.getForwardingRule(namespace, updatedCR.Spec.ResourceRef.Name)
 	if err != nil {
 		return fmt.Errorf("failed to find forwarding rule: %q", err)
```
prameshj, line 274: You can use %w to wrap errors. Helps implement Unwrap methods to identify original error - https://blog.golang.org/go1.13-errors#TOC_3.3

pkg/psc/controller.go
```diff
@@ -114,6 +114,15 @@ func NewController(ctx *context.ControllerContext) *Controller {
 			controller.enqueueServiceAttachment(cur)
 		},
 	})
+
+	ctx.ServiceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(service interface{}) {
+			controller.addServiceToMetrics(service)
```
prameshj, line 120: Do we want to count all services or just the ILB ones?

swetharepakula: Currently, we want to count all the services. Since there are metrics that count the number of ILBs I think we don't need a specifically count just the ILB ones. Counting the services gives us an estimate of the ratio of service attachments to ilbs to services.

pkg/psc/controller.go
swetharepakula: Fixed, thanks!

pkg/psc/controller.go
swetharepakula: Fixed

pkg/psc/controller.go
swetharepakula: Oh yes! thanks for pointing that out. I will make a follow up pr to fix all of these in the psc controller.

pkg/psc/controller.go
prameshj: sounds good.

PR #1480, opened 2021-06-19
Use the new ServiceAttachment fields over deprecated ones

swetharepakula:
* ProducerForwardingRule -> TargetService
 * ConsumerForwardingRules -> ConnectedEndpoints
 * ConsumerForwardingRule.ForwardingRule -> ConnectedEndpoint.Endpoint

For reference: PSC API [changes](https://cloud.google.com/vpc/docs/configure-private-service-connect-producer#api-changes)

/assign @freehan @prameshj

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -250,13 +250,13 @@ func (c *Controller) processServiceAttachment(key string) error {
 	saName := c.saNamer.ServiceAttachment(namespace, name, string(updatedCR.UID))
 	desc := sautils.ServiceAttachmentDesc{URL: updatedCR.SelfLink}
 	gceSvcAttachment := &beta.ServiceAttachment{
-		ConnectionPreference:   svcAttachment.Spec.ConnectionPreference,
-		Name:                   saName,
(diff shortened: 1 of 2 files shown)


pkg/psc/controller.go
```diff
@@ -532,16 +532,16 @@ func validateUpdate(existingSA, desiredSA *beta.ServiceAttachment) error {
 		return fmt.Errorf("serviceAttachment connection preference cannot be updated from %s to %s", existingSA.ConnectionPreference, desiredSA.ConnectionPreference)
 	}
 
-	existingFR, err := cloud.ParseResourceURL(existingSA.ProducerForwardingRule)
+	existingFR, err := cloud.ParseResourceURL(existingSA.TargetService)
 	if err != nil {
 		return fmt.Errorf("serviceAttachment existing forwarding rule has malformed URL: %q", err)
```
prameshj, line 537: Should the error reflect the new field name "target service"

swetharepakula: good catch! thanks, fixed in the next patch.

PR #1492, opened 2021-06-23
Update Ingress Sync Queue to use multiple workers

spencerhance:
- Decreases ingress sync time
Tested by batch creating 30 ingresses that use the same service and 5 workers
Before: ~20 minutes for all ingresses to sync
After: ~2 minutes for all ingresses to sync
- Note: There is a mutex around SyncBackends(), so that portion of the code is still effectively single-threaded.

- Configurable with `--num-ingress-workers` flag

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -81,6 +81,7 @@ var (
 		NodePortRanges                   PortRanges
 		ResyncPeriod                     time.Duration
 		NumL4Workers                     int
+		NumIngressWorkers                int
 		RunIngressController             bool
 		RunL4Controller                  bool
 		Version                          bool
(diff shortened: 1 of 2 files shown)


pkg/flags/flags.go
```diff
@@ -197,6 +198,8 @@ the pod secrets for creating a Kubernetes client.`)
 		`Relist and confirm cloud resources this often.`)
 	flag.IntVar(&F.NumL4Workers, "num-l4-workers", 5,
 		`Number of parallel L4 Service worker goroutines.`)
+	flag.IntVar(&F.NumIngressWorkers, "num-ingress-workers", 5,
```
freehan, line 201: let us default to 1

freehan (commented):
can we add a mutex lock on the SyncBackend as well.

pkg/flags/flags.go
spencerhance: Done

spencerhance:
> can we add a mutex lock on the SyncBackend as well.

Done, added some more comments explaining this as well

PR #1493, opened 2021-06-23
Wrap errors using '%w' in the format strings

swetharepakula:
* log lines should use '%s' or '%q'

/assign @prameshj

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -206,7 +206,7 @@ func (c *Controller) enqueueServiceAttachment(obj interface{}) {
 func (c *Controller) addServiceToMetrics(obj interface{}) {
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/psc/controller.go
```diff
@@ -345,7 +345,7 @@ func (c *Controller) garbageCollectServiceAttachments() {
 		}
 		key, err := cache.MetaNamespaceKeyFunc(sa)
 		if err != nil {
-			klog.V(4).Infof("failed to generate key for service attachment: %s/%s: %w", sa.Namespace, sa.Name, err)
+			klog.V(4).Infof("failed to generate key for service attachment: %s/%s: %s", sa.Namespace, sa.Name, err)
```
prameshj, line 348: %q for consistency?

prameshj (commented):
lgtm except one nit.

pkg/psc/controller.go
swetharepakula: that sounds reasonable to me. thanks!

PR #1494, opened 2021-06-24
Filter errors when marking sync failures for metrics

swetharepakula:
The following errors will not be counted as failed syncs for PSC metrics:
  * GCE Resource Not Found
  * GCE Bad Gateway
  * Service Not Found in Store
  * FwdRule IP does not match Service

/assign @prameshj @freehan

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -17,6 +17,7 @@ package psc
 
 import (
 	context2 "context"
+	"errors"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -26,6 +27,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	beta "google.golang.org/api/compute/v0.beta"
+	"google.golang.org/api/googleapi"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/psc/controller_test.go
```diff
@@ -728,6 +729,70 @@ func TestShouldProcess(t *testing.T) {
...
+		err           error
+		expectedError error
+	}{
+		{
+			desc:          "google api status not found error",
+			err:           utils.FakeGoogleAPINotFoundErr(),
+			expectedError: nil,
+		},
+		{
+			desc:          "google api bad gateway error",
```
prameshj, line 745: Also add a case for google api with a different status code that you are not filtering - like 409(Conflict) or 400(Bad Request)

pkg/psc/controller.go
```diff
@@ -638,3 +648,21 @@ func shouldProcess(old, cur *sav1beta1.ServiceAttachment) bool {
...
+
+// filterError filters out errors that should not be considered sync errors for metrics
+func filterError(err error) error {
+	var apiError *googleapi.Error
+	if errors.As(err, &apiError) {
+		if utils.IsHTTPErrorCode(apiError, http.StatusNotFound) || utils.IsHTTPErrorCode(apiError, http.StatusBadGateway) {
+			return nil
+		}
+	}
+
```
prameshj, line 660: nit: newline can be skipped here and in line 666.

prameshj (commented):
lgtm except one testcase suggestion.

pkg/psc/controller_test.go
swetharepakula: Did some rework on the test and code. Changed the code to filter out BadRequests and now test both BadRequests and BadGateways where I expect BadGateways to return an error.

pkg/psc/controller.go
swetharepakula: Fixed

PR #1499, opened 2021-06-28
Add standard k8s resource description for ServiceAttachments

swetharepakula:
GCE ServiceAttachment description will have a link to the cluster and a link to the K8s Service Attachment CR

/assign @prameshj

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -72,6 +72,9 @@ var (
 		GCEOperationPollInterval         time.Duration
 		GCERateLimit                     RateLimitSpecs
 		GCERateLimitScale                float64
+		GKEClusterName                   string
+		GKEClusterHash                   string
+		GKEClusterType                   string
 		HealthCheckPath                  string
 		HealthzPort                      int
 		InCluster                        bool
@@ -230,6 +233,9 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableBackendConfigHealthCheck, "enable-backendconfig-healthcheck", false, "Enable configuration of HealthChecks from the BackendConfig")
 	flag.BoolVar(&F.EnablePSC, "enable-psc", false, "Enable PSC controller")
 	flag.BoolVar(&F.EnableIngressGAFields, "enable-ingress-ga-fields", false, "Enable using Ingress Class GA features")
+	flag.StringVar(&F.GKEClusterName, "gke-cluster-name", "", "The name of the GKE cluster this Ingress Controller will be interacting with")
+	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster hash of the GKE cluster this Ingress Controller will be interacting with")
+	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
 }
 
 type RateLimitSpecs struct {
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -41,6 +41,7 @@ import (
 	sav1beta1 "k8s.io/ingress-gce/pkg/apis/serviceattachment/v1beta1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/psc/metrics"
 	serviceattachmentclient "k8s.io/ingress-gce/pkg/serviceattachment/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/utils"
@@ -96,6 +97,16 @@ type Controller struct {
 	collector           metrics.PSCMetricsCollector
 
 	hasSynced func() bool
+
+	// The following are used to generate a cluster link for service attachment descs
+	// These values should only be used for providing information and not for
+	// any controller logic
+	// clusterName is the name of the cluster
+	clusterName string
+	// clusterLoc is the cluster zone or region
+	clusterLoc string
+	// regionalCluster indicates whether the cluster is regional or not.
+	regionalCluster bool
 }
 
 func NewController(ctx *context.ControllerContext) *Controller {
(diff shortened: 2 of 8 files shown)


pkg/flags/flags.go
```diff
@@ -230,6 +233,9 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableBackendConfigHealthCheck, "enable-backendconfig-healthcheck", false, "Enable configuration of HealthChecks from the BackendConfig")
 	flag.BoolVar(&F.EnablePSC, "enable-psc", false, "Enable PSC controller")
 	flag.BoolVar(&F.EnableIngressGAFields, "enable-ingress-ga-fields", false, "Enable using Ingress Class GA features")
+	flag.StringVar(&F.GKEClusterName, "gke-cluster-name", "", "The name of the GKE cluster this Ingress Controller will be interacting with")
+	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster has of the GKE cluster this Ingress Controller will be interacting with")
```
prameshj, line 237: nit: "The cluster hash"

pkg/flags/flags.go
```diff
@@ -230,6 +233,9 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
...
+	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "", "The type of cluster has of the GKE cluster this Ingress Controller will be interacting with")
```
prameshj, line 238: fix description.

pkg/psc/controller.go
```diff
@@ -96,6 +97,16 @@ type Controller struct {
 	collector           metrics.PSCMetricsCollector
 
 	hasSynced func() bool
+
+	// The following are used to generate a cluster link for service attachment descs
+	// These values  should only be used for providiing information and not for
```
prameshj, line 102: nit: "providing" typo

pkg/utils/serviceattachment/svcattachmentdesc.go
```diff
@@ -18,13 +18,33 @@ package serviceattachment
...
+	K8sCluster  string `json:"k8sCluster,omitempty"`
+}
+
+func NewServiceAttachmentDesc(namespace, crName, clusterName, location string, regional bool) ServiceAttachmentDesc {
+	locationKey := "regions"
+	if !regional {
+		locationKey = "zones"
+	}
+
+	cluster := fmt.Sprintf("/%s/%s/clusters/%s", locationKey, location, clusterName)
```
prameshj, line 38: Any existing utils we can use to generate cluster link from parameters? Just in case the format changes in future, it would be good to fix it in just one util function.

pkg/psc/controller_test.go
```diff
@@ -388,7 +393,7 @@ func TestServiceAttachmentConsumers(t *testing.T) {
 		{Endpoint: "consumer-fwd-rule-3", Status: "PENDING"},
 	}
 
-	desc := sautils.ServiceAttachmentDesc{URL: saCR.SelfLink}
+	desc := sautils.NewServiceAttachmentDesc(saCR.Namespace, saCR.Name, ClusterName, controller.cloud.Region(), true)
```
prameshj, line 396: The new util does not include saCR.SelfLink, shouldn't that still be set?

pkg/psc/controller.go
```diff
@@ -111,6 +122,17 @@ func NewController(ctx *context.ControllerContext) *Controller {
 		hasSynced:           ctx.HasSynced,
 		recorder:            ctx.Recorder,
 		collector:           ctx.ControllerMetrics,
+		clusterName:         flags.F.GKEClusterName,
+		regionalCluster:     ctx.RegionalCluster,
+	}
+	if controller.regionalCluster {
```
prameshj, line 128: I see controller.Cloud also has an "regional" boolean? Do you know if we can use that instead of accepting that as a commandline flag? https://github.com/kubernetes/ingress-gce/blob/bb88e6e55bfedb1bebc12ed4f68c0f338f634ea3/vendor/k8s.io/legacy-cloud-providers/gce/gce.go#L120

I see we are already relying on legacy-cloud-providers repo for the Region() and Zone() functions.

pkg/psc/controller_test.go
swetharepakula: Kubernetes has deprecated selflinks. In 1.21 I believe by default is no longer available and in 1.22 will be removed completely. This is one of the reason we are generating our own link instead of using the one on the CR.

pkg/utils/serviceattachment/svcattachmentdesc.go
swetharepakula: I was wondering if we should move it into the utils. However we moved the ServiceAttachment Desc into its own package to reduce the impact of importing it which is why I didn't create its own utils for this. I don't think there is an existing one yet. Perhaps adding a desc_utils package which has these functions?

pkg/psc/controller.go
swetharepakula: I originally did try to get the info from `controller.Cloud`, but unfortunately it isn't exposed and there is no method that returns it either. Even all the utils to convert the config bytes to a `Config` object are internal, so I think we have to stick to the commandline flag.

pkg/flags/flags.go
swetharepakula: Done.

pkg/flags/flags.go
swetharepakula: Done.

pkg/psc/controller.go
swetharepakula: Done.

pkg/utils/serviceattachment/svcattachmentdesc.go
swetharepakula: I created a new `descutils` package that has the link generation methods that can be used in the future for other resource descriptions.

swetharepakula:
Thanks @prameshj for the reviewing! I made an update that should address most of the comments. The only one I think that is not possible is pulling the regional information from the gce.Cloud object.

pkg/flags/flags.go
```diff
@@ -230,6 +233,9 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableBackendConfigHealthCheck, "enable-backendconfig-healthcheck", false, "Enable configuration of HealthChecks from the BackendConfig")
 	flag.BoolVar(&F.EnablePSC, "enable-psc", false, "Enable PSC controller")
 	flag.BoolVar(&F.EnableIngressGAFields, "enable-ingress-ga-fields", false, "Enable using Ingress Class GA features")
+	flag.StringVar(&F.GKEClusterName, "gke-cluster-name", "", "The name of the GKE cluster this Ingress Controller will be interacting with")
+	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster hash of the GKE cluster this Ingress Controller will be interacting with")
+	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
```
prameshj, line 238: set to ZONAL by default?

prameshj:
one minor comment, lgtm otherwise.

pkg/flags/flags.go
swetharepakula: That's a good idea. I updated with this change.

PR #1500, opened 2021-06-29
Emit success event after Service Attachment creation

swetharepakula:
/assign @prameshj

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -358,8 +358,14 @@ func (c *Controller) processServiceAttachment(key string) error {
 	}
 	klog.V(2).Infof("Created service attachment %s", saName)
 
-	_, err = c.updateServiceAttachmentStatus(updatedCR, gceSAKey)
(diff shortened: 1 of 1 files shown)


pkg/psc/controller.go
```diff
@@ -358,8 +358,14 @@ func (c *Controller) processServiceAttachment(key string) error {
 	}
 	klog.V(2).Infof("Created service attachment %s", saName)
 
-	_, err = c.updateServiceAttachmentStatus(updatedCR, gceSAKey)
+	updatedCR, err = c.updateServiceAttachmentStatus(updatedCR, gceSAKey)
 	klog.V(2).Infof("Updated Service Attachment %s/%s status", updatedCR.Namespace, updatedCR.Name)
+
+	if err != nil {
```
prameshj, line 364: err == nil

swetharepakula: good catch! thanks!

PR #1501, opened 2021-06-30
Fix API version for IG Linker

spencerhance:
This caused an extra sync loop if BackendService was not GA and was using alpha or beta only fields


bowei:
How was this tested?

/ok-to-test

bowei:
/hold 

for answer to the testing question

bowei:
/remove-lifecycle stale

@spencerhance ping

spencerhance:
@bowei this was tested by running e2e tests locally.  I modified the app-protocol test to use instance groups and tested to make sure that only one sync happened.

I also ran the full neg /instancegroup test to make sure it still passed.

spencerhance:
@bowei Can you remove the hold label if it lgtm?

Issue #1504, opened 2021-07-07
Drop import of k8s.io/kubernetes


howardjohn:
Currently we are importing this library. This requires a bunch of `replace` directives in go.mod, which is viral - any importers of this library *also* need to include them.

The current usage of the library is extremely small, it seems like it should be feasible to drop this import. This is also a general best practice, as far as I know, to not import k8s.io/kubernetes.

PR #1505, opened 2021-07-07
Drop k8s.io/kubernetes import

howardjohn:
Fixes https://github.com/kubernetes/ingress-gce/issues/1504


go.mod
```diff
@@ -18,39 +18,13 @@ require (
...
 	k8s.io/component-base v0.19.0
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
-	k8s.io/kubernetes v1.19.0
-	k8s.io/legacy-cloud-providers v0.0.0
+	k8s.io/legacy-cloud-providers v0.19.0
 	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
 )
-
-replace (
```
spencerhance, line 33: Curious why all the replace directives were removed?

howardjohn: They are only required because of k8s.io/kubernetes having invalid go.mod (because it isn't supposed to be imported). `require` statements are essentially viral - if your dependency has them then you need to include them as well, and soon enough the entire golang ecosystem has a mess of `require` statements.

howardjohn: Although I do have a slight concern about the `cloud.google.com/go` version changing. Someone (or the tests) with more knowledge here may want to validate that closer

pkg/apis/frontendconfig/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,7 +1,7 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright 2020 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
```
swetharepakula, line 4: I think most of these files will have a similar diff. Is there a reason for the codegen? I don't think we need to do this unless we are updating the APIs.

howardjohn: I just ran the hack/update-codegen script, which I assumed should be run on every commit?

pkg/utils/utils.go
```diff
@@ -429,13 +428,26 @@ func ListWithPredicate(nodeLister listers.NodeLister, predicate NodeConditionPre
...
+	ip, err := getPreferredNodeAddress(inputNode, []api_v1.NodeAddressType{api_v1.NodeInternalIP})
 	if err != nil {
 		klog.Errorf("Failed to get IP address for node %s", inputNode.Name)
 	}
 	return ip
 }
 
+// getPreferredNodeAddress returns the address of the provided node, using the provided preference order.
+// If none of the preferred address types are found, an error is returned.
+func getPreferredNodeAddress(node *api_v1.Node, preferredAddressTypes []api_v1.NodeAddressType) (string, error) {
```
swetharepakula, line 440: Add the corresponding tests for this function.

go.mod
swetharepakula: cc/ @freehan @prameshj

swetharepakula (commented):
Thanks for submitting this PR!  I think most of the PR looks good.
Two main suggestions: 
 1) Add the tests for the respective functions that we are copying over since we are not going to be maintaining those. 
 2) Do not include the codegen changes since there are no API changes.

pkg/apis/frontendconfig/v1beta1/zz_generated.deepcopy.go
swetharepakula: We only run it for API changes as most of the time changes are like this commit which have no functionality difference and are mostly updates to the Copyright and some of the generated code to match newest version of the generation tool. Since there is no API, can you remove this commit? Thanks!

pkg/utils/slice/slice.go
```diff
@@ -0,0 +1,36 @@
+package slice
```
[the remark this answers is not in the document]
freehan: need the package license header here as well

freehan (commented):
thanks!

howardjohn:
Thanks, updated to use existing function in one place and add header+test for other

pkg/l4/l4controller.go
```diff
@@ -21,12 +21,15 @@ import (
 	"reflect"
 	"sync"
```
swetharepakula, line 23: nit: remove the extra line before "time"

howardjohn: I would recommend making formatting (and verification of formatting) automated. I fixed it though

PR #1511, opened 2021-07-16
add lock to GC logic

freehan:
This is to help prevent multiple workers entering GC logic with conflicts as GC might walk thru shared resources.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -86,6 +86,10 @@ type LoadBalancerController struct {
 	// backendLock locks the SyncBackend function to avoid conflicts between
 	// multiple ingress workers.
 	backendLock sync.Mutex
+
+	// gcLock locks the GC logics to avoid conflicts between multiple ingress workers.
+	gcLock sync.Mutex
+
 	// linker implementations for backends
 	negLinker backends.Linker
 	igLinker  backends.Linker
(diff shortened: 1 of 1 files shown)


freehan:
/assign @swetharepakula @prameshj 

cc @spencerhance

pkg/controller/controller.go
```diff
@@ -587,8 +583,53 @@ func (lbc *LoadBalancerController) sync(key string) error {
...
+	// Garbage collection will occur regardless of an error occurring. If an error occurred,
+	// it could have been caused by quota issues; therefore, garbage collecting now may
+	// free up enough quota for the next sync to pass.
+	allIngresses := lbc.ctx.Ingresses().List()
+	frontendGCAlgorithm := frontendGCAlgorithm(ingExists, oldScope != nil, ing)
+	if gcErr := lbc.ingSyncer.GC(allIngresses, ing, frontendGCAlgorithm, newScope); gcErr != nil {
+		lbc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeWarning, events.GarbageCollection, "Error during garbage collection: %v", gcErr)
+		return fmt.Errorf("error during sync %v, error during GC %v", syncErr, gcErr)
+	}
+	return nil
```
swetharepakula, line 607: nit: instead of nil, return `syncErr` here then you can remove the error handling logic in sync().

or another way could be to move the error check into sync and make the function just return the error from gc without any wrapping.

swetharepakula (approved):
I think the PR looks good. Behavior (other than the lock) is the same before and after the refactor.

I only had one nit about the error handling done in postSyncGC. Currently we are forced to check the error in postSyncGC and again in sync. I am wondering if we can remove one of them.

bowei:
where is the test for this?

pkg/controller/controller.go
freehan: Done.

pkg/controller/controller.go
```diff
@@ -587,8 +583,53 @@ func (lbc *LoadBalancerController) sync(key string) error {
 		if err == nil && ingExists {
```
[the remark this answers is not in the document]
freehan: This preserve the existing behavior by throwing out the error to be.handled by the work queue.

freehan:
> where is the test for this?

There are many existing uint tests (including GC tests) runs the `sync` itself, which includes preSyncGC and postSyncGC

freehan:
/hold cancel 

Let us merge first and start soaking. If more test needed, can be added later.

PR #1516, opened 2021-07-27
Fix err msg

bowei:
Fix the error message for when HTTP and HTTPS are both disabled

### DIFF
--- a/pkg/loadbalancers/l7.go
+++ b/pkg/loadbalancers/l7.go
@@ -18,6 +18,7 @@ package loadbalancers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/loadbalancers/l7.go
```diff
@@ -40,8 +41,8 @@ import (
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
-const (
-	invalidConfigErrorMessage = "invalid ingress frontend configuration, please check your usage of the 'kubernetes.io/ingress.allow-http' annotation."
+var (
+	errAllProtocolsDisabled = errors.New("invalid configuration: both HTTP and HTTPS are disabled (kubernetes.io/ingress.allow-http is false and there no valid TLS configuration); your Ingress will not be able to serve any traffic")
```
prameshj, line 45: nit: add "is" - "there is no valid TLS..."

PR #1522, opened 2021-07-30
Implement flag-gated support for Endpoint Slices in the NEG controller.

alan-kut:


pkg/common/index/index_test.go
```diff
@@ -0,0 +1,78 @@
...
+	keys, e := EndpointSlicesByServiceFunc("Not an Endpoint Slice")
+	if len(keys) > 0 {
+		t.Errorf("Incorrect keys, got: %s, expected empty", keys)
+	}
+	if e != nil {
+		t.Errorf("Incorrect error, got: %v, expected nil", e)
+	}
+}
+
+func TestEndpointSlicesByServiceFuncReturnsEmptyForSlicesWithoutService(t *testing.T) {
```
swetharepakula, line 45: Combine this test with the one below and use the testcase pattern like: https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/utils_test.go#L36-L185

pkg/common/index/index.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
+)
+
+const (
+	EndpointSlicesByServiceIndex = "by-service-index"
+)
+
```
swetharepakula, line 28: Add comments to this function and below.

pkg/common/index/index.go
```diff
@@ -0,0 +1,43 @@
...
+package index
+
+import (
+	"fmt"
+
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
+)
+
+const (
+	EndpointSlicesByServiceIndex = "by-service-index"
```
swetharepakula, line 26: What is this constant used for?

pkg/common/index/index.go
```diff
@@ -0,0 +1,43 @@
+/*
```
swetharepakula, line 1: perhaps this would be better in utils package? Perhaps in utils/endpointslices or something there.

pkg/context/context.go
```diff
@@ -180,6 +181,14 @@ func NewControllerContext(
 		context.RegionalCluster = true
 	}
 
+	if config.EndpointSlicesEnabled {
```
swetharepakula, line 184: We should also conditionally create the endpoints informer on line 161.

pkg/flags/flags.go
```diff
@@ -236,6 +237,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.GKEClusterName, "gke-cluster-name", "", "The name of the GKE cluster this Ingress Controller will be interacting with")
 	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster hash of the GKE cluster this Ingress Controller will be interacting with")
 	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
+	flag.BoolVar(&F.EnableEndpointSlicesInNeg, "enable-endpoint-slices-in-neg", false, "Enable using Endpoint Slices API in NEG instead of Endpoints API")
```
swetharepakula, line 240: nit: maybe `EnableEndpointSlicesForNeg`

pkg/neg/controller.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	istioV1alpha3 "istio.io/api/networking/v1alpha3"
 	apiv1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
```
swetharepakula, line 26: nit: One recommendation is to import as `discovery`. This will make the update to the v1 easier. However feel free to ignore if you think having the v1beta1 is valuable.

pkg/neg/controller.go
```diff
@@ -133,6 +135,7 @@ func NewController(
 	enableNonGcpMode bool,
 	enableAsm bool,
 	asmServiceNEGSkipNamespaces []string,
+	enableEndpointSlices bool,
```
swetharepakula, line 138: nit: Instead of adding a bool, you can do a check based on whether the endpointSliceInformer is nil or not.

swetharepakula (commented):
I have not finished reviewing this PR yet. However to make the review process a little quicker I am adding my partial feedback.

One suggestion is to break this PR up into smaller commits if not different PRs, for example perhaps one just for introducing the indexer, one for adding the flag, one for adding logic to use the flag etc. This would make the reviews easier. Thanks!

pkg/common/index/index.go
alan-kut: Moved to the utils/endpointslices for now

pkg/common/index/index.go
alan-kut: This is the name of the index

pkg/context/context.go
alan-kut: The Endpoints Informer is used also in the pkg/controller/translator/translator.go

pkg/flags/flags.go
alan-kut: Done

pkg/neg/controller.go
alan-kut: Done

pkg/neg/controller.go
alan-kut: I'd rather check the boolean flag. If for some reason endpointSliceInformer would be initialized e.g. for different part of the ingress-gce then we still don't want to use it here unless the flag is true.

pkg/neg/controller.go
```diff
@@ -757,6 +774,9 @@ func (c *Controller) handleErr(err error, key interface{}) {
 }
 
 func (c *Controller) enqueueEndpoint(obj interface{}) {
+	e, ok := obj.(*apiv1.Endpoints)
+	if ok && e.Namespace != "kube-system" {
```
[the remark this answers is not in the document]
alan-kut: This should be removed. I had some additional logs when debugging and apparently I didn't remove all of it.

pkg/common/index/index_test.go
alan-kut: Done

pkg/common/index/index.go
alan-kut: Done

pkg/context/context.go
freehan: Let us add a TODO to remove translator.go dependency on EndpointsInformer

pkg/context/context.go
```diff
@@ -180,6 +181,14 @@ func NewControllerContext(
...
+		// Do not trigger periodic resync on EndpointSlice object.
+		// This aims improve NEG controller performance by avoiding unnecessary NEG sync that triggers for each NEG syncer.
+		// As periodic resync may temporary starve NEG API ratelimit quota.
+		context.EndpointSliceInformer = discoveryinformer.NewEndpointSliceInformer(kubeClient, config.Namespace, 0, utils.NewNamespaceIndexer())
+		context.EndpointSliceInformer.AddIndexers(cache.Indexers{endpointslices.EndpointSlicesByServiceIndex: endpointslices.EndpointSlicesByServiceFunc})
```
freehan, line 189: @robscott @wojtek-t 

Do you think this way is safe enough to consume EndpointSlices with custom Indexer? Can you think of any corner cases? The goal here is to easily find the corresponding EndpointSlices of a service from cache without traversing thru the whole cache and do label matching.

If this works, it would be a small piece of code to easily consume EndpointSlice. Might worth adding it into blogpost @robscott

pkg/flags/flags.go
```diff
@@ -237,8 +238,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.GKEClusterName, "gke-cluster-name", "", "The name of the GKE cluster this Ingress Controller will be interacting with")
 	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster hash of the GKE cluster this Ingress Controller will be interacting with")
 	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
-
 	flag.BoolVar(&F.EnableTrafficScaling, "enable-traffic-scaling", false, "Enable support for Service {max-rate-per-endpoint, capacity-scaler}")
+	flag.BoolVar(&F.EnableEndpointSlicesForNeg, "enable-endpoint-slices-for-neg", false, "Enable using Endpoint Slices API in NEG instead of Endpoints API")
```
freehan, line 242: Just EnableEndpointSlice might be good enough

pkg/neg/types/types.go
```diff
@@ -342,3 +344,62 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
 
 // EndpointPodMap is a map from network endpoint to a namespaced name of a pod
 type EndpointPodMap map[NetworkEndpoint]types.NamespacedName
+
+type EndpointsData struct {
```
freehan, line 348: add comment to say this is an abstraction of EndpointSlice and Endpoints object.

pkg/neg/syncers/utils.go
```diff
@@ -245,61 +245,53 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to retrieve associated zone of node %q: %w", *endpointAddress.NodeName, err)
+			}
+			if zoneNetworkEndpointMap[zone] == nil {
+				zoneNetworkEndpointMap[zone] = negtypes.NewNetworkEndpointSet()
+			}
 
-				if includeAllEndpoints || shouldPodBeInNeg(podLister, address.TargetRef.Namespace, address.TargetRef.Name) {
-					networkEndpoint := negtypes.NetworkEndpoint{IP: address.IP, Port: matchPort, Node: *address.NodeName}
+			if endpointAddress.Ready || shouldPodBeInNeg(podLister, endpointAddress.TargetRef.Namespace, endpointAddress.TargetRef.Name) {
```
freehan, line 276: add a TODO to consider relying on Terminating state instead of checking pod DeletionTimestamp

cc: @prameshj

pkg/neg/types/types.go
```diff
@@ -342,3 +344,62 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
...
+	result := make([]EndpointsData, 0, len(slices))
+	for _, slice := range slices {
+		ports := make([]PortData, 0)
+		addresses := make([]AddressData, 0)
+		for _, port := range slice.Ports {
+			ports = append(ports, PortData{Name: *port.Name, Port: *port.Port})
+		}
+		for _, ep := range slice.Endpoints {
+			// Ignore terminating endpoints. Nil means that endpoint is not terminating.
+			if ep.Conditions.Terminating != nil && *ep.Conditions.Terminating {
```
freehan, line 395: @prameshj 
I am wondering if NEG controller needs to distinguish Ready/NotReady/Terminating. 

The current implementation only distinguish `1. needs to be in NEG` and `2. does not need to be in NEG`(Pod has deletion timestamp).  

So in EndpointSlice, it translate to `1. Endpoint exists in EndpointSlice` `2. Endpoint does not exists or in terminiating state`. 

Is this good enough? Should we do more here to future proof?

freehan: So the current implementation is very aggressive in adding and removing endpoints. 
Attach gets triggered when pod started before Ready
Detach gets triggered immediately when pod got into terminating state.

pkg/utils/endpointslices/index.go
```diff
@@ -0,0 +1,54 @@
...
+	return fmt.Sprintf("%s/%s", namespace, serviceName)
+}
+
+/*
+Index function. If applied for an EndpointSlice with an associated service
+it returns properly formatted service name with a namespace as a key.
+Otherwise returns an empty keys array which means that the object will not
+be indexed.
+*/
+func EndpointSlicesByServiceFunc(obj interface{}) ([]string, error) {
```
freehan, line 44: index func for blog post
cc @robscott

pkg/context/context.go
wojtek-t: It was me who suggested using the indexer to Alan. So at the level of idea yes - I think it's exactly what we need.

[Conceptually, what the indexer does it is just kind-of decorating the Add/Update/Delete calls to additionally not just update the underlying store, but also the index [index is map from "value -> set of objects that match that value']].

So I don't see any corner-cases where it could cause any problems (it adds a bit of overhead but both cpu and memory it should be negligible compared to what the store itself is using.]


[Now - instead of calling AddIndexers, I suggest passing the indexer directly in line 188 - but that's just a minor comment.]

alan-kut: Done, added indexer directly in line 188

pkg/flags/flags.go
alan-kut: Done

pkg/neg/syncers/utils.go
alan-kut: Done

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
...
+		if !ok {
+			klog.Errorf("unexpected object type: %T", obj)
+			return
+		}
+		if endpointSlice, ok = tombstone.Obj.(*discovery.EndpointSlice); !ok {
+			klog.Errorf("unexpected object type: %T", obj)
+			return
+		}
+	}
+	serviceName, ok := endpointSlice.Labels[discovery.LabelServiceName]
```
freehan, line 798: missing checking ok?

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
...
+	// Check if serviceName contains a namespace
+	serviceNamespace, _, err := cache.SplitMetaNamespaceKey(serviceName)
```
freehan, line 800: do you need to check this format though?

if this is malfortmatted (not an object name), maybe just ignore this EndpointSlice?

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
...
+	if err != nil {
+		klog.Errorf("Failed to split service name: %v", err)
+		return
+	}
+	if len(serviceNamespace) > 0 || len(endpointSlice.GetNamespace()) == 0 {
+		c.endpointQueue.Add(serviceName)
+		return
+	}
+	// Prepend serviceName with the namespace of the EndpointSlice
+	c.endpointQueue.Add(endpointSlice.GetNamespace() + "/" + serviceName)
```
freehan, line 810: nit: use fmt.Sprintf("%s/%s", namespace, name)

pkg/neg/controller_test.go
```diff
@@ -1224,35 +1245,120 @@ func TestEnableNegCRD(t *testing.T) {
...
 }
 
-func getNodeEvent(controller *Controller, eventChan chan string) {
-	item, quit := controller.nodeQueue.Get()
+func TestEnqueueEndpoints(t *testing.T) {
+	namespace := "nmspc"
+	service := "svc"
+	key := namespace + "/" + service
+	t.Parallel()
+	testCases := []struct {
```
freehan, line 1253: might want to add some negative case. for instance, malformed service name label

pkg/neg/syncers/utils.go
```diff
@@ -245,61 +245,58 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
...
-					// Skip if the endpoint's pod not matching the subset lables.
-					if !shouldPodBeInDestinationRuleSubset(podLister, address.TargetRef.Namespace, address.TargetRef.Name, subsetLables) {
-						continue
-					}
-				}
-				if address.NodeName == nil {
-					klog.V(2).Infof("Endpoint %q in Endpoints %s/%s does not have an associated node. Skipping", address.IP, endpoints.Namespace, endpoints.Name)
+		for _, endpointAddress := range ed.Addresses {
+			// Apply the selector if Istio:DestinationRule subset labels provided.
+			if subsetLables != "" {
```
freehan, line 250: nit: subsetLabels

pkg/neg/types/types.go
freehan: I think the comment should be put on EndpointsData?

pkg/neg/types/types.go
```diff
@@ -342,3 +344,65 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
...
+// It contains all the information needed to set up
+// GCP Load Balancer.
+type AddressData struct {
+	TargetRef *apiv1.ObjectReference
+	NodeName  *string
+	Addresses []string
+	Ready     bool
+}
+
+func EndpointsDataFromEndpoints(ep *apiv1.Endpoints) []EndpointsData {
```
freehan, line 369: Add comment:

// EndpointsDataFromEndpoints converts all Addresses (Ready and NotReady) from the Endpoints object to the EndpointData list.

pkg/neg/types/types.go
```diff
@@ -342,3 +344,65 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
...
+		}
+		for _, addr := range subset.NotReadyAddresses {
+			addresses = append(addresses, AddressData{TargetRef: addr.TargetRef, NodeName: addr.NodeName, Addresses: []string{addr.IP}, Ready: false})
+		}
+		result = append(result, EndpointsData{Meta: &ep.ObjectMeta, Ports: ports, Addresses: addresses})
+	}
+	return result
+}
+
+func EndpointsDataFromEndpointSlices(slices []*discovery.EndpointSlice) []EndpointsData {
```
freehan, line 388: // EndpointsDataFromEndpoints converts all Addresses except for condition Terminating == True  from the Endpoints object to the EndpointData list.

pkg/neg/types/types.go
freehan: Need unit test

pkg/neg/types/types.go
freehan: Need unit test

pkg/neg/syncers/transaction.go
```diff
@@ -176,24 +198,44 @@ func (s *transactionSyncer) syncInternal() error {
...
-	if err != nil {
-		return err
-	}
+	var targetMap map[string]negtypes.NetworkEndpointSet
+	var endpointPodMap negtypes.EndpointPodMap
 
-	if !exists {
-		klog.Warningf("Endpoint %s/%s does not exist. Skipping NEG sync", s.Namespace, s.Name)
-		return nil
+	if s.enableEndpointSlices {
```
freehan, line 204: Need unit test for core transaction syncer function with EndpointSlice enabled. 

Might be able to reuse existing test and add one more dimension.

pkg/context/context.go
```diff
@@ -263,10 +273,17 @@ func (ctx *ControllerContext) HasSynced() bool {
 		ctx.BackendConfigInformer.HasSynced,
 		ctx.PodInformer.HasSynced,
 		ctx.NodeInformer.HasSynced,
-		ctx.EndpointInformer.HasSynced,
 		ctx.SvcNegInformer.HasSynced,
 	}
 
+	if ctx.EndpointInformer != nil {
```
swetharepakula, line 279: With this patch, is it possible for the endpoint informer to be nil?

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -81,16 +85,42 @@ func TestLocalGetEndpointSet(t *testing.T) {
...
+			endpointSets:        nil,
+			networkEndpointType: negtypes.VmIpEndpointType,
+		},
 	}
 	svcKey := fmt.Sprintf("%s/%s", testServiceName, testServiceNamespace)
 	ec := NewLocalL4ILBEndpointsCalculator(nodeLister, zoneGetter, svcKey)
 	for _, tc := range testCases {
-		retSet, _, err := ec.CalculateEndpoints(tc.endpoints, nil)
+		var endpointsData []negtypes.EndpointsData
+		if tc.useSlices {
```
swetharepakula, line 118: This is okay, but I think it would be more valuable to test the Endpoint - > EndpointData conversion and the EndpointSlice -> EndpointData conversion. Then this test just needs to test against EndpointData. I think there are similar comments on the conversion code: https://github.com/kubernetes/ingress-gce/pull/1522/commits/7019ba3683ea305fddf5211dfc83a0956baf2e90#r682796115

pkg/context/context.go
alan-kut: It's not, I reverted this part.

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
 	c.endpointQueue.Add(key)
 }
 
+func (c *Controller) enqueueEndpointSlice(obj interface{}) {
+	endpointSlice, ok := obj.(*discovery.EndpointSlice)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			klog.Errorf("unexpected object type: %T", obj)
```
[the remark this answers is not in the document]
alan-kut: Done

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
...
+			return
+		}
+		if endpointSlice, ok = tombstone.Obj.(*discovery.EndpointSlice); !ok {
+			klog.Errorf("unexpected object type: %T", obj)
```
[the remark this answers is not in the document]
alan-kut: Done

pkg/neg/controller.go
alan-kut: Yes, fixed.

pkg/neg/controller.go
alan-kut: I think it can be removed all together because I think the service name from the label never contains a namespace. I'll confirm this with Rob.

pkg/neg/controller.go
alan-kut: Moved this to the utils to make sure all the keys generated here has the same format.

pkg/neg/types/types.go
alan-kut: Yes, fixed

pkg/neg/types/types.go
alan-kut: Done.

pkg/neg/types/types.go
alan-kut: Done.

pkg/neg/syncers/utils.go
alan-kut: Done

pkg/neg/controller_test.go
alan-kut: Done.

pkg/neg/syncers/endpoints_calculator_test.go
alan-kut: Done

pkg/neg/syncers/transaction.go
alan-kut: Done

freehan:
Looks good overall.

freehan (approved):
/lgtm
/approve

Thanks!

pkg/neg/controller_test.go
```diff
@@ -323,7 +344,7 @@ func TestEnableNEGServiceWithIngress(t *testing.T) {
 func TestEnableNEGServiceWithL4ILB(t *testing.T) {
 	kubeClient := fake.NewSimpleClientset()
 	testContext := negtypes.NewTestContextWithKubeClient(kubeClient)
-	controller := newTestControllerWithParamsAndContext(kubeClient, testContext, true)
+	controller := newTestControllerWithParamsAndContext(kubeClient, testContext, true, false)
```
prameshj, line 347: Should these be multiple cases, once with epslices = true and false?

pkg/neg/controller_test.go
```diff
@@ -1224,35 +1245,134 @@ func TestEnableNegCRD(t *testing.T) {
...
+			if gotKey != wantedKey {
+				t.Errorf("Got key %q, want %q", gotKey, wantedKey)
 			} else {
-				t.Logf("Got expected nodeKey %s", nodeKey)
+				t.Logf("Got expected key %s", wantedKey)
 				return
 			}
 		case <-time.After(5 * time.Second):
-			t.Errorf("Timed out waiting for node enqueue %v", time.Now())
+			t.Errorf("Timed out waiting for enqueue %v", time.Now())
```
prameshj, line 1362: nit: "waiting for enqueue of key %s"

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -33,7 +33,7 @@ import (
 func TestLocalGetEndpointSet(t *testing.T) {
 	t.Parallel()
 	mode := negtypes.L4LocalMode
-	_, transactionSyncer := newL4ILBTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), mode)
+	_, transactionSyncer := newL4ILBTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), mode, false)
```
prameshj, line 36: Same comment as in another test - It would be good to have 2 cases - true and false, just for completeness.

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -104,7 +104,7 @@ func TestLocalGetEndpointSet(t *testing.T) {
 func TestClusterGetEndpointSet(t *testing.T) {
 	t.Parallel()
 	mode := negtypes.L4ClusterMode
-	_, transactionSyncer := newL4ILBTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), mode)
+	_, transactionSyncer := newL4ILBTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), mode, false)
```
prameshj, line 107: nit: add a comment that just the "false" case is enough since this test does not depend on endpoints at all.

pkg/neg/syncers/utils.go
```diff
@@ -245,61 +245,58 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
 		foundMatchingPort = true
 
 		// processAddressFunc adds the qualified endpoints from the input list into the endpointSet group by zone
```
prameshj, line 247: Comment is not applicable, can be removed.

pkg/neg/syncers/utils.go
```diff
@@ -245,61 +245,58 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
...
-				}
-				if zoneNetworkEndpointMap[zone] == nil {
-					zoneNetworkEndpointMap[zone] = negtypes.NewNetworkEndpointSet()
-				}
+			}
+			if endpointAddress.NodeName == nil {
+				klog.V(2).Infof("Endpoint %q in Endpoints %s/%s does not have an associated node. Skipping", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
+				continue
+			}
+			if endpointAddress.TargetRef == nil {
```
prameshj, line 264: Isn't this same check already done in line 251? Is this a dup?
I see this was the case in the previous version of this file too.

pkg/neg/syncers/utils.go
```diff
@@ -245,61 +245,58 @@ func toZoneNetworkEndpointMap(endpoints *apiv1.Endpoints, zoneGetter negtypes.Zo
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to retrieve associated zone of node %q: %w", *endpointAddress.NodeName, err)
+			}
+			if zoneNetworkEndpointMap[zone] == nil {
+				zoneNetworkEndpointMap[zone] = negtypes.NewNetworkEndpointSet()
+			}
 
-				if includeAllEndpoints || shouldPodBeInNeg(podLister, address.TargetRef.Namespace, address.TargetRef.Name) {
-					networkEndpoint := negtypes.NetworkEndpoint{IP: address.IP, Port: matchPort, Node: *address.NodeName}
+			// TODO: This check and EndpoinsData.Ready field may be deleted once Endpoints support is removed.
```
prameshj, line 276: nit: EndpointsData

pkg/neg/types/types.go
```diff
@@ -342,3 +344,70 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
...
 // EndpointPodMap is a map from network endpoint to a namespaced name of a pod
 type EndpointPodMap map[NetworkEndpoint]types.NamespacedName
+
+// Abstraction over Endpoints and EndpointSlices.
+// It contains all the information needed to set up
+// GCP Load Balancer.
+type EndpointsData struct {
+	Meta      *metav1.ObjectMeta
+	Ports     []PortData
+	Addresses []AddressData
```
prameshj, line 354: nit: add comment that this includes both ready and unready Endpoints, if derived from Endpoints object?

pkg/neg/types/types.go
```diff
@@ -342,3 +344,70 @@ func (key NegSyncerKey) GetAPIVersion() meta.Version {
...
+
+type AddressData struct {
+	TargetRef *apiv1.ObjectReference
+	NodeName  *string
+	Addresses []string
+	Ready     bool
+}
+
+// Converts API Endpoints to the EndpointsData abstraction.
+// All endpoints are converted, not ready addresses are converted with Reade=false.
```
prameshj, line 370: nit: Ready=false

pkg/utils/endpointslices/index.go
```diff
@@ -0,0 +1,44 @@
...
+	es, ok := obj.(*discovery.EndpointSlice)
+	if !ok {
+		return []string{}, nil
+	}
+	key, err := EndpointSlicesServiceKey(es)
+	if err != nil {
+		return []string{}, nil
```
prameshj, line 41: return the error?

prameshj (commented):
Sorry for the late review. Minor comments/nits only.

pkg/neg/controller_test.go
alan-kut: I don't think so. This case tests L4, it doesn't depend that much on EndpointSlices, especially in regard to the mocked controller.

pkg/neg/controller_test.go
alan-kut: Fixed in the next PR

pkg/neg/syncers/endpoints_calculator_test.go
alan-kut: I don't agree. With 5 boolean flags we would need to have 2^5 test cases for every test to have full completeness.
In this case the transactionSyncer is used only to get a NodeLister. NodeLister is the same regardless of the enableEndpointSlices flag.

pkg/neg/syncers/endpoints_calculator_test.go
alan-kut: Done

pkg/neg/syncers/utils.go
alan-kut: Done

pkg/neg/syncers/utils.go
alan-kut: You are right. Fixed.

pkg/neg/syncers/utils.go
alan-kut: Fixed

pkg/neg/types/types.go
alan-kut: Fixed

pkg/neg/types/types.go
alan-kut: Done

pkg/utils/endpointslices/index.go
alan-kut: Done

pkg/neg/controller.go
```diff
@@ -765,6 +782,34 @@ func (c *Controller) enqueueEndpoint(obj interface{}) {
...
+		}
+	}
+	serviceName, ok := endpointSlice.Labels[discovery.LabelServiceName]
+	// Check if serviceName contains a namespace
+	serviceNamespace, _, err := cache.SplitMetaNamespaceKey(serviceName)
+	if err != nil {
+		klog.Errorf("Failed to split service name: %v", err)
+		return
+	}
+	if len(serviceNamespace) > 0 || len(endpointSlice.GetNamespace()) == 0 {
```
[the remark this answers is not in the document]
alan-kut: Yes. It is handled correctly.

PR #1529, opened 2021-08-02
skip NEG API call for NEG linker if possible

freehan:
This avoids unnecessary NEG API calls if possible. That would speed up ingress processing/

### DIFF
--- a/pkg/backends/neg_linker_test.go
+++ b/pkg/backends/neg_linker_test.go
@@ -14,6 +14,8 @@ limitations under the License.
 package backends
 
 import (
+	"fmt"
+	"k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
 	"strings"
 	"testing"
 
@@ -34,6 +36,7 @@ import (
 
 func newTestNEGLinker(fakeNEG negtypes.NetworkEndpointGroupCloud, fakeGCE *gce.Cloud) *negLinker {
 	fakeBackendPool := NewPool(fakeGCE, defaultNamer)
+	ctx := negtypes.NewTestContext()
 
 	// Add standard hooks for mocking update calls. Each test can set a update different hook if it chooses to.
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaBackendServices.UpdateHook = mock.UpdateAlphaBackendServiceHook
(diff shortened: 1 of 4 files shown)


pkg/backends/neg_linker_test.go
```diff
@@ -42,7 +43,7 @@ func newTestNEGLinker(fakeNEG negtypes.NetworkEndpointGroupCloud, fakeGCE *gce.C
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaRegionBackendServices.UpdateHook = mock.UpdateAlphaRegionBackendServiceHook
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockBetaRegionBackendServices.UpdateHook = mock.UpdateBetaRegionBackendServiceHook
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockRegionBackendServices.UpdateHook = mock.UpdateRegionBackendServiceHook
-	return &negLinker{fakeBackendPool, fakeNEG, fakeGCE}
+	return &negLinker{fakeBackendPool, fakeNEG, fakeGCE, ctx.SvcNegInformer.GetIndexer()}
 }
 
 func TestLinkBackendServiceToNEG(t *testing.T) {
```
swetharepakula, line 49: Can you add a test case that uses the svcneg informer?

swetharepakula:
/approve

Overall lgtm. I think we should add one more test case for when the informer is populated.

pkg/backends/neg_linker_test.go
freehan: added test cases with svcneg Informer populated

pkg/backends/neg_linker_test.go
```diff
@@ -233,7 +263,9 @@ func TestBackendsForNEG(t *testing.T) {
 					SelfLink:            "/neg1",
 				},
 			},
-			sp: &utils.ServicePort{},
+			sp: &utils.ServicePort{
+				VMIPNEGEnabled: true,
```
prameshj, line 267: The L4 controller will set this to true when invoking negLinker.Link(), so this change is fine - https://github.com/kubernetes/ingress-gce/blob/80ac3c02662cde257307d1f48d6966c91127c6b8/pkg/l4/l4controller.go#L288

PR #1530, opened 2021-08-02
Update Backends if the settings change w/ same Backends

bowei:
This handles the case where the user updates their Service definition.


freehan (approved):
/lgtm
/approve

PR #1536, opened 2021-08-04
Check for v1 IngressClass resource

swetharepakula:
Check for v1 IngressClass resource instead of v1beta1.

### DIFF
--- a/cmd/glbc/app/init.go
+++ b/cmd/glbc/app/init.go
@@ -66,7 +66,7 @@ func IngressClassEnabled(client kubernetes.Interface) bool {
 	klog.V(2).Info("Checking if Ingress Class API exists")
 
 	err := wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
-		resourceList, err := client.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1beta1")
+		resourceList, err := client.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1")
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/glbc/app/init.go
```diff
@@ -66,7 +66,7 @@ func IngressClassEnabled(client kubernetes.Interface) bool {
 	klog.V(2).Info("Checking if Ingress Class API exists")
 
 	err := wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
-		resourceList, err := client.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1beta1")
+		resourceList, err := client.Discovery().ServerResourcesForGroupVersion("networking.k8s.io/v1")
```
liggitt, line 69: shouldn't there be a test that would have caught this?

swetharepakula: We depend on our e2e tests for validating this logic

liggitt: but... they didn't catch the issue before, so is an update to those needed?

swetharepakula: This logic is flag gated, and the flag is not enabled in the OSS tests which is why this was not caught.

liggitt: ok, thanks, will look for tests in the environments where the flag is enabled

freehan (commented):
/lgtm
/approve

swetharepakula:
/hold cancel
Canceling hold with the additional review.

PR #1538, opened 2021-08-04
Always enable pathtype

swetharepakula:
This PR will separate the path type from the ingress class. That way path type can be enabled as ingress class continues to be developed.

/assign @freehan

### DIFF
--- a/pkg/controller/translator/translator.go
+++ b/pkg/controller/translator/translator.go
@@ -19,8 +19,6 @@ import (
 	"strconv"
 	"strings"
 
-	"k8s.io/ingress-gce/pkg/flags"
-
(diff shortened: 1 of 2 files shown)


swetharepakula:
I am investigating the failures. Currently I cannot reproduce locally.

bowei:
Swetha -- do this mean that when the user uses one of the other path types, we now support it?

swetharepakula:
Bowei - Yes this means if a user tries `Exact` or `Prefix` our controllers will support this.

Test failure was because the branch needed to be rebased.

freehan (commented):
/lgtm
/approve

PR #1547, opened 2021-08-12
Add e2e test for v1beta1 BackendConfig features

skmatti:


cmd/e2e-test/security_policy_v1beta1_test.go
```diff
@@ -0,0 +1,122 @@
...
+	ctx := context.Background()
+	t.Parallel()
+
+	Framework.RunWithSandbox("Security Policy Enable (backendconfig v1beta1)", t, func(t *testing.T, s *e2e.Sandbox) {
+		policies := []*computebeta.SecurityPolicy{
+			buildPolicyAllowAll(fmt.Sprintf("enable-test-allow-all-%s", s.Namespace)),
+		}
+		defer func() {
+			if err := cleanupSecurityPolicies(ctx, t, Framework.Cloud, policies); err != nil {
+				t.Errorf("cleanupSecurityPolicies(...) =  %v, want nil", err)
```
bowei, line 46: t.Logf

We should not Errorf in the defer()

bowei: Is it not possible to write a generic test with all of the fields of BackendConfig so we don't need separate tests for v1beta1 on those fields?

skmatti: Done

PR #1551, opened 2021-08-14
Add e2e test to validate consistency between v1 and v1beta1 Backendconfig APIs

skmatti:


cmd/e2e-test/backend_config_test.go
```diff
@@ -148,3 +151,72 @@ func TestBackendConfigNegatives(t *testing.T) {
...
+		t.Logf("BackendConfig %s created using V1 API", bcKey)
+		gotV1beta1BC, err := Framework.BackendConfigClient.CloudV1beta1().BackendConfigs(backendConfig.Namespace).
+			Get(context.Background(), backendConfig.Name, metav1.GetOptions{})
+		if err != nil {
+			t.Fatalf("Error getting v1beta1 backendconfig %s: %v", bcKey, err)
+		}
+
+		if diff := cmp.Diff(v1beta1BackendConfig.Spec, gotV1beta1BC.Spec); diff != "" {
+			t.Fatalf("Unexpected v1beta1 backendconfig spec (-want +got):\n%s", diff)
+		}
```
bowei, line 205: we should also Diff against a hand written expected output, just in case there is a bug in both v1 and v1beta1

skmatti: v1beta1BackendConfig is the defined struct (not from the server). Is that what mean by hand written?. 

It think this test will capture the case where both v1 and v1beta1 have a bug

freehan (commented):
/lgtm
/approve

PR #1556, opened 2021-08-17
Small fixes for PR 1522

alan-kut:


pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -35,6 +35,7 @@ import (
 func TestLocalGetEndpointSet(t *testing.T) {
 	t.Parallel()
 	mode := negtypes.L4LocalMode
+	// The "enableEndpointSlices=false" case is enough since this test does not depend on endpoints at all.
```
prameshj, line 38: This does depend on endpoints to pick up the Subset.

pkg/utils/endpointslices/index_test.go
```diff
@@ -60,12 +61,15 @@ func TestEndpointSlicesByServiceFunc(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
 			keys, e := EndpointSlicesByServiceFunc(tc.obj)
-			if len(keys) != len(tc.expectedKeys) || (len(tc.expectedKeys) == 1 && keys[0] != tc.expectedKeys[0]) {
-				t.Errorf("Incorrect keys, got: %s, expected %s", keys, tc.expectedKeys)
+			if tc.expectError && e == nil {
```
prameshj, line 64: nit: optional, but you could do "gotError := e != nil" and check if tc.expectError and gotError are equal.

prameshj (commented):
/ok-to-test

looks good, except for one comment about Local traffic policy subsets test.

freehan:
So this PR from OSS community member was trying to fix the bug. https://github.com/kubernetes/ingress-gce/pull/1462

This one adds a field ServicePortName field in ServicePort struct, which is missing for the translation.

pkg/neg/syncers/endpoints_calculator_test.go
alan-kut: Done

pkg/utils/endpointslices/index_test.go
alan-kut: I don't see any clear way how to do this. Also the error message may change. I'd leave it as it is.

alan-kut:
I fixed the comment about Local traffic policy subsets test.

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -35,6 +35,7 @@ import (
 func TestLocalGetEndpointSet(t *testing.T) {
 	t.Parallel()
 	mode := negtypes.L4LocalMode
+	// The "enableEndpointSlices=false" case is enough since this test does not depend on endpoints to pick up the Subsets.
```
prameshj, line 38: This test does depend on endpoints to pick up the subsets. Subsets will only include nodes that have endpoints running and are valid LB candidates. We should test enableEndpointSlices=true here.

Other than this comment, the changes LGTM.

pkg/neg/syncers/utils.go
```diff
@@ -261,10 +260,6 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
 				klog.V(2).Infof("Endpoint %q in Endpoints %s/%s does not have an associated node. Skipping", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
 				continue
 			}
-			if endpointAddress.TargetRef == nil {
```
prameshj, line 264: Actually, I think we need this.. the previous check for TargerRef is inside an `if subsetLabels != ""`

pkg/neg/syncers/endpoints_calculator_test.go
alan-kut: Done

pkg/neg/syncers/utils.go
alan-kut: You are right, fixed.

alan-kut: Shouldn't it contain check: endpointAddress.TargetRef.Kind != "Pod" as well?

PR #1557, opened 2021-08-17
Support EndpointSlices in the controller/translator.

alan-kut:


pkg/controller/translator/translator.go
```diff
@@ -549,7 +556,33 @@ func listAll(store cache.Store, selector labels.Selector, appendFn cache.AppendF
 	return nil
 }
 
-func listEndpointTargetPorts(indexer cache.Indexer, namespace, name, targetPort string) []int {
+func listEndpointTargetPortsFromEndpointSlices(indexer cache.Indexer, namespace, name, targetPort string) []int {
+	if i, err := strconv.Atoi(targetPort); err == nil {
```
freehan, line 560: Add comment // if targetPort is integer, no need to translate to endpoint ports

pkg/controller/translator/translator.go
```diff
@@ -549,7 +556,33 @@ func listAll(store cache.Store, selector labels.Selector, appendFn cache.AppendF
 	return nil
 }
 
-func listEndpointTargetPorts(indexer cache.Indexer, namespace, name, targetPort string) []int {
+func listEndpointTargetPortsFromEndpointSlices(indexer cache.Indexer, namespace, name, targetPort string) []int {
```
freehan, line 559: add comment for both `listEndpointTargetPortsFromEndpointSlices` and `listEndpointTargetPortsFromEndpoints`:

return targetport if port number is specified. 

find the actual target port behind the named target port

pkg/controller/translator/translator.go
```diff
@@ -549,7 +556,33 @@ func listAll(store cache.Store, selector labels.Selector, appendFn cache.AppendF
...
+	if err != nil {
+		klog.Errorf("Failed to retrieve endpoint slices for service %s/%s: %v", namespace, name, err)
+		return []int{}
+	}
+
+	ret := []int{}
+	for _, sliceObj := range slices {
+		slice := sliceObj.(*discoveryapi.EndpointSlice)
+		for _, port := range slice.Ports {
+			if port.Protocol != nil && *port.Protocol == api_v1.ProtocolTCP && port.Name != nil && *port.Name == targetPort && port.Port != nil {
```
freehan, line 577: Looking at this piece of code. I realize the previous code has a bug in it. 

port.Name is actually the service port name. targetPort is the targetPort Name. They can be different. 

The correct implementation should be passing in service port name and look it up in the endpoint when targetport is named port.

freehan: Might worth fixing this bug.  


See a counter example for the current implementation. 
https://github.com/freehan/kubescript/blob/master/hack/namdeport/namedport.yaml

freehan:
So this PR from OSS community member was trying to fix the bug. #1462

This one adds a field ServicePortName field in ServicePort struct, which is missing for the translation.

pkg/controller/translator/translator.go
alan-kut: I'd leave fixing a bug into separate PR.

pkg/controller/translator/translator.go
alan-kut: Done

pkg/controller/translator/translator.go
alan-kut: Done

pkg/controller/translator/translator.go
alan-kut: The reasoning is that - fixing a bug and support Endpoint Slices are entirely different things. Having them separate would be more clear and safer.

freehan (approved):
/lgtm
/approve

PR #1559, opened 2021-08-24
add backoff retry in readiness poller

freehan:


pkg/neg/readiness/poller.go
```diff
@@ -142,6 +148,10 @@ func (p *poller) Poll(key negMeta) (retry bool, err error) {
 	// TODO(freehan): filter the NEs that are in interest once the API supports it
 	res, err := p.negCloud.ListNetworkEndpoints(key.Name, key.Zone /*showHealthStatus*/, true, key.SyncerKey.GetAPIVersion())
 	if err != nil {
+		// On receiving GCE API error, do not retry immediately. This is to prevent the reflector to overwhelm the GCE NEG API when
+		// rate limiting is in effect. This will allow readiness reflector to overwhelm the GCE NEG API and cause syncers to backoff.
```
prameshj, line 152: `This will allow readiness reflector to overwhelm the GCE NEG API and cause syncers to backoff.` - Should it be "This will prevent"?

pkg/neg/readiness/poller.go
```diff
@@ -142,6 +148,10 @@ func (p *poller) Poll(key negMeta) (retry bool, err error) {
 	// TODO(freehan): filter the NEs that are in interest once the API supports it
 	res, err := p.negCloud.ListNetworkEndpoints(key.Name, key.Zone /*showHealthStatus*/, true, key.SyncerKey.GetAPIVersion())
 	if err != nil {
```
prameshj, line 150: Should we check for a specific error type?

pkg/neg/readiness/poller.go
```diff
@@ -33,6 +34,11 @@ import (
 
 const (
 	healthyState = "HEALTHY"
+
+	// retryDelay is the delay to retry health status polling.
+	// GCE NEG API RPS quota is rate limited per every 10 seconds.
```
prameshj, line 39: nit: Is there an external doc we can reference for the RPS? Or maybe add the default per-project quota here?

pkg/neg/readiness/poller.go
```diff
@@ -142,6 +148,10 @@ func (p *poller) Poll(key negMeta) (retry bool, err error) {
 	// TODO(freehan): filter the NEs that are in interest once the API supports it
 	res, err := p.negCloud.ListNetworkEndpoints(key.Name, key.Zone /*showHealthStatus*/, true, key.SyncerKey.GetAPIVersion())
```
prameshj, line 149: Can we add a unit test by mocking List API to return an error?

pkg/neg/readiness/poller.go
freehan: I am thinking we should not. Because any error would cause an immediate retry

pkg/neg/readiness/poller.go
freehan: Yes. I will add it.

pkg/neg/readiness/poller.go
freehan: okay.

PR #1562, opened 2021-09-01
Add errors to healthcheck merge() func

spencerhance:
- HealthCheck.merge() and ToAlphaComputeHealthCheck() now optionally return errors
- merge() now returns an error if the specified protocol is invalid

Example error:

```
Warning  Sync    1s (x3 over 3s)     loadbalancer-controller  Error syncing to GCP: error running backend syncing routine: error ensuring health check: Protocol "http" is not valid, must be one of ["HTTP","HTTPS","HTTP2"]
```

Old error:

```
Health check "k8s1-xxx-yyy" update error: googleapi: Error 400: Invalid value for field 'resource.httpHealthCheck': ''. HTTP healthCheck missing., invalid
```

/assign @freehan

### DIFF
--- a/pkg/translator/healthchecks.go
+++ b/pkg/translator/healthchecks.go
@@ -127,24 +127,30 @@ func (hc *HealthCheck) Protocol() annotations.AppProtocol {
 
 // ToComputeHealthCheck returns a valid compute.HealthCheck object
 func (hc *HealthCheck) ToComputeHealthCheck() (*compute.HealthCheck, error) {
-	hc.merge()
+	if err := hc.merge(); err != nil {
+		return nil, err
+	}
 	return utils.ToV1HealthCheck(&hc.HealthCheck)
 }
 
 // ToBetaComputeHealthCheck returns a valid computebeta.HealthCheck object
 func (hc *HealthCheck) ToBetaComputeHealthCheck() (*computebeta.HealthCheck, error) {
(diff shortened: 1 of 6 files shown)


pkg/translator/healthchecks.go
```diff
@@ -167,7 +173,12 @@ func (hc *HealthCheck) merge() {
 	case annotations.ProtocolHTTP2:
 		http2 := computealpha.HTTP2HealthCheck(hc.HTTPHealthCheck)
 		hc.HealthCheck.Http2HealthCheck = &http2
+	default:
```
freehan, line 176: In this case, `merge` was doing a noop. I am wondering if there is any corner case relies on this.

freehan: In the case of noop, is the hc.HealthCheck still valid?

freehan (commented):
I think this needs a second pair of eyes.

freehan (approved):
LGTM from my side. Pending on a second pair of eyes over this.
/approve

pkg/translator/healthchecks.go
spencerhance: re offline convo: in current behavior sync loop still breaks but with a really opaque error

swetharepakula:
Great improvement over the existing errors!
I think change looks good
/lgtm

PR #1577, opened 2021-09-26
Updates E2E test images registry

claudiubelu:
We're moving away from google.com gcp projects. These images are now on community-owned infra.

Note that the porter image has been centralized into the agnhost image.

Related: https://github.com/kubernetes/k8s.io/issues/1522


prameshj:
Looks good to me. @freehan @swetharepakula @spencerhance any issues with making this switch?

freehan:
If switching causes problem, it would very visible in our test grid. Merging it and see


/lgtm
/approve

PR #1588, opened 2021-10-26
Add Rbs mode flag

kl52752:
Rbs mode flag will enalbe L4 External LoadBalancer Controller
RbsMode has 4 values and its behaviour for L4NetLbController
will be implemented later

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	flag "github.com/spf13/pflag"
-
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
@@ -87,7 +86,7 @@ var (
 		NumIngressWorkers                int
 		RunIngressController             bool
 		RunL4Controller                  bool
-		RunL4NetLBController             bool
+		L4elbRbsMode                     RbsMode
 		Version                          bool
 		WatchNamespace                   string
 		LeaderElection                   LeaderElectionConfiguration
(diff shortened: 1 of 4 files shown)


pkg/flags/flags.go
```diff
@@ -233,8 +232,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableV2FrontendNamer, "enable-v2-frontend-namer", false, "Enable v2 ingress frontend naming policy.")
 	flag.BoolVar(&F.RunIngressController, "run-ingress-controller", true, `Optional, whether or not to run IngressController as part of glbc. If set to false, ingress resources will not be processed. Only the L4 Service controller will be run, if that flag is set to true.`)
 	flag.BoolVar(&F.RunL4Controller, "run-l4-controller", false, `Optional, whether or not to run L4 Service Controller as part of glbc. If set to true, services of Type:LoadBalancer with Internal annotation will be processed by this controller.`)
-	// TODO(kl52752) run-l4netlb-controller is only for testing purpose it will be changed to —-l4elb-rbs-mode
-	flag.BoolVar(&F.RunL4NetLBController, "run-l4netlb-controller", false, `Optional, whether or not to run L4ELB Service Controller as part of glbc. If set to true, services of Type:LoadBalancer without Internal annotation will be processed by this controller.`)
+	flag.Var(&F.L4elbRbsMode, "l4elb-rbs-mode", `If set L4NetLbController will be run. Possible Values: opt-in, enabled, enforced, rollback`)
```
prameshj, line 235: Why do we need this info in the controller? This controller only needs to know if the rbs mode is enabled or not - i.e should the elb specific controller be spun up or not. How will this extra info(enabled vs enforced vs opt-in) be used?

freehan: +1 Please put details about each mode in description.

kl52752: done
I added comment to cover all possible values for rbs mode.
The idea of this flag is to distinct which services will be handled by L4NetLB controller and it will be done via opt-in annotation or finalizer. There is also rollback value  to migrate all RBS based LB back to target pool based (legacy)

pkg/flags/flags.go
```diff
@@ -233,8 +232,12 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
...
 	flag.BoolVar(&F.RunL4Controller, "run-l4-controller", false, `Optional, whether or not to run L4 Service Controller as part of glbc. If set to true, services of Type:LoadBalancer with Internal annotation will be processed by this controller.`)
-	// TODO(kl52752) run-l4netlb-controller is only for testing purpose it will be changed to —-l4elb-rbs-mode
-	flag.BoolVar(&F.RunL4NetLBController, "run-l4netlb-controller", false, `Optional, whether or not to run L4ELB Service Controller as part of glbc. If set to true, services of Type:LoadBalancer without Internal annotation will be processed by this controller.`)
+	flag.Var(&F.L4elbRbsMode, "l4elb-rbs-mode", `If set L4NetLbController will be run.
+	Possible Values:
+	opt-in: manages NetLB services that contains opt-in annotation or NetLB finalizer,
+	enabled: manages NetLB services that contains NetLB finalizer,
+	enforced: manages all NetLB services,
+	rollback: rollback changes to target pool based loadbalancer`)
```
freehan, line 240: I am not sure if rollback would ever be useful. I am saying this because your cluster basically would have 3 states.

1. only targetPool
2. mix targetPool and RBS
3. only RBS

From state 3, you actually could not roll back to 2. In your description, it sounds like rollback directly from 3 to 1. Is  it the intention? I see this would create more headache.

cezarygerard: The rollback was explicitly requested by @wojtek-t and @ahmedtd 

In this mode the controller will iterate over services, remove the finalizer and clan up the GCP resources, so k/k service controller can pick up the service and create the target pool.

freehan: The rollback sounds like a failsafe. But this fail safe is actually disruptive as there is no seamless migration from InstanceGroup to Targetpool.

I would say rollback is dangerous as it it converts all services back into TargetPool mode. However, there might be only one service having problem. Then in this case, it might be better to just roll back the problematic services and basically rollback from state 3 to state 2. This means you do not have a rollback mode. You just change back to enabled mode and fix the problematic service. Does it make sense?

kl52752: done, rollback mode was remoced

PR #1596, opened 2021-10-28
Implement delete function for L4NetLB

kl52752:
Add L4NetLB finalizer
Handle L4 External LoadBalancer service deletion
Add unit tests for L4NetLBController

### DIFF
--- a/pkg/l4netlb/l4netlbcontroller.go
+++ b/pkg/l4netlb/l4netlbcontroller.go
@@ -93,7 +93,7 @@ func NewL4NetLBController(
 		AddFunc: func(obj interface{}) {
 			addSvc := obj.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(addSvc.Namespace, addSvc.Name)
-			if shouldProcess(addSvc) {
+			if l4netLBc.shouldProcessService(addSvc, nil) {
 				klog.V(3).Infof("L4 External LoadBalancer Service %s added, enqueuing", svcKey)
 				l4netLBc.ctx.Recorder(addSvc.Namespace).Eventf(addSvc, v1.EventTypeNormal, "ADD", svcKey)
 				l4netLBc.svcQueue.Enqueue(addSvc)
@@ -108,7 +108,7 @@ func NewL4NetLBController(
 			curSvc := cur.(*v1.Service)
 			oldSvc := old.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(curSvc.Namespace, curSvc.Name)
-			if shouldProcessUpdate(oldSvc, curSvc) {
+			if l4netLBc.shouldProcessService(curSvc, oldSvc) {
 				klog.V(3).Infof("L4 External LoadBalancer Service %s updated, enqueuing", svcKey)
 				l4netLBc.svcQueue.Enqueue(curSvc)
 				l4netLBc.enqueueTracker.Track()
@@ -119,16 +119,45 @@ func NewL4NetLBController(
 	return l4netLBc
 }
 
-// shouldProcess checks if given service should be process by controller
-func shouldProcess(svc *v1.Service) bool {
-	needsNetLB, _ := annotations.WantsL4NetLB(svc)
-	return needsNetLB || needsDeletion(svc)
+// needsAddition checks if given service should be added by controller
+func needsAddition(newSvc, oldSvc *v1.Service) bool {
+	if oldSvc != nil {
+		return false
+	}
+	needsNetLB, _ := annotations.WantsL4NetLB(newSvc)
+	return needsNetLB
+}
+
+func (lc *L4NetLBController) needsUpdate(newSvc, oldSvc *v1.Service) bool {
+	if oldSvc == nil {
+		return false
+	}
+	//TODO(kl52752) Implement a condition to compare services
+	return false
 }
 
 // shouldProcessUpdate checks if given service should be process by controller
-func shouldProcessUpdate(oldSvc, newSvc *v1.Service) bool {
+func (lc *L4NetLBController) shouldProcessService(newSvc, oldSvc *v1.Service) bool {
 	//TODO(kl52752) add check for update
-	return needsDeletion(newSvc)
+	if needsAddition(newSvc, oldSvc) || lc.needsUpdate(newSvc, oldSvc) || needsDeletion(newSvc) {
+		l4netlb := loadbalancers.NewL4NetLB(newSvc, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(newSvc.Namespace), &lc.sharedResourcesLock)
+		return lc.isRbsBasedLBService(newSvc, l4netlb)
+	}
+	return false
+}
+
+// isRbsBasedLBService returns if the given LoadBalancer service is not legacy target pool based LoadBalancer.
+func (lc *L4NetLBController) isRbsBasedLBService(svc *v1.Service, l4 *loadbalancers.L4NetLB) bool {
+	// skip services that are being handled by the legacy service controller.
+	if utils.IsLegacyL4NetLBService(svc) {
+		klog.Warningf("Ignoring update for service %s:%s managed by service controller", svc.Namespace, svc.Name)
+		return false
+	}
+	if lc.hasLegacyForwardingRule(svc) {
+		klog.Warningf("Ignoring update for service %s:%s which have legacy forwarding rule", svc.Namespace, svc.Name)
+		return false
+	}
+	return true
 }
 
 func (lc *L4NetLBController) checkHealth() error {
@@ -198,7 +227,7 @@ func (lc *L4NetLBController) sync(key string) error {
 // Returns an error if processing the service update failed.
 func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.SyncResultNetLB {
 	l4netlb := loadbalancers.NewL4NetLB(service, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(service.Namespace), &lc.sharedResourcesLock)
-	if !lc.shouldProcessService(service, l4netlb) {
+	if !lc.isRbsBasedLBService(service, l4netlb) {
 		return nil
 	}
 
@@ -251,20 +280,6 @@ func (lc *L4NetLBController) ensureBackendLinking(port utils.ServicePort) error 
 		return err
 	}
 	return lc.igLinker.Link(port, lc.ctx.Cloud.ProjectID(), zones)
-}
-
-// shouldProcessService returns if the given LoadBalancer service should be processed by this controller.
-func (lc *L4NetLBController) shouldProcessService(svc *v1.Service, l4 *loadbalancers.L4NetLB) bool {
-	// skip services that are being handled by the legacy service controller.
-	if utils.IsLegacyL4NetLBService(svc) {
-		klog.Warningf("Ignoring update for service %s:%s managed by service controller", svc.Namespace, svc.Name)
-		return false
-	}
-	if lc.hasLegacyForwardingRule(svc) {
-		klog.Warningf("Ignoring update for service %s:%s which have legacy forwarding rule", svc.Namespace, svc.Name)
-		return false
-	}
-	return true
 }
 
 func (lc *L4NetLBController) ensureInstanceGroups(service *v1.Service, nodeNames []string) error {
--- a/pkg/l4netlb/l4netlbcontroller.go
+++ b/pkg/l4netlb/l4netlbcontroller.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	v1 "k8s.io/api/core/v1"
 	listers "k8s.io/client-go/listers/core/v1"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/ingress-gce/pkg/instances"
 	"k8s.io/ingress-gce/pkg/loadbalancers"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/common"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/klog"
@@ -91,24 +93,42 @@ func NewL4NetLBController(
 		AddFunc: func(obj interface{}) {
 			addSvc := obj.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(addSvc.Namespace, addSvc.Name)
-			needsNetLB, svcType := annotations.WantsL4NetLB(addSvc)
-			//TODO (kl52752) Add check for deletion
-			if needsNetLB {
+			if shouldProcess(addSvc) {
 				klog.V(3).Infof("L4 External LoadBalancer Service %s added, enqueuing", svcKey)
 				l4netLBc.ctx.Recorder(addSvc.Namespace).Eventf(addSvc, v1.EventTypeNormal, "ADD", svcKey)
 				l4netLBc.svcQueue.Enqueue(addSvc)
 				l4netLBc.enqueueTracker.Track()
 			} else {
-				klog.V(4).Infof("Ignoring add for non external lb service %s based on %v", svcKey, svcType)
+				klog.V(4).Infof("Ignoring add for non external lb service %s", svcKey)
 			}
 		},
 		// Deletes will be handled in the Update when the deletion timestamp is set.
 		UpdateFunc: func(old, cur interface{}) {
-			//TODO(kl52752) add implementation and check for deletion
+			//TODO(kl52752) add implementation for update
+			curSvc := cur.(*v1.Service)
+			oldSvc := old.(*v1.Service)
+			svcKey := utils.ServiceKeyFunc(curSvc.Namespace, curSvc.Name)
+			if shouldProcessUpdate(oldSvc, curSvc) {
+				klog.V(3).Infof("L4 External LoadBalancer Service %s updated, enqueuing", svcKey)
+				l4netLBc.svcQueue.Enqueue(curSvc)
+				l4netLBc.enqueueTracker.Track()
+			}
 		},
 	})
 	ctx.AddHealthCheck("service-controller health", l4netLBc.checkHealth)
 	return l4netLBc
+}
+
+// shouldProcess checks if given service should be process by controller
+func shouldProcess(svc *v1.Service) bool {
+	needsNetLB, _ := annotations.WantsL4NetLB(svc)
+	return needsNetLB || needsDeletion(svc)
+}
+
+// shouldProcessUpdate checks if given service should be process by controller
+func shouldProcessUpdate(oldSvc, newSvc *v1.Service) bool {
+	//TODO(kl52752) add check for update
+	return needsDeletion(newSvc)
 }
 
 func (lc *L4NetLBController) checkHealth() error {
@@ -153,9 +173,17 @@ func (lc *L4NetLBController) sync(key string) error {
 		klog.V(3).Infof("Ignoring sync of non-existent service %s", key)
 		return nil
 	}
-	var result *loadbalancers.SyncResultNetLB
+	if needsDeletion(svc) {
+		klog.V(3).Infof("Deleting L4 External LoadBalancer resources for service %s", key)
+		result := lc.garbageCollectRBSNetLB(key, svc)
+		if result == nil {
+			return nil
+		}
+		return result.Error
+	}
+
 	if wantsNetLB, _ := annotations.WantsL4NetLB(svc); wantsNetLB {
-		result = lc.syncInternal(svc)
+		result := lc.syncInternal(svc)
 		if result == nil {
 			// result will be nil if the service was ignored(due to presence of service controller finalizer).
 			return nil
@@ -174,7 +202,10 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.Sy
 		return nil
 	}
 
-	// #TODO(kl52752) Add ensure finalizer for NetLB
+	if err := common.EnsureServiceFinalizer(service, common.NetLBFinalizerV2, lc.ctx.KubeClient); err != nil {
+		return &loadbalancers.SyncResultNetLB{Error: fmt.Errorf("Failed to attach L4 External LoadBalancer finalizer to service %s/%s, err %w", service.Namespace, service.Name, err)}
+	}
+
 	nodeNames, err := utils.GetReadyNodeNames(lc.nodeLister)
 	if err != nil {
 		return &loadbalancers.SyncResultNetLB{Error: err}
@@ -182,7 +213,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.Sy
 
 	if err := lc.ensureInstanceGroups(service, nodeNames); err != nil {
 		lc.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncInstanceGroupsFailed",
-			"Error syncing instance group: %v", err)
+			"Error syncing instance group, err: %v", err)
 		return &loadbalancers.SyncResultNetLB{Error: err}
 	}
 
@@ -191,13 +222,13 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.Sy
 	syncResult := l4netlb.EnsureFrontend(nodeNames, service)
 	if syncResult.Error != nil {
 		lc.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncExternalLoadBalancerFailed",
-			"Error ensuring Resource for L4 External LoadBalancer: %v", syncResult.Error)
+			"Error ensuring Resource for L4 External LoadBalancer, err: %v", syncResult.Error)
 		return syncResult
 	}
 
 	if err = lc.ensureBackendLinking(l4netlb.ServicePort); err != nil {
 		lc.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncExternalLoadBalancerFailed",
-			"Error linking instance groups to backend service: %v", err)
+			"Error linking instance groups to backend service, err: %v", err)
 		syncResult.Error = err
 		return syncResult
 	}
@@ -205,7 +236,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.Sy
 	err = lc.ensureServiceStatus(service, syncResult.Status)
 	if err != nil {
 		lc.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncExternalLoadBalancerFailed",
-			"Error updating L4 External LoadBalancer: %v", err)
+			"Error updating L4 External LoadBalancer, err: %v", err)
 		syncResult.Error = err
 		return syncResult
 	}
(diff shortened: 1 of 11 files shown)


pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -92,8 +93,7 @@ func NewL4NetLBController(
 			addSvc := obj.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(addSvc.Namespace, addSvc.Name)
 			needsNetLB, svcType := annotations.WantsL4NetLB(addSvc)
-			//TODO (kl52752) Add check for deletion
-			if needsNetLB {
+			if needsNetLB || l4netLBc.needsDeletion(addSvc) {
```
freehan, line 96: in Event Handler. 

We typically do not need to judge the intention of the user based on the object spec (deletion or not). Controller typically need to determine whether the object is "relevant" or not. If yes, then enqueue for the worker process to sync it. If not, then do not enqueue.

Then in the sync process, check user intention and act appropriately.

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -104,7 +104,16 @@ func NewL4NetLBController(
...
 		// Deletes will be handled in the Update when the deletion timestamp is set.
 		UpdateFunc: func(old, cur interface{}) {
-			//TODO(kl52752) add implementation and check for deletion
+			//TODO(kl52752) add implementation for update
+			curSvc := cur.(*v1.Service)
+			svcKey := utils.ServiceKeyFunc(curSvc.Namespace, curSvc.Name)
+			isDeleted := l4netLBc.needsDeletion(curSvc)
+			if isDeleted {
+				klog.V(3).Infof("L4 External LoadBalancer Service %s updated, needsDeletion = %v enqueuing", svcKey, isDeleted)
+				l4netLBc.ctx.Recorder(curSvc.Namespace).Eventf(curSvc, v1.EventTypeNormal, "UPDATE", svcKey)
```
freehan, line 113: Please note that periodic resync on service objects are enabled. Every 10 mins, this will emit an event for all relevant Services.

pkg/l4netlb/l4netlbcontroller.go
freehan: I would recommend you funnel Add/Update event to the common process function to determine if you want to sync it or not


```
       ...
       if needProcess(obj) {

             l4netLB.svcQeue.Enqueue(curSvc)
             ...
       }
}

       ...
       if needProcess(cur) {

             l4netLB.svcQeue.Enqueue(curSvc)
             ...
       }
}
```

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -174,15 +191,18 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.Sy
 		return nil
 	}
 
-	// #TODO(kl52752) Add ensure finalizer for NetLB
+	if err := common.EnsureServiceFinalizer(service, common.NetLBFinalizerV2, lc.ctx.KubeClient); err != nil {
```
freehan, line 194: Finalizer should be added before actual LB provisioning starts. This will prevent losing the finalizer state due to controller disruption.

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -246,3 +274,54 @@ func (lc *L4NetLBController) ensureServiceStatus(svc *v1.Service, newStatus *v1.
 	}
 	return patch.PatchServiceLoadBalancerStatus(lc.ctx.KubeClient.CoreV1(), svc, *newStatus)
 }
+
+func (lc *L4NetLBController) hasLegacyForwardingRule(svc *v1.Service) bool {
```
freehan, line 278: nit: comment for the function

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -246,3 +274,54 @@ func (lc *L4NetLBController) ensureServiceStatus(svc *v1.Service, newStatus *v1.
...
+	l4netlb := loadbalancers.NewL4NetLB(svc, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(svc.Namespace), &lc.sharedResourcesLock)
+	frName := utils.LegacyForwardingRuleName(svc)
+	existingFR := l4netlb.GetForwardingRule(frName, meta.VersionGA)
+	return existingFR != nil && existingFR.BackendService == "" && existingFR.Target != ""
+}
+
+func (lc *L4NetLBController) needsDeletion(svc *v1.Service) bool {
```
freehan, line 285: nit: comment for the function

needsDeletion return true if svc required deleting RBS based NetLB

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -246,3 +274,54 @@ func (lc *L4NetLBController) ensureServiceStatus(svc *v1.Service, newStatus *v1.
...
+		return false
+	}
+	if common.IsDeletionCandidateForGivenFinalizer(svc.ObjectMeta, common.NetLBFinalizerV2) {
+		return true
+	}
+	needsNetLB, _ := annotations.WantsL4NetLB(svc)
+	return !needsNetLB
+}
+
+func (lc *L4NetLBController) processServiceDeletion(key string, svc *v1.Service) *loadbalancers.SyncResult {
```
freehan, line 296: add comment

recommend changing the function name to `GC` as it is not necessarily service deletion. Or something more specific like `GarbageCollectRBSNetLB`

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -246,3 +274,54 @@ func (lc *L4NetLBController) ensureServiceStatus(svc *v1.Service, newStatus *v1.
...
+		return result
+	}
+	if err := lc.ensureServiceStatus(svc, &v1.LoadBalancerStatus{}); err != nil {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
+			"Error reseting L4 External LoadBalancer status to empty, err: %v", err)
+		result.Error = fmt.Errorf("Failed to reset L4 External LoadBalancer status, err: %w", err)
+		return result
+	}
+	igName := lc.ctx.ClusterNamer.InstanceGroup()
+	if err := lc.instancePool.DeleteInstanceGroup(igName); err != err {
```
freehan, line 313: do you need to try delete IG here? 

I assume IG controller needs to know when IG is needed right? And then IG controller will manage IG lifecycle.

freehan: One benefit of keeping the IG logic here is that potentially IG gets clean up more promptly. NO need to wait for periodic GC at the IG controller. May be worth adding a comment here to describe the reasoning.

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 	return result
 }
 
+// EnsureLoadBalancerDeleted performs a cleanup of all GCE resources for the given loadbalancer service.
+// It is static IP address, helath check, firewall rules and backend service
```
freehan, line 149: typo health check

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResult {
+	sharedHC := !helpers.RequestsOnlyLocalTraffic(svc)
```
freehan, line 151: nit: no need to keep this a var right? just need to call the function directly later?

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+	}
+	name := l4netlb.ServicePort.BackendName()
+	if err = ensureAddressDeleted(l4netlb.cloud, name, l4netlb.cloud.Region()); err != nil {
+		klog.Errorf("Failed to delete address for service %s - %v", l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		result.GCEResourceInError = annotations.AddressResource
+	}
+	hcName, hcFwName := l4netlb.namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	// delete fw rules
+	deleteFunc := func(name string) error {
```
freehan, line 175: nit: deleteFirewallFunc

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+	// delete firewall rule allowing load balancer source ranges
+	err = deleteFunc(name)
+	if err != nil {
+		klog.Errorf("Failed to delete firewall rule %s for service %s - %v", name, l4netlb.NamespacedName.String(), err)
+		result.GCEResourceInError = annotations.FirewallRuleResource
+		result.Error = err
+	}
+
+	// delete firewall rule allowing healthcheck source ranges
+	err = deleteFunc(hcFwName)
```
freehan, line 195: How do we currently configures firewall rules? 

Do we keep 2 firewall rules for each LB? I thought we only have one firewall per NetLB. 

If so, is it possible to merge them?  An alternative is to maintain only 1 firewall rule to allow health check ranges for all NetLBs. 

This is to prevent the controller taking too much firewall quota for new NetLBs.

pkg/loadbalancers/l4netlb.go
freehan: also, do you delete static IP address?

freehan: if the user manually promote an emphameral IP address to static, what is the behavior today? I think we just leave it alone?

pkg/l4netlb/l4netlbcontroller.go
kl52752: done, although the update function (in future - not implemented yet) have to check if service was updated so need Process will take 2 params `needProcessUpdate(newSvc, oldSvc)`

pkg/l4netlb/l4netlbcontroller.go
kl52752: done - comment removed

pkg/l4netlb/l4netlbcontroller.go
kl52752: Well before finalizer is set there is only check if we need to process this service so I thought that this is a good place to put it

pkg/l4netlb/l4netlbcontroller.go
kl52752: done

pkg/l4netlb/l4netlbcontroller.go
kl52752: done

pkg/l4netlb/l4netlbcontroller.go
kl52752: I removed deletion of IG here.

pkg/loadbalancers/l4netlb.go
kl52752: I invoke `ensureAddressDeleted` and it calls API Addresses delete

kl52752: In kubernetes/kubernetes static IP address is also deleted

kl52752: https://github.com/kubernetes/kubernetes/blob/c9ddd248b6d58c4c973d507be07261b96ef0cfbc/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_external.go#L335

pkg/loadbalancers/l4netlb.go
kl52752: it is passed to l4netlb.namer.L4HealthCheck so imo it is more readable as var. It moved it down, line before the function where it is used

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: For Internal Load Balancer Firewall rules are created the same way (2 for each LB)
I'm afraid that we cannot merge this rules, looking to the code we have 
1. rule for LB and for common we open LB port por all traffic fg: port 8080 range:0.0.0.0/0
2. for Health check we have hcPort and l4 load balancer source ranges fg port: 32586 range:[35.191.0.0/16 209.85.152.0/22 209.85.204.0/22 130.211.0.0/22]
so from what I understand if I would like to merge those rules I will end up with ports:  8080,32586 for range 0.0.0.0/0

@prameshj can you confirm that?

prameshj: I agree, it would be hard to merge them. One firewall rule is for LB traffic, allowlisting source ranges specified in the service spec (per service firewall rule)

Another firewall rule is to allow healthchecks from the Google Healthcheck source ranges (per-service for Local trafficpolicy, shared rule for Cluster traffic policy).

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -91,26 +92,43 @@ func NewL4NetLBController(
...
 }
 
+// shouldProcess checks if given service should be porcess by controller
+func shouldProcess(svc *v1.Service) bool {
+	needsNetLB, _ := annotations.WantsL4NetLB(svc)
+	return needsNetLB || needsDeletion(svc)
+}
+
+// shouldProcessUpdate checks if given service should be porcess by controller
+func shouldProcessUpdate(oldSvc, newSvc *v1.Service) bool {
```
freehan, line 127: I am wondering if you want to enable periodic reconciliation for the LBs. I think you want to right? 

With this you only cehck if newSvc needsDeletion or not and do not do periodic reconciliation. 

@prameshj

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 	return result
 }
 
+// EnsureLoadBalancerDeleted performs a cleanup of all GCE resources for the given loadbalancer service.
+// It is static IP address, health check, firewall rules and backend service
```
freehan, line 149: This function does not clean up static IP, right?

One corner case is this: after user created the LB type service, user manually promoted the VIP from empheral to static IP. In this case, we should not delete user's manual promoted static IP. 

Also check how the existing NetLB handles this.

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+	if err = utils.IgnoreHTTPNotFound(composite.DeleteForwardingRule(l4netlb.cloud, key, meta.VersionGA)); err != nil {
+		klog.Errorf("Failed to delete forwarding rule %s for service %s - %v", frName, l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		result.GCEResourceInError = annotations.ForwardingRuleResource
+	}
+	name := l4netlb.ServicePort.BackendName()
+	if err = ensureAddressDeleted(l4netlb.cloud, name, l4netlb.cloud.Region()); err != nil {
+		klog.Errorf("Failed to delete address for service %s - %v", l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		result.GCEResourceInError = annotations.AddressResource
```
freehan, line 170: does the result need to maintain a list of GCEResourceInError? or just one?

pkg/loadbalancers/l4netlb.go
```diff
@@ -145,6 +145,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+		result.GCEResourceInError = annotations.BackendServiceResource
+		result.Error = err
+	}
+
+	// Delete healthcheck
+	if sharedHC {
+		l4netlb.sharedResourcesLock.Lock()
+		defer l4netlb.sharedResourcesLock.Unlock()
+	}
+	err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l4netlb.cloud, hcName, l4netlb.scope))
```
freehan, line 214: One suggestion, always check dedicated HC despite it is sharedHC.
This is prevent a LB service change from ExternalTrafficPolicy from Local to Cluster during the lifecycle of the LB and we leak the health check.

freehan (approved):
overall looking good.

pkg/l4netlb/l4netlbcontroller.go
kl52752: Oh, I see :) Removed

pkg/l4netlb/l4netlbcontroller.go
kl52752: Yes, I will be added in next Task when I will handle service Update

pkg/loadbalancers/l4netlb.go
kl52752: yes, you are right  I don't clean up static IP which can be owned by User. 
In legacy NetLB region address is clean up in case there is some error in create/update attempt

pkg/loadbalancers/l4netlb.go
kl52752: fixed

pkg/loadbalancers/l4netlb.go
kl52752: Shouldn't change ExternalTrafficPolicy from Local to Cluster be handled in Update? 
In this function I handled only deleting resources so If HC is shared and some other resource is using it then Health Check won't be deleted, 
I intend to post PR with Update functionality after this one is merged.

freehan: Well. That sounds quite messy as every sync. It will need to check if the other HC setup exists. 

Instead, I would suggest just do this during GC, so that it is one time operation.

pkg/l4netlb/l4netlbcontroller.go
freehan: Okay

freehan (commented):
Looks good overall. 

Just waiting for @prameshj for a scan.

pkg/loadbalancers/l4netlb.go
prameshj: Good idea. Can we add a TODO to address this in deletion codepath?

ILB Controller has this same issue as well.

pkg/loadbalancers/l4netlb.go
```diff
@@ -147,6 +147,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+		result.GCEResourceInError += fmt.Sprintf(" %s", annotations.ForwardingRuleResource)
+	}
+	name := l4netlb.ServicePort.BackendName()
+	if err = ensureAddressDeleted(l4netlb.cloud, name, l4netlb.cloud.Region()); err != nil {
+		klog.Errorf("Failed to delete address for service %s - %v", l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		result.GCEResourceInError += fmt.Sprintf(" %s", annotations.AddressResource)
+	}
+	sharedHC := !helpers.RequestsOnlyLocalTraffic(svc)
+	hcName, hcFwName := l4netlb.namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
```
prameshj, line 175: I just realised that this will generate the same healthcheck name(when shared) for ILB and NetLB. It can be tricky to use the same healthcheck for ILB and NetLB - we need to make sure the params are same in both cases, if we are sharing, otherwise the reconcile function - https://github.com/kubernetes/ingress-gce/blob/458f1dda1a4de50ff8b965510a2246e797bde3d2/pkg/healthchecks/healthchecks_l4.go#L155 will keep updating healthcheck in each service sync.

pkg/utils/utils.go
```diff
@@ -658,6 +658,16 @@ func IsSubsettingL4ILBService(svc *api_v1.Service) bool {
...
 }
 
+// IsL4NetLBService returns true if the given LoadBalancer service is managed by L4NetLBController.
+func IsL4NetLBService(svc *api_v1.Service) bool {
+	return slice.ContainsString(svc.ObjectMeta.Finalizers, common.NetLBFinalizerV2, nil)
+}
+
+// IsLegacyL4NetLBService returns true if the given LoadBalancer service is managed by service controller.
+func IsLegacyL4NetLBService(svc *api_v1.Service) bool {
+	return slice.ContainsString(svc.ObjectMeta.Finalizers, common.LegacyNetLBFinalizer, nil)
```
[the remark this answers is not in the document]
kl52752: No, only new or migrated services will have this finalizer so this is the only controller that can add this finalizer
@cezarygerard can you confirm?

pkg/utils/common/finalizer.go
```diff
@@ -39,6 +39,10 @@ const (
 	ILBFinalizerV2 = "gke.networking.io/l4-ilb-v2"
 	// NegFinalizerKey is the finalizer used by neg controller to ensure NEG CRs are deleted after corresponding negs are deleted
 	NegFinalizerKey = "networking.gke.io/neg-finalizer"
+	// LegacyNetLBFinalizer key is used to identify L4 Externall LoadBalancer services whose resources are managed by service controller.
+	LegacyNetLBFinalizer = "gke.networking.io/l4-netlb-v1"
+	// NetLBFinalizerV2 is the finalizer used by newer controllers that manage L4 Externall LoadBalancer services.
```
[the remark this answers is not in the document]
kl52752: done

pkg/test/gcehooks.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func InsertForwardingRuleHook(ctx context.Context, key *meta.Key, obj *compute.ForwardingRule, m *cloud.MockForwardingRules) (b bool, e error) {
+	if obj.IPAddress == "" {
+		obj.IPAddress = FwIPAddress
+	}
+	return false, nil
+}
+
+func DeleteForwardingRulesdHookError(ctx context.Context, key *meta.Key, m *cloud.MockForwardingRules) (bool, error) {
```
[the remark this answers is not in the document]
kl52752: done

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -54,6 +55,70 @@ func TestEnsureL4NetLoadBalancer(t *testing.T) {
...
+	if err := l4NetLB.EnsureLoadBalancerDeleted(svc); err.Error != nil {
+		t.Errorf("UnexpectedError %v", err.Error)
+	}
+	ensureNetLBResourceDeleted(t, svc, l4NetLB)
+}
+func ensureNetLBResourceDeleted(t *testing.T, apiService *v1.Service, l4NetLb *L4NetLB) {
+	t.Helper()
+
+	resourceName := l4NetLb.ServicePort.BackendName()
+	hcName, hcFwName := l4NetLb.namer.L4HealthCheck(apiService.Namespace, apiService.Name, true)
```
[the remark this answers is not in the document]
kl52752: done

pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -265,3 +288,129 @@ func TestForwardingRuleWithPortRange(t *testing.T) {
 		deleteNetLBService(lc, svc)
 	}
 }
+
+func TestProcessServiceCreate(t *testing.T) {
+	svc, lc := createAndSyncNetLBSvc(t)
+	nodePort := int32(0)
```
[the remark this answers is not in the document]
kl52752: fixed, Added NodePort to service manifest

pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -265,3 +288,129 @@ func TestForwardingRuleWithPortRange(t *testing.T) {
...
+	key, _ := common.KeyFunc(svc)
+	err := lc.sync(key)
+	if err != nil {
+		t.Errorf("Failed to sync service %s, err %v", svc.Name, err)
+	}
+	svc, err = lc.ctx.KubeClient.CoreV1().Services(svc.Namespace).Get(context.TODO(), svc.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("Failed to lookup service %s, err %v", svc.Name, err)
+	}
+	if len(svc.Status.LoadBalancer.Ingress) > 0 && svc.Status.LoadBalancer.Ingress[0].IP == test.FwIPAddress {
```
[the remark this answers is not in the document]
kl52752: done

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -246,3 +284,50 @@ func (lc *L4NetLBController) ensureServiceStatus(svc *v1.Service, newStatus *v1.
 	}
 	return patch.PatchServiceLoadBalancerStatus(lc.ctx.KubeClient.CoreV1(), svc, *newStatus)
 }
+
+// hasLegacyForwardingRule return true if forwarding rule is target pool based
+func (lc *L4NetLBController) hasLegacyForwardingRule(svc *v1.Service) bool {
+	l4netlb := loadbalancers.NewL4NetLB(svc, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(svc.Namespace), &lc.sharedResourcesLock)
+	frName := utils.LegacyForwardingRuleName(svc)
+	existingFR := l4netlb.GetForwardingRule(frName, meta.VersionGA)
+	return existingFR != nil && existingFR.BackendService == "" && existingFR.Target != ""
```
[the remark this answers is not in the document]
kl52752: done

pkg/loadbalancers/l4netlb.go
```diff
@@ -147,6 +147,87 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
+	if err != nil {
+		klog.Errorf("Failed to create key for forwarding rule resources with name %s for service %s - %v", frName, l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		return result
+	}
+	// If any resource deletion fails, log the error and continue cleanup.
+	if err = utils.IgnoreHTTPNotFound(composite.DeleteForwardingRule(l4netlb.cloud, key, meta.VersionGA)); err != nil {
+		klog.Errorf("Failed to delete forwarding rule %s for service %s - %v", frName, l4netlb.NamespacedName.String(), err)
+		result.Error = err
+		result.GCEResourceInError += fmt.Sprintf(" %s", annotations.ForwardingRuleResource)
```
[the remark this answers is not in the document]
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: I checked and the only thing that will differ is the description because I added LB type to description. So maybe I can change the description to 
`This resource is shared by all L4 ILBs and XLBs Services using ExternalTrafficPolicy: Cluster.`

kl52752: There is another difference in params passing to HC it is `scope` and `Region` but 
`func needToUpdateHealthChecks` doesn't check them

pkg/l4netlb/l4netlbcontroller.go
kl52752: I refactor those functions and added to new commit https://github.com/kubernetes/ingress-gce/pull/1596/commits/fdb02f1c738e2ae62c7d269c0286ecf4fe0af993

pkg/utils/utils.go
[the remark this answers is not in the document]
prameshj: I am referring to the legacy finalizer. I thought they should be added by service controller?

pkg/loadbalancers/l4netlb.go
prameshj: can we address that in a separate PR? Changing description + making sure that the same lock is used should be ok i think.

prameshj:
/lgtm

Open items are: 1) identify how the legacy finalizer will be added 2) Handle shared healthchecks.

PR #1597, opened 2021-10-29
Readme: update version mapping

zioproto:
trivial: update readme with latest version mappings of GKE and ingress-gce

This is getting complex to keep up to date for GKE customers. For example this list does not capture the detail that some early GKE 1.19 releases are on ingress-gce tag  1.10.15. This is not relevant anymore because those version are not available anymore to customers.


zioproto:
Cc: @clsacramento 

We should think of a better way to provide this information in the future.

PR #1599, opened 2021-11-05
Fix ports in L4NetLB

kl52752:
In L4NetLB NodePort is used by namer to create BackendService name.
Previously NodePort was not set and by default its value was 0 so
all backends had the same name and we there were error in creating
multiple services
To test NodePort set multi-services test were written

Change port ranges calculation for creating forwarding rule.
Previously we assigned first continue port Range from service,
now it is calculated from service assigned ports to
create boundary range MINPORT-MAXPORT.

### DIFF
--- a/pkg/l4netlb/l4netlbcontroller_test.go
+++ b/pkg/l4netlb/l4netlbcontroller_test.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package l4netlb
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog"
+	"k8s.io/legacy-cloud-providers/gce"
+
+	"testing"
+
+	ga "google.golang.org/api/compute/v1"
+	"k8s.io/ingress-gce/pkg/composite"
+	ingctx "k8s.io/ingress-gce/pkg/context"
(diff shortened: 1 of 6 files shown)


kl52752:
\assign @freehan

pkg/loadbalancers/l4netlb.go
```diff
@@ -79,8 +79,10 @@ func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, n
 		backendPool:         backends.NewPool(cloud, namer),
 	}
 	portId := utils.ServicePortID{Service: l4netlb.NamespacedName}
-	l4netlb.ServicePort = utils.ServicePort{ID: portId,
+	l4netlb.ServicePort = utils.ServicePort{
+		ID:           portId,
 		BackendNamer: l4netlb.namer,
+		NodePort:     int64(service.Spec.Ports[0].NodePort),
```
prameshj, line 85: We should only set this if the value is non-zero. It is possible to disable NodePort allocation for LoadBalancer services using https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1864-disable-lb-node-ports. 

Also, what is this used for? To annotate the instance group with the service port?

pkg/utils/utils.go
```diff
@@ -612,6 +612,29 @@ func GetPortsAndProtocol(svcPorts []api_v1.ServicePort) (ports []string, portRan
...
+			minPort = svcPort.Port
+		}
+		if svcPort.Port > maxPort {
+			maxPort = svcPort.Port
+		}
+	}
+	return minPort, maxPort
+}
+
+func GetPortMinMaxRangeAndProtocol(svcPorts []api_v1.ServicePort) (portRange, protocol string) {
```
prameshj, line 629: test for this util?

Also, nit: You can drop the Get prefix and call this MinMaxPortRangeAndProtocol.

pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -0,0 +1,194 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
prameshj, line 2: nit: should this test addition be part of a separate commit/PR? It is not adding a test for the multiple port forwarding rule right?

I think this test should also verify forwarding rule(in addition to backend service) and have a testcase for multiple ports forwarding rule.

pkg/loadbalancers/l4netlb.go
kl52752: Node porst is used by Namer to create backend service name `sp.BackendNamer.IGBackend(sp.NodePort)`

pkg/utils/utils.go
kl52752: done + test added

pkg/l4netlb/l4netlbcontroller_test.go
kl52752: I added here test for multiservice because this way I can test changes with Node Port where we calculate new name for backend based on NodePort. I will add test for multiport forwarding rule.

kl52752: done added test for multiple ports forwarding rule

pkg/loadbalancers/l4netlb.go
kl52752: As I can see it is also used in L7 LB

PR #1600, opened 2021-11-05
Trigger sync for certain node changes

swetharepakula:
* VmIpPort NEGs should be synced when node change results in a zone
   being added or removed

/assign @freehan @prameshj

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -283,27 +283,29 @@ func NewController(
 		})
 	}
 
-	if negController.runL4 {
-		nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-			AddFunc: func(obj interface{}) {
-				node := obj.(*apiv1.Node)
-				negController.enqueueNode(node)
-			},
-			DeleteFunc: func(obj interface{}) {
-				node := obj.(*apiv1.Node)
-				negController.enqueueNode(node)
-			},
-			UpdateFunc: func(old, cur interface{}) {
-				oldNode := old.(*apiv1.Node)
-				currentNode := cur.(*apiv1.Node)
-				candidateNodeCheck := utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
-				if candidateNodeCheck(oldNode) != candidateNodeCheck(currentNode) {
-					klog.Infof("Node %q has changed, enqueueing", currentNode.Name)
-					negController.enqueueNode(currentNode)
-				}
-			},
-		})
+	nodeEventHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			node := obj.(*apiv1.Node)
+			negController.enqueueNode(node)
+		},
+		DeleteFunc: func(obj interface{}) {
+			node := obj.(*apiv1.Node)
+			negController.enqueueNode(node)
+		},
 	}
+
+	if negController.runL4 {
+		nodeEventHandler.UpdateFunc = func(old, cur interface{}) {
+			oldNode := old.(*apiv1.Node)
+			currentNode := cur.(*apiv1.Node)
+			candidateNodeCheck := utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
+			if candidateNodeCheck(oldNode) != candidateNodeCheck(currentNode) {
+				klog.Infof("Node %q has changed, enqueueing", currentNode.Name)
+				negController.enqueueNode(currentNode)
+			}
+		}
+	}
+	nodeInformer.AddEventHandler(nodeEventHandler)
 
 	if enableAsm {
 		negController.enableASM = enableAsm
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -93,6 +93,10 @@ type syncerManager struct {
 	// This will make all NEGs created by NEG controller to be NON_GCP_PRIVATE_IP_PORT type.
 	enableNonGcpMode     bool
 	enableEndpointSlices bool
+
+	// zoneMap keeps track of the last set of zones the neg controller
+	// has seen. zoneMap is protected by the mu mutex.
+	zoneMap map[string]struct{}
 }
 
 func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer,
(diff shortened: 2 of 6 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -466,6 +468,29 @@ func (s *transactionSyncer) commitPods(endpointMap map[string]negtypes.NetworkEn
 	}
 }
 
+func (s *transactionSyncer) isZoneChange() bool {
```
freehan, line 471: add comment

pkg/neg/manager.go
```diff
@@ -268,8 +282,24 @@ func (manager *syncerManager) Sync(namespace, name string) {
...
+	for _, zone := range zones {
+		newZoneMap[zone] = struct{}{}
+	}
+
+	needSync := false
+	if !reflect.DeepEqual(manager.zoneMap, newZoneMap) {
+		needSync = true
+	}
+	manager.zoneMap = newZoneMap
 	for key, syncer := range manager.syncerMap {
```
freehan, line 301: add a comment to say the this is to trigger sync whenever the list of zone changes

pkg/neg/syncers/transaction.go
```diff
@@ -466,6 +468,29 @@ func (s *transactionSyncer) commitPods(endpointMap map[string]negtypes.NetworkEn
...
+	for _, ref := range negCR.Status.NetworkEndpointGroups {
+		id, err := cloud.ParseResourceURL(ref.SelfLink)
+		if err != nil {
+			klog.Warningf("unable to parse selflink %s", ref.SelfLink)
+			continue
+		}
+		existingZones.Insert(id.Key.Zone)
+	}
+
+	zones, err := s.zoneGetter.ListZones(negtypes.NodePredicateForEndpointCalculatorMode(s.EpCalculatorMode))
```
freehan, line 488: log the error?

pkg/neg/manager.go
```diff
@@ -93,6 +93,10 @@ type syncerManager struct {
 	// This will make all NEGs created by NEG controller to be NON_GCP_PRIVATE_IP_PORT type.
 	enableNonGcpMode     bool
 	enableEndpointSlices bool
+
+	// zoneMap keeps track of the last set of zones the neg controller
+	// has seen
+	zoneMap map[string]struct{}
```
freehan, line 99: add comment to say this is protected by mu mutex

pkg/neg/manager.go
```diff
@@ -268,8 +282,24 @@ func (manager *syncerManager) Sync(namespace, name string) {
 func (manager *syncerManager) SyncNodes() {
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
+
+	zones, err := manager.zoneGetter.ListZones(utils.AllNodesPredicate)
```
freehan, line 286: I would recommend just put this whole section into a separate function to make it cleaner. 

And make sure it does not block in anyway.

pkg/neg/manager.go
```diff
@@ -268,8 +282,24 @@ func (manager *syncerManager) Sync(namespace, name string) {
...
+	if err != nil {
+		klog.V(3).Infof("Unable to initialize zone map in neg manager: %s", err)
```
freehan, line 288: when err? is the zones going to be nil or empty? would that be a problem?

pkg/neg/manager.go
prameshj, line 301: nit: you could add: ```if key.NegType == negtypes.VmIpEndpointType { needSync = true ``` to simplify the if condition.

pkg/neg/manager_test.go
```diff
@@ -1413,6 +1413,127 @@ func TestGarbageCollectionNegCrdEnabled(t *testing.T) {
...
+			desc:          "vm ip neg syncer is stopped",
+			expectSync:    false,
+			syncerStopped: true,
+			negType:       negtypes.VmIpEndpointType,
+		},
+		{
+			desc:          "vm ip port neg syncer is stopped",
+			expectSync:    false,
+			syncerStopped: true,
+			negType:       negtypes.VmIpEndpointType,
```
prameshj, line 1473: Fix negType.

pkg/neg/manager.go
```diff
@@ -268,8 +282,24 @@ func (manager *syncerManager) Sync(namespace, name string) {
 func (manager *syncerManager) SyncNodes() {
```
prameshj, line 282: SyncNodes will only be invoked on node changes if ILB Subsetting is enabled - https://github.com/kubernetes/ingress-gce/blob/6af12c2a66bfc8b59c9b98ff7be9ef1999f1c51d/pkg/neg/controller.go#L286-L303

Otherwise we do not enqueue node events. Should that be changed too?

swetharepakula: That's a good catch! Yes we need to change that so it will always invoke now.

pkg/neg/manager.go
swetharepakula: I pulled out the section and now if we error here, we just return false for the zone change, and not update the zone map. The transaction syncer also does a check, so next periodic sync will pick up the zone change.

pkg/neg/manager.go
swetharepakula: Since needSync is a check done overall, and the neg type is done by syncer, I don't think we can simplify the if condition.

pkg/neg/manager.go
swetharepakula: Removed if condition, so handlers are always added.

pkg/neg/controller.go
```diff
@@ -283,27 +283,25 @@ func NewController(
...
+	nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			node := obj.(*apiv1.Node)
+			negController.enqueueNode(node)
+		},
+		DeleteFunc: func(obj interface{}) {
+			node := obj.(*apiv1.Node)
+			negController.enqueueNode(node)
+		},
+		UpdateFunc: func(old, cur interface{}) {
```
prameshj, line 295: Should we enqueue node updates just for subsetting? Does L7 care if a node became an L4 Lb candidate or not?

swetharepakula: That's a good point, I will update so only add and delete are added when L4 is not enabled.

freehan:
Looks good overall. 
Wait for @prameshj for the final LGTM

PR #1603, opened 2021-11-10
[Cherrypick #1600 into 1.13]  Trigger sync for certain node changes

swetharepakula:
* VmIpPort NEGs should be synced when node change results in a zone
   being added or removed
 * Node informer event handlers are initialized in all cases


freehan (approved):
/lgtm
/approve

PR #1606, opened 2021-11-10
Unified  health check description for ILB and XLB resource

kl52752:
When External or Internal L4 LoadBalancer creates a shared health check
it should set description indicating that this resource is shared
among all L4 LoadBalancers despise their type.

/assign @prameshj

### DIFF
--- a/pkg/firewalls/firewalls_l4.go
+++ b/pkg/firewalls/firewalls_l4.go
@@ -36,7 +36,7 @@ func EnsureL4FirewallRule(cloud *gce.Cloud, fwName, lbIP, nsName string, sourceR
 	if err != nil {
 		return err
 	}
-	fwDesc, err := utils.MakeL4LBServiceDescription(nsName, lbIP, meta.VersionGA, sharedRule, l4Type)
+	fwDesc, err := utils.MakeL4LBServiceDescription(nsName, lbIP, meta.VersionGA, sharedRule)
 	if err != nil {
 		klog.Warningf("EnsureL4FirewallRule(%v): failed to generate description for L4 %s rule, err: %v", fwName, l4Type.ToString(), err)
 	}
--- a/pkg/healthchecks/healthchecks_l4.go
+++ b/pkg/healthchecks/healthchecks_l4.go
@@ -44,6 +44,7 @@ const (
 
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
+// If ExternalTrafficPolicy is Cluster then health check is shared among all ILBs and XLBs
 func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
@@ -59,7 +60,7 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
 	selfLink := ""
 	key, err := composite.CreateKey(cloud, name, scope)
 	if err != nil {
-		return nil, selfLink, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", name, err)
+		return nil, selfLink, fmt.Errorf("Failed to create composite key for healthcheck %s of type %v - %w", name, l4Type.ToString(), err)
 	}
 	hc, err := composite.GetHealthCheck(cloud, key, meta.VersionGA)
 	if err != nil {
@@ -71,10 +72,16 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
 	if scope == meta.Regional {
 		region = cloud.Region()
 	}
-	expectedHC := NewL4HealthCheck(name, svcName, shared, path, port, l4Type, scope, region)
+
+	// If TrafficPolicyLocal is equal to "Cluster" then ILBs and XLBs should share HC
+	// but because they use HC with different scope there will be 2 health checks with the same name
+	// one Global and other one Regional.
+	// Since ILB supports Regional HC also it is recommended to change ILB HC scope to Regional to limit HC resources.
+	// https://cloud.google.com/load-balancing/docs/health-check-concepts#lb_guide
+	expectedHC := NewL4HealthCheck(name, svcName, shared, path, port, scope, region)
 	if hc == nil {
 		// Create the healthcheck
-		klog.V(2).Infof("Creating healthcheck %s for service %s, shared = %v", name, svcName, shared)
+		klog.V(2).Infof("Creating healthcheck %s for service %s of type %v, shared = %v", name, svcName, l4Type.ToString(), shared)
 		err = composite.CreateHealthCheck(cloud, key, expectedHC)
 		if err != nil {
 			return nil, selfLink, err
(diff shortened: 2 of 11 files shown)


pkg/firewalls/firewalls_l4.go
```diff
@@ -36,9 +36,9 @@ func EnsureL4FirewallRule(cloud *gce.Cloud, fwName, lbIP, nsName string, sourceR
 	if err != nil {
 		return err
 	}
-	fwDesc, err := utils.MakeL4LBServiceDescription(nsName, lbIP, meta.VersionGA, sharedRule, l4Type)
+	fwDesc, err := utils.MakeL4LBServiceDescription(nsName, lbIP, meta.VersionGA, sharedRule)
```
prameshj, line 39: Will firewall rule be shared between ILB and External LB? IIRC, the source ranges for healthchecks are different in case of ILB vs ExternalLB. So, we need to handle modifying source ranges if they are being shared. Easiest is to keep them separate.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -44,18 +44,18 @@ const (
...
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType) (string, string, int32, string, error) {
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
 	if !sharedHC {
 		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
 	}
 	nn := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
-	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, nn, sharedHC, hcPath, hcPort, scope, l4Type)
+	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, nn, sharedHC, hcPath, hcPort, scope)
```
prameshj, line 54: How about we keep the l4Type parameter, but we don't include it in description for shared healthcheck?

pkg/firewalls/firewalls_l4.go
```diff
@@ -36,9 +36,9 @@ func EnsureL4FirewallRule(cloud *gce.Cloud, fwName, lbIP, nsName string, sourceR
...
 	if err != nil {
-		klog.Warningf("EnsureL4FirewallRule(%v): failed to generate description for L4 %s rule, err: %v", fwName, l4Type.ToString(), err)
+		klog.Warningf("EnsureL4FirewallRule(%v): failed to generate description for L4 rule, err: %v", fwName, err)
```
prameshj, line 41: Adding Lb type in these logs is good to have. Can we just change the MakeL4LBServiceDescription to not take the lbType param?

prameshj (commented):
Can we verify this on a real cluster?

Create 2 ILBs with Subsetting - Local and Cluster traffic policy.
Create 2 External LBs with RBS - Local and Cluster traffic policy and make sure they all show healthy backends?

pkg/firewalls/firewalls_l4.go
prameshj: Followed up offline. The Source ranges are same for L4 internal and external LBs(https://github.com/kubernetes/ingress-gce/blob/aeae5e419e42d2e5b6c8be4571e87d1e2da6e2ff/vendor/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer.go#L94), they are only different for L7 LBs. So it is fine to reuse the firewall rule.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -44,18 +44,18 @@ const (
 
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
```
prameshj, line 46: Add a comment that the healthcheck is shared by L4 ILB and XLB, in case of ExternalTrafficPolicy: Cluster?

pkg/healthchecks/healthchecks_l4.go
kl52752: done

pkg/firewalls/firewalls_l4.go
kl52752: done

pkg/healthchecks/healthchecks_l4.go
kl52752: done

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -44,6 +44,7 @@ const (
...
+// If ExternalTrafficPollicy is Cluster then health check is shared among all ILBs and XLBs
```
prameshj, line 47: Super nit: "TrafficPollicy" -> "TrafficPolicy"
Also end with a ".".

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -71,10 +72,10 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
 	if scope == meta.Regional {
 		region = cloud.Region()
 	}
-	expectedHC := NewL4HealthCheck(name, svcName, shared, path, port, l4Type, scope, region)
+	expectedHC := NewL4HealthCheck(name, svcName, shared, path, port, scope, region)
```
prameshj, line 75: Regional healthcheck is supported by ILB and XLB - Can you expand the doc to include this line - https://cloud.google.com/load-balancing/docs/health-check-concepts#lb_guide

So we could change to use Regional scope always, in l4.go as well.

pkg/healthchecks/healthchecks_l4.go
kl52752: done

pkg/healthchecks/healthchecks_l4.go
prameshj: Was this resolved?

pkg/loadbalancers/l4_test.go
```diff
@@ -1181,7 +1178,7 @@ func assertInternalLbResources(t *testing.T, apiService *v1.Service, l *L4, node
 	}
 
 	// Check that HealthCheck is created
-	healthcheck, err := composite.GetHealthCheck(l.cloud, meta.GlobalKey(hcName), meta.VersionGA)
+	healthcheck, err := composite.GetHealthCheck(l.cloud, meta.RegionalKey(hcName, l.cloud.Region()), meta.VersionGA)
```
prameshj, line 1181: Can we check that the healthcheck scope and region value matches expectation?

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -45,40 +43,30 @@ const (
...
-	var region string
-	if scope == meta.Regional {
-		region = cloud.Region()
-	}
 
-	// If TrafficPolicyLocal is equal to "Cluster" then ILBs and XLBs should share HC
-	// but because they use HC with different scope there will be 2 health checks with the same name
-	// one Global and other one Regional.
-	// Since ILB supports Regional HC also it is recommended to change ILB HC scope to Regional to limit HC resources.
+	// Regional healthcheck is supported by ILB and XLB
```
prameshj, line 67: Can you move the comments on line 67 and 68, to line 45? 

nit: end the comments with "."

prameshj (commented):
Looks good to me. Will add the lgtm tag once the comments are addressed.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -45,40 +43,30 @@ const (
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
```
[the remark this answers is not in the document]
kl52752: done

pkg/healthchecks/healthchecks_l4.go
kl52752: done

pkg/loadbalancers/l4_test.go
kl52752: We can check the region, scope is not present in ga.healthcheck so in Mock conversion it is reseted

prameshj: Thanks, resolving since the region check has been added.

PR #1607, opened 2021-11-12
Update L4NetLB services

kl52752:
Add function to L4NetLBController which handles External L4 LoadBalancer services update

### DIFF
--- a/pkg/l4netlb/l4netlbcontroller.go
+++ b/pkg/l4netlb/l4netlbcontroller.go
@@ -18,6 +18,7 @@ package l4netlb
 
 import (
 	"fmt"
+	"reflect"
 	"sync"
 	"time"
 
@@ -35,6 +36,7 @@ import (
 	"k8s.io/ingress-gce/pkg/loadbalancers"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/common"
+	utilslb "k8s.io/ingress-gce/pkg/utils/loadbalancers"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/klog"
@@ -104,7 +106,6 @@ func NewL4NetLBController(
 		},
 		// Deletes will be handled in the Update when the deletion timestamp is set.
 		UpdateFunc: func(old, cur interface{}) {
-			//TODO(kl52752) add implementation for update
 			curSvc := cur.(*v1.Service)
 			oldSvc := old.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(curSvc.Namespace, curSvc.Name)
@@ -112,6 +113,7 @@ func NewL4NetLBController(
 				klog.V(3).Infof("L4 External LoadBalancer Service %s updated, enqueuing", svcKey)
 				l4netLBc.svcQueue.Enqueue(curSvc)
 				l4netLBc.enqueueTracker.Track()
+				return
 			}
 		},
 	})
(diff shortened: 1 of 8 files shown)


pkg/loadbalancers/portslb.go
```diff
@@ -0,0 +1,73 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package loadbalancers
+
+import v1 "k8s.io/api/core/v1"
```
freehan, line 19: consider moving this file into `pkg/utils/loadbalancers/ports.go` as its a purely util function

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -104,14 +105,23 @@ func NewL4NetLBController(
...
 			svcKey := utils.ServiceKeyFunc(curSvc.Namespace, curSvc.Name)
 			if l4netLBc.shouldProcessService(curSvc, oldSvc) {
 				klog.V(3).Infof("L4 External LoadBalancer Service %s updated, enqueuing", svcKey)
 				l4netLBc.svcQueue.Enqueue(curSvc)
 				l4netLBc.enqueueTracker.Track()
+				return
+			}
+			// Enqueue NetLB services periodically for reasserting that resources exist.
+			needsNetLb, _ := annotations.WantsL4NetLB(curSvc)
+			if needsNetLb && reflect.DeepEqual(old, cur) {
```
freehan, line 119: can you merge this section with the above Enqueue check?

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -347,3 +415,9 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
 	lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeNormal, "DeletedLoadBalancer", "Deleted L4 External LoadBalancer")
 	return result
 }
+
+func (lc *L4NetLBController) hasBackendForwardingRule(svc *v1.Service) bool {
```
freehan, line 419: nit comment

freehan: is this function used?

pkg/l4netlb/l4netlbcontroller.go
```diff
@@ -347,3 +415,9 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
...
+	l4netlb := loadbalancers.NewL4NetLB(svc, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(svc.Namespace), &lc.sharedResourcesLock)
+	existingFR := l4netlb.GetForwardingRule(l4netlb.GetFRName(), meta.VersionGA)
```
freehan, line 421: this is very expensive, every sync will trigger a Get ForwardingRule API call. This eats up GCE API RPS unnecessarily. The same could be checked by validating the forwarding-rule annotation.

pkg/l4netlb/l4netlbcontroller.go
kl52752: No, this is a merge error now this function is renamed to hasLegacyForwardingRule, will remove

pkg/l4netlb/l4netlbcontroller.go
kl52752: This function was deleted but this logic is moved to `hasLegacyForwardingRule`.
Service Controller doesn't set forwarding rule annotation and here we need to distinguish between TP (legacy) and RBS based loadbalancers. I can check the annotation first but it will be only useful for Updates and Deletions (because it will be set by new controller) but for Addition I need to check forwarding rule target.

freehan: That is good enough. I want to avoid making GCE API calls for every service reconciliation.

pkg/utils/loadbalancers/ports.go
```diff
@@ -0,0 +1,73 @@
...
+	ports := []*v1.ServicePort{}
+	for i := range service.Spec.Ports {
+		sp := &service.Spec.Ports[i]
+		ports = append(ports, sp)
+	}
+	return ports
+}
+
+// PortsEqualForLBService checks if ports in loadbalancer are equal
+func PortsEqualForLBService(x, y *v1.Service) bool {
```
freehan, line 31: add unit testing for util functions

freehan (commented):
LGTM overall

pkg/utils/loadbalancers/ports.go
kl52752: done

PR #1613, opened 2021-11-18
Update PSC controller to use V1 GCE Service Attachment API

swetharepakula:
Updates k8s-cloud-provider to v1.18.0

/assign @freehan @prameshj


bowei (approved):
/lgtm

PR #1618, opened 2021-11-24
Clean up health check on service deletion

kl52752:
When service ExternalTrafficPolicy is updated a new helath check is
created. In update function we don't try to delete old health check due
to performance reason. When service is deleted we need to check both
shared and non-shared health checks to prevent resource leak.

### DIFF
--- a/pkg/l4/l4controller_test.go
+++ b/pkg/l4/l4controller_test.go
@@ -23,6 +23,9 @@ import (
 	"time"
 
 	"k8s.io/ingress-gce/pkg/loadbalancers"
+	"k8s.io/ingress-gce/pkg/utils"
+
+	"net/http"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -31,7 +34,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/util/retry"
-	"k8s.io/cloud-provider"
+	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/context"
@@ -39,7 +42,6 @@ import (
 	"k8s.io/ingress-gce/pkg/utils/common"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/legacy-cloud-providers/gce"
-	"net/http"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/l4/l4controller_test.go
```diff
@@ -225,6 +227,23 @@ func TestProcessCreateOrUpdate(t *testing.T) {
...
+	}
+	for _, isShared := range []bool{true, false} {
+		hcName, _ := l4c.namer.L4HealthCheck(newSvc.Namespace, newSvc.Name, isShared)
+		if !isHealthCheckDeleted(l4c, hcName) {
+			t.Errorf("Health check %s should be deleted", hcName)
+		}
+	}
+}
+
+func isHealthCheckDeleted(l4c *L4Controller, hcName string) bool {
```
prameshj, line 244: Can you modify the function here - https://github.com/kubernetes/ingress-gce/blob/b7961f5ababe9473fe9c3f7f62b297b730bc0151/pkg/loadbalancers/l4_test.go#L1260 to check for both healthchecks?

pkg/l4/l4controller_test.go
```diff
@@ -225,6 +227,23 @@ func TestProcessCreateOrUpdate(t *testing.T) {
 	}
 	validateSvcStatus(newSvc, false, t)
 	prevMetrics.ValidateDiff(test.GetL4LatencyMetric(t), &test.L4ILBLatencyMetricInfo{CreateCount: 1, UpdateCount: 1, DeleteCount: 1, UpperBoundSeconds: 1}, t)
+	newSvc.DeletionTimestamp = &v1.Time{}
```
prameshj, line 230: This test changes from Cluster to Local traffic policy, so it doesn't repro the leak. Can we have a separate test here that 1) creates lb with Local policy 2) changes to cluster policy 3) verifies that HC is leaked 4) deletes service 5) verifies healthcheck is cleaned up?

prameshj (commented):
Changes look good.. one suggestion for unit test.

pkg/l4/l4controller_test.go
kl52752: done

pkg/l4/l4controller_test.go
kl52752: done

pkg/loadbalancers/l4.go
```diff
@@ -169,21 +169,42 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
+			if !utils.IsInUsedByError(err) {
+				klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s - %v", l.NamespacedName.String(), err)
+				return err
+			}
+			// Ignore deletion error due to health check in use by another resource.
+			// This will be hit if this is a shared healthcheck.
+			klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
+		} else {
+			// We don't need firewall rule any more so delete it.
+			err = deleteFunc(hcFwName)
```
prameshj, line 194: You can remove the firewall rule deletion in line 157. This new logic you added is the correct way to delete it.

pkg/loadbalancers/l4.go
```diff
@@ -169,21 +169,42 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
+		err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l.cloud, hcName, meta.Global))
+		if err != nil {
+			if !utils.IsInUsedByError(err) {
+				klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s - %v", l.NamespacedName.String(), err)
+				return err
+			}
+			// Ignore deletion error due to health check in use by another resource.
+			// This will be hit if this is a shared healthcheck.
+			klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
+		} else {
```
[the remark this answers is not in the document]
kl52752: done, in line 208 I delete only fw rule for non-shared HC and fw rule for shared hc is deleted after hc deletion.

pkg/loadbalancers/l4.go
kl52752: done, in line 157 I delete only fw rule for non-shared HC and fw rule for shared hc is deleted after hc deletion.

pkg/loadbalancers/l4.go
```diff
@@ -152,28 +151,40 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
-		klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
-	} else {
-		// Delete healthcheck firewall rule if healthcheck deletion is successful.
-		err = deleteFwFunc(hcFwName)
+		err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l.cloud, hcName, meta.Global))
 		if err != nil {
-			klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, l.NamespacedName.String(), err)
-			result.GCEResourceInError = annotations.FirewallForHealthcheckResource
+			if !utils.IsInUsedByError(err) {
+				klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s - %v", l.NamespacedName.String(), err)
```
prameshj, line 168: How about setting `result.GCEResourceInError = annotations.HealthcheckResource` here and just return? 

So `deleteHcFunc := func(sharedHC bool) {`

Right now, we are returning nil even upon error in deleting firewall rule so that result.GCEResourceInError  is set correctly. Removing the return value might be clearer.

pkg/loadbalancers/l4netlb.go
```diff
@@ -205,43 +205,58 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
 		result.GCEResourceInError = annotations.FirewallRuleResource
 		result.Error = err
 	}
-
+	// Delete fw rules for non-shared health check only.
```
prameshj, line 208: This block should be removed right?

pkg/loadbalancers/l4netlb.go
```diff
@@ -205,43 +205,58 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
...
-			klog.Errorf("Failed to delete healthcheck for service %s - %v", l4netlb.NamespacedName.String(), err)
-			result.GCEResourceInError = annotations.HealthcheckResource
-			result.Error = err
-			return result
+	// We don't delete health check during service update so
+	// it is possible that there might be some health check leak
+	// when service TrafficPolicyLocal is updated and new a health check was created.
+	// When service is deleted we need to check both health checks shared and non-shared
+	// and delete them if needed.
+	deleteHcFunc := func(sharedHC bool) error {
```
prameshj, line 231: same comment as in l4.go - `deleteHcFunc := func(sharedHC bool) {`

prameshj:
Couple of small comments, otherwise lgtm.

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/loadbalancers/l4.go
```diff
@@ -152,31 +151,41 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
-	err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l.cloud, hcName, meta.Global))
-	if err != nil {
-		if !utils.IsInUsedByError(err) {
-			klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s, err %v", l.NamespacedName.String(), err)
-			result.GCEResourceInError = annotations.HealthcheckResource
-			result.Error = err
-			return result
+	// We don't delete health check during service update so
+	// it is possible that there might be some health check leak
+	// when service TrafficPolicyLocal is updated and a new health check was created.
```
prameshj, line 156: nit: Change "when service TrafficPolicyLocal is updated" to "when externalTrafficPolicy is changed from Cluster to Local"

pkg/loadbalancers/l4netlb.go
```diff
@@ -205,46 +205,58 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
 		result.GCEResourceInError = annotations.FirewallRuleResource
 		result.Error = err
 	}
-
+	// Shared Health check might be used by other service
```
prameshj, line 208: This block still shows up in the diff.. Should be removed?

pkg/loadbalancers/l4netlb.go
```diff
@@ -205,46 +205,58 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
...
-	err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l4netlb.cloud, hcName, l4netlb.scope))
-	if err != nil {
-		if !utils.IsInUsedByError(err) {
-			klog.Errorf("Failed to delete healthcheck for service %s - %v", l4netlb.NamespacedName.String(), err)
-			result.GCEResourceInError = annotations.HealthcheckResource
-			result.Error = err
-			return result
+	// We don't delete health check during service update so
+	// it is possible that there might be some health check leak
+	// when service TrafficPolicyLocal is updated and new a health check was created.
```
prameshj, line 227: Same nit comment as in l4.go

pkg/loadbalancers/l4netlb.go
```diff
@@ -205,46 +205,58 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
...
+				result.GCEResourceInError = annotations.HealthcheckResource
+				result.Error = err
+				return
+			}
+			// Ignore deletion error due to health check in use by another resource.
+			// This will be hit if this is a shared healthcheck.
+			klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
+		} else {
+			// Delete healthcheck firewall rule if healthcheck deletion is successful.
+			err = deleteFirewallFunc(hcFwName)
```
prameshj, line 249: optional - make the name deleteFwFunc to be consistent with the ILB source code.

prameshj:
Couple of nits + 1 code block which should be removed. I can lgtm once that is addressed.

pkg/loadbalancers/l4.go
kl52752: Shouldn't it be the "from Local to Cluster"? I thought that the problem is when non-shared (it's Local) health check stays unused. For shared clusters it's not a big deal because there is a chance that some other process use it.

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: done as above

pkg/loadbalancers/l4.go
prameshj: You are right, it is Local to Cluster.

PR #1619, opened 2021-11-24
Add back original CDN test for older Ingress Versions

swetharepakula:
* older ingress versions will be unable to use the new expanded config
   options and cannot pass the new tests

/assign @prameshj @bowei

### DIFF
--- a/cmd/e2e-test/cdn_test.go
+++ b/cmd/e2e-test/cdn_test.go
@@ -37,7 +37,108 @@ import (
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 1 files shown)


prameshj:
TestCDN test has been reverted to the version in https://github.com/kubernetes/ingress-gce/blob/d06cb4252a492076571425106eb95df806553cd2/cmd/e2e-test/cdn_test.go#L37 with all ExpandedCDN functionality being tested in tests of the form "TestExpandedCDN*".

This looks good to me.

/lgtm
/approve

PR #1625, opened 2021-11-30
Fix firewall rule deletion in L4 LBs

kl52752:
When L4 LoadBalancer is deleted (Internal or External) firewall rule was deleted upfront.
But when ExternalTrafficPolicy is set to Cluster health check is shared
among all ILB or XLB services and removing firewall rule will make those
services unreachable. So in service deletion we need to check if health
check is not in use by another service,

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -132,9 +132,8 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult 
 		result.Error = err
 		result.GCEResourceInError = annotations.AddressResource
 	}
-	hcName, hcFwName := l.namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	// delete fw rules
-	deleteFunc := func(name string) error {
+	deleteFwFunc := func(name string) error {
 		err := firewalls.EnsureL4FirewallRuleDeleted(l.cloud, name)
 		if err != nil {
 			if fwErr, ok := err.(*firewalls.FirewallXPNError); ok {
@@ -146,21 +145,13 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult 
 		return nil
 	}
 	// delete firewall rule allowing load balancer source ranges
-	err = deleteFunc(name)
+	err = deleteFwFunc(name)
 	if err != nil {
 		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", name, l.NamespacedName.String(), err)
 		result.GCEResourceInError = annotations.FirewallRuleResource
 		result.Error = err
 	}
-
-	// delete firewall rule allowing healthcheck source ranges
-	err = deleteFunc(hcFwName)
-	if err != nil {
-		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, l.NamespacedName.String(), err)
-		result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-		result.Error = err
-	}
-	// delete backend service
+	// Delete backend service
 	err = utils.IgnoreHTTPNotFound(l.backendPool.Delete(name, meta.VersionGA, meta.Regional))
 	if err != nil {
 		klog.Errorf("Failed to delete backends for internal loadbalancer service %s, err  %v", l.NamespacedName.String(), err)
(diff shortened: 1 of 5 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -184,6 +187,14 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
 		// Ignore deletion error due to health check in use by another resource.
 		// This will be hit if this is a shared healthcheck.
 		klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
+	} else if sharedHC {
```
prameshj, line 190: We can make this just "} else {"

Can we change the logic slightly so that the healthcheck firewall rule is always deleted after healthcheck is deleted?

That way, you can get rid of the additions in lines 155 to 165.

pkg/loadbalancers/l4netlb.go
```diff
@@ -200,14 +200,18 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *SyncResu
...
 		result.Error = err
 	}
-
-	// delete firewall rule allowing healthcheck source ranges
-	err = deleteFirewallFunc(hcFwName)
-	if err != nil {
-		klog.Errorf("Failed to delete firewall rule %s for service %s - %v", hcFwName, l4netlb.NamespacedName.String(), err)
-		result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-		result.Error = err
+	if !sharedHC {
```
prameshj, line 203: Same comment as in l4.go. We can move hc fw deletion to the end.

pkg/loadbalancers/l4.go
```diff
@@ -152,15 +152,18 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
-	err = deleteFunc(hcFwName)
-	if err != nil {
-		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, l.NamespacedName.String(), err)
-		result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-		result.Error = err
+	if !sharedHC {
+		// Delete firewall rule allowing healthcheck source ranges.
+		// Here we delete only firewall rule for non-shared health checks because we own the HC.
+		// For shared HC we need to check if resource is not used by another service.
+		err = deleteFunc(hcFwName)
```
prameshj, line 159: In line 137, can we rename this to deleteFwFunc()?

pkg/loadbalancers/l4_test.go
```diff
@@ -539,6 +539,55 @@ func TestEnsureInternalLoadBalancerDeletedTwiceDoesNotError(t *testing.T) {
...
+	_, _, result := ensureService(fakeGCE, namer, nodeNames, vals.ZoneName, 8080, t)
+	if result != nil && result.Error != nil {
+		t.Fatalf("Error ensuring service err: %v", result.Error)
+	}
+	svc2, l, result := ensureService(fakeGCE, namer, nodeNames, vals.ZoneName, 8081, t)
+	if result != nil && result.Error != nil {
+		t.Fatalf("Error ensuring service err: %v", result.Error)
+	}
+
+	// Delete the loadbalancer
```
prameshj, line 560: super nit: add "." at the end of comments.

prameshj (commented):
Thanks for fixing this and adding a comprehensive test case!

pkg/loadbalancers/l4.go
```diff
@@ -152,15 +152,18 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *SyncResult
...
 		result.Error = err
 	}
-
-	// delete firewall rule allowing healthcheck source ranges
-	err = deleteFunc(hcFwName)
-	if err != nil {
-		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, l.NamespacedName.String(), err)
-		result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-		result.Error = err
+	if !sharedHC {
```
prameshj, line 155: It would be great to cherrypick the l4.go + unit test changes into 1.14 and 1.13 branches as well.

pkg/loadbalancers/l4.go
kl52752: done

pkg/loadbalancers/l4.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/loadbalancers/l4_test.go
kl52752: done

PR #1626, opened 2021-11-30
Firewall CR migration

sugangli:
This PR is for migrating existing direct GCE FW configuration to our firewall CR approach.  Only L7 firewall changes are included here. L4 changes will come later.  Major changes:

1. Most of the new CR-related functions are under the second commit. Firewall CRUDs are supported here.
2. Add firewall informers so that the ingress-gce can watch the status and populate to the related LB services/ingresses. It is mainly for populating the XPN error and its relevant gcloud cmd to the service/ingress's event, but it also serves general service updating purpose. 

Here is the algorithm in high-level:
-  ingress-gce creates the CR with pending status
- Platform Firewall Controller(borg) configures the firewall based on the CRs, and updates the status
- ingress-gce watches the status, and enqueue in the service/ingress if needed. Then it uses the service/ingress ensuring code path to ensure(read) the firewall CR status and populate the event.

3. Two flags are added : EnableFirewallCR and EnableFWControllerEnforcement. When the latter is false, ingress-gce configures firewall CRs and gce firewalls at the same time, but Platform FW Controller(borg) will not enforce the CRs. When EnableFWControllerEnforcement is true, we simply drop the direct GCE firewall configurations. We use cluster proto as the source of truth to turn this flag on/off.


pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
 		},
 	})
 
+	ctx.FirewallInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			addFW := obj.(*v1beta1.GCPFirewall)
+			for _, ing := range ctx.Ingresses().List() {
+				name := fwc.ctx.ClusterNamer.FirewallRule()
+				if name == addFW.Name {
```
skmatti, line 129: Is it better to move this before the for loop?.

```
name := fwc.ctx.ClusterNamer.FirewallRule()
if name != addFW.Name {
 return
}

for _, ing := range ctx.Ingresses().List() {
  fwc.ctx.Recorder(...)
}
fwc.queue.Enqueue(queueKey)
```

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
...
+					fwc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeNormal, events.SyncIngress, FirewallReasonAdded, addFW.Name)
+				}
+			}
+		},
+		UpdateFunc: func(old, cur interface{}) {
+			curFW := cur.(*v1beta1.GCPFirewall)
+			foundXPN := false
+			for _, ing := range ctx.Ingresses().List() {
+				name := fwc.ctx.ClusterNamer.FirewallRule()
+				if name == curFW.Name {
```
skmatti, line 139: Same here, move this logic outside the for loop?.

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
...
+						// Enqueue again if XPN permission error is presented in order to populate the gcloud cmd to ALL the ingresses
+						if con.Reason == string(v1beta1.FirewallRuleReasonXPNPermissionError) {
+							foundXPN = true
+						}
+						fwc.ctx.Recorder(ing.Namespace).Eventf(ing, apiv1.EventTypeNormal, events.SyncIngress, FirewallReasonUpdated,
+							curFW.Name, con.Status, con.Reason)
+					}
+				}
+			}
+			if foundXPN {
```
skmatti, line 150: Don't we need to requeue for all the cases?. 

What if the GCPFirewall Spec changes?. We need to requeue then too?.

pkg/firewalls/firewalls.go
```diff
@@ -82,11 +93,12 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 		ports.Insert(fr.nodePortRanges...)
 	}
 	ports.Insert(additionalPorts...)
+	klog.V(3).Infof("ports %v", ports)
```
skmatti, line 96: Intended to remove this debug line?

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+		klog.V(3).Infof("ensureFirewallCR Create CR :%v", expectedFW)
+		_, err = fw.Create(context.Background(), expectedFW, metav1.CreateOptions{})
+		return err
+	}
+	for _, con := range currentFW.Status.Conditions {
+		if con.Reason == string(v1beta1.FirewallRuleReasonXPNPermissionError) ||
+			con.Reason == string(v1beta1.FirewallRuleReasonInvalid) ||
+			con.Reason == string(v1beta1.FirewallRuleReasonSyncError) {
+			// Use recorder to log the cmd
+			klog.V(3).Infof("ensureFirwallCR(%v): Could not enforce Firewall CR:%v", currentFW.Name, con.Reason)
```
skmatti, line 166: ensureFirewallCR

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	firewallCR.Spec.Ingress = &v1beta1.GCPFirewallIngress{
+		Source: &v1beta1.IngressSource{
+			IPBlocks: src_cidrs,
+		},
+		Destination: &v1beta1.IngressDestination{
+			IPBlocks: dst_cidrs,
+		},
+	}
+
+	firewallCR.Status.Conditions = []metav1.Condition{
```
skmatti, line 244: Setting status on creation is not allowed. If we want do this, this needs to be done in API definition using kubebuilder default value - https://github.com/kubernetes/cloud-provider-gcp/blob/43757f3f46ae41820665c462aca8a67068de26fa/crd/apis/gcpfirewall/v1beta1/types.go#L180

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	return fr.updateFirewall(expectedFirewall)
 }
 
+//ensureFirewallCR creates/updates the firewall CR
```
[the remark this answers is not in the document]
sugangli: In controller.go updateFunc, we don't have the context of the FW parameter so we cannot populate the gcloud cmd accordingly. Therefore, we use that as a trigger so that the ensureFirewallCR can read the CR.

pkg/firewalls/firewalls.go
sugangli: Removed.

pkg/firewalls/controller.go
sugangli: Updated.

pkg/firewalls/controller.go
sugangli: Updated.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+func ensureFirewallCR(client firewallclient.Interface, expectedFW *v1beta1.GCPFirewall) error {
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	currentFW, err := fw.Get(context.Background(), expectedFW.Name, metav1.GetOptions{})
+
+	klog.V(3).Infof("ensureFirewallCR Get CR :%v", currentFW)
+
+	if currentFW == nil || err != nil {
```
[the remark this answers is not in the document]
sugangli: Updated.

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
 		},
 	})
 
+	ctx.FirewallInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
sugangli: Updated.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,113 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	}
+
+	return firewallCR, nil
+}
+
 // GC deletes the firewall rule.
 func (fr *FirewallRules) GC() error {
 	name := fr.namer.FirewallRule()
 	klog.V(3).Infof("Deleting firewall %q", name)
+	err := deleteFirewallCR(fr.firewallClient, name)
```
rahulkjoshi, line 252: Only `if flags.F.EnableFirewallCR`

pkg/firewalls/firewalls.go
```diff
@@ -100,6 +112,24 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+		klog.V(3).Infof("Firewall CR is enabled.")
+		expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{})
+		if err != nil {
+			return err
+		}
+		err = ensureFirewallCR(fr.firewallClient, expectedFirewallCR)
+		if err != nil && strings.Contains(err.Error(), string(v1beta1.FirewallRuleReasonXPNPermissionError)) {
+			gcloudCmd := gce.FirewallToGCloudCreateCmd(expectedFirewall, fr.cloud.NetworkProjectID())
+			klog.V(3).Infof("Could not create L7 firewall on XPN cluster: %v. Raising event for cmd: %q", err, gcloudCmd)
+			return newFirewallXPNError(err, gcloudCmd)
```
rahulkjoshi, line 125: Also only return if enforcement is enabled right?

pkg/firewalls/firewalls.go
```diff
@@ -100,6 +112,24 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 		},
 		TargetTags: targetTags,
 	}
+	if flags.F.EnableFirewallCR {
+		klog.V(3).Infof("Firewall CR is enabled.")
+		expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{})
+		if err != nil {
+			return err
+		}
+		err = ensureFirewallCR(fr.firewallClient, expectedFirewallCR)
```
rahulkjoshi, line 121: Add a comment that the first time ensure is called (i.e. when we're first making a CR) it will always succeed, even if there's a XPN issue.

On a subsequent sync loop, we'll see the status change and then raise the error up the stack

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
...
+		// We enqueue the sync task for all update events for these two reasons:
+		// 1. Make sure the CR is consistent with the FW configuration.
+		// 2. Force the controller to read the CR status and populate proper errors.
+		UpdateFunc: func(old, cur interface{}) {
+			curFW := cur.(*v1beta1.GCPFirewall)
+			name := fwc.ctx.ClusterNamer.FirewallRule()
+			if name != curFW.Name {
+				return
+			}
+			for _, ing := range ctx.Ingresses().List() {
```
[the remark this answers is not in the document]
sugangli: Discussed offline. To be consistent with the original GCE FW configuration, we will only emit XPN errors event for every ingress.

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
 		},
 	})
 
+	// FW CRs will be updated/deleted by the PFW controller or the user. Ingress controller need to watch such events
+	// and act accordingly.
+	ctx.FirewallInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
sugangli: Removed.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,113 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	firewallCR.Spec.Action = v1beta1.ActionAllow
+	firewallCR.Spec.Disabled = !flags.F.EnableFWControllerEnforcement
+	var protocolPorts []v1beta1.ProtocolPort
+
+	for _, item := range ports {
+		s := strings.Split(item, "-")
+		var protocolPort v1beta1.ProtocolPort
+		startport, err := strconv.Atoi(s[0])
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,113 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+		}
+		sp32 := int32(startport)
+		if len(s) > 1 {
+			endport, err := strconv.Atoi(s[1])
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,113 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	return fr.updateFirewall(expectedFirewall)
 }
 
+//ensureFirewallCR creates/updates the firewall CR
+func ensureFirewallCR(client firewallclient.Interface, expectedFW *v1beta1.GCPFirewall) error {
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	currentFW, err := fw.Get(context.Background(), expectedFW.Name, metav1.GetOptions{})
+
+	klog.V(3).Infof("ensureFirewallCR Get CR :%v", currentFW)
```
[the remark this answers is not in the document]
sugangli: Changed tp `%+v`.  Don't see that util, but we probably can remove those logs before we release it.

pkg/firewalls/controller.go
```diff
@@ -119,6 +121,43 @@ func NewFirewallController(
...
+		AddFunc: func(obj interface{}) {
+			addFW := obj.(*v1beta1.GCPFirewall)
+			name := fwc.ctx.ClusterNamer.FirewallRule()
+			if name != addFW.Name {
```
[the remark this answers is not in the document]
sugangli: Removed.

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,113 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+func deleteFirewallCR(client firewallclient.Interface, name string) error {
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	klog.V(3).Infof("Delete CR :%v", name)
+	return fw.Delete(context.Background(), name, metav1.DeleteOptions{})
+}
+
+//NewFirewallCR constructs the firewall CR from name, ports and ranges
+func NewFirewallCR(name string, ports, srcRanges, dstRanges []string) (*v1beta1.GCPFirewall, error) {
+	firewallCR := &v1beta1.GCPFirewall{}
+	firewallCR.Name = name
```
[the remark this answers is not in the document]
sugangli: Updated.

sugangli:
@cezarygerard, @panslava, could you please take a look?

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+		}
+	}
+
+	if !reflect.DeepEqual(currentFW.Spec, expectedFW.Spec) { //Update the current firewall CR
+		klog.V(3).Infof("ensureFirewallCR Update CR currentFW.Spec: %+v, expectedFW.Spec: %+v", currentFW.Spec, expectedFW.Spec)
+		currentFW.Spec = expectedFW.Spec
+		_, err = fw.Update(context.Background(), currentFW, metav1.UpdateOptions{})
+		return err
+	}
+	return err
```
aojea, line 181: return nil? 

I think that is clear if we are explicit or we may get here with some edge case in the future

pkg/firewalls/firewalls.go
```diff
@@ -116,10 +146,121 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+}
+
 // GC deletes the firewall rule.
 func (fr *FirewallRules) GC() error {
 	name := fr.namer.FirewallRule()
 	klog.V(3).Infof("Deleting firewall %q", name)
+	if flags.F.EnableFirewallCR {
+		err := deleteFirewallCR(fr.firewallClient, name)
+		if flags.F.EnableFWControllerEnforcement {
+			return err
```
aojea, line 261: if you return here and this is a cluster that got upgraded, and there is still old firewall rules ... should this not delete these old rules too?

sugangli: Good point. In fact, for the upgraded clusters, we will have both the firewall CR and the firewall rules, so we need to GC both of them. I have added a GCCR() to handle CRs separately.

cmd/glbc/main.go
```diff
@@ -133,6 +134,11 @@ func main() {
 		}
 	}
 
+	firewallClient, err := firewallclient.NewForConfig(kubeConfig)
```
aojea, line 137: I can see that the other CRDs clients try to ensure that the CRD exist,  do we need that too?

sugangli: No, we don't. The CRD will be ensured by the platform firewall controller (a borg job),  as it will be shared by multiple entities, like ingress controller, service controller,  cluster API server, etc.

aojea:
/hold 

we need to clarify first the CRDs source https://github.com/kubernetes/cloud-provider-gcp/issues/431

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +42,55 @@ func portRanges() []string {
 
 func TestFirewallPoolSync(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges())
+	flags.F.EnableFirewallCR = true
```
swetharepakula, line 45: can we test both when the flag is true and false. In the false cases, we should make sure that the CR is not created.

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +42,55 @@ func portRanges() []string {
...
 		t.Fatal(err)
 	}
-	verifyFirewallRule(fwp, ruleName, nodes, srcRanges, portRanges(), t)
+	verifyFirewallRule(fwp, simpleclient, ruleName, nodes, srcRanges, portRanges(), t)
 }
 
 func TestFirewallPoolSyncNodes(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges())
+	flags.F.EnableFirewallCR = true
```
swetharepakula, line 57: same comment as above.

pkg/firewalls/firewalls.go
```diff
@@ -100,6 +112,24 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 		},
 		TargetTags: targetTags,
 	}
+	if flags.F.EnableFirewallCR {
```
swetharepakula, line 115: Instead of using the flag here, set a bool on the pool and use that instead of the global variable.

pkg/firewalls/controller.go
```diff
@@ -119,6 +120,29 @@ func NewFirewallController(
 		},
 	})
 
+	if flags.F.EnableFirewallCR {
```
swetharepakula, line 123: Add this as a bool to the controller and use that instead.

swetharepakula (commented):
To reduce the dependency, I am wondering if we should instead use a dynamic client, especially since this controller will not be ensuring the Firewall CRD.

At this point. 1.21 is old. If we are updating dependencies perhaps we should update to something newer. This perhaps can be done in a followup.

pkg/firewalls/firewalls_test.go
```diff
@@ -279,7 +298,46 @@ func validateXPNError(err error, op string, t *testing.T) {
 	}
 }
 
-func verifyFirewallRule(fwp *fakeFirewallsProvider, ruleName string, expectedNodes, expectedCIDRs, expectedPorts []string, t *testing.T) {
+func verifyFirewallRule(fwp *fakeFirewallsProvider, firewallclient *firewallclient.Clientset, ruleName string, expectedNodes, sourceRanges, expectedPorts []string, t *testing.T) {
+	// Verify if the firewall CR was created
+	if flags.F.EnableFirewallCR {
+		fw := firewallclient.NetworkingV1beta1().GCPFirewalls()
```
cezarygerard, line 304: can you move it to separate func?

(I mean everything inside this  `if flags.F.EnableFirewallCR {` )

pkg/firewalls/controller.go
cezarygerard: +1

pkg/firewalls/firewalls_test.go
```diff
@@ -279,7 +298,46 @@ func validateXPNError(err error, op string, t *testing.T) {
...
+		actualSrcRanges := sets.NewString()
+		for _, ipblock := range actualFW.Spec.Ingress.Source.IPBlocks {
+			actualSrcRanges.Insert(string(ipblock))
+		}
+
+		if !actualSrcRanges.Equal(srcranges) {
+			t.Errorf("actual SrcRanges(%v) does not equal to expected SrcRanges(%v)", actualSrcRanges, srcranges)
+		}
+	}
 	// Verify firewall rule was created
```
cezarygerard, line 341: do you still validate firewall rule even if EnableFirewallCR was true?

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	return fr.updateFirewall(expectedFirewall)
 }
 
+// ensureFirewallCR creates/updates the firewall CR
+// On CR update, it will read the conditions to see if there are errors updated by PFW controller.
+// If the Spec was updated by others, it will reconcile the Spec.
+func ensureFirewallCR(client firewallclient.Interface, expectedFW *v1beta1.GCPFirewall) error {
```
cezarygerard, line 152: Rename expectedFW to expectedFWCR

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	currentFW, err := fw.Get(context.Background(), expectedFW.Name, metav1.GetOptions{})
```
cezarygerard, line 154: currentFW to currentFWCR

pkg/firewalls/interfaces.go
```diff
@@ -25,6 +25,7 @@ type SingleFirewallPool interface {
 	// Sync syncs firewall rules with the cloud
 	Sync(nodeNames, additionalPorts, additionalRanges []string, allowNodePort bool) error
 	GC() error
+	GCCR() error
```
cezarygerard, line 28: Extending the existing firewalls.go code add a lot of complexity

Why not just add new implementation of SingleFirewallPool
that will do all the  firewallCR management?

this would simplyfy the code in so many places

the controller would pick it's firewall Pool based on flag(or corresponding bool arg) during the creation time, and you will never have to alter the logic based on the flag

I think such approach would be much cleaner even if the 2 implementations would have some common code

pkg/flags/flags.go
```diff
@@ -257,6 +259,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableMultipleIGs, "enable-multiple-igs", false, "Enable using multiple unmanaged instance groups")
 	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
+	flag.BoolVar(&F.EnableFirewallCR, "enable-firewall-cr", false, "Enable generating firewall CR")
+	flag.BoolVar(&F.EnableFWControllerEnforcement, "enable-fw-controller-enforcement", false, "Enable the GCP firewall controller to enforce the CRs")
```
cezarygerard, line 263: Can you extend the comment there

does enable-fw-controller-enforcement mean that ingress does not create any firewalls on GCP, only the firewallCRs?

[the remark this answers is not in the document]
sugangli: I will add that to the description. In practice, EnableFWControllerEnforcement check is always under EnableFirewallCR = true, and we don't expect users play with these flags. The whole point of having two flags is for testing and safe rollout.

pkg/firewalls/firewalls.go
```diff
@@ -100,6 +112,24 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
 		TargetTags: targetTags,
 	}
+	if flags.F.EnableFirewallCR {
+		klog.V(3).Infof("Firewall CR is enabled.")
+		expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{})
+		if err != nil {
+			return err
+		}
+		err = ensureFirewallCR(fr.firewallClient, expectedFirewallCR)
+		if err != nil && strings.Contains(err.Error(), string(v1beta1.FirewallRuleReasonXPNPermissionError)) {
```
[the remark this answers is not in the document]
sugangli: We could.  But either we convert string(v1beta1.FirewallRuleReasonXPNPermissionError) to an error, or convert err to a string. Does it make a difference?

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(3).Infof("ensureFirewallCR Create CR :%+v", expectedFW)
+			_, err = fw.Create(context.Background(), expectedFW, metav1.CreateOptions{})
+		}
+		return err
+	}
+	for _, con := range currentFW.Status.Conditions {
+		if con.Reason == string(v1beta1.FirewallRuleReasonXPNPermissionError) ||
```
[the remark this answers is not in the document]
sugangli: At this point, we only have one type(https://github.com/kubernetes/cloud-provider-gcp/blob/master/crd/apis/gcpfirewall/v1beta1/types.go#L192). But the reason may not be empty(https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition),  so it should be safe to check Reason directly.

vendor/modules.txt
```diff
@@ -408,8 +408,8 @@ k8s.io/api/scheduling/v1beta1
 k8s.io/api/storage/v1
 k8s.io/api/storage/v1alpha1
 k8s.io/api/storage/v1beta1
-# k8s.io/apiextensions-apiserver v0.20.0
-## explicit; go 1.15
+# k8s.io/apiextensions-apiserver v0.21.1
```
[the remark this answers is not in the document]
sugangli: Agree. We could do it in a separate PR.

pkg/firewalls/firewalls.go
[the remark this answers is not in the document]
sugangli: Made the change.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	currentFW, err := fw.Get(context.Background(), expectedFW.Name, metav1.GetOptions{})
+
+	klog.V(3).Infof("ensureFirewallCR Get CR :%+v", currentFW)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(3).Infof("ensureFirewallCR Create CR :%+v", expectedFW)
+			_, err = fw.Create(context.Background(), expectedFW, metav1.CreateOptions{})
+		}
+		return err
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+		if con.Reason == string(v1beta1.FirewallRuleReasonXPNPermissionError) ||
+			con.Reason == string(v1beta1.FirewallRuleReasonInvalid) ||
+			con.Reason == string(v1beta1.FirewallRuleReasonSyncError) {
+			// Use recorder to emit the cmd in Sync()
+			klog.V(3).Infof("ensureFirewallCR(%v): Could not enforce Firewall CR Reason:%v", currentFW.Name, con.Reason)
+			return fmt.Errorf(con.Reason)
+		}
+	}
+
+	if !reflect.DeepEqual(currentFW.Spec, expectedFW.Spec) { //Update the current firewall CR
```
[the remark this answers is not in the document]
sugangli: Good point. Updated.

pkg/firewalls/firewalls_test.go
sugangli: Added a negative case.

pkg/firewalls/firewalls_test.go
sugangli: We only need to test the false case once, right? False cases are the same, i.e., no CR created.

pkg/firewalls/firewalls.go
sugangli: Updated. Passed both flags as arguments to the constructors.

pkg/firewalls/controller.go
sugangli: Updated.

pkg/firewalls/firewalls.go
```diff
@@ -17,12 +17,20 @@ limitations under the License.
...
+	"strconv"
 	"strings"
 
 	"google.golang.org/api/compute/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/cloud-provider-gcp/crd/apis/gcpfirewall/v1beta1"
+	firewallclient "k8s.io/cloud-provider-gcp/crd/client/gcpfirewall/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/flags"
```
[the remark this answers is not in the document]
sugangli: Updated.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+
+//deleteFirewallCR deletes the firewall CR
+func deleteFirewallCR(client firewallclient.Interface, name string) error {
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	klog.V(3).Infof("Delete CR :%v", name)
+	return fw.Delete(context.Background(), name, metav1.DeleteOptions{})
+}
+
+//NewFirewallCR constructs the firewall CR from name, ports and ranges
+func NewFirewallCR(name string, ports, srcRanges, dstRanges []string) (*v1beta1.GCPFirewall, error) {
```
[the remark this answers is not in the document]
sugangli: In the folllowing PR, l4 firewall will need to call it in a different package.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +146,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
 func (fr *FirewallRules) GC() error {
 	name := fr.namer.FirewallRule()
 	klog.V(3).Infof("Deleting firewall %q", name)
 	return fr.deleteFirewall(name)
 }
 
+// GCCR deletes the firewall CR
+// For the upgraded clusters with EnableFirewallCR = true, the firewall CR and the firewall co-exist.
+// We need to delete both of them every time.
+func (fr *FirewallRules) GCCR() error {
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls_test.go
sugangli: Good point! I moved it before I saw this comment.

pkg/firewalls/firewalls_test.go
sugangli: Yes, we do. They are in separated functions now.

pkg/firewalls/firewalls.go
sugangli: Done.

pkg/firewalls/firewalls.go
sugangli: Done.

pkg/firewalls/interfaces.go
sugangli: I agree that it might look cleaner if they are separate. However, considering firewall rules config and CRs config share lots of common logics(create, update, delete, etc) and parameters, a new implementation might introduce lots of duplicated codes. The major complexity comes from the EnableFWControllerEnforcement and XPNPermissionError handling.  Other than those, two flows looks identical.

pkg/flags/flags.go
sugangli: Yes.  Extended.

aojea:
> /hold
> 
> we need to clarify first the CRDs source [kubernetes/cloud-provider-gcp#431](https://github.com/kubernetes/cloud-provider-gcp/issues/431)

/hold cancel

this was resolved, and the conclusion is that it will keep be working as today

pkg/firewalls/firewalls_test.go
swetharepakula: that's right

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +41,65 @@ func portRanges() []string {
 
 func TestFirewallPoolSync(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges())
+	simpleclient := firewallclient.NewSimpleClientset()
```
swetharepakula, line 44: nit: rename to fwClient to make it easier to follow

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +41,65 @@ func portRanges() []string {
...
 		t.Fatal(err)
 	}
 	verifyFirewallRule(fwp, ruleName, nodes, srcRanges, portRanges(), t)
+	// verify no CRs are generated.
+	verifyFirewallCR(simpleclient, ruleName, srcRanges, portRanges(), false, t)
+
+	fp = NewFirewallPool(simpleclient, fwp, defaultNamer, srcRanges, portRanges(), true, false)
+	if err := fp.Sync(nodes, nil, nil, true); err != nil {
+		t.Fatal(err)
+	}
```
swetharepakula, line 57: should we verifyFirewallRule after the sync?

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +41,65 @@ func portRanges() []string {
...
+	verifyFirewallCR(simpleclient, ruleName, srcRanges, portRanges(), true, t)
+
 }
 
 func TestFirewallPoolSyncNodes(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges())
+	simpleclient := firewallclient.NewSimpleClientset()
```
swetharepakula, line 64: same comment as above, rename to fwClient.

pkg/firewalls/firewalls_test.go
```diff
@@ -39,49 +41,65 @@ func portRanges() []string {
...
 		t.Errorf("unexpected err when syncing firewall, err: %v", err)
 	}
 	verifyFirewallRule(fwp, ruleName, nodes, srcRanges, portRanges(), t)
+	verifyFirewallCR(simpleclient, ruleName, srcRanges, portRanges(), true, t)
 }
 
 func TestFirewallPoolSyncSrcRanges(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges())
+	simpleclient := firewallclient.NewSimpleClientset()
```
swetharepakula, line 93: same comment as above.

pkg/firewalls/firewalls_test.go
```diff
@@ -279,7 +328,54 @@ func validateXPNError(err error, op string, t *testing.T) {
 	}
 }
 
-func verifyFirewallRule(fwp *fakeFirewallsProvider, ruleName string, expectedNodes, expectedCIDRs, expectedPorts []string, t *testing.T) {
+func verifyFirewallCR(firewallclient *firewallclient.Clientset, ruleName string, sourceRanges, expectedPorts []string, crEnabled bool, t *testing.T) {
+	if crEnabled {
```
swetharepakula, line 332:
```
if !crEnabled{
    return
}
```

pkg/firewalls/controller.go
```diff
@@ -161,6 +186,11 @@ func (fwc *FirewallController) sync(key string) error {
 		if err := fwc.firewallPool.GC(); err != nil {
```
swetharepakula, line 186: Should this now be flag gated? We only GC if `enableFWControllerEnforcement` is false?

pkg/firewalls/firewalls.go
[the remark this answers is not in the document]
sugangli: It could an error of creation or an error of getting the CR but not the NotFound. Is it really necessary to put this in the comment?

pkg/firewalls/controller.go
sugangli: Done.

pkg/context/context.go
```diff
@@ -170,6 +176,7 @@ func NewControllerContext(
 		PodInformer:             informerv1.NewPodInformer(kubeClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer()),
 		NodeInformer:            informerv1.NewNodeInformer(kubeClient, config.ResyncPeriod, utils.NewNamespaceIndexer()),
 		SvcNegInformer:          informersvcneg.NewServiceNetworkEndpointGroupInformer(svcnegClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer()),
+		FirewallInformer:        informerfirewall.NewGCPFirewallInformer(firewallClient, config.ResyncPeriod, utils.NewNamespaceIndexer()),
```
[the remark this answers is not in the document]
sugangli: Updated. We are checking whether if firewallClient is nil.

cmd/glbc/main.go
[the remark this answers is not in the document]
sugangli: Done.

pkg/context/context.go
```diff
@@ -397,6 +404,7 @@ func (ctx *ControllerContext) Start(stopCh chan struct{}) {
 	go ctx.ServiceInformer.Run(stopCh)
 	go ctx.PodInformer.Run(stopCh)
 	go ctx.NodeInformer.Run(stopCh)
+	go ctx.FirewallInformer.Run(stopCh)
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/controller.go
```diff
@@ -161,6 +186,11 @@ func (fwc *FirewallController) sync(key string) error {
...
 			klog.Errorf("Could not garbage collect firewall pool, got error: %v", err)
 		}
+		if flags.F.EnableFirewallCR {
```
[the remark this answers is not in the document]
sugangli: Added two bool to the controller.

pkg/firewalls/controller_test.go
```diff
@@ -41,16 +42,17 @@ import (
...
 
 	ctxConfig := context.ControllerContextConfig{
 		Namespace:             api_v1.NamespaceAll,
 		ResyncPeriod:          1 * time.Minute,
 		DefaultBackendSvcPort: test.DefaultBeSvcPort,
 	}
-
-	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, nil, nil, nil, fakeGCE, defaultNamer, "" /*kubeSystemUID*/, ctxConfig)
-	fwc := NewFirewallController(ctx, []string{"30000-32767"})
+	kubeConfig := NewFakeKubeConfig()
```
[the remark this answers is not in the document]
sugangli: Good catch! Removed it.

pkg/firewalls/firewalls.go
```diff
@@ -100,6 +116,24 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
 		TargetTags: targetTags,
 	}
+	if fr.enableFirewallCR {
+		klog.V(3).Infof("Firewall CR is enabled.")
+		expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{}, fr.enableFWControllerEnforcement)
+		if err != nil {
+			return err
+		}
+		err = ensureFirewallCR(fr.firewallClient, expectedFirewallCR)
+		if err != nil && errors.Is(err, fmt.Errorf(string(v1beta1.FirewallRuleReasonXPNPermissionError))) {
```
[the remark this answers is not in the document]
sugangli: Either way works.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +150,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
 
+// ensureFirewallCR creates/updates the firewall CR
+// On CR update, it will read the conditions to see if there are errors updated by PFW controller.
+// If the Spec was updated by others, it will reconcile the Spec.
+func ensureFirewallCR(client firewallclient.Interface, expectedFWCR *v1beta1.GCPFirewall) error {
+	fw := client.NetworkingV1beta1().GCPFirewalls()
+	currentFWCR, err := fw.Get(context.Background(), expectedFWCR.Name, metav1.GetOptions{})
+
+	klog.V(3).Infof("ensureFirewallCR Get CR :%+v", currentFWCR)
+
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +150,127 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
...
+	if err != nil {
+		// Create the CR if it is not found.
+		if api_errors.IsNotFound(err) {
+			klog.V(3).Infof("ensureFirewallCR Create CR :%+v", expectedFWCR)
```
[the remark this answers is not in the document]
sugangli: The rest of the comments are using simple present. It will be better if we keep it consistent with others?

pkg/firewalls/firewalls_test.go
sugangli: Done.

pkg/firewalls/firewalls_test.go
sugangli: Done.

pkg/firewalls/firewalls_test.go
sugangli: Done.

pkg/firewalls/firewalls_test.go
sugangli: Done.

pkg/firewalls/firewalls_test.go
sugangli: Updated.

pkg/firewalls/controller.go
sugangli: Yes. I am 90% sure we will support the legacy firewall in the beginning, which means PFW controller will handle the old FW's deletion if it is in the enforce mode.

pkg/flags/flags.go
```diff
@@ -257,6 +259,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableMultipleIGs, "enable-multiple-igs", false, "Enable using multiple unmanaged instance groups")
 	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
+	flag.BoolVar(&F.EnableFirewallCR, "enable-firewall-cr", false, "Enable generating firewall CR")
+	flag.BoolVar(&F.EnableFWControllerEnforcement, "enable-fw-controller-enforcement", false, "Enable the GCP firewall controller to enforce the CRs. If set to true, Ingress Controller stops creating GCE firewall rules. We can only enable this if enable-firewall-cr sets to true.")
```
swetharepakula, line 263: I think we should make this flag relative to the ingress-gce controllers. It feels conceptually weird to enable something else enforcing firewalls. Instead we should probably change this to EnableFWEnforcement and default to true. Then when the platform firewall controller is ready to be turned on, we will change EnableFWEnforcement to be false.

alternatively, we could do a flag that says disableFWEnforcement, but I slightly prefer the above to avoid having the double negative.

sugangli: How about call it DisableFWEnforcement and default to false? It sounds more intuitive as we are deprecating ingress controller enforcing gce FWs. EnableXXX sounds more like we are adding XXX to the controller.

pkg/firewalls/firewalls.go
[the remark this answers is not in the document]
swetharepakula: I think we should update the others to be like `Creating` or `Created` depending on whether the action we are logging has happened or not. If you think the error has enough context, I am fine with this as is.

pkg/firewalls/firewalls.go
swetharepakula, line 126: we should log the error if it is not this case.

pkg/firewalls/firewalls.go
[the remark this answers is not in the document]
swetharepakula: I was more wondering if the error by itself has enough context, so whatever is consuming, we should be able to understand in the logs what caused it.

pkg/firewalls/fakes.go
```diff
@@ -21,17 +21,25 @@ import (
...
 type fakeFirewallsProvider struct {
 	fw               map[string]*compute.Firewall
+	fwcr             map[string]*v1beta1.GCPFirewall
 	networkProjectID string
 	networkURL       string
 	onXPN            bool
 	fwReadOnly       bool
 }
 
+func NewFakeKubeConfig() *rest.Config {
```
swetharepakula, line 39: Is this used somewhere?

swetharepakula (commented):
I think this mostly looks good to me. I have thought about the two flags, and I think the enforcement flag should be renamed to indicate a behavior in ingress-gce instead of a behavior on an external component.

Thanks Sugang!

pkg/context/context.go
```diff
@@ -397,6 +406,9 @@ func (ctx *ControllerContext) Start(stopCh chan struct{}) {
 	go ctx.ServiceInformer.Run(stopCh)
 	go ctx.PodInformer.Run(stopCh)
 	go ctx.NodeInformer.Run(stopCh)
+	if ctx.FirewallInformer != nil {
+		go ctx.FirewallInformer.Run(stopCh)
+	}
```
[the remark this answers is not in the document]
sugangli: Updated.

cmd/glbc/main.go
```diff
@@ -279,7 +287,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 		})
 	}
 
-	fwc := firewalls.NewFirewallController(ctx, flags.F.NodePortRanges.Values())
+	fwc := firewalls.NewFirewallController(ctx, flags.F.NodePortRanges.Values(), flags.F.EnableFirewallCR, flags.F.EnableFWControllerEnforcement)
```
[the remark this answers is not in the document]
sugangli: Per discussion, I renamed it to DisableFWEnforcement. Since they mean the same thing (disable ingress controller FW enforcement = enable FW controller enforcement), I renamed all of them without changing the logic.

pkg/firewalls/controller.go
```diff
@@ -56,20 +57,25 @@ type FirewallController struct {
 	translator   *translator.Translator
 	nodeLister   cache.Indexer
 	hasSynced    func() bool
+	enableCR     bool
+	enforceCR    bool
```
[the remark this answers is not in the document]
sugangli: See the comment above.

pkg/firewalls/controller.go
```diff
@@ -158,8 +187,15 @@ func (fwc *FirewallController) sync(key string) error {
 
 	// If there are no more ingresses, then delete the firewall rule.
 	if len(gceIngresses) == 0 {
-		if err := fwc.firewallPool.GC(); err != nil {
-			klog.Errorf("Could not garbage collect firewall pool, got error: %v", err)
+		if !fwc.enforceCR {
```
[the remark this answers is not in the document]
sugangli: Added the check in the cmd/glbc/main.go. We set DisableFWEnforcement to false as long as EnableFirewallCR is false.

pkg/firewalls/firewalls.go
sugangli: We log every action(create, update, and CR get). If there is an error, we should be able the see the action log first.

pkg/firewalls/fakes.go
sugangli: Good catch. Deleted.

pkg/controller/controller_test.go
```diff
@@ -72,7 +74,7 @@ func newLoadBalancerController() *LoadBalancerController {
 		DefaultBackendSvcPort: test.DefaultBeSvcPort,
 		HealthCheckPath:       "/",
 	}
-	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, nil, nil, nil, fakeGCE, namer, "" /*kubeSystemUID*/, ctxConfig)
+	ctx := context.NewControllerContext(nil, kubeClient, backendConfigClient, nil, firewallClient, nil, nil, nil, fakeGCE, namer, "" /*kubeSystemUID*/, ctxConfig)
```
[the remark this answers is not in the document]
sugangli: Updated.

cmd/glbc/main.go
```diff
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
+	firewallclient "k8s.io/cloud-provider-gcp/crd/client/gcpfirewall/clientset/versioned"
```
cezarygerard, line 44: please rename to firewallcrclient

cmd/glbc/main.go
```diff
@@ -133,6 +134,13 @@ func main() {
 		}
 	}
 
+	var firewallClient firewallclient.Interface
```
cezarygerard, line 137: firewallCRClient

here and further

cmd/glbc/main.go
```diff
@@ -279,7 +287,12 @@ func runControllers(ctx *ingctx.ControllerContext) {
 		})
 	}
 
-	fwc := firewalls.NewFirewallController(ctx, flags.F.NodePortRanges.Values())
+	if !flags.F.EnableFirewallCR {
```
cezarygerard, line 290: I think we should panic here

pkg/flags/flags.go
```diff
@@ -257,6 +259,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableMultipleIGs, "enable-multiple-igs", false, "Enable using multiple unmanaged instance groups")
 	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
+	flag.BoolVar(&F.EnableFirewallCR, "enable-firewall-cr", false, "Enable generating firewall CR")
```
cezarygerard, line 262: why not StringVar with 3 valid values?

pkg/firewalls/controller.go
```diff
@@ -119,6 +125,29 @@ func NewFirewallController(
 		},
 	})
 
+	if enableCR {
+		// FW CRs will be updated/deleted by the PFW controller or the user. Ingress controller need to watch such events
+		// and act accordingly.
+		ctx.FirewallInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
cezarygerard, line 131: ctx.FirewallInformer.AddEventHandler

what we get updates on? Firewall CRs rigth?
when the Platform Firewall Controller changes the CR?
or when customer changes something there?

pkg/firewalls/controller.go
```diff
@@ -119,6 +125,29 @@ func NewFirewallController(
...
+			// We enqueue the sync task for all update events for these two reasons:
+			// 1. Make sure the CR is consistent with the FW configuration.
+			// 2. Force the controller to read the CR status and populate proper errors.
+			UpdateFunc: func(old, cur interface{}) {
+				curFW := cur.(*v1beta1.GCPFirewall)
+				name := fwc.ctx.ClusterNamer.FirewallRule()
+				if name != curFW.Name {
+					return
+				}
+				fwc.queue.Enqueue(queueKey)
```
cezarygerard, line 141: what should be the queueKey?

I am confused why we always use the fake?

```
	// queueKey is a "fake" key which can be enqueued to a task queue.
	queueKey = &v1.Ingress{
		ObjectMeta: metav1.ObjectMeta{Name: "queueKey"},
	}
```

@swetharepakula ?

pkg/firewalls/interfaces.go
cezarygerard: please see my proposal here:
https://github.com/cezarygerard/ingress-gce/commit/450a41f7e9f2074fcf79dd35443c1cf96a395bc2

on this branch in my fork
https://github.com/cezarygerard/ingress-gce/commits/composite-l7-firewalls

the firewall logic is much cleaner, fewer variables are passed around, way fewer 'if' statements


What is the problem with XPN error handling?
The dry run mode can be utilized by simply creating a composite firewall

@swetharepakula WDYT?

cmd/glbc/main.go
cezarygerard: firewallClient would suggest this is provisioning resources on the GCP

pkg/flags/flags.go
sugangli: Technically I am fine with either way. But StringVar does not validate the values and we need to check it by ourselves? Plus, the other features flags are using Bool so it might be more consistent?

pkg/firewalls/controller.go
sugangli: The PFW controller will update the status of the CR, and some status requires the ingress controller to take actions. For example, the ingress controller needs to populate gcloud command for the user if there is an XPNPermission error.

pkg/firewalls/controller.go
```diff
@@ -50,26 +51,31 @@ var (
...
+	disableFWEnforcement bool
 }
 
 // NewFirewallController returns a new firewall controller.
 func NewFirewallController(
 	ctx *context.ControllerContext,
-	portRanges []string) *FirewallController {
-	firewallPool := NewFirewallPool(ctx.Cloud, ctx.ClusterNamer, gce.L7LoadBalancerSrcRanges(), portRanges)
+	portRanges []string,
+	enableCR, disableFWEnforcement bool) *FirewallController {
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/firewalls.go
```diff
@@ -116,13 +150,126 @@ func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []str
 	return fr.updateFirewall(expectedFirewall)
 }
 
+// ensureFirewallCR creates/updates the firewall CR
+// On CR update, it will read the conditions to see if there are errors updated by PFW controller.
```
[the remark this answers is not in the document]
sugangli: Done.

pkg/firewalls/controller.go
```diff
@@ -119,6 +125,29 @@ func NewFirewallController(
 		},
 	})
 
+	if enableCR {
+		// FW CRs will be updated/deleted by the PFW controller or the user. Ingress controller need to watch such events
```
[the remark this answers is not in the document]
sugangli: Done.

cmd/glbc/main.go
sugangli: Done.

cmd/glbc/main.go
sugangli: Done.

cmd/glbc/main.go
sugangli: Agreed. Updated.

pkg/firewalls/controller.go
swetharepakula: I didn't catch this initially. Even if we will only have one, I think it makes sense to either have a constant for the name of the single firewall CR or to just enqueue the CR from the `cur` in update func.
My preference is for the later, because I think it makes the most sense.

We should be checking that the firewall cr that we see is the one we expect before adding the queueKey.

pkg/firewalls/interfaces.go
sugangli: I just updated the PR according to your proposal.  I keep the bool flags as we don't need them everywhere, and it makes the fw pool logic cleaner. I will remove the comment-out proposal in pkg/firewalls/controller.go once we reach to an agreement.

sugangli:
> 

I refactored L7 firewall pool code path according to what @cezarygerard  proposed. Waiting for his comment.  We have few more PRs to come after this. But I am only able to spend 20% of my time on this project.

cezarygerard:
it's LGTM overall

thank you for applying my suggestions

let's just let's just remove this comment

"//// if we had enum-like flag 'crMode' to guard the firewall CR creation the above code would look ~like:"
and I will tag the change

PR #1628, opened 2021-12-03
Refactor, extract instancePool, translator

panslava:


pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -217,7 +216,6 @@ func newL4NetLBServiceController() *L4NetLBController {
 	}
 
 	lc := NewL4NetLBController(ctx, stopCh)
-	lc.Init()
```
kl52752, line 220: you can just `return NewL4NetLBController(ctx, stopCh)`

panslava: thank you, fixed!

Issue #1590, opened 2021-10-27
Migrate glbc leader election client to lease lock


rf232:
Context: kubernetes/kubernetes#80289

freehan:
merged.

PR #1634, opened 2021-12-10
Migrate gce-ingress to configmaps-leases leaderelection lock

wojtek-t:
Ref https://github.com/kubernetes/ingress-gce/issues/1590

/assign @bowei @freehan

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -224,7 +224,9 @@ func makeLeaderElectionConfig(ctx *ingctx.ControllerContext, client clientset.In
 	}
 	// add a uniquifier so that two processes on the same host don't accidentally both become active
 	id := fmt.Sprintf("%v_%x", hostname, rand.Intn(1e6))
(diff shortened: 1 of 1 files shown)


freehan (approved):
/lgtm
/approve

freehan:
Does this require graceful migration between the 2 mechanism? Or they are okay to migrate directly?

wojtek-t:
> Does this require graceful migration between the 2 mechanism? Or they are okay to migrate directly?

This one is graceful migration (intermediate step) towards eventual step (lease-based election).
So this PR can just be deployed as is without any additional action.

@freehan ^^

acumino:
Why do we need this https://github.com/kubernetes/ingress-gce/blob/b589a5813e6b4fe9f35d7674df43c4825380bcb0/pkg/flags/flags.go#L50 if any way we are setting the resource lock explicitly here https://github.com/kubernetes/ingress-gce/blob/b589a5813e6b4fe9f35d7674df43c4825380bcb0/cmd/glbc/main.go#L228-L236

PR #1635, opened 2021-12-13
Extract NodeController from l7 lb

panslava:


pkg/controller/node.go
```diff
@@ -70,13 +73,15 @@ func NewNodeController(ctx *context.ControllerContext) *NodeController {
 // Run the queue to process updates for the controller. This must be run in a
 // separate goroutine (method will block until queue shutdown).
 func (c *NodeController) Run() {
+	defer c.Shutdown()
```
kl52752, line 76: this will not work this queue is blocking (see comment above :))

kl52752: fixed

pkg/controller/node.go
```diff
@@ -70,13 +73,15 @@ func NewNodeController(ctx *context.ControllerContext) *NodeController {
...
 	start := time.Now()
 	for !c.hasSynced() {
 		klog.V(2).Infof("Waiting for hasSynced (%s elapsed)", time.Now().Sub(start))
 		time.Sleep(1 * time.Second)
 	}
 	klog.V(2).Infof("Caches synced (took %s)", time.Now().Sub(start))
-	c.queue.Run()
+	go c.queue.Run()
+	<-c.stopCh
```
prameshj, line 84: Maybe add c.Shutdown after this line, instead of defer?

panslava: Agree, this feels safer, fixed!

PR #1636, opened 2021-12-13
Add Network Tier support to L4 NetLB

kl52752:


cezarygerard:
No new tests in forwarding_rules_test.go ?
Maybe we should cover the network tier in the unit tests?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -226,7 +226,7 @@ func (l *L4) ensureForwardingRule(loadBalancerName, bsLink string, options gce.I
 	// If the network is not a legacy network, use the address manager
 	if !l.cloud.IsLegacyNetwork() {
 		nm := types.NamespacedName{Namespace: l.Service.Namespace, Name: l.Service.Name}.String()
-		addrMgr = newAddressManager(l.cloud, nm, l.cloud.Region(), subnetworkURL, loadBalancerName, ipToUse, cloud.SchemeInternal)
+		addrMgr = newAddressManager(l.cloud, nm, l.cloud.Region(), subnetworkURL, loadBalancerName, ipToUse, cloud.SchemeInternal, cloud.NetworkTierPremium)
```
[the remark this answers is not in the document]
kl52752: Yes,  ILB Forwarding rules can only be created in Premium tier.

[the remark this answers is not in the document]
kl52752: I also added test cases you asked in comment above

freehan:
/ok-to-test

Need to rebase

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -337,11 +337,20 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
 	ipToUse := l4lbIPToUse(l4netlb.Service, existingFwdRule, "")
 	klog.V(2).Infof("ensureExternalForwardingRule(%v): LoadBalancer IP %s", loadBalancerName, ipToUse)
 
+	// If network tier annotation in Service Spec is present
+	// check if it match network tiers form forwarding rule and external ip Address.
+	// If they do not match, tear down the existing resources with the wrong tier.
+	netTier, isFromSvc := utils.GetNetworkTier(l4netlb.Service)
+	if isFromSvc {
+		if err := l4netlb.tearDownResourcesWithWrongNetworkTier(existingFwdRule, ipToUse, netTier); err != nil {
```
freehan, line 345: There is also a project wide network tier default config in GCE API. [
](https://cloud.google.com/network-tiers/docs/using-network-service-tiers#setting_the_tier_for_all_resources_in_a_project)


I am wondering if the old NetLB created, does the service controller specify network tier during creation? If not, then it uses the project level default. Please bear in mind that some user might change the default config. Will this accidentally tear down user's old LB?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -400,6 +410,26 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
 	return composite.GetForwardingRule(l4netlb.cloud, key, fr.Version)
 }
 
+func (l4netlb *L4NetLB) tearDownResourcesWithWrongNetworkTier(existingFwdRule *composite.ForwardingRule, ipToUse string, svcNetTier cloud.NetworkTier) error {
```
freehan, line 413: add comment

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -400,6 +410,26 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
...
+	if existingFwdRule != nil && existingFwdRule.NetworkTier != svcNetTier.ToGCEValue() {
+		l4netlb.deleteForwardingRule(existingFwdRule.Name, meta.VersionGA)
+	}
+	if ipToUse != "" {
+		addr, err := l4netlb.cloud.GetRegionAddressByIP(l4netlb.cloud.Region(), ipToUse)
+		if err == nil && addr != nil && addr.NetworkTier != svcNetTier.ToGCEValue() {
```
freehan, line 419: suggest to write it this way:

addr, err := l4netlb.cloud.GetRegionAddressByIP(l4netlb.cloud.Region(), ipToUse)

if utils.IsNotFoundError(err) {
    return nil
}

....// delete and check error

pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -742,3 +743,27 @@ func TestControllerShouldNotProcessServicesWithLegacyFwR(t *testing.T) {
 		t.Errorf("Service should not be marked for update")
 	}
 }
+
+func TestControllerUserIPWithStandardNetworkTier(t *testing.T) {
+	// Network Tier from User Static Address should match network tier from forwarding rule.
+	// Premium Network Tier is default for creating forwarding rule so if User wants to use Standard Network Tier for Static Address
+	// should include network tier annotation in Service.
```
prameshj, line 750: nit: "they" should include network tier...

pkg/l4netlb/l4netlbcontroller_test.go
```diff
@@ -742,3 +743,27 @@ func TestControllerShouldNotProcessServicesWithLegacyFwR(t *testing.T) {
...
+	lc := newL4NetLBServiceController()
+
+	svc := test.NewL4NetLBService(8080, defaultNodePort)
+	svc.Spec.LoadBalancerIP = usersIP
+	addNetLBService(lc, svc)
+	key, _ := common.KeyFunc(svc)
+	addUsersStaticAddress(lc, cloud.NetworkTierStandard)
+	// Simulate block address reservation because address is owned by User.
+	lc.ctx.Cloud.Compute().(*cloud.MockGCE).MockAddresses.InsertHook = test.InsertAddressErrorHook
+	if err := lc.sync(key); err == nil {
```
prameshj, line 761: Maybe we should test the case where a mismatch results in an expected error?

pkg/loadbalancers/forwarding_rules.go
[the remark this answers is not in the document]
prameshj: Add comment above this line that ILB always Premium tier?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -337,11 +337,20 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
 	ipToUse := l4lbIPToUse(l4netlb.Service, existingFwdRule, "")
 	klog.V(2).Infof("ensureExternalForwardingRule(%v): LoadBalancer IP %s", loadBalancerName, ipToUse)
 
+	// If network tier annotation in Service Spec is present
+	// check if it match network tiers form forwarding rule and external ip Address.
```
prameshj, line 341: nit: s/form/from

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -400,6 +410,26 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
...
 }
 
+func (l4netlb *L4NetLB) tearDownResourcesWithWrongNetworkTier(existingFwdRule *composite.ForwardingRule, ipToUse string, svcNetTier cloud.NetworkTier) error {
+	if existingFwdRule != nil && existingFwdRule.NetworkTier != svcNetTier.ToGCEValue() {
+		l4netlb.deleteForwardingRule(existingFwdRule.Name, meta.VersionGA)
+	}
+	if ipToUse != "" {
+		addr, err := l4netlb.cloud.GetRegionAddressByIP(l4netlb.cloud.Region(), ipToUse)
+		if err == nil && addr != nil && addr.NetworkTier != svcNetTier.ToGCEValue() {
+			if ipToUse == l4netlb.Service.Spec.LoadBalancerIP {
```
prameshj, line 420: Why does this mean it is user-owned? Is it because GCE-reserved IPs are released immediately?Can you add a comment?

Vlad1mir-D:
This feature is so long awaited.
Thank you guys, please keep this up!

pkg/loadbalancers/forwarding_rules.go
kl52752: done

pkg/loadbalancers/forwarding_rules.go
kl52752: done

pkg/l4netlb/l4netlbcontroller_test.go
kl52752: done

pkg/l4netlb/l4netlbcontroller_test.go
kl52752: Well actually this is the case. I added check for error message and some comment.

pkg/loadbalancers/forwarding_rules.go
kl52752: done

pkg/loadbalancers/forwarding_rules.go
kl52752: No. it is because if User want to create LB with specific address he needs to add it to Spec loadBalancerIP field. When Ingress-GCE creates ephemeral IP address for LB this field (loadBalancerIP) is not populated and remain empty. In the address manager the check to distinguish between User own and Controller own address is address name (for controller own address its name should match load balancer name) but since I have service here it is easier for me to check service spec.

kl52752: Or maybe this assumption is wrong. The user can specify ip address in service spec that is not reserved then the controller will try to reserve this address and if operation is successful then this will be controller owned address?
Does it make and sense?

pkg/loadbalancers/forwarding_rules.go
kl52752: The old NetLB checks for service Network Tier annotation and if it is not present then it sets NT for default (Premium). https://github.dev/kubernetes/kubernetes/blob/489fb9bee3f626b3eeb120a5af89ad8c2b2f1c20/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_external.go#L95
Then this Network Tier is passed to ensure address IP and forwarding rule.
Resources with wrong NT will only be teared down if the service annotation is present in other case there will be an error ensuring LB (but the resources will not be changed).

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -752,3 +753,31 @@ func TestControllerShouldNotProcessServicesWithLegacyFwR(t *testing.T) {
...
+
+	lc := newL4NetLBServiceController()
+
+	svc := test.NewL4NetLBService(8080, defaultNodePort)
+	svc.Spec.LoadBalancerIP = usersIP
+	addNetLBService(lc, svc)
+	key, _ := common.KeyFunc(svc)
+	addUsersStaticAddress(lc, cloud.NetworkTierStandard)
+	// Simulate block address reservation because address is owned by User.
+	lc.ctx.Cloud.Compute().(*cloud.MockGCE).MockAddresses.InsertHook = test.InsertAddressErrorHook
```
prameshj, line 773: Do you need this ErrorHook? The test needs to: 1) Create Address with standard tier 2) Sync LB and see the WrongNetworkTierError when creating forwarding rule, right?

pkg/loadbalancers/address_manager_test.go
```diff
@@ -112,6 +126,22 @@ func TestAddressManagerExternallyOwned(t *testing.T) {
...
+// owned by the controller and it's network tier doesn't match expected.
+func TestAddressManagerExternallyOwnedWrongNetworkTier(t *testing.T) {
+	svc, err := fakeGCECloud(vals)
+	require.NoError(t, err)
+	targetIP := "1.1.1.1"
+
+	addr := &compute.Address{Name: "my-important-address", Address: targetIP, AddressType: string(cloud.SchemeInternal), NetworkTier: string(cloud.NetworkTierStandard)}
+	err = svc.ReserveRegionAddress(addr, vals.Region)
+	require.NoError(t, err, "")
+	svc.Compute().(*cloud.MockGCE).MockAddresses.InsertHook = test.InsertAddressErrorHook
```
prameshj, line 139: Same question as my other comment - why do we need this Error hook? Even without this, HoldAddress() should invoke validateAddress that will throw the mismatched tier error?

pkg/loadbalancers/forwarding_rules.go
prameshj: `l4netlb.Service.Spec.LoadBalancerIP != ""` does not guarantee that user reserved the ip. As you mentioned, user can specify an IP(without reserving) and expect the controller to reserve it.
Also, when service Update is processed, or if the controller restarts due to master upgrade, it will get a service object with LoadBalancer IP populated. That does not mean it is user-reserved.

However, addr != nil means user reserved it, right?

pkg/loadbalancers/forwarding_rules.go
prameshj: @freehan What do we expect to happen in case user changed the project-wide default. Example - 1) Default was premium, they created NetLB and reserved IP in premium tier.
2) They changed project default to Standard.  They did not change the tier for the address.
3) GetNetworkTier in controller still picks premium as default(since it is a const string). So no errors in LB sync loop.
But future NetLBs that don't specify any tier annotation, will also be created in premium default. Their project-wide setting wont be applied. This should be ok?

pkg/loadbalancers/forwarding_rules.go
kl52752: `However, addr != nil means user reserved it, right?` no this might means that this is Update operation and address was reserved by controller previously. Well I think I need to check address name as it is in address manager.

pkg/l4lb/l4netlbcontroller_test.go
kl52752: Yes you are right this mock is not necessary here

pkg/loadbalancers/forwarding_rules.go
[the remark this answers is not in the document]
kl52752: done

pkg/loadbalancers/address_manager_test.go
kl52752: removed

prameshj:
Couple of nits, otherwise LGTM.

pkg/utils/utils.go
```diff
@@ -734,3 +734,22 @@ func GetBasePath(cloud *gce.Cloud) string {
...
+func GetNetworkTier(service *api_v1.Service) (cloud.NetworkTier, bool) {
+	l, ok := service.Annotations[annotations.NetworkTierAnnotationKey]
+	if !ok {
+		return cloud.NetworkTierDefault, false
+	}
+
+	v := cloud.NetworkTier(l)
+	switch v {
+	case cloud.NetworkTierStandard:
+		fallthrough
```
freehan, line 749: why is this fallthrough?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -337,11 +338,22 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string, existingFwdR
...
 	// If the network is not a legacy network, use the address manager
 	if !l4netlb.cloud.IsLegacyNetwork() {
 		nm := types.NamespacedName{Namespace: l4netlb.Service.Namespace, Name: l4netlb.Service.Name}.String()
-		addrMgr = newAddressManager(l4netlb.cloud, nm, l4netlb.cloud.Region(), "", loadBalancerName, ipToUse, cloud.SchemeExternal)
+		addrMgr := newAddressManager(l4netlb.cloud, nm, l4netlb.cloud.Region() /*subnetURL = */, "", loadBalancerName, ipToUse, cloud.SchemeExternal, netTier)
+
+		// If network tier annotation in Service Spec is present
+		// check if it match network tiers from forwarding rule and external ip Address.
+		// If they do not match, tear down the existing resources with the wrong tier.
+		if isFromSvc {
```
freehan, line 351: This means the GCE project wide default will be effective right? I thought we decided to ignore the project level default and treat the annotation as source of truth. If the annotation is not present, then use Premium tier.

pkg/annotations/service.go
```diff
@@ -61,6 +61,10 @@ const (
 
 	// BackendConfigKey is GA version of backend config key.
 	BackendConfigKey = "cloud.google.com/backend-config"
+	// NetworkTierAnnotationKey is annotated on a Service object to indicate which
+	// network tier a GCP LB should use.
+	// The valid values are "Standard" and "Premium" (default).
+	NetworkTierAnnotationKey = "cloud.google.com/network-tier"
```
freehan, line 67: // The valid values are "Standard" and "Premium" (default, if unspecified).

freehan (changes requested):

freehan:
Need rebase as well

pkg/loadbalancers/forwarding_rules.go
freehan: If the existing service controller supports the GCE project wide default, then we should be backward compatible. Can you double check?

pkg/utils/utils.go
kl52752: changed.

pkg/loadbalancers/forwarding_rules.go
kl52752: Yes, you are right project level network tier is ignored here.
`isFromSvc ` means that annotation is present in service spec and according to svc controllers codebase we should tear down resources`only when` annotation is present in service. Maybe the name is wrong I can rename it to `isFromAnnotation`?

pkg/annotations/service.go
kl52752: done

freehan (approved):
/lgtm
/approve

kl52752:
@freehan @prameshj I found some bug related to ILB we cannot create internal IP address and set Network Tier (even for Premium). I fixed it and added UT.

PR #1638, opened 2021-12-14
Add PSC upgrade test

swetharepakula:
* covers change in CRD version and usage of GCE GA API


cmd/e2e-test/upgrade/psc.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+// PreUpgrade implements e2e.UpgradeTest.PreUpgrade.
+func (p *PSC) PreUpgrade() error {
+	err := e2e.CreateSubnet(p.s, e2e.PSCSubnetName, e2e.PSCSubnetPurpose)
+	if err != nil {
+		// Subnet Creation could fail because of a leaked subnet from a previous run.
+		// Instead of failing the test, attempt to reuse subnet.
+		p.t.Logf("error creating psc subnet %s: %q", e2e.PSCSubnetName, err)
```
prameshj, line 63: nit: upper case log messages.

cmd/e2e-test/upgrade/psc.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// PostUpgrade implements e2e.UpgradeTest.PostUpgrade
+func (p *PSC) PostUpgrade() error {
+	// ensure that service attachment is queryable and is configured as expected
+	if _, err := e2e.WaitForServiceAttachment(p.s, saName); err != nil {
+		p.t.Fatalf("failed waiting for service attachment: %q", err)
+	}
+
+	gceSAURL := pscUpdateAndValidate(p.t, p.s, p.framework)
```
prameshj, line 94: Is this to ensure that updates to CR continue to work after upgrade? Maybe consider adding a comment?

prameshj (commented):
Looks good to me.

cmd/e2e-test/upgrade/psc.go
swetharepakula: done

cmd/e2e-test/upgrade/psc.go
swetharepakula: Good point! I am testing to make sure the CR is updatable as well as the underlying GCE PSC resource. Since we switch from using the beta API to the GA API too. 

I have added a comment explaining this too. Thanks!

PR #1642, opened 2021-12-20
External and Internal Load Balancers backend should use the same k8s2 name convention

kl52752:
Function creating names for ILB aligning with k8s2 convention is
VMIPNEG.
Since we use the same name for neg, backend service and firewall rules I
rename this function to L4Backend. Now we can use the same function for
NetLB services and NetLB does not use NEG so VMIPNEG name is not
suitable here.

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -80,10 +80,10 @@ func getILBOptions(svc *corev1.Service) gce.ILBOptions {
 func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4LBSyncResult {
 	klog.V(2).Infof("EnsureInternalLoadBalancerDeleted(%s): attempting delete of load balancer resources", l.NamespacedName.String())
 	result := &L4LBSyncResult{SyncType: SyncTypeDelete, StartTime: time.Now()}
-	// All resources use the NEG Name, except forwarding rule.
-	name, ok := l.namer.VMIPNEG(svc.Namespace, svc.Name)
+	// All resources use the L4Backend Name, except forwarding rule.
+	name, ok := l.namer.L4Backend(svc.Namespace, svc.Name)
 	if !ok {
-		result.Error = fmt.Errorf("Namer does not support L4 VMIPNEGs")
+		result.Error = fmt.Errorf("Namer does not support L4 Backends")
 		return result
 	}
 	frName := l.GetFRName()
@@ -198,8 +198,8 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 	}
 
 	l.Service = svc
-	// Use the same resource name for NEG, BackendService as well as FR, FWRule.
-	name, ok := l.namer.VMIPNEG(l.Service.Namespace, l.Service.Name)
+	// All resources use the L4Backend name, except forwarding rule.
+	name, ok := l.namer.L4Backend(l.Service.Namespace, l.Service.Name)
 	if !ok {
 		result.Error = fmt.Errorf("Namer does not support L4 VMIPNEGs")
 		return result
--- a/pkg/utils/namer/interfaces.go
+++ b/pkg/utils/namer/interfaces.go
@@ -57,9 +57,10 @@ type BackendNamer interface {
 	// NEG returns the gce neg name based on the service namespace, name
 	// and target port.
 	NEG(namespace, name string, Port int32) string
-	// VMIPNEG returns the gce neg name based on the service namespace and name.
-	// The second output parameter indicates if the namer supports VM_IP_NEGs.
-	VMIPNEG(namespace, name string) (string, bool)
+	// L4Backend returns the name for L4 LB backend resources, based on the service namespace and name.
+	// It supports ILB with subsetting enabled (VM_IP_NEGs) and NetLB with RBS enabled.
+	// The second output parameter indicates if the namer is supported.
+	L4Backend(namespace, name string) (string, bool)
 	// InstanceGroup constructs the name for an Instance Group.
 	InstanceGroup() string
 	// NamedPort returns the name for a named port.
--- a/pkg/utils/namer/l4_namer.go
+++ b/pkg/utils/namer/l4_namer.go
@@ -45,11 +45,11 @@ func NewL4Namer(kubeSystemUID string, namer *Namer) *L4Namer {
 	return &L4Namer{v2Prefix: defaultPrefix + schemaVersionV2, v2ClusterUID: clusterUID, Namer: namer}
 }
 
-// VMIPNEG returns the gce VM_IP_NEG name based on the service namespace and name
-// NEG naming convention:
+// L4Backend returns the gce L4 Backend name based on the service namespace and name
+// Naming convention:
 //   k8s2-{uid}-{ns}-{name}-{suffix}
 // Output name is at most 63 characters.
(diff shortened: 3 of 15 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -81,7 +81,7 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4LbSyncRes
 	klog.V(2).Infof("EnsureInternalLoadBalancerDeleted(%s): attempting delete of load balancer resources", l.NamespacedName.String())
 	result := &L4LbSyncResult{SyncType: SyncTypeDelete, StartTime: time.Now()}
 	// All resources use the NEG Name, except forwarding rule.
```
prameshj, line 83: Nit: update comment - "All resources use the L4Backend name, except forwarding rule."

pkg/loadbalancers/l4.go
```diff
@@ -81,7 +81,7 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4LbSyncRes
...
-	name, ok := l.namer.VMIPNEG(svc.Namespace, svc.Name)
+	name, ok := l.namer.L4Backend(svc.Namespace, svc.Name)
 	if !ok {
 		result.Error = fmt.Errorf("Namer does not support L4 VMIPNEGs")
```
prameshj, line 86: nit: "does not support L4 Backends"

pkg/loadbalancers/l4.go
```diff
@@ -199,7 +199,7 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 
 	l.Service = svc
 	// Use the same resource name for NEG, BackendService as well as FR, FWRule.
```
prameshj, line 201: nit: can you update this too? looks like it was incorrect from before.

"All resources use the L4Backend name, except forwarding rule."

pkg/utils/namer/l4_namer.go
```diff
@@ -45,11 +45,11 @@ func NewL4Namer(kubeSystemUID string, namer *Namer) *L4Namer {
 	return &L4Namer{v2Prefix: defaultPrefix + schemaVersionV2, v2ClusterUID: clusterUID, Namer: namer}
 }
 
-// VMIPNEG returns the gce VM_IP_NEG name based on the service namespace and name
-// NEG naming convention:
+// L4Backend returns the gce L4 Backend name based on the service namespace and name
+// Nnaming convention:
```
prameshj, line 49: Nit: remove extra "n" in Naming.

pkg/utils/namer/l4_namer.go
```diff
@@ -73,7 +73,7 @@ func (namer *L4Namer) L4ForwardingRule(namespace, name, protocol string) string
 // L4HealthCheck returns the name of the L4 ILB Healthcheck and the associated firewall rule.
```
prameshj, line 73: nit: replace ILB with LB?

pkg/utils/namer/interfaces.go
```diff
@@ -57,9 +57,9 @@ type BackendNamer interface {
 	// NEG returns the gce neg name based on the service namespace, name
 	// and target port.
 	NEG(namespace, name string, Port int32) string
-	// VMIPNEG returns the gce neg name based on the service namespace and name.
-	// The second output parameter indicates if the namer supports VM_IP_NEGs.
-	VMIPNEG(namespace, name string) (string, bool)
+	// L4Backend returns the gce name based on the service namespace and name.
```
prameshj, line 60: nit: "..returns the name for L4 LB backend resources, based on the service namespace and name."?

prameshj (commented):
Looks good, except got a few nits.

pkg/utils/namer/interfaces.go
kl52752: done

pkg/utils/namer/l4_namer.go
kl52752: done

pkg/utils/namer/l4_namer.go
kl52752: done

pkg/loadbalancers/l4.go
kl52752: done

pkg/loadbalancers/l4.go
kl52752: done

pkg/loadbalancers/l4.go
kl52752: done

pkg/utils/serviceport.go
```diff
@@ -71,9 +72,10 @@ func (sp *ServicePort) GetDescription() Description {
 func (sp *ServicePort) BackendName() string {
 	if sp.NEGEnabled {
 		return sp.BackendNamer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.Port)
-	} else if sp.VMIPNEGEnabled {
-		negName, _ := sp.BackendNamer.VMIPNEG(sp.ID.Service.Namespace, sp.ID.Service.Name)
-		return negName
+	} else if sp.VMIPNEGEnabled || sp.L4RBSEnabled {
+		// Use L4 Backend name for both Internal and External LoadBalancers
+		backendName, _ := sp.BackendNamer.L4Backend(sp.ID.Service.Namespace, sp.ID.Service.Name)
```
freehan, line 77: @prameshj I am wondering if this would cause any race condition when user switch the loadbalancer type from NetLB to ILB?

pkg/utils/namer/interfaces.go
```diff
@@ -57,9 +57,9 @@ type BackendNamer interface {
 	// NEG returns the gce neg name based on the service namespace, name
 	// and target port.
 	NEG(namespace, name string, Port int32) string
-	// VMIPNEG returns the gce neg name based on the service namespace and name.
-	// The second output parameter indicates if the namer supports VM_IP_NEGs.
-	VMIPNEG(namespace, name string) (string, bool)
+	// L4Backend returns the name for L4 LB backend resources, based on the service namespace and name.
+	// The second output parameter indicates if the namer supports L4 Name.
```
freehan, line 61: add comment to say this supports ILB with subsetting enabled (VM_IP_NEGs) and NetLB with RBS enabled.

kl52752: done

pkg/utils/serviceport.go
kl52752: Yes, you are right, there is a race condition in case ILB <-> XLB update. One problem here is that "new service" will try to update the Backend Service or Forwarding rule scheme and this action will return an error from googleapis.
One solution might be to check for finalizers before adding svc to the queue. ILB controller will check L4NetLBFinalizer and XLB will check ILB_v2 finalizer. Then when the updated service finishes cleaning its stack it will remove the finalizer and the other controller will pick up this service and create its own stack.
WDYT? Does it make any sense?

kl52752: I tested update XLB -> ILB and ILB -> XLB and there are some error trying to ensure `new` load balancer but when `old` 
LB is successfully removed then `new` LB is ensured without errors and works fine. This transition takes up to 1min.

freehan: SGTM

freehan (approved):
/lgtm
/approve

PR #1643, opened 2021-12-21
Fix instance selection for node Tag

kl52752:
If in proejct there are multiple instances in one zone there might be an
error in chosing instance for node tag.
Instance should be select from cluster's instance group.


freehan (commented):
/lgtm

PR #1647, opened 2021-12-30
Extract common parts for L4 LoadBalancers - Part 3

kl52752:
Extract common L4 LB firewall functions

### DIFF
--- a/pkg/firewalls/firewalls_l4.go
+++ b/pkg/firewalls/firewalls_l4.go
@@ -18,48 +18,62 @@ package firewalls
 
 import (
 	"strings"
+	"sync"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"google.golang.org/api/compute/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/ingress-gce/pkg/utils"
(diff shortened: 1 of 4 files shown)


pkg/firewalls/firewalls_l4.go
```diff
@@ -18,60 +18,75 @@ package firewalls
...
+	"k8s.io/client-go/tools/record"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog"
 	"k8s.io/legacy-cloud-providers/gce"
 )
 
-func EnsureL4FirewallRule(cloud *gce.Cloud, fwName, lbIP, nsName string, sourceRanges, portRanges, nodeNames []string, proto string, sharedRule bool, l4Type utils.L4LBType) error {
-	existingFw, err := cloud.GetFirewall(fwName)
+// FirewallParams holds all data needed to create firewall for L4 LB
+type FirewallParams struct {
```
freehan, line 33: I am wondering if it is necessary to expose the unused Firewall fields. For instance, DestRanges etc...

freehan (commented):
LGTM overall.

@prameshj wait for second pair of eyes

pkg/firewalls/firewalls_l4.go
kl52752: Yes you are right, removed.

PR #1648, opened 2021-12-30
L4NetLB e2e tests

kl52752:


freehan (commented):
are you sure you want to use the ingress e2e framework to test NetLB? It has some pros and cons .

I would suggest you still put it in a different guitar test. That would require you to copy some of the code of ingress-gce guitar test.

cmd/e2e-test/l4netlb_test.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		if svc1Params.HcName == svc2ParamsLocal.HcName {
+			t.Fatalf("Services should not share health checks")
+		}
+
+		// Update Service ExternalTrafficPolicy to Cluster
+		if _, err := e2e.EnsureL4NetLBEchoService(s, svcName2, nil, replicas, v1.ServiceExternalTrafficPolicyTypeCluster); err != nil {
+			t.Fatalf("Error updating echo service %s/%s: %q", s.Namespace, svcName2, err)
+		}
+		time.Sleep(60 * time.Second)
```
freehan, line 79: sleep is sometimes quite fragile. Suggest just fold the sleep time into whatever Timeout in the following wait functions.

cmd/e2e-test/l4netlb_test.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// Test scenario
+// 1. Create NetLB service.
+// 2. Verify service resources and Forwarding rule schema EXTERNAL.
+// 3. Update service annotation to Internal.
+// 4. Verify service resources and Forwarding rule schema INTERNAL.
+// 5. Remove service annotation Internal.
+// 6. Verify service resources and Forwarding rule schema EXTERNAL.
+func TestL4NetLbMigrateToILB(t *testing.T) {
```
freehan, line 115: for this kind of transition test, I would suggest write it in this structure. So that it could minimize the duplicated code.

```
for _, tc := range []struct{
     desc  string
     transition func()
     validateSomething bool
}{
   {
        desc: "initial state"
        transition: func() {
               // change service object from 1 spec to another
     
        }
        validateSomething: false
   },
   { 
       desc: "Internal -> External"
       ...
    }
} {
     // pre transition check
    tc.transition
   
   // post transition common check
   WaitForSomethingToHappen(...)
   
   if tc.validateSomething {
        // do additional validation
    }


}

```

pkg/e2e/helpers.go
```diff
@@ -458,6 +458,54 @@ func CheckSvcEvents(s *Sandbox, svcName, msgType, message string, ignoreMessages
 	return false, nil
 }
 
+// CheckSvcUpdateEvents checks that the service has 2 consecutive events update and ensure.
+func CheckSvcUpdateEvents(s *Sandbox, svcName, updateMsg, ensureMsg string) (bool, error) {
```
freehan, line 462: Checking event might be a bit fragile. If the event emission changed in the controller implementation, then the e2e test would fail immediately. 

I would recommend examine the API contract (Service Status updated with LB VIP) directly. And validate with sending traffic to VIP. (But because the implementation, traffic would not actually hit the ILB. It will be captured by iptables first. ) If you want to validate with traffic, the client needs to be outside of the cluster. Check the ILB guitar test where it creates a separate VM to send traffic to ILB.

cmd/e2e-test/l4netlb_test.go
kl52752: done

cmd/e2e-test/l4netlb_test.go
kl52752: you are right this test is not well written, due to issue with ILB check (it needs to be done from internal vm) I removed this test and it will be added in guitar test

pkg/e2e/helpers.go
kl52752: removed

PR #1649, opened 2021-12-30
Add instance group management enabling flags

kl52752:


pkg/flags/flags.go
```diff
@@ -245,6 +247,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
 	flag.BoolVar(&F.EnableTrafficScaling, "enable-traffic-scaling", false, "Enable support for Service {max-rate-per-endpoint, capacity-scaler}")
 	flag.BoolVar(&F.EnableEndpointSlices, "enable-endpoint-slices", false, "Enable using Endpoint Slices API instead of Endpoints API")
+	flag.BoolVar(&F.EnableMultipleIgs, "enable-multiple-igs", false, "Enable using unmanaged instance group management")
+	flag.IntVar(&F.MaxIgSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
```
freehan, line 251: would this default to 15?

1000 means that you would need to precreate 1000 IGs?

kl52752: No max-ig-size is a number of instances we can add to Unmanaged Instance Group, default value is 1000 but for testing purposes we would like this value to be modifiable. When max-ig-size for IG is reached then new IG will be created and new nodes will be added there.

freehan (approved):
/lgtm
/approve

PR #1652, opened 2022-01-11
update k8s api to 0.22.2

anthonyho007:
to fix ingress-gce apis import error due to https://github.com/kubernetes/kube-openapi/pull/234
```
../../../../pkg/mod/k8s.io/ingress-gce@v1.10.13/pkg/apis/backendconfig/v1/zz_generated.openapi.go:49:3: cannot use "github.com/go-openapi/spec".Schema{...} (type "github.com/go-openapi/spec".Schema) as type "k8s.io/kube-openapi/pkg/validation/spec".Schema in field value
```


gillesdouaire:
Running into this issue when working on Kubernetes 1.22 compatibility, we have a strong dependency on this. This PR is much needed and appreciated 👍

gillesdouaire:
@bowei @swetharepakula

swetharepakula:
Thanks for the PR! We have been holding back on this because we have needed elements from older K8s releases for the e2e tests. Our e2e tests need to be able to run against K8s versions earlier than 1.22 which has been our biggest blocker. We will reevaluate our decision and see if we are in a position to update now.

However before this change can be made, the dependency changes require ingress-gce to be updated to use Go 1.16 or higher.

/ok-to-test

swetharepakula:
The repo has been updated to use Golang 1.16, which unblocks the upgrade to k8s api 0.22.2. Please try rebasing. The test failures are not flakes but due to the golang versions

PR #1653, opened 2022-01-12
avoid updating BackendService if possible

freehan:


bowei:
@freehan where are the tests for this change?

PR #1658, opened 2022-01-26
Change l4elb-rbs-mode flag to simple boolean flag run-l4-netlb-controller

cezarygerard:
Change l4elb-rbs-mode flag to simple boolean flag run-l4-netlb-controller

The flag is used in main.go to enable L4 Net LB controller in ingress-gce


freehan (commented):
/lgtm
/approve

PR #1666, opened 2022-02-02
Add check for forwarding rule name in annotation for l4 rbs

panslava:
We were checking only for TCPForwardingRule, or UDPForwardingRule key existence in annotations, and were considering such services as RBS L4 NetLB

But ILB services also have these keys, but with different values

This pr adds check if value of this annotation matches L4NetLB naming

### DIFF
--- a/pkg/l4lb/l4netlbcontroller_test.go
+++ b/pkg/l4lb/l4netlbcontroller_test.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -34,7 +35,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/cloud-provider/service/helpers"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/composite"
 	ingctx "k8s.io/ingress-gce/pkg/context"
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -827,6 +827,47 @@ func TestIsRBSBasedService(t *testing.T) {
...
+		},
+		{
+			createControllerAndService: func() (*L4NetLBController, *v1.Service) {
+				svc, l4netController := createAndSyncLegacyNetLBSvc(t)
+				l4netlb := loadbalancers.NewL4NetLB(svc, l4netController.ctx.Cloud, meta.Regional, l4netController.namer, l4netController.ctx.Recorder(svc.Namespace), &l4netController.sharedResourcesLock)
+				svc.Annotations = map[string]string{annotations.UDPForwardingRuleKey: l4netlb.GetFRName()}
+				return l4netController, svc
+			},
+			isRBSService: true,
+			desc:         "Should detect RBS by TCP forwarding rule annotation with matching name",
```
kl52752, line 848:
```suggestion
			desc:         "Should detect RBS by UDP forwarding rule annotation with matching name",
```

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -827,6 +827,47 @@ func TestIsRBSBasedService(t *testing.T) {
...
+				}
+				return l4netController, svc
+			},
+			isRBSService: false,
+			desc:         "Should not detect RBS by forwarding rule annotations without matching name",
+		},
+		{
+			createControllerAndService: func() (*L4NetLBController, *v1.Service) {
+				_, l4netController := createAndSyncLegacyNetLBSvc(t)
+				ilbSvc := test.NewL4ILBService(false, 8080)
```
kl52752, line 865: I think that you should add here TCP/UDP annotation with ilb naming scheme.

kl52752 (changes requested):

pkg/l4lb/l4netlbcontroller_test.go
panslava: Thank you, fixed

pkg/l4lb/l4netlbcontroller_test.go
panslava: Thank you, added!

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -827,6 +829,53 @@ func TestIsRBSBasedService(t *testing.T) {
 			isRBSService: false,
 			desc:         "Should not detect RBS by forwarding rule pointed to target pool",
 		},
+		{
```
cezarygerard, line 832: there is so many things that could be improved with this TestIsRBSBasedService
- use simple test struct outside of for loop (such approach has IDE support)
- keep the test case struct simple
- build the required structs inside the for loop
- limit anonymous functions calls
- you don't have to run all tests for a faeture in single loop, just run those with similar setup
- use 	t.Run(testCase.desc, func(t *testing.T) so test frameworks can see different test cases
- format test error messages with got before want

Check the refactore version here: https://github.com/kubernetes/ingress-gce/pull/1669


For testing the forwarding rule annotation you can use simple, non-table test:

```
func TestForwardingRuleAnnotations(t *testing.T) {
	svc := test.NewL4LegacyNetLBService(8080, 30234)
	controller := newL4NetLBServiceController()
	l4netlb := loadbalancers.NewL4NetLB(svc, controller.ctx.Cloud, meta.Regional, controller.namer, controller.ctx.Recorder(svc.Namespace), &controller.sharedResourcesLock)
	frName := l4netlb.GetFRName()

	// just make sure the APi is not called once
	controller.ctx.Cloud.Compute().(*cloud.MockGCE).MockForwardingRules.GetHook = func (ctx context.Context, key *meta.Key, m *cloud.MockForwardingRules) (bool, *ga.ForwardingRule, error) {
		t.Errorf("this should not be called")
		return true, nil, nil
	}
	expectRBSService := true

	svc.Annotations = map[string]string{annotations.TCPForwardingRuleKey: frName}
	if expectRBSService != controller.isRBSBasedService(svc) {
		t.Errorf("Should detect RBS by TCP forwarding rule annotation. isRBSBasedService(%v) = %v, want %v", svc, false, expectRBSService)
	}

	svc.Annotations = map[string]string{annotations.UDPForwardingRuleKey: frName}
	if expectRBSService != controller.isRBSBasedService(svc) {
		t.Errorf("Should detect RBS by UDP forwarding rule annotation. isRBSBasedService(%v) = %v, want %v", svc, false, expectRBSService)
	}
}
```

plus add one test case into my refactoring from pull/1669

```
		{
			desc:         "Should not detect RBS by forwarding rule annotations without matching name",
			annotations : map[string]string{annotations.UDPForwardingRuleKey: "fr-1", annotations.TCPForwardingRuleKey: "fr-2"},
			expectRBSService: false,
		},
```

panslava: thank you, fixed

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -775,64 +778,111 @@ func TestControllerUserIPWithStandardNetworkTier(t *testing.T) {
...
 		{
-			createControllerAndService: func() (*L4NetLBController, *v1.Service) {
-				svc, l4netController := createAndSyncLegacyNetLBSvc(t)
-				// Add Forwarding Rule pointing to RBS backend
-				l4netController.ctx.Cloud.Compute().(*cloud.MockGCE).MockForwardingRules.GetHook = test.GetRBSForwardingRule
-				return l4netController, svc
-			},
-			isRBSService: true,
-			desc:         "Should detect RBS by forwarding rule",
+			desc:             "Should detect RBS by finalizer when service contains legacy finalizer",
```
kl52752, line 806:
```suggestion
			desc:             "Should detect RBS by finalizer when service contains both legacy and NetLB finalizers",
```

cezarygerard:
@bowei @freehan

as for now this is the top priority PR for RBS
 we need this to enable RBS controller on nightly tests

freehan (approved):
/lgtm
/approve

PR #1667, opened 2022-02-02
Add L4 Metrics - Part 1 refactor

kl52752:
Refactor L4ILBServiceState so we can use it to both Internal and
External L4 LoadBalancer.
I moved L4 ILB metrics test to new file because I will add there tests
for L4 NetLB in next part.
I added NewL4NetLBMetricsStates in this PR to show how those metrics
will be used by NetLB controller.


cezarygerard:
@kl52752 

I miss why you want to combine L4ILB metrics with L4NetLB metrics?

They should stay separate, we need to be able to distinguish them, even if some features/options for ILB/NetLB of are named the same

I would suggest defining L4NetServiceState
with hardcoded list of features that can be counted
we are not going to add many more new types to benefit from generic counting

kl52752:
@cezarygerard my intention was not to combine ILB and NetLB Metrics together but only create common type so we can use it in `L4SyncResult` because this struct is common for both L4 loadbalnacers and I didn't want to put there 2 fields with metric one for ILB and second for NetLB.
I tried to "hardcoded" metric value for both LB in functions for ILB:
in `newL4ILBSyncResult` you can see that Metrics for ILB are created using ` metrics.NewL4NetLBMetricsStates()` where I initialise Metric map with all needed keys for L4 ILB. And during metric collection in func `ValidateILBMetrics` I check if given status has all L4 ILB metrics added.
And for NetLB is/will be the same scenario but I will add and validate different keys set.
I didn't create L4NetServiceState because it will be a lot of code duplication and IMO there is no need to do this the only thing what is important is the fact that we always need to create L4ServiceState with appropriate function either for ILB or NetLB and because for the validation of metric state in metric controller we can easily see the error if this condition will not be met.

kl52752:
@cezarygerard  And also this is only refactoring part I will add next part with actual L4 NetLB metrics implementation. I wanted to separate those 2 topics.

kl52752:
replaced with https://github.com/kubernetes/ingress-gce/pull/1671

PR #1668, opened 2022-02-02
Add multiple zones support to fake instance groups

panslava:
We are using FakeInstanceGroups from `pkg/instances/fakes.go` as a mock for testing interactions with GCE instance groups api.  

This PR rewrites FakeInstanceGroups so it can support multiple instance groups in multiple zones

This PR does not involve any production code changes, only test code changes

### DIFF
--- a/pkg/instances/fakes.go
+++ b/pkg/instances/fakes.go
@@ -24,21 +24,23 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"google.golang.org/api/compute/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/test"
 	"k8s.io/ingress-gce/pkg/utils"
-	"k8s.io/ingress-gce/pkg/utils/namer"
 )
 
-// NewFakeInstanceGroups creates a new FakeInstanceGroups.
-func NewFakeInstanceGroups(nodes sets.String, namer *namer.Namer) *FakeInstanceGroups {
+// NewEmptyFakeInstanceGroups creates a new FakeInstanceGroups without zones, igs or instances.
+func NewEmptyFakeInstanceGroups() *FakeInstanceGroups {
 	return &FakeInstanceGroups{
-		instances:        nodes,
-		listResult:       getInstanceList(nodes),
-		namer:            namer,
-		zonesToInstances: map[string][]string{},
+		zonesToIGsToInstances: map[string]IGsToInstances{},
 	}
 }
 
-// InstanceGroup fakes
+// NewFakeInstanceGroups creates a new FakeInstanceGroups.
+func NewFakeInstanceGroups(zonesToIGsToInstances map[string]IGsToInstances) *FakeInstanceGroups {
+	return &FakeInstanceGroups{
+		zonesToIGsToInstances: zonesToIGsToInstances,
+	}
+}
 
 // FakeZoneLister records zones for nodes.
 type FakeZoneLister struct {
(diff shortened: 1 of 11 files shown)


pkg/instances/fakes.go
```diff
@@ -25,16 +25,15 @@ import (
 	"google.golang.org/api/compute/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/ingress-gce/pkg/utils"
-	"k8s.io/ingress-gce/pkg/utils/namer"
 )
 
 // NewFakeInstanceGroups creates a new FakeInstanceGroups.
-func NewFakeInstanceGroups(nodes sets.String, namer *namer.Namer) *FakeInstanceGroups {
+func NewFakeInstanceGroups(zoneToIGsToInstances map[string]map[*compute.InstanceGroup]sets.String) *FakeInstanceGroups {
```
cezarygerard, line 31: can you 'override' this 'constructor' method with second one to ease the instances_test.go code?

```
func NewFakeInstanceGroupsSingleZone(nodes sets.String, igName, zone string) *FakeInstanceGroups {
	ig := &compute.InstanceGroup{Name: igName}
	zoneToIGsToInstances := map[string]map[*compute.InstanceGroup]sets.String{
		zone: {
			ig: sets.NewString([]string{"n1", "n2"}...),
		},
	}
	return NewFakeInstanceGroups(zoneToIGsToInstances)
}
```

pkg/instances/instances_test.go
```diff
@@ -40,7 +41,12 @@ func newNodePool(f *FakeInstanceGroups, zone string) NodePool {
 }
 
 func TestNodePoolSync(t *testing.T) {
-	f := NewFakeInstanceGroups(sets.NewString([]string{"n1", "n2"}...), defaultNamer)
+	ig := &compute.InstanceGroup{Name: defaultNamer.InstanceGroup()}
```
cezarygerard, line 44: overriding the NewFakeInstanceGroupsSingleZone method mentioned above
simplifies this code
```
	ig := &compute.InstanceGroup{Name: defaultNamer.InstanceGroup()}
	f := NewFakeInstanceGroups(map[string]map[*compute.InstanceGroup]sets.String{
		defaultZone: {
			ig: sets.NewString([]string{"n1", "n2"}...),
		},
	})
```

into
```
	f := NewFakeInstanceGroupsSingleZone(sets.NewString("n1", "n2"), defaultNamer.InstanceGroup(), defaultZone)
```
here and 5 further places

pkg/instances/fakes.go
panslava: We spoke with Cezary and decided that it would be better not to add this extra constructor

pkg/instances/instances_test.go
panslava: We spoke with Cezary and decided that it would be better not to add this extra constructor

pkg/instances/fakes.go
```diff
@@ -24,21 +24,23 @@ import (
...
-// NewFakeInstanceGroups creates a new FakeInstanceGroups.
-func NewFakeInstanceGroups(nodes sets.String, namer *namer.Namer) *FakeInstanceGroups {
+// NewEmptyFakeInstanceGroups creates a new FakeInstanceGroups without zones, igs or instances.
+func NewEmptyFakeInstanceGroups() *FakeInstanceGroups {
 	return &FakeInstanceGroups{
-		instances:        nodes,
-		listResult:       getInstanceList(nodes),
-		namer:            namer,
-		zonesToInstances: map[string][]string{},
+		ZonesToIGsToInstances: map[string]map[*compute.InstanceGroup]sets.String{},
```
cezarygerard, line 34: this type is really mouthful

maybe you can declare the 
`type IGsWithInstances map[*compute.InstanceGroup]sets.String`

which makes 
`ZonesToIGsToInstances map[string]IGsWithInstances`

pkg/instances/instances_test.go
```diff
@@ -139,7 +159,11 @@ func TestSetNamedPorts(t *testing.T) {
 }
 
 func TestGetInstanceReferences(t *testing.T) {
-	pool := newNodePool(NewFakeInstanceGroups(sets.NewString([]string{"ig"}...), defaultNamer), defaultZone)
+	pool := newNodePool(NewFakeInstanceGroups(map[string]map[*compute.InstanceGroup]sets.String{
+		defaultZone: {
+			&compute.InstanceGroup{Name: "ig"}: sets.NewString([]string{"ig"}...),
```
cezarygerard, line 164: &compute.InstanceGroup{Name: "ig"}: sets.NewString("ig"),

pkg/instances/fakes.go
```diff
@@ -60,113 +62,112 @@ func (z *FakeZoneLister) GetZoneForNode(name string) (string, error) {
...
+	if err != nil {
+		return err
 	}
-	f.zonesToInstances[zone] = append(f.zonesToInstances[zone], instanceNames...)
+
+	f.ZonesToIGsToInstances[zone][ig].Insert(instanceNames...)
 	return nil
 }
 
 // GetInstancesByZone returns the zone to instances map.
```
cezarygerard, line 143: this func is not used in the code

pkg/instances/fakes.go
```diff
@@ -60,113 +62,112 @@ func (z *FakeZoneLister) GetZoneForNode(name string) (string, error) {
...
-	instances        sets.String
-	instanceGroups   []*compute.InstanceGroup
-	getResult        *compute.InstanceGroup
-	listResult       *compute.InstanceGroupsListInstances
-	calls            []int
-	namer            *namer.Namer
-	zonesToInstances map[string][]string
+	getResult             *compute.InstanceGroup
+	calls                 []int
+	ZonesToIGsToInstances map[string]map[*compute.InstanceGroup]sets.String
```
cezarygerard, line 67: this does not have to be exported
and ZonesToIGsToInstances should not be referred from test code

pkg/instances/instances_test.go
```diff
@@ -40,7 +41,12 @@ func newNodePool(f *FakeInstanceGroups, zone string) NodePool {
 }
 
 func TestNodePoolSync(t *testing.T) {
-	f := NewFakeInstanceGroups(sets.NewString([]string{"n1", "n2"}...), defaultNamer)
+	ig := &compute.InstanceGroup{Name: defaultNamer.InstanceGroup()}
+	f := NewFakeInstanceGroups(map[string]map[*compute.InstanceGroup]sets.String{
```
cezarygerard, line 45: why 'f'?
goland has mare than 26 different variable names

maybe 'fakeInstanceGroups'
or at least fakeIGs

pkg/instances/instances_test.go
```diff
@@ -51,31 +57,41 @@ func TestNodePoolSync(t *testing.T) {
 	f.calls = []int{}
```
cezarygerard, line 57: f.calls = nil
is it better?

pkg/instances/instances_test.go
panslava: it was there before :( I will change

pkg/instances/instances_test.go
panslava: fixed and in all other places

pkg/instances/instances_test.go
cezarygerard: didnt' notice it was here before 
thanks for fixing

pkg/instances/instances_test.go
```diff
@@ -40,24 +41,43 @@ func newNodePool(f *FakeInstanceGroups, zone string) NodePool {
 	return pool
 }
 
+func getInstancesSet(f *FakeInstanceGroups, ig *compute.InstanceGroup) (sets.String, error) {
```
cezarygerard, line 44: this should be part of FakeInstanceGroups struct


```
func (f *FakeInstanceGroups,) getInstancesSet(ig *compute.InstanceGroup) (sets.String, error) {
...
```


it will be useful on testing other implementations of NodePool interface

pkg/instances/instances_test.go
cezarygerard: fair enough

pkg/instances/instances_test.go
panslava: I am not sure if we want to extend FakeInstanceGroups interface with some new public method, so it will not match real cloud.InstanceGroups

But at the same time I can imagine, that this function can be reused

I a bit refactored and moved this function to test/utils in new commit

PR #1672, opened 2022-02-09
L4 Metrics refactoring

kl52752:
- Splitting L4SyncResult to ILB and NetLB specific
- Extracting consts for L4 ILB Metrics Name
- Moving L4 ILB metrics test to new file
- Making metric test function generic


pkg/test/utils.go
```diff
@@ -333,15 +333,15 @@ type L4ILBLatencyMetricInfo struct {
...
 	var createCount, updateCount, deleteCount uint64
 	var createSum, updateSum, deleteSum float64
-	var result L4ILBLatencyMetricInfo
+	var result L4LBLatencyMetricInfo
 
-	latencyMetric, err := getPrometheusMetric("l4_ilb_sync_duration_seconds")
+	latencyMetric, err := getPrometheusMetric(metricName)
 	if err != nil {
-		t.Errorf("Failed to get L4 ILB prometheus metric 'l4_ilb_sync_duration_seconds', err: %v", err)
+		t.Errorf("Failed to get L4 LB prometheus metric '%s', err: %v", metricName, err)
```
cezarygerard, line 344: the 't' variable should not be asserted here

the GetL4LatencyMetric should return error
checked in test code

golang style ;-)


https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#assert

pkg/test/utils.go
```diff
@@ -333,15 +333,15 @@ type L4ILBLatencyMetricInfo struct {
 	deleteSum         float64
 }
 
-// GetL4LatencyMetric gets the current state of the l4_ilb_sync_duration_seconds metric.
-func GetL4LatencyMetric(t *testing.T) *L4ILBLatencyMetricInfo {
+// GetL4LatencyMetric gets the current state of the sync_duration_seconds metric for ILB or NetLB.
+func GetL4LatencyMetric(t *testing.T, metricName string) *L4LBLatencyMetricInfo {
```
cezarygerard, line 337: Instead of extending the GetL4LatencyMetric with new args you can override it:

```
func GetL4ILBLatencyMetric() (*L4LBLatencyMetricInfo, error) {
	return getL4LatencyMetric("l4_ilb_sync_duration_seconds")
}

func GetL4NetLBLatencyMetric() (*L4LBLatencyMetricInfo, error) {
	return getL4LatencyMetric("l4_ilb_sync_duration_seconds")
}

func getL4LatencyMetric(metricName string) (*L4LBLatencyMetricInfo, error) {
```

this will result in minimal changes to 
pkg/l4lb/l4controller_test.go

(apart from handling error from GetL4ILBLatencyMetric which is style guide requirement)

pkg/test/utils.go
```diff
@@ -322,8 +322,8 @@ func getPrometheusMetric(name string) (*dto.MetricFamily, error) {
 	return nil, nil
 }
 
-// L4LatencyMetricInfo holds the state of the l4_ilb_sync_duration_seconds metric.
-type L4ILBLatencyMetricInfo struct {
+// L4LatencyMetricInfo holds the state of the sync_duration_seconds metric for ILB or NetLB.
+type L4LBLatencyMetricInfo struct {
```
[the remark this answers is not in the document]
kl52752: Can you be a little more specific? I can't see any :)

pkg/test/utils.go
kl52752: done

pkg/test/utils.go
kl52752: This is not a part of this review I can create new PR with refactoring to align this test with golang style.

cezarygerard: sgtm

pkg/test/utils.go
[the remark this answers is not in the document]
cezarygerard: no typo, my bad ;-)

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -363,7 +363,8 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.L4
 		syncResult.Error = fmt.Errorf("failed to set resource annotations, err: %w", err)
 		return syncResult
 	}
-	return nil
+	syncResult.MetricsState.InSuccess = true
+	return syncResult
```
freehan, line 367: was this a bug ?
curious if this can be catch by a unit test?

pkg/loadbalancers/l4netlb.go
```diff
@@ -122,16 +135,17 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 
 	bs, err := l4netlb.backendPool.EnsureL4BackendService(name, hcLink, protocol, string(l4netlb.Service.Spec.SessionAffinity), string(cloud.SchemeExternal), l4netlb.NamespacedName, meta.VersionGA)
 	if err != nil {
+		klog.Errorf("Failed to ensure backend service - %v", err)
 		result.GCEResourceInError = annotations.BackendServiceResource
-		result.Error = fmt.Errorf("Failed to ensure backend service - %v", err)
+		result.Error = err
```
freehan, line 140: do you want to wrap around the error with %w for unwrapping later?

Check for ILB error metrics handling. With the wraping, it would allow error processing later in the metrics pipeline. 

For instance, you could determine if the error is an GCE error.

pkg/loadbalancers/l4netlb.go
```diff
@@ -122,16 +135,17 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
 		return result
 	}
 	result.Annotations[annotations.BackendServiceKey] = name
 	fr, err := l4netlb.ensureExternalForwardingRule(bs.SelfLink)
 	if err != nil {
-
+		klog.Errorf("L4 NetLB service: Failed to ensure forwarding rule - %v", err)
```
freehan, line 146: %w

freehan (commented):
lgtm overall.

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/l4lb/l4netlbcontroller.go
kl52752: yes, it was a bug :) this function is a part of sync() function which is tested in unit tests and sync returns error not status. This is snippet for extracting error from result.
```        result := lc.syncInternal(svc)
		if result == nil {
			// result will be nil if the service was ignored(due to presence of service controller finalizer).
			return nil
		}
                return result.Error
```
As you can see when we tested `sync` we checked that it didn't return the error and sync result is inside of this function and for happy path it was correct behaviour because when result is nil we returns nil error. 
 But I created tests for metrics and it was failing because we need not nil result in happy path to publish metrics. SO this test will be added in next review.

PR #1673, opened 2022-02-09
Refactor Address Manager to return if address is managed by controller

kl52752:
This change is needed because I would like to add metric for IP addresses which are managed by controller and this metric will be collected in L4NetLb controller.

### DIFF
--- a/pkg/loadbalancers/address_manager.go
+++ b/pkg/loadbalancers/address_manager.go
@@ -29,6 +29,15 @@ import (
 	"k8s.io/klog"
 )
 
+// IPAddressType defines if IP address is Managed by controller
+type IPAddressType int
+
+const (
+	IPAddrUndefined IPAddressType = iota // IP Address type could not be determine due to error is address provisioning.
+	IPAddrManaged
+	IPAddrUnmanaged
+)
+
 // Original file in https://github.com/kubernetes/legacy-cloud-providers/blob/6aa80146c33550e908aed072618bd7f9998837f6/gce/gce_address_manager.go
 type addressManager struct {
 	logPrefix   string
@@ -60,8 +69,8 @@ func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetU
 
 // HoldAddress will ensure that the IP is reserved with an address - either owned by the controller
 // or by a user. If the address is not the addressManager.name, then it's assumed to be a user's address.
-// The string returned is the reserved IP address.
-func (am *addressManager) HoldAddress() (string, error) {
+// The string returned is the reserved IP address and IPAddressType indicating if IP address is managed by controller.
+func (am *addressManager) HoldAddress() (string, IPAddressType, error) {
 	// HoldAddress starts with retrieving the address that we use for this load balancer (by name).
 	// Retrieving an address by IP will indicate if the IP is reserved and if reserved by the user
 	// or the controller, but won't tell us the current state of the controller's IP. The address
@@ -72,7 +81,7 @@ func (am *addressManager) HoldAddress() (string, error) {
 	// Get the address in case it was orphaned earlier
 	addr, err := am.svc.GetRegionAddress(am.name, am.region)
 	if err != nil && !utils.IsNotFoundError(err) {
-		return "", err
+		return "", IPAddrUndefined, err
 	}
 
 	if addr != nil {
@@ -80,7 +89,7 @@ func (am *addressManager) HoldAddress() (string, error) {
 		validationError := am.validateAddress(addr)
 		if validationError == nil {
 			klog.V(4).Infof("%v: address %q already reserves IP %q Type %q. No further action required.", am.logPrefix, addr.Name, addr.Address, addr.AddressType)
-			return addr.Address, nil
+			return addr.Address, IPAddrManaged, nil
 		}
 
 		klog.V(2).Infof("%v: deleting existing address because %v", am.logPrefix, validationError)
@@ -89,7 +98,7 @@ func (am *addressManager) HoldAddress() (string, error) {
 			if utils.IsNotFoundError(err) {
 				klog.V(4).Infof("%v: address %q was not found. Ignoring.", am.logPrefix, addr.Name)
 			} else {
-				return "", err
+				return "", IPAddrUndefined, err
 			}
 		} else {
 			klog.V(4).Infof("%v: successfully deleted previous address %q", am.logPrefix, addr.Name)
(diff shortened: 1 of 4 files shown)


pkg/loadbalancers/address_manager.go
```diff
@@ -141,55 +151,55 @@ func (am *addressManager) ensureAddressReservation() (string, error) {
...
 		// The address with this name is checked at the beginning of 'HoldAddress()', but for some reason
 		// it was re-created by this point. May be possible that two controllers are running.
 		klog.Warningf("%v: address %q unexpectedly existed with IP %q.", am.logPrefix, addr.Name, am.targetIP)
-	} else {
-		// If the retrieved address is not named with the loadbalancer name, then the controller does not own it, but will allow use of it.
-		klog.V(4).Infof("%v: address %q was already reserved with name: %q, description: %q", am.logPrefix, am.targetIP, addr.Name, addr.Description)
-		am.tryRelease = false
+		return addr.Address, Managed, nil
 	}
+	// If the retrieved address is not named with the loadbalancer name, then the controller does not own it, but will allow use of it.
```
cezarygerard, line 198: why you dropped the `else {` clause
it changes  the logic slightly

pkg/loadbalancers/address_manager.go
```diff
@@ -29,6 +29,14 @@ import (
 	"k8s.io/klog"
 )
 
+type IPAddressType int
```
cezarygerard, line 32: please ad some comment on what this type means

cezarygerard:
Please add the description  why we need this change

pkg/loadbalancers/address_manager.go
```diff
@@ -60,8 +68,8 @@ func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetU
 
 // HoldAddress will ensure that the IP is reserved with an address - either owned by the controller
 // or by a user. If the address is not the addressManager.name, then it's assumed to be a user's address.
-// The string returned is the reserved IP address.
-func (am *addressManager) HoldAddress() (string, error) {
+// The string returned is the reserved IP address and IPAddressType indicating if IP address is managed by controller.
```
cezarygerard, line 71: is it managed by controller or managed by user?

pkg/loadbalancers/address_manager.go
cezarygerard, line 32: Optional:
you could move the `type IPAddressType int` to separate package 'address' or 'addresstype' so the consts would always be referred as:Managed
addresstype.Unmanaged / addresstype.Managed / addresstype.Unmanaged

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -340,7 +339,7 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
 	klog.V(2).Infof("ensureExternalForwardingRule(%s): LoadBalancer IP %s", frName, ipToUse)
 
 	netTier, isFromAnnotation := utils.GetNetworkTier(l4netlb.Service)
-
+	isIPManaged := Managed
```
cezarygerard, line 342: you assume the address is managed by default ,is it correct?

pkg/loadbalancers/address_manager.go
kl52752: It's added in comment `IP address is managed by controller`

pkg/loadbalancers/address_manager.go
kl52752: done

kl52752: I agree that addresstype.Unmanaged is more descriptive than just Managed but to create a new poackage for one type, I don't think it is a good Idea, maybe I will rename the consts? Managed -> IPAddrManaged WDYT?

pkg/loadbalancers/forwarding_rules.go
kl52752: yes, the address will be managed by User if the User add it's address in Service.Spec.LoadBalancersIP

pkg/loadbalancers/address_manager.go
```diff
@@ -29,6 +29,15 @@ import (
 	"k8s.io/klog"
 )
 
+// IPAddressType defines if IP address is Managed by controller
+type IPAddressType int
+
+const (
+	IpAddrUndefined IPAddressType = iota
```
cezarygerard, line 36: 'IPAddrUndefined'
go/go-style/decisions#initialisms

same in the next names

pkg/loadbalancers/address_manager.go
```diff
@@ -29,6 +29,15 @@ import (
 	"k8s.io/klog"
 )
 
+// IPAddressType defines if IP address is Managed by controller
+type IPAddressType int
+
+const (
+	IPAddrUndefined IPAddressType = iota
```
freehan, line 36: add comments here to elaborate what each type implies

freehan (commented):
just nit: lgtm overall

pkg/loadbalancers/address_manager.go
cezarygerard: I'm ok with current new better names

anyway in golang I have not seen using new package as readability hack ;-)

pkg/loadbalancers/address_manager.go
cezarygerard: @kl52752 ping

why not leave

```
	if am.isManagedAddress(addr) {
		// The address with this name is checked at the beginning of 'HoldAddress()', but for some reason
		// it was re-created by this point. May be possible that two controllers are running.
		klog.Warningf("%v: address %q unexpectedly existed with IP %q.", am.logPrefix, addr.Name, am.targetIP)
		return addr.Address, IpAddrManaged, nil
	} else {
		// If the retrieved address is not named with the loadbalancer name, then the controller does not own it, but will allow use of it.
		klog.V(4).Infof("%v: address %q was already reserved with name: %q, description: %q", am.logPrefix, am.targetIP, addr.Name, addr.Description)
		am.tryRelease = false
		return addr.Address, IpAddrUnmanaged, nil
	}
```

pkg/loadbalancers/address_manager.go
kl52752: in line 32 there is a comment for typed `// IPAddressType defines if IP address is Managed by controller` so IMO it is readable that `IPAddrManaged`  means that IP address is managed by controller and `IPAddrUnmanaged` that IP address is not managed by controller. IPAddrUndefined is a tricky case because this means that there was an error in address ensuring so we could not determine its state. So maybe I will add only this comment?

PR #1676, opened 2022-02-15
Remove debug print statement

swetharepakula:
assign @freehan


freehan (approved):
/lgtm
/approve

bowei:
can we `git grep 'fmt.Print' | grep -v _test` are there any more places where we print?

PR #1686, opened 2022-03-28
add enqueue timestamp to sync function

kl52752:
This PR will enable measuring latency time starting from enqueue time
    till the end of processing time. I changed object type qhich is added to
    periodic queue. Previously we added  service Key which was a string. I
    changed this type into a struct so we can add timestam and service key
    there.

    In next PR I will use enqueu timestamp to measure latency for L4NetLB
    services.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -144,14 +144,15 @@ func NewLoadBalancerController(
 	ctx.IngressInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			addIng := obj.(*v1.Ingress)
+			ingKey := common.NamespacedName(addIng)
 			if !utils.IsGLBCIngress(addIng) {
-				klog.V(4).Infof("Ignoring add for ingress %v based on annotation %v", common.NamespacedName(addIng), annotations.IngressClassKey)
+				klog.V(4).Infof("Ignoring add for ingress %v based on annotation %v", ingKey, annotations.IngressClassKey)
 				return
 			}
 
-			klog.V(2).Infof("Ingress %v added, enqueuing", common.NamespacedName(addIng))
+			klog.V(2).Infof("Ingress %v added, enqueuing", ingKey)
 			lbc.ctx.Recorder(addIng.Namespace).Eventf(addIng, apiv1.EventTypeNormal, events.SyncIngress, "Scheduled for sync")
-			lbc.ingQueue.Enqueue(obj)
+			lbc.ingQueue.Enqueue(utils.NewTask(ingKey))
 		},
 		DeleteFunc: func(obj interface{}) {
 			delIng := obj.(*v1.Ingress)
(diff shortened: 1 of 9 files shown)


pkg/utils/taskqueue.go
```diff
@@ -35,6 +37,12 @@ type TaskQueue interface {
 	NumRequeues(obj interface{}) int
 }
 
+// TimestampTask defines service key and enqueue timestamp
```
freehan, line 40: nit:
How about `s/service key/object key`?

This is no only keys for services right? also ingresses

pkg/controller/controller.go
```diff
@@ -144,45 +144,48 @@ func NewLoadBalancerController(
...
-				klog.V(4).Infof("Ignoring add for ingress %v based on annotation %v", common.NamespacedName(addIng), annotations.IngressClassKey)
+				klog.V(4).Infof("Ignoring add for ingress %v based on annotation %v", ingKey, annotations.IngressClassKey)
 				return
 			}
 
-			klog.V(2).Infof("Ingress %v added, enqueuing", common.NamespacedName(addIng))
+			klog.V(2).Infof("Ingress %v added, enqueuing", ingKey)
 			lbc.ctx.Recorder(addIng.Namespace).Eventf(addIng, apiv1.EventTypeNormal, events.SyncIngress, "Scheduled for sync")
-			lbc.ingQueue.Enqueue(obj)
+			lbc.ingQueue.Enqueue(utils.TimestampTask{EnqueueTime: time.Now(), SvcKey: ingKey})
```
freehan, line 155: Just to reduce the duplicate code: 

One option: 
```
func NewTask(key string) {
      return utils.TimestampTask{EnqueueTime: time.Now(), SvcKey: key}
}
```

The other option is to generate a timestamp inside Enqueue function. 

I will suggest go with the second option so that no change from client side.

pkg/utils/taskqueue.go
```diff
@@ -95,13 +106,9 @@ func (t *PeriodicTaskQueueWithMultipleWorkers) Run() {
...
 	for _, obj := range objs {
-		key, err := t.keyFunc(obj)
-		if err != nil {
-			klog.Errorf("Couldn't get key for object %+v (type %T): %v", obj, obj, err)
-			return
-		}
+		key := obj.(TimestampTask).SvcKey
 		klog.V(4).Infof("Enqueue key=%q (%v)", key, t.resource)
-		t.queue.Add(key)
+		t.queue.Add(obj)
```
freehan, line 111: hmmmm I think this part voids this whole PR. 

queue is automatically deduped. If the same key gets added into the queue multiple times, it should be merged into one. However, with the timestamp, it is no longer able to dedup it.

pkg/utils/taskqueue.go
```diff
@@ -73,7 +84,7 @@ func (t *PeriodicTaskQueueWithMultipleWorkers) runInternal(workerId int) {
 			return
 		}
 		klog.V(4).Infof("Worker-%d: Syncing %v (%v)", workerId, key, t.resource)
-		if err := t.sync(key.(string)); err != nil {
+		if err := t.sync(key.(TimestampTask)); err != nil {
```
freehan, line 87: I would recommend measuring the time spent here directly instead of relying on the controller to implement sync function with internal to measure it.

freehan (commented):
This PR needs some rethinking. we should discuss in our weekly meeting

pkg/utils/taskqueue.go
kl52752: maybe just an object? Will suite both cases

pkg/utils/taskqueue.go
kl52752: Yes, good catch it is a bug. All I need ot do is to define custom String function for struct and pass only object key.
https://go.dev/play/

pkg/utils/taskqueue.go
kl52752: The idea is to pass enqueue timestamp to sync function so at the end of the sync in `publishMetrics` method we will calculate total time (queue time + processing time). This will be implemented in next part.

pkg/controller/controller.go
kl52752: I added NewTaskFunction, we can discussed the other option in the meeting

pkg/utils/taskqueue.go
kl52752: There are metrics related to the queue, the problem is that we cannot access them in controller and include them to latency metric. What I want to achieve is to have one duration metric from the time where service was enqueued till the end of the processing. So I thought that the easiest way to achieve that is to pass enqueue timestamp to sync function. I know that right now only L4NetLB controller will use this function but maybe in future we will implement this behaviour in other controllers? If it is not the case I can implement new periodic queue for L4NetLB controller only.

pkg/utils/taskqueue.go
kl52752: I will add tests to cover this case.

kl52752: Ok my experiment has a bug so the behaviour will be like this
"The key type can be any [comparable](https://golang.org/ref/spec#Comparison_operators) type; notably, this excludes functions, maps, and slices. The value type can be any type, including custom types or interface{}.
Interface values are comparable. Two interface values are equal if they have [identical](https://go.dev/ref/spec#Type_identity) dynamic types and equal dynamic values or if both have value nil.
Struct values are comparable if all their fields are comparable. Two struct values are equal if their corresponding non-[blank](https://go.dev/ref/spec#Blank_identifier) fields are equal."
And unfortunately we cannot define custom comparable operator

kl52752:
This pull request has a bug. Closing it

PR #1687, opened 2022-03-30
Upgrade golang version to 1.16

freehan:


bowei:
Why not 1.18? Why 1.16 specifically.

freehan:
No particular reason for 1.18. Mostly to align with k/k on using 1.16.

bowei:
K/k has moved to 1.18 I think

PR #1688, opened 2022-03-31
Delete service and wait for NEG deletion in ASM e2e tests

panslava:


pkg/e2e/fixtures.go
```diff
@@ -452,6 +452,11 @@ func CreatePorterService(s *Sandbox, name string) error {
 	return err
 }
 
+// DeletePorterService deletes the K8s porter service
+func DeletePorterService(s *Sandbox, svcName string) error {
```
freehan, line 456: Is this used?

freehan (commented):
ping me when fixed

pkg/e2e/fixtures.go
panslava: No, forgot to remove... Thanks

PR #1691, opened 2022-04-05
Improve RBS service deletion

cezarygerard:
In L4 RBS load balancer controller

1) remove service annotations, for  correct XLB->ILB transition when customer changes service annotations

2) try to delete the instance group to cleanup the instance group, this will
  - succeed when the last Service/LB using the instance group is deleted
  - fail if other L4/L7 load balancer is using the instance group
otherwise the IG can lea

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -403,24 +403,45 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
 	l4netLB := loadbalancers.NewL4NetLB(svc, lc.ctx.Cloud, meta.Regional, lc.namer, lc.ctx.Recorder(svc.Namespace), &lc.sharedResourcesLock)
 	lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeNormal, "DeletingLoadBalancer",
 		"Deleting L4 External LoadBalancer for %s", key)
+
 	result := l4netLB.EnsureLoadBalancerDeleted(svc)
 	if result.Error != nil {
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancerFailed",
 			"Error deleting L4 External LoadBalancer, err: %v", result.Error)
 		return result
 	}
+
 	if err := updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{}); err != nil {
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
 			"Error reseting L4 External LoadBalancer status to empty, err: %v", err)
 		result.Error = fmt.Errorf("Failed to reset L4 External LoadBalancer status, err: %w", err)
 		return result
 	}
+
+	// Remove LB annotations from the Service when processing the finalizer.
+	if err := updateAnnotations(lc.ctx, svc, nil); err != nil {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
+			"Error removing resource annotations: %v: %v", err)
+		result.Error = fmt.Errorf("Failed to reset resource annotations, err: %w", err)
+		return result
+	}
+
 	if err := common.EnsureDeleteServiceFinalizer(svc, common.NetLBFinalizerV2, lc.ctx.KubeClient); err != nil {
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancerFailed",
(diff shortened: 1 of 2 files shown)


bowei:
Can we add more details to the commit message regarding what this fixes

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -415,12 +415,31 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
 		result.Error = fmt.Errorf("Failed to reset L4 External LoadBalancer status, err: %w", err)
 		return result
 	}
+
+	// Also remove any LB annotations from the service metadata
```
bowei, line 419: // Remove LB annotations from the Service when processing the finalizer.

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -415,12 +415,31 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
...
+	if err := updateAnnotations(lc.ctx, svc, nil); err != nil {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
+			"Error resetting resource annotations for load balancer: %v", err)
+		result.Error = fmt.Errorf("failed to reset resource annotations, err: %w", err)
```
bowei, line 423: If we bail out here, the deletion will be retried?

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -415,12 +415,31 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
 		result.Error = fmt.Errorf("Failed to reset L4 External LoadBalancer status, err: %w", err)
 		return result
 	}
+
+	// Also remove any LB annotations from the service metadata
+	if err := updateAnnotations(lc.ctx, svc, nil); err != nil {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
+			"Error resetting resource annotations for load balancer: %v", err)
```
bowei, line 422: Error removing resource annotations: %v

pkg/l4lb/l4netlbcontroller.go
cezarygerard: yes, the deletion will be retried.


this line will make it retry deletion if annotations are still there:
https://github.com/kubernetes/ingress-gce/blob/master/pkg/l4lb/l4netlbcontroller.go#L240

pkg/l4lb/l4netlbcontroller.go
cezarygerard: done

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -403,24 +403,45 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
...
 	if err := common.EnsureDeleteServiceFinalizer(svc, common.NetLBFinalizerV2, lc.ctx.KubeClient); err != nil {
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancerFailed",
 			"Error removing finalizer from L4 External LoadBalancer, err: %v", err)
 		result.Error = fmt.Errorf("Failed to remove L4 External LoadBalancer finalizer, err: %w", err)
 		return result
 	}
+
+	// Try to delete instance group, instancePool.DeleteInstanceGroup ignores errors if resource is in use or not found.
+	// TODO(cezarygerard) replace with multi-IG management
+	if err := lc.instancePool.DeleteInstanceGroup(lc.namer.InstanceGroup()); err != nil {
```
bowei, line 438: can you add more information in the commit message why we need to add this extra deletion?

Under what conditions would this make a difference?

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -61,6 +62,11 @@ const (
 	userAddrName         = "UserStaticAddress"
 )
 
+var (
+	serviceAnnotationKeys = []string{annotations.FirewallRuleKey, annotations.BackendServiceKey, annotations.HealthcheckKey,
```
bowei, line 66: break lines for easier maintenance

```
serviceAnnotationKeys = []string{
  annotations.FirewallRuleKey,
 ...
}
```

pkg/l4lb/l4netlbcontroller.go
cezarygerard: done

pkg/l4lb/l4netlbcontroller_test.go
cezarygerard: done

cezarygerard:
@bowei 
all resolved

freehan (approved):
/lgtm
/approve

PR #1693, opened 2022-04-06
Add new flag for configuring metrics export frequency

cezarygerard:
- Add new flag for configuring metrics export.
- Modify tests to use naive fake for controller metrics 

This will allow us to increase export frequency to comply with sampling theorem: https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem


bowei:
Can you put the reason to make this a flag instead of just changing the constant?

bowei:
(Add to commit message)

cezarygerard:
Added the rationale to the commit message

PR #1695, opened 2022-04-13
Avoid deleting firewall rules for health check when firewall rule is …

cezarygerard:
…used by some existing health check

NetLB and ILB use different health check but the same firewall rules for health checks
this may result in deleting firewall rules for running health check of the other type

alternative solution to the problem: https://github.com/kubernetes/ingress-gce/pull/1695](https://github.com/kubernetes/ingress-gce/pull/1696

### DIFF
--- a/pkg/healthchecks/healthchecks_l4.go
+++ b/pkg/healthchecks/healthchecks_l4.go
@@ -50,9 +50,19 @@ func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4Re
 	if !sharedHC {
 		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
 	}
-	nn := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
-	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, nn, sharedHC, hcPath, hcPort, scope, l4Type)
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
 	return hcLink, hcFwName, hcPort, hcName, err
+}
+
+// HealthCheckDeleted checks if given health check was deleted
+func HealthCheckDeleted(cloud *gce.Cloud, hcName string, scope meta.KeyType) (bool, error) {
+	key, err := composite.CreateKey(cloud, hcName, scope)
+	if err != nil {
+		return false, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", hcName, err)
+	}
+	_, err = composite.GetHealthCheck(cloud, key, meta.VersionGA)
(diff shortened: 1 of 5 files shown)


pkg/healthchecks/healthchecks_l4.go
```diff
@@ -50,11 +50,21 @@ func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4Re
...
 	}
-	nn := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
-	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, nn, sharedHC, hcPath, hcPort, scope, l4Type)
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
 	return hcLink, hcFwName, hcPort, hcName, err
 }
 
+// HealthCheckDeleted checks if given health check was deleted
+func HealthCheckDeleted(cloud *gce.Cloud, hcName string, scope meta.KeyType) (bool, error) {
```
kl52752, line 59: I think that it checks if given HealthCheck even exist. It can not even be created right?

pkg/loadbalancers/l4.go
```diff
@@ -150,37 +150,70 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncRe
...
+		return "", nil
+	}
+	// Health check deleted, now delete the firewall rule
+	return l.deleteHealthCheckFirewall(hcName, hcFwName, sharedHC)
+}
+
+func (l *L4) deleteHealthCheckFirewall(hcName, hcFwName string, sharedHC bool) (string, error) {
+	if sharedHC {
+		// Check if XLB health check is present
+		deleted, err := healthchecks.HealthCheckDeleted(l.cloud, hcName, meta.Regional)
```
kl52752, line 187: This should be `exists` not deleted it is confusing because your are deleting HC for ILB and here you check HC for XLB and you don't want to delete it just checks if it exists because it is needed. 
So function and the variable should says exist not deleted.

pkg/loadbalancers/l4.go
```diff
@@ -150,37 +150,70 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncRe
...
+			klog.V(2).Infof("Failed to delete regional health check firewall rule %s: health check in use.", hcName)
+			return "", nil
 		}
 	}
-	for _, isShared := range []bool{true, false} {
-		deleteHcFunc(isShared)
+
+	// Delete healthcheck firewall rule if no healthcheck uses the firewall rule
+	err := l.deleteFirewall(hcFwName)
+	if err != nil {
```
kl52752, line 200: why not keep it one liner?  
`if err := l.deleteFirewall(hcFwName); err != nil `
It is not that long and readable.

cezarygerard:
/hold
wait with merge until pull/1696 or pull/1695 is deemed a better solution for the problem

PR #1696, opened 2022-04-13
Create new namer for L4 health check and L4 health check forwarding rule

cezarygerard:
the new name for shared(per-node) L4 health check forwarding rule will prevent l4netlb controller from deleting firewall rules used by l4-ilb-subsetting load balancers

alternative solution to the problem: https://github.com/kubernetes/ingress-gce/pull/1695

### DIFF
--- a/pkg/utils/namer/l4_regional_hc_namer.go
+++ b/pkg/utils/namer/l4_regional_hc_namer.go
@@ -1,0 +1,33 @@
+package namer
+
+import (
+	"strings"
+)
+
+const (
+	regionalHcSuffix = "regional"
+)
+
+// L4RegionalHealthCheckNamer is a  namer use only for health check and firewall resources
+// it suffixes firewall rule name with '-regional' string
+type L4RegionalHealthCheckNamer struct {
+	l4namer L4ResourcesNamer
+}
+
+// NewL4RegionalHealthCheckNamer returns new instance of L4RegionalHealthCheckNamer
+func NewL4RegionalHealthCheckNamer(l4namer L4ResourcesNamer) *L4RegionalHealthCheckNamer {
+	l4rhcn := &L4RegionalHealthCheckNamer{
+		l4namer: l4namer,
+	}
+	return l4rhcn
+}
+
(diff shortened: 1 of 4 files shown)


pkg/utils/namer/l4_regional_hc_namer_test.go
```diff
@@ -0,0 +1,56 @@
+package namer
```
kl52752, line 1: add header

pkg/utils/namer/l4_regional_hc_namer.go
```diff
@@ -0,0 +1,33 @@
+package namer
```
kl52752, line 1: add header

pkg/utils/namer/l4_regional_hc_namer.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"strings"
+)
+
+const (
+	regionalHcSuffix = "regional"
+)
+
+// L4RegionalHealthCheckNamer is a  namer use only for health check and firewall resources
```
kl52752, line 11: maybe.`for regional health check and its corresponding firewallrule`?

pkg/utils/namer/l4_regional_hc_namer.go
```diff
@@ -0,0 +1,33 @@
...
+// NewL4RegionalHealthCheckNamer returns new instance of L4RegionalHealthCheckNamer
+func NewL4RegionalHealthCheckNamer(l4namer L4ResourcesNamer) *L4RegionalHealthCheckNamer {
+	l4rhcn := &L4RegionalHealthCheckNamer{
+		l4namer: l4namer,
+	}
+	return l4rhcn
+}
+
+// L4HealthCheck returns the name of the L4 LB Healthcheck and the associated firewall rule.
+// For shared health checks the firewall rule name is suffixed with '-regional' string
```
kl52752, line 26: can you add both patters here for shared and non-shared?

cezarygerard:
/hold
wait with merge until pull/1696 or pull/1695 is deemed a better solution for the problem

freehan:
I think this approach might make more sense than the other one.

PR #1697, opened 2022-04-14
limit number of gce API calls to get forwarding rule

cezarygerard:
this fix prevents forwarding rule check for internal services and  non-loadbalancer type services

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -247,14 +247,28 @@ func (lc *L4NetLBController) hasRBSForwardingRule(svc *v1.Service) bool {
 }
 
 // isRBSBasedService checks if service has either RBS annotation, finalizer or RBSForwardingRule
+// tries to filter out all other service types before proceeding to checking forwarding rule
 func (lc *L4NetLBController) isRBSBasedService(svc *v1.Service) bool {
-	if svc == nil {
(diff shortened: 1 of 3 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -251,6 +251,13 @@ func (lc *L4NetLBController) isRBSBasedService(svc *v1.Service) bool {
 	if svc == nil {
 		return false
 	}
+
+	WantsL4NetLB, _ := annotations.WantsL4NetLB(svc)
```
kl52752, line 255: Didi you tested this change in case XLB -> ILB migration? I can see that it is called before `isDeleted` and I wonder if it will be conflicted with scenario where we want to delete XLB stack when changing to ILB.

kl52752: Actually it is called twice before and in `needsDeletion`

cezarygerard: I have added new unit test to ensure this cases

I have changed the code to keep the forwarding rule check as the last resort

I am closing this conversation as the code snippet above is outdated

please see the new code

aojea:
/ok-to-test
/reopen

@cezarygerard should we move forward with this? seems an important improvement at a quick look, specially on environments with a lots of services

PR #1698, opened 2022-04-14
Do not include User Errors in L4NetLBInError metric

kl52752:
There are errors related to Network Tier that are caused by User misconfigurations and they should be filtered out from ServicesInError metrics.
This errors are:
1. Ip address and forwarding rule Network Tier mismatch
2. User created service in Standard Tier and deleted the annotation, this action will changed desired service Network Tier to PREMIUM but IP address and forwarding rule should be tear down first because we cannot update Network Tier in this resources. We can tear down resources only when Network Tier from annotation mismatch with network tier in IP address or forwarding rule.

### DIFF
--- a/pkg/l4lb/l4netlbcontroller_test.go
+++ b/pkg/l4lb/l4netlbcontroller_test.go
@@ -19,7 +19,6 @@ package l4lb
 import (
 	"context"
 	"fmt"
-	"google.golang.org/api/googleapi"
 	"math/rand"
 	"net/http"
 	"reflect"
@@ -28,6 +27,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"google.golang.org/api/googleapi"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -604,7 +605,7 @@ func TestProcessServiceCreationFailed(t *testing.T) {
 		key, _ := common.KeyFunc(svc)
 		err := lc.sync(key)
 		if err == nil || err.Error() != param.expectedError {
-			t.Errorf("Error mismatch '%v' != '%v'", err.Error(), param.expectedError)
+			t.Errorf("Error mismatch '%v' != '%v'", err, param.expectedError)
 		}
 	}
 }
@@ -876,10 +877,6 @@ func updateAndAssertExternalTrafficPolicy(newSvc *v1.Service, lc *L4NetLBControl
 	return nil
 }
 
-func isWrongNetworkTierError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "does not have the expected network tier")
-}
-
 func TestControllerUserIPWithStandardNetworkTier(t *testing.T) {
 	// Network Tier from User Static Address should match network tier from forwarding rule.
 	// Premium Network Tier is default for creating forwarding rule so if User wants to use Standard Network Tier for Static Address
@@ -893,7 +890,7 @@ func TestControllerUserIPWithStandardNetworkTier(t *testing.T) {
 	key, _ := common.KeyFunc(svc)
 	addUsersStaticAddress(lc, cloud.NetworkTierStandard)
 	// Sync should return error that Network Tier mismatch because we cannot tear User Managed Address.
-	if err := lc.sync(key); !isWrongNetworkTierError(err) {
+	if err := lc.sync(key); !utils.IsNetworkTierError(err) {
 		t.Errorf("Expected error when trying to ensure service with wrong Network Tier, err: %v", err)
 	}
 	svc.Annotations[annotations.NetworkTierAnnotationKey] = string(cloud.NetworkTierStandard)
--- a/pkg/loadbalancers/address_manager.go
+++ b/pkg/loadbalancers/address_manager.go
@@ -165,6 +165,19 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 		return addr.Address, IPAddrManaged, nil
 	}
 
+	if utils.IsNetworkTierMismatchGCEError(reserveErr) {
+		receivedNetworkTier := cloud.NetworkTierPremium
+		if receivedNetworkTier == am.networkTier {
+			// We don't have information of current ephemeral IP address Network Tier since
+			// we try to reserve the address so we need to check against desired Network Tier and set the opposite one.
+			// This is just for error message.
+			receivedNetworkTier = cloud.NetworkTierStandard
+		}
+		resource := fmt.Sprintf("Reserved static IP (%v)", am.name)
+		networkTierError := utils.NewNetworkTierErr(resource, string(am.networkTier), string(receivedNetworkTier))
+		return "", IPAddrUndefined, networkTierError
+	}
+
 	if !utils.IsHTTPErrorCode(reserveErr, http.StatusConflict) && !utils.IsHTTPErrorCode(reserveErr, http.StatusBadRequest) {
 		// If the IP is already reserved:
 		//    by an internal address: a StatusConflict is returned
@@ -187,7 +200,7 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 
 	// Check that the address attributes are as required.
 	if err := am.validateAddress(addr); err != nil {
-		return "", IPAddrUndefined, err
+		return "", IPAddrUndefined, fmt.Errorf("address (%q) validation failed, err: %w", addr.Name, err)
 	}
 
 	if am.isManagedAddress(addr) {
@@ -205,13 +218,13 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 
 func (am *addressManager) validateAddress(addr *compute.Address) error {
 	if am.targetIP != "" && am.targetIP != addr.Address {
-		return fmt.Errorf("address %q does not have the expected IP %q, actual: %q", addr.Name, am.targetIP, addr.Address)
+		return fmt.Errorf("IP mismatch, expected %q, actual: %q", am.targetIP, addr.Address)
 	}
 	if addr.AddressType != string(am.addressType) {
-		return fmt.Errorf("address %q does not have the expected address type %q, actual: %q", addr.Name, am.addressType, addr.AddressType)
+		return fmt.Errorf("address type mismatch, expected %q, actual: %q", am.addressType, addr.AddressType)
 	}
 	if addr.NetworkTier != am.networkTier.ToGCEValue() {
-		return fmt.Errorf("address %q does not have the expected network tier %q, actual: %q", addr.Name, am.networkTier.ToGCEValue(), addr.NetworkTier)
+		return utils.NewNetworkTierErr(fmt.Sprintf("Static IP (%v)", am.name), am.networkTier.ToGCEValue(), addr.NetworkTier)
 	}
 	return nil
 }
@@ -238,7 +251,7 @@ func (am *addressManager) TearDownAddressIPIfNetworkTierMismatch() error {
 	}
 	if addr != nil && addr.NetworkTier != am.networkTier.ToGCEValue() {
 		if !am.isManagedAddress(addr) {
-			return fmt.Errorf("User specific address IP (%v) network tier mismatch %v != %v ", am.targetIP, addr.NetworkTier, am.networkTier)
+			return utils.NewNetworkTierErr(fmt.Sprintf("User specific address IP (%v)", am.name), string(am.networkTier), addr.NetworkTier)
 		}
 		klog.V(3).Infof("Deleting IP address %v because has wrong network tier", am.targetIP)
 		am.svc.DeleteRegionAddress(addr.Name, am.targetIP)
--- a/pkg/loadbalancers/address_manager_test.go
+++ b/pkg/loadbalancers/address_manager_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/mock"
+	"k8s.io/ingress-gce/pkg/test"
 	"k8s.io/ingress-gce/pkg/utils"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
@@ -148,8 +149,11 @@ func TestAddressManagerExternallyOwnedWrongNetworkTier(t *testing.T) {
 	err = svc.ReserveRegionAddress(addr, vals.Region)
 	require.NoError(t, err, "")
 	mgr := newAddressManager(svc, testSvcName, vals.Region, testSubnet, testLBName, targetIP, cloud.SchemeInternal, cloud.NetworkTierPremium)
+	svc.Compute().(*cloud.MockGCE).MockAddresses.InsertHook = test.InsertAddressNetworkErrorHook
 	_, _, err = mgr.HoldAddress()
-	require.Error(t, err, "does not have the expected network tier")
+	if err == nil || !utils.IsNetworkTierError(err) {
+		t.Fatalf("mgr.HoldAddress() = %v, utils.IsNetworkTierError(err) = %t, want %t", err, utils.IsNetworkTierError(err), true)
+	}
 }
 
 // TestAddressManagerExternallyOwned tests the case where the address exists but isn't
--- a/pkg/loadbalancers/forwarding_rules.go
+++ b/pkg/loadbalancers/forwarding_rules.go
@@ -388,6 +388,11 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
 	}
 
 	if existingFwdRule != nil {
+		if existingFwdRule.NetworkTier != fr.NetworkTier {
+			resource := fmt.Sprintf("Forwarding rule (%v)", frName)
+			networkTierMismatchError := utils.NewNetworkTierErr(resource, existingFwdRule.NetworkTier, fr.NetworkTier)
+			return nil, IPAddrUndefined, networkTierMismatchError
+		}
 		equal, err := Equal(existingFwdRule, fr)
 		if err != nil {
 			return existingFwdRule, IPAddrUndefined, err
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -142,6 +142,8 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service) 
 	result.Annotations[annotations.BackendServiceKey] = name
 	fr, ipAddrType, err := l4netlb.ensureExternalForwardingRule(bs.SelfLink)
 	if err != nil {
+		// User can misconfigure the forwarding rule if Network Tier will not match service level Network Tier.
+		result.MetricsState.IsUserError = utils.IsUserError(err)
 		result.GCEResourceInError = annotations.ForwardingRuleResource
 		result.Error = fmt.Errorf("Failed to ensure forwarding rule - %w", err)
 		return result
--- a/pkg/loadbalancers/l4netlb_test.go
+++ b/pkg/loadbalancers/l4netlb_test.go
@@ -38,12 +38,9 @@ import (
 )
 
 const (
-	managedIP    = true
-	unmanagedIP  = false
-	premiumTier  = true
-	standardTier = false
-	usersIP      = "35.10.211.60"
-	userAddrName = "UserStaticAddress"
+	usersIP        = "35.10.211.60"
+	usersIPPremium = "35.10.211.70"
+	userAddrName   = "UserStaticAddress"
 )
 
 func TestEnsureL4NetLoadBalancer(t *testing.T) {
@@ -71,7 +68,7 @@ func TestEnsureL4NetLoadBalancer(t *testing.T) {
 		t.Errorf("Annotations error: %v", err)
 	}
 	assertNetLbResources(t, svc, l4netlb, nodeNames)
-	if err := checkMetrics(result.MetricsState, managedIP, premiumTier); err != nil {
+	if err := checkMetrics(result.MetricsState /*isManaged = */, true /*isPremium = */, true /*isUserError =*/, false); err != nil {
 		t.Errorf("Metrics error: %v", err)
 	}
 }
@@ -279,11 +276,12 @@ func assertNetLbResources(t *testing.T, apiService *v1.Service, l4NetLb *L4NetLB
 	}
 }
 
-func TestMetricsForUserStaticService(t *testing.T) {
+func TestMetricsForStandardNetworkTier(t *testing.T) {
 	nodeNames := []string{"test-node-1"}
 	vals := gce.DefaultTestClusterValues()
 	fakeGCE := getFakeGCECloud(vals)
-	createUserStaticIP(fakeGCE, vals.Region)
+	createUserStaticIPInStandardTier(fakeGCE, vals.Region)
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockForwardingRules.GetHook = test.GetRBSForwardingRuleInStandardTier
 
 	svc := test.NewL4NetLBRBSService(8080)
 	svc.Spec.LoadBalancerIP = usersIP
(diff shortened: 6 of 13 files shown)


pkg/test/gcehooks.go
```diff
@@ -99,6 +100,11 @@ func GetRBSForwardingRule(ctx context.Context, key *meta.Key, m *cloud.MockForwa
 	return true, &fwRule, nil
 }
 
+func GetRBSForwardingRuleInStandardTier(ctx context.Context, key *meta.Key, m *cloud.MockForwardingRules) (bool, *compute.ForwardingRule, error) {
+	fwRule := compute.ForwardingRule{BackendService: bsUrl, LoadBalancingScheme: string(cloud.SchemeExternal), NetworkTier: cloud.NetworkTierStandard.ToGCEValue()}
```
cezarygerard, line 104: not needed, put a dummy string "some_standard_tier_bs"

pkg/utils/utils.go
```diff
@@ -168,6 +169,16 @@ func IsInUsedByError(err error) bool {
 	return strings.Contains(apiErr.Message, "being used by")
 }
 
+func isNetworkMismatchGCEError(err error) bool {
+	var re = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)
```
cezarygerard, line 173: can you unit test this func isNetworkMismatchGCEError?


maybe it would be better to have some more lenient check
just 
"The network tier of "
PLUS

"Address must be the same"

pkg/metrics/metrics.go
```diff
@@ -502,11 +502,17 @@ func (im *ControllerMetrics) computeL4NetLBMetrics() map[feature]int {
...
 		l4NetLBInError:            0,
+		l4NetLBInUserError:        0,
 	}
 
 	for key, state := range im.l4NetLBServiceMap {
 		klog.V(6).Infof("NetLB Service %s has metrics %+v", key, state)
 		counts[l4NetLBService]++
+		if state.IsUserError {
+			counts[l4NetLBInUserError]++
+			// Skip counting other features if the service is in error state.
```
cezarygerard, line 513: // Skip counting other features if the service is in user-caused error state.

pkg/test/gcehooks.go
kl52752: actaully no we need whole url for forwarding rule compression https://github.dev/kubernetes/ingress-gce/blob/cc2910140315638c2da19dca3e9ea5e8f6746d00/pkg/loadbalancers/forwarding_rules.go#L450

pkg/utils/utils.go
kl52752: added UT

pkg/metrics/metrics.go
kl52752: done

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -877,7 +878,7 @@ func updateAndAssertExternalTrafficPolicy(newSvc *v1.Service, lc *L4NetLBControl
 }
 
 func isWrongNetworkTierError(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "does not have the expected network tier")
+	return err != nil && strings.Contains(err.Error(), "network tier mismatch")
```
freehan, line 881: consider switching to use the error wrapping/unwrapping mechanism to evaluate error types?

https://golangbyexample.com/wrapping-and-unwrapping-error-golang/#Unwrap_an_error

pkg/utils/utils.go
freehan: I am curious if there is a GCE error type specific for this in the GCE go client?

pkg/loadbalancers/address_manager.go
```diff
@@ -163,6 +163,8 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 
 		klog.V(4).Infof("%v: successfully created address %q which reserved IP %q", am.logPrefix, addr.Name, addr.Address)
 		return addr.Address, IPAddrManaged, nil
+	} else if utils.IsNetworkTierError(reserveErr) {
```
freehan, line 166: move this section to below line 151. 
Consider something like this
```
const ErrNetTierMismatch = fmt.Errorf("Address network tier mismatch")


if reserveErr != nil {
     return "", IPAddrUndefined, fmt.Errorf("Expect network tier = %s, %w", am.networkTier, ErrNetTierMismatch)
}
```

pkg/loadbalancers/l4netlb.go
```diff
@@ -142,6 +142,8 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 	result.Annotations[annotations.BackendServiceKey] = name
 	fr, ipAddrType, err := l4netlb.ensureExternalForwardingRule(bs.SelfLink)
 	if err != nil {
+		// User can misconfigure the forwarding rule if Network Tier will not match desired.
+		result.MetricsState.IsUserError = utils.IsNetworkTierError(err)
```
freehan, line 146: I would suggest have a function to evaluate error 
```
func IsUserError(err Error) bool {
   if utils.IsNetworkTierError(err) {
       return true
   }
   return false
}
```

pkg/loadbalancers/address_manager.go
kl52752: actually all errors are validated below so this else is just not necessary.

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/l4lb/l4netlbcontroller_test.go
kl52752: done, or at least I tried to :)

pkg/utils/utils.go
```diff
@@ -168,6 +171,28 @@ func IsInUsedByError(err error) bool {
 	return strings.Contains(apiErr.Message, "being used by")
 }
 
+// IsNetworkTierMismatchGCEError checks if error is a GCE network tier mismatch for external IP
+func IsNetworkTierMismatchGCEError(err error) bool {
```
cezarygerard, line 175: make this func return false if err == nil

pkg/utils/utils.go
```diff
@@ -168,6 +171,28 @@ func IsInUsedByError(err error) bool {
...
+	var re = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)
+	return re.MatchString(err.Error())
+}
+
+// IsNetworkTierError checks if wrapped error is a Network Tier Mismatch error
+func IsNetworkTierError(err error) bool {
```
cezarygerard, line 181: make this func return false if err == nil

this will simplify such expressions in your tests:

 if result.Error == nil || !utils.IsNetworkTierError(result.Error)

pkg/utils/utils.go
kl52752: I don't like this idea the controller always checks if returned error is not nil then we checks specific errors if needed. Returning error here will only simplified test code but in controller there will be a duplication.

pkg/utils/utils.go
kl52752: I don't like this idea the controller always checks if returned error is not nil then we checks specific errors if needed. Returning error here will only simplified test code but in controller there will be a duplication.

bowei:
Can we fix the commit message to have enough context to understand the change etc.

pkg/utils/utils.go
```diff
@@ -168,6 +171,28 @@ func IsInUsedByError(err error) bool {
...
+	netTierError := errors.Unwrap(err)
```
bowei, line 182: We should use `errors.As(..)` to unwrap errors. This code assumes that there is only one level of wrapping for the error?

pkg/utils/utils_test.go
```diff
@@ -1221,3 +1221,60 @@ func TestMinMaxPortRangeAndProtocol(t *testing.T) {
 		}
 	}
 }
+
+func TestIsNetworkMismatchGCEError(t *testing.T) {
+	for _, tc := range []struct {
+		err      error
+		expected bool
```
bowei, line 1228: [minor] expected => want

pkg/utils/utils_test.go
```diff
@@ -1221,3 +1221,60 @@ func TestMinMaxPortRangeAndProtocol(t *testing.T) {
...
+			err:      fmt.Errorf("The network tier of external IP is"),
+			expected: false,
+		},
+		{
+			err:      fmt.Errorf("Some dummy string"),
+			expected: false,
+		},
+	} {
+		if IsNetworkTierMismatchGCEError(tc.err) != tc.expected {
+			t.Errorf("isNetworkMismatchGCEError for error %s should return %v", tc.err.Error(), tc.expected)
```
bowei, line 1252: if got := IsNetworkTierMismatchGCEError(tc.err); got != tc.want {
  t.Errorf("IsNetworkTierMismatchGCEError(%v) = %v, want %v", tc.err, got, tc.want)
}

pkg/utils/utils_test.go
```diff
@@ -1221,3 +1221,60 @@ func TestMinMaxPortRangeAndProtocol(t *testing.T) {
...
+		}
+	}
+}
+
+func TestIsNetworkMismatchError(t *testing.T) {
+	netTierMismatchError := fmt.Errorf("%s err: expected PREMIUM", NetworkTierMismatchStr)
+	for _, tc := range []struct {
+		err      error
+		expected bool
```
bowei, line 1261: expected => want

pkg/utils/utils_test.go
```diff
@@ -1221,3 +1221,60 @@ func TestMinMaxPortRangeAndProtocol(t *testing.T) {
...
+			err:      fmt.Errorf("%s err: wrong error is not wraped", netTierMismatchError),
+			expected: false,
+		},
+		{
+			err:      fmt.Errorf("Wrong error."),
+			expected: false,
+		},
+	} {
+		if IsNetworkTierError(tc.err) != tc.expected {
+			t.Errorf("isNetworkMismatchGCEError for error %s should return %v", tc.err.Error(), tc.expected)
```
bowei, line 1277: See above:

```
if got := ...; got != tc.want {
  ...
}
```

pkg/test/gcehooks.go
```diff
@@ -30,6 +30,7 @@ import (
 
 const (
 	FwIPAddress = "10.0.0.1"
+	bsUrl       = "https://www.googleapis.com/compute/v1/projects/test-project/regions/us-central1/backendServices/k8s2-axyqjz2d-default-netbtest-hgray14h"
```
bowei, line 33: should document this -- is the string significant or just a random string.

pkg/loadbalancers/address_manager.go
```diff
@@ -165,6 +165,11 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 		return addr.Address, IPAddrManaged, nil
 	}
 
+	if utils.IsNetworkTierMismatchGCEError(reserveErr) {
+		wrongNetworkTierEWrror := fmt.Errorf("network tier mismatch err:%w", reserveErr)
```
bowei, line 169: space between err: %w

pkg/loadbalancers/address_manager.go
```diff
@@ -165,6 +165,11 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
...
+		return "", IPAddrUndefined, fmt.Errorf("failed to reserve address %q in Network Tier %s, err: %w", am.name, am.networkTier, wrongNetworkTierEWrror)
```
bowei, line 170: Seems like overkill to wrap an error twice?

pkg/loadbalancers/address_manager.go
```diff
@@ -238,7 +243,8 @@ func (am *addressManager) TearDownAddressIPIfNetworkTierMismatch() error {
 	}
 	if addr != nil && addr.NetworkTier != am.networkTier.ToGCEValue() {
 		if !am.isManagedAddress(addr) {
-			return fmt.Errorf("User specific address IP (%v) network tier mismatch %v != %v ", am.targetIP, addr.NetworkTier, am.networkTier)
+			networkTierMismatchError := fmt.Errorf("%s %v != %v ", utils.NetworkTierMismatchStr, addr.NetworkTier, am.networkTier)
+			return fmt.Errorf("Error tearing down the User specific address IP (%v) err: %w", am.networkTier, networkTierMismatchError)
```
bowei, line 247: Do we need to wrap errors twice?

pkg/loadbalancers/address_manager_test.go
```diff
@@ -149,7 +149,9 @@ func TestAddressManagerExternallyOwnedWrongNetworkTier(t *testing.T) {
 	require.NoError(t, err, "")
 	mgr := newAddressManager(svc, testSvcName, vals.Region, testSubnet, testLBName, targetIP, cloud.SchemeInternal, cloud.NetworkTierPremium)
 	_, _, err = mgr.HoldAddress()
-	require.Error(t, err, "does not have the expected network tier")
+	if err == nil || !utils.IsNetworkTierError(err) {
+		t.Fatalf("Hold Address should return Network Tier Mismatch error")
```
bowei, line 153: mgr.HoldAddress() = %v, utils.IsNetworkTierError(err)  = %t, want %t

is this test supposed to pass if err == nil?

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -388,6 +388,10 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
 	}
 
 	if existingFwdRule != nil {
+		if existingFwdRule.NetworkTier != fr.NetworkTier {
+			networkTierError := fmt.Errorf("%s, old: %s, new: %s", utils.NetworkTierMismatchStr, existingFwdRule.NetworkTier, fr.NetworkTier)
```
bowei, line 392: Double wrapped error.

Why not 

err := fmt.Errorf("...")
return nil, err

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -38,12 +38,15 @@ import (
 )
 
 const (
-	managedIP    = true
-	unmanagedIP  = false
-	premiumTier  = true
-	standardTier = false
-	usersIP      = "35.10.211.60"
-	userAddrName = "UserStaticAddress"
+	managedIP      = true
```
bowei, line 41: I find this stylistically quite confusing.
There is nothing to prevent accidentally using managedIP in place of premiumTier.

Can we just document at the func call site instead of using these constants?

e.g. 

```
if err := checkMetrics(..., /*managedIP*/=true, ...)
```

pkg/utils/utils.go
bowei: If we are doing MustCompile -- I recommend initializing it in `init()` so the compilation failure will be detected immediately

pkg/loadbalancers/address_manager_test.go
kl52752: No, this test checks that error was returned and this is NetworkTierError.
Error msg fixed

pkg/loadbalancers/forwarding_rules.go
kl52752: because I wanted to include which resource returned the "NetworkTierError".
changed.

pkg/loadbalancers/l4netlb_test.go
kl52752: changed

pkg/utils/utils.go
kl52752: done

pkg/utils/utils.go
kl52752: done

pkg/utils/utils_test.go
kl52752: done

pkg/utils/utils_test.go
kl52752: done

pkg/utils/utils_test.go
kl52752: done

pkg/test/gcehooks.go
kl52752: this is a significant url, was craeted based on project and region set in test. We need whole url for forwarding rule validation.

pkg/loadbalancers/address_manager.go
kl52752: obsolete, done

pkg/loadbalancers/address_manager.go
kl52752: fixed

pkg/loadbalancers/forwarding_rules.go
kl52752: done

pkg/loadbalancers/address_manager_test.go
kl52752: done

pkg/test/gcehooks.go
kl52752: done

pkg/loadbalancers/address_manager.go
```diff
@@ -165,6 +165,19 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 		return addr.Address, IPAddrManaged, nil
 	}
 
+	if utils.IsNetworkTierMismatchGCEError(reserveErr) {
+		receivedNetworkTier := cloud.NetworkTierPremium
+		if receivedNetworkTier == am.networkTier {
+			// We don't have information of current ephemeral IP address Network Tier since
+			// we tries to reserve the address so we need to check against desired Network Tier and set the opposite one.
+			// This is just for error message.
```
panslava, line 173: Sorry, I do not finally get was this code is doing...
So if `am.networkTier == cloud.NetworkTierPremium` we change  `receivedNetworkTier` to standard, but is not address manager only used for premium tier? 

(also small thing, but should be "we try")

pkg/loadbalancers/address_manager.go
```diff
@@ -187,7 +200,7 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 
 	// Check that the address attributes are as required.
 	if err := am.validateAddress(addr); err != nil {
-		return "", IPAddrUndefined, err
+		return "", IPAddrUndefined, fmt.Errorf("address (%q) validation fails err: %w", addr.Name, err)
```
panslava, line 203: sorry for nitpicking, but probably something like `address (%q) validation failed, err: %w` would be better to understand

pkg/loadbalancers/address_manager.go
```diff
@@ -205,13 +218,14 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
 
 func (am *addressManager) validateAddress(addr *compute.Address) error {
 	if am.targetIP != "" && am.targetIP != addr.Address {
-		return fmt.Errorf("address %q does not have the expected IP %q, actual: %q", addr.Name, am.targetIP, addr.Address)
+		return fmt.Errorf("IP mismatch expected %q, actual: %q", am.targetIP, addr.Address)
```
panslava, line 221: sorry for nitpicking, 
but probably would be better to add comma
`IP mismatch, expected %q, actual: %q`

pkg/loadbalancers/address_manager.go
```diff
@@ -205,13 +218,14 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
...
 	}
 	if addr.AddressType != string(am.addressType) {
-		return fmt.Errorf("address %q does not have the expected address type %q, actual: %q", addr.Name, am.addressType, addr.AddressType)
+		return fmt.Errorf("address type mismatch expected %q, actual: %q", am.addressType, addr.AddressType)
```
panslava, line 224: sorry for nitpicking,
but probably would be better to add comma
`address type mismatch, expected %q, actual: %q`

pkg/loadbalancers/address_manager.go
```diff
@@ -238,7 +252,9 @@ func (am *addressManager) TearDownAddressIPIfNetworkTierMismatch() error {
 	}
 	if addr != nil && addr.NetworkTier != am.networkTier.ToGCEValue() {
 		if !am.isManagedAddress(addr) {
-			return fmt.Errorf("User specific address IP (%v) network tier mismatch %v != %v ", am.targetIP, addr.NetworkTier, am.networkTier)
+			resource := fmt.Sprintf("User specific address IP (%v)", am.name)
+			networkTierMismatchError := utils.NewNetworkTierErr(resource, string(am.networkTier), addr.NetworkTier)
+			return networkTierMismatchError
```
panslava, line 257: I can be wrong with this, but I have a feeling, that writing this in one line makes it actually easier to read, cause you immediately understand that we just return error
however, it is long line... so discussible

```
return utils.NewNetworkTierErr(fmt.Sprintf("User specific address IP (%v)", am.name), string(am.networkTier), addr.NetworkTier)
```

pkg/loadbalancers/address_manager.go
```diff
@@ -205,13 +218,14 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
...
+		return fmt.Errorf("IP mismatch expected %q, actual: %q", am.targetIP, addr.Address)
 	}
 	if addr.AddressType != string(am.addressType) {
-		return fmt.Errorf("address %q does not have the expected address type %q, actual: %q", addr.Name, am.addressType, addr.AddressType)
+		return fmt.Errorf("address type mismatch expected %q, actual: %q", am.addressType, addr.AddressType)
 	}
 	if addr.NetworkTier != am.networkTier.ToGCEValue() {
-		return fmt.Errorf("address %q does not have the expected network tier %q, actual: %q", addr.Name, am.networkTier.ToGCEValue(), addr.NetworkTier)
+		resource := fmt.Sprintf("Static IP (%v)", am.name)
+		return utils.NewNetworkTierErr(resource, am.networkTier.ToGCEValue(), addr.NetworkTier)
```
panslava, line 228: I can be wrong with this, but I have a feeling, that writing this in one line makes it actually easier to read, cause you immediately understand that we just return error
however, it is long line... so discussible

```
return utils.NewNetworkTierErr(fmt.Sprintf("Static IP (%v)", am.name), string(am.networkTier), addr.NetworkTier)
```

pkg/loadbalancers/l4netlb.go
```diff
@@ -142,6 +142,8 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 	result.Annotations[annotations.BackendServiceKey] = name
 	fr, ipAddrType, err := l4netlb.ensureExternalForwardingRule(bs.SelfLink)
 	if err != nil {
+		// User can misconfigure the forwarding rule if Network Tier will not match desired.
```
panslava, line 145: Can you add a bit of clarification in this comment about "will not match desired ... what"

pkg/utils/utils.go
```diff
@@ -91,6 +92,28 @@ const (
 	LabelAlphaNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
 )
 
+var re *regexp.Regexp
+
+func init() {
+	re = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)
+}
```
panslava, line 99: I understand, Bowei commented about this, and asked to put into init(), but is there any benefit to do it in init, instead of just call regexp.MustCompile on global level?

Something like

```var re = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)```

File: pkg/utils/utils.go [hunk 29]
```diff
@@ -91,6 +92,28 @@ const (
...
+
+// NetworkTierError is a struct to define error caused by User misconfiguration of Network Tier.
+type NetworkTierError struct {
+	resource   string
+	desiredNT  string
+	receivedNT string
+}
+
+// Error function print out Network Tier error.
```
panslava, line 108: [nitpicking] 
should be `prints`

pkg/utils/utils.go
```diff
@@ -168,6 +191,32 @@ func IsInUsedByError(err error) bool {
...
+	var netTierError *NetworkTierError
+	return errors.As(err, &netTierError)
+	// if strings.Contains(err.Error(), NetworkTierMismatchStr) {
+	// 	return true
+	// }
+	// netTierError := errors.Unwrap(err)
+	// if netTierError == nil {
+	// 	return false
+	// }
+	// return strings.Contains(netTierError.Error(), NetworkTierMismatchStr)
```
panslava, line 210: should these comments be removed?

panslava (changes requested):

pkg/loadbalancers/address_manager.go
kl52752: This is just for logging purposes. The thing is that we tried to reserve an ephemeral IP address but this action returned an error that Network Tier mismatch. We do not get the address before this action so we don't know what the actual Network Tier is. What we know is that we tried to reserve the address with Desired (standard or premium) Network Tier. So if we want to log that Network Tier mismatch and print it like `actual: %v != desired %v` we need to set actual to the opposite one we chose to reserve address in since there was a mismatch. I know that this is a bit mixed up but it is only for logging purposes.

pkg/loadbalancers/address_manager.go
kl52752: done

pkg/loadbalancers/address_manager.go
kl52752: done

pkg/loadbalancers/address_manager.go
kl52752: done

pkg/loadbalancers/address_manager.go
kl52752: done

pkg/loadbalancers/address_manager.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/utils/utils.go
kl52752: because `init()` function is call only once when package is first loaded. Changing `var re = regexp.MustCompile` in global scope MustCompalie would be called multiple times (just like it was before).

pkg/utils/utils.go
kl52752: done

pkg/utils/utils.go
kl52752: YES:) done

cezarygerard:

kl52752:
@bowei all comments are resolved PTAL

pkg/utils/utils.go
```diff
@@ -91,6 +92,28 @@ const (
 	LabelAlphaNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
 )
 
+var re *regexp.Regexp
```
bowei, line 95: we shouldn't call this `re` -- that is too generic to be a variable at package scope.

networkTierErrorRegexp

pkg/utils/utils.go
```diff
@@ -91,6 +92,28 @@ const (
...
+
+func init() {
+	re = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)
```
bowei, line 98: I think you can just do

var (
  networkTierErrorRegexp = regexp.MustCompile...

bowei (commented):
minor comments

pkg/loadbalancers/address_manager.go
```diff
@@ -165,6 +165,19 @@ func (am *addressManager) ensureAddressReservation() (string, IPAddressType, err
...
+	if utils.IsNetworkTierMismatchGCEError(reserveErr) {
+		receivedNetworkTier := cloud.NetworkTierPremium
+		if receivedNetworkTier == am.networkTier {
+			// We don't have information of current ephemeral IP address Network Tier since
+			// we tries to reserve the address so we need to check against desired Network Tier and set the opposite one.
+			// This is just for error message.
+			receivedNetworkTier = cloud.NetworkTierStandard
+		}
+		resource := fmt.Sprintf("Reserved static IP (%v)", am.name)
+		networkTierError := utils.NewNetworkTierErr(resource, string(am.networkTier), string(receivedNetworkTier))
```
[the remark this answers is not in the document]
kl52752: I kind of did, I just need to pass what resource is in error and wha are expected values and there are differen resources like IP address or forwarding rule

PR #1699, opened 2022-04-15
Modify reporting error metrics from L4 RBS services -  only report service in error if resync deadline elapsed

kl52752:
Add resync timestamp to service metric state
    
    When service sync ends with an error we wants to try to resync it for
    some period of time and after that time if result is still error we will
    increment L4NetLBSInError metric.
    
    Retry period is fixed 20min. It is becasue resync period is 10min.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -82,6 +82,7 @@ var (
 		NegGCPeriod                      time.Duration
 		NodePortRanges                   PortRanges
 		ResyncPeriod                     time.Duration
+		L4NetLBProvisionDeadline         time.Duration
 		NumL4Workers                     int
 		NumIngressWorkers                int
 		RunIngressController             bool
@@ -204,6 +205,8 @@ the pod secrets for creating a Kubernetes client.`)
 		`Path to kubeconfig file with authorization and master location information.`)
 	flag.DurationVar(&F.ResyncPeriod, "sync-period", 30*time.Second,
 		`Relist and confirm cloud resources this often.`)
+	flag.DurationVar(&F.L4NetLBProvisionDeadline, "l4-netlb-provision-deadline", 20*time.Minute,
+		`Deadline latency for L4 NetLB provisioning.`)
 	flag.IntVar(&F.NumL4Workers, "num-l4-workers", 5,
 		`Number of parallel L4 Service worker goroutines.`)
 	flag.IntVar(&F.NumIngressWorkers, "num-ingress-workers", 1,
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -375,7 +375,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.L4
 		syncResult.Error = fmt.Errorf("failed to set resource annotations, err: %w", err)
 		return syncResult
 	}
-	syncResult.MetricsState.InSuccess = true
+	syncResult.SetMetricsForSuccessfulService()
 	return syncResult
 }
 
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -66,6 +66,12 @@ type L4NetLBSyncResult struct {
 	StartTime          time.Time
 }
 
+// SetMetricsForSuccessfulService should be call after successful sync.
+func (r *L4NetLBSyncResult) SetMetricsForSuccessfulService() {
+	r.MetricsState.FirstSyncErrorTime = nil
+	r.MetricsState.InSuccess = true
+}
+
 // NewL4NetLB creates a new Handler for the given L4NetLB service.
 func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder, lock *sync.Mutex) *L4NetLB {
 	l4netlb := &L4NetLB{cloud: cloud,
@@ -100,7 +106,8 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service) 
 	result := &L4NetLBSyncResult{
 		Annotations: make(map[string]string),
 		StartTime:   time.Now(),
-		SyncType:    SyncTypeCreate}
+		SyncType:    SyncTypeCreate,
+	}
 
 	// If service already has an IP assigned, treat it as an update instead of a new Loadbalancer.
 	if len(svc.Status.LoadBalancer.Ingress) > 0 {
--- a/pkg/metrics/l4metrics_test.go
+++ b/pkg/metrics/l4metrics_test.go
@@ -19,6 +19,7 @@ package metrics
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 )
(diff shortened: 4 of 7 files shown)


pkg/metrics/metrics.go
```diff
@@ -38,6 +38,8 @@ const (
 	versionVar = "INGRESS_VERSION"
 	// Dummy float so we can used bool based timeseries
 	versionValue = 1.0
+	// l4rbsResyncPeriod defines the time in which we expect that L4 NetLB service will be sync successfully.
+	l4rbsResyncPeriod = 20 * time.Minute
```
cezarygerard, line 42: 1) maybe this should be a flag?
2) rename? maybe l4rbsSyncDeadline? or maxL4rbsSyncPeriod?

pkg/metrics/types.go
```diff
@@ -79,6 +81,8 @@ type L4NetLBServiceState struct {
 	IsPremiumTier bool
 	// InSuccess specifies if the NetLB service VIP is configured.
 	InSuccess bool
+	// RetryStartTime specifies the time timestamp when the service started retry sync.
+	RetryStartTime *time.Time
```
cezarygerard, line 85: FirstSyncErrorTime?

pkg/metrics/metrics.go
kl52752: done

pkg/metrics/types.go
kl52752: done

pkg/flags/flags.go
```diff
@@ -204,6 +205,8 @@ the pod secrets for creating a Kubernetes client.`)
 		`Path to kubeconfig file with authorization and master location information.`)
 	flag.DurationVar(&F.ResyncPeriod, "sync-period", 30*time.Second,
 		`Relist and confirm cloud resources this often.`)
+	flag.DurationVar(&F.L4NetLBProvisionMaxPeriod, "l4-netlb-provision-max-period", 20*time.Minute,
+		`Time during which the L4 NetLB service should be provision.`)
```
bowei, line 209: Deadline latency for L4 NetLB provisioning.

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -456,6 +456,9 @@ func (lc *L4NetLBController) publishMetrics(result *loadbalancers.L4NetLBSyncRes
 	switch result.SyncType {
 	case loadbalancers.SyncTypeCreate, loadbalancers.SyncTypeUpdate:
 		klog.V(4).Infof("External L4 Loadbalancer for Service %s ensured, updating its state %v in metrics cache", namespacedName, result.MetricsState)
+		if !result.MetricsState.InSuccess {
+			result.MetricsState.FirstSyncErrorTime = &result.StartTime
```
bowei, line 460: I find it unintuitive to hide setting this inside PublishMetrics.
My expectation is that `result` would be `const` inside this method (yes golang does not have such a thing).

Why not set this value where InSuccess is being set?

pkg/metrics/l4metrics_test.go
```diff
@@ -282,17 +291,31 @@ func TestComputeL4NetLBMetrics(t *testing.T) {
...
 			serviceStates: []L4NetLBServiceState{
-				newL4NetLBServiceState(isSuccess, unmanaged, standard),
-				newL4NetLBServiceState(isSuccess, unmanaged, premium),
-				newL4NetLBServiceState(isSuccess, unmanaged, premium),
-				newL4NetLBServiceState(isSuccess, managed, premium),
-				newL4NetLBServiceState(isSuccess, managed, premium),
-				newL4NetLBServiceState(isSuccess, managed, standard),
-				newL4NetLBServiceState(isError, managed, premium),
-				newL4NetLBServiceState(isError, managed, standard),
+				newL4NetLBServiceState(isSuccess, unmanaged, standard /*resyncTime = */, nil),
```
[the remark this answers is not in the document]
kl52752: done

pkg/flags/flags.go
kl52752: done

pkg/l4lb/l4netlbcontroller.go
bowei: My recommendation would be to have a method that sets MetricStat.InSuccess and set both values at the same time if the fields are meant to be sync'd up.
I stand by my comment that callers of this method will not expect that PublishMetrics modifieds its argument...

kl52752: done

cezarygerard:

kl52752:
@bowei all comments are resolved PTAL

pkg/flags/flags.go
```diff
@@ -204,6 +205,8 @@ the pod secrets for creating a Kubernetes client.`)
 		`Path to kubeconfig file with authorization and master location information.`)
 	flag.DurationVar(&F.ResyncPeriod, "sync-period", 30*time.Second,
 		`Relist and confirm cloud resources this often.`)
+	flag.DurationVar(&F.L4NetLBProvisionMaxPeriod, "l4-netlb-provision-max-period", 20*time.Minute,
```
bowei, line 208: l4-netlb-provision-deadline

(match the words in the comment, flag, etc to make this easy to maintain).

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -375,10 +375,16 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.L4
 		syncResult.Error = fmt.Errorf("failed to set resource annotations, err: %w", err)
 		return syncResult
 	}
-	syncResult.MetricsState.InSuccess = true
+	setMetricsForSuccessfulService(syncResult)
 	return syncResult
 }
 
+func setMetricsForSuccessfulService(syncResult *loadbalancers.L4NetLBSyncResult) {
```
bowei, line 382: Prefer methods if the variables being manipulated are local to the struct

```
func (r*L4NetLBSyncResult) setSuccessfulSync() {
  r.MetricsState.FirstSyncErrorTime = nil
	r.MetricsState.InSuccess = true
}
```

pkg/loadbalancers/l4netlb.go
```diff
@@ -66,6 +66,17 @@ type L4NetLBSyncResult struct {
 	StartTime          time.Time
 }
 
+// newSyncResult creates L4NetLBSyncResult for L4NetLB sync function.
+func newSyncResult(syncType string) *L4NetLBSyncResult {
+	result := &L4NetLBSyncResult{
+		Annotations: make(map[string]string),
+		StartTime:   time.Now(),
+		SyncType:    syncType}
```
bowei, line 74: \n 
}

pkg/flags/flags.go
kl52752: done

pkg/loadbalancers/l4netlb.go
kl52752: done

pkg/metrics/l4metrics_test.go
```diff
@@ -38,6 +39,12 @@ const (
 	disableCustomSubnet = false
 )
 
+var (
+	currTime    = time.Now()
+	before5min  = currTime.Add(-5 * time.Minute)
+	before25min = currTime.Add(-25 * time.Minute)
```
bowei, line 45: We should be careful that this doesn't result in a flaky test. For example, the t.Parallel() blocks waiting for all the single threaded tests to run first and then runs all the parallel etc together.

This means there may be a non-trivial amount of time between the var init and the tests running.

bowei: Isn't it safer to take the curTime and curTime + X in the test itself?

kl52752: But our function (newL4NetLBServiceState) is taking the reference of time and you cannot take address of time.Now() or time.Now().Add(). 
Do you think it would be better to create local variables for each test?

bowei: ye

kl52752: done

Issue #1702, opened 2022-05-03
Link to gke-self-managed.sh broken


kundan2707:
LInk for gke-self-managed.sh  in https://github.com/kubernetes/ingress-gce/tree/master/docs/deploy/gke#readme is broken

swetharepakula:
Thanks for catching and fixing this @kundan2707 !

PR #1703, opened 2022-05-03
corrected link for gke-self-managed.sh

kundan2707:
Corrected broken link for gke-self-managed.sh in https://github.com/kubernetes/ingress-gce/tree/master/docs/deploy/gke#important-prerequisites

#1702


swetharepakula (approved):
Thanks for your change @kundan2707 !

/lgtm
/approve

PR #1705, opened 2022-05-05
Rewrite L4 healthchecks creation and deletion

cezarygerard:
- create a common singleton like struct fot l4 health checks
- new struct holds mutex for common resources
- delete shared healtcheck firewall rules safely

without proper guarding of shared healtcheck firewall rules, the NetLB controller may delete the ILB healthcheck firewall or vice versa

### DIFF
--- a/hack/run-local-glbc.sh
+++ b/hack/run-local-glbc.sh
@@ -43,5 +43,7 @@ ${GLBC} \
     --running-in-cluster=false \
     --logtostderr --v=${V} \
     --config-file-path=${GCECONF} \
+    --run-l4-controller \
+    --run-l4-netlb-controller \
     "${@}" \
     2>&1 | tee -a /tmp/glbc.log
--- a/pkg/firewalls/firewalls_l4.go
+++ b/pkg/firewalls/firewalls_l4.go
@@ -50,7 +50,7 @@ func EnsureL4FirewallRule(cloud *gce.Cloud, nsName string, params *FirewallParam
 	if err != nil {
 		return err
 	}
-	fwDesc, err := utils.MakeL4LBServiceDescription(nsName, params.IP, meta.VersionGA, sharedRule, params.L4Type)
+	fwDesc, err := utils.MakeL4LBFirewallDescription(nsName, params.IP, meta.VersionGA, sharedRule)
 	if err != nil {
 		klog.Warningf("EnsureL4FirewallRule(%v): failed to generate description for L4 %s rule, err: %v", params.Name, params.L4Type.ToString(), err)
 	}
@@ -104,8 +104,8 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
 }
 
 func firewallRuleEqual(a, b *compute.Firewall) bool {
-	return a.Description == b.Description &&
-		len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
+	// let's skip description not to trigger flood of updates
+	return len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
 		a.Allowed[0].IPProtocol == b.Allowed[0].IPProtocol &&
 		utils.EqualStringSets(a.Allowed[0].Ports, b.Allowed[0].Ports) &&
 		utils.EqualStringSets(a.SourceRanges, b.SourceRanges) &&
--- a/pkg/healthchecks/healthchecks_l4.go
+++ b/pkg/healthchecks/healthchecks_l4.go
@@ -18,6 +18,11 @@ package healthchecks
 
 import (
 	"fmt"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/events"
+	"k8s.io/ingress-gce/pkg/firewalls"
+	"strconv"
+	"sync"
 
 	cloudprovider "github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -42,26 +47,85 @@ const (
 	gceHcUnhealthyThreshold = int64(3)
 )
 
+var instance *l4HealthChecks
+
+type l4HealthChecks struct {
+	mutex           sync.Mutex
+	cloud           *gce.Cloud
+	recorderFactory events.RecorderProducer
+}
+
+func Initialize(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
+	instance = &l4HealthChecks{
+		cloud:           cloud,
+		recorderFactory: recorderFactory,
+	}
+}
+func Fake(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
+	return &l4HealthChecks{
+		cloud:           cloud,
+		recorderFactory: recorderFactory,
+	}
+}
+
+func GetL4() *l4HealthChecks {
+	return instance
+}
+
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
+	// mutex?
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
 	if !sharedHC {
 		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
+		l4hc.mutex.Lock()
+		defer l4hc.mutex.Unlock()
+
 	}
-	nn := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
-	_, hcLink, err := ensureL4HealthCheckInternal(cloud, hcName, nn, sharedHC, hcPath, hcPort, scope, l4Type)
-	return hcLink, hcFwName, hcPort, hcName, err
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	_, hcLink, err := l4hc.ensureL4HealthCheckInternal(hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
+	if err != nil {
+		return "", "", "", annotations.HealthcheckResource, err
+	}
+	err = l4hc.ensureFirewall(svc, hcFwName, hcPort, sharedHC, nodeNames)
+	if err != nil {
+		return "", "", "", annotations.FirewallForHealthcheckResource, err
+	}
+
+	return hcLink, hcFwName, hcName, "", err
 }
 
-func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
-	selfLink := ""
-	key, err := composite.CreateKey(cloud, name, scope)
-	if err != nil {
-		return nil, selfLink, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", name, err)
+func (l4hc *l4HealthChecks) DeleteHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error) {
+	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	if sharedHC {
+		l4hc.mutex.Lock()
+		defer l4hc.mutex.Unlock()
 	}
-	hc, err := composite.GetHealthCheck(cloud, key, meta.VersionGA)
+	NamespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	err := utils.IgnoreHTTPNotFound(l4hc.deleteHealthCheck(hcName, scope))
+	if err != nil {
+		if !utils.IsInUsedByError(err) {
+			klog.Errorf("Failed to delete healthcheck for service %s - %v", NamespacedName.String(), err)
+			return annotations.HealthcheckResource, err
+		}
+		// Ignore deletion error due to health check in use by another resource.
+		// This will be hit if this is a shared healthcheck.
+		klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
+		return "", nil
+	}
+	// Health check deleted, now delete the firewall rule
+	return l4hc.deleteHealthCheckFirewall(svc, hcName, hcFwName, sharedHC, l4Type)
+}
+
+func (l4hc *l4HealthChecks) ensureL4HealthCheckInternal(name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
+	selfLink := ""
+	key, err := composite.CreateKey(l4hc.cloud, name, scope)
+	if err != nil {
+		return nil, selfLink, fmt.Errorf("Failed to create key for for healthcheck with name %s for service %s", name, svcName.String())
+	}
+	hc, err := composite.GetHealthCheck(l4hc.cloud, key, meta.VersionGA)
 	if err != nil {
 		if !utils.IsNotFoundError(err) {
 			return nil, selfLink, err
@@ -69,17 +133,17 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
 	}
 	var region string
 	if scope == meta.Regional {
-		region = cloud.Region()
+		region = l4hc.cloud.Region()
 	}
 	expectedHC := NewL4HealthCheck(name, svcName, shared, path, port, l4Type, scope, region)
 	if hc == nil {
 		// Create the healthcheck
 		klog.V(2).Infof("Creating healthcheck %s for service %s, shared = %v", name, svcName, shared)
-		err = composite.CreateHealthCheck(cloud, key, expectedHC)
+		err = composite.CreateHealthCheck(l4hc.cloud, key, expectedHC)
 		if err != nil {
 			return nil, selfLink, err
 		}
-		selfLink = cloudprovider.SelfLink(meta.VersionGA, cloud.ProjectID(), "healthChecks", key)
+		selfLink = cloudprovider.SelfLink(meta.VersionGA, l4hc.cloud.ProjectID(), "healthChecks", key)
 		return expectedHC, selfLink, nil
 	}
 	selfLink = hc.SelfLink
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
 	}
 	mergeHealthChecks(hc, expectedHC)
 	klog.V(2).Infof("Updating healthcheck %s for service %s", name, svcName)
-	err = composite.UpdateHealthCheck(cloud, key, expectedHC)
+	err = composite.UpdateHealthCheck(l4hc.cloud, key, expectedHC)
 	if err != nil {
 		return nil, selfLink, err
 	}
 	return expectedHC, selfLink, err
 }
 
-func DeleteHealthCheck(cloud *gce.Cloud, name string, scope meta.KeyType) error {
-	key, err := composite.CreateKey(cloud, name, scope)
+func (l4hc *l4HealthChecks) ensureFirewall(svc *corev1.Service, hcFwName string, hcPort int32, sharedHC bool, nodeNames []string) error {
+	// Add firewall rule for healthchecks to nodes
+	hcFWRParams := firewalls.FirewallParams{
+		PortRanges:   []string{strconv.Itoa(int(hcPort))},
+		SourceRanges: gce.L4LoadBalancerSrcRanges(),
+		Protocol:     string(corev1.ProtocolTCP),
+		Name:         hcFwName,
+		NodeNames:    nodeNames,
+	}
+	err := firewalls.EnsureL4LBFirewallForHc(svc, sharedHC, &hcFWRParams, l4hc.cloud, &l4hc.mutex, l4hc.recorderFactory.Recorder(svc.Namespace))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (l4hc *l4HealthChecks) deleteHealthCheck(name string, scope meta.KeyType) error {
+	// always check mutex
+	key, err := composite.CreateKey(l4hc.cloud, name, scope)
 	if err != nil {
 		return fmt.Errorf("Failed to create composite key for healthcheck %s - %w", name, err)
 	}
-	return composite.DeleteHealthCheck(cloud, key, meta.VersionGA)
+	return composite.DeleteHealthCheck(l4hc.cloud, key, meta.VersionGA)
+}
+
+func (l4hc *l4HealthChecks) deleteHealthCheckFirewall(svc *corev1.Service, hcName, hcFwName string, sharedHC bool, l4Type utils.L4LBType) (string, error) {
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+
+	safeToDelete, err := l4hc.healthCheckFirewallSafeToDelete(hcName, sharedHC, l4Type)
+	if err != nil {
+		klog.Errorf("Failed to delete health check firewall rule %s for service %s - %v", hcFwName, namespacedName.String(), err)
+		return annotations.HealthcheckResource, err
+	}
+	if !safeToDelete {
+		klog.V(2).Infof("Failed to delete health check firewall rule %s: health check in use.", hcName)
+		return "", nil
+	}
+	// Delete healthcheck firewall rule if no healthcheck uses the firewall rule
+	err = l4hc.deleteFirewall(hcFwName, svc)
+	if err != nil {
+		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, namespacedName.String(), err)
+		return annotations.FirewallForHealthcheckResource, err
+	}
+	return "", nil
+}
+
+func (l4hc *l4HealthChecks) healthCheckFirewallSafeToDelete(hcName string, sharedHC bool, l4Type utils.L4LBType) (bool, error) {
+	if !sharedHC {
+		return true, nil
+	}
+	var scopeToCheck meta.KeyType
+	scopeToCheck = meta.Regional
+	if l4Type == utils.XLB {
+		scopeToCheck = meta.Global
+	}
+	key, err := composite.CreateKey(l4hc.cloud, hcName, scopeToCheck)
+	if err != nil {
+		return false, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", hcName, err)
+	}
+	_, err = composite.GetHealthCheck(l4hc.cloud, key, meta.VersionGA)
+	return utils.IsNotFoundError(err), nil
+}
+
+func (l4hc *l4HealthChecks) deleteFirewall(name string, svc *corev1.Service) error {
+	err := firewalls.EnsureL4FirewallRuleDeleted(l4hc.cloud, name)
+	if err != nil {
+		if fwErr, ok := err.(*firewalls.FirewallXPNError); ok {
+			recorder := l4hc.recorderFactory.Recorder(svc.Namespace)
+			recorder.Eventf(svc, corev1.EventTypeNormal, "XPN", fwErr.Message)
+			return nil
+		}
+		return err
+	}
+	return nil
 }
 
 func NewL4HealthCheck(name string, svcName types.NamespacedName, shared bool, path string, port int32, l4Type utils.L4LBType, scope meta.KeyType, region string) *composite.HealthCheck {
--- a/pkg/loadbalancers/interfaces.go
+++ b/pkg/loadbalancers/interfaces.go
@@ -18,7 +18,10 @@ package loadbalancers
 
 import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/networking/v1"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/namer"
 )
 
 // LoadBalancerPool is an interface to manage the cloud resources associated
@@ -36,4 +39,12 @@ type LoadBalancerPool interface {
 	Shutdown(ings []*v1.Ingress) error
 	// HasUrlMap returns true if an URL map exists in GCE for given ingress.
 	HasUrlMap(ing *v1.Ingress) (bool, error)
+}
+
+// L4HealthChecks defines methods for creating adn deleting health checks (and their firewall rules) for l4 services
+type L4HealthChecks interface {
+	// EnsureL4HealthCheck creates health check (and firewall rule) for l4 service
+	EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error)
+	// DeleteHealthCheck deletes health check (and firewall rule) for l4 service
+	DeleteHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error)
 }
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -18,9 +18,7 @@ package loadbalancers
 
 import (
 	"fmt"
-	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
@@ -49,11 +47,11 @@ type L4 struct {
 	scope       meta.KeyType
 	namer       namer.L4ResourcesNamer
 	// recorder is used to generate k8s Events.
-	recorder            record.EventRecorder
-	Service             *corev1.Service
-	ServicePort         utils.ServicePort
-	NamespacedName      types.NamespacedName
-	sharedResourcesLock *sync.Mutex
+	recorder       record.EventRecorder
+	Service        *corev1.Service
+	ServicePort    utils.ServicePort
+	NamespacedName types.NamespacedName
+	l4HealthChecks L4HealthChecks
 }
 
 // L4ILBSyncResult contains information about the outcome of an L4 ILB sync. It stores the list of resource name annotations,
@@ -69,8 +67,15 @@ type L4ILBSyncResult struct {
 }
 
 // NewL4Handler creates a new L4Handler for the given L4 service.
-func NewL4Handler(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder, lock *sync.Mutex) *L4 {
-	l := &L4{cloud: cloud, scope: scope, namer: namer, recorder: recorder, Service: service, sharedResourcesLock: lock}
+func NewL4Handler(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder) *L4 {
+	l := &L4{
+		cloud:          cloud,
+		scope:          scope,
+		namer:          namer,
+		recorder:       recorder,
+		Service:        service,
+		l4HealthChecks: healthchecks.GetL4(),
+	}
 	l.NamespacedName = types.NamespacedName{Name: service.Name, Namespace: service.Namespace}
 	l.backendPool = backends.NewPool(l.cloud, l.namer)
 	l.ServicePort = utils.ServicePort{ID: utils.ServicePortID{Service: l.NamespacedName}, BackendNamer: l.namer,
@@ -150,37 +155,26 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncRe
 	// when externalTrafficPolicy is changed from Local to Cluster and a new health check was created.
 	// When service is deleted we need to check both health checks shared and non-shared
 	// and delete them if needed.
-	deleteHcFunc := func(sharedHC bool) {
-		hcName, hcFwName := l.namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
-		if sharedHC {
-			l.sharedResourcesLock.Lock()
-			defer l.sharedResourcesLock.Unlock()
-		}
-		err = utils.IgnoreHTTPNotFound(healthchecks.DeleteHealthCheck(l.cloud, hcName, meta.Global))
-		if err != nil {
-			if !utils.IsInUsedByError(err) {
-				klog.Errorf("Failed to delete healthcheck for internal loadbalancer service %s - %v", l.NamespacedName.String(), err)
-				result.GCEResourceInError = annotations.HealthcheckResource
-				result.Error = err
-				return
-			}
-			// Ignore deletion error due to health check in use by another resource.
-			// This will be hit if this is a shared healthcheck.
-			klog.V(2).Infof("Failed to delete healthcheck %s: health check in use.", hcName)
-		} else {
-			// Delete healthcheck firewall rule if healthcheck deletion is successful.
-			err = deleteFwFunc(hcFwName)
-			if err != nil {
-				klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, l.NamespacedName.String(), err)
-				result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-				result.Error = err
-			}
-		}
-	}
 	for _, isShared := range []bool{true, false} {
-		deleteHcFunc(isShared)
+		resourceInError, err := l.l4HealthChecks.DeleteHealthCheck(svc, l.namer, isShared, meta.Global, utils.ILB)
+		if err != nil {
+			result.GCEResourceInError = resourceInError
+			result.Error = err
+		}
 	}
 	return result
+}
+
+func (l *L4) deleteFirewall(name string) error {
+	err := firewalls.EnsureL4FirewallRuleDeleted(l.cloud, name)
+	if err != nil {
+		if fwErr, ok := err.(*firewalls.FirewallXPNError); ok {
+			l.recorder.Eventf(l.Service, corev1.EventTypeNormal, "XPN", fwErr.Message)
+			return nil
+		}
+		return err
+	}
+	return nil
 }
 
 // GetFRName returns the name of the forwarding rule for the given ILB service.
@@ -221,18 +215,10 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 
 	// create healthcheck
 	sharedHC := !helpers.RequestsOnlyLocalTraffic(l.Service)
-	ensureHCFunc := func() (string, string, int32, string, error) {
-		if sharedHC {
-			// Take the lock when creating the shared healthcheck
-			l.sharedResourcesLock.Lock()
-			defer l.sharedResourcesLock.Unlock()
-		}
-		return healthchecks.EnsureL4HealthCheck(l.cloud, l.Service, l.namer, sharedHC, meta.Global, utils.ILB)
-	}
+	hcLink, hcFwName, hcName, resourceInErr, err := l.l4HealthChecks.EnsureL4HealthCheck(l.Service, l.namer, sharedHC, meta.Global, utils.ILB, nodeNames)
 
-	hcLink, hcFwName, hcPort, hcName, err := ensureHCFunc()
 	if err != nil {
-		result.GCEResourceInError = annotations.HealthcheckResource
+		result.GCEResourceInError = resourceInErr
 		result.Error = err
 		return result
 	}
@@ -253,7 +239,6 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 		Protocol:     string(protocol),
 		Name:         name,
 		NodeNames:    nodeNames,
-		L4Type:       utils.ILB,
 	}
 
 	if err := firewalls.EnsureL4LBFirewallForNodes(l.Service, &nodesFWRParams, l.cloud, l.recorder); err != nil {
@@ -262,22 +247,6 @@ func (l *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service)
 		return result
 	}
 	result.Annotations[annotations.FirewallRuleKey] = name
-
-	// Add firewall rule for healthchecks to nodes
-	hcFWRParams := firewalls.FirewallParams{
-		PortRanges:   []string{strconv.Itoa(int(hcPort))},
-		SourceRanges: gce.L4LoadBalancerSrcRanges(),
-		Protocol:     string(corev1.ProtocolTCP),
-		Name:         hcFwName,
-		NodeNames:    nodeNames,
-		L4Type:       utils.ILB,
-	}
-	err = firewalls.EnsureL4LBFirewallForHc(l.Service, sharedHC, &hcFWRParams, l.cloud, l.sharedResourcesLock, l.recorder)
-	if err != nil {
-		result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-		result.Error = err
-		return result
-	}
 	result.Annotations[annotations.FirewallRuleForHealthcheckKey] = hcFwName
 
 	// Check if protocol has changed for this service. In this case, forwarding rule should be deleted before
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -18,8 +18,6 @@ package loadbalancers
 
 import (
 	"fmt"
-	"strconv"
-	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
@@ -47,11 +45,11 @@ type L4NetLB struct {
 	scope       meta.KeyType
 	namer       namer.L4ResourcesNamer
 	// recorder is used to generate k8s Events.
-	recorder            record.EventRecorder
-	Service             *corev1.Service
-	ServicePort         utils.ServicePort
-	NamespacedName      types.NamespacedName
-	sharedResourcesLock *sync.Mutex
+	recorder       record.EventRecorder
+	Service        *corev1.Service
+	ServicePort    utils.ServicePort
+	NamespacedName types.NamespacedName
+	l4HealthChecks L4HealthChecks
 }
 
 // L4NetLBSyncResult contains information about the outcome of an L4 NetLB sync. It stores the list of resource name annotations,
@@ -67,15 +65,15 @@ type L4NetLBSyncResult struct {
 }
 
 // NewL4NetLB creates a new Handler for the given L4NetLB service.
-func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder, lock *sync.Mutex) *L4NetLB {
+func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder) *L4NetLB {
 	l4netlb := &L4NetLB{cloud: cloud,
-		scope:               scope,
-		namer:               namer,
-		recorder:            recorder,
-		Service:             service,
-		sharedResourcesLock: lock,
-		NamespacedName:      types.NamespacedName{Name: service.Name, Namespace: service.Namespace},
-		backendPool:         backends.NewPool(cloud, namer),
+		scope:          scope,
+		namer:          namer,
+		recorder:       recorder,
+		Service:        service,
+		NamespacedName: types.NamespacedName{Name: service.Name, Namespace: service.Namespace},
+		backendPool:    backends.NewPool(cloud, namer),
+		l4HealthChecks: healthchecks.GetL4(),
 	}
 	portId := utils.ServicePortID{Service: l4netlb.NamespacedName}
 	l4netlb.ServicePort = utils.ServicePort{
@@ -84,6 +82,7 @@ func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, n
 		NodePort:     int64(service.Spec.Ports[0].NodePort),
 		L4RBSEnabled: true,
 	}
+
 	return l4netlb
 }
 
@@ -108,25 +107,19 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service) 
 	}
 
 	l4netlb.Service = svc
+
 	sharedHC := !helpers.RequestsOnlyLocalTraffic(svc)
-	ensureHCFunc := func() (string, string, int32, string, error) {
-		if sharedHC {
-			// Take the lock when creating the shared healthcheck
-			l4netlb.sharedResourcesLock.Lock()
-			defer l4netlb.sharedResourcesLock.Unlock()
-		}
-		return healthchecks.EnsureL4HealthCheck(l4netlb.cloud, l4netlb.Service, l4netlb.namer, sharedHC, l4netlb.scope, utils.XLB)
-	}
-	hcLink, hcFwName, hcPort, hcName, err := ensureHCFunc()
+	hcLink, hcFwName, hcName, resourceInErr, err := l4netlb.l4HealthChecks.EnsureL4HealthCheck(l4netlb.Service, l4netlb.namer, sharedHC, l4netlb.scope, utils.XLB, nodeNames)
+
 	if err != nil {
-		result.GCEResourceInError = annotations.HealthcheckResource
+		result.GCEResourceInError = resourceInErr
 		result.Error = fmt.Errorf("Failed to ensure health check %s - %w", hcName, err)
 		return result
 	}
 	result.Annotations[annotations.HealthcheckKey] = hcName
 
 	name := l4netlb.ServicePort.BackendName()
-	protocol, res := l4netlb.createFirewalls(name, hcLink, hcFwName, hcPort, nodeNames, sharedHC)
+	protocol, res := l4netlb.createFirewalls(name, nodeNames)
 	if res.Error != nil {
 		return res
 	}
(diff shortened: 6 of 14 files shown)


pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
+	// mutex?
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
 	if !sharedHC {
 		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
+		l4hc.mutex.Lock()
```
kl52752, line 83: You should lock the mutex if health check is shared otherwise sync is not necessary.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
+	return hcLink, hcFwName, hcName, "", err
+}
+
+func (l4hc *l4HealthChecks) DeleteHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error) {
+	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	if sharedHC {
+		l4hc.mutex.Lock()
+		defer l4hc.mutex.Unlock()
+	}
+	NamespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
```
kl52752, line 106: why camel case not `namespacedName`?

pkg/healthchecks/healthchecks_l4.go
cezarygerard: good catch!

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
 }
 
-func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
+func (l4hc *l4HealthChecks) ensureL4HealthCheckInternal(name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
 	selfLink := ""
-	key, err := composite.CreateKey(cloud, name, scope)
+	key, err := composite.CreateKey(l4hc.cloud, name, scope)
 	if err != nil {
-		return nil, selfLink, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", name, err)
+		return nil, selfLink, fmt.Errorf("Failed to create key for for healthcheck with name %s for service %s", name, svcName.String())
```
kl52752, line 126: remove duplicated `for`, add namespace also

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	err = composite.UpdateHealthCheck(l4hc.cloud, key, expectedHC)
 	if err != nil {
 		return nil, selfLink, err
 	}
 	return expectedHC, selfLink, err
 }
 
-func DeleteHealthCheck(cloud *gce.Cloud, name string, scope meta.KeyType) error {
-	key, err := composite.CreateKey(cloud, name, scope)
+func (l4hc *l4HealthChecks) ensureFirewall(svc *corev1.Service, hcFwName string, hcPort int32, sharedHC bool, nodeNames []string) error {
```
kl52752, line 163: add comment that for ILB and NetLB fwrule for shared HC is also shared (even they have hc in different scope)

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	if err != nil {
+		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, namespacedName.String(), err)
+		return annotations.FirewallForHealthcheckResource, err
+	}
+	return "", nil
+}
+
+func (l4hc *l4HealthChecks) healthCheckFirewallSafeToDelete(hcName string, sharedHC bool, l4Type utils.L4LBType) (bool, error) {
+	if !sharedHC {
+		return true, nil
```
kl52752, line 211: maybe instead of returning bool you can create dedicated error for "NotSafeToDelete"? Just a proposal.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	}
+	var scopeToCheck meta.KeyType
+	scopeToCheck = meta.Regional
```
kl52752, line 214: `scopeToCheck := meta.Regional` no need to split it to 2 lines

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
+
+func GetL4() *l4HealthChecks {
+	return instance
+}
+
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
+	// mutex?
```
kl52752, line 78: delete

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+func (l4hc *l4HealthChecks) ensureFirewall(svc *corev1.Service, hcFwName string, hcPort int32, sharedHC bool, nodeNames []string) error {
+	// Add firewall rule for healthchecks to nodes
+	hcFWRParams := firewalls.FirewallParams{
+		PortRanges:   []string{strconv.Itoa(int(hcPort))},
+		SourceRanges: gce.L4LoadBalancerSrcRanges(),
+		Protocol:     string(corev1.ProtocolTCP),
+		Name:         hcFwName,
+		NodeNames:    nodeNames,
+	}
+	err := firewalls.EnsureL4LBFirewallForHc(svc, sharedHC, &hcFWRParams, l4hc.cloud, &l4hc.mutex, l4hc.recorderFactory.Recorder(svc.Namespace))
```
kl52752, line 172: just return firewalls.Ens...

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	safeToDelete, err := l4hc.healthCheckFirewallSafeToDelete(hcName, sharedHC, l4Type)
+	if err != nil {
+		klog.Errorf("Failed to delete health check firewall rule %s for service %s - %v", hcFwName, namespacedName.String(), err)
+		return annotations.HealthcheckResource, err
+	}
+	if !safeToDelete {
+		klog.V(2).Infof("Failed to delete health check firewall rule %s: health check in use.", hcName)
+		return "", nil
+	}
+	// Delete healthcheck firewall rule if no healthcheck uses the firewall rule
```
kl52752, line 200: add dot at the end

pkg/loadbalancers/interfaces.go
```diff
@@ -37,3 +40,11 @@ type LoadBalancerPool interface {
 	// HasUrlMap returns true if an URL map exists in GCE for given ingress.
 	HasUrlMap(ing *v1.Ingress) (bool, error)
 }
+
+// L4HealthChecks defines methods for creating adn deleting health checks (and their firewall rules) for l4 services
```
kl52752, line 44: typo `adn ` -> `and`

pkg/loadbalancers/interfaces.go
```diff
@@ -37,3 +40,11 @@ type LoadBalancerPool interface {
...
+type L4HealthChecks interface {
+	// EnsureL4HealthCheck creates health check (and firewall rule) for l4 service
+	EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error)
+	// DeleteHealthCheck deletes health check (and firewall rule) for l4 service
```
kl52752, line 48: add comment related to shared resource

pkg/loadbalancers/l4netlb.go
```diff
@@ -84,6 +82,7 @@ func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, n
 		NodePort:     int64(service.Spec.Ports[0].NodePort),
 		L4RBSEnabled: true,
 	}
+
```
kl52752, line 85: remove

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
+}
+
+func GetL4() *l4HealthChecks {
+	return instance
+}
+
 // EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
```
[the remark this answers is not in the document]
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: svcName is of type: types.NamespacedName
the String() method prints namespace either

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +153,87 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	}
+	err := firewalls.EnsureL4LBFirewallForHc(svc, sharedHC, &hcFWRParams, l4hc.cloud, &l4hc.mutex, l4hc.recorderFactory.Recorder(svc.Namespace))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (l4hc *l4HealthChecks) deleteHealthCheck(name string, scope meta.KeyType) error {
+	// always check mutex
```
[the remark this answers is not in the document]
cezarygerard: you mean rename deleteHealthCheck -> delete?

pkg/healthchecks/healthchecks_l4.go
cezarygerard: I think error should be an exceptions state, not the expected logic flow.

pkg/healthchecks/healthchecks_l4.go
cezarygerard: otherwise the following line does not compile

	key, err := composite.CreateKey(l4hc.cloud, hcName, scopeToCheck)

with:
../healthchecks/healthchecks_l4.go:227:38: cannot use scopeToCheck (type string) as type meta.KeyType in argument to composite.CreateKey


meta.KeyType is just redefined string
beauty of golang...

pkg/healthchecks/healthchecks_l4.go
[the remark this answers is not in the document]
cezarygerard: you just mean the comment.
done.

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: deleted the comment

pkg/healthchecks/healthchecks_l4.go
cezarygerard: haha, what a foolish code ;-)

cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/loadbalancers/interfaces.go
cezarygerard: done

pkg/loadbalancers/l4netlb.go
cezarygerard: done

pkg/loadbalancers/interfaces.go
cezarygerard: in interface?

I will add in implementation.

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -55,7 +55,8 @@ func TestEnsureL4NetLoadBalancer(t *testing.T) {
 	svc := test.NewL4NetLBRBSService(8080)
 	namer := namer_util.NewL4Namer(kubeSystemUID, namer_util.NewNamer(vals.ClusterName, "cluster-fw"))
 
-	l4netlb := NewL4NetLB(svc, fakeGCE, meta.Regional, namer, record.NewFakeRecorder(100), &sync.Mutex{})
+	l4netlb := NewL4NetLB(svc, fakeGCE, meta.Regional, namer, record.NewFakeRecorder(100))
+	l4netlb.l4HealthChecks = healthchecks.Fake(fakeGCE, &test.FakeRecorderSource{})
```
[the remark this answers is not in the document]
cezarygerard: The reason I am always using new instance of healthchecks for each test case  (despite it is a singleton in production code) is that test are running in parallel.
If I was using same healthchecks than may assertions would fails because tests use the same service names, healthchecks names, firewall names etc. Asserting on one resource being deleted may fail due to other test creating resource of the same name.
 
Also I need that healthchecks use the same fakeGCE as the rest of the test case.

BTW it is important that all controllers in a single test case use the same fakeHC, I have fixed this in 
- TestHealthCheckFirewallDeletionWithILB
- TestHealthCheckFirewallDeletionWithNetLB

cezarygerard:
@bowei I have just answered all questions from Katarzyna.
Please review this code as we need this before we go GA with RBS feature

pkg/loadbalancers/l4netlb_test.go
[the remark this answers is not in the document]
kl52752: I meant in a single test case.

pkg/loadbalancers/interfaces.go
kl52752: Ack

pkg/healthchecks/healthchecks_l4.go
kl52752: Ack

pkg/healthchecks/healthchecks_l4.go
kl52752: you can just cast it
`scopeToCheck := meta.KeyType(meta.Regional)`

pkg/healthchecks/healthchecks_l4.go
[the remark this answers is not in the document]
kl52752: yes the comment :) sorry for being not precise

pkg/healthchecks/healthchecks_l4.go
kl52752: Ack

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
 	gceHcUnhealthyThreshold = int64(3)
 )
 
-// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
+var instance *l4HealthChecks
```
bowei, line 50: Can you explain the rationale behind adding a singleton. Seems like it mostly just removes the Cloud and recorder argument? If we find that a problem because it's too verbose -- maybe just put those in a single struct and pass that as an argument instead of having to deal with the complexities of having a Singleton?

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
...
+}
+
+func Initialize(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
+	instance = &l4HealthChecks{
+		cloud:           cloud,
+		recorderFactory: recorderFactory,
+	}
+}
+
+func Fake(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
```
bowei, line 65: This seems like it doesn't actually create a fake?

bowei (commented):
Biggest question is why introduce a singleton? Seems like it's mostly to reduce verbosity?

pkg/firewalls/firewalls_l4.go
```diff
@@ -104,8 +102,8 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
 }
 
 func firewallRuleEqual(a, b *compute.Firewall) bool {
-	return a.Description == b.Description &&
-		len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
+	// let's skip description not to trigger flood of updates
```
kl52752, line 105: I'm not sure if we wants to do this. For non-shared HC description contains `ServiceIP: ip`, so if IP will change then the health check description will be obsolete, it probably is needed only for debugging purposes but still it might be confusing.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
 	gceHcUnhealthyThreshold = int64(3)
 )
 
-// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
+var instance *l4HealthChecks
+
+type l4HealthChecks struct {
```
kl52752, line 52: add comment

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
...
+	mutex           sync.Mutex
+	cloud           *gce.Cloud
+	recorderFactory events.RecorderProducer
+}
+
+func Initialize(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
```
kl52752, line 58: Add comment, or maybe rename it to InitializeL4 (or sth) since package is healthcheck

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
...
+}
+
+func Fake(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
+	return &l4HealthChecks{
+		cloud:           cloud,
+		recorderFactory: recorderFactory,
+	}
+}
+
+func GetL4() *l4HealthChecks {
```
kl52752, line 72: Add comment

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,111 @@ const (
...
+// for the healthcheck. If healthcheck is shared (external traffic policy 'cluster') then firewall rules will be shared
+// regardless of scope param.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
 	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
-	if !sharedHC {
+
+	if sharedHC {
```
kl52752, line 84: instead 
```
       hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
	if sharedHC {
		// lock out entire EnsureL4HealthCheck process
		l4hc.mutex.Lock()
		defer l4hc.mutex.Unlock()
	} else {
		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
	}
```
do this 
```

        hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
	if sharedHC {
                hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
		// lock out entire EnsureL4HealthCheck process
		l4hc.mutex.Lock()
		defer l4hc.mutex.Unlock()
	} 
	```

File: pkg/healthchecks/healthchecks_l4.go (hunk 5 above)
Comment (kl52752) [reply pkg/healthchecks/healthchecks_l4.go:217]: Ack

File: pkg/loadbalancers/l4netlb_test.go [hunk 23]
```diff
@@ -146,11 +148,59 @@ func TestDeleteL4NetLoadBalancerWithSharedHC(t *testing.T) {
...
+	result := l4NetLB.EnsureFrontend(nodeNames, netlbSvc)
+	if result.Error != nil {
+		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+	}
+	if len(result.Status.Ingress) == 0 {
+		t.Errorf("Got empty loadBalancer status using handler %v", l4NetLB)
+	}
+	assertNetLbResources(t, netlbSvc, l4NetLB, nodeNames)
+
+	// Delete the NetLB loadbalancer.
```
kl52752, line 183: Can you explain why you delete it before creating ILB resource. 
I thought that this test steps would be like:
1. Create L4NetLB and check that Regioanl HC and fwrule was created
2. Create L4ILB and verify that new HC was created (Global one) and verify that fwrule is the same
3. Delete one of the LB and verify that fwrule was not deleted and one HC was deleted.
Does it make any sense?

pkg/healthchecks/healthchecks_l4.go
cezarygerard: You are right, 
My rationale was to name it Fake so it is clearly only for test purposes.

pkg/loadbalancers/l4netlb_test.go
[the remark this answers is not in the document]
cezarygerard: yes, single test case must use the same l4HealthChecks in every controller

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +47,103 @@ const (
...
+var instance *l4HealthChecks
+
+type l4HealthChecks struct {
+	mutex           sync.Mutex
+	cloud           *gce.Cloud
+	recorderFactory events.RecorderProducer
+}
+
+func Initialize(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
+	instance = &l4HealthChecks{
```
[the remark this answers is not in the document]
cezarygerard: it's run from the main.go, so it should not be run twice

Anyway, I have added the init mutex variable and nil checks to prevent double initialization

```

// InitializeL4 creates singleton instance, must be run before GetL4() func
func InitializeL4(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
	if instance == nil {
		initLock.Lock()
		defer initLock.Unlock()

		if instance == nil {
			instance = &l4HealthChecks{
				cloud:           cloud,
				recorderFactory: recorderFactory,
			}
		}
	}
}

```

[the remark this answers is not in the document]
cezarygerard: Added a mutex guard, 
btw see my comment on singleton pattern

pkg/healthchecks/healthchecks_l4.go
cezarygerard: It is not about cloud and recorder args.

We have a real bug in healthcheck deletion.

Deletion of shared health check tries  to remove firewall rule for health check
and the healthcheck from ILB and NetLB use the same firewall rule

Checking if healthcheck-under-deletion is in use is not enough (deleting NetLB Healthcheck will not fail if ILB Healthcheck is in use)
cross-checking firewalls is still subject to race conditions (one healthcheck is deleted, the other is added ad the same time)

so we need a way to safely healthchecks using the same mutex object across the NetLB controller and ILB controller.

Creating a common  l4HealthChecks struct makes a good place for cross-checking healthchecks and simplifies code in NetLB controller and ILB controller.

cezarygerard: Now the open question whether it is worh

it make initialisation in l4.go and l4netlb.go much easier
but on the other hand the tests setups are more complicated

maybe it is enough to create an instance in main and pass it via context, just like any other single instance struct in this code

pkg/loadbalancers/l4netlb_test.go
cezarygerard: it does not delete before ILB

there is 
```
	_, l4ilb, ilbResult := ensureService(fakeGCE, namer, nodeNames, vals.ZoneName, 8081, t)
```
few lines before
there are assertions after resource creation (also inside the ensureService)


there there are assertions for gcp resources after creation

I have added checks for healthcheck deletion
the other resources deletion is checked in other tests

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/firewalls/firewalls_l4.go
cezarygerard: I have changed this logic, for non shared health checks

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -18,6 +18,11 @@ package healthchecks
 
 import (
 	"fmt"
+	"k8s.io/ingress-gce/pkg/annotations"
```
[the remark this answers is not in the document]
cezarygerard: done

pkg/firewalls/firewalls_l4.go
```diff
@@ -104,8 +104,8 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
...
+	return len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
```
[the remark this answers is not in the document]
cezarygerard: done

changed the logic a little bit for non-shared health checks and firewalls according to https://github.com/kubernetes/ingress-gce/pull/1705#discussion_r866520978

pkg/firewalls/firewalls_l4.go
```diff
@@ -104,8 +104,8 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
 }
 
 func firewallRuleEqual(a, b *compute.Firewall) bool {
-	return a.Description == b.Description &&
-		len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
+	// let's skip description not to trigger flood of updates
```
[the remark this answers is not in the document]
cezarygerard: done

changed the logic a little bit for non-shared health checks and firewalls according to https://github.com/kubernetes/ingress-gce/pull/1705#discussion_r866520978

pkg/healthchecks/healthchecks_l4.go
cezarygerard: I have tried to sketch the old way for dependency injection: from  context down to controllers na handlers
https://github.com/cezarygerard/ingress-gce/pull/3

it does not seem much simpler

bowei:
Can you make sure the Git commits follow good conventions (e.g. have one line + description describing the bug being fixed etc.)

pkg/firewalls/firewalls_l4.go
```diff
@@ -103,13 +103,19 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
 	return nil
 }
 
-func firewallRuleEqual(a, b *compute.Firewall) bool {
-	return a.Description == b.Description &&
-		len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
+func firewallRuleEqual(a, b *compute.Firewall, skipDescription bool) bool {
+	fwrEqual := len(a.Allowed) == 1 &&
+		len(a.Allowed) == len(b.Allowed) &&
```
kl52752, line 108: maybe instaed of creating new variable `fwrEqual` you can do :
```
 .. previous check..
 && (skipDescription || a.Description == b.Description)

File: pkg/healthchecks/healthchecks_l4.go [hunk 29]
```diff
@@ -42,44 +46,125 @@ const (
...
+func FakeL4(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
+	instance = &l4HealthChecks{
+		cloud:           cloud,
+		recorderFactory: recorderFactory,
+	}
+	return instance
+}
+
+// GetL4 returns singleton instance, must be run after InitializeL4
+func GetL4() *l4HealthChecks {
```
kl52752, line 87: why don't you put Init func here? (and make Init function not exported).
Checking `instance == nil` is cheap and then you don't have to worry that someone will call GetL4 before Init function. 
I also think that it would be better to create this object inside context and keep  it there (sinc all controllers have context and context is explicitly initialise before controllers are run so it will be safe) instead of this singleton pattern but this is just my opinion.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+	return instance
+}
+
+// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, and the firewall rule required
+// for the healthcheck. If healthcheck is shared (external traffic policy 'cluster') then firewall rules will be shared
+// regardless of scope param.
```
kl52752, line 93: I think that this comment is not complete. I don't understand (without knowing the context) `firewall rules will be shared regardless of scope param` what is scope param? Just write that ILB creates Global HC and NetLB Regional and firewall rule is shared for both.

pkg/loadbalancers/l4netlb_test.go
kl52752: I see, naming `ensureService` was misleading.
Resolved

cezarygerard:
@bowei done

cezarygerard:

pkg/firewalls/firewalls_l4.go
```diff
@@ -103,13 +103,19 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
 	return nil
 }
 
-func firewallRuleEqual(a, b *compute.Firewall) bool {
-	return a.Description == b.Description &&
-		len(a.Allowed) == 1 && len(a.Allowed) == len(b.Allowed) &&
+func firewallRuleEqual(a, b *compute.Firewall, skipDescription bool) bool {
+	fwrEqual := len(a.Allowed) == 1 &&
```
bowei, line 107: Seems like we are assuming the length is 1?

Should we do something more robust to avoid potential problems in the future wrt to the contract:

```
if len(a.Allowed) != len(b.Allowed) { return false }
for i:= range a.Allowed {
  aa := a.Allowed[i]
  ba := b.Allowed[i]
  if ...
}
if utils.EqualStringSets(...)
if utils.EqualStringSets(...)
if utils.EqualStringSets(...)
if ! skipDescription  { ... }
return true
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+
+type l4HealthChecks struct {
+	mutex           sync.Mutex
+	cloud           *gce.Cloud
+	recorderFactory events.RecorderProducer
+}
+
+// InitializeL4 creates singleton instance, must be run before GetL4() func
+func InitializeL4(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
+	if instance == nil {
```
bowei, line 64: log info here

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+			instance = &l4HealthChecks{
+				cloud:           cloud,
+				recorderFactory: recorderFactory,
+			}
+		}
+	}
+}
+
+// FakeL4 creates instance of l4HealthChecks> USe for test only.
+func FakeL4(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
```
bowei, line 78: does this need to be exported?

pkg/healthchecks/healthchecks_l4.go
bowei: [minor] It's probably more "Go"ish to call this L4()

pkg/healthchecks/healthchecks_l4.go
bowei: +1 we should be very explicit in our comments, otherwise it ends up with more questions than answers:

Suggest (if accurate)

```
EnsureL4HealthCheck and firewall rules exist for the L4
LoadBalancer Service.

The healthcheck and firewall will be shared between different K8s
Services for ExternalTrafficPolicy = Cluster, as the same
configuration is used across all Services of this type.

Firewall rules are always created at in the Global scope (vs
Regional). This means that one Firewall rule is created for
Services of different scope (Global vs Regional).
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
 	gceHcUnhealthyThreshold = int64(3)
 )
 
-// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
+var (
+	// instance is a sinngleton instance, created by InitializeL4
```
bowei, line 50: typo sinngleton

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+	instance *l4HealthChecks
+	// mutex for preventing multiple initialization
```
bowei, line 52: // initLock for preventing ...

pkg/healthchecks/healthchecks_l4.go
bowei, line 64:
```
if instance != nil {
  log.Error  double init
  return
}
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
-	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
-	if !sharedHC {
-		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
+	hcPath, hcPort := helpers.GetServiceHealthCheckPathPort(svc)
+	if sharedHC {
+		hcPath, hcPort = gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
+		// lock out entire EnsureL4HealthCheck process
```
bowei, line 100: Let's make this a lot more descriptive to make it easier for people to understand:

```
// We need to acquire a controller-wide mutex to ensure that in the case of a healthcheck shared between loadbalancers that the sync of the GCE resources is not performed in parallel. 
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+func GetL4() *l4HealthChecks {
+	return instance
+}
+
+// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, and the firewall rule required
+// for the healthcheck. If healthcheck is shared (external traffic policy 'cluster') then firewall rules will be shared
+// regardless of scope param.
 // If the healthcheck already exists, it is updated as needed.
-func EnsureL4HealthCheck(cloud *gce.Cloud, svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, string, int32, string, error) {
+func (l4hc *l4HealthChecks) EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error) {
```
bowei, line 95: If it's not a huge change, we should change the return value to a struct with named fields.

```
type EnsureL4HealthCheckResult struct {
  HCLink string
  ...
}
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
 	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
-	hcPath, hcPort := gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
-	if !sharedHC {
-		hcPath, hcPort = helpers.GetServiceHealthCheckPathPort(svc)
+	hcPath, hcPort := helpers.GetServiceHealthCheckPathPort(svc)
+	if sharedHC {
+		hcPath, hcPort = gce.GetNodesHealthCheckPath(), gce.GetNodesHealthCheckPort()
```
bowei, line 99: we should log.V(3).Infof or higher

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
-	return hcLink, hcFwName, hcPort, hcName, err
+
+	return hcLink, hcFwName, hcName, "", err
 }
 
-func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
+// DeleteHealthCheck deletes health check (and firewall rule) for l4 service. Checks if shared resources are safe to delete.
+func (l4hc *l4HealthChecks) DeleteHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error) {
+	if sharedHC {
+		// lock out entire DeleteHealthCheck process
```
bowei, line 121: same comment as above, make this easier for future maintainers to understand.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
+
+	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	err := utils.IgnoreHTTPNotFound(l4hc.deleteHealthCheck(hcName, scope))
+	if err != nil {
+		if !utils.IsInUsedByError(err) {
+			klog.Errorf("Failed to delete healthcheck for service %s - %v", namespacedName.String(), err)
+			return annotations.HealthcheckResource, err
+		}
+		// Ignore deletion error due to health check in use by another resource.
```
bowei, line 134: Move this comment above the if statement  `if !utils.IsInUsedByError`

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +174,85 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
 	if err != nil {
 		return nil, selfLink, err
 	}
 	return expectedHC, selfLink, err
 }
 
-func DeleteHealthCheck(cloud *gce.Cloud, name string, scope meta.KeyType) error {
-	key, err := composite.CreateKey(cloud, name, scope)
+// ensureFirewall rule for L4 service.
+// The firewall rules are the same for ILB and NetLB that use external traffic policy 'local' (sharedHC == true)
```
bowei, line 185: [minor] clean up comment formatting

```
// ensureFirewall rule for `svc`.
//
// L4 ILB and L4 NetLB Services with ExternalTrafficPolicy=Local use the same firewall
// rule at global scope. 
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +174,85 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	// Delete healthcheck firewall rule if no healthcheck uses the firewall rule.
+	err = l4hc.deleteFirewall(hcFwName, svc)
+	if err != nil {
+		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, namespacedName.String(), err)
+		return annotations.FirewallForHealthcheckResource, err
+	}
+	return "", nil
+}
+
+func (l4hc *l4HealthChecks) healthCheckFirewallSafeToDelete(hcName string, sharedHC bool, l4Type utils.L4LBType) (bool, error) {
```
bowei, line 228: we should name this func closer to what it indicates rather than the interpreted meaning.

isn't this just healthcheckExists()?

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -89,19 +174,85 @@ func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.Na
...
+	if err != nil {
+		return false, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", hcName, err)
+	}
+	_, err = composite.GetHealthCheck(l4hc.cloud, key, meta.VersionGA)
+	return utils.IsNotFoundError(err), nil
+}
+
+func (l4hc *l4HealthChecks) deleteFirewall(name string, svc *corev1.Service) error {
+	err := firewalls.EnsureL4FirewallRuleDeleted(l4hc.cloud, name)
+	if err != nil {
```
bowei, line 247: if err == nil { return nil }

log.Error the error

Also, we should have a comment saying why we ignore the FIrewallXPNError

should we use err.As() instead of the type case?

pkg/loadbalancers/interfaces.go
```diff
@@ -37,3 +40,11 @@ type LoadBalancerPool interface {
 	// HasUrlMap returns true if an URL map exists in GCE for given ingress.
 	HasUrlMap(ing *v1.Ingress) (bool, error)
 }
+
+// L4HealthChecks defines methods for creating and deleting health checks (and their firewall rules) for l4 services
+type L4HealthChecks interface {
+	// EnsureL4HealthCheck creates health check (and firewall rule) for l4 service
+	EnsureL4HealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) (string, string, string, string, error)
```
bowei, line 47: see comment about this return type

pkg/loadbalancers/l4.go
```diff
@@ -150,39 +155,28 @@ func (l *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncRe
...
-		deleteHcFunc(isShared)
-	}
 	return result
 }
 
+func (l *L4) deleteFirewall(name string) error {
+	err := firewalls.EnsureL4FirewallRuleDeleted(l.cloud, name)
+	if err != nil {
+		if fwErr, ok := err.(*firewalls.FirewallXPNError); ok {
+			l.recorder.Eventf(l.Service, corev1.EventTypeNormal, "XPN", fwErr.Message)
```
bowei, line 172: we should probably log info here

pkg/loadbalancers/l4netlb.go
```diff
@@ -211,53 +191,43 @@ func (l4netlb *L4NetLB) EnsureLoadBalancerDeleted(svc *corev1.Service) *L4NetLBS
...
-		} else {
-			// Delete healthcheck firewall rule if healthcheck deletion is successful.
-			err = deleteFwFunc(hcFwName)
-			if err != nil {
-				klog.Errorf("Failed to delete firewall rule %s for service %s - %v", hcFwName, l4netlb.NamespacedName.String(), err)
-				result.GCEResourceInError = annotations.FirewallForHealthcheckResource
-				result.Error = err
-			}
+			result.GCEResourceInError = resourceInError
+			result.Error = err
```
bowei, line 205: Do we need to break if we get an error from this loop?

bowei (commented):
Make sure we have sufficient logging for debugging.

pkg/healthchecks/healthchecks_l4.go
cezarygerard: It is being used in pkg/loadbalancer  unit tests

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,44 +46,125 @@ const (
...
-// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
+var (
+	// instance is a sinngleton instance, created by InitializeL4
+	instance *l4HealthChecks
+	// mutex for preventing multiple initialization
+	initLock = &sync.Mutex{}
+)
+
+type l4HealthChecks struct {
+	mutex           sync.Mutex
```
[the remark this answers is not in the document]
cezarygerard: done.

added comment
renamed to sharedResourcesLock as it was in previous implementation

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: Error, not warning?
ok, done error

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: it is actually inversed

if healthcheck exists then the function will return false
also it always returns true for shared healthchecks

maybe: healthCheckFirewallInUse ?

cezarygerard: but it still inverst entire logic

pkg/loadbalancers/interfaces.go
cezarygerard: fixed

but to declare the struct I must have moved the interfaces to the to the provider package instead of client package to avoid circular dependencies

pkg/healthchecks/healthchecks_l4.go
cezarygerard: fixed

but to declare the return struct I must have moved the interfaces to the to the provider package instead of client package to avoid circular dependencies

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

actually the share healthcheck firewall rule if ExternalTrafficPolicy=Cluster

pkg/loadbalancers/l4netlb.go
cezarygerard: no, customer might have updated the service from ExternalTrafficPolicy=Local to  ExternalTrafficPolicy=Cluster, and now they are deleting the service.

We do not try to cleanup firewall rules during ExternalTrafficPolicy change, so we must cleanup here

cezarygerard:
@bowei 
done and answered all comments

pkg/loadbalancers/l4netlb.go
bowei: can we add a comment describing why

pkg/firewalls/firewalls_l4.go
```diff
@@ -103,13 +103,29 @@ func EnsureL4FirewallRuleDeleted(cloud *gce.Cloud, fwName string) error {
...
+	if len(a.Allowed) != len(b.Allowed) {
+		return false
+	}
+	for i := range a.Allowed {
+		if !allowRulesEqual(a.Allowed[i], b.Allowed[i]) {
+			return false
+		}
+	}
+
+	srcAndTargetEqual := utils.EqualStringSets(a.SourceRanges, b.SourceRanges) &&
```
bowei, line 116: it's easier to maintain if we make this straight line code:

```
if ! utils.EqualStringSets() { return false }
if ! utils.EqualStringSets() { return false }
if ! skipDescription && a.Description != b.Description {
  return false
}
return true
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,242 @@ const (
 	gceHcUnhealthyThreshold = int64(3)
 )
 
-// EnsureL4HealthCheck creates a new HTTP health check for an L4 LoadBalancer service, based on the parameters provided.
+var (
+	// instance is a singleton instance, created by InitializeL4
+	instance *l4HealthChecks
+	// instanceLock to prevent duplicate initialization.
+	instanceLock = &sync.Mutex{}
```
bowei, line 53: go style puts the mutex in the same block as the protected var:

```
struct {
  lock
  var
  var

  nonProtectedVar
  ...
}
```

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,242 @@ const (
...
+)
+
+type l4HealthChecks struct {
+	// Lock access to shared resources - node healthcecks and their firewall rules
```
bowei, line 57: We should be more explicit:

// sharedResourceLock serializes operations on the healthcheck and firewall
// resources shared across multiple Services.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,242 @@ const (
...
+}
+
+// InitializeL4 creates singleton instance, must be run before GetL4() func
+func InitializeL4(cloud *gce.Cloud, recorderFactory events.RecorderProducer) {
+	if instance != nil {
+		klog.Error("Multiple L4 Healthchecks initialization attempts")
+		return
+	}
+
+	instanceLock.Lock()
```
bowei, line 70: You need to put this first, no? the instance != nil needs to be guarded.

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,242 @@ const (
...
+	defer instanceLock.Unlock()
+
+	if instance == nil {
```
bowei, line 73: you don't need this if any more

pkg/healthchecks/healthchecks_l4.go
bowei: can we use this comment?

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,242 @@ const (
...
+	var scopeToCheck meta.KeyType
+	scopeToCheck = meta.Regional
+	if l4Type == utils.XLB {
+		scopeToCheck = meta.Global
+	}
+	key, err := composite.CreateKey(l4hc.cloud, hcName, scopeToCheck)
+	if err != nil {
+		return false, fmt.Errorf("Failed to create composite key for healthcheck %s - %w", hcName, err)
+	}
+	_, err = composite.GetHealthCheck(l4hc.cloud, key, meta.VersionGA)
```
bowei, line 267: Don't we need to take the lock here to avoid something creating in parallel to the deletion?
This has a lock read unlock; A ; lock write unlock issue -- there might be some operation occurring at A that is a race.

pkg/healthchecks/healthchecks_l4.go
cezarygerard: Usually the singleton has 2 checks, 
1) to only lock the mutex if the instance is not initialized
and 
2) to make sure that instance was not initialized between first check and lock acquisition.


The code used to look like this:
```
	if instance == nil {
		initLock.Lock()
		defer initLock.Unlock()

		if instance == nil {
			instance = &l4HealthChecks{
				cloud:           cloud,
				recorderFactory: recorderFactory,
			}
		}
	}
```

in the comnent https://github.com/kubernetes/ingress-gce/pull/1705#discussion_r874400663
you asked me to log double init attempts

so I changed the above code into

```
	if instance != nil {
		klog.Error("Multiple L4 Healthchecks initialization attempts")
		return
	}

	instanceLock.Lock()
	defer instanceLock.Unlock()

	if instance == nil {
		instance = &l4HealthChecks{
			cloud:           cloud,
			recorderFactory: recorderFactory,
		}
		klog.V(3).Infof("Initialized L4 Healthchecks")
	}

```
which is equivalent of double checked locking.

I can just lock out entire function, 

```
	instanceLock.Lock()
	defer instanceLock.Unlock()
	
	if instance != nil {
		klog.Error("Multiple L4 Healthchecks initialization attempts")
		return
	}

	instance = &l4HealthChecks{
		cloud:           cloud,
		recorderFactory: recorderFactory,
	}
	klog.V(3).Infof("Initialized L4 Healthchecks")
```


as InitializeL4 is only called once (not in every healthchecks.GetL4() call ) 

iare you ok with the last option?

pkg/healthchecks/healthchecks_l4.go
cezarygerard: TBH, I don't get this race issue here
Could  you please elaborate?

please note that EnsureL4HealthCheck and DeleteHealthCheck execute 
l4hc.sharedResourcesLock.Lock()
very early before any API calls

so 
between healthCheckFirewallSafeToDelete call and actual deletion, the sharedResourcesLock is still possessed by the same thread

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: done

pkg/firewalls/firewalls_l4.go
cezarygerard: done

hack/run-local-glbc.sh
```diff
@@ -43,5 +43,7 @@ ${GLBC} \
     --running-in-cluster=false \
     --logtostderr --v=${V} \
     --config-file-path=${GCECONF} \
+    --run-l4-controller \
```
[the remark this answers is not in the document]
cezarygerard: undone the change

pkg/healthchecks/healthchecks_l4.go
cezarygerard: @kl52752 
- The l4HealthChecks needs recorder factory which is not easily accessible from l4.go and l4netlb.go
- I have tried to create L4HealthChecks in context but I get import cycle  error due to pkg/firewalls using the context

@bowei 
Changed.

pkg/loadbalancers/l4netlb.go
cezarygerard: added more comment line here
there is also full rationale in the comments just before the for loop

pkg/firewalls/firewalls_l4.go
cezarygerard: refactored this whole func

pkg/healthchecks/healthchecks_l4.go
cezarygerard: yes, used your comment

pkg/firewalls/firewalls_l4.go
cezarygerard: done

pkg/healthchecks/healthchecks_l4.go
cezarygerard: logged the error, added comment

The type cast code has been moved from l4.go/l4netlb.go without changing the logic, therefore I am hesitant to change to errors.As(...)

cezarygerard:
@bowei @kl52752 
I have answered all your comments

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,248 @@ const (
...
+	}
+	if !safeToDelete {
+		klog.V(3).Infof("Failed to delete health check firewall rule %s: health check in use.", hcName)
+		return "", nil
+	}
+	klog.V(3).Infof("Deleting healthcheck firewall rule named: %s", hcFwName)
+	// Delete healthcheck firewall rule if no healthcheck uses the firewall rule.
+	err = l4hc.deleteFirewall(hcFwName, svc)
+	if err != nil {
+		klog.Errorf("Failed to delete firewall rule %s for internal loadbalancer service %s, err %v", hcFwName, namespacedName.String(), err)
```
panslava, line 253: I think we should remove "internal" from this log

pkg/healthchecks/healthchecks_l4.go
```diff
@@ -42,69 +46,248 @@ const (
...
 }
 
-func ensureL4HealthCheckInternal(cloud *gce.Cloud, name string, svcName types.NamespacedName, shared bool, path string, port int32, scope meta.KeyType, l4Type utils.L4LBType) (*composite.HealthCheck, string, error) {
+// DeleteHealthCheck deletes health check (and firewall rule) for l4 service. Checks if shared resources are safe to delete.
+func (l4hc *l4HealthChecks) DeleteHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error) {
+
+	hcName, hcFwName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+
+	klog.V(3).Infof("Trying to delete L4 healthcheck: %s and firewall rule %s from service %s/%s, shared: %v", hcName, hcFwName, svc.Name, svc.Namespace, sharedHC)
```
panslava, line 149: You can use `namespacedName.String()` instead `svc.Namespace, svc.Name`

pkg/healthchecks/healthchecks_l4.go
cezarygerard: good catch

PR #1707, opened 2022-05-06
Truncate nodes list to maximum instance group size (1000) before adding to instance group

panslava:
This fixes a bug for clusters bigger than instance group max size, cause we were trying to add more than allowed number of nodes to instance group and gcp api returned an error

### DIFF
--- a/pkg/instances/instances.go
+++ b/pkg/instances/instances.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/api/compute/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/ingress-gce/pkg/events"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 
@@ -333,6 +334,23 @@ func (i *Instances) Sync(nodes []string) (err error) {
 			return err
 		}
 		kubeNodes := sets.NewString(nodes...)
+
+		// Individual InstanceGroup has a limit for 1000 instances in it.
+		// As a result, it's not possible to add more to it.
+		if len(kubeNodes) > flags.F.MaxIgSize {
+			// List() will return a sorted list so the kubeNodesList truncation will have a stable set of nodes.
+			kubeNodesList := kubeNodes.List()
+
+			// Store first 10 truncated nodes for logging
+			truncatedNodesSample := kubeNodesList[flags.F.MaxIgSize:]
+			maxTruncatedNodesSampleSize := 10
+			if len(truncatedNodesSample) > maxTruncatedNodesSampleSize {
+				truncatedNodesSample = truncatedNodesSample[:maxTruncatedNodesSampleSize]
+			}
+
+			klog.Warningf("Total number of kubeNodes: %d, truncating to maximum Instance Group size = %d. Instance group name: %s. First %d truncated instances: %v", len(kubeNodesList), flags.F.MaxIgSize, igName, len(truncatedNodesSample), truncatedNodesSample)
+			kubeNodes = sets.NewString(kubeNodesList[:flags.F.MaxIgSize]...)
+		}
 
 		// A node deleted via kubernetes could still exist as a gce vm. We don't
 		// want to route requests to it. Similarly, a node added to kubernetes
--- a/pkg/instances/instances_test.go
+++ b/pkg/instances/instances_test.go
@@ -23,6 +23,7 @@ import (
 
 	"google.golang.org/api/compute/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/test"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 )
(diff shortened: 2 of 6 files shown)


pkg/instances/instances_test.go
```diff
@@ -116,6 +119,42 @@ func TestNodePoolSync(t *testing.T) {
 		t.Fatalf(
 			"Did not expect any calls, got %+v", fakeIGs.calls)
 	}
+
+	// KubeNodes: n1, n2, ..., n1000, n1001
+	// GCENodes:
```
kl52752, line 124: Can you add test where GCENodes are not empty and some might not be in KubeNodes list?

pkg/instances/instances_test.go
```diff
@@ -116,6 +119,42 @@ func TestNodePoolSync(t *testing.T) {
...
+	// Try to add 1001 instances to the instance group. Should truncate to 1000 instances
+	fakeIGs = NewFakeInstanceGroups(map[string]IGsToInstances{
+		defaultZone: {
+			ig: sets.NewString(),
+		},
+	})
+	pool = newNodePool(fakeIGs, defaultZone)
+	pool.EnsureInstanceGroupsAndPorts(defaultNamer.InstanceGroup(), []int64{80})
+
+	fakeIGs.calls = []int{}
```
kl52752, line 134: can you explain why it is needed?

kl52752 (commented):
tiny comments, beside that lgtm

pkg/instances/instances_test.go
panslava: Done

pkg/instances/instances_test.go
panslava: Not really needed for this test, removed

pkg/instances/instances_test.go
```diff
@@ -41,6 +42,8 @@ func newNodePool(f *FakeInstanceGroups, zone string) NodePool {
 }
 
 func TestNodePoolSync(t *testing.T) {
```
cezarygerard, line 44: this should be a table test

cezarygerard (approved):
please see my 1 comment

pkg/instances/instances_test.go
panslava: Done

cezarygerard:
Slavik,

Thank you for refactoring the tests!

cezarygerard:

pkg/instances/instances.go
```diff
@@ -334,6 +335,13 @@ func (i *Instances) Sync(nodes []string) (err error) {
 		}
 		kubeNodes := sets.NewString(nodes...)
 
+		// Individual InstanceGroup has a limit for 1000 instances in it.
+		// As a result, it's not possible to add more to it.
+		if len(kubeNodes) > flags.F.MaxIgSize {
```
bowei, line 340: Should this truncation be stable... e.g. should we sort the list before truncation?
I don't know if there is an explicit guarantee that the result of kubeNodes.List() is always going to return the same set of Nodes.

pkg/instances/instances.go
```diff
@@ -334,6 +335,13 @@ func (i *Instances) Sync(nodes []string) (err error) {
...
+			klog.Warningf("Limiting number of VMs for InstanceGroup %s to %d", igName, flags.F.MaxIgSize)
```
bowei, line 341: I think we need to print more info here:

* Number of entries total 
* A sample of the VMs ignored (print at most the first 10)
* We should say "Nodes" instead of "VMs" (keep the naming consistent)
*

pkg/instances/instances.go
panslava: That is a good question, I was curious about this as well

.List() actually "returns the contents as a sorted string slice."
https://github.com/kubernetes/ingress-gce/blob/master/vendor/k8s.io/apimachinery/pkg/util/sets/string.go#L170

panslava: To be sure, I've added to unit tests assertion

// call sync one more time and check that it will be no-op and will not cause any api calls

pkg/instances/instances.go
panslava: Done

pkg/instances/instances.go
bowei: Let's add a comment here so future reviewers don't have the same question

// List() will return a sorted list so the kubeNodesList truncation will have a stable set of nodes.

bowei:
/approve

Ping back when you add the comment.

pkg/instances/instances.go
panslava: Done

panslava:
Okay, we discussed with Cezary and agreed on merging this, with hacky way of making the tests pass by doing 

`flags.F.MaxIgSize = 1000`

but will refactor this in other pull request

panslava:
Request following this one, with refactoring https://github.com/kubernetes/ingress-gce/pull/1715

pkg/instances/instances.go
```diff
@@ -334,6 +335,23 @@ func (i *Instances) Sync(nodes []string) (err error) {
...
 
+		// Individual InstanceGroup has a limit for 1000 instances in it.
+		// As a result, it's not possible to add more to it.
+		if len(kubeNodes) > flags.F.MaxIgSize {
+			// List() will return a sorted list so the kubeNodesList truncation will have a stable set of nodes.
+			kubeNodesList := kubeNodes.List()
+
+			// Store first 10 truncated nodes for logging
+			truncatedNodesSample := kubeNodesList[flags.F.MaxIgSize:]
+			maxTruncatedNodesSampleSize := 10
```
bowei, line 346: suggestion:

```
truncateForLog := func(l []string) []string {
  if len(l) <= 10 { return l }
  return l[:10]
}
```

klog.Warning(..., truncateForLog(kubeNodesList))

This way we don't pollute the variable state space and make it very clear what the intent is...

PR #1709, opened 2022-05-09
Fix for backend service update

kl52752:
We should update backend service with ig llinks
only when ig links changed.

### DIFF
--- a/pkg/backends/regional_ig_linker.go
+++ b/pkg/backends/regional_ig_linker.go
@@ -18,10 +18,10 @@ import (
 
 	cloudprovider "github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/instances"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/klog"
 )
 
 // RegionalInstanceGroupLinker handles linking backends to InstanceGroups.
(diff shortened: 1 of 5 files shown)


cezarygerard (approved):
lgtm

pkg/backends/regional_ig_linker.go
```diff
@@ -46,30 +45,26 @@ func (linker *RegionalInstanceGroupLinker) Link(sp utils.ServicePort, projectID
...
 	}
-
-	addIGs := sets.String{}
-	for _, igLink := range igLinks {
-		path, err := utils.RelativeResourceName(igLink)
-		if err != nil {
-			return fmt.Errorf("failed to parse instance group %s: %w", igLink, err)
-		}
-		addIGs.Insert(path)
+	be, err := linker.backendPool.Get(sp.BackendName(), meta.VersionGA, meta.Regional)
```
panslava, line 48: for me it is not very clear, what stands behind `be`, is it `BackEndservice` ?

panslava: (oh, it is probably coming from ig_linker.go)

panslava (approved):

cezarygerard:

bowei:
Let's fix the commit message so that the change is obvious when looking at the git history:

```
Only update BackendService IG links when they change

<more details...>
```

pkg/backends/regional_ig_linker.go
```diff
@@ -46,33 +45,29 @@ func (linker *RegionalInstanceGroupLinker) Link(sp utils.ServicePort, projectID
...
+		return err
 	}
 	if len(addIGs) == 0 {
 		return nil
 	}
-	// TODO(kl52752) Check for existing links and add only new one
+
 	var newBackends []*composite.Backend
-	for _, igLink := range addIGs.List() {
+	for _, igLink := range addIGs {
```
bowei, line 61: Is it an issue for unit tests that this order may not be stable due to the use of a map in the sets.String? Should the tests account for that?

pkg/backends/regional_ig_linker.go
```diff
@@ -46,33 +45,29 @@ func (linker *RegionalInstanceGroupLinker) Link(sp utils.ServicePort, projectID
...
+	bs, err := linker.backendPool.Get(sp.BackendName(), meta.VersionGA, meta.Regional)
+	if err != nil {
+		return err
+	}
+	addIGs, err := getInstanceGroupsToAdd(bs, igLinks)
+	if err != nil {
+		return err
 	}
 	if len(addIGs) == 0 {
 		return nil
```
bowei, line 57: Add a log here that there was no change to the Backends.

pkg/backends/regional_ig_linker.go
```diff
@@ -46,33 +45,29 @@ func (linker *RegionalInstanceGroupLinker) Link(sp utils.ServicePort, projectID
 		igSelfLink := cloudprovider.SelfLink(meta.VersionGA, projectID, "instanceGroups", key)
 		igLinks = append(igLinks, igSelfLink)
 	}
```
bowei, line 47: Let's log at the beginning of this method (Link()) 

log.Infof("Link(%v, %q, %v)", sp, projectID, zones)

pkg/backends/regional_ig_linker.go
```diff
@@ -46,33 +45,29 @@ func (linker *RegionalInstanceGroupLinker) Link(sp utils.ServicePort, projectID
...
-	be, err := linker.backendPool.Get(sp.BackendName(), meta.VersionGA, meta.Regional)
-	if err != nil {
-		return err
-	}
-	be.Backends = newBackends
 
-	if err := linker.backendPool.Update(be); err != nil {
+	bs.Backends = newBackends
+
+	if err := linker.backendPool.Update(bs); err != nil {
```
bowei, line 70: let's log Updating backendPool

bowei (commented):
Mostly LGTM. Let's add logging so if there is an issue with this code path, we have enough information to debug.

pkg/backends/regional_ig_linker.go
kl52752: done

pkg/backends/regional_ig_linker.go
kl52752: done

pkg/backends/regional_ig_linker.go
kl52752: done

pkg/backends/regional_ig_linker.go
kl52752: Added UT

pkg/backends/regional_ig_linker.go
```diff
@@ -39,40 +39,38 @@ func NewRegionalInstanceGroupLinker(instancePool instances.NodePool, backendPool
...
+	bs, err := linker.backendPool.Get(sp.BackendName(), meta.VersionGA, meta.Regional)
+	if err != nil {
+		return err
+	}
+	addIGs, err := getInstanceGroupsToAdd(bs, igLinks)
+	if err != nil {
+		return err
 	}
 	if len(addIGs) == 0 {
+		klog.V(3).Infof("No backends to add for %s, skipping update.", sp.BackendName())
```
bowei, line 59: Can we have this at V(2) so it appears in the normal log? Is it too noisy?

bowei: We can do this as a follow up

pkg/backends/regional_ig_linker.go
```diff
@@ -39,40 +39,38 @@ func NewRegionalInstanceGroupLinker(instancePool instances.NodePool, backendPool
...
 	}
-	be, err := linker.backendPool.Get(sp.BackendName(), meta.VersionGA, meta.Regional)
-	if err != nil {
-		return err
-	}
-	be.Backends = newBackends
 
-	if err := linker.backendPool.Update(be); err != nil {
+	bs.Backends = newBackends
+	klog.V(3).Infof("Update Backend %s, with %d backends.", sp.BackendName(), len(addIGs))
```
bowei, line 72: Same comment here V(2)

PR #1711, opened 2022-05-09
Fix HttpsRedirects Lifecycle

spencerhance:
* Fixes a bug preventing the disabling and deletion of HttpsRedirects
due to UrlMap Ensure happening before the TargetProxy Ensure
* EnsureRedirectUrlMap() now does not return an error if deletion fails
* Prevents premature removal of the `redirect-url-map` annotation (which prevented deletion)

Our e2e test does not catch this because it wipes out the annotation on ingress update().  This will be fixed in a follow-up.

### DIFF
--- a/pkg/loadbalancers/url_maps.go
+++ b/pkg/loadbalancers/url_maps.go
@@ -119,7 +119,10 @@ func (l *L7) ensureRedirectURLMap() error {
 			return nil
 		} else {
 			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); err != nil {
-				return err
+				// Do not block LB sync if this fails
+				klog.Errorf("DeleteUrlMap(%s) = %v", key, err)
+				// Signal to the rest of the controller that the UrlMap still exists
+				l.redirectUm = &composite.UrlMap{Name: key.Name}
 			}
 		}
 		return nil
--- a/pkg/loadbalancers/target_proxies.go
+++ b/pkg/loadbalancers/target_proxies.go
@@ -38,7 +38,6 @@ func (l *L7) checkProxy() (err error) {
 	// TODO(shance): move to translator
 	var umName string
 	if flags.F.EnableFrontendConfig {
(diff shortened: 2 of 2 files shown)


pkg/loadbalancers/url_maps.go
```diff
@@ -119,7 +119,10 @@ func (l *L7) ensureRedirectURLMap() error {
 			return nil
 		} else {
 			if err := composite.DeleteUrlMap(l.cloud, key, l.Versions().UrlMap); err != nil {
-				return err
+				// Do not block LB sync if this fails
+				klog.Errorf("DeleteUrlMap(%s) = %v", key, err)
+				// Signal to the rest of the controller that the UrlMap still exists
+				l.redirectUm = &composite.UrlMap{Name: key.Name}
```
swetharepakula, line 125: Does this mean that the deletion will fail first time around, and then after the target proxy is removed, the url map will be cleaned up on a retry?

swetharepakula, line 125: Does this mean that the deletion will fail first time around, and then after the target proxy is removed, the url map will be cleaned up on a retry?

swetharepakula:
Thanks Spencer!
/approve
/lgtm

swetharepakula:
Thanks Spencer!
/approve
/lgtm

pkg/loadbalancers/url_maps.go
swetharepakula: Resolved offline. The issue is an order of operations problem (url map deletion is attempted before the target proxy). Without a large refactor, the simplest way to fix is to not fail url map deletion and allow the rest of the deletion flow to go on and allow the url map to be deleted on a retry.

swetharepakula: Resolved offline. The issue is an order of operations problem (url map deletion is attempted before the target proxy). Without a large refactor, the simplest way to fix is to not fail url map deletion and allow the rest of the deletion flow to go on and allow the url map to be deleted on a retry.

PR #1715, opened 2022-05-18
Add maxIGSize to NodePool constructor, instead of using global flag

panslava:
This is refactoring of NodePool, following from this https://github.com/kubernetes/ingress-gce/pull/1707

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -107,9 +107,9 @@ var (
 		EnableIngressGAFields          bool
 		EnableTrafficScaling           bool
 		EnableEndpointSlices           bool
-		EnableMultipleIgs              bool
 		EnablePinhole                  bool
-		MaxIgSize                      int
+		EnableMultipleIGs              bool
+		MaxIGSize                      int
 	}{
 		GCERateLimitScale: 1.0,
 	}
@@ -249,8 +249,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableTrafficScaling, "enable-traffic-scaling", false, "Enable support for Service {max-rate-per-endpoint, capacity-scaler}")
 	flag.BoolVar(&F.EnableEndpointSlices, "enable-endpoint-slices", false, "Enable using Endpoint Slices API instead of Endpoints API")
 	flag.BoolVar(&F.EnablePinhole, "enable-pinhole", false, "Enable Pinhole firewall feature")
-	flag.BoolVar(&F.EnableMultipleIgs, "enable-multiple-igs", false, "Enable using unmanaged instance group management")
-	flag.IntVar(&F.MaxIgSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
+	flag.BoolVar(&F.EnableMultipleIGs, "enable-multiple-igs", false, "Enable using multiple unmanaged instance groups")
+	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
 }
 
--- a/pkg/instances/instances.go
+++ b/pkg/instances/instances.go
@@ -25,7 +25,6 @@ import (
 	"google.golang.org/api/compute/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/ingress-gce/pkg/events"
-	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog"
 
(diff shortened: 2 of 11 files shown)


pkg/flags/flags.go
```diff
@@ -244,8 +244,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
 	flag.BoolVar(&F.EnableTrafficScaling, "enable-traffic-scaling", false, "Enable support for Service {max-rate-per-endpoint, capacity-scaler}")
 	flag.BoolVar(&F.EnableEndpointSlices, "enable-endpoint-slices", false, "Enable using Endpoint Slices API instead of Endpoints API")
-	flag.BoolVar(&F.EnableMultipleIgs, "enable-multiple-igs", false, "Enable using unmanaged instance group management")
-	flag.IntVar(&F.MaxIgSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
+	flag.BoolVar(&F.EnableMultipleIGs, "enable-multiple-igs", false, "Enable using multiple unmanaged instance groups")
+	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
```
bowei, line 248: Not clear to me why we need to expose this as a flag. We can still keep it as var in the pool for testing purposes, but are we really going to change this constant?

pkg/instances/instances.go
```diff
@@ -47,22 +46,33 @@ type Instances struct {
...
+	Cloud      InstanceGroups
+	Namer      namer.BackendNamer
+	Recorders  recorderSource
+	BasePath   string
+	ZoneLister ZoneLister
+	MaxIGSize  int
+}
+
+// NewNodePool creates a new node pool using NodePoolConfig.
+func NewNodePool(config NodePoolConfig) NodePool {
```
bowei, line 68: take a pointer here unless you really want to copy things

pkg/instances/instances.go
```diff
@@ -337,19 +347,19 @@ func (i *Instances) Sync(nodes []string) (err error) {
...
 		// Individual InstanceGroup has a limit for 1000 instances in it.
 		// As a result, it's not possible to add more to it.
-		if len(kubeNodes) > flags.F.MaxIgSize {
+		if len(kubeNodes) > i.maxIGSize {
 			// List() will return a sorted list so the kubeNodesList truncation will have a stable set of nodes.
 			kubeNodesList := kubeNodes.List()
 
 			// Store first 10 truncated nodes for logging
-			truncatedNodesSample := kubeNodesList[flags.F.MaxIgSize:]
+			truncatedNodesSample := kubeNodesList[i.maxIGSize:]
```
bowei, line 355: suggestion:

truncateForLog := func(l []string) []string {
  if len(l) <= 10 { return l }
  return l[:10]
}

pkg/instances/instances.go
panslava: Fixed

pkg/instances/instances.go
panslava: Done

pkg/flags/flags.go
panslava: We will use this for easier end2end testing, so on some testing environment we would be able to set this flag to 5 (or other small value), and then test that our multiple instance group management actually working with google cloud apis, but without creating 1000+ nodes

pkg/instances/instances_test.go
```diff
@@ -84,9 +89,9 @@ func TestNodePoolSync(t *testing.T) {
 			defaultZone: {
 				ig: testCase.gceNodes,
 			},
-		})
+		}, maxIGSize)
```
cezarygerard, line 92: could you create the map[string]IGsToInstances outside of the function call?
just make it a variable 
```
readableVarName := map[string]IGsToInstances{
			defaultZone: {
				ig: testCase.gceNodes,
			},
		}
fakeGCEInstanceGroups := NewFakeInstanceGroups(readableVarName, maxIGSize)
```

pkg/instances/instances_test.go
```diff
@@ -144,12 +149,13 @@ func TestNodePoolSync(t *testing.T) {
 }
 
 func TestSetNamedPorts(t *testing.T) {
+	maxIGSize := 1000
 	fakeIGs := NewFakeInstanceGroups(map[string]IGsToInstances{
```
cezarygerard, line 153: same here,create asisgn the map to variable

pkg/instances/instances_test.go
```diff
@@ -198,11 +204,12 @@ func TestSetNamedPorts(t *testing.T) {
 }
 
 func TestGetInstanceReferences(t *testing.T) {
+	maxIGSize := 1000
 	pool := newNodePool(NewFakeInstanceGroups(map[string]IGsToInstances{
```
cezarygerard, line 208: same here,
add some variables for readability

pkg/instances/instances_test.go
panslava: Done

pkg/instances/instances_test.go
panslava: Done

pkg/instances/instances_test.go
panslava: Done

cezarygerard (approved):

PR #1717, opened 2022-05-19
Implement migration from L4 Legacy Target Pool based services to L4 RBS

panslava:
- Switch target of existing forwarding rule from target pool to regional
  backend service
- Clean up GCP resources used only by legacy service: target pool,
  health checks, firewall rules for health checks

### DIFF
--- a/pkg/loadbalancers/forwarding_rules.go
+++ b/pkg/loadbalancers/forwarding_rules.go
@@ -393,6 +393,23 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
 			networkTierMismatchError := utils.NewNetworkTierErr(resource, existingFwdRule.NetworkTier, fr.NetworkTier)
 			return nil, IPAddrUndefined, networkTierMismatchError
 		}
+
+		if existingFwdRule.Target != "" {
+			// If transitioning from Legacy L4 Net LB Service to RBS,
+			// instead of forwarding rule recreation, call set-target
+			// on forwarding rule, to minimize down-time
+			err = composite.SetProxyForForwardingRule(l4netlb.cloud, key, existingFwdRule, bsLink)
+			if err != nil {
+				return nil, IPAddrUndefined, err
+			}
+			err = l4netlb.cleanLegacyServiceResources()
+			if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/loadbalancers/forwarding_rules.go
```diff
@@ -393,6 +393,23 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
 			networkTierMismatchError := utils.NewNetworkTierErr(resource, existingFwdRule.NetworkTier, fr.NetworkTier)
 			return nil, IPAddrUndefined, networkTierMismatchError
 		}
+
+		if existingFwdRule.Target != "" {
+			// If transitioning from Legacy L4 Net LB Service to RBS,
+			// instead of forwarding rule recreation, call set-target
+			// on forwarding rule, to minimize down-time
+			err = composite.SetProxyForForwardingRule(l4netlb.cloud, key, existingFwdRule, bsLink)
```
kl52752, line 401: So from what I understand this is a place when you wiped out Target field from forwarding rule and set BackendService instead, right?
So my question is if any deletion from `cleanLegacyServiceResources` fails (not becausee of `in used by another resource` reason) then some resources might leaked (since forwarding rule point to BackendService and it is not legacy servicee any more). Is it right?

kl52752 (changes requested):

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -393,6 +393,23 @@ func (l4netlb *L4NetLB) ensureExternalForwardingRule(bsLink string) (*composite.
...
+			if err != nil {
+				return nil, IPAddrUndefined, err
+			}
+			err = l4netlb.cleanLegacyServiceResources()
```
kl52752, line 405: Is this error going to be exported in metric `L4NetLBsInError`?

pkg/loadbalancers/forwarding_rules.go
panslava: that is true, if some deletion fails with unexpected error, the resource will be leaked, cause neither ingress-gce nor k/k service-controller will clean it

PR #1722, opened 2022-06-01
Refactor L4NetLB

panslava:


pkg/loadbalancers/l4netlb.go
```diff
@@ -80,6 +82,7 @@ func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, n
 		NamespacedName: types.NamespacedName{Name: service.Name, Namespace: service.Namespace},
```
kl52752, line 82: why you need this? in new function?

pkg/loadbalancers/l4netlb.go
```diff
@@ -100,118 +112,115 @@ func (l4netlb *L4NetLB) createKey(name string) (*meta.Key, error) {
 // been created. It is health check, firewall rules, backend service and forwarding rule.
```
kl52752, line 112: I don't like this design here the error is hidden and it is easy to forget about error handling from function.

PR #1732, opened 2022-06-14
Multi cherry pick bugfixes from maste rto release 1.17

cezarygerard:
This fixes recent errors we found in scale testing


swetharepakula:
Can you add the original PR numbers so it will be easy to track?

PR #1748, opened 2022-07-13
Fix DestinationRanges update when IP changes

sugangli:
L4 firewall's DestinationRanges was not being updated when the forwardingrule's IP was updated.

### DIFF
--- a/pkg/firewalls/firewalls_l4.go
+++ b/pkg/firewalls/firewalls_l4.go
@@ -122,6 +122,10 @@ func firewallRuleEqual(a, b *compute.Firewall, skipDescription bool) bool {
 		}
 	}
 
+	if !utils.EqualStringSets(a.DestinationRanges, b.DestinationRanges) {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/firewalls/firewalls_l4.go
```diff
@@ -122,6 +122,10 @@ func firewallRuleEqual(a, b *compute.Firewall, skipDescription bool) bool {
 		}
 	}
 
+	if !utils.EqualStringSets(a.DestinationRanges, b.DestinationRanges) {
```
swetharepakula, line 125: Can you add a test for this too?

swetharepakula (commented):
Thanks Sugang for the bug fix!

Can you add a corresponding unit test for the change. Otherwise LGTM

pkg/firewalls/firewalls_l4.go
sugangli: Updated. PTAL.

swetharepakula (commented):
Thanks for the change!

/lgtm
/approve

PR #1749, opened 2022-07-19
Fix BackendConfig.securityPolicy is not removed after config updates

### LINKED ISSUE #1706: BackendConfig.securityPolicy is not removed after config updates
I have the following _BackendConfig_ with _securityPolicy_ that applies rate limite on a public API we expose via google load balancers. 

```
kind: BackendConfig
metadata:
  name: api
spec:
  healthCheck:
    checkIntervalSec: 15
    timeoutSec: 15
    healthyThreshold: 1
    unhealthyThreshold: 2
    type: HTTP
    requestPath: /api/
    port: 80
  securityPolicy:
    name: rate-limit
```

The policy itself works fine to create the BackendConfig and update the Load Balancers with the desired policies, but it doesn't update ir accordingly when we remove the policy.

During releases, our team need to run performance tests against the APIs behind this policy to ensure it has not decreased performance, but unfortunatelly due to a bug in the `ingress-gce` we can't disable the policy after it is applied, causing our tests to be blocked.

I've investigated the [source code](https://github.com/kubernetes/ingress-gce/blob/37a3586364079f4af0726a0fc444135e42a56bf7/pkg/backends/features/securitypolicy.go#L35) and identified that it skips the policySetup when the `securityPolicy` is not defined.

In order to solve this issue, we need a way to either set the policy to `None` like in the ui, or it should check if a policy in the resource matches the one defined in the BackendConfig to remove it if they don't match.

songrx1997:
Fix [issue](https://github.com/kubernetes/ingress-gce/issues/1706) #1706 BackendConfig.securityPolicy is not removed after config updates

### DIFF
--- a/pkg/backends/features/securitypolicy.go
+++ b/pkg/backends/features/securitypolicy.go
@@ -31,12 +31,11 @@ import (
 // EnsureSecurityPolicy ensures the security policy link on backend service.
 // TODO(mrhohn): Emit event when attach/detach security policy to backend service.
 func EnsureSecurityPolicy(cloud *gce.Cloud, sp utils.ServicePort, be *composite.BackendService) error {
-	if sp.BackendConfig.Spec.SecurityPolicy == nil {
-		return nil
-	}
-
-	if be.Scope != meta.Global {
-		return fmt.Errorf("cloud armor security policies not supported for %s backend service %s", be.Scope, be.Name)
+	var desiredPolicyName string
+	if sp.BackendConfig.Spec.SecurityPolicy != nil {
+		desiredPolicyName = sp.BackendConfig.Spec.SecurityPolicy.Name
+	} else {
+		desiredPolicyName = ""
 	}
 
 	existingPolicyName, err := utils.KeyName(be.SecurityPolicy)
@@ -44,9 +43,13 @@ func EnsureSecurityPolicy(cloud *gce.Cloud, sp utils.ServicePort, be *composite.
 	if be.SecurityPolicy != "" && err != nil {
 		return err
 	}
-	desiredPolicyName := sp.BackendConfig.Spec.SecurityPolicy.Name
+
 	if existingPolicyName == desiredPolicyName {
 		return nil
+	}
+
+	if be.Scope != meta.Global {
+		return fmt.Errorf("cloud armor security policies not supported for %s backend service %s", be.Scope, be.Name)
 	}
 
 	klog.V(2).Infof("Set security policy in backend service %s (%s:%s) to %q", be.Name, sp.ID.Service.String(), sp.ID.Port.String(), desiredPolicyName)
--- a/pkg/backends/features/securitypolicy_test.go
+++ b/pkg/backends/features/securitypolicy_test.go
@@ -78,7 +78,7 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 			expectSetCall: true,
 		},
 		{
-			desc: "remove-policy",
+			desc: "unset-policy-empty-policy-string",
 			currentBackendService: &composite.BackendService{
 				Scope:          meta.Global,
 				SecurityPolicy: "https://www.googleapis.com/compute/projects/test-project/global/securityPolicies/policy-1",
@@ -90,6 +90,15 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 					},
 				},
 			},
+			expectSetCall: true,
+		},
+		{
+			desc: "unset-policy-no-specified-policy",
+			currentBackendService: &composite.BackendService{
+				Scope:          meta.Global,
+				SecurityPolicy: "https://www.googleapis.com/compute/projects/test-project/global/securityPolicies/policy-1",
+			},
+			desiredConfig: &backendconfigv1.BackendConfig{},
 			expectSetCall: true,
 		},
 		{
@@ -112,15 +121,6 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 			desiredConfig:         &backendconfigv1.BackendConfig{},
 		},
 		{
-			desc: "no-specified-policy",
-			currentBackendService: &composite.BackendService{
-				Scope:          meta.Global,
-				SecurityPolicy: "https://www.googleapis.com/compute/projects/test-project/global/securityPolicies/policy-1",
-			},
-			desiredConfig: &backendconfigv1.BackendConfig{},
-			expectSetCall: false,
-		},
-		{
 			desc: "regional backend service",
 			currentBackendService: &composite.BackendService{
 				Scope: meta.Regional,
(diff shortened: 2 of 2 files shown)


swetharepakula:
Thanks for the change Ruixian! My main comment is that we want to still be able to exit early if we see that no change needs to be done for security policy. Can you also add a unit test for this change?

songrx1997:
Hi @swetharepakula thanks for you comment.

> Thanks for the change Ruixian! My main comment is that we want to still be able to exit early if we see that no change needs to be done for security policy. 
For this, we need to know the desired security policy and current security policy. If they are different, we need to change the policy. Otherwise we do nothing and exit. So I think only after [getting both the desired and current policy](https://github.com/kubernetes/ingress-gce/pull/1749/commits/a5bbb02dc540961ab69644ff95c93e2ae43db789#diff-694a3074761a6ab43364827c08bb7533a5532940ce0c8a11c418f4ed266d9ce2R51) can we know whether changes need to be done. 

> Can you also add a unit test for this change?
Will do this, thanks for reminding!

songrx1997:
Add a new [commit](https://github.com/kubernetes/ingress-gce/pull/1749/commits/143afcc0f94e36f0942e119c1aa511845365e59a) to update the test cases.

Added a new test case `remove-policy-2` for the situation:
Current Backend Service with securityPolicy &&  BackendConfig without securityPolicy ->  need to make a call to remove the policy. (This is the case in #1706 )

Removed `no-specified-policy`. We should make a call and remove the policy in this situation. So the case is no longer applicable.

pkg/backends/features/securitypolicy_test.go
```diff
@@ -78,7 +78,7 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 			expectSetCall: true,
 		},
 		{
-			desc: "remove-policy",
+			desc: "remove-policy-1",
```
swetharepakula, line 81: nit: maybe rename to `change-security-policy`

pkg/backends/features/securitypolicy_test.go
```diff
@@ -92,6 +92,15 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 			},
 			expectSetCall: true,
 		},
+		{
+			desc: "remove-policy-2",
```
swetharepakula, line 96: nit: rename to `remove-policy` or `unset-policy`

pkg/backends/features/securitypolicy_test.go
```diff
@@ -111,15 +120,6 @@ func TestEnsureSecurityPolicy(t *testing.T) {
 			currentBackendService: &composite.BackendService{Scope: meta.Global},
 			desiredConfig:         &backendconfigv1.BackendConfig{},
 		},
-		{
-			desc: "no-specified-policy",
-			currentBackendService: &composite.BackendService{
-				Scope:          meta.Global,
-				SecurityPolicy: "https://www.googleapis.com/compute/projects/test-project/global/securityPolicies/policy-1",
```
swetharepakula, line 118: modify this test so that the SecurityPolicy is not set in the currentBackendService

swetharepakula (commented):
Few nits left. This is almost ready to merge. Thanks!

pkg/backends/features/securitypolicy_test.go
songrx1997: Thanks @swetharepakula , this case would unset the security policy since the name field in the config is an empty string. I think `unset-policy-empty-policy-string` would be more clear.

pkg/backends/features/securitypolicy_test.go
songrx1997: Thanks @swetharepakula , this case would unset the security policy but it is different than the previous one. This one have no security policy field in the config. So I think `unset-policy-no-policy` would be more clear.

pkg/backends/features/securitypolicy_test.go
songrx1997: Thanks @swetharepakula , this is the `unset-policy-no-policy` test. Since it is similar to `unset-policy-empty-policy-string`, I put them together to make it clear.

pkg/backends/features/securitypolicy.go
```diff
@@ -31,20 +31,23 @@ import (
...
 func EnsureSecurityPolicy(cloud *gce.Cloud, sp utils.ServicePort, be *composite.BackendService) error {
-	if sp.BackendConfig.Spec.SecurityPolicy == nil {
-		return nil
-	}
-
 	if be.Scope != meta.Global {
 		return fmt.Errorf("cloud armor security policies not supported for %s backend service %s", be.Scope, be.Name)
 	}
 
+	var desiredPolicyName string
```
swetharepakula, line 38: @spencerhance, is there any problem because we are checking the scope before checking if security policy is nil? I am wondering if there are some cases where we will accidentally return an error but previously we would have moved on because security policy was nil?

pkg/backends/features/securitypolicy.go
```diff
@@ -31,24 +31,27 @@ import (
...
-		return nil
-	}
-
-	if be.Scope != meta.Global {
-		return fmt.Errorf("cloud armor security policies not supported for %s backend service %s", be.Scope, be.Name)
+	var desiredPolicyName string
+	if sp.BackendConfig.Spec.SecurityPolicy != nil {
+		desiredPolicyName = sp.BackendConfig.Spec.SecurityPolicy.Name
+	} else {
+		desiredPolicyName = ""
```
spencerhance, line 38: desiredPolicyName should be an empty string by default right?

pkg/backends/features/securitypolicy.go
spencerhance: Hm I think it should be okay because the policies would have to be different as well as the scope being regional.  Since you can't set the security policy on a regional BackendService I think we should be okay.

spencerhance (approved):
One possible scenario is that there is some customer who got stuck in this state accidentally and will be suprised when their policy is removed.  That being said, it's probably a very rare occurrence given how important securityPolicy is.

pkg/backends/features/securitypolicy.go
songrx1997: Yes it should be an empty string.

songrx1997:
> One possible scenario is that there is some customer who got stuck in this state accidentally and will be suprised when their policy is removed. That being said, it's probably a very rare occurrence given how important securityPolicy is.

Thanks @spencerhance Do you mean the situation that customers forget to put SecurityPolicy in the backendConfig and accidentally unset their security policy?

pkg/backends/features/securitypolicy.go
swetharepakula: My comment isn't applicable anymore. The previous version checked the scope before checking if there was a change to the security policy. I think this version of the code doesn't run into the issue I mentioned above. Thanks for reviewing Spencer!

pkg/backends/features/securitypolicy_test.go
swetharepakula: I think "regional backend service with no specified security policy" covers the case I was thinking of.

swetharepakula (commented):
Thanks for the change Ruixian!

/lgtm
/approve

PR #1754, opened 2022-07-20
Fix creation of NEGs in the new zone when cluster spans to the new zone

gauravkghildiyal:
* Propagate node update events for both NEG types (GCE_VM_IP_PORT + GCE_VM_IP)

* Configure separate zoneMaps for the NEG types (`vmIpZoneMap` and `vmIpPortZoneMap`). These zoneMaps will be updated using the same predicate function that is eventually used inside `transactionSyncer.isZoneChange()` for the respective NEG types.

* Introduce a new function `NodePredicateForNetworkEndpointType` that will be the source of truth for the type of predicate function to be used for a given NEG type.
  

For future reference, here's another related PR: https://github.com/kubernetes/ingress-gce/pull/1600

/assign @swetharepakula

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -292,7 +292,7 @@ func NewController(
 		})
 	}
 
-	nodeEventHandler := cache.ResourceEventHandlerFuncs{
+	nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			node := obj.(*apiv1.Node)
 			negController.enqueueNode(node)
@@ -301,20 +301,20 @@ func NewController(
 			node := obj.(*apiv1.Node)
 			negController.enqueueNode(node)
 		},
-	}
-
-	if negController.runL4 {
-		nodeEventHandler.UpdateFunc = func(old, cur interface{}) {
+		UpdateFunc: func(old, cur interface{}) {
 			oldNode := old.(*apiv1.Node)
 			currentNode := cur.(*apiv1.Node)
-			candidateNodeCheck := utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
-			if candidateNodeCheck(oldNode) != candidateNodeCheck(currentNode) {
-				logger.Info("Node has changed, enqueueing", "node", klog.KObj(currentNode))
+
+			vmIpCandidateNodeCheck := negtypes.NodePredicateForNetworkEndpointType(negtypes.VmIpEndpointType)
+			vmIpPortCandidateNodeCheck := negtypes.NodePredicateForNetworkEndpointType(negtypes.VmIpPortEndpointType)
+
+			if vmIpCandidateNodeCheck(oldNode) != vmIpCandidateNodeCheck(currentNode) ||
+				vmIpPortCandidateNodeCheck(oldNode) != vmIpPortCandidateNodeCheck(currentNode) {
+				logger.Info("Node has changed, enqueueing", "node", currentNode.Name)
 				negController.enqueueNode(currentNode)
 			}
-		}
-	}
-	nodeInformer.AddEventHandler(nodeEventHandler)
+		},
+	})
 
 	if enableAsm {
 		negController.enableASM = enableAsm
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -94,11 +94,12 @@ type syncerManager struct {
 	enableNonGcpMode     bool
 	enableEndpointSlices bool
 
-	// zoneMap keeps track of the last set of zones the neg controller
-	// has seen. zoneMap is protected by the mu mutex.
-	zoneMap map[string]struct{}
-
 	logger klog.Logger
+
+	// zone maps keep track of the last set of zones the neg controller has seen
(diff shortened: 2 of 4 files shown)


pkg/neg/manager.go
```diff
@@ -283,33 +270,62 @@ func (manager *syncerManager) SyncNodes() {
...
-		klog.Warningf("Unable to list zones: %s", err)
-		return false
-	}
+	vmIpSyncState, vmIpPortSyncState, nonGCPPrivateSyncState := zonesCalculationPending, zonesCalculationPending, zonesCalculationPending
 
-	newZoneMap := make(map[string]struct{})
-	for _, zone := range zones {
-		newZoneMap[zone] = struct{}{}
-	}
+	for key, syncer := range manager.syncerMap {
```
swetharepakula, line 294: Since this a map, the order is not guaranteed. What could happen is that the later syncers already see that the node is ready. So on the second loop, when the order changes, the first syncer which checks the zone may already hav the new zone but later syncers may not.

pkg/neg/manager.go
```diff
@@ -283,35 +270,14 @@ func (manager *syncerManager) SyncNodes() {
...
 		if needSync && !syncer.IsStopped() {
 			syncer.Sync()
 		}
 	}
 }
 
-// updateZoneMap updates the manager's zone map with the current zones and returns true if the
-// zones have changed. The caller must obtain mu mutex before calling this function
-func (manager *syncerManager) updateZoneMap() bool {
-	zones, err := manager.zoneGetter.ListZones(utils.AllNodesPredicate)
```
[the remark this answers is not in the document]
gauravkghildiyal: Resolving since outdated.

pkg/neg/controller.go
```diff
@@ -292,20 +292,21 @@ func NewController(
...
 			oldNode := old.(*apiv1.Node)
 			currentNode := cur.(*apiv1.Node)
-			candidateNodeCheck := utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
-			if candidateNodeCheck(oldNode) != candidateNodeCheck(currentNode) {
+
+			// 	- CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes() function decides the candidate nodes for the GCE_VM_IP kind of NEGs.
+			// 	- CandidateNodesPredicate() function decides the candidate nodes for the GCE_VM_IP_PORT kind of NEGs
+			//
+			// If we find a change in the candidacy of the node according to any of the two functions above, we should enqueue the node.
+			if utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes(oldNode) != utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes(currentNode) ||
```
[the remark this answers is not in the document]
gauravkghildiyal: Thanks. That makes sense. 

The reason why I initially thought it shouldn't be a problem was based on the need that we want to minimize the number of calls made to GCE APIs. Since `isZoneChange()` function was just making calls to the Kubernetes Cluster API, I thought it would be okay.

But it's true that there was scope of improvement there as we can save these O(number of nodes) operations used by `isZoneChange()` **for each NEG**, by just calculating it at the beginning and **not doing it for each NEG**

pkg/neg/manager.go
gauravkghildiyal: Thanks for pointing that out. I was consciously relying on the assumption that we were dealing with a slice where new NEGs are added at the end - where I believe things would have worked. Knowing that it's a map totally changes the game and doing things this was is definitely a big red flag.

pkg/neg/manager.go
```diff
@@ -743,52 +782,31 @@ func patchNegStatus(svcNegClient svcnegclient.Interface, oldNeg, newNeg negv1bet
...
-	for port, portInfo1 := range p1 {
-		portInfo2, ok := p2[port]
-		if !ok {
-			continue
-		}
+// updateZoneMap updates the existingZoneMap with the latest zones and returns
+// true if the zones have changed. The caller must obtain mu mutex of the
+// manager before calling this function since it modifies the passed
+// existingZoneMap.
+func updateZoneMap(existingZoneMap *map[string]struct{}, candidateNodePredicate utils.NodeConditionPredicate, zoneGetter negtypes.ZoneGetter, logger klog.Logger) bool {
```
swetharepakula, line 796: Can you move this function up to where it was before to reduce the git diff?

gauravkghildiyal: The reason for moving it down was because it's no longer a struct method for `syncerManager` but an individual one and I wanted to have all struct methods together and the rest towards the end. 

I don't have any strong opinion on either of the two so please do let me know how you'd like me to proceed. Thanks!

swetharepakula: In general I do agree with that. In this case, I am trying to preserve the git history so that we can track it back to the original commit easily.

Let's move this function back. Thanks!

pkg/neg/manager.go
swetharepakula: marking as resolved as no longer relevant.

pkg/neg/manager.go
gauravkghildiyal: Done :)

swetharepakula (commented):
Thanks for the change!

/approve
/lgtm

PR #1769, opened 2022-08-01
Add golang-ci with errcheck enabled to test suite to ensure all errors produced are checked

michaelasp:
Add the errcheck utility to check and see if there are any functions that produce an error which do not get checked after. Also fixes some issues related to this in sections, no functionality changes, we generally just log the error if it happens.


pkg/e2e/helpers.go
```diff
@@ -99,11 +99,6 @@ type WaitForIngressOptions struct {
 // This is needed for ConfigMap search.
 var Scheme = runtime.NewScheme()
 
-func init() {
```
swetharepakula, line 102: Don't we still need to add to scheme. The scheme created in line 100 doesn't have any types registered.

michaelasp: Oops you're right, forgot the init function autoruns on a per package basis so thought it could be removed since nothing called it 😅, reverted with an error check. In this case is this a panic worthy error if we can't register the scheme? I think so but let me know what you think.

build/test.sh
```diff
@@ -50,3 +50,16 @@ if [ -n "${ERRS}" ]; then
 fi
 echo "PASS"
 echo
+
+echo -n "Checking errcheck: "
+go install github.com/kisielk/errcheck@latest 2>/dev/null
```
panslava, line 55: I see that it just tries to install errcheck "globally" 

But I am curious, if this is the best approach, cause in main kubernetes repo, they using a separate go module, to keep track of such "utility" dependencies https://github.com/kubernetes/kubernetes/tree/master/hack/tools

michaelasp: Mmm this may be a better way, especially if we plan to install more dependencies in the future. Let me look into this. Thanks for the suggestion!

michaelasp: Done, very similar to how the kubernetes repo handles it. PTAL!

pkg/e2e/helpers.go
swetharepakula: Let's just log the error for now. That will be most similar to the existing behavior. Thanks!

hack/tools/tools.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
swetharepakula, line 2: nit: update to 2022

michaelasp:
Modified upon the idea from @panslava that we can use golang-ci instead. Won't compile until #1779 is resolved since golang-ci uses a higher go version. This should allow for more extensibility in the future and the ability to add more linters!

hack/tools/tools.go
michaelasp: Done

michaelasp:
Works as intended with newest go version bump, @panslava @swetharepakula PTAL!

pkg/e2e/helpers.go
michaelasp: Docne

build/test.sh
```diff
@@ -50,3 +50,27 @@ if [ -n "${ERRS}" ]; then
...
+# Set gobin here, install dependencies after this
+cd "hack/tools"
+export GOBIN=$PWD/bin
+export PATH=$GOBIN:$PATH
+export CGO_ENABLED=0
+# Install golangci-lint
+echo "Installing golangci-lint"
+echo
+go install github.com/golangci/golangci-lint/cmd/golangci-lint > /dev/null
+cd "../.."
```
panslava, line 63: it is going to work for now, but what about doing something like

`go mod download`

so we will just download all the packages that are in hack/tools ? (or it is not going to work?)

panslava:
Thank you, great job with CI!

LGTM from me, but I want someone else to also take a look

.golangci.yaml
```diff
@@ -0,0 +1,28 @@
...
+    exclude-functions:
+      - fmt.Fprintf
+      - fmt.Fprint
+      - (net/http.ResponseWriter).Write
+      - (*net/http.Server).Shutdown
+      - (*flag.FlagSet).Parse
+      - (*os.File).Close
+      - (io.Closer).Close
+      - (flag.Value).Set
+      - k8s.io/apimachinery/pkg/util/wait.PollUntil
```
panslava, line 28: Should we have a task to also fix error handling for these functions?

hack/tools/go.mod
```diff
@@ -0,0 +1,165 @@
+module k8s.io/ingress-gce/hack/tools
+
+go 1.19
```
cezarygerard, line 3: I think we went for go1.18 in PR/1784

cezarygerard:
Hi @michaelasp 

can you add 2-line README.md file to hack/tools
with info
1) what the tools do
2) how they are run

thanks

hack/tools/go.mod
michaelasp: Done, thanks!

michaelasp:
Added readme and modified the go version to match the actual version.

swetharepakula:
Thanks @michaelasp!

This is a great addition!

/lgtm
/approve

PR #1772, opened 2022-08-05
Add Custom Response Header Logic and the Unit Tests for it

songrx1997:


pkg/backends/features/customresponseheaders.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
swetharepakula, line 2: nit: change to 2022

swetharepakula:
Thanks for the change Ruixian!

A couple of small nits, otherwise lgtm! Please rebase on master and update the PR.

@spencerhance, please review as well. Thanks!

songrx1997:
> Thanks for the change Ruixian!
> 
> A couple of small nits, otherwise lgtm! Please rebase on master and update the PR.
> 
> @spencerhance, please review as well. Thanks!

Thanks @swetharepakula!  Just updated the PR for nits.

spencerhance (approved):
Thanks @songrx1997!

Generally lgtm.  We will have to be somewhat cautious in rolling this out since I think it will wipe out any existing response headers that somewhat has set manually.  (Unsupported behavior but still a risk factor we've seen before).

PR #1773, opened 2022-08-05
fix for get NodePort from service spec

kl52752:
service of type LoadBalancer has check that fields Spec.ports needs to be added but other type of services like ClusterIp does not have those requirements. To fix this issue we need to check is Spec.Ports array is not empty and it is good to check at the very beginning of `shouldProcess` that this service is of type LoadBalancer

### DIFF
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -80,6 +80,13 @@ func (r *L4NetLBSyncResult) SetMetricsForSuccessfulServiceSync() {
 	r.MetricsState.InSuccess = true
 }
 
+func getNodePort(service *corev1.Service) int64 {
+	if len(service.Spec.Ports) == 0 {
+		return 0
+	}
+	return int64(service.Spec.Ports[0].NodePort)
+}
+
 // NewL4NetLB creates a new Handler for the given L4NetLB service.
 func NewL4NetLB(service *corev1.Service, cloud *gce.Cloud, scope meta.KeyType, namer namer.L4ResourcesNamer, recorder record.EventRecorder) *L4NetLB {
 	l4netlb := &L4NetLB{cloud: cloud,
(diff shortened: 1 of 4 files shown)


pkg/loadbalancers/l4netlb.go
```diff
@@ -80,6 +80,13 @@ func (r *L4NetLBSyncResult) SetMetricsForSuccessfulServiceSync() {
 	r.MetricsState.InSuccess = true
 }
 
+func getNodePort(service *corev1.Service) int64 {
+	if len(service.Spec.Ports) == 0 {
+		return 0
+	}
+	return int64(service.Spec.Ports[0].NodePort)
+}
+
```
panslava, line 89: Let's add unit tests for this

pkg/utils/utils.go
```diff
@@ -782,3 +782,11 @@ func GetNetworkTier(service *api_v1.Service) (cloud.NetworkTier, bool) {
 		return cloud.NetworkTierDefault, false
 	}
 }
+
+// IsLoadBalancerService checks if kubernetes service is type of LoadBalancer.
```
panslava, line 786: Let's add unit tests for this

pkg/loadbalancers/l4netlb.go
kl52752: This is very simple function I don't think that there is any value in testing if len called on empty array will return 0.
In test `isRbsBasedService` we check this and the behaviour of test shows that it was properly initialized

pkg/utils/utils.go
kl52752: As above I think the value in test that checks business logic not language standard.

pkg/loadbalancers/l4netlb.go
```diff
@@ -80,6 +80,13 @@ func (r *L4NetLBSyncResult) SetMetricsForSuccessfulServiceSync() {
 	r.MetricsState.InSuccess = true
 }
 
+func getNodePort(service *corev1.Service) int64 {
```
panslava, line 83: Should we add some condition if service is nil?

PR #1775, opened 2022-08-08
[Cherrypick #1774 in release-1.19]  Fix RBS crash, when getting NodePorts for Services without ports

cezarygerard:
Fix RBS crash, when getting NodePorts for Services without ports


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -234,7 +234,7 @@ func (lc *L4NetLBController) hasForwardingRuleAnnotation(svc *v1.Service, frName
 
 // isRBSBasedService checks if service has either RBS annotation, finalizer or RBSForwardingRule
 func (lc *L4NetLBController) isRBSBasedService(svc *v1.Service) bool {
-	if svc == nil {
+	if !utils.IsLoadBalancerServiceType(svc) {
```
swetharepakula, line 237: Why is this change necessary? Doesn't the annotation and finalizer check afterwards cover all the cases? Is this why non RBS services were causing issues?

swetharepakula (commented):
/lgtm

PR #1776, opened 2022-08-09
Add Custom response header e2e test

songrx1997:
/assign @swetharepakula

### DIFF
--- a/cmd/e2e-test/customresponseheaders_test.go
+++ b/cmd/e2e-test/customresponseheaders_test.go
@@ -1,0 +1,139 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+
+	v1 "k8s.io/api/networking/v1"
+	"k8s.io/ingress-gce/pkg/annotations"
+	backendconfig "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
+	"k8s.io/ingress-gce/pkg/e2e"
+	"k8s.io/ingress-gce/pkg/e2e/adapter"
+	"k8s.io/ingress-gce/pkg/fuzz"
+	"k8s.io/ingress-gce/pkg/fuzz/features"
+	"k8s.io/ingress-gce/pkg/utils"
+)
+
+func TestCustomResponseHeaders(t *testing.T) {
+	t.Parallel()
+
+	ing := fuzz.NewIngressBuilder("", "ingress-1", "").
+		AddPath("test.com", "/", "service-1", v1.ServiceBackendPort{Number: 80}).
+		Build()
+
+	for _, tc := range []struct {
+		desc     string
+		beConfig *backendconfig.BackendConfig
+	}{
+		{
+			desc: "http one path w/ Custom Response header.",
+			beConfig: fuzz.NewBackendConfigBuilder("", "backendconfig-1").
+				AddCustomResponseHeader("X-Test-Header: test").
+				Build(),
+		},
+	} {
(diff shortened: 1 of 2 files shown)


swetharepakula (commented):
This generally looks good to me. Some small suggestions. Can you rebase and update the PR as well. Thanks!

pkg/fuzz/helpers.go
```diff
@@ -447,6 +447,15 @@ func (b *BackendConfigBuilder) AddCustomRequestHeader(header string) *BackendCon
 	return b
 }
 
+// AddCustomResponseHeader adds a custom response header to the BackendConfig.
+func (b *BackendConfigBuilder) AddCustomResponseHeader(header string) *BackendConfigBuilder {
+	if b.backendConfig.Spec.CustomResponseHeaders == nil {
+		b.backendConfig.Spec.CustomResponseHeaders = &backendconfig.CustomResponseHeadersConfig{}
+	}
+	b.backendConfig.Spec.CustomResponseHeaders.Headers = append(b.backendConfig.Spec.CustomResponseHeaders.Headers, header)
+	return b
```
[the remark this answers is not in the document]
songrx1997: Thanks Swetha! We are initializing the struct ```CustomResponseHeaders``` in the if statement, not the list ```CustomResponseHeaders.Headers```.  So we are appending that header to a nil list, which is a part of the struct.

cmd/e2e-test/customresponseheaders_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+			backendConfigAnnotation := map[string]string{
+				annotations.BetaBackendConfigKey: `{"default":"backendconfig-1"}`,
+			}
+			bcCRUD := adapter.BackendConfigCRUD{C: Framework.BackendConfigClient}
+			tc.beConfig.Namespace = s.Namespace
+			if _, err := bcCRUD.Create(tc.beConfig); err != nil {
+				t.Fatalf("error creating BackendConfig: %v", err)
+			}
+			t.Logf("BackendConfig created (%s/%s) ", s.Namespace, tc.beConfig.Name)
```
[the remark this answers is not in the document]
songrx1997: Just updated this with the desc in the log

pkg/fuzz/helpers.go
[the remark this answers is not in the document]
swetharepakula: You don't need the if condition because append is able to handle the nil list in the struct.

Append will deal with initializing the list.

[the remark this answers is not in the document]
swetharepakula: Oh, missed that. LGTM!

cmd/e2e-test/customresponseheaders_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+			backendConfigAnnotation := map[string]string{
+				annotations.BetaBackendConfigKey: `{"default":"backendconfig-1"}`,
+			}
+			bcCRUD := adapter.BackendConfigCRUD{C: Framework.BackendConfigClient}
+			tc.beConfig.Namespace = s.Namespace
+			if _, err := bcCRUD.Create(tc.beConfig); err != nil {
+				t.Fatalf("error creating BackendConfig: %v", err)
+			}
+			t.Logf("\"%s\"   BackendConfig created (%s/%s) ", tc.desc, s.Namespace, tc.beConfig.Name)
```
swetharepakula, line 66: maybe create a test logger that will do this automatically or create a helper method that wraps the logger?

songrx1997: Added a helper to prepend the desc to the log string.

cmd/e2e-test/customresponseheaders_test.go
```diff
@@ -0,0 +1,138 @@
...
+		t.Logf(logHelper(testDesc, "Backend service %q has expected custom response headers"), bs.GA.Name)
+		numBsWithCRH = numBsWithCRH + 1
+	}
+	if numBsWithCRH != 1 {
+		return fmt.Errorf("unexpected number of backend service has custom response headers: got %d, want 1", numBsWithCRH)
+	}
+	return nil
+}
+
+func logHelper(desc string, format string) string {
```
swetharepakula, line 136: Maybe rename to logWithDesc?

swetharepakula: It looks like with the default logging in golang, we are limited in our options. Lets rename to something like `appendDesc` so it is clear what the helper function is doing. Please also add a comment as to why we are doing this.

For the format maybe something like `[%s]: %s`?

songrx1997: Updated the naming and format.

swetharepakula (commented):
Thanks for the change!

/lgtm
/approve

PR #1793, opened 2022-08-26
Refactor NEG Syncers to use composition

panslava:
This is my personal opinion, how to do https://github.com/kubernetes/ingress-gce/pull/1792 better , without using "inheritance" like approach, but composition instead
But, I can be wrong


panslava:
Closing this, cause this was just experiment

PR #1807, opened 2022-09-12
Add Ingress Support for L7-ILB HTTP+HTTPS on same LB

sawsa307:
Previously this was not allowed by the L7-ILB, but now it is. Sharing the same VIP can be used only when specifying a static VIP.

### DIFF
--- a/cmd/e2e-test/ilb_test.go
+++ b/cmd/e2e-test/ilb_test.go
@@ -221,7 +221,6 @@ func TestILBStaticIP(t *testing.T) {
 	})
 }
 
-// TODO(shance): Remove the SetAllowHttp() calls once L7-ILB supports sharing VIPs
 func TestILBHttps(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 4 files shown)


spencerhance:
A couple quick notes:

Can you add more info about the bug fix in the PR Title and Description?

Also can you update the commit message to describe what it's doing?

Usually a good format for a commit message is formatted something like:

```
Fix <brief problem description> bug.

* More detail
* Even more detail
```

spencerhance:
Also the PR needs to be rebased:

```
git fetch upstream
git rebase upstream/master
```

pkg/loadbalancers/l7.go
```diff
@@ -147,8 +147,9 @@ func (l7 *L7) edgeHop() error {
 	}
 
 	// Check for invalid L7-ILB HTTPS config before attempting sync
-	if utils.IsGCEL7ILBIngress(l7.runtimeInfo.Ingress) && sslConfigured && l7.runtimeInfo.AllowHTTP {
-		l7.recorder.Eventf(l7.runtimeInfo.Ingress, corev1.EventTypeWarning, "WillNotConfigureFrontend", "gce-internal Ingress class does not currently support both HTTP and HTTPS served on the same IP (kubernetes.io/ingress.allow-http must be false when using HTTPS).")
+
+	if utils.IsGCEL7ILBIngress(l7.runtimeInfo.Ingress) && sslConfigured && l7.runtimeInfo.AllowHTTP && l7.runtimeInfo.StaticIPName == "" {
+		l7.recorder.Eventf(l7.runtimeInfo.Ingress, corev1.EventTypeWarning, "WillNotConfigureFrontend", "gce-internal Ingress class does not currently support both HTTP and HTTPS served on the same IP without the same non-static IP. Please configure a static-ip and attach it to the ingress with the kubernetes.io/ingress.regional-static-ip-name annotation.")
```
spencerhance, line 152: I think we can reword this to:

"gce-internal Ingress class must be configured with a static-ip to use both HTTP and HTTPS served on the same IP. Please configure a static-ip with Purpose=SHARED_LOADBALANCER_VIP and attach it to the ingress with the kubernetes.io/ingress.regional-static-ip-name annotation."

pkg/loadbalancers/l7.go
```diff
@@ -147,8 +147,9 @@ func (l7 *L7) edgeHop() error {
 	}
 
 	// Check for invalid L7-ILB HTTPS config before attempting sync
-	if utils.IsGCEL7ILBIngress(l7.runtimeInfo.Ingress) && sslConfigured && l7.runtimeInfo.AllowHTTP {
-		l7.recorder.Eventf(l7.runtimeInfo.Ingress, corev1.EventTypeWarning, "WillNotConfigureFrontend", "gce-internal Ingress class does not currently support both HTTP and HTTPS served on the same IP (kubernetes.io/ingress.allow-http must be false when using HTTPS).")
+
```
spencerhance, line 150: We should probably update the unit tests as well:

e.g. `TestCreateHTTPSILBLoadBalancerAllowHTTP` in `loadbalancers_test.go`

spencerhance (commented):
Some comments but otherwise LGTM!

swetharepakula:
Thanks for the change David! Can you update the commit message with a little more description of how the feature works. Perhaps a sentence that explains that the same VIP can be used only when specifying a static VIP?

pkg/loadbalancers/l7.go
sawsa307: Done. Thank you!

pkg/loadbalancers/l7.go
sawsa307: Updated test `TestCreateHTTPSILBLoadBalancerAllowHTTP`.

sawsa307:
> Thanks for the change David! Can you update the commit message with a little more description of how the feature works. Perhaps a sentence that explains that the same VIP can be used only when specifying a static VIP?

spencerhance (approved):

cmd/e2e-test/ilb_test.go
```diff
@@ -244,9 +244,8 @@ func TestILBHttps(t *testing.T) {
 			desc: "https ILB one path, pre-shared cert",
 			ingBuilder: fuzz.NewIngressBuilder("", ingressPrefix+"1", "").
 				AddPath("test.com", "/", serviceName, port80).
-				ConfigureForILB().
-				SetAllowHttp(false),
-			numForwardingRules: 1,
+				ConfigureForILB(),
```
swetharepakula, line 247: We should probably keep some of these tests because we want to test this configuration too. Perhaps add another case instead of changing the existing ones since the existing ones are still valid.

pkg/loadbalancers/loadbalancers_test.go
```diff
@@ -268,18 +268,30 @@ func TestCreateHTTPSILBLoadBalancer(t *testing.T) {
...
-		Ingress:   newILBIngress(),
+		AllowHTTP:    true,
+		TLS:          []*translator.TLSCerts{createCert("key", "cert", "name")},
+		UrlMap:       gceUrlMap,
+		Ingress:      newILBIngress(),
+		StaticIPName: ipName,
 	}
 
-	if _, err := j.pool.Ensure(lbInfo); err == nil {
-		t.Fatalf("j.pool.Ensure(%v) = nil, want err", lbInfo)
```
swetharepakula, line 282: Since this feature is will not be available for all versions, we still do want to keep this e2e test for the older versions, and write a new test. We also would want to keep this test for cases where static ip is not provided.

swetharepakula:
Code mostly LGTM. A few adjustments are needed for the tests so ensure we don't lose coverage on certain configurations..

swetharepakula:
/lgtm

@spencerhance please review one more time and I can approve once you think the PR is ready. Thanks!

sawsa307:
> If I understand correctly, this PR should allow Ingress with an internal L7 LB to have HTTP and HTTPS simultaneously. Still, from the [GKE official documentation for internal ingress](https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balance-ingress), I'm not sure how to achieve this. From the documentation it says:
> 
> image
> 
> This means that either HTTP or HTTPS can be active, not both. Is the documentation out of date, or I'm not reading it correctly?

Sorry the documentation is out-dated. We will update that.

PR #1782, opened 2022-08-19
Add Dual-Stack support to L4 ILB

panslava:
Add support for `spec.ipFamilies` and `spec.ipFamiliyPolicy` in L4 ILB

- Add new feature-flag `--enable-l4ilb-dual-stack`, all the new logic is hidden behind it
- Support all the transitions between ipFamilies, controller sync will provide resources for required ipFamily and clean resources for not required, but only if they exist in annotations
- On service deletion, clean up all the resources, no matter if they exist in annotations

PR #1808, opened 2022-09-13
Refactor healthchecksl4, change functions order, rename functions

panslava:
1) Moved functions to be defined in order of execution
2) Renamed EnsureHealthCheck -> EnsureHealthCheckWithFirewall, DeleteHealthCheck -> DeleteHealthCheckWithFirewall

No functional changes, just small refactoring

This is needed for Dual-Stack https://github.com/kubernetes/ingress-gce/pull/1782, but doing this in separate PR

### DIFF
--- a/pkg/healthchecksl4/healthchecksl4.go
+++ b/pkg/healthchecksl4/healthchecksl4.go
@@ -95,7 +95,7 @@ func GetInstance() *l4HealthChecks {
 	return instance
 }
 
-// EnsureHealthCheck and firewall rules exist for the L4
+// EnsureHealthCheckWithFirewall exist for the L4
 // LoadBalancer Service.
 //
 // The healthcheck and firewall will be shared between different K8s
@@ -105,7 +105,7 @@ func GetInstance() *l4HealthChecks {
 // Firewall rules are always created at in the Global scope (vs
 // Regional). This means that one Firewall rule is created for
 // Services of different scope (Global vs Regional).
-func (l4hc *l4HealthChecks) EnsureHealthCheck(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) *EnsureL4HealthCheckResult {
+func (l4hc *l4HealthChecks) EnsureHealthCheckWithFirewall(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) *EnsureL4HealthCheckResult {
 	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
 
 	hcName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
@@ -121,7 +121,7 @@ func (l4hc *l4HealthChecks) EnsureHealthCheck(svc *corev1.Service, namer namer.L
 	}
 	klog.V(3).Infof("L4 Healthcheck %s, path: %q, port %d", hcName, hcPath, hcPort)
 
-	_, hcLink, err := l4hc.ensureL4HealthCheckInternal(hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
+	hcLink, err := l4hc.ensureHealthCheck(hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
 	if err != nil {
 		return &EnsureL4HealthCheckResult{
 			GceResourceInError: annotations.HealthcheckResource,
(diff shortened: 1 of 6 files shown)


pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -226,6 +200,32 @@ func (l4hc *l4HealthChecks) ensureFirewall(svc *corev1.Service, hcFwName string,
 	return firewalls.EnsureL4LBFirewallForHc(svc, sharedHC, &hcFWRParams, l4hc.cloud, l4hc.recorderFactory.Recorder(svc.Namespace))
 }
 
+// DeleteHealthCheckWithFirewall deletes health check (and firewall rule) for l4 service. Checks if shared resources are safe to delete.
+func (l4hc *l4HealthChecks) DeleteHealthCheckWithFirewall(svc *corev1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error) {
+	hcName := namer.L4HealthCheck(svc.Namespace, svc.Name, sharedHC)
+	hcFwName := namer.L4HealthCheckFirewall(svc.Namespace, svc.Name, sharedHC)
+	namespacedName := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}
+	klog.V(3).Infof("Trying to delete L4 healthcheck: %s and firewall rule %s from service %s, shared: %v", hcName, hcFwName, namespacedName.String(), sharedHC)
```
kl52752, line 208: maybe it would be useful to log the scope also?

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -226,6 +200,32 @@ func (l4hc *l4HealthChecks) ensureFirewall(svc *corev1.Service, hcFwName string,
...
+	if sharedHC {
+		// We need to acquire a controller-wide mutex to ensure that in the case of a healthcheck shared between loadbalancers that the sync of the GCE resources is not performed in parallel.
+		l4hc.sharedResourcesLock.Lock()
+		defer l4hc.sharedResourcesLock.Unlock()
+	}
+
+	err := l4hc.hcProvider.Delete(hcName, scope)
+	if err != nil {
+		// Ignore deletion error due to health check in use by another resource.
+		if !utils.IsInUsedByError(err) {
```
kl52752, line 218: maybe we should check the value of sharedHC to see if this error is really related to shared health check?

pkg/healthchecksl4/interfaces.go
```diff
@@ -10,10 +10,10 @@ import (
 
 // L4HealthChecks defines methods for creating and deleting health checks (and their firewall rules) for l4 services
 type L4HealthChecks interface {
-	// EnsureHealthCheck creates health check (and firewall rule) for l4 service
-	EnsureHealthCheck(svc *v1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) *EnsureL4HealthCheckResult
-	// DeleteHealthCheck deletes health check (and firewall rule) for l4 service
-	DeleteHealthCheck(svc *v1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType) (string, error)
+	// EnsureHealthCheckWithFirewall creates health check with firewall rule for l4 service
```
kl52752, line 13: add the "." at the end

pkg/healthchecksl4/interfaces.go
```diff
@@ -10,10 +10,10 @@ import (
...
+	EnsureHealthCheckWithFirewall(svc *v1.Service, namer namer.L4ResourcesNamer, sharedHC bool, scope meta.KeyType, l4Type utils.L4LBType, nodeNames []string) *EnsureL4HealthCheckResult
+	// DeleteHealthCheckWithFirewall deletes health check with firewall rule for l4 service
```
kl52752, line 15: add the "." at the end

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -121,7 +121,7 @@ func (l4hc *l4HealthChecks) EnsureHealthCheck(svc *corev1.Service, namer namer.L
 	}
 	klog.V(3).Infof("L4 Healthcheck %s, path: %q, port %d", hcName, hcPath, hcPort)
 
-	_, hcLink, err := l4hc.ensureL4HealthCheckInternal(hcName, namespacedName, sharedHC, hcPath, hcPort, scope, l4Type)
```
kl52752, line 124: `ensureL4HealthCheckInternal` is this function used? I don't see it deleted?

pkg/healthchecksl4/healthchecksl4.go
panslava: I agree, that can be useful, but I don't want to change a lot of "logic" in this PR

Also, what if user decided to use our non-shared "healthcheck" for his own resources (which is weird case, but possible), so I am not sure, how we should act in this situation

pkg/healthchecksl4/healthchecksl4.go
panslava: Thank you, done

pkg/healthchecksl4/interfaces.go
panslava: Thank you, done

pkg/healthchecksl4/interfaces.go
panslava: Thank you done

PR #1810, opened 2022-09-13
Add checking for firewall source ranges in l4_test

panslava:
Instead of checking that len(firewall.SourceRanges) != 0 check that it actually has expected ranges

### DIFF
--- a/pkg/loadbalancers/l4_test.go
+++ b/pkg/loadbalancers/l4_test.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"google.golang.org/api/compute/v1"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/backends"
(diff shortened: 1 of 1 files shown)


pkg/loadbalancers/l4_test.go
```diff
@@ -1519,7 +1521,11 @@ func verifyNodesFirewall(l4 *L4, nodeNames []string) error {
 		return fmt.Errorf("failed to create description for resources, err %w", err)
 	}
 
-	return verifyFirewall(l4, nodeNames, fwName, fwDesc)
+	sourceRanges, err := servicehelper.GetLoadBalancerSourceRanges(l4.Service)
+	if err != nil {
+		return fmt.Errorf("servicehelper.GetLoadBalancerSourceRanges(%v) returned error %v, want nil", l4.Service, err)
```
kl52752, line 1526: you can add %+v for l4.Service to print fields also, but maybe only namespace name of the service is needed here?

pkg/loadbalancers/l4_test.go
```diff
@@ -1531,19 +1537,22 @@ func verifyHealthCheckFirewall(l4 *L4, nodeNames []string) error {
...
 		return fmt.Errorf("failed to fetch firewall rule %q - err %w", firewallName, err)
 	}
 	if !utils.EqualStringSets(nodeNames, firewall.TargetTags) {
 		return fmt.Errorf("expected firewall rule target tags '%v', Got '%v'", nodeNames, firewall.TargetTags)
 	}
-	if len(firewall.SourceRanges) == 0 {
-		return fmt.Errorf("unexpected empty source range for firewall rule %v", firewall)
+	sort.Strings(expectedSourceRanges)
+	sort.Strings(firewall.SourceRanges)
+	if !reflect.DeepEqual(expectedSourceRanges, firewall.SourceRanges) {
```
kl52752, line 1553: why do you need deepEqual? You can just check if returned diff is not an empty string.

panslava: thank you, changed just to diff

pkg/loadbalancers/l4_test.go
panslava: Thank you, added printing fields

I like being more "verbose" in test, especially on errors, so for me printing whole service is not that bad :)

PR #1811, opened 2022-09-13
Remove go test command with -i flag, cause it was deprecated

panslava:
This flag was deprecated, and does not seam to give any improvement https://github.com/golang/go/issues/41696


build/test.sh
```diff
@@ -55,7 +54,6 @@ echo
 cd "hack/tools"
 export GOBIN=$PWD/bin
 export PATH=$GOBIN:$PATH
-export CGO_ENABLED=0
```
panslava, line 58: This line is duplicated, we already do

`export CGO_ENABLED=0`

in the top of the file

PR #1812, opened 2022-09-13
Change healthchecksl4 sync mechanism

panslava:
Removed shared global struct, instead only use shared mutex 
for fake, return healthchecks instance with independent mutex

I also want to make our package pass go test -race, and current fake approach makes it fail

### DIFF
--- a/pkg/healthchecksl4/healthchecksl4.go
+++ b/pkg/healthchecksl4/healthchecksl4.go
@@ -24,10 +24,10 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/cloud-provider/service/helpers"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/composite"
-	"k8s.io/ingress-gce/pkg/events"
 	"k8s.io/ingress-gce/pkg/firewalls"
 	"k8s.io/ingress-gce/pkg/healthchecksprovider"
 	"k8s.io/ingress-gce/pkg/utils"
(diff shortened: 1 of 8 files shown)


pkg/l4lb/l4controller_test.go
```diff
@@ -70,7 +69,6 @@ func newServiceController(t *testing.T, fakeGCE *gce.Cloud) *L4Controller {
 	for _, n := range nodes {
 		ctx.NodeInformer.GetIndexer().Add(n)
 	}
-	healthchecksl4.Fake(ctx.Cloud, ctx)
```
kl52752, line 73: So right now all test for controllers are creating the health checks with shared mutex because it is created in NewHandler function. Is it desired?

panslava: Before this PR, NewHandler creates L4 with "synced" healthchecks, and then in all the tests, after NewHandler(), we override them with "Fake" health checks, which are supposed to be not synced. This approach is still questionable, but to fix it, it needs bigger refactoring

but still, if you look at the current Fake code

```
func Fake(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
	instance = &l4HealthChecks{
		cloud:           cloud,
		recorderFactory: recorderFactory,
		hcProvider:      healthchecksprovider.NewHealthChecks(cloud, meta.VersionGA),
	}
	return instance
}
``` 
mutex `sharedResourcesLock sync.Mutex` created automatically, cause it is member of struct by "value", so it is suppose to be "unsynced" and it's own for each of result of Fake()

But, now, if multiple tests running in parallel, it can be a race condition in Fake(), if 2 goroutines execute at first
```
instance = &l4HealthChecks{
		cloud:           cloud,
		recorderFactory: recorderFactory,
		hcProvider:      healthchecksprovider.NewHealthChecks(cloud, meta.VersionGA),
	}
```

then `return instance` will return the same instance for both of them

kl52752: but in controller test you don't use Fake for health check you just depend on the controller to create it's own l4 hander and it is using default [function](https://github.dev/panslava/ingress-gce/blob/change-healthchecks_l4-mutex/pkg/l4lb/l4controller.go#L217) and [here](https://github.dev/panslava/ingress-gce/blob/change-healthchecks_l4-mutex/pkg/loadbalancers/l4.go#L78).  
ANd in test code you create for every controller different [fakeGCE](https://github.com/panslava/ingress-gce/blob/4a6258b69c4296cde9b4e40ff1f70d9cec393ad0/pkg/l4lb/l4controller_test.go#L184) 
And for me there is no race condition unless test use the same fakeGCE (It probably is within one test not across the tests which run in parallel).
ANd for me it looks like all test are locked with the same mutex and use different fakeGCE.
Does it make sense to you?

panslava: You are right, just seems that l4controller_test and l4netlbcontroller_test don't use Parallel (cause they check "global" metrics") so it will not help if the mutex of health checks will be not shared

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -47,52 +47,36 @@ const (
...
 }
 
-// Fake creates instance of l4HealthChecks. Use for test only.
-func Fake(cloud *gce.Cloud, recorderFactory events.RecorderProducer) *l4HealthChecks {
-	instance = &l4HealthChecks{
-		cloud:           cloud,
-		recorderFactory: recorderFactory,
-		hcProvider:      healthchecksprovider.NewHealthChecks(cloud, meta.VersionGA),
+// FakeNotSynced creates instance of l4HealthChecks with independent lock. Use for test only.
+func FakeNotSynced(cloud *gce.Cloud, recorder record.EventRecorder) *l4HealthChecks {
```
cezarygerard, line 73: why not call this just Fake?

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -47,52 +47,36 @@ const (
 )
 
 var (
-	// instanceLock to prevent duplicate initialization.
-	instanceLock = &sync.Mutex{}
-	// instance is a singleton instance, created by Initialize
-	instance *l4HealthChecks
+	// sharedLock used to prevent race condition between shared health checks and firewalls.
+	sharedLock = &sync.Mutex{}
```
cezarygerard, line 51: So this variable should only exist once? I like this

cezarygerard:
see my 2 small comments
but overall LGTM
mu better than my singleton attempt ;-)

pkg/healthchecksl4/healthchecksl4.go
panslava: Okay, I renamed to Fake, difference is not that big, 
I just thought that it is so similar to just normal l4HealthChecks, that it is not that clear, what is "fake" here

PR #1814, opened 2022-09-14
Fix deletion of address in l4

panslava:
We were trying to delete address with backend service name, but we only reserve address with forwarding rule name

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -118,7 +118,7 @@ func (l4 *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncR
 		result.Error = err
 		result.GCEResourceInError = annotations.ForwardingRuleResource
 	}
-	if err = ensureAddressDeleted(l4.cloud, name, l4.cloud.Region()); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -118,7 +118,7 @@ func (l4 *L4) EnsureInternalLoadBalancerDeleted(svc *corev1.Service) *L4ILBSyncR
 		result.Error = err
 		result.GCEResourceInError = annotations.ForwardingRuleResource
 	}
-	if err = ensureAddressDeleted(l4.cloud, name, l4.cloud.Region()); err != nil {
+	if err = ensureAddressDeleted(l4.cloud, frName, l4.cloud.Region()); err != nil {
```
kl52752, line 121: can you fix also ensureFirewallRule [function](https://github.dev/panslava/ingress-gce/blob/l4-delete-proper-address/pkg/loadbalancers/forwarding_rules.go#L198). Argument name indicates that forwarding rule name == loadbalancer name but this is not true. I think this might be the reason of original bug.
YOu can put frName inside this [function](https://github.dev/panslava/ingress-gce/blob/l4-delete-proper-address/pkg/loadbalancers/l4.go#L250) and later use Name form returned forwaridng rule. 
WDYT?

panslava: thank you, good suggestion, done

PR #1821, opened 2022-09-18
Use namer.L4Firewall and namer.L4Backend instead of servicePort

panslava:
It is better to have single source of naming, which is L4Namer in l4netlb
This PR changes getting Backend Service name and Firewall name with using namer, instead of ServicePort
This PR also helps to add DualStack implementation 

/assign @kl52752

### DIFF
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -48,7 +48,6 @@ type L4NetLB struct {
 	// recorder is used to generate k8s Events.
 	recorder        record.EventRecorder
 	Service         *corev1.Service
-	ServicePort     utils.ServicePort
 	NamespacedName  types.NamespacedName
 	healthChecks    healthchecksl4.L4HealthChecks
 	forwardingRules ForwardingRulesProvider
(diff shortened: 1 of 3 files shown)


pkg/loadbalancers/l4netlb.go
```diff
@@ -125,26 +125,25 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
 	if hcResult.Err != nil {
 		result.GCEResourceInError = hcResult.GceResourceInError
 		result.Error = fmt.Errorf("Failed to ensure health check %s - %w", hcResult.HCName, hcResult.Err)
 		return result
 	}
 	result.Annotations[annotations.HealthcheckKey] = hcResult.HCName
+	result.Annotations[annotations.FirewallRuleForHealthcheckKey] = hcResult.HCFirewallRuleName
 
-	name := l4netlb.ServicePort.BackendName()
+	bsName := l4netlb.namer.L4Backend(l4netlb.Service.Namespace, l4netlb.Service.Name)
```
kl52752, line 136: Is `ServicePort` used after this refactoring? If not maybe you should delete it.

panslava: Thank you, good catch, done, 
but it needed to move a bit of logic upper to l4netlbcontroller, so now it's also changed

PR #1824, opened 2022-09-21
Add a new NetL4LB flag

code-elinka:
Before this change there was only one flag "num-l4-workers" to keep a count of internal and external L4 load balancer goroutines. It's better to have 2 separate flags so the new "num-net-l4-workers" flag was added.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -191,6 +191,7 @@ func main() {
 		Namespace:             flags.F.WatchNamespace,
 		ResyncPeriod:          flags.F.ResyncPeriod,
 		NumL4Workers:          flags.F.NumL4Workers,
+		NumL4NetLBWorkers:     flags.F.NumL4NetLBWorkers,
 		DefaultBackendSvcPort: defaultBackendServicePort,
 		HealthCheckPath:       flags.F.HealthCheckPath,
 		FrontendConfigEnabled: flags.F.EnableFrontendConfig,
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -84,6 +84,7 @@ var (
 		ResyncPeriod                     time.Duration
 		L4NetLBProvisionDeadline         time.Duration
 		NumL4Workers                     int
+		NumL4NetLBWorkers                int
 		NumIngressWorkers                int
 		RunIngressController             bool
 		RunL4Controller                  bool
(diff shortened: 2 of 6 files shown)


cmd/glbc/main.go
```diff
@@ -191,6 +191,7 @@ func main() {
 		Namespace:             flags.F.WatchNamespace,
 		ResyncPeriod:          flags.F.ResyncPeriod,
 		NumL4Workers:          flags.F.NumL4Workers,
+		NumNetL4Workers:       flags.F.NumNetL4Workers,
```
panslava, line 194: We usually use term "L4NetLB", so let's rename to `NumL4NetLBWorkers"

pkg/flags/flags.go
```diff
@@ -209,7 +210,9 @@ the pod secrets for creating a Kubernetes client.`)
 	flag.DurationVar(&F.L4NetLBProvisionDeadline, "l4-netlb-provision-deadline", 20*time.Minute,
 		`Deadline latency for L4 NetLB provisioning.`)
 	flag.IntVar(&F.NumL4Workers, "num-l4-workers", 5,
-		`Number of parallel L4 Service worker goroutines.`)
+		`Number of parallel internal (ILB) L4 Service worker goroutines.`)
```
panslava, line 213: `internal (ILB)` feels repetitive

let's say

`Number of parallel L4 Internal Load Balancer Service worker goroutines.`

pkg/flags/flags.go
```diff
@@ -209,7 +210,9 @@ the pod secrets for creating a Kubernetes client.`)
...
+	flag.IntVar(&F.NumNetL4Workers, "num-net-l4-workers", 5,
+		`Number of parallel external (XLB/NetLB) L4 Service worker goroutines.`)
```
panslava, line 215: Saying both XLB and NetLB is not needed (btw, in this code base, I rarely see term XLB, usually we use NetLB or ELB)

let's say

`Number of parallel L4 External Load Balancer Service worker goroutines.`

panslava (changes requested):
Thank you, just some small nits, otherwise lgtm

cmd/glbc/main.go
code-elinka: Done, thank you

pkg/flags/flags.go
code-elinka: Okay, I renamed also the flag itself to `num-l4-net-workers`. I did not add `-lb-` part to keep the flag shorter

code-elinka:
Tested it locally: 
```
 GLBC=bin/amd64/glbc hack/run-local-glbc.sh --run-l4-netlb-controller --num-l4-net-workers=3 --run-l4-controller
 ```
 Got:  NumL4Workers:5 (default if the flag is not entered),  NumL4NetLBWorkers:3 (from the flag)

panslava (approved):

PR #1826, opened 2022-09-24
Refactor instances package, move node controller out of l7

panslava:
- Rename instances package -> instancegroups. This package controls instance groups, not "instances" themselves, similar to "neg" package
- Move controller/node.go out of l7 controller, but to instancegroups package (cause it is used not only for l7, but also for l4)
- Unexport only internal methods, rename functions for clarity

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/ingress-gce/pkg/frontendconfig"
 	"k8s.io/ingress-gce/pkg/ingparams"
+	"k8s.io/ingress-gce/pkg/instancegroups"
 	"k8s.io/ingress-gce/pkg/l4lb"
 	"k8s.io/ingress-gce/pkg/psc"
 	"k8s.io/ingress-gce/pkg/serviceattachment"
@@ -357,8 +358,14 @@ func runControllers(ctx *ingctx.ControllerContext) {
 
 	ctx.Start(stopCh)
 
-	nodeController := controller.NewNodeController(ctx, stopCh)
-	go nodeController.Run()
+	igControllerParams := &instancegroups.ControllerConfig{
+		NodeInformer: ctx.NodeInformer,
+		IGManager:    ctx.InstancePool,
+		HasSynced:    ctx.HasSynced,
+		StopCh:       stopCh,
+	}
+	igController := instancegroups.NewController(igControllerParams)
+	go igController.Run()
 
 	// The L4NetLbController will be run when RbsMode flag is Set
 	if flags.F.RunL4NetLBController {
--- a/pkg/instancegroups/controller.go
+++ b/pkg/instancegroups/controller.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package controller
+package instancegroups
 
 import (
 	"time"
@@ -22,21 +22,19 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/ingress-gce/pkg/context"
-	"k8s.io/ingress-gce/pkg/instances"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
 )
 
-// NodeController synchronizes the state of the nodes to the unmanaged instance
+// Controller synchronizes the state of the nodes to the unmanaged instance
 // groups.
-type NodeController struct {
+type Controller struct {
 	// lister is a cache of the k8s Node resources.
 	lister cache.Indexer
 	// queue is the TaskQueue used to manage the node worker updates.
 	queue utils.TaskQueue
-	// instancePool is a NodePool to manage kubernetes nodes.
-	instancePool instances.NodePool
+	// igManager is an interface to manage instance groups.
+	igManager Manager
 	// hasSynced returns true if relevant caches have done their initial
 	// synchronization.
 	hasSynced func() bool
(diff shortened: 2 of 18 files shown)


cezarygerard:
1) What is the benefit of this change? What do we need it for?
2) can you try to merge regional IG linker, and IG linker. The regional IG linker is not necessary

swetharepakula (commented):
Can we break this down to make it easier to review? One commit would just be moving it to a new package and then move the refactors to a new commit. Currently a little difficult to understand what the implications might be.

cezarygerard:
@swetharepakula , this change is rather simple
Slavik has just moved node related stull from  pkg/controller to pkg/instancegroups
there should be no change in logic

@panslava can you confirm?

panslava:
I agree with you @cezarygerard, and most of the changes are just "renaming"

But, I can still split into commits
1) Moving node controller out
2) Renamings

panslava:
Uff, sorry, actually, I've tried again, and it does not make much sense to me to split it in multiple commits, I can only split in separate commit renaming package instances -> instancegroups, but all other renamings are kind of have to go together with moving Node Controller out of l7, otherwise, commits will not be self-sufficient 

So, if it is possible, can we go please with current PR ?

cezarygerard:
I am good with the code, LGTM from my side
but I would love to see LGTM from @swetharepakula or @spencerhance

cmd/glbc/main.go
```diff
@@ -358,8 +359,14 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
 	ctx.Start(stopCh)
 
-	nodeController := controller.NewNodeController(ctx, stopCh)
-	go nodeController.Run()
+	igControllerParams := &instancegroups.ControllerConfig{
+		NodeInformer: ctx.NodeInformer,
+		IGManager:    ctx.InstancePool,
+		HasSynced:    ctx.HasSynced,
+		StopCh:       stopCh,
+	}
```
swetharepakula, line 367: is there a reason to specify the objects rather than passing in the ctx?

pkg/instancegroups/interfaces.go
```diff
@@ -14,36 +14,33 @@ See the License for the specific language governing permissions and
...
+// Manager is an interface to sync kubernetes nodes to google cloud instance groups
+// through the Provider interface. It handles zones opaquely using the zoneLister.
+type Manager interface {
 	EnsureInstanceGroupsAndPorts(name string, ports []int64) ([]*compute.InstanceGroup, error)
 	DeleteInstanceGroup(name string) error
 
-	// TODO: Refactor for modularity
-	Add(groupName string, nodeNames []string) error
-	Remove(groupName string, nodeNames []string) error
-	Sync(nodeNames []string) error
```
swetharepakula, line 40: Why are we removing these?

pkg/instancegroups/interfaces.go
```diff
@@ -14,36 +14,33 @@ See the License for the specific language governing permissions and
...
+// ZoneLister manages lookups for GCE instance groups/instances to zones.
+type ZoneLister interface {
+	ListZones(predicate utils.NodeConditionPredicate) ([]string, error)
+	GetZoneForNode(name string) (string, error)
 }
 
-// InstanceGroups is an interface for managing gce instances groups, and the instances therein.
-type InstanceGroups interface {
+// Provider is an interface for managing gce instances groups, and the instances therein.
+type Provider interface {
```
swetharepakula, line 43: Why the change in the interface name? InstanceGroups seems clearer.

pkg/instancegroups/manager.go
```diff
@@ -268,20 +268,20 @@ func (i *Instances) splitNodesByZone(names []string) map[string][]string {
...
+func (m *manager) getInstanceReferences(zone string, nodeNames []string) (refs []*compute.InstanceReference) {
 	for _, nodeName := range nodeNames {
-		refs = append(refs, &compute.InstanceReference{Instance: fmt.Sprintf(i.instanceLinkFormat, zone, canonicalizeInstanceName(nodeName))})
+		refs = append(refs, &compute.InstanceReference{Instance: fmt.Sprintf(m.instanceLinkFormat, zone, canonicalizeInstanceName(nodeName))})
 	}
 	return refs
 }
 
 // Add adds the given instances to the appropriately zoned Instance Group.
-func (i *Instances) Add(groupName string, names []string) error {
```
swetharepakula, line 279: Are we no longer exporting because we were never using out of the package?

pkg/instancegroups/interfaces.go
swetharepakula: I see these functions we are no longer exporting since they are internal only.

swetharepakula (commented):
I agree this is a no-op. I had a couple questions on using ctx versus a new ControllerConfig (not a blocker).

Also the rename from InstanceGroups interface to Provider seems a little confusing to me.

It would also be good to get @spencerhance's opinion on the L7 implications. I don't think this has any behavior changes that will affect L7, but it would be good to confirm.

pkg/instancegroups/controller.go
```diff
@@ -84,15 +89,25 @@ func (c *NodeController) Run() {
...
+}
+
 // Shutdown shuts down the goroutine that processes node updates.
-func (c *NodeController) Shutdown() {
+func (c *Controller) Shutdown() {
 	c.queue.Shutdown()
 }
 
-func (c *NodeController) sync(key string) error {
+func (c *Controller) sync(_ string) error {
```
aojea, line 107: in my experience logging the key for debugging is worth it, 
```
start := time.Now()
klog.V(4).Infof("Start processing %s", key)
defer klog.V(4).Infof("Processing key %s took %v", time.Since(start)")
```

pkg/instancegroups/controller_test.go
```diff
@@ -0,0 +1,79 @@
...
+		{
+			"new heartbeat",
+			func(node *api_v1.Node) {
+				node.Status.Conditions[0].LastHeartbeatTime = meta_v1.NewTime(time.Now())
+			},
+			false,
+		},
+	}
+
+	for _, tc := range testCases {
```
aojea, line 46: use subtests better, it avoid test pollution between testcases, `t.Run`https://go.dev/blog/subtests

pkg/instancegroups/controller.go
panslava: Added

pkg/instancegroups/interfaces.go
panslava: TBH, it all starts from Inteliji linter, that says that interfaces and structs should not start with package name

And I find it reasonable actually

Inside this package, `Provider` makes sense, in a local context
Outside the package, it will be instancegroups.Provider which is also sounds good to me, better than instancegroups.InstanceGroupsProvider which is unreasonably repetitive 

And why word "Provider"? Because we are trying to use this word now for things that abstracts cloud API, we already have healthchecks provider, forwarding rules provider

pkg/instancegroups/controller.go
```diff
@@ -14,47 +14,52 @@ See the License for the specific language governing permissions and
...
 }
 
-// NewNodeController returns a new node update controller.
-func NewNodeController(ctx *context.ControllerContext, stopCh chan struct{}) *NodeController {
-	c := &NodeController{
-		lister:       ctx.NodeInformer.GetIndexer(),
-		instancePool: ctx.InstancePool,
-		hasSynced:    ctx.HasSynced,
-		stopCh:       stopCh,
+type ControllerConfig struct {
```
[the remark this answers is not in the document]
panslava: I am not strongly pushing on this, just to me feels like 4 arguments is already a lot for the function, and that's why I decided to move into single struct

But, I think, this is maybe not very golangish way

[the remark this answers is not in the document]
aojea: yeah, I think it is golangish so you can add new config options without modifying the constructor signature, my point is that you are only refactoring names and also not adding new inputs, so it does seem like something unnecessary here.

panslava:
/assign spencerhance
please take a look :)

spencerhance (approved):
seems reasonable to me

PR #1827, opened 2022-09-25
Add Multiple Instance Group Management

panslava:


cezarygerard:
before we merge this

can you try to combine ig_linker and regional_ig_linker into 1 struct/object?

PR #1828, opened 2022-09-26
Change l4 healthcheck timeout

code-elinka:
According to l4 healthcheck constants
the default set up has timeout = 8 seconds
and threshold = 3 times. As result it can
lead to 24 (8*3) seconds of lost requests.
This change is supposed to decrease this
time to only 6 seconds.

### DIFF
--- a/pkg/healthchecksl4/healthchecksl4.go
+++ b/pkg/healthchecksl4/healthchecksl4.go
@@ -38,12 +38,13 @@ import (
 
 const (
 	// L4 Load Balancer parameters
-	gceHcCheckIntervalSeconds = int64(8)
-	gceHcTimeoutSeconds       = int64(1)
+	gceSharedHcCheckIntervalSeconds = int64(8) // Shared Health check Interval
+	gceLocalHcCheckIntervalSeconds  = int64(3) // Local Health check Interval
+	gceHcTimeoutSeconds             = int64(1)
 	// Start sending requests as soon as one healthcheck succeeds.
-	gceHcHealthyThreshold = int64(1)
-	// Defaults to 3 * 8 = 24 seconds before the LB will steer traffic away.
-	gceHcUnhealthyThreshold = int64(3)
+	gceHcHealthyThreshold         = int64(1)
+	gceSharedHcUnhealthyThreshold = int64(3) // 3  * 8 = 24 seconds before the LB will steer traffic away
+	gceLocalHcUnhealthyThreshold  = int64(2) // 2  * 3 = 6 seconds before the LB will steer traffic away
 )
 
 var (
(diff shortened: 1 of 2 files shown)


panslava:
yeh, trash is hard to hold for 3 times consecutive

panslava:
We want to introduce this change only to services with external traffic policy local, so this needs a bit more changes

code-elinka:
Wrong change. `ExternalTrafficPolicy=Local` is not the same with XLB

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -302,12 +303,26 @@ func newL4HealthCheck(name string, svcName types.NamespacedName, shared bool, pa
 	if err != nil {
 		klog.Warningf("Failed to generate description for L4HealthCheck %s, err %v", name, err)
 	}
+
+	// Default time for LB to move the traffic differs for shared/local external traffic policies
```
cezarygerard, line 307: this can be enclosed in private methods, it will read better
hcCheckInterval :=  getHelthcheckInterval(shared)
hcUnhealthyThreshold := getHelthcheckUnhealthyThreshold(shared)

pkg/healthchecksl4/healthchecksl4_test.go
```diff
@@ -109,6 +121,32 @@ func TestCompareHealthChecks(t *testing.T) {
 	}
 }
 
+// Checks that correct health check constants used for different traffic policies types (local/shared)
```
cezarygerard, line 124: // Checks that correct health check constants used for share and non shared healthchecks.

Nowhere in healthchecksl4 we refer to traffic policy

pkg/healthchecksl4/healthchecksl4.go
code-elinka: Thank you, I decided not to use `get_` prefix and not to create a variable for that.

pkg/healthchecksl4/healthchecksl4_test.go
```diff
@@ -38,16 +38,107 @@ func TestMergeHealthChecks(t *testing.T) {
...
+			wantCheckIntervalSec:   gceLocalHcCheckIntervalSeconds,
+			wantTimeoutSec:         gceHcTimeoutSeconds,
+			wantHealthyThreshold:   gceHcHealthyThreshold,
+			wantUnhealthyThreshold: gceLocalHcUnhealthyThreshold + 1,
+		},
+	}
+	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
 			// healthcheck intervals and thresholds are common for Global and Regional healthchecks. Hence testing only Global case.
 			wantHC := newL4HealthCheck("hc", types.NamespacedName{Name: "svc", Namespace: "default"}, false, "/", 12345, utils.ILB, meta.Global, "")
```
aojea, line 144: this doesn't exercise shared/local that is the thing we just changed,no?

pkg/healthchecksl4/healthchecksl4_test.go
```diff
@@ -27,7 +27,7 @@ import (
 
 func TestMergeHealthChecks(t *testing.T) {
 	t.Parallel()
-	for _, tc := range []struct {
+	testCases := []struct {
```
aojea, line 30: if we add a `shared bool` var here we could this new dimension

PR #1833, opened 2022-10-13
increase test coverage

aojea:
This just exercise the new thresholds added depending on the shared value (local/cluster)

Lot of boilerplate instead of making the test dynamic for simplicity


panslava:
thank you!

/lgtm

PR #1835, opened 2022-10-13
Create Error State Interface

sawsa307:
Added error checking functions in transactionSyncer.go for syncer's error state get/set/reset.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -93,6 +93,14 @@ type transactionSyncer struct {
 	enableEndpointSlices bool
 
 	logger klog.Logger
+
+	// inError indicates if the syncer is in any of 4 error scenarios
+	// 1. Endpoint counts from EPS is different from calculated endpoint list
+	// 2. EndpontSlice has missing or invalid data
+	// 3. Attach/Detach EP fails due to incorrect batch information
+	// 4. Endpoint count from EPS or calculated endpoint list is 0
+	// Need to grab syncLock first for any reads or writes based on this value
+	inError bool
 }
 
 func NewTransactionSyncer(
@@ -136,6 +144,7 @@ func NewTransactionSyncer(
 		svcNegClient:         svcNegClient,
 		customName:           customName,
 		enableEndpointSlices: enableEndpointSlices,
+		inError:              false,
 		logger:               logger,
 	}
 	// Syncer implements life cycle logic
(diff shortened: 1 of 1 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -93,6 +93,9 @@ type transactionSyncer struct {
 	enableEndpointSlices bool
 
 	logger klog.Logger
+
+	// grab syncLock first if any decision needs to be based on this value
+	inError bool
```
swetharepakula, line 98: Can you expand the comment to indicate what it means to be in error state?

pkg/neg/syncers/transaction.go
```diff
@@ -184,6 +188,11 @@ func (s *transactionSyncer) syncInternal() error {
 }
 
 func (s *transactionSyncer) syncInternalImpl() error {
+	// TODO: enter degraded mode before reset and sync
```
swetharepakula, line 191: Can you expand the comment to indicate when do we enter the error state and why we will reset here? 

What was the finalized decision of when we run in degraded mode?

swetharepakula (commented):
Couple of nits on the comments. Please update the commit message since we no longer have the error state interface.

pkg/neg/syncers/transaction.go
sawsa307: Update the comments.

pkg/neg/syncers/transaction.go
sawsa307: Updated the comment. We will run in degraded mode at the beginning of next sync, and after degraded mode procedures are done, we reset the error boolean.

pkg/neg/syncers/transaction.go
```diff
@@ -93,6 +93,14 @@ type transactionSyncer struct {
 	enableEndpointSlices bool
 
 	logger klog.Logger
+
+	// inError indicates if the syncer is in any of 4 error scenarios
+	// 1. Endpoint counts from EPS is different from calculated endpoint list
+	// 2. EndpontSlice has missing or invalid data
+	// 3. Attach/Detach EP fails due to incorrect batch information
+	// 4. Endpoint count from EPS or calculated endpoint list is 0
+	// Need to grab syncLock first if any decision needs to be based on this value
```
swetharepakula, line 102: nit: lock is needed for writes as well.

swetharepakula (commented):
One nit, otherwise this LGTM. Thanks David!

pkg/neg/syncers/transaction.go
sawsa307: Fixed. Thanks!

swetharepakula (commented):
Thanks David!

/lgtm
/approve

PR #1836, opened 2022-10-13
Add check for endpoint counts equal

sawsa307:
Add check to make sure the total number of endpoints from EndpointDataequals to the number of endpoints from endpointPodMap. If not, the syncer will enter the error state. 
EndpointData may contain duplicated endpoint, but endpointPodMap removes the duplicated endpoints, and dupCount stores the number of duplicated it removed. We compare the endpoint counts with duplicates.
Change the function definition of NetworkEndpointsCalculator.CalculateEndpoints() to return another variable that represents the count of duplicated endpoints.

### DIFF
--- a/pkg/neg/syncers/transaction_test.go
+++ b/pkg/neg/syncers/transaction_test.go
@@ -1409,6 +1409,362 @@ func TestUnknownNodes(t *testing.T) {
 	}
 }
 
+func TestInvalidEndpointInfoEndpointCountsDiffer(t *testing.T) {
+	t.Parallel()
+	_, transactionSyncer := newTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), negtypes.VmIpPortEndpointType, false, true)
+
+	instance1 := testInstance1
+	instance2 := testInstance2
+	instance4 := testInstance4
+	testPortName := "port1"
+	testServiceNamespace := "namespace"
+	port80 := int32(80)
+	port81 := int32(81)
+	testIP1 := "10.100.1.1"
+	testIP2 := "10.100.1.2"
+	testIP3 := "10.100.2.2"
+	testIP4 := "10.100.4.1"
+	testPodName1 := "pod1"
+	testPodName2 := "pod2"
+	testPodName3 := "pod3"
+	testPodName4 := "pod4"
+
+	endpointPodMap := map[negtypes.NetworkEndpoint]types.NamespacedName{
+		{
+			IP:   testIP1,
+			Port: "80",
+			Node: instance1,
+		}: {
+			Namespace: testServiceNamespace,
+			Name:      testPodName1,
+		},
+		{
+			IP:   testIP2,
+			Port: "80",
+			Node: instance1,
+		}: {
+			Namespace: testServiceNamespace,
+			Name:      testPodName2,
+		},
+		{
+			IP:   testIP3,
+			Port: "81",
+			Node: instance2,
+		}: {
+			Namespace: testServiceNamespace,
+			Name:      testPodName3,
+		},
+		{
+			IP:   testIP4,
+			Port: "81",
+			Node: instance4,
+		}: {
+			Namespace: testServiceNamespace,
+			Name:      testPodName4,
+		},
+	}
+
+	testCases := []struct {
+		desc          string
+		endpointsData []negtypes.EndpointsData
+		dupCount      int
+		expect        bool
+	}{
+		{
+			desc: "counts equal, endpointData has no duplicated endpoints",
+			endpointsData: []negtypes.EndpointsData{
+				{
+					Meta: &metav1.ObjectMeta{
+						Name:      testServiceName + "-1",
+						Namespace: testServiceNamespace,
+					},
+					Ports: []negtypes.PortData{
+						{
+							Name: testPortName,
+							Port: port80,
+						},
+					},
+					Addresses: []negtypes.AddressData{
+						{
+							TargetRef: &corev1.ObjectReference{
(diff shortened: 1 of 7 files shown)


aojea:
> the syncer will enter the error state.

What does the syncer does in error state? Because it seems it just reset the value , so we just flip the state, or am I missing something?

sawsa307:
> > the syncer will enter the error state.
> 
> What does the syncer does in error state? Because it seems it just reset the value , so we just flip the state, or am I missing something?

When the syncer is in error state, we will do degraded mode procedures, which we will redo endpoint calculations and send attach/detach calls with correct endpoint information.

aojea:
> When the syncer is in error state, we will do degraded mode procedures,

ah, but that is in a follow up right? I can't see how it is done here 😄

sawsa307:
> > When the syncer is in error state, we will do degraded mode procedures,
> 
> ah, but that is in a follow up right? I can't see how it is done here 😄

Yes you're right. For now we reset the inError boolean so it is a no-op, but in the future when we implement the degraded mode functionality(which will be in a separate PR), it will be used to trigger degraded mode if the syncer is in error state.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,382 @@ func TestUnknownNodes(t *testing.T) {
 	}
 }
 
+func TestCompareEndpointCounts(t *testing.T) {
+	t.Parallel()
+	_, transactionSyncer := newTestTransactionSyncer(negtypes.NewAdapter(gce.NewFakeGCECloud(gce.DefaultTestClusterValues())), negtypes.VmIpPortEndpointType, false, true)
+
+	instance1 := testInstance1
+	instance2 := testInstance2
+	instance4 := testInstance4
```
swetharepakula, line 1418: Can we just use the original instances?

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,382 @@ func TestUnknownNodes(t *testing.T) {
...
+	}
+
+	testCases := []struct {
+		desc          string
+		endpointsData []negtypes.EndpointsData
+		dupCount      int
+		expect        bool
+	}{
+		{
+			desc: "counts equal, endpointData has not-ready endpoints",
```
swetharepakula, line 1466: based on the discussion above, we do not need to filter for not-ready endpoints.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,382 @@ func TestUnknownNodes(t *testing.T) {
...
+			IP:   "10.100.4.3",
+			Port: "80",
+			Node: instance4,
+		}: {
+			Namespace: testServiceNamespace,
+			Name:      "pod8",
+		},
+	}
+
+	testCases := []struct {
```
swetharepakula, line 1459: How about the cases for with no duplicated endpoints where counts are equal and counts are not?

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,382 @@ func TestUnknownNodes(t *testing.T) {
...
+			dupCount: 1,
+			expect:   true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			got := transactionSyncer.endpointCountsDiffer(tc.endpointsData, endpointPodMap, tc.dupCount)
+			if got != tc.expect {
+				t.Errorf("For case %q, expect equal to be %v, but got %v", tc.desc, tc.expect, got)
```
swetharepakula, line 1782: nit: Got before expectation/want.

Since you are running with subtests, you do not need to add the description in the error message. I would add something about what you are checking so it will be easier to discern from the test logs. Maybe something like:
```
t.Errorf("endpointCountsDiffer() = %t,  expected %t", tc.desc, tc.expect, got)
```

pkg/neg/syncers/utils.go
```diff
@@ -285,7 +286,13 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
 						networkEndpoint.Node = ""
 					}
 					zoneNetworkEndpointMap[zone].Insert(networkEndpoint)
-					networkEndpointPodMap[networkEndpoint] = types.NamespacedName{Namespace: endpointAddress.TargetRef.Namespace, Name: endpointAddress.TargetRef.Name}
+
+					// increment the count for duplicated endpoint
+					if _, contains := networkEndpointPodMap[networkEndpoint]; contains {
+						dupCount += 1
+					} else {
```
swetharepakula, line 293: This slightly changes the behavior of the function. In the old version, we would write blindly, so that means if we have duplicate endpoints the last one is the one we store. In this version we the one we store is the first one.

To maintain original behavior always set the endpoint in the networkEndpointPodMap.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,354 @@ func TestUnknownNodes(t *testing.T) {
...
+	instance2 := testInstance2
+	instance4 := testInstance4
+	testPortName := "port1"
+	testServiceNamespace := "namespace"
+	port80 := int32(80)
+	port81 := int32(81)
+
+	endpointPodMap := map[negtypes.NetworkEndpoint]types.NamespacedName{
+		{
+			IP:   "10.100.1.2",
```
swetharepakula, line 1426: make all the ips and names variables. This will reduce chance of testing errors, as the same variable is used and we don't have to maintain strings.

swetharepakula (commented):
couple nits. Otherwise lgtm!

/approve

swetharepakula:
/lgtm

Thanks!

PR #1837, opened 2022-10-14
Fix ports read/write races in neg tests

panslava:
In general, I want to make all unit tests pass with -race, this is 1 step towards 

Current approach with `ports` variable in neg tests has race, cause we use global variable, and we modify it from different tests

/assign swetharepakula
/assign cezarygerard
/cc code-elinka
/cc aojea

### DIFF
--- a/pkg/neg/controller_test.go
+++ b/pkg/neg/controller_test.go
@@ -1689,32 +1689,35 @@ func newTestIngress(name string) *networkingv1.Ingress {
 	}
 }
 
-var ports = []apiv1.ServicePort{
-	{
-		Port:       80,
-		TargetPort: intstr.FromInt(8080),
-	},
-	{
-		Port:       443,
-		TargetPort: intstr.FromString(testNamedPort),
-	},
-	{
-		Name:       testNamedPort,
-		Port:       8081,
-		TargetPort: intstr.FromInt(8081),
-	},
-	{
-		Port:       8888,
-		TargetPort: intstr.FromInt(8888),
-	},
-	{
-		Name:       testNamedPortWithNumber,
-		Port:       8881,
-		TargetPort: intstr.FromInt(8882),
-	},
+func servicePorts() []apiv1.ServicePort {
+	return []apiv1.ServicePort{
+		{
+			Port:       80,
+			TargetPort: intstr.FromInt(8080),
+		},
+		{
+			Port:       443,
+			TargetPort: intstr.FromString(testNamedPort),
+		},
+		{
+			Name:       testNamedPort,
+			Port:       8081,
+			TargetPort: intstr.FromInt(8081),
+		},
+		{
+			Port:       8888,
+			TargetPort: intstr.FromInt(8888),
+		},
+		{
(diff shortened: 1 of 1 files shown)


aojea:
why not remote `t.Parallel()` from the tests that share the variable?

panslava:
> why not removing `t.Parallel()` from the tests that share the variable?

Do you think it is better? I feel like running tests in parallel can be good to find some weird errors (like these global variables, but not in real code), that we would never found if we ran everything single threaded

aojea:
> Do you think it is better? I feel like running tests in parallel can be good to find some weird errors (like these global variables, but not in real code), that we would never found if we ran everything single threaded

in this case is a design choice , if you have a global variable to share in multiple test is a conscious decision, what is important though is to run with `-race` , we should always enable the race detector, t.Parallel() is more an optimisation to me ... but I can be wrong on this, of course

panslava:
> > Do you think it is better? I feel like running tests in parallel can be good to find some weird errors (like these global variables, but not in real code), that we would never found if we ran everything single threaded
> 
> in this case is a design choice , if you have a global variable to share in multiple test is a conscious decision, what is important though is to run with `-race` , we should always enable the race detector, t.Parallel() is more an optimisation to me ... but I can be wrong on this, of course

Tbh this tests were written not by me, but this global variable is actually not intended to be shared between tests (in the way that there is no shared logic based on this variable), as I understand, this is just some useful “const”

aojea:
And are the test modifying the variable?

panslava:
> And are the test modifying the variable?

They do modify, but only for their own usage, like if they understand default const is not enough, they will append to it something extra https://github.com/kubernetes/ingress-gce/blob/c6d26454e482a142e7805f7114b22e8ac73803f0/pkg/neg/controller_test.go#L1771

aojea:
Oh, that is wrong thenz, it will pollute all the other tests

panslava:
> Oh, that is wrong thenz, it will pollute all the other tests

So that’s what I am trying to fix in this PR, at first, I am taking a copy of this global ports, and then modify only copy

aojea:
What do you think about using a function instead?
It is more intuitive and less error prone to avoid similar issues in the future 
```
func servicePorts() []apiv1.ServicePort
```

panslava:
> What do you think about using a function instead? It is more intuitive and less error prone to avoid similar issues in the future
> 
> ```
> func servicePorts() []apiv1.ServicePort
> ```

thanks! good suggestion! changed

aojea:
/lgtm

Thanks

aojea:
why don't you send a PR enabling the `race detector` and adding a commit per fix?

this way is easier to review and to fix

panslava:
@aojea the problem is, there are other tests that are failing race detector, so I can not enable it yet... and other tests are harder to fix (at least not as obvious)

panslava:
or you suggesting something else? Maybe I didn't understand completely the suggestion

aojea:
https://github.com/kubernetes/ingress-gce/pull/1838

PR #1839, opened 2022-10-18
Reserve ILB IP address before deleting FR on protocol change

panslava:
This fixes a bug, which could result in losing ip address. On Service Protocol change, we delete forwarding rule before changing Backend Service. But we didn't reserve address before forwrading rule rule deletion, so after deleting forwarding rule, this address could be reserved from some other place.


aojea:
no tests?

panslava:
it's hard, maybe, I need to add a hook for fake gce, and check on forwarding rule deletion, that address of the forwarding rule is reserved

aojea:
> it's hard, maybe, I need to add a hook for fake gce, and check on forwarding rule deletion, that address of the forwarding rule is reserved

and an e2e instead?

cezarygerard:
there is a line of code in thatd does not follow the best practices:
pkg/loadbalancers/forwarding_rules.go:ensureForwardingRule

		equal, err := Equal(existingFwdRule, fr)
		if err != nil {
			return existingFwdRule, err
		}
		
		
please change it to 

		equal, err := Equal(existingFwdRule, fr)
		if err != nil {
			return nil, err
		}

PR #1842, opened 2022-10-20
Add check for invalid or missing data in EPS

sawsa307:
Check if EndpointSlice has missing or invalid data zone/nodeName. If so, the syncer will enter the error state.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -242,7 +242,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		}
 		endpointsData := negtypes.EndpointsDataFromEndpointSlices(endpointSlices)
 		targetMap, endpointPodMap, dupCount, err = s.endpointsCalculator.CalculateEndpoints(endpointsData, currentMap)
-		if s.invalidEndpointInfo(endpointsData, endpointPodMap, dupCount) {
+		if s.invalidEndpointInfo(endpointsData, endpointPodMap, dupCount) || s.isZoneMissing(targetMap) {
 			s.setErrorState()
 		}
 		if err != nil {
@@ -362,6 +362,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 //  1. The endpoint count from endpointData doesn't equal to the one from endpointPodMap:
 //     endpiontPodMap removes the duplicated endpoints, and dupCount stores the number of duplicated it removed
 //     and we compare the endpoint counts with duplicates
+//  2. There is at least one endpoint in endpointData with missing nodeName
 func (s *transactionSyncer) invalidEndpointInfo(eds []negtypes.EndpointsData, endpointPodMap negtypes.EndpointPodMap, dupCount int) bool {
 	// Endpoint count from EndpointPodMap
 	countFromPodMap := len(endpointPodMap) + dupCount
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -352,6 +355,29 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	return utilerrors.NewAggregate(errList)
 }
 
+// isNodeNameMissing returns true if there is endpoint missing nodeName
```
swetharepakula, line 358: Can we refactor this a little bit so that we only go through the eds one time? Perhaps we can combine with the validations in https://github.com/kubernetes/ingress-gce/pull/1836?

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,256 @@ func TestUnknownNodes(t *testing.T) {
...
+							Addresses: []string{testIP4},
+							Ready:     true,
+						},
+					},
+				},
+			},
+			expect: false,
+		},
+		{
+			desc: "contains missing nodeName",
```
swetharepakula, line 1508: nit: "at least one endpoint is missing a nodeName"

pkg/neg/syncers/transaction_test.go
```diff
@@ -1409,6 +1409,256 @@ func TestUnknownNodes(t *testing.T) {
...
+				},
+			},
+			expect: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			if got := transactionSyncer.isNodeNameMissing(tc.endpointsData); got != tc.expect {
+				t.Errorf("isNodeNameMissing() = %t, expected %t", got, tc.expect)
```
swetharepakula, line 1582: nit: include what we got first, and then the expected

swetharepakula: ignore my comment, I read the line incorrectly

pkg/neg/syncers/transaction.go
```diff
@@ -370,14 +371,31 @@ func (s *transactionSyncer) invalidEndpointInfo(eds []negtypes.EndpointsData, en
 	countFromEndpointData := 0
 	for _, ed := range eds {
 		countFromEndpointData += len(ed.Addresses)
+		for _, endpointAddress := range ed.Addresses {
+			nodeName := endpointAddress.NodeName
+			if nodeName == nil || len(*nodeName) == 0 {
+				s.logger.Info("Detected error when checking nodeName, marking syncer in error state", "endpoint", endpointAddress, "endpointData", eds)
```
swetharepakula, line 377: Missed this in the previous PR. We aren't actually setting error state here so we shouldn't add that to this log.

Instead maybe just write detected error when checking node name.

swetharepakula: Same for the log line below.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1765,6 +1765,291 @@ func TestInvalidEndpointInfoEndpointCountsDiffer(t *testing.T) {
 	}
 }
 
+func TestInvalidEndpointInfoNodeNameMissing(t *testing.T) {
```
swetharepakula, line 1768: You could probably combine this with the test above and just introduce other cases.

swetharepakula (commented):
Couple more small comments.

swetharepakula (commented):
Thanks David!

/lgtm
/approve

PR #1843, opened 2022-10-20
Add statusCodeHandler to 404-server-with-metrics which serves status code given in URL path

yiyangy:
Add to 404 server with metrics the functionality to return 500 status code (InternalServerError). The server returns 500 if request path is /servererror.


bowei:
Why not make it generic so we don't have to change this later:

```
/statuscode/400 => returns 400
...
/statuscode/500 => returns 500
/statuscode/501 => returns 501
```

We should limit these to 4?? and 5??, but this should cover any cases we need.

PR #1844, opened 2022-10-20
Add statusCodeHandler to 404-server-with-metrics which serves status code given in URL path

yiyangy:
Add to 404 server with metrics the functionality to return status code. The status code to return is given in the URL path of request in the format of /statuscode/<code>, e.g. /statuscode/500 .

### DIFF
--- a/cmd/404-server-with-metrics/server-with-metrics.go
+++ b/cmd/404-server-with-metrics/server-with-metrics.go
@@ -26,6 +26,8 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -47,6 +49,8 @@ var (
 	isProd            = flag.Bool("is_prod", true, "Indicates if the server is running in production.")
 )
 
+const statusCodePrefix string = "/statuscode/"
+
 func main() {
 	flag.Parse()
 	klog.InitFlags(nil)
@@ -61,7 +65,7 @@ func main() {
 	server.registerHandlers()
 	klog.Infof("Default 404 server is running with GOMAXPROCS(%d) on %s:%d\n", runtime.GOMAXPROCS(-1), hostName, *port)
 
-	// The main http server for handling NotFound and healthzrequests
+	// The main http server for handling NotFound, StatusCode and healthz requests
 	go func() {
 		err := server.httpServer.ListenAndServe()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -245,6 +250,25 @@ func (s *server) notFoundHandler() http.HandlerFunc {
...
 }
 
+// statusCodeHandler returns a status code given by URL path in the request.
+// If URL path does not contain a valid status code, returns 404 (NotFound).
+func (s *server) statusCodeHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		p := strings.TrimPrefix(r.URL.Path, statusCodePrefix)
+		code, err := strconv.Atoi(p)
+		if err != nil {
+			code = http.StatusNotFound
```
bowei, line 260: we should return an error message here instead of dropping through...

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -245,6 +250,25 @@ func (s *server) notFoundHandler() http.HandlerFunc {
 	}
 }
 
+// statusCodeHandler returns a status code given by URL path in the request.
+// If URL path does not contain a valid status code, returns 404 (NotFound).
+func (s *server) statusCodeHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		p := strings.TrimPrefix(r.URL.Path, statusCodePrefix)
+		code, err := strconv.Atoi(p)
```
bowei, line 258: We should only allow codes between [400, 599] to avoid potential for any abuse use case.

cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -245,6 +250,25 @@ func (s *server) notFoundHandler() http.HandlerFunc {
...
+		if err != nil {
+			code = http.StatusNotFound
+		}
+		statusText := http.StatusText(code)
+		if statusText == "" {
+			code = http.StatusNotFound
+			statusText = http.StatusText(code)
+		}
+		w.WriteHeader(code)
+		fmt.Fprintf(w, "response %d (%s) \n", code, statusText)
```
bowei, line 268: we probably don't need the text...

cmd/404-server-with-metrics/server-with-metrics.go
yiyangy: Done.

cmd/404-server-with-metrics/server-with-metrics.go
yiyangy: Done.

cmd/404-server-with-metrics/server-with-metrics.go
yiyangy: Done.

bowei (approved):

PR #1845, opened 2022-10-24
Add more logs to L4 NetLB

panslava:
This should not introduce any change in logic, but involves some refactoring

I've tried to split it by multiple commits

PR #1847, opened 2022-10-26
Remove emitting Update event on top-level in L4 NetLB

panslava:
I've added this event in https://github.com/kubernetes/ingress-gce/pull/1845 thinking it will give more info,

but I found out that proper update event is emitted in needsUpdate function, emitting event on top level will spam events even on periodic enqueues, which we don't want


cezarygerard:
/lgtm

great job!

PR #1848, opened 2022-10-27
 updated ingress api version

### LINKED ISSUE #1816: Replace `extensions/v1beta1` `Ingress` in documentation
There are many instances of `extensions/v1beta1` `Ingress` in this repository, e.g.:
https://github.com/kubernetes/ingress-gce/blob/051cfde145f34c8721e39df778e262fb4a256c67/LIMITATIONS.md#latency

Per https://endoflife.date/google-kubernetes-engine, it appears that 1.21 will expire at the end of the year. That will be the last version of GKE that supported `extensions/v1beta1` `Ingress` which was deprecated in 1.14 -- https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

kundan2707:
Updated apiversion for ingress from  extensions/v1beta1 to networking.k8s.io/v1beta1
Fixes #1816

### DIFF
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -33,7 +33,7 @@ It takes ~1m to spin up a loadbalancer (this includes acquiring the public IP), 
 
 Assume one creates the following simple Ingress:
 ```yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: test-ingress
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -53,7 +53,7 @@ Either shutdown the controller satisfying the Ingress, or use the
 `ingress.class` annotation:
 
 ```yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: test
--- a/docs/troubleshooting/troubleshooting.md
+++ b/docs/troubleshooting/troubleshooting.md
@@ -174,7 +174,7 @@ $ kubectl exec test-701078429-s5kca -- curl --cacert /var/run/secrets/kubernetes
     "/apis/certificates.k8s.io",
     "/apis/certificates.k8s.io/v1alpha1",
     "/apis/extensions",
-    "/apis/extensions/v1beta1",
+    "/apis/networking.k8s.io/v1",
     "/apis/policy",
     "/apis/policy/v1alpha1",
     "/apis/rbac.authorization.k8s.io",


kundan2707:
@cezarygerard  can you please spare your time for this review.

docs/faq/README.md
```diff
@@ -53,7 +53,7 @@ Either shutdown the controller satisfying the Ingress, or use the
 `ingress.class` annotation:
```
aojea, line 53: this annotation was deprecated in v1 in favor of a field https://docs.konghq.com/kubernetes-ingress-controller/latest/concepts/ingress-versions/#ingress-class

docs/troubleshooting/troubleshooting.md
```diff
@@ -174,7 +174,7 @@ $ kubectl exec test-701078429-s5kca -- curl --cacert /var/run/secrets/kubernetes
     "/apis/certificates.k8s.io",
     "/apis/certificates.k8s.io/v1alpha1",
     "/apis/extensions",
-    "/apis/extensions/v1beta1",
+    "/apis/networking.k8s.io/v1",
```
aojea, line 177: if you want to fix this output you have to update more apis that are no longer on that version

kundan2707:
@aojea  OK i will updated all required APIs and commit again

docs/faq/README.md
sawsa307: In this public doc https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#deprecated_annotation, it describes we should still use ingress.class.

LIMITATIONS.md
```diff
@@ -33,7 +33,7 @@ It takes ~1m to spin up a loadbalancer (this includes acquiring the public IP),
 
 Assume one creates the following simple Ingress:
 ```yaml
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: test-ingress
```
swetharepakula, line 39: We need to add the annotation for this to be valid yaml. In v1 either the annotation or the ingress class name field is required, but GKE doesn't use the class name.

docs/faq/README.md
swetharepakula: @kundan2707 we can also link to the GKE public docs here: https://cloud.google.com/kubernetes-engine/docs/how-to/custom-ingress-controller

swetharepakula (commented):
A few other suggestions to make the yamls correct. Thanks for the PR!

jsoref:
@kundan2707 did you want to update this PR? It looks like you probably only need to add a couple of ingressClass items...

jsoref:
@panslava who does one need to bribe to get simple things integrated?

PR #1857, opened 2022-11-11
Bump k8s-cloud-provider to v1.20.0

spencerhance:
Re-vendoring required running go mod tidy as well; I split everything into separate commits to make it more clear.


aojea:
/lgtm cancel

> I1111 08:41:43.268]     gen_test.go:1831: UrlMap should contain 19 fields. Got 18


is like some code needs to be regenerated?

spencerhance:
Thanks @aojea!  Good thing I added that test :)  Re-ran the codegen script and it looks good now.

PR #1858, opened 2022-11-12
wait on startup to get permissions

aojea:
The glbc component needs to get a CRD and have permission to read it, this is not inmidiate when bootstrapping a cluster so, instead of crashing and depending on a external component to restart the pod, we can active poll on this conditions before crashing.

This has the benefit of reducing the noise on the logs with panics, that may be misinterpreted by external tools processing the logs. It also improves the bootstrap latency.

We can observe this behavior in any of the jobs

```
F1112 06:42:03.275192       1 main.go:115] Failed to ensure BackendConfig CRD: failed to verify the existence of BackendConfig CRD: customresourcedefinitions.apiextensions.k8s.io "backendconfigs.cloud.google.com" is forbidden: User "system:controller:glbc" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0x1)
```

where the glbc controller panics multiple times because it does not has permissions
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-ingress-gce-e2e/1591319036715601920/artifacts/e2e-c4f27e13ef-58614-master/glbc.log

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -31,6 +31,10 @@ import (
 )
 
 const (
+	// Sleep interval to check the CRD is present.
+	checkCRDPresentInterval = time.Second
+	// Timeout to check the CRD is present.
+	checkCRDPresentTimeout = 60 * time.Second
 	// Sleep interval to check the Established condition of CRD.
 	checkCRDEstablishedInterval = time.Second
 	// Timeout for checking the Established condition of CRD.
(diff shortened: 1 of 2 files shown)


aojea:

just taking any of the jobs runs we can see these panics on the logs https://storage.googleapis.com/kubernetes-jenkins/logs/ci-ingress-gce-e2e/1591319036715601920/artifacts/e2e-c4f27e13ef-58614-master/glbc.log

The time of 1 minute is taken from the observation of that logs

cmd/glbc/main.go
```diff
@@ -111,7 +113,20 @@ func main() {
 	// TODO(rramkumar): Reuse this CRD handler for other CRD's coming.
 	crdHandler := crd.NewCRDHandler(crdClient)
 	backendConfigCRDMeta := backendconfig.CRDMeta()
-	if _, err := crdHandler.EnsureCRD(backendConfigCRDMeta, true); err != nil {
+	// The CRD and/or its permission may take a while, wait for 1 minute to avoid
+	// crashing so we can reduce the noise on the logs.
+	if err := wait.PollImmediate(3*time.Second, 1*time.Minute, func() (bool, error) {
```
spencerhance, line 118: I think we should either wrap all the bootstrapping logic in the retry loop or extend the comment to explain why we only have to retry on the first step.

spencerhance (commented):
Thanks for the PR!

cmd/glbc/main.go
aojea: I think that the comment is explaining that, or the CRD is not present or the glbc controller doesn't gave permissions

```
F1112 06:42:03.275192       1 main.go:115] Failed to ensure BackendConfig CRD: failed to verify the existence of BackendConfig CRD: customresourcedefinitions.apiextensions.k8s.io "backendconfigs.cloud.google.com" is forbidden: User "system:controller:glbc" cannot get resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0x1)
```

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-ingress-gce-e2e/1591319036715601920/artifacts/e2e-c4f27e13ef-58614-master/glbc.log

aojea: you can not wrap the entire bootstrapping logic because we are looking for a specific response in this specific request, however, this indicates that my comment is not great , let me update it

aojea: hmm, I think that this is not the right place to add this, my fault, I didn't check that EnsureCRD already retries, we should look for these errors there

aojea:
/hold cancel

now I think I got it right

spencerhance (approved):
lgtm👍

pkg/crd/crd.go
```diff
@@ -78,16 +82,33 @@ func (h *CRDHandler) EnsureCRD(meta *CRDMeta, namespacedScoped bool) (*apiextens
...
+		// CRD doesn't exist, create it
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		// Fail on any other error
+		if err != nil {
+			return false, fmt.Errorf("failed to verify the existence of %v CRD: %v", meta.kind, err)
+		}
+		// CRD exists, get current resource version and update it
+		updateCRD = true
```
swetharepakula, line 102: In a followup we should check if we even need to ensure. The case I am thinking is say we graduate an API, but had to rollback. We will fail here because in the rollback we try ensure a version without the graduated API.

swetharepakula:
Thanks Antonio!

/lgtm
/approve

aojea:
woot , no panics now

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-ingress-gce-e2e/1593131924900024320/artifacts/e2e-77e2f432cb-58614-master/glbc.log

PR #1859, opened 2022-11-16
Update e2e.EnsureIngress() to check Annotations in addition to Spec

spencerhance:
In addition to fixing the e2e func, the following fixes are included in the addl. commits:
* Fix TestILBStaticIP bug where the annotations were not being applied by using the newly fixed EnsureIngress() func instead.
* Fix a missing negAnnotation in the ILB e2e test.

### DIFF
--- a/cmd/e2e-test/ilb_test.go
+++ b/cmd/e2e-test/ilb_test.go
@@ -159,7 +159,7 @@ func TestILBStaticIP(t *testing.T) {
 			}
 		}
 
-		_, err := e2e.CreateEchoService(s, "service-1", nil)
+		_, err := e2e.CreateEchoService(s, "service-1", negAnnotation)
 		if err != nil {
 			t.Fatalf("e2e.CreateEchoService(s, service-1, nil) = _, %v; want _, nil", err)
 		}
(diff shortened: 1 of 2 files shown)


cmd/e2e-test/ilb_test.go
```diff
@@ -191,9 +191,7 @@ func TestILBStaticIP(t *testing.T) {
 		var gclb *fuzz.GCLB
 		for i, testIng := range []*v1.Ingress{testIngDisabled, testIngEnabled, testIngDisabled} {
 			t.Run(fmt.Sprintf("Transition-%d", i), func(t *testing.T) {
-				newIng := ing.DeepCopy()
-				newIng.Spec = testIng.Spec
-				ing, err = crud.Patch(ing, newIng)
+				ing, err = crud.Patch(ing, testIng)
```
aojea, line 194: do we need to change this too ?

https://github.com/kubernetes/ingress-gce/blob/8cabaedcf3f758aeb4ebb1df0c5f294d5c2eae23/pkg/e2e/fixtures.go#L279

aojea:
/ok-to-test

should not be better one commit per change?

cmd/e2e-test/ilb_test.go
spencerhance: Good point!  Just updated the PR.

PR #1863, opened 2022-11-23
Change e2e-test dockerfile

code-elinka:
To support multiarch, e2e-test dockerfile
should have non-arch dependant packages.

### DIFF
--- a/Dockerfile.e2e-test
+++ b/Dockerfile.e2e-test
@@ -12,12 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:9
+# Check End of Life Date wiki.debian.org/DebianReleases
+# and replace with a new one if needed
+FROM debian:11
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+ARG CLOUD_SDK_VERSION="416.0.0-0"
+ENV CLOUD_SDK_VERSION=$CLOUD_SDK_VERSION
 
 COPY cmd/e2e-test/run.sh /run.sh
-RUN apt-get update && apt-get install -y curl python
-RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-206.0.0-linux-x86_64.tar.gz | tar -zxf -
+
+# Download Google Cloud GPG and install google-cloud-cli
+# See recommended instructions:
+# cloud.google.com/sdk/docs/install#installation_instructions
+RUN apt-get update && \
+  apt-get install -y curl python apt-transport-https ca-certificates gnupg && \
+  echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] \
+  http://packages.cloud.google.com/apt cloud-sdk main" | \
+  tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
+  curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+  apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - && \
+  apt-get update -y && apt-get install google-cloud-cli=$CLOUD_SDK_VERSION -y
 
 ENTRYPOINT ["/run.sh"]


Dockerfile.e2e-test
```diff
@@ -18,6 +18,14 @@ ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
 
 COPY cmd/e2e-test/run.sh /run.sh
 RUN apt-get update && apt-get install -y curl python
-RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-206.0.0-linux-x86_64.tar.gz | tar -zxf -
+RUN echo "Installing gcloud SDK, kubectl ..." \
```
aojea, line 21: why is this required in the e2e test image?

Dockerfile.e2e-test
```diff
@@ -18,6 +18,14 @@ ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
 
 COPY cmd/e2e-test/run.sh /run.sh
 RUN apt-get update && apt-get install -y curl python
-RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-206.0.0-linux-x86_64.tar.gz | tar -zxf -
```
aojea, line 21: common pattern for multiarch image is to use the env variable to build the url, see this example https://github.com/kubernetes-sigs/kind/blob/7d26b50a129faf1a7024ce1337d516aada1db118/images/base/Dockerfile#L32

`curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-206.0.0-linux-$TARGET_ARCH.tar.gz`

don't know if it is needed, but if the new URL download binaries without the right architecture it will fail, or is the new approach installer handling this internally?

aojea: it seems the sdk shows how to do it here

https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/emulators/Dockerfile

code-elinka: I am also not sure if it is used at all :D 

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-network/ingress-gce-e2e.yaml

Dockerfile.e2e-test
code-elinka: Removed installation step, kept only downloading

Dockerfile.e2e-test
```diff
@@ -15,9 +15,16 @@
 FROM debian:9
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
```
aojea, line 17: why not using `ARG_ARCH` ?

is not useful?

code-elinka: `ARG_ARCH` can be `arm/arm64/amd64/..` but the curl link has only `x86_64` and `arm` options. Nothing else. I added comments to point to that. WDYT?

code-elinka:
@aojea @cezarygerard @panslava Do you have other comments?

panslava:
@code-elinka in general looks good to me, but being honest, I am not really in the context of the issue this PR solves, let's assign to Cezary and Antonio

/assign @cezarygerard 
/assign @aojea

Dockerfile.e2e-test
code-elinka: @aojea Changed to use official installation
https://cloud.google.com/sdk/docs/install#installation_instructions

code-elinka:
@aojea Please check. I hope the list of comments will help to brush off the dust from this PR 😸

Dockerfile.e2e-test
```diff
@@ -12,12 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:9
+# Check End of Life Date wiki.debian.org/DebianReleases
+# and replace with a new one if needed
+FROM debian:11
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+ARG CLOUD_SDK_VERSION="412.0.0-0"
```
aojea, line 20: should we use last one? https://cloud.google.com/sdk/docs/release-notes

code-elinka: Let's use the last one. I checked, we have it with debian:11. 
```
elinka@instance-2:~$ sudo apt list google-cloud-cli -a
Listing... Done
google-cloud-cli/cloud-sdk-bullseye,cloud-sdk 416.0.0-0 all
<...>
```

aojea:
/lgtm
/approve

let's iterate on this

aojea:
I don't have approval here XD

aojea:
ddi you rebase on the latest master?

code-elinka:
> ddi you rebase on the latest master?

@aojea Yes, I rebased on your no-alpine changes. 

https://github.com/code-elinka/ingress-gce/commits/multiarch

code-elinka:

Could you please take a look? I changed debian version and gcloud-cli link (so it doesn't depend on architecture).

PR #1864, opened 2022-11-23
Delete make cmd

code-elinka:
This should be merged after https://github.com/kubernetes/test-infra/pull/28095. Because only test-infra repo uses push-e2e make command.


Makefile
```diff
@@ -50,12 +50,7 @@ CONTAINER_PREFIX ?= ingress-gce
 VERSION ?= $(shell git describe --tags --always --dirty)
 
 # Set to 1 to print more verbose output from the build.
-VERBOSE ?= 0
+VERBOSE ?= 1
```
aojea, line 53: verbose by default ?

code-elinka: Lol, thank you! changed back

PR #1865, opened 2022-11-23
Add metrics for error state occurence

sawsa307:
Add metrics to report the number of occurrence of each error state within each sync.


sawsa307:
Now merged in https://github.com/kubernetes/ingress-gce/pull/1877

PR #1867, opened 2022-11-23
Adds Subnet field to e2e-test

spencerhance:
This is an optional field that is necessary for ILB tests running in clusters that are not using the "default" network and subnet.

Tested in an example VM.


swetharepakula:
/ok-to-test
/lgtm
/approve

Thanks Spencer!

PR #1872, opened 2022-11-29
Rework the health check for the loadbalancer controller

gauravkghildiyal:
## _Why do we need this?_
There have been several bug reports about customers concerned about API calls to `k8s-ingress-svc-acct-permission-check-probe` returning 404s -- this was also flagged out by partner teams since it results in lots of RESOURCE_NOT_FOUND errors within GCE APIs. 

https://github.com/kubernetes/ingress-gce/pull/1104 is an earlier related change (accompanied by the addition to the public documentation at https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#implementation_details). This change was an improvement over the earlier approach of probing backend service called `foo`.  

**We can improve this further by probing using an API call which does not error out.**

## _What does this PR propose?_

* We can use the `GCE Regions.Get` call (https://cloud.google.com/compute/docs/reference/rest/v1/regions/get) in place of fetching a random backend service. The benefit being:
    - The call won't error out
    - It uses regional APIs for improved compliance.
* We only call the GCE API after 60 seconds from the previous successful call. 
    - This avoids spikes in API calls to GCE, which can result because of issues with the k8s liveness probe or if the liveness probe is configured to be too frequent.
    - Note that we still do not independently ping GCE API "every" 60 seconds but only lazily ping when the health check is invoked. This is because having a totally independent pinging frequency would only have been helpful if we were utilizing the results of the ping outside the health check -- since they are only used within the health check, we should just ensure that they are "atleast" 60 seconds apart and avoid unnecessary pings.

## _Is this change safe?_
It's reasonable to think that the reason for using the `BackendService.GET` call could be to check if we have the associated IAM permission. But that does not seem to be the case because of the following reasons:

* We can see from the **existing code** here https://github.com/kubernetes/ingress-gce/blob/091ae8b07bd92d6781ce036ea67acb19d48e14a2/pkg/controller/controller.go#L306-L309 that we pass the health check even in case of a permissions error -- meaning even right now we don't care if we have the permission or not.
* Taking a look at how this code evolved, the original reason for making this API call was stated in the following comment https://github.com/kubernetes/ingress-gce/blob/73afef4bec27a6539da13f6cf93ae09d832c6dde/controllers/gce/controller/cluster_manager.go#L77-L79 but the comment got lost in the refactoring done in https://github.com/kubernetes/ingress-gce/pull/424. This comment further solidifies the reasoning that this API call was to check if gce client is able to make API calls.

### DIFF
--- a/cmd/glbc/app/clients.go
+++ b/cmd/glbc/app/clients.go
@@ -110,7 +110,11 @@ func NewGCEClient() *gce.Cloud {
 			// user has no need for Ingress in this case. If they grant
 			// permissions to the node they will have to restart the controller
 			// manually to re-create the client.
-			// TODO: why do we bail with success out if there is a permission error???
+			//
+			// Our aim here is not to validate if we have an IAM permission but just
+			// to ensure that gce client is working properly and able to make API
+			// calls. The choice of fetching BackendServices below is completely
+			// arbitrary.
 			if _, err = cloud.ListGlobalBackendServices(); err == nil || utils.IsHTTPErrorCode(err, http.StatusForbidden) {
 				return cloud
 			}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controller
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -37,7 +38,7 @@ import (
 	frontendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/frontendconfig/v1beta1"
 	"k8s.io/ingress-gce/pkg/backends"
 	"k8s.io/ingress-gce/pkg/common/operator"
-	"k8s.io/ingress-gce/pkg/context"
+	ingcontext "k8s.io/ingress-gce/pkg/context"
 	legacytranslator "k8s.io/ingress-gce/pkg/controller/translator"
 	"k8s.io/ingress-gce/pkg/events"
 	"k8s.io/ingress-gce/pkg/flags"
@@ -56,10 +57,15 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	LoadBalancerControllerName     = "loadbalancer-controller"
+	lastSuccessfulGCEPingThreshold = 1 * time.Minute
+)
+
 // LoadBalancerController watches the kubernetes api and adds/removes services
 // from the loadbalancer, via loadBalancerConfig.
 type LoadBalancerController struct {
-	ctx *context.ControllerContext
+	ctx *ingcontext.ControllerContext
 
 	nodeLister cache.Indexer
 
@@ -105,7 +111,7 @@ type LoadBalancerController struct {
 
 // NewLoadBalancerController creates a controller for gce loadbalancers.
 func NewLoadBalancerController(
-	ctx *context.ControllerContext,
+	ctx *ingcontext.ControllerContext,
 	stopCh chan struct{}) *LoadBalancerController {
 
 	broadcaster := record.NewBroadcaster()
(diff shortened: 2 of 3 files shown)


cmd/glbc/app/clients.go
```diff
@@ -110,7 +110,10 @@ func NewGCEClient() *gce.Cloud {
 			// user has no need for Ingress in this case. If they grant
 			// permissions to the node they will have to restart the controller
 			// manually to re-create the client.
-			// TODO: why do we bail with success out if there is a permission error???
+			//
+			// Our aim here is not to validate if we have an IAM permission but just
+			// to ensure that gce client is working properly and able to make API
+			// calls.
```
spencerhance, line 116: Can you add a comment here mentioning that the choice of BackendServices is completely arbitrary.

spencerhance (approved):
lgtm

pkg/controller/controller.go
```diff
@@ -298,16 +299,22 @@ func NewLoadBalancerController(
...
+		defer cancel()
+		// Our aim here is not to validate if we have an IAM permission but just
+		// to ensure that gce client is working properly and able to make API
+		// calls.
+		_, err := ctx.Cloud.Compute().Projects().Get(ctxWithTimeout, ctx.Cloud.ProjectID())
 
 		// If this container is scheduled on a node without compute/rw it is
 		// effectively useless, but it is healthy. Reporting it as unhealthy
 		// will lead to container crashlooping.
 		if utils.IsHTTPErrorCode(err, http.StatusForbidden) {
```
bowei, line 312: I think we should rethink this whole thing and redo it rather than trying to carry on the existing assumptions. This was originally written probably because the controller was being installed by ordinary users onto their cluster directly. Very few people today are doing this in production.

gauravkghildiyal: As discussed offline, I've made a few changes:
- Started using a regional API
- Prevent spikes in GCE pings because of some malfunction or misconfiguration.

pkg/controller/controller.go
```diff
@@ -875,3 +870,50 @@ func frontendGCAlgorithm(ingExists bool, scopeChange bool, ing *v1.Ingress) util
 		return utils.NoCleanUpNeeded
 	}
 }
+
+// lbcHealthCheck returns a health check function that can verify if regional
+// GCE APIs are reachable. An invocation of the returned function will only
+// reach out to GCE if there is a gap of `lastSuccessfulGCEPingThreshold` from
+// the previous GCE ping.
+func lbcHealthCheck(ingCtx *ingcontext.ControllerContext, lastSuccessfulGCEPingThreshold time.Duration) func() error {
```
bowei, line 878: This feels more complicated than it needs to be. Can we not do something like this:

```
func lbcHealthCheck(...) {
	status := struct {
		lock sync.Mutex
		err error
	}

	check := func() {
		ctx, cancel := context.WithTimeout(...)
		defer cancel()
		
		_, err := ingCtx.Cloud.Compute()...
		if utils.IsHTTPErrorCode(err, http.StatusForbidden) {
			err = nil
		}
		status.lock.Lock()
		defer status.lock.Unlock()
		status.err = err

		if err != nil {
			log.Errorf(...)
		}
	}

	checkForever := func() {
		for {
			check()
		}
	}

  var running bool

	return func() error {
    if ! running {  
      checkForever() 
      running = true
    }

		status.lock.Lock()
		defer status.lock.Unlock()

		return status.err
	}
}
```

gauravkghildiyal: Yes I agree thats another way of doing it but I deliberately did things this way because of the reason mentioned in the description -- I know the description is relatively bigger now so copying the relevant lines :) 
>Note that we still do not independently ping GCE API "every" 60 seconds but only lazily ping when the health check is invoked. This is because having a totally independent pinging frequency would only have been helpful if we were utilizing the results of the ping outside the health check -- since they are only used within the health check, we should just ensure that they are "atleast" 60 seconds apart and avoid unnecessary pings.

End behaviour being:
1. Don't make unnecessary API requests.
2. If the external health check has a frequency like 2 minutes (viz. >60 seconds), always make API calls at that configured frequency

gauravkghildiyal: Let me know if you would still prefer the change and I'll do that.

swetharepakula:
/remove-lifecycle rotten

@gauravkghildiyal do we still need this PR?

gauravkghildiyal:
@swetharepakula Ah yes I'll get to it eventually XD. Thanks for the reminder and removing the rotten lifecycle.

MaesterZ:
For anyone landing on this issue in 2025, I am using this dummy resource in Terraform to avoid this error spam in logs:
```
resource "google_compute_backend_service" "k8s_permission_check_probe" {
  project = <gcp_project_id>
  name    = "k8s-ingress-svc-acct-permission-check-probe"
}
```

PR #1879, opened 2022-12-06
Fix L4 NetLB RBS cleanup when Service Type changed

panslava:
This fixes issue, that on switching Service Type from LoadBalancer to ClusterIP or NodePort, RBS resources were not cleaned up

### DIFF
--- a/pkg/l4lb/l4netlbcontroller_test.go
+++ b/pkg/l4lb/l4netlbcontroller_test.go
@@ -516,6 +516,65 @@ func TestProcessServiceDeletion(t *testing.T) {
 	deleteNetLBService(lc, svc)
 }
 
+func TestProcessRBSServiceTypeTransition(t *testing.T) {
+	testCases := []struct {
+		desc      string
+		finalType v1.ServiceType
+	}{
+		{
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -516,6 +516,65 @@ func TestProcessServiceDeletion(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			svc, lc := createAndSyncNetLBSvc(t)
+			if lc.needsDeletion(svc) {
+				t.Errorf("Service should not be marked for deletion")
+			}
+
+			svc.Spec.Type = tc.finalType
```
aojea, line 541: is it safe to modify the original object? or should you create a deepcopy?

aojea:
lgtm but I'm not familiar enough with the controller logic so I defer to @cezarygerard for final approval

cezarygerard:
@panslava  I have extended this PR with little improved logic
please review the pull/1881

PR #1881, opened 2022-12-09
NetLB controller: cleanup RBS resources in case service type changed

cezarygerard:
NetLB RBS Controller: Change needsDeletion and shouldProcessService logic 

1) only process if old or new version of the service was/is RBS
2) only delete if service has RBS annotation/finalizer/forwarding rule, even if the type changed from LoadBalacer to other type

This will make controller cleanup RBS resources in case service type changed.

This PR builds on top of pull/1879 from @panslava


panslava:
Good job in saving up on calls!

/lgtm
/hold

unhold when you think we are ready to merge

PR #1886, opened 2022-12-14
Add delay for retrying NEG polling for unhealthy pods

alexkats:
Add delay for retrying NEG polling for unhealthy pods and exponential backoff for GCE API errors


alexkats:
Not going to merge now, as another change should fix the underlying issue as well. Closing for now, but it may be reopened if it's needed later.

PR #1890, opened 2022-12-22
NetLB: Treat LoadBalancer IP setup errors as user errors.

mmamczur:
Change the errors caused by user misconfiguring the Load Balancer IP in cases where:
- the requested LB external IP is not reserved in the project
- attempting to use an IP already used by another LB
- attempting to use a reserved IP of another type (internal IP).

This will affect numbers reported in the metrics, marking the above errors as user errors.

### DIFF
--- a/pkg/loadbalancers/forwarding_rules.go
+++ b/pkg/loadbalancers/forwarding_rules.go
@@ -18,6 +18,8 @@ package loadbalancers
 
 import (
 	"fmt"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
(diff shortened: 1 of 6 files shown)


pkg/loadbalancers/forwarding_rules_test.go
```diff
@@ -227,3 +238,47 @@ func TestForwardingRulesEqual(t *testing.T) {
...
+	fakeGCE.ReserveRegionAddress(addr, fakeGCE.Region())
+	insertError := &googleapi.Error{Code: http.StatusBadRequest, Message: "Invalid value for field 'resource.IPAddress': '1.1.1.1'. Specified IP address is in-use and would result in a conflict., invalid"}
+	fakeGCE.Compute().(*cloud.MockGCE).MockForwardingRules.InsertHook = test.InsertForwardingRuleErrorHook(insertError)
+	_, _, err := l4.ensureExternalForwardingRule("link")
+
+	require.Error(t, err)
+	assert.True(t, utils.IsIPConfigurationError(err))
+}
+
+type fakeForwardingRulesProvider struct {
```
aojea, line 272: is this used somewhere?

mmamczur: removed
it's a leftover, I didn't notice it when committing, thanks for pointing it out!

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -446,3 +451,9 @@ func l4lbIPToUse(svc *v1.Service, fwdRule *composite.ForwardingRule, requestedSu
 	}
 	return fwdRule.IPAddress
 }
+
+const addressAlreadyInUseMessage = "Specified IP address is in-use and would result in a conflict."
```
cezarygerard, line 455: - consts are typically at the very beginning of the file

WDYT?

cezarygerard (commented):
one small comment, p for discussion

LGTM overall

pkg/loadbalancers/forwarding_rules.go
mmamczur: I moved it to the const block in the beginning of the file

PR #1896, opened 2023-01-04
Reimplement getting load balancer source ranges

panslava:
- Introduce custom error type, to identify user errors
- Cover with unit tests (was not done in old helper)
- Needed to add IPv6 support later (in next PR)

/assign cezarygerard

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -482,7 +482,7 @@ func (l4 *L4) ensureIPv4NodesFirewall(nodeNames []string, ipAddress string, resu
 	}()
 
 	// ensure firewalls
-	sourceRanges, err := helpers.GetLoadBalancerSourceRanges(l4.Service)
+	sourceRanges, err := utils.ServiceSourceRanges(l4.Service)
 	if err != nil {
 		result.Error = err
 		return
@@ -490,7 +490,7 @@ func (l4 *L4) ensureIPv4NodesFirewall(nodeNames []string, ipAddress string, resu
 	// Add firewall rule for ILB traffic to nodes
 	nodesFWRParams := firewalls.FirewallParams{
 		PortRanges:        portRanges,
-		SourceRanges:      sourceRanges.StringSlice(),
+		SourceRanges:      sourceRanges,
 		DestinationRanges: []string{ipAddress},
 		Protocol:          string(protocol),
 		Name:              firewallName,
--- a/pkg/utils/sourceranges.go
+++ b/pkg/utils/sourceranges.go
@@ -1,0 +1,86 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/net"
+)
+
+const (
+	allowAllIPv4Range = "0.0.0.0/0"
+)
+
+func ServiceSourceRanges(service *v1.Service) ([]string, error) {
+	ipRanges, err := getAllSourceRanges(service)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(ipRanges) == 0 {
+		return []string{allowAllIPv4Range}, nil
+	}
+	return ipRanges.StringSlice(), nil
+}
+
+func getAllSourceRanges(service *v1.Service) (net.IPNetSet, error) {
+	// if SourceRange field is specified, ignore sourceRange annotation
+	if len(service.Spec.LoadBalancerSourceRanges) > 0 {
+		specs := service.Spec.LoadBalancerSourceRanges
+		ipnets, err := net.ParseIPNets(specs...)
+
+		if err != nil {
+			return nil, NewInvalidSpecLoadBalancerSourceRangesError(specs, err)
+		}
+		return ipnets, err
+	}
+
+	val, ok := service.Annotations[v1.AnnotationLoadBalancerSourceRangesKey]
+	if !ok {
+		return nil, nil
+	}
+	val = strings.TrimSpace(val)
+	specs := strings.Split(val, ",")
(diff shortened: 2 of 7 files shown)


pkg/utils/sourceranges.go
```diff
@@ -0,0 +1,86 @@
...
+			return nil, NewInvalidSpecLoadBalancerSourceRangesError(specs, err)
+		}
+		return ipnets, err
+	}
+
+	val, ok := service.Annotations[v1.AnnotationLoadBalancerSourceRangesKey]
+	if !ok {
+		return nil, nil
+	}
+	val = strings.TrimSpace(val)
```
aojea, line 43: this is a common error on users, do you want to allow the values with extra spaces as valid?

aojea: because you are not doing that for the spec and I think you should be consistent, Kubernetes will allow it and will fail later (unless people uses kube-proxy that Rob fixed https://github.com/kubernetes/kubernetes/pull/94107/files)

aojea: oh, this is just a copy of the old code, why do you want to have the same code duplicate? should you not delete the old one?

pkg/utils/sourceranges_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"net"
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestServiceSourceRanges(t *testing.T) {
```
aojea, line 13: missing tests with traling spaces

pkg/utils/sourceranges.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	return ipRanges.StringSlice(), nil
+}
+
+func getAllSourceRanges(service *v1.Service) (net.IPNetSet, error) {
+	// if SourceRange field is specified, ignore sourceRange annotation
+	if len(service.Spec.LoadBalancerSourceRanges) > 0 {
+		specs := service.Spec.LoadBalancerSourceRanges
+		ipnets, err := net.ParseIPNets(specs...)
+
```
aojea, line 32: nit , we can remove this blank line

pkg/utils/sourceranges.go
panslava: Added trimming spaces per each range value, both in annotation and spec

pkg/utils/sourceranges_test.go
panslava: Added

pkg/utils/sourceranges.go
panslava: Done

pkg/loadbalancers/l4.go
```diff
@@ -482,15 +482,15 @@ func (l4 *L4) ensureIPv4NodesFirewall(nodeNames []string, ipAddress string, resu
 	}()
 
 	// ensure firewalls
-	sourceRanges, err := helpers.GetLoadBalancerSourceRanges(l4.Service)
```
aojea, line 485: IMHO, Ideally we should remove all these helper dependencies from the cloud-provider 

```
grep -r "k8s.io/cloud-provider/service/helpers" pkg/
pkg/l4lb/l4netlbcontroller_test.go:     "k8s.io/cloud-provider/service/helpers"
pkg/l4lb/l4lbcommon.go: "k8s.io/cloud-provider/service/helpers"
pkg/utils/patch/patch.go:       svchelpers "k8s.io/cloud-provider/service/helpers"
pkg/neg/controller.go:  "k8s.io/cloud-provider/service/helpers"
pkg/loadbalancers/l4netlb_test.go:      servicehelper "k8s.io/cloud-provider/service/helpers"
pkg/loadbalancers/l4netlb.go:   "k8s.io/cloud-provider/service/helpers"
pkg/loadbalancers/l4.go:        "k8s.io/cloud-provider/service/helpers"
pkg/loadbalancers/l4_test.go:   servicehelper "k8s.io/cloud-provider/service/helpers"
pkg/healthchecksl4/healthchecksl4.go:   "k8s.io/cloud-provider/service/helpers"
```

cc: @swetharepakula

pkg/utils/sourceranges_test.go
```diff
@@ -0,0 +1,90 @@
...
+		expectedSourceRanges []string
+		expectError          error
+	}{
+		{
+			desc:                 "Should return allow all for no specs or annotations",
+			expectedSourceRanges: []string{"0.0.0.0/0"},
+		},
+		{
+			desc:                 "Should parse ranges from spec",
+			specRanges:           []string{" 192.168.0.1/10", " 132.8.0.1/8 "},
```
aojea, line 27: this was a kubernetes API validation mistake, but the ecosystem moved on allowing them 
https://github.com/kubernetes/kubernetes/pull/94107

so it seems correct to allow it here too

pkg/utils/sourceranges.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/net"
+)
+
+const (
+	allowAllIPv4Range = "0.0.0.0/0"
```
aojea, line 12: how does this work for IPv6 services? should not add the rules based on the existing Service ipFamilies?

ii.e. if service has ipv4 , add 0.0.0.0, if service has ipv6 add ::/0, or is the firewall in GCE agnostic of the family of the zero range?

panslava: I will add IPv6 in the next pr, now I wanted to keep logic kind of the same as before

PR #1899, opened 2023-01-10
Fix logs, use proper format strings argument types

panslava:
Simple logs fix

We should add logcheck linter to auto verify klog calls, created task for it


aojea:
/lgtm

yep, we should have the linter, let me see if I can work on it

PR #1904, opened 2023-01-18
NEG: Add wrappers for NEG Windows test.

shettyg:
With this commit, you can run the NEG tests for Windows with: -test.run=TestWindowsNEG.

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -33,6 +33,14 @@ import (
 )
 
 func TestNEG(t *testing.T) {
+	testNEGOS(t, e2e.Linux)
+}
+
+func TestWindowsNEG(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/e2e-test/neg_test.go
```diff
@@ -138,6 +146,14 @@ func TestNEG(t *testing.T) {
 }
 
 func TestNEGTransition(t *testing.T) {
+	testNEGTransitionOS(t, e2e.Linux)
+}
+
+func TestWindowsNEGTransition(t *testing.T) {
```
aojea, line 152: naive question, how do you skip this test if there are no windows nodes?

or is ok to fail in that case?

shettyg: It is okay to fail in that case.

swetharepakula (commented):
Thanks for the change!

/lgtm
/approve

PR #1906, opened 2023-01-19
Add delay for retrying NEG polling for unhealthy pods

alexkats:
This change introduces delay for health checks in NEG polling to avoid busy waiting and quota exceeded errors.

### DIFF
--- a/pkg/neg/readiness/poller.go
+++ b/pkg/neg/readiness/poller.go
@@ -41,7 +41,8 @@ const (
 	// GCE NEG API RPS quota is rate limited per every 100 seconds.
 	// Make this retry delay to match the rate limiting interval.
 	// More detail: https://cloud.google.com/compute/docs/api-rate-limits
-	retryDelay = 100 * time.Second
+	retryDelay   = 100 * time.Second
+	hcRetryDelay = time.Second
 )
 
 // negMeta references a GCE NEG resource
@@ -253,6 +254,10 @@ func (p *poller) processHealthStatus(key negMeta, healthStatuses []*composite.Ne
 		if patchCount < len(target.endpointMap) {
 			retry = true
 		}
+	}
+
+	if retry {
+		<-p.clock.After(hcRetryDelay)
 	}
 
 	// If we didn't patch all of the endpoints, we must keep polling for health status
--- a/pkg/neg/readiness/poller_test.go
+++ b/pkg/neg/readiness/poller_test.go
@@ -19,16 +19,17 @@ package readiness
 import (
 	"context"
 	"fmt"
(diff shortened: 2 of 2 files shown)


pkg/neg/readiness/poller.go
```diff
@@ -255,6 +256,10 @@ func (p *poller) processHealthStatus(key negMeta, healthStatuses []*composite.Ne
 		}
 	}
 
+	if retry {
+		<-p.clock.After(hcRetryDelay)
```
aojea, line 260: Is not easier use time.Sleep?

alexkats: I used this approach for 2 reasons:
1. Consistency with other usages in the code
2. It gives more control in the tests using fake clock

aojea: The clock has also a Sleep method so it can be mocked, afaik the After spawns a goroutine... No strong opinions just curiosity

alan-kut: +1 for p.clock.Sleep

pkg/neg/readiness/poller_test.go
```diff
@@ -559,6 +566,10 @@ func TestProcessHealthStatus(t *testing.T) {
 	res := []*composite.NetworkEndpointWithHealthStatus{}
 
 	// processHealthStatus should not crash when pollMap does not have corresponding key.
+	go func() {
+		time.Sleep(2 * time.Second)
```
alan-kut, line 570: Why do you need to sleep in test?

pkg/neg/readiness/poller.go
alexkats: Well, I know about `Sleep` in clock, but in the fake clock it does the same as `Step`. So it doesn't give enough control in the tests. Here're 2 examples:
1. When you need to check some condition before delay and after it, you can't block on it in tests.
2. When you want tests to fail with timeout if delay was not expected. When using `After` the execution will be blocked without `Step`, with `Sleep` it won't.

Also, AFAIK `After` doesn't spawn a goroutine, it has pretty the same system calls as `Sleep`, but uses a channel for communication

pkg/neg/readiness/poller_test.go
alexkats: Thanks for pointing out, checked again and realized this whole goroutine is useless, since processHealthCheck is not retried

pkg/neg/readiness/poller.go
aojea: you are right, my apologies, I should have checked it before , I mentally confused with another thing

apologies for the nitpicking

swetharepakula:
/lgtm
/approve

Thanks Alex!

PR #1910, opened 2023-01-23
Bump go version to 1.19

sugangli:


aojea:
this only bumps the go.mod, if you want to build with go 1.19 you have to switch the build image

https://github.com/kubernetes/ingress-gce/blob/f3c56e747e4098fc553a32f23ca2b9419bc8cde2/Makefile#L45

sugangli:
> this only bumps the go.mod, if you want to build with go 1.19 you have to switch the build image
> 
> https://github.com/kubernetes/ingress-gce/blob/f3c56e747e4098fc553a32f23ca2b9419bc8cde2/Makefile#L45

Oh right. Forgot about that. Updated.

PR #1912, opened 2023-01-24
Add metrics for syncer states

sawsa307:
Added metrics to collect the current count of syncers in each status.

### DIFF
--- a/pkg/neg/metrics/neg_metrics_collector.go
+++ b/pkg/neg/metrics/neg_metrics_collector.go
@@ -30,6 +30,9 @@ var (
 	syncResultLabel = "result"
 	syncResultKey   = "sync_result"
 
+	syncerStateLabel = "state"
+	syncerStateKey   = "syncer_state"
+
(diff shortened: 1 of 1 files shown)


pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -39,6 +42,16 @@ var (
 		},
 		[]string{syncResultLabel},
 	)
+
+	// syncerSyncerStatus tracks the count of syncer in different statuses
```
swetharepakula, line 46: nit: different states

swetharepakula (commented):
small nit, otherwise ready to be merged

pkg/neg/metrics/neg_metrics_collector.go
sawsa307: Updated. Thanks!

swetharepakula (commented):
/lgtm
/approve

PR #1913, opened 2023-01-25
make verify

aojea:
It seems we were not running correctly the verify, this PR
- adds verify and update targets to the makefile
- split the verify options into different scripts to simplify its maintainnce
- fix one gofmt error

The generation still doesn't work, it will be fixed by https://github.com/kubernetes/ingress-gce/pull/1884 but this PR adds an option to verify the code generated is correct

Next step is to add a new job to run `make verify` on the CI https://github.com/kubernetes/test-infra/pull/28507


aojea:
/close

superseded by https://github.com/kubernetes/ingress-gce/pull/1914

Issue #1870, opened 2022-11-29
Pin dependencies of update-codegen.sh


spencerhance:
While looking at #1855 it came up that we are not pinning versions of the two dependencies:

1. k8s.io/code-generator
2. k8s.io/kube-openapi/cmd/openapi-gen

We should pin these so that we can have reproducible results when running our API generation.

We should be able to use the same approach as golangci-lint in `hack/tools/tools.go` to install the dependencies.  Then we can access them like how it's done in `build/test.sh`.

swetharepakula:
cc/ @gauravkghildiyal

Issue #1855, opened 2022-11-06
Cannot use the script /hack/update-codegen.sh


julienMichaud:
Hi,

Im working on a PR which allow users to disable QUIC support throught the FrontendConfig CR. 

Im executing the script /hack/update-codegen.sh to generate the new files but I get the error :

`./hack/update-codegen.sh: line 32: ../code-generator/generate-groups.sh: No such file or directory`

Can you please tell me what im doing wrong ? 

Here is the attribute I added 

```
type FrontendConfigSpec struct {
	SslPolicy       *string              `json:"sslPolicy,omitempty"`
    + QuicOverride    *string              `json:"quicOverride,omitempty"`
	RedirectToHttps *HttpsRedirectConfig `json:"redirectToHttps,omitempty"`
}

```
Thanks

aojea:
It seems the scripts tries to find an script on the vendor folder for package `k8s.io/code-generator`

https://github.com/kubernetes/ingress-gce/blob/8195b0bb284e56265d606ee967bd4612d17a6662/hack/update-codegen.sh#L24

However, it is not present there ... @swetharepakula does this ring a bell? how `code-generator` dependency works in this repo?

spencerhance:
So this is a fun one, since code-generator is not in vendor anymore the script only works with the `|| echo ../code-generator` part, i.e. you need to git clone or go install it into your gopath next to ingress-gce.

For example, if you have this repo cloned into `go/src/k8s.io/ingress-gce`, you need to clone `k8s.io/code-generator` into `go/src/k8s.io/code-generator`.

Additionally, you will have to install openapi-gen with `go install k8s.io/kube-openapi/cmd/openapi-gen@latest` if  you haven't already.

julienMichaud:
thanks guys, @spencerhance I did what you said, here are the logs 

```
Generating composite types
WARNING: No description for type: BackendService.circuitBreakers
WARNING: No description for type: BackendService.connectionDraining
WARNING: No description for type: BackendService.connectionTrackingPolicy
WARNING: No description for type: BackendService.subsetting
WARNING: No description for type: HealthCheck.http2HealthCheck
WARNING: No description for type: HealthCheck.udpHealthCheck
WARNING: No description for type: HealthCheck.httpsHealthCheck
WARNING: No description for type: HealthCheck.grpcHealthCheck
WARNING: No description for type: HealthCheck.sslHealthCheck
WARNING: No description for type: HealthCheck.tcpHealthCheck
[...]

PR #1914, opened 2023-01-25
Fix code gen process and pin dependencies for update-codegen.sh

aojea:
Closes https://github.com/kubernetes/ingress-gce/issues/1870
Closes https://github.com/kubernetes/ingress-gce/issues/1855

This incorporates the great work of @gauravkghildiyal fixing the update codegen script.

In addition:
- add verify scripts and makefile target to use `make verify`
- make the update codegen script non-interactive so it can run on CI
- bump the golang ci lint version to fix a problem
- run update codegen to update current code


pkg/composite/gen.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
aojea, line 2: @spencerhance do we need to update these fails constantly?
I mean, every year this is going to fail

swetharepakula: I agree, this doesn't feel necessary.

PR #1915, opened 2023-01-25
use bash for verify scripts

aojea:
otherwise the scripts fails because `pipefail` is a bash only thing

```
hack/verify-all.sh
hack/verify-all.sh: 19: set: Illegal option -o pipefail
make: *** [Makefile:68: verify] Error 2
+ EXIT_VALUE=2
```

### DIFF
--- a/hack/verify-all.sh
+++ b/hack/verify-all.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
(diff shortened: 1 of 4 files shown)


aojea:
you can see it is passing now

swetharepakula (commented):
/lgtm
/approve

swetharepakula:
Thanks Antonio!

PR #1916, opened 2023-01-26
Inline dependency on legacy-cloud-providers

wstcliyu:
Remove the dependency on legacy-cloud-providers for pkg/annotations/service.go
to facilitate import into google3 as this library does not exist in google3.


MrHohn (approved):
/lgtm
/approve

PR #1917, opened 2023-01-26
Cleanup node port utils GetNodePorts

cezarygerard:
It is not used anymore after PR #1907


panslava:
/lgtm 

why not include in 1.21?

PR #1920, opened 2023-01-26
Add an environment variable for disabling the instance group controller

ingwonsong:
This PR proposes a new environment variable `EXPERIMENTAL_ONLY_FOR_TESTING_DISABLE_INSTANCE_GROUP_CONTROLLER` for disabling the instance group controller for the testing env.

When we are doing a test or dev on non-GCP, the instance group controller is not needed, so this env var provides the way to turn off the instance group controller. Please note that it is disabled only when the env var has the value `true` (case insensitive).

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -358,14 +358,16 @@ func runControllers(ctx *ingctx.ControllerContext) {
 
 	ctx.Start(stopCh)
 
-	igControllerParams := &instancegroups.ControllerConfig{
-		NodeInformer: ctx.NodeInformer,
-		IGManager:    ctx.InstancePool,
-		HasSynced:    ctx.HasSynced,
-		StopCh:       stopCh,
+	if flags.F.RunInstanceGroupController {
+		igControllerParams := &instancegroups.ControllerConfig{
+			NodeInformer: ctx.NodeInformer,
+			IGManager:    ctx.InstancePool,
+			HasSynced:    ctx.HasSynced,
+			StopCh:       stopCh,
+		}
+		igController := instancegroups.NewController(igControllerParams)
+		go igController.Run()
 	}
-	igController := instancegroups.NewController(igControllerParams)
-	go igController.Run()
 
 	// The L4NetLbController will be run when RbsMode flag is Set
 	if flags.F.RunL4NetLBController {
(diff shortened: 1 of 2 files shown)


aojea:
/assign @cezarygerard @swetharepakula 

> When we are doing a test or dev on non-GCP, the instance group controller is not needed, so this flag provides the way to turn off the instance group controller. Default is true to keep the compatibility.

exposing a flag just for testing sounds like a footgun for users, can we make this option more hidden?

ingwonsong:
> exposing a flag just for testing sounds like a footgun for users, can we make this option more hidden?

Make sense. Then, what do you think about an env var?

aojea:
> Make sense. Then, what do you think about an env var?

I'm in favor of that,  exaggerating :) `EXPERIMENTAL_ONLY_FOR_TESTING_DISABLE_INSTANCE_GROUP_CONTROLLER`

ingwonsong:
> I'm in favor of that, exaggerating :) `EXPERIMENTAL_ONLY_FOR_TESTING_DISABLE_INSTANCE_GROUP_CONTROLLER`

Sounds great. Will change to use it.

ingwonsong:
> > I'm in favor of that, exaggerating :) `EXPERIMENTAL_ONLY_FOR_TESTING_DISABLE_INSTANCE_GROUP_CONTROLLER`
> 
> Sounds great. Will change to use it.

Done.

cmd/glbc/main.go
```diff
@@ -358,14 +359,16 @@ func runControllers(ctx *ingctx.ControllerContext) {
 
 	ctx.Start(stopCh)
 
-	igControllerParams := &instancegroups.ControllerConfig{
-		NodeInformer: ctx.NodeInformer,
-		IGManager:    ctx.InstancePool,
-		HasSynced:    ctx.HasSynced,
-		StopCh:       stopCh,
+	if strings.ToLower(os.Getenv("EXPERIMENTAL_ONLY_FOR_TESTING_DISABLE_INSTANCE_GROUP_CONTROLLER")) != "true" {
```
aojea, line 362: I suggested this name as a kind of joke but this works for me 😄

ingwonsong: Yes. it works 😄

ingwonsong:
At this time, just forked version would be enough, because this is just for testing.

PR #1922, opened 2023-01-27
Add E2E test for reduced ASM mode

swetharepakula:
* when ASM enabled, neg controller should create negs for services and will not trigger based on destination rules or subset

### DIFF
--- a/cmd/e2e-test/asm_test.go
+++ b/cmd/e2e-test/asm_test.go
@@ -10,8 +10,10 @@ import (
 
 	istioV1alpha3 "istio.io/api/networking/v1alpha3"
 	apiv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/e2e"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/e2e-test/asm_test.go
```diff
@@ -285,3 +287,95 @@ func TestNoIstioASM(t *testing.T) {
 
 	})
 }
+
+// TestGSMService tests the service/destinationrule, it can't run parallel with other tests.
+func TestGSMService(t *testing.T) {
+	_ = apiv1.ComponentStatus{}
```
bowei, line 293: What is this used for?

cmd/e2e-test/asm_test.go
```diff
@@ -285,3 +287,95 @@ func TestNoIstioASM(t *testing.T) {
...
+
+	// negStatus is used to verify GC in the zones NEGs were created. The actual negStatus is not important
+	// and therefore we can use the last generated neg status.
+	var negStatus *annotations.NegStatus
+
+	// This test case will need two namespaces, one will in asm-skip-namespaces.
+	Framework.RunWithSandbox("TestASMService", t, func(t *testing.T, sSkip *e2e.Sandbox) {
```
bowei, line 300: I recommend extending the RunWithSandbox to be able to give multiple namespaces than nesting like this, which can make future maintenance tricky as I'm not sure what the semantics would be.

cmd/e2e-test/asm_test.go
```diff
@@ -285,3 +287,95 @@ func TestNoIstioASM(t *testing.T) {
...
+		Framework.RunWithSandbox("TestASMService", t, func(t *testing.T, s *e2e.Sandbox) {
+			// Enable ASM mode
+			ctx := context.Background()
```
bowei, line 303: Move this down to where it is used.

cmd/e2e-test/asm_test.go
```diff
@@ -285,3 +287,95 @@ func TestNoIstioASM(t *testing.T) {
...
+
+			asmConfig := map[string]string{"enable-asm": "true",
+				"asm-skip-namespaces": fmt.Sprintf("kube-system,istio-system,%s", sSkip.Namespace)}
```
bowei, line 306: could use "..." + sSkip.Namespace instead of fmt [minor]

cmd/e2e-test/asm_test.go
```diff
@@ -285,3 +287,95 @@ func TestNoIstioASM(t *testing.T) {
...
+			if err := e2e.EnsureConfigMap(s, asmConfigNamespace, asmConfigName,
+				asmConfig); err != nil {
+				t.Error(err)
```
bowei, line 309: t.Error or t.Fatal?

cmd/e2e-test/asm_test.go
swetharepakula: I am not sure, this is a duplication of the original ASM test with removal of the destination rule pieces. I will test locally and ensure we don't need that.

cmd/e2e-test/asm_test.go
swetharepakula: Extended the sandbox to be able to create and track additional namespaces.

PR #1924, opened 2023-01-27
Export various stats about services in the metrics exported by this c…

mmamczur:
…ontroller.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -113,10 +113,11 @@ type ControllerContext struct {
 
 // ControllerContextConfig encapsulates some settings that are tunable via command line flags.
 type ControllerContextConfig struct {
-	Namespace         string
-	ResyncPeriod      time.Duration
-	NumL4Workers      int
-	NumL4NetLBWorkers int
+	Namespace               string
+	ResyncPeriod            time.Duration
+	NumL4Workers            int
+	NumL4NetLBWorkers       int
+	NumServiceMetricWorkers int
 	// DefaultBackendSvcPortID is the ServicePort for the system default backend.
 	DefaultBackendSvcPort utils.ServicePort
 	HealthCheckPath       string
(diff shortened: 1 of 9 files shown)


pkg/metrics/metrics.go
```diff
@@ -644,6 +746,29 @@ func (im *ControllerMetrics) computeServiceMetrics() map[feature]int {
...
+}
+
+func (im *ControllerMetrics) computeServiceStatsMetrics() {
+	im.Lock()
+	defer im.Unlock()
+	start := im.clock.Now()
+	// reset the existing counters so that services that were deleted are reported as 0 (there would be stale values otherwise).
+	im.serviceStatsMetricCounts.zeroCounts()
+	for key, serviceState := range im.serviceStatsMap {
+		// remove data for services that does no longer have fresh data (could have been deleted and the delete event was missed).
```
aojea, line 760: this can not happen .. well if there is a bug in the informer

pkg/context/context.go
```diff
@@ -152,7 +153,7 @@ func NewControllerContext(
 		ClusterNamer:            clusterNamer,
 		L4Namer:                 namer.NewL4Namer(string(kubeSystemUID), clusterNamer),
 		KubeSystemUID:           kubeSystemUID,
-		ControllerMetrics:       metrics.NewControllerMetrics(flags.F.MetricsExportInterval, flags.F.L4NetLBProvisionDeadline),
+		ControllerMetrics:       metrics.NewControllerMetrics(flags.F.MetricsExportInterval, flags.F.L4NetLBProvisionDeadline, config.ResyncPeriod),
```
aojea, line 156: I think you don't need to resync, the resync generate a fake event update on each item of the cache to be processed again

pkg/metrics/metrics.go
aojea: oh, I see, so you store the data in another cache

aojea: I see what happens, previous implementation is based on those caches ... just FYI, the informer has already a cache that represents current state of the cluster, this additional cache is not necessary, the informer keeps it up to date , it should be something like 

```
```
	serviceLister := ServiceInformer.GetIndexer()
	services, err := listers.NewServiceLister(serviceLister).List(labels.Everything())
        if err != nil {

       }
      for _, service := range services {
     // get metrics from service
   } 
```

PR #1926, opened 2023-01-31
Remove Services.Get() from syncNegStatusAnnotation()

gauravkghildiyal:
/assign @swetharepakula

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -17,14 +17,12 @@ limitations under the License.
 package neg
 
 import (
-	"context"
 	"fmt"
 	"time"
 
 	apiv1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
 	v1 "k8s.io/api/networking/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


mmamczur:
we tried a very similar thing along with adding more goroutines for the controller itself. It seemed to definitely help in manual testing.

I'm not sure why it was getting the service from a client before though. Can patch fail if we give it an outdated version?

gauravkghildiyal:
@swetharepakula gave some historical context for this: Fetching the service from the api-server was needed when `patch` didn't exist and we had to work with `update`s APIs. When we replaced the APIs to use `patch` instead of `update`, replacing this specific Get call to the api-server most likely got missed.

There are existing places in our code ([example1](https://github.com/gauravkghildiyal/ingress-gce/blob/fa6f5a6df062f1ba7c419fc068db4bb2fbbb276b/pkg/neg/syncers/transaction.go#L665), [example2](https://github.com/gauravkghildiyal/ingress-gce/blob/fa6f5a6df062f1ba7c419fc068db4bb2fbbb276b/pkg/neg/syncers/transaction.go#L693)) where we do fetch from the informer and then perform the update and I feel this should be fine too.

There's atleast one other place I know where we could also fetch from the informer. That, along with increasing the number of workers and a few other optimizations is something that I'm working on and will soon send them in separate PRs.

mmamczur:
let's have this merged then (I can't review nor lgtm yet)
/lgtm

swetharepakula (commented):
/lgtm
/approve

gauravkghildiyal:
> New changes are detected. LGTM label has been removed.

FYI: Just rebased -- no new changes

aojea:
you have to remove some unused code now

> pkg/neg/controller.go:20:2: imported and not used: "context"
> pkg/neg/controller.go:27:2: imported and not used: "k8s.io/apimachinery/pkg/apis/meta/v1" as metav1

PR #1927, opened 2023-02-01
Add metrics to track syncer staleness

sawsa307:
Added metrics to track the sync staleness of NEG syncers, where staleness is defined as how long since the syncer last syncs.

### DIFF
--- a/pkg/neg/metrics/metrics.go
+++ b/pkg/neg/metrics/metrics.go
@@ -31,6 +31,7 @@ const (
 	negOpLatencyKey          = "neg_operation_duration_seconds"
 	negOpEndpointsKey        = "neg_operation_endpoints"
 	lastSyncTimestampKey     = "sync_timestamp"
+	syncerStalenessKey       = "syncer_staleness"
 
 	resultSuccess = "success"
 	resultError   = "error"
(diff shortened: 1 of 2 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -127,6 +128,17 @@ var (
...
 	)
+
+	// SyncerStaleness tracks for every syncer, how long has it been since the last sync occured
+	SyncerStaleness = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      syncerStalenessKey,
+			Help:      "The duration that NEG syncers have been stale(out of sync)",
+			// custom buckets - [1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048 (~34min), 4096(~68min), +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
```
swetharepakula, line 139: Increase to one more bucket. I don't think we should see anything that high, but I am thinking one more bucket after an hour would be good.

pkg/neg/syncers/transaction.go
```diff
@@ -149,6 +152,7 @@ func NewTransactionSyncer(
 		enableEndpointSlices: enableEndpointSlices,
 		inError:              false,
 		logger:               logger,
+		lastSyncTimestamp:    time.Now(),
```
swetharepakula, line 155: This is I think is available on the svcneg CR already. Can we reuse?

swetharepakula (commented):
A couple nits, but otherwise this looks good to me.

pkg/neg/metrics/metrics.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
sawsa307: Hi Swetha are you referring to this part? 
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/controller.go#L397-L400
These two are different since this one is tracking at controller level, and the above one is tracking at the neg/syncer level.

swetharepakula: Discussed offline, I was referring to https://github.com/kubernetes/ingress-gce/blob/master/pkg/apis/svcneg/v1beta1/types.go#L56. We set this every sync, so I think we can reuse that value.

swetharepakula (commented):
/lgtm
/approve

PR #1928, opened 2023-02-01
Trigger updating DualStack ILB Service if ipFamilies changed

panslava:
Currently, updating ipFamilies do not trigger L4 ILB service update. This is a bug fix

/assign cezarygerard
/assign mmamczur

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -482,6 +482,11 @@ func (l4c *L4Controller) needsUpdate(oldService *v1.Service, newService *v1.Serv
 			oldService.Spec.HealthCheckNodePort, newService.Spec.HealthCheckNodePort)
 		return true
 	}
+	if l4c.enableDualStack && !reflect.DeepEqual(oldService.Spec.IPFamilies, newService.Spec.IPFamilies) {
+		recorder.Eventf(newService, v1.EventTypeNormal, "IPFamilies", "%v -> %v",
+			oldService.Spec.IPFamilies, newService.Spec.IPFamilies)
+		return true
+	}
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -482,6 +482,11 @@ func (l4c *L4Controller) needsUpdate(oldService *v1.Service, newService *v1.Serv
 			oldService.Spec.HealthCheckNodePort, newService.Spec.HealthCheckNodePort)
 		return true
 	}
+	if l4c.enableDualStack && !reflect.DeepEqual(oldService.Spec.IPFamilies, newService.Spec.IPFamilies) {
```
aojea, line 485: or maybe comparing ClusterIPs ?

panslava: I was considering it, but I think IPFamilies are better,

All the logic in sync depends on IPFamilies (to identify if service needs ipv4 or ipv6), and, we don't really need info about exactly clusterIPs changes (even though they should not change without ipFamilies..), we only need to know ipFamilies state

mmamczur: do we care about ordering here?
if it changes from IPv4,IPv6 -> IPv6,IPv4 should this trigger an update?

panslava: good point, just, by kubernetes design, first ipFamily should never change, because 1st ipFamily correlates with ClusterIP which is immutable (not ClusterIPs)

panslava: but maybe I should not rely on kubernetes still?

mmamczur: so it's super unlikely anyway because it would be rejected by the apiserver
I'd leave that since one extra 'update' won't hurt us

panslava: Thanks! I think so too :)

PR #1929, opened 2023-02-02
Change resync period to 10 minutes

panslava:
With short resync period, when running locally, it is easy to miss bugs, when service update is not triggered, but service is quickly updated due to resync period (example https://github.com/kubernetes/ingress-gce/pull/1928) 

/assign cezarygerard
/assign aojea


aojea:
hmm, is this the field that goes to the informers, because if that is the case this is wrong and can cause a lot of performance issues

aojea:
worth reading https://groups.google.com/g/kubernetes-sig-api-machinery/c/PbSCXdLDno0 

This has a lot of performance implications, why do we need such short resync period, that means that we'll be processing the cache every that time

PR #1930, opened 2023-02-02
Add metrics to track endpointslice staleness

sawsa307:
Added metrics to track the sync staleness of endpointslices, where staleness is defined as how long since it has been last processed.

### DIFF
--- a/pkg/neg/metrics/metrics.go
+++ b/pkg/neg/metrics/metrics.go
@@ -32,6 +32,7 @@ const (
 	negOpEndpointsKey        = "neg_operation_endpoints"
 	lastSyncTimestampKey     = "sync_timestamp"
 	syncerStalenessKey       = "syncer_staleness"
+	epsStalenessKey          = "endpointslice_staleness"
 
 	resultSuccess = "success"
 	resultError   = "error"
@@ -135,6 +136,15 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      syncerStalenessKey,
 			Help:      "The duration of a syncer since it last syncs",
+		},
+	)
+
+	// EPSStaleness tracks for every endpoint slice, how long since it was last processed
+	EPSStaleness = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      epsStalenessKey,
+			Help:      "The duration for an endpoint slice since it was last processed by syncer",
 			// custom buckets - [1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048 (~34min), 4096(~68min), 8192(~136min), +Inf]
 			Buckets: prometheus.ExponentialBuckets(1, 2, 14),
 		},
@@ -152,6 +162,7 @@ func RegisterMetrics() {
 		prometheus.MustRegister(LastSyncTimestamp)
 		prometheus.MustRegister(InitializationLatency)
 		prometheus.MustRegister(SyncerStaleness)
+		prometheus.MustRegister(EPSStaleness)
 
 		RegisterSyncerMetrics()
 	})
@@ -185,6 +196,10 @@ func PublishNegInitializationMetrics(latency time.Duration) {
 
 func PublishNegSyncerStalenessMetrics(syncerStaleness time.Duration) {
 	SyncerStaleness.Observe(syncerStaleness.Seconds())
+}
+
+func PublishNegEPSStalenessMetrics(epsStaleness time.Duration) {
+	EPSStaleness.Observe(epsStaleness.Seconds())
 }
 
 func getResult(err error) string {
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -245,8 +245,25 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			return nil
 		}
 		endpointSlices := make([]*discovery.EndpointSlice, len(slices))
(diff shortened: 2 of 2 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -127,6 +128,17 @@ var (
...
 	)
+
+	// EPSStaleness tracks for every endpoint slice, how long since it was last processed
+	EPSStaleness = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      epsStalenessKey,
+			Help:      "The duration for an endpoint slice since it was last processed by syncer",
+			// custom buckets - [1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048 (~34min), 4096(~68min), +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
```
swetharepakula, line 139: Can you extend by one more bucket? Thanks!

pkg/neg/syncers/transaction.go
```diff
@@ -240,9 +244,31 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+			endpointslice := slice.(*discovery.EndpointSlice)
+			endpointSlices[i] = endpointslice
+			key := keyFunc(endpointslice.ObjectMeta.Namespace, endpointslice.ObjectMeta.Name)
+			if lastSyncTime, contains := s.epsLastSyncTimestampMap[key]; contains {
+				staleness := currTime.Sub(lastSyncTime)
+				metrics.PublishNegEPSStalenessMetrics(staleness)
+				s.logger.V(3).Info("Endpoint slice syncs", "Namespace/Name", key, "staleness", staleness)
+
+				// remove from previous map since this eps is updated
+				delete(s.epsLastSyncTimestampMap, key)
```
swetharepakula, line 259: For this metric, I don't think we need to track the lastSyncTimestamp per EPS. we have the following cases:
1. EPS newly created since last syncer sync
2. EPS existed previously and was processed during the lastSyncer Sync

Since we already know when we last synced the syncer (from the svcNeg CR), you just need to check if the EPS creation timestamp is before or after the lastSyncTimestamp. If the creationTS is after now() then you publish now()-creationTS() for that endpoint's staleness. If creationTS is before now() then you publish now()-syncerLastSync timestamp.

Otherwise you have to add logic that will allow you to recognize when an EPS was deleted as you will never delete those from the map.

pkg/neg/syncers/transaction.go
```diff
@@ -245,8 +245,27 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			return nil
 		}
 		endpointSlices := make([]*discovery.EndpointSlice, len(slices))
+		negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.NegSyncerKey.NegName)
+		if err != nil {
+			s.logger.Error(err, "unable to retrieve neg from the store", "neg", klog.KRef(s.Namespace, s.NegName))
+			return err
```
swetharepakula, line 251: since this is only for metrics, do not return an error.

pkg/neg/syncers/transaction.go
```diff
@@ -245,8 +245,27 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+			var epsStaleness time.Duration
+			lastSyncTimestamp := negCR.Status.LastSyncTime
+			epsCreationTimestamp := endpointslice.ObjectMeta.CreationTimestamp
+
+			// if this endpoint slice is newly created/created after last sync
+			if lastSyncTimestamp.Before(&epsCreationTimestamp) {
+				epsStaleness = time.Since(epsCreationTimestamp.Time)
+			} else {
+				epsStaleness = time.Since(lastSyncTimestamp.Time)
+			}
```
swetharepakula, line 266: We can remove the else and you don't need line 257 either.
```
epsStaleness := time.Since(lastSyncTimestamp.Time)
if lastSyncTimestamp.Before(&epsCreationTimestamp) {
    epsStaleness = time.Since(epsCreationTimestamp.Time)
} 
```

pkg/neg/syncers/transaction.go
```diff
@@ -246,7 +246,25 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		}
 		endpointSlices := make([]*discovery.EndpointSlice, len(slices))
 		for i, slice := range slices {
-			endpointSlices[i] = slice.(*discovery.EndpointSlice)
+			endpointslice := slice.(*discovery.EndpointSlice)
+			endpointSlices[i] = endpointslice
+
+			negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.NegSyncerKey.NegName)
```
swetharepakula, line 252: move this outside of the for loop. There is only one NEG per syncer.

pkg/neg/syncers/transaction.go
```diff
@@ -246,7 +246,25 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+			if err != nil {
+				s.logger.Error(err, "unable to retrieve neg from the store", "neg", klog.KRef(s.Namespace, s.NegName))
```
swetharepakula, line 254: you can add a continue after this. Then you don't need the else afterwards.

pkg/neg/syncers/transaction.go
```diff
@@ -246,7 +246,25 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+			} else {
+				var epsStaleness time.Duration
```
swetharepakula, line 256: remove this. Initialize when you will use it

swetharepakula (commented):
/lgtm
/approve

PR #1931, opened 2023-02-02
Refactor syncNetworkEndpoints

sawsa307:
Add a waitgroup that waits for the completeness of all NEG operations.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -97,16 +97,18 @@ type transactionSyncer struct {
 
 	logger klog.Logger
 
-	// inError indicates if the syncer is in any of 4 error scenarios
+	// errorState indicates if the syncer is in any of 4 error scenarios
 	// 1. Endpoint counts from EPS is different from calculated endpoint list
 	// 2. EndpontSlice has missing or invalid data
 	// 3. Attach/Detach EP fails due to incorrect batch information
 	// 4. Endpoint count from EPS or calculated endpoint list is 0
 	// Need to grab syncLock first for any reads or writes based on this value
-	inError bool
+	errorState string
 
 	// syncCollector collect sync related metrics
 	syncCollector metrics.SyncerMetricsCollector
+
+	wg sync.WaitGroup
 }
 
 func NewTransactionSyncer(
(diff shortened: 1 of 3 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -447,43 +450,66 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
...
 }
 
+// collectSyncResult spins off go routines to execute NEG operations
+func (s *transactionSyncer) collectSyncResult(attachBatchMap map[string][]EndpointBatch, detachBatchMap map[string][]EndpointBatch) {
+	s.logger.V(3).Info("Start collecting result.")
+
+	for zone, batches := range attachBatchMap {
+		for _, batch := range batches {
+			s.logger.V(3).Info("wg increse by one")
+			s.wg.Add(1)
```
swetharepakula, line 479: I think we can greatly simplify this refactor by adding to the wait groups into attachNetworkEndpoints and detachNetworkEndpoints instead which will be marked as done in operationInternal.

swetharepakula: Just realized that attach and detach NetworkEndpoints will start off more go routines so we cannot just add inside attach and detach.

sawsa307: Thank you! As discussed offline, now collectSyncResult will only wait for all goroutine to finish and emit metrics .

pkg/neg/syncers/transaction.go
```diff
@@ -465,25 +467,35 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
...
 		return err
 	}
+	go s.collectSyncResult()
 	return nil
 }
 
+// collectSyncResult collects the result of the sync and emits the metrics for sync result
+func (s *transactionSyncer) collectSyncResult() {
+	s.wg.Wait()
+	s.logger.V(3).Info("Done collecting result.")
```
swetharepakula, line 477: Shouldn't this publish something about the result or metric?

pkg/neg/syncers/transaction.go
```diff
@@ -465,25 +467,34 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
...
 	}
+	go s.collectSyncResult()
 	return nil
 }
 
+// collectSyncResult collects the result of the sync and emits the metrics for sync result
+func (s *transactionSyncer) collectSyncResult() {
+	s.wg.Wait()
+}
+
```
swetharepakula, line 478: For this function, what is the proposed way to collect the results? How will you get the aggregate result of all the goroutines?

swetharepakula (commented):
/lgtm
/approve

PR #1937, opened 2023-02-07
Add L4 NetLB Dual-Stack Metrics

panslava:
This PR adds counting service per ipFamilies, ipFamilyPolicy and Status for L4 NetLB DualStack

As ILB and NetLB DualStack metrics store the same data, I changed type L4ILBDualStackServiceState to be just general L4DualStackServiceState

I will add PersistentError (status to report after 20 mins of errors) in the next PR for both ILB and NetLB


cezarygerard:
we discussed it at te desk
good job Slavik!

PR #1940, opened 2023-02-10
add aojea as approver

aojea:


OWNERS
```diff
@@ -1,4 +1,5 @@
 approvers:
+- aojea
```
panslava, line 2: is it fine, that you still listed in reviewers?

aojea: yes, is ok

PR #1941, opened 2023-02-10
[Cherry pick #1907] to release 1.20 Remove using node ports for L4 NetLB RBS Services

cezarygerard:
[Cherry pick #1907] to release 1.20 Remove using node ports for L4 NetLB RBS Services

NodePorts are not needed for NetLB RBS and were copied from legacy implementation. They were also slowing down our controller, because we were updating same instance group with node port for every service


aojea:
don't worry for the verify job, it is no longer required in this branch, it will merge once someone lgtm the PR

image


news PRs should not run it

PR #1943, opened 2023-02-10
Export various stats about services in the metrics exported by this c…

mmamczur:
…ontroller.

### DIFF
--- a/pkg/servicemetrics/servicemetrics.go
+++ b/pkg/servicemetrics/servicemetrics.go
@@ -1,0 +1,369 @@
+package servicemetrics
+
+import (
+	"fmt"
+	"k8s.io/utils/net"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/common"
+	"k8s.io/klog/v2"
+	"k8s.io/legacy-cloud-providers/gce"
+)
+
+const (
+	// Names of the labels used by the service metrics.
+	labelType                  = "type"
+	labelExternalTrafficPolicy = "external_traffic_policy"
+	labelInternalTrafficPolicy = "internal_traffic_policy"
+	labelSessionAffinityConfig = "session_affinity_config"
+	labelProtocol              = "protocol"
+	labelIPFamilies            = "ip_families"
+	labelIPFamilyPolicy        = "ip_family_policy"
+	labelIsStaticIPv4          = "is_static_ip_v4"
+	labelIsStaticIPv6          = "is_static_ip_v6"
+	labelNetworkTier           = "network_tier"
+	labelGlobalAccess          = "global_access"
+	labelCustomSubnet          = "custom_subnet"
+	labelNumberOfPorts         = "number_of_ports"
+
+	// possible values for the service_type label
+	serviceTypeSubsettingILB = "SubsettingILB"
+	serviceTypeRBSXLB        = "RBSXLB"
+	serviceTypeLegacyILB     = "LegacyILB"
+	serviceTypeLegacyXLB     = "LegacyXLB"
+
+	// sessionAffinityTimeoutDefault is the default timeout value for a service session affinity.
+	sessionAffinityTimeoutDefault = 10800
+
+	//  possible values for the session_affinity_config label.
+	sessionAffinityBucketMoreThanDefault = "10800+"
+	sessionAffinityBucketDefault         = "10800"
+	sessionAffinityBucketLessThanDefault = "0-10799"
+	sessionAffinityBucketNone            = "None"
+)
+
+var (
+	serviceL4ProtocolStatsCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "service_l4_protocol_stats",
+			Help: "Number of services broken down by various stats",
+		},
+		[]string{
+			labelType,
+			labelExternalTrafficPolicy,
+			labelInternalTrafficPolicy,
+			labelSessionAffinityConfig,
+			labelProtocol,
+			labelNumberOfPorts,
+		},
+	)
+	serviceIPStackStatsCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "service_ip_stack_stats",
+			Help: "Number of services broken down by various stats",
+		},
+		[]string{
+			labelType,
+			labelExternalTrafficPolicy,
+			labelInternalTrafficPolicy,
+			labelIPFamilies,
+			labelIPFamilyPolicy,
+			labelIsStaticIPv4,
+			labelIsStaticIPv6,
+		},
+	)
+	serviceGCPFeaturesStatsCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "service_gcp_features_stats",
+			Help: "Number of services broken down by various stats",
+		},
+		[]string{
+			labelType,
+			labelNetworkTier,
+			labelGlobalAccess,
+			labelCustomSubnet,
+		},
+	)
+)
+
+func init() {
+	klog.V(3).Infof("Registering Service stats usage metrics %v", serviceL4ProtocolStatsCount)
+	prometheus.MustRegister(serviceL4ProtocolStatsCount)
+	prometheus.MustRegister(serviceIPStackStatsCount)
+	prometheus.MustRegister(serviceGCPFeaturesStatsCount)
+}
+
+// Controller is the controller that exposes and populates metrics containing various stats about Services in the cluster.
+type Controller struct {
+	ctx             *context.ControllerContext
+	stopCh          chan struct{}
+	svcQueue        utils.TaskQueue
+	metricsInterval time.Duration
+	serviceInformer cache.SharedIndexInformer
+}
+
+// NewController creates a new Controller.
+func NewController(ctx *context.ControllerContext, exportInterval time.Duration, stopCh chan struct{}) *Controller {
+	svcMetrics := &Controller{
+		ctx:             ctx,
+		stopCh:          stopCh,
+		serviceInformer: ctx.ServiceInformer,
+		metricsInterval: exportInterval,
+	}
+	return svcMetrics
+}
+
+// Run starts the controller until stopped via the stop channel.
+func (c *Controller) Run() {
+	klog.Infof("Starting Service Metric Stats controller")
+	go func() {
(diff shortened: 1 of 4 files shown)


pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+	NetworkTier  string
+	GlobalAccess bool
+	CustomSubnet bool
+}
+
+func (c *Controller) export() {
+	serviceLister := c.serviceInformer.GetIndexer()
+	services, err := listers.NewServiceLister(serviceLister).List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to list services err=%v", err)
```
aojea, line 119: shouldn't you log and return? 

if you have an error most probable `services` is nil and we'll record unknown metrics

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+		Protocol:              getProtocol(service.Spec.Ports),
+	}
+	ipStack := &serviceIPStackMetricState{
+		Type:                  serviceType,
+		ExternalTrafficPolicy: externalTrafficPolicy,
+		InternalTrafficPolicy: internalTrafficPolicy,
+		IPFamilies:            getIPFamilies(service.Spec.IPFamilies),
+		IPFamilyPolicy:        getIPFamilyPolicy(service.Spec.IPFamilyPolicy),
+		IsStaticIPv4:          service.Spec.LoadBalancerIP != "",
+		IsStaticIPv6:          false,
```
aojea, line 192: @panslava is this assumption right? there are never StaticIPv6?

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func getInternalTrafficPolicy(service *v1.Service) string {
+	if service.Spec.InternalTrafficPolicy == nil {
+		return string(v1.ServiceInternalTrafficPolicyCluster)
+	}
+	return string(*service.Spec.InternalTrafficPolicy)
+}
+
+func getPortsBucket(ports []v1.ServicePort) string {
```
aojea, line 218: prometheus histogram metrics only need to receive the data and it already. manages the bucket, do we need to export this as a string?

pkg/servicemetrics/servicemetrics.go
panslava: for now they are not supported in Google Cloud

pkg/servicemetrics/servicemetrics.go
mmamczur: so here we want to limit the number of possible values in the label and we decided on these specific 'bucket' ourselves. This is intended to limit the cardinality of the metric

pkg/servicemetrics/servicemetrics.go
mmamczur: I should, fixed!

pkg/servicemetrics/servicemetrics_test.go
```diff
@@ -0,0 +1,688 @@
...
+		{
+			desc: "2-5",
+			ports: []v1.ServicePort{{
+				Name: "p1",
+				Port: 80,
+			},
+				{
+					Name: "p2",
+					Port: 123456,
+				}},
```
cezarygerard, line 405: please improve the formatting

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+	updatePrometheusMetrics(l4ProtocolState, ipStackState, gcpFeaturesState)
+}
+
+func groupServices(services []*v1.Service) (map[serviceL4ProtocolMetricState]int64, map[serviceIPStackMetricState]int64, map[serviceGCPFeaturesMetricState]int64) {
+	l4ProtocolState := make(map[serviceL4ProtocolMetricState]int64)
+	ipStackState := make(map[serviceIPStackMetricState]int64)
+	gcpFeaturesState := make(map[serviceGCPFeaturesMetricState]int64)
+
+	for _, service := range services {
+		l4Protocol, ipStack, gcpFeatures := metricsForService(service)
```
cezarygerard, line 133: metrics from service?

pkg/servicemetrics/servicemetrics_test.go
```diff
@@ -0,0 +1,688 @@
...
+		IsStaticIPv6:          false,
+	}
+	gcpFeaturesKeyS1AndS2 := serviceGCPFeaturesMetricState{
+		Type:         "RBSXLB",
+		NetworkTier:  "Premium",
+		GlobalAccess: false,
+		CustomSubnet: false,
+	}
+
+	l4ProtocolState, ipStackState, gcpFeaturesState := groupServices(services)
```
cezarygerard, line 270: groupServices is a misleading name

maybe calculateMetrics(allServices)  /* if you also rename services variable to allServices */

cezarygerard:
Michał

overall LGTM from my side

please take from my comments whatever you find useful and we can commit soon
drop the comments that you find unuseful

thanks!

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+		l4ProtocolState[*l4Protocol]++
+		ipStackState[*ipStack]++
+		gcpFeaturesState[*gcpFeatures]++
+	}
+	return l4ProtocolState, ipStackState, gcpFeaturesState
+}
+
+func updatePrometheusMetrics(l4ProtocolState map[serviceL4ProtocolMetricState]int64, ipStackState map[serviceIPStackMetricState]int64, gcpFeaturesState map[serviceGCPFeaturesMetricState]int64) {
```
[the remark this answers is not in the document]
mmamczur: done

pkg/servicemetrics/servicemetrics_test.go
```diff
@@ -0,0 +1,688 @@
...
+	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/test"
+	"k8s.io/ingress-gce/pkg/utils/common"
+	namer_util "k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/legacy-cloud-providers/gce"
+)
+
+const clusterUID = "aaaaa"
+
+func createController() *Controller {
```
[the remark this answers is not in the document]
mmamczur: removed

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+	for serviceStat, count := range l4ProtocolState {
+		serviceL4ProtocolStatsCount.With(prometheus.Labels{
+			"type":                    serviceStat.Type,
```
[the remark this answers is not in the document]
mmamczur: added consts for all labels

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+			return "RBSXLB"
+		}
+		return "LegacyXLB"
+	}
+	return ""
+}
+
+func getSessionAffinityConfig(service *v1.Service) string {
+	if service.Spec.SessionAffinity != v1.ServiceAffinityClientIP {
+		return ""
```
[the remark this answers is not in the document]
mmamczur: changed

pkg/servicemetrics/servicemetrics_test.go
mmamczur: done

pkg/servicemetrics/servicemetrics.go
mmamczur: done

pkg/servicemetrics/servicemetrics_test.go
mmamczur: changed

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,304 @@
...
+		NumberOfPorts:         getPortsBucket(service.Spec.Ports),
+		Protocol:              getProtocol(service.Spec.Ports),
+	}
+	ipStack := &serviceIPStackMetricState{
+		Type:                  serviceType,
+		ExternalTrafficPolicy: externalTrafficPolicy,
+		InternalTrafficPolicy: internalTrafficPolicy,
+		IPFamilies:            getIPFamilies(service.Spec.IPFamilies),
+		IPFamilyPolicy:        getIPFamilyPolicy(service.Spec.IPFamilyPolicy),
+		IsStaticIPv4:          service.Spec.LoadBalancerIP != "",
```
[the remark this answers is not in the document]
mmamczur: ok both will now check if the provided val is a valid IP (either v4 of v6)

swetharepakula (commented):
this mostly looks good to me. I would recommend customizing the buckets in prometheus if the goal is to reduce the cardinality and remove some of the logic necessary to maintain the buckets.

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,369 @@
...
+	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/common"
+	"k8s.io/klog/v2"
+	"k8s.io/legacy-cloud-providers/gce"
+)
+
+const (
+	// Names of the labels used by the service metrics.
+	labelType                  = "type"
```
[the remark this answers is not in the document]
mmamczur: in general this is the service type, so could be ClusterIP, NodePort etc with a caveat that instead of just having LoadBalancer we have special values for each varian of them, like LegacyXLB, RBSXLB etc
we considered having separate label but we wanted to limit the number of possible label value combinatons

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,369 @@
...
+	labelExternalTrafficPolicy = "external_traffic_policy"
+	labelInternalTrafficPolicy = "internal_traffic_policy"
+	labelSessionAffinityConfig = "session_affinity_config"
+	labelProtocol              = "protocol"
+	labelIPFamilies            = "ip_families"
+	labelIPFamilyPolicy        = "ip_family_policy"
+	labelIsStaticIPv4          = "is_static_ip_v4"
+	labelIsStaticIPv6          = "is_static_ip_v6"
```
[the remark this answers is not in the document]
mmamczur: LB that would have 2 static IPs?
is this possible? We could also set both these to true if we ever support a case like that

pkg/servicemetrics/servicemetrics.go
```diff
@@ -0,0 +1,369 @@
...
+	sessionAffinityBucketMoreThanDefault = "10800+"
+	sessionAffinityBucketDefault         = "10800"
+	sessionAffinityBucketLessThanDefault = "0-10799"
+	sessionAffinityBucketNone            = "None"
+)
+
+var (
+	serviceL4ProtocolStatsCount = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "service_l4_protocol_stats",
```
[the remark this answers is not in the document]
mmamczur: something like service_l4_config_stats?

mmamczur:
> this mostly looks good to me. I would recommend customizing the buckets in prometheus if the goal is to reduce the cardinality and remove some of the logic necessary to maintain the buckets.

what do you mean about the buckets, can you have them for the label values (and not the metric value)?

PR #1944, opened 2023-02-11
Add new metric status PersistentError for L4 DualStack

panslava:
Report PersistentError if service is in Error after 20 minutes of
retries

Should be merged after NetLB Metrics https://github.com/kubernetes/ingress-gce/pull/1937

/assign cezarygerard

### DIFF
--- a/pkg/metrics/types.go
+++ b/pkg/metrics/types.go
@@ -78,10 +78,11 @@ type L4DualStackServiceStatus string
 const StatusSuccess = L4DualStackServiceStatus("Success")
 const StatusUserError = L4DualStackServiceStatus("UserError")
 const StatusError = L4DualStackServiceStatus("Error")
+const StatusPersistentError = L4DualStackServiceStatus("PersistentError")
 
-// L4DualStackServiceState defines ipFamilies, ipFamilyPolicy and status
-// of L4 ILB DualStack service
-type L4DualStackServiceState struct {
+// L4DualStackServiceLabels defines ipFamilies, ipFamilyPolicy and status
+// of L4 DualStack service
+type L4DualStackServiceLabels struct {
 	// IPFamilies stores spec.ipFamilies of Service
 	IPFamilies string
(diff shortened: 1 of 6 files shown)


pkg/metrics/types.go
```diff
@@ -78,20 +78,29 @@ type L4DualStackServiceStatus string
...
-type L4DualStackServiceState struct {
+// L4DualStackServiceLabels defines ipFamilies, ipFamilyPolicy and status
+// of L4 DualStack service
+type L4DualStackServiceLabels struct {
 	// IPFamilies stores spec.ipFamilies of Service
 	IPFamilies string
 	// IPFamilyPolicy specifies spec.IPFamilyPolicy of Service
 	IPFamilyPolicy string
-	// Status specifies status of L4 DualStack Service
+	// Status specifies status of L4 ILB DualStack
```
cezarygerard, line 90: ILB?

panslava: fixed

pkg/metrics/types.go
```diff
@@ -90,8 +91,16 @@ type L4DualStackServiceState struct {
 	Status L4DualStackServiceStatus
 }
 
+// L4DualStackServiceState defines ipFamilies, ipFamilyPolicy, status and tracks
+// FirstSyncErrorTime of L4 DualStack service
+type L4DualStackServiceState struct {
+	L4DualStackServiceLabels
+	// FirstSyncErrorTime specifies the time timestamp when the service sync ended up with error for the first time.
+	FirstSyncErrorTime *time.Time
```
cezarygerard, line 99: FirstSyncTime?
you always set it, not only when error happens

panslava: It is funny logic we have, but on every sync we initially set the service to Error state, and only on successfuly sync update it to success

That's why on every beggining on sync we also set FirstSyncErrorTime -- considering that service is in Error state

So I think we should keep the name

PR #1945, opened 2023-02-11
Add L4 DualStack Sync Latency metrics

panslava:
Add a new metric, similar to L4 ILB, NetLB Latency, which will report latencies if L4 DualStack enabled, with ipfamilies as a new label

Should be merged after https://github.com/kubernetes/ingress-gce/pull/1937

### DIFF
--- a/pkg/l4lb/metrics/metrics.go
+++ b/pkg/l4lb/metrics/metrics.go
@@ -27,8 +27,10 @@ const (
 	statusSuccess                                  = "success"
 	statusError                                    = "error"
 	L4ilbLatencyMetricName                         = "l4_ilb_sync_duration_seconds"
+	L4ILBDualStackLatencyMetricName                = "l4_ilb_dualstack_sync_duration_seconds"
 	L4ilbErrorMetricName                           = "l4_ilb_sync_error_count"
 	L4netlbLatencyMetricName                       = "l4_netlb_sync_duration_seconds"
+	L4NetLBDualStackLatencyMetricName              = "l4_netlb_dualstack_sync_duration_seconds"
 	L4netlbErrorMetricName                         = "l4_netlb_sync_error_count"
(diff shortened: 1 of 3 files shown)


pkg/l4lb/metrics/metrics.go
```diff
@@ -56,14 +59,24 @@ var (
 		},
 		l4LBSyncLatencyMetricsLabels,
 	)
+	l4ILBDualStackSyncLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: L4ILBDualStackLatencyMetricName,
+			Help: "Latency of an L4 ILB DualStack Sync",
```
cezarygerard, line 65: maybe we can just use 
0.5,1,2,4,8 and so on?

the funny 0.9375s comes from extension of old buckets

pkg/l4lb/metrics/metrics.go
```diff
@@ -75,6 +88,16 @@ var (
 		},
 		l4LBSyncLatencyMetricsLabels,
 	)
+	l4NetLBDualStackSyncLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: L4NetLBDualStackLatencyMetricName,
+			Help: "Latency of an L4 NetB DualStack Sync",
+			// custom buckets - [0.9375s, 1.875s, 3.75s, 7.5s, 30s, 60s, 120s, 240s(4min), 480s(8min), 960s(16m), 3840s(64min), 7680s(128m) +Inf]
```
cezarygerard, line 95: same, use 0.5, 1, 2...

pkg/l4lb/metrics/metrics.go
panslava: Agree, fixed

pkg/l4lb/metrics/metrics.go
panslava: Agree, fixed

PR #1946, opened 2023-02-11
configmapcontroller use the informer cache

aojea:
Instead of doing a Get to the apiserver use the informer cache


swetharepakula (commented):
/lgtm
/approve

PR #1947, opened 2023-02-11
Add health status metric for L4 ILB DualStack Controller

panslava:


pkg/l4lb/l4controller.go
```diff
@@ -155,6 +156,9 @@ func (l4c *L4Controller) checkHealth() error {
 		// Log here, context/http handler do no log the error.
 		klog.Error(msg)
 		l4metrics.PublishL4FailedHealthCheckCount(l4ILBControllerName)
+		if l4c.enableDualStack {
+			l4metrics.PublishL4FailedHealthCheckCount(l4ILBDualStackControllerName)
```
aojea, line 160: where is this metric recorded? is not the same controller always?

panslava: This is reported under same metric for L4 ILB , L4 NetLB currently (which maybe strange a bit) called l4_failed_healthcheck_count

https://github.com/kubernetes/ingress-gce/blob/4c98f1568ff74ff310c96b138371c49cf3ee637c/pkg/l4lb/metrics/metrics.go#L87

And the metric has label "controller_name", so for DualStack it will report to the same metric with new label value controller_name = l4-ilb-dualstack-controller

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -354,11 +357,16 @@ func (lc *L4NetLBController) checkHealth() error {
 	// if lastEnqueue time is more than 15 minutes before the last sync time, the controller is falling behind.
 	// This indicates that the controller was stuck handling a previous update, or sync function did not get invoked.
 	syncTimeLatest := lastEnqueueTime.Add(enqueueToSyncDelayThreshold)
```
panslava, line 359: This is funny, how we reference const from the other file :))

PR #1950, opened 2023-02-14
Refactor procedure to compute endpoint target map

sawsa307:
Refactor the code that compute endpoint target map into a separate function computeTargetMap.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -186,17 +186,14 @@ func (s *transactionSyncer) syncInternal() error {
 	defer s.syncLock.Unlock()
 
 	start := time.Now()
-	result := s.syncInternalImpl()
+	err := s.syncInternalImpl()
 
-	s.updateStatus(result.Error)
-	metrics.PublishNegSyncMetrics(string(s.NegSyncerKey.NegType), string(s.endpointsCalculator.Mode()), result.Error, start)
-	if result.Error != nil {
-		s.syncCollector.UpdateSyncer(s.NegSyncerKey, result)
-	}
-	return result.Error
+	s.updateStatus(err)
+	metrics.PublishNegSyncMetrics(string(s.NegSyncerKey.NegType), string(s.endpointsCalculator.Mode()), err, start)
+	return err
 }
 
-func (s *transactionSyncer) syncInternalImpl() *negtypes.NegSyncResult {
+func (s *transactionSyncer) syncInternalImpl() error {
 	// TODO(cheungdavid): for now we reset the boolean so it is a no-op, but
 	// in the future, it will be used to trigger degraded mode if the syncer is in error state.
 	if s.inErrorState() {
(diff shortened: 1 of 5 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -248,13 +241,13 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+		if s.inErrorState() {
+			s.resetErrorState()
+		} else {
+			targetMap, endpointPodMap, err = s.computeTargetMap(endpointSlices)
+			if err != nil {
+				return err
+			}
 		}
 	} else {
 		ep, exists, err := s.endpointLister.Get(
```
aojea, line 253: Interesting, since EndpointSlices is GA since 1.21, maybe we can start removing all the Endpoints code

sawsa307: Hi Antontio. Regarding this, I asked Swetha about this and this change is already on the calendar, and it will be in a separate PR. For now, we refactor this part of code since in degraded mode, we will do this calculation differently.

pkg/neg/syncers/transaction.go
```diff
@@ -265,16 +258,13 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
-		}
-		if err != nil {
-			return fmt.Errorf("endpoints calculation error in mode %q, err: %w", s.endpointsCalculator.Mode(), err)
+		if s.inErrorState() {
+			s.resetErrorState()
+		} else {
+			targetMap, endpointPodMap, err = s.computeTargetMap(endpointSlices)
+			if err != nil {
+				return err
+			}
```
swetharepakula, line 267: Why is the computation in the else block? Don't we always need to compute the target map?

pkg/neg/syncers/transaction.go
swetharepakula: Made https://github.com/kubernetes/ingress-gce/pull/1973 to remove endpoints codepath

pkg/neg/syncers/transaction.go
```diff
@@ -268,18 +262,13 @@ func (s *transactionSyncer) syncInternalImpl() *negtypes.NegSyncResult {
...
-			s.setErrorState(result.Result)
-			return result
-		}
-		if valid, result := s.CheckEndpointInfo(endpointsData, endpointPodMap, dupCount); !valid {
-			s.setErrorState(result.Result)
-			return result
-		}
-		if err != nil {
-			return negtypes.NewNegSyncResult(fmt.Errorf("endpoints calculation error in mode %q, err: %w", s.endpointsCalculator.Mode(), err), negtypes.ResultOtherError)
+		if s.inErrorState() {
```
swetharepakula, line 265: this means we will not compute the target map if we are in error state? What do we do in that case?

pkg/neg/syncers/transaction.go
sawsa307: Remove the condition. Now we only do computeTargetMap.

sawsa307:
Closed due to change of implementation

PR #1956, opened 2023-02-15
Endpoints Calculator should check the endpoint is ready

aojea:
The fact that the endpointslice Conditions.Ready and Terminating
prevent that Unready endpoints are used, but this is a bit of a
fortunate coincidence, the endpoints calculator should verify
the endpoint is serving traffic.

You can run the test that is in a different commit to verify it fails without the fix

/assign @swetharepakula

### DIFF
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -77,6 +77,10 @@ func (l *LocalL4ILBEndpointsCalculator) CalculateEndpoints(eds []types.Endpoints
 				l.logger.V(2).Info("Address inside Endpoints does not have an associated pod. Skipping", "address", addr.Addresses, "endpoints", klog.KRef(ed.Meta.Namespace, ed.Meta.Name))
 				continue
 			}
+			if !addr.Ready {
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -77,6 +77,10 @@ func (l *LocalL4ILBEndpointsCalculator) CalculateEndpoints(eds []types.Endpoints
 				l.logger.V(2).Info("Address inside Endpoints does not have an associated pod. Skipping", "address", addr.Addresses, "endpoints", klog.KRef(ed.Meta.Namespace, ed.Meta.Name))
 				continue
 			}
+			if !addr.Ready {
+				l.logger.V(2).Info("Address inside Endpoints is not ready. Skipping", "address", addr.Addresses, "endpoints", klog.KRef(ed.Meta.Namespace, ed.Meta.Name))
+				continue
+			}
```
swetharepakula, line 83: For ILB case, since we use the healthchecknodeport to decide which nodes to route traffic, isn't it okay to ignore endpoint readiness?

The healthcheck should fail for this node if this is the only endpoint and it is not ready so load balancer will not route traffic until the endpoint is ready.

The benefit I see in not adding the readiness check is that we add the node sooner and can reduce the latency of attaching the endpoint if this is a new node.

aojea:
I think that Swetha already fixed this in other pr

PR #1957, opened 2023-02-15
don't consider ANY pod that is not terminating

aojea:
The neg syncer was doing some wrong assumptions:

1. Endpoints that are not Ready can have Pods with deletionTimestamp not nil.
The Endpoint and EndpointSlice controller will remove the Pod from the corresponding object once it gets deleted (DeletionTimestamp != nil)

2. Any pod with deletionTimestamp == nil is valid, this is the most dangerous assumption, since a pod can be Unready and be used.

/kind bug
/assign @swetharepakula @bowei

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	apiv1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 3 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -271,12 +270,12 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
...
-			// TODO: This check and EndpointsData.Ready field may be deleted once Endpoints support is removed.
-			// The purpose of this check is to handle endpoints in terminating state.
-			// The Endpoints API doesn't have terminating field. Terminating endpoints are marked as not ready.
-			// This check support this case. For not ready endpoints it checks if the endpoint is not yet ready or terminating.
-			// The EndpointSlices API has terminating field which solves this problem.
-			if endpointAddress.Ready || shouldPodBeInNeg(podLister, endpointAddress.TargetRef.Namespace, endpointAddress.TargetRef.Name) {
+			// Only consider endpoints that are Ready.
+			// TODO: In order to consider Terminating endpoints, the AddressData type should add a new field to
+			// carry this information based on the EndpointSlices Conditions Terminating = true and Serving = true.
+			// Terminating endpoints should only be used with ExternalTrafficPolicy = Local and if there are any
```
swetharepakula, line 276: In the L7 case we don't care about Local or Cluster

swetharepakula (commented):
I have one nit on the comment because I don't think it completely applies to this part of the code.

The only issue with this is that Endpoints do not have the terminating status on the endpoint itself, which is why this logic was added. I will be creating a PR to remove the entire endpoints support, which would remove this code as well. However until we do that, I do not think we should remove this due to correctness of the Endpoints path. I will try to have that PR ready by the end of the week, so we can merge this one in too.

pkg/neg/syncers/utils.go
aojea: I tried to complete the existing one, I don't feel it fits too

aojea:
> The only issue with this is that Endpoints do not have the terminating status on the endpoint itself, which is why this logic was added.

this logic is wrong, or I'm reading it wrong, the terminating status means pod.deletionTimestsamp != nil, and this logic does not consider those pods.

In addition, it doesn't check the phase of the pods, that means that can be adding notReady or failing pods

aojea:
> The only issue with this is that Endpoints do not have the terminating status on the endpoint itself, which is why this logic was added.

This is a new EndpointData that Alan added to consolidate slices and Endpoints, we can add Terminating status there if we need it, but besides that, this is not handling terminating ... but I'd love if you can double and triple check my assumptions

aojea:
/hold

talked offline, @swetharepakula has to land different changes around this code and she'll fix this during that refactor

swetharepakula:
This has been added as part of #1973.

PR #1953, opened 2023-02-15
Fill missing information in degraded mode

sawsa307:
Create endpoints and fill missing information during degraded mode. Add endpoints to the desire map and endpointPodMap.

PR #1952, opened 2023-02-14
Create skeleton code for degraded mode procedure

sawsa307:
Create skeleton code for degraded mode procedure. Implementation for filling missing information and filter invalid pod will come in separate PRs.

PR #1958, opened 2023-02-17
Filter terminal pods in degraded mode

sawsa307:
Filter terminal pods, which are pods with failed or succeeded phase.

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -328,6 +328,10 @@ func validateAndAddEndpoints(ep negtypes.AddressData, zoneGetter negtypes.ZoneGe
 		}
 		pod, ok := obj.(*apiv1.Pod)
 		if !ok {
+			klog.V(2).Infof("Endpoint %q does not correspond to a pod object. Skipping", address)
+			continue
+		}
+		if !validatePod(pod) {
 			klog.V(2).Infof("Endpoint %q does not correspond to a valid pod resource. Skipping", address)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -325,6 +325,18 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	return s.syncNetworkEndpoints(addEndpoints, removeEndpoints)
 }
 
+// isValidPod checks if this pod is a valid pod resource
+// it returns false if the pod:
+// 1. doesn't exist
+// 2. is in terminal state
+func (s *transactionSyncer) isValidPod(pod *apiv1.Pod) bool {
```
swetharepakula, line 332: will there be a followup PR to use this method?

sawsa307: Yes it will be used in another PR together with #1953.

pkg/neg/syncers/transaction.go
```diff
@@ -325,6 +325,18 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+	// Terminal Pod means a pod is in PodFailed or PodSucceeded phase
+	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded {
+		return false
+	}
+	return true
```
swetharepakula, line 337: should we also verify that the endpoint is the one we expected or will that occur in a different check?

sawsa307: This will only check if the pod resource is valid. Check for endpoints will occur in #1952.

pkg/neg/syncers/utils.go
```diff
@@ -356,6 +360,19 @@ func validateAndAddEndpoints(ep negtypes.AddressData, zoneGetter negtypes.ZoneGe
...
+// validatePod checks if this pod is a valid pod resource
+// it returns false if the pod:
+// 1. is in terminal state
+func validatePod(pod *apiv1.Pod) bool {
+	// Terminal Pod means a pod is in PodFailed or PodSucceeded phase
+	phase := pod.Status.Phase
+	if phase == apiv1.PodFailed || phase == apiv1.PodSucceeded {
+		klog.V(2).Info("Pod %s/%s is a terminal pod with status %v, skipping", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name, phase)
+		return false
+	}
```
swetharepakula, line 372: should we check that the ip matches?

swetharepakula: also where will the pod labels be verified?

sawsa307: Checking if IP matches is included in the PR that does pod ip and node's podCIDR check, #1963.
Pod labels verification is included in #1966.
This PR is solely filtering out pods that are in terminal state.

swetharepakula (commented):
/lgtm
/approve

PR #1959, opened 2023-02-17
Fix waitGroup race condition bug

sawsa307:
Initialize one waitGroup for each sync now to avoid race condition due to multiple syncs all work on the same waitGroup.


swetharepakula (commented):
/lgtm
/approve
/ok-to-test

PR #1962, opened 2023-02-20
Filter pods from non-existent node in degraded mode

sawsa307:
Filter pods from non-existent node in degraded mode


swetharepakula (commented):
/lgtm
/approve

PR #1963, opened 2023-02-21
Filter pods that have out of range IP

sawsa307:
Filter pods have IPs outside of the corresponding nodes' IP ranges.

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -371,15 +371,11 @@ func getEndpointPod(
 
 // toZoneNetworkEndpointMap translates addresses in endpoints object into zone and endpoints map, and also return the count for duplicated endpoints
 // we will not raise error in degraded mode for misconfigured endpoints, instead they will be filtered directly
-func toZoneNetworkEndpointMapDegradedMode(
-	eds []negtypes.EndpointsData,
-	zoneGetter negtypes.ZoneGetter,
-	podLister, nodeLister cache.Indexer,
-	servicePortName string,
-	networkEndpointType negtypes.NetworkEndpointType,
-) ZoneNetworkEndpointMapResult {
+func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, podLister, nodeLister cache.Indexer, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, enableDualStackNEG bool) ZoneNetworkEndpointMapResult {
 	zoneNetworkEndpointMap := map[string]negtypes.NetworkEndpointSet{}
 	networkEndpointPodMap := negtypes.EndpointPodMap{}
+	dupCount := 0
+	ipsForPod := ipsForPod(eds)
 	for _, ed := range eds {
 		matchPort := ""
 		for _, port := range ed.Ports {
@@ -392,7 +388,7 @@ func toZoneNetworkEndpointMapDegradedMode(
 			continue
 		}
 		for _, endpointAddress := range ed.Addresses {
-			if endpointAddress.AddressType != discovery.AddressTypeIPv4 {
+			if !enableDualStackNEG && endpointAddress.AddressType != discovery.AddressTypeIPv4 {
 				klog.Infof("Skipping non IPv4 address in degraded mode: %q, in endpoint slice %s/%s", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
 				continue
 			}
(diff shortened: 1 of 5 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -325,6 +327,33 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+	}
+	obj, exists, err := s.nodeLister.GetByKey(pod.Spec.NodeName)
+	if err != nil || !exists {
+		return false
+	}
+	node, isNode := obj.(*apiv1.Node)
+	if !isNode {
+		return false
+	}
+	_, podCIDR, err := netset.ParseCIDRSloppy(node.Spec.PodCIDR)
```
aojea, line 349: you can omit the custom parsers here and use the std library ones, net.ParseCIDR()

pkg/neg/syncers/transaction.go
```diff
@@ -325,6 +327,33 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+	podIP := net.ParseIP(pod.Status.PodIP)
```
aojea, line 350: the CNI not necessarily have to use `node.Spec.PodCIDR` , that means that this relation may not be always true, also, if the CNI is supposed to use the CIDR this double check should not be necessary, because this can only happen if the CNI is wrong.

In addition, there is a plan to support multiple PodCIDRs, so node.Spec.PodCIDRs will contain a slice of networks.

sawsa307: The checks is related to the problem of preemptible node with the buggy 1.22 node eviction logic. We are verify this because it might cause the NEG not to be sync completely. We are operation it on the assumption that there might be a bug somewhere else. 
Currently since we are only supporting single network ipv4, we only do check for ipv4. As we add support for the other feature, this will be updated. 
@swetharepakula

aojea: then it is better if you iterate over the node.Spec.PodCIDRs list too , and returns true if there is a match, and false if there is no match

aojea: I commented in the other PR you sent with a sketch of a solution

pkg/neg/syncers/transaction.go
sawsa307: Hi Antonio, i found in the comment section for ParseCIDRSloppy, it says `We're choosing to keep it for compat with potential stored values.`, where the values refer to CIDR with leading '0' characters on numbers. Is there a specific reason that we can be sure that there is no need to consider CIDR like this?

aojea: yeah, that is a good question, we had to do it in kubernetes to keep compatibility, imagine someone created a service with the ip 10.001.1.1, if he upgrades that Service will be invalid, and we don't want to break it.

For components that use the kubernetes API they have more freedom to make their choice and say, since version X if we found that IP we are going to reject it, so please update your manifests.

I prefer for the ecosystem to "force" this evolution and not keep the old parsers, it should be a problem only for Kubernetes/kubernetes not for the others.

pkg/neg/syncers/utils_test.go
```diff
@@ -1522,85 +1586,35 @@ func TestValidatePod(t *testing.T) {
...
+func addPodsToLister(podLister cache.Indexer, endpointSlices []*discovery.EndpointSlice) {
+	for _, eps := range endpointSlices {
+		addressType := eps.AddressType
+		for _, ep := range eps.Endpoints {
+			podNamespace := ep.TargetRef.Namespace
+			podName := ep.TargetRef.Name
+			for _, addr := range ep.Addresses {
+				addresses := []v1.PodIP{{IP: addr}}
+				if addressType == discovery.AddressTypeIPv4 {
+					addresses = append(addresses, v1.PodIP{IP: "2001:db8::68"})
```
sawsa307, line 1598: add ipv6

swetharepakula (commented):
/lgtm
/approve

swetharepakula (commented):
/hold

pkg/neg/syncers/utils.go
```diff
@@ -562,6 +572,41 @@ func ipsForPod(eds []negtypes.EndpointsData) map[types.NamespacedName]negtypes.N
...
+	for _, podCIDR := range node.Spec.PodCIDRs {
+		podCIDR = strings.TrimSpace(podCIDR)
+		_, ipnet, err := net.ParseCIDR(podCIDR)
+		if err != nil {
+			// swallow errors for CIDRs that are invalid
+			continue
+		}
+		ipnets = append(ipnets, ipnet)
+	}
+	podIP := net.ParseIP(pod.Status.PodIP)
```
swetharepakula, line 601: this only checks for the ipv4 address. we need to verify the ipv6 address as well

pkg/neg/syncers/utils_test.go
```diff
@@ -2122,85 +2244,38 @@ func TestParseIPAddress(t *testing.T) {
...
-				Name:      fmt.Sprintf("pod%v", i),
-			},
-			Status: corev1.PodStatus{
-				Phase: v1.PodRunning,
-			},
-			Spec: corev1.PodSpec{
-				NodeName: testInstance1,
-			},
-		})
+func addPodsToLister(podLister cache.Indexer, endpointSlices []*discovery.EndpointSlice) {
```
swetharepakula, line 2247: add a comment about what this function is doing and how it uses the EPS

swetharepakula (commented):
typo in the first commit for dual stack

For the second commit, we should check all the IPs we are interested in, not just what is in PodIP

swetharepakula:
/hold cancel

initial to stop the lgtm. I initially had only looked at the first commit

pkg/neg/syncers/utils.go
```diff
@@ -560,6 +571,70 @@ func ipsForPod(eds []negtypes.EndpointsData) map[types.NamespacedName]negtypes.N
 	return result
 }
 
+// podContainsEndpointAddress checks the pod's existing PodIP(s)
+// and return error if the given endpoint's IP address does not any of them.
+// If dual stack is enabled, it would return error if either ipv4 or ipv6 address does not match to the pod IPs
+func podContainsEndpointAddress(networkEndpoint negtypes.NetworkEndpoint, pod *apiv1.Pod, enableDualStackNEG bool) error {
+	endpointIPs := []string{networkEndpoint.IP}
+	if enableDualStackNEG {
```
swetharepakula, line 579: check that there is a dual stack endpoint. It is possible dual stack is enabled but this service is not a dual stack service.

pkg/neg/syncers/utils.go
```diff
@@ -560,6 +571,70 @@ func ipsForPod(eds []negtypes.EndpointsData) map[types.NamespacedName]negtypes.N
...
+			continue
+		}
+		ipnets = append(ipnets, ipnet)
+	}
+	podIPs := []net.IP{}
+	if enableDualStackNEG {
+		// get the pod's ipv4 and ipv6 address and convert them from type apiv1.PodIP to net.IP
+		for _, podIP := range pod.Status.PodIPs {
+			podIPs = append(podIPs, net.ParseIP(podIP.IP))
+		}
```
swetharepakula, line 618: since the pod ips have already been verified, use the network endpoint we formed instead. This might be a dual stack cluster, but it might not be a dual stack service so both endpoints may not need to be verified.

swetharepakula (commented):
The pod IP validation needs to consider a few more edge cases

pkg/neg/syncers/utils.go
```diff
@@ -493,6 +502,65 @@ func ipsForPod(eds []negtypes.EndpointsData) map[types.NamespacedName]negtypes.N
 	return result
 }
 
+// podContainsEndpointAddress checks the pod's existing PodIP(s)
+// and return error if the given endpoint's IP address does not any of them.
+// If this is a dual stack endpoint, we would validate both IPs
+func podContainsEndpointAddress(networkEndpoint negtypes.NetworkEndpoint, pod *apiv1.Pod) error {
+	endpointIPs := []string{networkEndpoint.IP}
```
swetharepakula, line 509: add todo to remove when single stack ipv6 is supported

pkg/neg/syncers/utils_test.go
```diff
@@ -1782,6 +1792,57 @@ func TestDegradedModeValidateEndpointInfo(t *testing.T) {
 			expectedEndpointMap: endpointMap,
 			expectedPodMap:      podMap,
 		},
+		{
+			desc: "endpoint with IP not matching to its pod, endpoint should be removed",
```
swetharepakula, line 1796: add one for the ipv6 address as well

swetharepakula (commented):
few small comments, otherwise this is ready to merge

swetharepakula (commented):
/lgtm
/approve

PR #1964, opened 2023-02-22
don't process unsupported LB Services with multiple protocoles

aojea:
Since kubernetes 1.24 Kubernetes allot to configure LoadBalancer Services with multiplt protocols, typically UDP and TCP.

The ingress-gce L4 controllers assume that there is only one protocol, and is defined by the first port in the ServicePorts array.

We should not process these services because they work only for one protocol, causing confusion to users since they can see that it partially works, and is hard to find out what is the problem

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -192,6 +192,10 @@ func (l4c *L4Controller) shouldProcessService(service *v1.Service) bool {
 		klog.Warningf("Ignoring update for service %s:%s managed by service controller", service.Namespace, service.Name)
 		return false
 	}
+	if utils.HasMultipleProtocols(service.Spec.Ports) {
+		klog.Warningf("Ignoring update for service %s:%s using different ports protocols", service.Namespace, service.Name)
+		return false
+	}
 	frName := utils.LegacyForwardingRuleName(service)
 	// Processing should continue if an external forwarding rule exists. This can happen if the service is transitioning from External to Internal.
 	// The external forwarding rule might not be deleted by the time this controller starts processing the service.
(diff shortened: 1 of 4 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -192,6 +192,10 @@ func (l4c *L4Controller) shouldProcessService(service *v1.Service) bool {
 		klog.Warningf("Ignoring update for service %s:%s managed by service controller", service.Namespace, service.Name)
 		return false
 	}
+	if utils.HasMultipleProtocols(service.Spec.Ports) {
+		klog.Warningf("Ignoring update for service %s:%s using different ports protocols", service.Namespace, service.Name)
```
mmamczur, line 196: should we mention that it's not supported for GKE currently?
Probably if I saw "Ignoring update for service default:myservice using different ports protocols" I'd think "but why ignoring?"

panslava: We should emit customer facing event, logs are internal only

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -232,6 +232,10 @@ func (lc *L4NetLBController) shouldProcessService(newSvc, oldSvc *v1.Service) bo
 	if !(lc.isRBSBasedService(newSvc) || lc.isRBSBasedService(oldSvc)) {
 		return false
 	}
+	if utils.HasMultipleProtocols(newSvc.Spec.Ports) {
```
panslava, line 235: Correct me if I am wrong -- but this will prevent service deletion?

aojea:
@cezarygerard please don't forget about this, I'm going to close it and let you handle it

PR #1966, opened 2023-02-23
Filter pods that don't belong to the service in question

sawsa307:
Filter pods that don't have labels match to its service label selector.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -180,7 +180,7 @@ func NewTransactionSyncer(
 	return syncer
 }
 
-func GetEndpointsCalculator(nodeLister, podLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger, enableDualStackNEG bool) negtypes.NetworkEndpointsCalculator {
+func GetEndpointsCalculator(podLister, nodeLister, serviceLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger, enableDualStackNEG bool) negtypes.NetworkEndpointsCalculator {
 	serviceKey := strings.Join([]string{syncerKey.Name, syncerKey.Namespace}, "/")
 	if syncerKey.NegType == negtypes.VmIpEndpointType {
 		nodeLister := listers.NewNodeLister(nodeLister)
@@ -195,6 +195,7 @@ func GetEndpointsCalculator(nodeLister, podLister cache.Indexer, zoneGetter negt
 		zoneGetter,
 		podLister,
 		nodeLister,
+		serviceLister,
 		syncerKey.PortTuple.Name,
 		syncerKey.NegType,
 		logger,
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -48,6 +48,10 @@ const (
 	minRetryDelay = 5 * time.Second
 	maxRetryDelay = 600 * time.Second
 	separator     = "||"
+
+	// managedByEPSControllerValue is a unique value used with LabelManagedBy to indicate
+	// the EndpointSlice is managed by the endpoint slice controller.
+	managedByEPSControllerValue = "endpointslice-controller.k8s.io"
 )
 
 // encodeEndpoint encodes ip and instance into a single string
@@ -371,7 +375,7 @@ func getEndpointPod(
 
 // toZoneNetworkEndpointMap translates addresses in endpoints object into zone and endpoints map, and also return the count for duplicated endpoints
 // we will not raise error in degraded mode for misconfigured endpoints, instead they will be filtered directly
-func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, podLister, nodeLister cache.Indexer, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, enableDualStackNEG bool) ZoneNetworkEndpointMapResult {
+func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, podLister, nodeLister, serviceLister cache.Indexer, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, enableDualStackNEG bool) ZoneNetworkEndpointMapResult {
 	zoneNetworkEndpointMap := map[string]negtypes.NetworkEndpointSet{}
 	networkEndpointPodMap := negtypes.EndpointPodMap{}
 	dupCount := 0
@@ -387,6 +391,8 @@ func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGett
 		if len(matchPort) == 0 {
 			continue
 		}
+		serviceName := ed.Meta.Labels[discovery.LabelServiceName]
+		isCustomEPS := ed.Meta.Labels[discovery.LabelManagedBy] != managedByEPSControllerValue
 		for _, endpointAddress := range ed.Addresses {
 			if !enableDualStackNEG && endpointAddress.AddressType != discovery.AddressTypeIPv4 {
 				klog.Infof("Skipping non IPv4 address in degraded mode: %q, in endpoint slice %s/%s", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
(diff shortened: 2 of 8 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -339,10 +342,15 @@ func (s *transactionSyncer) isValidPod(pod *apiv1.Pod) bool {
 	if err != nil || !exists {
 		return false
 	}
-	_, isNode := obj.(*apiv1.Node)
+	node, isNode := obj.(*apiv1.Node)
 	if !isNode {
 		return false
 	}
+	_, podCIDR, err := netset.ParseCIDRSloppy(node.Spec.PodCIDR)
```
aojea, line 350: you don't need to use the custom parsers, and is more complete if you iterate over all the PodCIDRs, you can add a helper for that

```
func nodeContainsPodIP(node *v1.Node, pod *v1.Pod) bool {
     if node == nil || pod == nil {
        return false
      }
         ipnets := []*net.IPNet{}
 	for _, podCIDR := range node.Spec.PodCIDRs {
		podCIDR = strings.TrimSpace(podCIDR)
		_, ipnet, err := net.ParseCIDR(podCIDR)
		if err != nil {
                         // swallow errors for CIDRs that are invalid
			continue
		}
		ipnets = append(ipnets, ipnet)
	}
	_, ipnet, err := net.ParseCIDR(node.Spec.PodCIDR)
	if err == nil {
              // swallow errors for CIDRs that are invalid
	     ipnets = append(ipnets, ipnet)
        }
         podIP := net.ParseIP(pod.Status.PodIP)
	for _, net := range ipnets {
           if net.Contains(podIP) {
                return true
                 }
      }
       return false
  }

File: pkg/neg/syncers/transaction.go [hunk 2]
```diff
@@ -351,6 +352,21 @@ func (s *transactionSyncer) isValidPod(pod *apiv1.Pod) bool {
 	if !podCIDR.Contains(podIP) {
 		return false
 	}
+	podLabels := pod.ObjectMeta.Labels
```
aojea, line 355: hmm, how can we end in this situation?

pkg/neg/syncers/transaction.go
sawsa307: Yes I saw your comment in the other PR #1963, and I was about to address it there, but thank you for writing this out!

pkg/neg/syncers/transaction.go
sawsa307: Hi Antonio, could you elaborate on which situation you are referring to?

pkg/neg/syncers/transaction.go
sawsa307: since node.Spec.PodCIDR is guaranteed to be the zeroth entry in node.Spec.PodCIDRs, I think we can skip the check outside the loop

aojea: > since node.Spec.PodCIDR is guaranteed to be the zeroth entry in node.Spec.PodCIDRs

good catch

pkg/neg/syncers/utils.go
```diff
@@ -395,6 +399,16 @@ func validatePod(pod *apiv1.Pod, nodeLister cache.Indexer) bool {
 		klog.V(2).Info("Pod %s/%s has an IP %v that is outside of the node's allocated IP range(s) %v, skipping", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name, pod.Status.PodIP, node.Spec.PodCIDRs)
 		return false
 	}
+	service := getService(serviceLister, pod.ObjectMeta.Namespace, serviceName)
+	if service == nil {
+		klog.V(2).Info("Endpoint does not correspond to an existing service %s, skipping", serviceName)
+		return false
+	}
+	// for custom endpoint slice, we won't check the pod's labels
+	if !isCustomEPS && !podBelongsToService(pod, service) {
```
sawsa307, line 408: continue if custom eps

pkg/neg/syncers/transaction.go
```diff
@@ -180,7 +180,7 @@ func NewTransactionSyncer(
 	return syncer
 }
 
-func GetEndpointsCalculator(nodeLister, podLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger, enableDualStackNEG bool) negtypes.NetworkEndpointsCalculator {
+func GetEndpointsCalculator(podLister, nodeLister, serviceLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger, enableDualStackNEG bool) negtypes.NetworkEndpointsCalculator {
```
swetharepakula, line 183: why are you moving the params around?

pkg/neg/syncers/utils.go
```diff
@@ -387,6 +387,8 @@ func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGett
 		if len(matchPort) == 0 {
 			continue
 		}
+		serviceName := ed.Meta.Labels[discovery.LabelServiceName]
+		isCustomEPS := ed.Meta.Labels[discovery.LabelManagedBy] != "endpointslice-controller.k8s.io"
```
swetharepakula, line 391: is there a constant we can use? otherwise define this as a constant

pkg/neg/syncers/utils.go
```diff
@@ -563,6 +576,19 @@ func nodeContainsPodIP(node *apiv1.Node, networkEndpoint negtypes.NetworkEndpoin
...
 }
 
+// podBelongsToService checks the pod's labels
+// and return error if any label specified in the service's label selector is not in the pod's labels
+func podBelongsToService(pod *apiv1.Pod, service *apiv1.Service) error {
+	podLabels := pod.ObjectMeta.Labels
+	serviceLabels := service.Spec.Selector
+	for key, val1 := range serviceLabels {
+		if val2, contains := podLabels[key]; !contains || val1 != val2 {
+			return fmt.Errorf("%w: pod %s has labels not match to its service %s's label selector", negtypes.ErrEPPodLabelMismatch, pod.Name, service.Name)
```
swetharepakula, line 586: Perhaps this as the msg instead: "%w: pod %s/%s has labels that do not match the service %s/%s's label selector"

Include the namespace for both the pod and service

pkg/neg/syncers/utils_test.go
```diff
@@ -2354,13 +2499,15 @@ func getTestEndpointSlices(name, namespace string) []*discovery.EndpointSlice {
 	port81 := int32(81)
 	port8081 := int32(8081)
 	protocolTCP := v1.ProtocolTCP
+	managedByController := "endpointslice-controller.k8s.io"
```
swetharepakula, line 2502: use a constant instead

pkg/neg/syncers/transaction.go
sawsa307: To make it consistent with the order we pass parameters to NewL7EndpointsCalculator()

pkg/neg/syncers/utils.go
sawsa307: I check and the only place we define this const is in endpoint slice controller repo as an not exported const, so I create one in this file.

pkg/neg/syncers/utils.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/utils_test.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
swetharepakula: In this case since we added one more indexer, it requires us to change every instance to compile but in the future try not to change the order especially when the types are the same as this may not be caught in all instances.

pkg/neg/syncers/utils.go
```diff
@@ -48,6 +48,10 @@ const (
 	minRetryDelay = 5 * time.Second
 	maxRetryDelay = 600 * time.Second
 	separator     = "||"
+
+	// controllerName is a unique value used with LabelManagedBy to indicated
+	// the EndpointSlice is managed by the endpoint slice controller.
+	controllerName = "endpointslice-controller.k8s.io"
```
swetharepakula, line 54: maybe `managedByEPSControllerValue`

sawsa307: Updated. Thanks!

pkg/neg/syncers/utils.go
```diff
@@ -48,6 +48,10 @@ const (
 	minRetryDelay = 5 * time.Second
 	maxRetryDelay = 600 * time.Second
 	separator     = "||"
+
+	// managedByEPSControllerValue is a unique value used with LabelManagedBy to indicated
```
swetharepakula, line 52: typo: `to indicate`

sawsa307: Updated!

swetharepakula (commented):
/lgtm
/approve

PR #1967, opened 2023-02-23
Replace fetch from apiserver with informer cache while processing SvcNeg

gauravkghildiyal:
/assign @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #1969, opened 2023-02-24
bump kubernetes dependencies to 1.26.1

aojea:
Current version of the Kubernetes APIs is too old 1.20

Update the repo to latest stable version Kubernetes 1.26.1

There was necessary to adapt the codebase:
- compile with 1.20, that is the latest golang version and the one supported in current kubernetes version
- migrate from the deprecated and removed `apimachinery/clock` library to `k8s.io/utils/clock`https://github.com/kubernetes/utils/pull/224 in the `pkg/neg`folder
- migrate from v1.LoadBalancerStatus to networkingv1.IngressLoadbalancerStatus  https://github.com/kubernetes/kubernetes/pull/106242
- migrate from v1.Handler to v1.ProbeHandler https://github.com/kubernetes/kubernetes/pull/105979
- update golangci-lint to not fail on the new signature of `k8s.io/client-go/tools/cache/cache.SharedInformer.AddEventHandler`


cezarygerard:
lgtm overall
THANK YOU!

go.mod
```diff
@@ -1,78 +1,79 @@
 module k8s.io/ingress-gce
 
-go 1.18
+go 1.19
```
swetharepakula, line 3: should this be 1.20?

swetharepakula:
Thank you for this change! This looks good to me. The only question I had was whether we should regenerate go.mod with Go 1.20. But I think that doesn't need to keep this PR from merging.

/lgtm

go.mod
aojea: https://go.dev/ref/mod

> go directive
> A go directive indicates that a module was written assuming the semantics of a given version of Go. The version must be a valid Go release version: a positive integer followed by a dot and a non-negative integer (for example, 1.9, 1.14).

it was to be conservative and not make a big jump

PR #1970, opened 2023-02-27
Revert "bump kubernetes dependencies to 1.26.1"

aojea:
Reverts kubernetes/ingress-gce#1969

Too many changes and most probably one of them broke the downstream CI


aojea:
/close

it seems the problem was introduced in a different commit

PR #1972, opened 2023-02-27
Add names to queues used inside the NEG controller for enabling queue metrics

gauravkghildiyal:
Adding names to the queues has the side effect of enabling queue based metrics.

Refer https://github.com/kl52752/ingress-gce/blob/cdf339e0abde40ae1c1edf69ffccb095c9054c69/pkg/metrics/workqueue.go#L41-L90 for list of metrics.

/assign @swetharepakula


aojea:
/ok-to-test
/lgtm
/approve

hehe, this is one of the classics ;)

PR #1973, opened 2023-02-27
Only use endpoint slices to generate network endpoints

swetharepakula:
removes support for using endpoints to generate network endpoints

This PR combines https://github.com/kubernetes/ingress-gce/pull/1957 and does additional cleanup

/assign @aojea
/assign @bowei


pkg/flags/flags.go
```diff
@@ -253,7 +253,6 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.GKEClusterHash, "gke-cluster-hash", "", "The cluster hash of the GKE cluster this Ingress Controller will be interacting with")
 	flag.StringVar(&F.GKEClusterType, "gke-cluster-type", "ZONAL", "The cluster type of the GKE cluster this Ingress Controller will be interacting with")
 	flag.BoolVar(&F.EnableTrafficScaling, "enable-traffic-scaling", false, "Enable support for Service {max-rate-per-endpoint, capacity-scaler}")
-	flag.BoolVar(&F.EnableEndpointSlices, "enable-endpoint-slices", false, "Enable using Endpoint Slices API instead of Endpoints API")
```
bowei, line 256: Do we not use this flag anymore? Usually when retiring flags, you keep the command line param, but make it log so existing callers on the executable aren't broken. Then fix all of the callers and then delete the flag. If we can fix all of the associated call sites, then ok, we can delete this in one shot.

swetharepakula: We will be able to fix the call sites, so this is safe to delete.

aojea: so bowei nailed it i the review 😄

PR #1979, opened 2023-02-28
Fix NEG endpoint attachment for L4 NEGs.

mmamczur:
The L4 EndpointCalculators do not compute pod endpoints data. It always failed checks in CheckEndpointInfo but the failure was ignored. Now the code returns upon failure after changes from ff4751f08645101e5c0fc60857726d59d4a9d25a. This change skips the checks for calculators that don't return the pods data as these checks compare it with endpoints from slices.

### DIFF
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -109,6 +109,12 @@ func (l *LocalL4ILBEndpointsCalculator) CalculateEndpoints(eds []types.Endpoints
 	return subsetMap, nil, 0, err
 }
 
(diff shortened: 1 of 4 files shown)


aojea:
/ok-to-test
/assign @swetharepakula 
@sawsa307

swetharepakula:
I do think this fixes the issue for L4 but it is adding another layer of complexity to an already complicated piece of code. I would prefer to revert the original metrics PR instead.

cc/ @sawsa307

aojea:
/hold

the original PR was reverted in https://github.com/kubernetes/ingress-gce/pull/1980

sawsa307:
> /hold
> 
> the original PR was reverted in #1980

#1982 is the actual revert. #1980 is a PR dependent on #1982 so it is reverted as well.

PR #1981, opened 2023-02-28
Add smaller bucket sizes for workqueue metrics

gauravkghildiyal:
Some events related to NEG workqueues take subsecond time for completion. Having more buckets within the 1 second duration should help estimate us better if the NEG Controller is working as expected -- even a slight increase on these <1 second processing times could drastically effect the overall performance when the NEG Controller is managing too many events.


swetharepakula:
This seems reasonable to me. Cezary can you take a look and confirm this will not negatively affect existing workqueue metrics for l4.

\assign @cezarygerard

PR #1983, opened 2023-03-02
Refactor error state checking

sawsa307:
Create function ValidateEndpoints and move endpoint validation logics into EndpointCalculator. Now it would correctly validate endpoints for L4 and L7 endpoint calculators. Validation for L4 endpoints is currently a no-op.


sawsa307:
cc @mmamczur

pkg/neg/syncers/transaction.go
```diff
@@ -259,16 +261,15 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 	endpointsData := negtypes.EndpointsDataFromEndpointSlices(endpointSlices)
 	targetMap, endpointPodMap, dupCount, err = s.endpointsCalculator.CalculateEndpoints(endpointsData, currentMap)
-	if valid, reason := s.isValidEPField(err); !valid {
```
mmamczur, line 262: thanks for getting rid of this part as well, it looked super confusing with this isValidEPField potentially ignoring errors

mmamczur:
LGTM, probably best choice that you move this validation to the calculator

pkg/neg/types/sync_results.go
```diff
@@ -50,6 +50,17 @@ var (
 	ResultOtherError        = "OtherError"
 	ResultInProgress        = "InProgress"
 	ResultSuccess           = "Success"
+
+	ErrorStateReason = map[error]string{
```
swetharepakula, line 54: add a comment as to why we need this.

pkg/neg/types/sync_results.go
```diff
@@ -50,6 +50,17 @@ var (
 	ResultOtherError        = "OtherError"
```
swetharepakula, line 50: all the results should be constants.

sawsa307: Updated. Thanks!

pkg/neg/types/sync_results.go
sawsa307: Added. Thanks!

PR #1985, opened 2023-03-06
Add flag for Transparent Health Check

DamianSawicki:
This is a boolean feature flag to enable Transparent Health Checks (THC).

The purpose of the THC feature is to implement integration of

 * health checks in GCE called Unified Health Checks (UHC) and
 * health check mechanism in K8s,

so we can reflect K8s readiness status in UHC healthy status. The goal is to provide consistent health check results for GKE and UHC, showing the same status and operating on the same probers results.

K8s readiness status is a signal to enable the traffic flow for the POD. For GSLB, the health information is stored in the UHC configured for the BackendService. When those two signals are synchronized, there will be no disruption on traffic when the Backend will be marked as healthy. 

**Naming:** The THC health check mechanism is “transparent” from Kubernetes point of view because the actual K8s POD’s status is exposed to the GCE UHC system, visible for Users in GCE, and used by GSLB to route the traffic.

**Background:** In GKE the integration of Services with GCLB is built on top of GCE BackendServices. Backend Services are health checked using UHC and this health check signal is used to control traffic flow to the Backends (PODs). UHC healthy statuses are independent from Kubernetes health checks, which can create situations where both systems report different readiness values for the same backend and can lead to the discrepancy in marking backends ready to receive traffic. The K8s signal to receive the status is called POD’s readiness and UHC signal to start sending the traffic is called UHC healthy status.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -108,6 +108,7 @@ var (
 		EnablePSC                      bool
 		EnableIngressGAFields          bool
 		EnableTrafficScaling           bool
+		EnableTransparentHealthChecks  bool
 		EnablePinhole                  bool
 		EnableL4ILBDualStack           bool
 		EnableL4NetLBDualStack         bool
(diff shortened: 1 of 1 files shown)


DamianSawicki:
\assign kl52752

PR #1990, opened 2023-03-09
Add getPodLabelMap function for NegLabelPropagation and add unit test for it

songrx1997:
Add EndpointAnnotationMap to negtypes

Add getPodLabelMap function and truncateLabel function for label propagation 
 
Add unit tests

### DIFF
--- a/pkg/neg/syncers/labels/label_test.go
+++ b/pkg/neg/syncers/labels/label_test.go
@@ -1,0 +1,185 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package labels
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+)
+
+func TestGetPodLabelMap(t *testing.T) {
+	t.Parallel()
+	testContext := negtypes.NewTestContext()
+	podLister := testContext.PodInformer.GetIndexer()
+
+	namespace := "ns1"
+	name := "n1"
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+			Labels: map[string]string{
+				"app.kubernetes.io/name":    "pod-with-long-name",
+				"app.kubernetes.io/version": "v1",
+				"foo-key":                   "foo",
+			},
+		},
+	}
+	podLister.Add(pod)
+
+	for _, tc := range []struct {
+		desc      string
+		lpConfig  PodLabelPropagationConfig
+		expect    PodLabelMap
+		expectErr bool
+	}{
+		{
+			desc: "No short key for both labels, no truncation needed for both",
+			lpConfig: PodLabelPropagationConfig{
+				Labels: []Label{
+					{
+						Key:               "app.kubernetes.io/name",
+						MaxLabelSizeBytes: 40,
+					},
+					{
+						Key:               "app.kubernetes.io/version",
+						MaxLabelSizeBytes: 30,
+					},
+				},
+			},
+			expect: PodLabelMap{
+				"app.kubernetes.io/name":    "pod-with-long-name",
+				"app.kubernetes.io/version": "v1",
+			},
+			expectErr: false,
+		},
+		{
+			desc: "Short key for both labels, truncation needed for the first one.",
+			lpConfig: PodLabelPropagationConfig{
+				Labels: []Label{
+					{
+						Key:               "app.kubernetes.io/name",
+						ShortKey:          "name",
+						MaxLabelSizeBytes: 20,
+					},
+					{
+						Key:               "app.kubernetes.io/version",
+						ShortKey:          "version",
+						MaxLabelSizeBytes: 30,
+					},
+				},
+			},
+			expect: PodLabelMap{
+				"name":    "pod-with-long-na",
+				"version": "v1",
+			},
+			expectErr: true,
+		},
+		{
+			desc: "No short key for both labels, truncation needed for the first one, truncation failed for the second one",
+			lpConfig: PodLabelPropagationConfig{
+				Labels: []Label{
+					{
+						Key:               "app.kubernetes.io/name",
+						MaxLabelSizeBytes: 30,
+					},
+					{
+						Key:               "app.kubernetes.io/version",
+						MaxLabelSizeBytes: 20,
+					},
+				},
+			},
+			expect: PodLabelMap{
+				"app.kubernetes.io/name": "pod-with",
+			},
+			expectErr: true,
+		},
+	} {
+		ret, err := GetPodLabelMap(pod, tc.lpConfig)
+		if !reflect.DeepEqual(ret, tc.expect) {
+			t.Errorf("For test case %q, got label map %+v, want %+v", tc.desc, ret, tc.expect)
+		}
+		if (err != nil) != tc.expectErr {
+			t.Errorf("For test case %q, got error %v, expectErr: %t ", tc.desc, err, tc.expectErr)
+		}
+	}
+}
+
+func TestTruncateLabel(t *testing.T) {
+	for _, tc := range []struct {
+		desc              string
+		key               string
+		label             string
+		maxLabelSizeBytes int
+		expect            string
+		expectErr         error
+	}{
+		{
+			desc:              "No truncation needed",
+			key:               "app.kubernetes.io/name",
+			label:             "pod-with-long-name",
+			maxLabelSizeBytes: 40,
+			expect:            "pod-with-long-name",
+			expectErr:         nil,
+		},
+		{
+			desc:              "No truncation needed with label shorter than minLabelLength",
+			key:               "app.kubernetes.io/name",
+			label:             "pod",
+			maxLabelSizeBytes: 25,
+			expect:            "pod",
+			expectErr:         nil,
+		},
+		{
+			desc:              "Truncation needed",
+			key:               "app.kubernetes.io/name",
+			label:             "pod-with-long-name",
+			maxLabelSizeBytes: 30,
+			expect:            "pod-with",
+			expectErr:         ErrLabelTruncated,
+		},
+		{
+			desc:              "Truncation failed with key length exceeding limit",
+			key:               "app.kubernetes.io/name",
+			label:             "pod-with-long-name",
+			maxLabelSizeBytes: 20,
+			expect:            "",
+			expectErr:         ErrLabelTruncationFailed,
+		},
+		{
+			desc:              "Truncation failed with truncated label shorter than minLabelLength",
+			key:               "app.kubernetes.io/name",
+			label:             "pod-with-long-name",
+			maxLabelSizeBytes: 25,
+			expect:            "",
+			expectErr:         ErrLabelTruncationFailed,
+		},
+	} {
+		label, err := truncatePodLabel(tc.key, tc.label, tc.maxLabelSizeBytes)
+		if tc.expect != label {
+			t.Errorf("For test case %q, got label value: %s, want: %s,", tc.desc, label, tc.expect)
+		}
+		if !errors.Is(err, tc.expectErr) {
+			t.Errorf("For test case %q, got error: %v, want: %v", tc.desc, err, tc.expectErr)
+		}
+	}
+}
--- a/pkg/neg/syncers/labels/labels.go
+++ b/pkg/neg/syncers/labels/labels.go
@@ -19,7 +19,11 @@ package labels
 import (
 	"errors"
 
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/utils"
 )
 
 // PodLabelPropagationConfig contains a list of configurations for labels to be propagated to GCE network endpoints.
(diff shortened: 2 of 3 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -424,3 +425,54 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
...
+		if !ok {
+			klog.V(3).Infof("No Label %s found in pod %s", label.Key, key)
+			continue
+		}
+		lpKey := label.Key
+		if label.ShortKey != "" {
+			lpKey = label.ShortKey
+		}
+		if truncated, ok := truncateLabel(lpKey, val, label.MaxLabelSizeBytes); ok {
+			labelMap[lpKey] = truncated
```
swetharepakula, line 461: when will the event be emitted that this key was truncated? We would need the service this happened on.

swetharepakula (commented):
Mostly looks good. Need to add some logging and way to publish events when we truncate labels.

pkg/neg/syncers/utils.go
```diff
@@ -424,3 +425,54 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
...
+		return labelMap
+	}
+	key := fmt.Sprintf("%s/%s", namespace, name)
+	obj, exists, err := podLister.GetByKey(key)
+	if err != nil {
+		klog.Errorf("Failed to retrieve pod %s from pod lister: %v", key, err)
+		return labelMap
+	}
+	if !exists {
+		return labelMap
```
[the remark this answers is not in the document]
songrx1997: log added

pkg/neg/syncers/utils.go
```diff
@@ -424,3 +425,54 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
...
+		val, ok := pod.Labels[label.Key]
+		if !ok {
+			klog.V(3).Infof("No Label %s found in pod %s", label.Key, key)
+			continue
+		}
+		lpKey := label.Key
+		if label.ShortKey != "" {
+			lpKey = label.ShortKey
+		}
+		if truncated, ok := truncateLabel(lpKey, val, label.MaxLabelSizeBytes); ok {
```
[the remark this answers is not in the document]
songrx1997: events added for truncation failed and truncated

pkg/neg/syncers/utils.go
```diff
@@ -466,3 +467,63 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
 	}
 	return endpointBatch, nil
 }
+
+// getPodLabelMap will return a json string of a label map extracted from a pod according to PodLabelPropagationConfig
+// The return map json has the pod label key as key and label value as value
+func getPodLabelMap(namespace, name string, podLister cache.Indexer, lpConfig negtypes.PodLabelPropagationConfig, recorder record.EventRecorder) map[string]string {
```
swetharepakula, line 473: Change this to take in a logger instead of using klog. This way we will get the contextual logging.

songrx1997: Logger added to the function

pkg/neg/syncers/utils.go
```diff
@@ -466,3 +467,63 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
 	}
 	return endpointBatch, nil
 }
+
+// getPodLabelMap will return a json string of a label map extracted from a pod according to PodLabelPropagationConfig
+// The return map json has the pod label key as key and label value as value
+func getPodLabelMap(namespace, name string, podLister cache.Indexer, lpConfig negtypes.PodLabelPropagationConfig, recorder record.EventRecorder, logger klog.Logger) map[string]string {
+	labelMap := make(map[string]string, 0)
+	if podLister == nil {
```
bowei, line 475: This is weird. Why do we expect podLister to ever be nil?

pkg/neg/syncers/utils.go
```diff
@@ -466,3 +467,63 @@ func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.N
 	}
 	return endpointBatch, nil
 }
+
+// getPodLabelMap will return a json string of a label map extracted from a pod according to PodLabelPropagationConfig
+// The return map json has the pod label key as key and label value as value
+func getPodLabelMap(namespace, name string, podLister cache.Indexer, lpConfig negtypes.PodLabelPropagationConfig, recorder record.EventRecorder, logger klog.Logger) map[string]string {
```
bowei, line 473: This func eats the error from the Lister. It's better to pass the error up to the stack and let the caller decide what to do with the error.

songrx1997: Done. Add error returns for pod related issues.

pkg/neg/syncers/utils.go
songrx1997: Removed

songrx1997:
Separated the label truncation from the getLabelMap function. 
Will do the truncation after [getting the endpoints to be attached](https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction.go#L276) in order to 
1. Prevent truncation event emission for existing endpoints
2. Collect label propagation truncation metrics

pkg/neg/syncers/label_propagation.go
```diff
@@ -0,0 +1,58 @@
...
+		val, ok := pod.Labels[label.Key]
+		if ok {
+			labelMap[label.Key] = val
+		}
+	}
+	return labelMap
+}
+
+// truncateLabel will return the truncated value of a pod label and the result of the truncation
+func truncateLabel(key, val string, maxLabelSizeBytes int, pod *v1.Pod, recorder record.EventRecorder) (string, bool) {
```
bowei, line 40: We should mandate some minimum label size. It doesn't make sense to truncate the label down to 0 length.

You should not mix utility functions with things that do side effects (e.g. do stuff with events). Functions should do one thing at one abstraction level.

If maxTotalLen is in bytes, you need to encode the string to bytes and compare based on that value.

```go
const minLabelLength = 5 // for example

// truncatePodKeyLabel returns the potentially truncated label value to ensure that 
// len(key) + len(label) <= maxTotalLen. If it is not possible to fit, the func will return an error.
func truncatePodKeyLabel(key, label string, maxTotalLen int) (string, error) {
  if len(key) + minLabelLength > maxTotalLen {
    return fmt.Errorf("not enough space for the label (%d > %d)",len(key), minLabelLength, maxTotalLen),
  }
  if len(key) + len(label) > maxTotalLen {
    label = label[:maxTotalLen - len(key)]
  }
  return label, nil
}
```

put the event code as usage of the above func

```
label, err := truncatePodKeyLabel(key, label, maxTotalLen)
if err != nil {
  // Record the error
}

```

bowei:
Don't mix functional logic with side effects. Keep your functions focused on doing on thing.

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+			if label.ShortKey != "" {
+				lpKey = label.ShortKey
+			}
+			truncated, err, ok := truncatePodLabel(lpKey, val, label.MaxLabelSizeBytes)
+			if err != nil {
+				errs = append(errs, err)
+			}
+			if ok {
+				labelMap[lpKey] = truncated
+			}
```
swetharepakula, line 66: This behavior makes sense because we expect that truncation to be returned as an error. However this means for other errors we would keep the key and add an empty string. We should omit the the key in the other error case.

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+	labelBytes := []byte(label)
+	totalLen := len(keyBytes) + len(labelBytes)
+	if totalLen <= maxTotalSize {
+		return label, nil, true
+	}
+	if len(keyBytes)+minLabelLength > maxTotalSize {
+		return "", fmt.Errorf("Label `%s:%s` propagation failed because the key exceeded the limit, length: %d, limit %d", key, label, len(keyBytes)+minLabelLength, maxTotalSize), false
+	}
+	truncated := string(labelBytes[:maxTotalSize-len(keyBytes)])
+	return truncated, fmt.Errorf("Label `%s:%s` was truncated to `%s:%s` because the total length exceeded the limit, length: %d, limit: %d", key, label, key, truncated, len(key)+len(label), maxTotalSize), true
```
swetharepakula, line 88: This should not be an error case. We should just depend on the truncated bool here instead.

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+// EndpointAnnotationMap is a map of network endpoint, endpoint annotations.
+type EndpointAnnotationMap map[negtypes.NetworkEndpoint]PodLabelMap
+
+const minLabelLength = 5
+
+// getPodLabelMap will return the label map extracted from a pod according to PodLabelPropagationConfig
+// The returned map has the pod label key as key and label value as value
+func getPodLabelMap(pod *v1.Pod, lpConfig PodLabelPropagationConfig, recorder record.EventRecorder) (PodLabelMap, error) {
+	labelMap := PodLabelMap{}
+	errs := []error{}
```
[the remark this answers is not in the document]
songrx1997: Updated, thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+	}
+	if len(errs) != 0 {
+		return labelMap, utils.JoinErrs(errs)
+	}
+	return labelMap, nil
+}
+
+// truncatePodLabel returns the potentially truncated label value to ensure that len(key) + len(label) <= maxTotalSize.
+// This function will return an error if it is not possible to fit or truncation happens
+func truncatePodLabel(key, label string, maxTotalSize int) (string, error, bool) {
```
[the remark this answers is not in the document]
songrx1997: Updated, thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+	labelMap := PodLabelMap{}
+	errs := []error{}
+	for _, label := range lpConfig.Labels {
+		val, ok := pod.Labels[label.Key]
+		if ok {
+			lpKey := label.Key
+			if label.ShortKey != "" {
+				lpKey = label.ShortKey
+			}
+			truncated, err, ok := truncatePodLabel(lpKey, val, label.MaxLabelSizeBytes)
```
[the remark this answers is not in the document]
songrx1997: Changed it to truncatedVal, thanks!

pkg/neg/syncers/labels/labels.go
songrx1997: The bool returned by the function indicates whether the truncated value is valid.

Changed the naming of `ok` to `valid` to make it more readable and added more comment to `truncatePodLabel` function to clarify this.

pkg/neg/syncers/labels/labels.go
songrx1997: According to offline discussion, will handle this as an error.

bowei:
Ruixian -- can you resolve comments that are done?

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+// PodLabelMap is a map of pod label key, label values.
+type PodLabelMap map[string]string
+
+// EndpointAnnotationMap is a map of network endpoint, endpoint annotations.
+type EndpointAnnotationMap map[negtypes.NetworkEndpoint]PodLabelMap
+
+const minLabelLength = 5
+
+// getPodLabelMap will return the label map extracted from a pod according to PodLabelPropagationConfig
+// The returned map has the pod label key as key and label value as value
```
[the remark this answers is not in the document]
songrx1997: Document added, thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+// This function will return an error if it is not possible to fit or truncation happens
+func truncatePodLabel(key, label string, maxTotalSize int) (string, error, bool) {
+	keyBytes := []byte(key)
+	labelBytes := []byte(label)
+	totalLen := len(keyBytes) + len(labelBytes)
+	if totalLen <= maxTotalSize {
+		return label, nil, true
+	}
+	if len(keyBytes)+minLabelLength > maxTotalSize {
+		return "", fmt.Errorf("Label `%s:%s` propagation failed because the key exceeded the limit, length: %d, limit %d", key, label, len(keyBytes)+minLabelLength, maxTotalSize), false
```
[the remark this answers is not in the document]
songrx1997: Defined error types for this and removed the bool here.
Will handle the errors in the caller according to error types.

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
+		}
+	}
+	if len(errs) != 0 {
+		return labelMap, utils.JoinErrs(errs)
+	}
+	return labelMap, nil
+}
+
+// truncatePodLabel returns the potentially truncated label value to ensure that len(key) + len(label) <= maxTotalSize.
+// This function will return an error if it is not possible to fit or truncation happens
```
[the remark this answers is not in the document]
songrx1997: Removed the bool and added comment to describe the return values, thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
...
 	MaxLabelSizeBytes int
 }
+
+// PodLabelMap is a map of pod label key, label values.
+type PodLabelMap map[string]string
+
+// EndpointAnnotationMap is a map of network endpoint, endpoint annotations.
+type EndpointAnnotationMap map[negtypes.NetworkEndpoint]PodLabelMap
+
+const minLabelLength = 5
```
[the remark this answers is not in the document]
songrx1997: Comment added, thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -27,3 +36,54 @@ type Label struct {
 	ShortKey          string
 	MaxLabelSizeBytes int
 }
+
+// PodLabelMap is a map of pod label key, label values.
+type PodLabelMap map[string]string
+
+// EndpointAnnotationMap is a map of network endpoint, endpoint annotations.
+type EndpointAnnotationMap map[negtypes.NetworkEndpoint]PodLabelMap
```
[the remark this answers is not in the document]
songrx1997: Regarding this, the naming is based on the field name `Annotations` in NetworkEndpoint object defined [here](https://github.com/kubernetes/ingress-gce/blob/master/pkg/composite/gen.go#L2392).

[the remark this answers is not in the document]
swetharepakula: I think PodLabel might be easier to follow here, because annotations could mean annotation on the network endpoint as well as the annotation on the endpoint slice resource itself.

pkg/neg/syncers/labels/labels.go
```diff
@@ -44,3 +48,55 @@ var (
 	ErrLabelTruncated        = errors.New("label is truncated")
 	ErrLabelTruncationFailed = errors.New("failed to truncate label")
 )
+
+// minLabelLength defines the minimum label value length when doing truncation.
```
swetharepakula, line 52: nit: the minimum space left for the label value

The label value could be in reality be shorter

pkg/neg/syncers/labels/labels.go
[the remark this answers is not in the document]
songrx1997: Make sense, just changed it to `EndpointPodLabelMap `. Thanks!

pkg/neg/syncers/labels/labels.go
```diff
@@ -44,3 +48,55 @@ var (
 	ErrLabelTruncated        = errors.New("label is truncated")
 	ErrLabelTruncationFailed = errors.New("failed to truncate label")
 )
+
+// minLabelLength defines the minimum space left for the label value after truncation.
```
swetharepakula, line 52: remove `after truncation`. Its for when the key is too long and there is no space for the label.

pkg/neg/syncers/labels/labels.go
```diff
@@ -44,3 +48,55 @@ var (
...
+			}
+			truncatedVal, err := truncatePodLabel(lpKey, val, label.MaxLabelSizeBytes)
+			if err != nil {
+				errs = append(errs, err)
+			}
+
+			// Add the label to the map only if the truncation result is valid
+			if !errors.Is(err, ErrLabelTruncationFailed) {
+				labelMap[lpKey] = truncatedVal
+			}
```
swetharepakula, line 76: Flip the conditional
```
if errors.Is(err, ErrLabelTruncated) {
    labelMap[lpKey] = truncatedVal
}

```

pkg/neg/syncers/labels/label_test.go
```diff
@@ -0,0 +1,168 @@
...
+						MaxLabelSizeBytes: 20,
+					},
+				},
+			},
+			expect: PodLabelMap{
+				"app.kubernetes.io/name": "pod-with",
+			},
+		},
+	} {
+		ret, _ := GetPodLabelMap(pod, tc.lpConfig)
```
swetharepakula, line 112: add an expectation on whether an error is returned or not.

pkg/neg/syncers/labels/label_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+		{
+			desc:              "Truncation failed with truncated label shorter than minLabelLength",
+			key:               "app.kubernetes.io/name",
+			label:             "pod-with-long-name",
+			maxLabelSizeBytes: 25,
+			expect:            "",
+		},
+	} {
+		label, _ := truncatePodLabel(tc.key, tc.label, tc.maxLabelSizeBytes)
```
swetharepakula, line 163: add an expectation on whether an error is returned or not.

swetharepakula (commented):
A few small changes needed. This PR is almost ready to be merged.

pkg/neg/syncers/labels/labels.go
songrx1997: Updated, thanks!

pkg/neg/syncers/labels/label_test.go
songrx1997: Added expectErr for this, thanks!

pkg/neg/syncers/labels/label_test.go
songrx1997: Added expectErr for this, thanks!

pkg/neg/syncers/labels/labels.go
songrx1997: Changed this to 
```
if err == nil || errors.Is(err, ErrLabelTruncated)
```

pkg/neg/syncers/labels/label_test.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+			expectErr: true,
+		},
+	} {
+		ret, err := GetPodLabelMap(pod, tc.lpConfig)
+		if !reflect.DeepEqual(ret, tc.expect) {
+			t.Errorf("For test case %q, label map output = %+v, but got %+v", tc.desc, tc.expect, ret)
+		}
+		if (err != nil) != tc.expectErr {
+			t.Errorf("For test case %q, expect error %v, but got %v", tc.desc, tc.expectErr, err)
```
swetharepakula, line 122: nit: do got before want.

%t is used for bools

gauravkghildiyal: +1 to `got before want`

As a general note, it would also be nice if we can start using `cmp.Diff` instead of `reflect.DeepEqual` -- atleast for newer tests. (No need to do it now, lets target getting this in)

Refer: https://google.github.io/styleguide/go/decisions#equality-comparison-and-diffs

songrx1997: Updated, thanks for the suggestions!

swetharepakula (commented):
/lgtm
/approve

PR #1992, opened 2023-03-09
Fix error caused by unhashable error list

sawsa307:
Fix error caused by unhashable error aggregation in getErrorStateReason()


pkg/neg/syncers/transaction.go
```diff
@@ -347,6 +347,10 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 
 // getErrorStateReason returns the reason of the error state based returned error
 func (s *transactionSyncer) getErrorStateReason(err error) string {
+	// these errors only occurs on NEG creation, and they are not related to endpoint sync
```
swetharepakula, line 350: we should indicate that they are also not a relevant error state reason.

swetharepakula (commented):
/lgtm
/approve
/ok-to-test

PR #1993, opened 2023-03-09
Create the flag for degraded mode

sawsa307:
Create a flag that enables degraded mode endpoint calculation.


swetharepakula (commented):
/lgtm
/approve
/ok-to-test

PR #1994, opened 2023-03-09
Update duplicate endpoint handling

sawsa307:
When we encounter duplicate endpoints, we would always choose the one with an alphabetically lower name so we have a consistent and reliable way to break the tie.

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -272,9 +272,12 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
 				}
 				zoneNetworkEndpointMap[zone].Insert(networkEndpoint)
 
-				// increment the count for duplicated endpoint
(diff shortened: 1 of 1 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -272,9 +272,13 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
 				}
 				zoneNetworkEndpointMap[zone].Insert(networkEndpoint)
 
-				// increment the count for duplicated endpoint
+				// increment the count for duplicate endpoint
 				if _, contains := networkEndpointPodMap[networkEndpoint]; contains {
 					dupCount += 1
+					existingPod, contains := networkEndpointPodMap[networkEndpoint]
```
swetharepakula, line 278: the same call is being made in line 276. Instead just set existingPod there

swetharepakula (commented):
/lgtm
/approve

PR #1996, opened 2023-03-10
Add field Annotations to NetworkEndpoint struct

songrx1997:


pkg/neg/types/network_endpoint.go
```diff
@@ -22,9 +22,10 @@ import (
 
 // NetworkEndpoint contains the essential information for each network endpoint in a NEG
 type NetworkEndpoint struct {
-	IP   string
-	Port string
-	Node string
+	IP          string
+	Port        string
+	Node        string
+	Annotations string
```
swetharepakula, line 28: This may cause a problem once we start populating this field since we use this as a key in our EndpointSet which is used to calculate the differences in desired and current. Since we should not consider an endpoint with annotations different from ones without an annotation. we shouldn't add the annotations here.

PR #2006, opened 2023-03-13
Fix PublishL4NetLBDualStackSyncLatency

panslava:
Order of execution was wrong, function was fast returned  

and some small logs improvements 

/assign aojea


code-elinka:
Do you know what happened to the workflow?

PR #2007, opened 2023-03-13
Add log of processing time for the service metrics.

mmamczur:


pkg/servicemetrics/servicemetrics.go
```diff
@@ -175,6 +177,7 @@ func (c *Controller) export() {
 	l4ProtocolState, ipStackState, gcpFeaturesState := calculateMetrics(allServices)
 
 	updatePrometheusMetrics(l4ProtocolState, ipStackState, gcpFeaturesState)
+	klog.Infof("Exported service metrics in %v ms", time.Since(start).Microseconds())
```
panslava, line 180: Sorry for nit-picking, but why exactly Microseconds (maybe just time.Since()), and feels like "ms" is milliseconds

PR #2008, opened 2023-03-14
Adjust health check Description when BackendConfig used

DamianSawicki:
This is the first preparatory PR in a series of two. The intention of the second PR is to detect the removal of `BackendConfig` CRD and to update health check configuration according to the existing input (rather than keep the existing configuration from `BackendConfig`).

The present PR updates the health check `Description` when `BackendConfig` is used. Additionally, test cases are modified to verify health check `Description`s.

### DIFF
--- a/pkg/healthchecks/healthchecks_test.go
+++ b/pkg/healthchecks/healthchecks_test.go
@@ -770,6 +770,7 @@ func (*syncSPFixture) hc() *compute.HealthCheck {
 			Port:        80,
 			RequestPath: "/",
 		},
+		Description: translator.DescriptionForDefaultHealthChecks,
 	}
 }
 
@@ -809,6 +810,7 @@ func (*syncSPFixture) neg() *compute.HealthCheck {
 			PortSpecification: "USE_SERVING_PORT",
 			RequestPath:       "/",
 		},
+		Description: translator.DescriptionForDefaultNEGHealthChecks,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/healthchecks/healthchecks_test.go
```diff
@@ -1237,14 +1251,16 @@ func TestSyncServicePort(t *testing.T) {
...
 				filter := func(hc *compute.HealthCheck) {
-					hc.Description = ""
 					hc.SelfLink = ""
 				}
-				filter(gotHC)
-				filter(tc.wantComputeHC)
-
-				if !reflect.DeepEqual(gotHC, tc.wantComputeHC) {
-					t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(gotHC), pretty.Sprint(tc.wantComputeHC))
+				// Create copies to be modified by the above func filter as otherwise resync may try to revert modifications.
```
kl52752, line 1256: Add tc := tc at the beginning of the for loop so your tc won't be override.
The re-sync should not change your values so try if the tc fix work if not maybe there is some bug in the test case.

DamianSawicki: I'm not sure how important it is not to overwrite `tc.wantComputeHC`, but I think we should not to overwrite `gotHC`. In the previous version of the code, `Description` was overwritten in `filter` to `""`, and then re-sync – after I added checking `Description` in `calculateDiff` – changed the value of `Description` back. This is how I noticed that it would be better to use copies. Since then, I removed overwriting `Description` from `filter` and added `Description`s to `want`s, so copies are not technically necessary, but I think they still make sense. I will experiment with your suggestion.

PR #2009, opened 2023-03-14
Move BackoffHandler and RetryHandler to a separate package

alexkats:
As BackoffHandler and RetryHandler are already general enough (and not specific to NEG syncer), it makes sense to move them to a more common package, to be able to use it in other places. 

Also there was added a `DecreaseDelay` function to the interface for gradual decrease of the delay.

### DIFF
--- a/pkg/backoff/backoff.go
+++ b/pkg/backoff/backoff.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package syncers
+package backoff
 
 import (
 	"fmt"
--- a/pkg/backoff/backoff_test.go
+++ b/pkg/backoff/backoff_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package syncers
+package backoff
 
 import (
 	"testing"
--- a/pkg/neg/syncers/backoff.go
+++ b/pkg/neg/syncers/backoff.go
@@ -26,17 +26,19 @@ import (
 
 var ErrRetriesExceeded = fmt.Errorf("maximum retry exceeded")
 
-// backoffHandler handles delays for back off retry
-type backoffHandler interface {
-	// NextRetryDelay returns the delay for next retry or error if maximum number of retries exceeded.
-	NextRetryDelay() (time.Duration, error)
-	// ResetRetryDelay resets the retry delay
-	ResetRetryDelay()
+// BackoffHandler handles delays for back off retry
+type BackoffHandler interface {
+	// NextDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextDelay() (time.Duration, error)
+	// ResetDelay resets the retry delay
+	ResetDelay()
+	// DecreaseDelay returns the decreased delay for next retry
+	DecreaseDelay() time.Duration
 }
 
-// exponentialBackOffHandler is a backoff handler that returns retry delays semi-exponentially with random jitter within boundary.
-// exponentialBackOffHandler returns ErrRetriesExceeded when maximum number of retries has reached.
-type exponentialBackOffHandler struct {
+// exponentialBackoffHandler is a backoff handler that returns retry delays semi-exponentially with random jitter within boundary.
+// exponentialBackoffHandler returns ErrRetriesExceeded when maximum number of retries has reached.
+type exponentialBackoffHandler struct {
 	lock           sync.Mutex
 	lastRetryDelay time.Duration
 	retryCount     int
@@ -45,8 +47,8 @@ type exponentialBackOffHandler struct {
 	maxRetryDelay  time.Duration
 }
 
-func NewExponentialBackendOffHandler(maxRetries int, minRetryDelay, maxRetryDelay time.Duration) *exponentialBackOffHandler {
-	return &exponentialBackOffHandler{
+func NewExponentialBackoffHandler(maxRetries int, minRetryDelay, maxRetryDelay time.Duration) BackoffHandler {
+	return &exponentialBackoffHandler{
 		lastRetryDelay: time.Duration(0),
 		retryCount:     0,
 		maxRetries:     maxRetries,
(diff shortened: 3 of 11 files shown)


pkg/backoff/backoff.go
```diff
@@ -72,10 +74,26 @@ func (handler *exponentialBackOffHandler) NextRetryDelay() (time.Duration, error
...
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+	handler.retryCount = 0
+	if handler.lastRetryDelay == handler.minRetryDelay {
+		handler.lastRetryDelay = time.Duration(0)
+	} else {
+		handler.lastRetryDelay = handler.lastRetryDelay*2 - wait.Jitter(handler.lastRetryDelay, 0.5)
+		if handler.lastRetryDelay < handler.minRetryDelay {
+			handler.lastRetryDelay = handler.minRetryDelay
+		}
```
swetharepakula, line 96: Based on the name of the function, this seems a little unintuitive as the delay is actually getting increased. This happens when lastRetryDelay has been set to 0 like in line 91, so this would cause the delay to keep going between minRetryDelay and 0 if you keep calling DecreaseDelay.

I think reducing to 0 might make more sense here.

pkg/backoff/backoff_test.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
swetharepakula, line 2: nit: I recommend in future PRs, make one commit that renames the package and put all the changes in a second commit, so that files like this would be recognized as a rename. That helps keep the git history of the file.

pkg/backoff/backoff_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+func verifyIntervalDelay(t *testing.T, delay, expectedMinDelay, expectedMaxDelay time.Duration) {
+	if delay < expectedMinDelay || delay > expectedMaxDelay {
+		t.Errorf("Expect retry delay between %v and %v, but got %v", expectedMinDelay, expectedMaxDelay, delay)
+	}
+}
+
+func verifyIncreasedDelay(t *testing.T, delay, expectedInitialDelay time.Duration) {
```
swetharepakula, line 48: nit: initialDelay instead of expectedInitialDelay

pkg/backoff/backoff_test.go
```diff
@@ -0,0 +1,122 @@
...
+	expectedMaxDelay := expectedInitialDelay * 2
+	if expectedMaxDelay > testMaxRetryDelay {
+		expectedMaxDelay = testMaxRetryDelay
+	}
+	verifyIntervalDelay(t, delay, expectedInitialDelay, expectedMaxDelay)
+}
+
+func verifyDecreasedDelay(t *testing.T, delay, expectedInitialDelay time.Duration) {
```
swetharepakula, line 56: same comment as below.

pkg/backoff/backoff_test.go
```diff
@@ -0,0 +1,122 @@
...
+	delay := handler.DecreaseDelay()
+	verifyDecreasedDelay(t, delay, expectDelay)
+	expectDelay = delay
+
+	for i := 0; i < testRetry; i++ {
+		delay, err := handler.NextDelay()
+		verifyError(t, err, nil)
+		verifyIncreasedDelay(t, delay, expectDelay)
+		expectDelay = delay
+	}
```
swetharepakula, line 109: Can this be combined with the for loop on line 93?

swetharepakula (commented):
I have one question about Decrease delay and how to handle one case. Otherwise the rest of the changes look good.

pkg/backoff/backoff.go
alexkats: Yeah, my bad here, but I'd rather change condition on line 90 to `<=` instead of `==`

pkg/backoff/backoff_test.go
alexkats: Not really, because I need to invoke `DecreaseDelay` in between, but I introduced helper function and put this loop there

pkg/backoff/backoff_test.go
alexkats: Decided to split it in this PR as well, thanks for the suggestion

pkg/backoff/backoff.go
alexkats: Also added a test for this situation

pkg/neg/syncers/backoff.go
```diff
@@ -26,17 +26,19 @@ import (
...
-	// ResetRetryDelay resets the retry delay
-	ResetRetryDelay()
+// BackoffHandler handles delays for back off retry
+type BackoffHandler interface {
+	// NextDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextDelay() (time.Duration, error)
+	// ResetDelay resets the retry delay
+	ResetDelay()
+	// DecreaseDelay returns the decreased delay for next retry
+	DecreaseDelay() time.Duration
```
aojea, line 36: I'm not familiar with this kind of method, in theory the backoff should increase, what is the use case to decrease?

pkg/neg/syncers/backoff_test.go
```diff
@@ -27,39 +27,84 @@ const (
...
+
+func verifyIncreasedDelay(t *testing.T, delay, initialDelay time.Duration) {
+	expectedMaxDelay := initialDelay * 2
+	if expectedMaxDelay > testMaxRetryDelay {
+		expectedMaxDelay = testMaxRetryDelay
+	}
+	verifyIntervalDelay(t, delay, initialDelay, expectedMaxDelay)
+}
 
+func verifyDecreasedDelay(t *testing.T, delay, initialDelay time.Duration) {
```
aojea: same as above, this function seems to be used only in one test and with hardcoded value, I don't feel we should generalize it and it makes you more difficult to read the test, since you have to go here to know that is checking that this means half of the delay

pkg/neg/syncers/backoff.go
aojea: is there any place I can read more about it?
It's hard to review without that context

aojea:
ok, the I defer to someone with more context, I personally want to get all the expected feature in one PR with different commits , so I can review the whole context and iterate over the commits

PR #2012, opened 2023-03-15
Process IPv6 Endpoints 

swetharepakula:
This PR moves the logic of skipping IPv6 endpoints to the individual endpoint calculators. This allows L4 to use the ipv6 endpoints and l7 to ignore them.


panslava:
thank you @swetharepakula ! good job

I verified on my cluster, and l4 ILB ipv6 singlestack etp local is working

code looks fine as well

I can give lgtm, or you want someone else to also take a look?

PR #2014, opened 2023-03-15
Add code-elinka to the owners list

code-elinka:


OWNERS
```diff
@@ -9,5 +9,6 @@ approvers:
 - panslava
 - cezarygerard
 - kl52752
+- code-elinka
```
aojea, line 12: do you use to skip the ladder and go to superpowers directly? don't you start as reviewer first?

code-elinka: @aojea valid comment! :crab: 
Check it now please

PR #2016, opened 2023-03-15
Enable Degraded mode endpoint calculation

sawsa307:
Use results from degraded mode endpoint calculation only when the flag enableDegradedMode is true and we are in error state. Flag is created in #1993.

### DIFF
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -199,7 +199,7 @@ func (l *L7EndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
 func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
-	return toZoneNetworkEndpointMap(eds, l.zoneGetter, l.servicePortName, l.networkEndpointType, l.lpConfig)
+	return toZoneNetworkEndpointMap(eds, l.podLister, l.zoneGetter, l.servicePortName, l.networkEndpointType, l.lpConfig)
 }
 
 func nodeMapToString(nodeMap map[string][]*v1.Node) string {
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -61,7 +61,7 @@ func (l *LocalL4ILBEndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *LocalL4ILBEndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+func (l *LocalL4ILBEndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet, _ bool) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
 	// List all nodes where the service endpoints are running. Get a subset of the desired count.
 	zoneNodeMap := make(map[string][]*v1.Node)
 	processedNodes := sets.String{}
@@ -147,7 +147,7 @@ func (l *ClusterL4ILBEndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *ClusterL4ILBEndpointsCalculator) CalculateEndpoints(_ []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+func (l *ClusterL4ILBEndpointsCalculator) CalculateEndpoints(_ []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet, _ bool) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
 	// In this mode, any of the cluster nodes can be part of the subset, whether or not a matching pod runs on it.
 	nodes, _ := utils.ListWithPredicate(l.nodeLister, utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes)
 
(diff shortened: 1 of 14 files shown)


pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -198,8 +204,12 @@ func (l *L7EndpointsCalculator) Mode() types.EndpointsCalculatorMode {
...
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
-	return toZoneNetworkEndpointMap(eds, l.zoneGetter, l.servicePortName, l.networkEndpointType, l.lpConfig)
+func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ map[string]types.NetworkEndpointSet, inErrorState bool) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+	if l.enableDegradedMode && inErrorState {
+		return toZoneNetworkEndpointMapDegradedMode(eds, l.zoneGetter, l.podLister, l.nodeLister, l.servicePortName, l.networkEndpointType)
+	} else {
+		return toZoneNetworkEndpointMap(eds, l.zoneGetter, l.servicePortName, l.networkEndpointType, l.lpConfig)
+	}
```
swetharepakula, line 212: When degraded mode is calculated, we want to always do both calculations so we can emit metrics about the diff. Error state will dictate which calculation we will use.

perhaps something like this. The metrics for now could be a comment with a todo.
```
if !l.enableDegradedMode {
    return toZoneNetworkEndpointMap (...)
}

legacyCalculations := toZoneNetworkEndpointMap (...)
degradedModeCalculations := toZoneNetworkEndpointMapDegradedMode (...)
publishCalculationDiffMetric()

if errorState {
    return degradedModeCalculations
}
return legacyCalculations
```

pkg/neg/syncers/transaction_test.go
```diff
@@ -1431,8 +1431,8 @@ func newTestTransactionSyncer(fakeGCE negtypes.NetworkEndpointGroupCloud, negTyp
 		testContext.NodeInformer.GetIndexer(),
 		testContext.SvcNegInformer.GetIndexer(),
 		reflector,
-		GetEndpointsCalculator(testContext.NodeInformer.GetIndexer(), testContext.PodInformer.GetIndexer(), fakeZoneGetter,
-			svcPort, mode, klog.TODO(), negtypes.PodLabelPropagationConfig{}),
+		GetEndpointsCalculator(testContext.NodeInformer.GetIndexer(), testContext.PodInformer.GetIndexer(), testContext.SvcNegInformer.GetIndexer(),
```
swetharepakula, line 1434: Are there any tests for the L7 calculate endpoints path that will test the degraded mode logic added?

pkg/neg/syncers/endpoints_calculator.go
sawsa307: Hi Swetha I have this change in #2026. A few thing I want to address: I think we don't need to calculate both when we are in errorState, since the result from legacy will always be nil.

pkg/neg/syncers/transaction_test.go
sawsa307: Updpated in endpoints_calculator_test.go. Thank you!

pkg/neg/syncers/transaction.go
```diff
@@ -193,6 +193,7 @@ func (s *transactionSyncer) syncInternal() error {
 	start := time.Now()
 	err := s.syncInternalImpl()
 	if err != nil {
+		s.logger.V(3).Info("Setting error state", "err", err, "errorState", s.getErrorStateReason(err))
```
bowei, line 191: You should have setErrorState() take in the err directly.

Put the logging in the func

pkg/neg/syncers/transaction.go
```diff
@@ -261,8 +262,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	endpointsData := negtypes.EndpointsDataFromEndpointSlices(endpointSlices)
 	targetMap, endpointPodMap, dupCount, err = s.endpointsCalculator.CalculateEndpoints(endpointsData, currentMap)
 	if err != nil {
-		s.setErrorState(s.getErrorStateReason(err))
-		return fmt.Errorf("endpoints calculation error in mode %q, err: %w", s.endpointsCalculator.Mode(), err)
+		return err
```
bowei, line 260: why don't we set errorState here?

pkg/neg/syncers/utils.go
```diff
@@ -218,7 +218,7 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 }
 
 // toZoneNetworkEndpointMap translates addresses in endpoints object into zone and endpoints map, and also return the count for duplicated endpoints
-func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, lpConfig negtypes.PodLabelPropagationConfig) (map[string]negtypes.NetworkEndpointSet, negtypes.EndpointPodMap, int, error) {
+func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, podLister cache.Indexer, zoneGetter negtypes.ZoneGetter, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, lpConfig negtypes.PodLabelPropagationConfig) (map[string]negtypes.NetworkEndpointSet, negtypes.EndpointPodMap, int, error) {
```
bowei, line 221: [style] this is a pretty long line, suggest

```
func toZoneNetworkEndpointMap(
  eds []negtypes.EndpointsData, 
  podLister cache.Indexer,
  zoneGetter negtypes.ZoneGetter,
  servicePortName string,
  networkEndpointType negtypes.NetworkEndpointType,
  lpConfig negtypes.PodLabelPropagationConfig
) (map[string]negtypes.NetworkEndpointSet, negtypes.EndpointPodMap, int, error) {
```

suggest create a struct for the return value

```
type zoneNetworkEndpointMapResult struct {
  networkEndpointSet map[string]...
  endpointPodMap ...
  duplicateCount int
}
```

pkg/neg/syncers/transaction.go
sawsa307: Because when we return error, we will jump back to line 195, which will set error state if err != nil. This is a duplicate setErrorState.

pkg/neg/syncers/transaction.go
sawsa307: Got it! Let me update that.

pkg/neg/syncers/utils.go
sawsa307: Understood! Let me update that

pkg/neg/syncers/utils.go
```diff
@@ -251,19 +251,31 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
...
 			}
 			zone, err := zoneGetter.GetZoneForNode(*endpointAddress.NodeName)
 			if err != nil {
-				return nil, nil, dupCount, negtypes.ErrNodeNotFound
+				return nil, nil, dupCount, negtypes.ErrEPNodeNotFound
 			}
 			if zone == "" {
-				klog.V(2).Info("Detected unexpected error when checking missing zone")
-				return nil, nil, dupCount, negtypes.ErrEPMissingZone
+				klog.V(2).Info("Detected unexpected error when checking zone")
```
bowei, line 265: klog Error

pkg/neg/syncers/utils.go
```diff
@@ -251,19 +251,31 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
...
+				return nil, nil, dupCount, negtypes.ErrEPZoneMissing
+			}
+
+			key := fmt.Sprintf("%s/%s", endpointAddress.TargetRef.Namespace, endpointAddress.TargetRef.Name)
+			obj, exists, err := podLister.GetByKey(key)
+			if err != nil || !exists {
+				klog.V(2).Infof("Endpoint %q in Endpoints %s/%s does not correspond to an existing pod. Skipping", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
```
bowei, line 272: why aren't these errors?

bowei: log error

sawsa307: I'll update the message.

pkg/neg/syncers/utils.go
```diff
@@ -251,19 +251,31 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
...
+				return nil, nil, dupCount, negtypes.ErrEPPodNotFound
+			}
+			_, ok := obj.(*apiv1.Pod)
+			if !ok {
+				klog.V(2).Infof("Endpoint %q in Endpoints %s/%s does not correspond to an existing pod. Skipping", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
```
bowei, line 277: This doesn't seem to match what is happening here. !ok means we couldn't convert the type of an `any` to Pod. Change the log message to 

Errorf("couldn't convert obj to *Pod: got %T", obj)

Why is this counted as ErrEPPodNotFound?

sawsa307: I want to convey that the pod namespace and name of this pod doesn't match to a `Pod` resouce. Should I add another type of error as return error?

PR #2018, opened 2023-03-16
Flag for changes regarding health checks with BackendConfig

DamianSawicki:
~~Safeguarding the main change from https://github.com/kubernetes/ingress-gce/pull/2008/ with a flag.~~

Safeguard the changes from https://github.com/kubernetes/ingress-gce/pull/2008/ with a flag. Test dynamically enabling this flag.

### DIFF
--- a/pkg/healthchecks/healthcheck.go
+++ b/pkg/healthchecks/healthcheck.go
@@ -35,6 +35,7 @@ func (c *fieldDiffs) add(field, oldv, newv string) {
 }
 func (c *fieldDiffs) String() string { return strings.Join(c.f, ", ") }
 func (c *fieldDiffs) hasDiff() bool  { return len(c.f) > 0 }
+func (c *fieldDiffs) size() int64    { return int64(len(c.f)) }
 
 func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCheckConfig) *fieldDiffs {
 	var changes fieldDiffs
@@ -73,6 +74,7 @@ func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCh
 		changes.add("Port", strconv.FormatInt(old.Port, 10), strconv.FormatInt(new.Port, 10))
 	}
 	if old.Description != new.Description {
+		// TODO(DamianSawicki): Emit an event.
 		changes.add("Description", old.Description, new.Description)
 	}
 
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -131,7 +131,15 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		hc.UpdateFromBackendConfig(bchcc)
 	}
 
-	changes := calculateDiff(existingHC, hc, bchcc)
+	filter := func(hc *translator.HealthCheck) *translator.HealthCheck {
+		var ans = *hc // Shallow copy.
+		if !flags.F.EnableUpdateCustomHealthCheckDescription {
+			ans.Description = ""
+		}
+		return &ans
+	}
+
+	changes := calculateDiff(filter(existingHC), filter(hc), bchcc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
 		err := h.update(hc)
--- a/pkg/healthchecks/healthchecks_test.go
+++ b/pkg/healthchecks/healthchecks_test.go
@@ -429,6 +429,96 @@ func TestHealthCheckUpdate(t *testing.T) {
 	}
 }
 
+func getSingletonHealthcheck(t *testing.T, c *gce.Cloud) *compute.HealthCheck {
+	t.Helper()
+	// Verify the health check exists
+	var computeHCs []*compute.HealthCheck
+	computeHCs, _ = c.Compute().HealthChecks().List(context.Background(), nil)
+	if len(computeHCs) != 1 {
+		t.Fatalf("Got %d healthchecks, want 1\n%s", len(computeHCs), pretty.Sprint(computeHCs))
+	}
+	return utils.DeepCopyComputeHealthCheck(computeHCs[0]) // Make a copy to avoid reading an overwritten version later.
+}
+
+// Test changing the value of the flag EnableUpdateCustomHealthCheckDescription from false to true.
+func TestRolloutUpdateCustomHCDescription(t *testing.T) {
+	// No parallel() because we modify the value of the flags EnableBackendConfigHealthCheck and EnableUpdateCustomHealthCheckDescription.
+	oldEnableBC := flags.F.EnableBackendConfigHealthCheck
+	oldUpdateDescription := flags.F.EnableUpdateCustomHealthCheckDescription
+	flags.F.EnableBackendConfigHealthCheck = true
+	// Start with EnableUpdateCustomHealthCheckDescription = false.
+	flags.F.EnableUpdateCustomHealthCheckDescription = false
+	defer func() {
+		flags.F.EnableBackendConfigHealthCheck = oldEnableBC
+		flags.F.EnableUpdateCustomHealthCheckDescription = oldUpdateDescription
+	}()
+
+	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
+
+	var (
+		defaultSP       *utils.ServicePort = testSPs["HTTP-80-reg-nil"]
+		backendConfigSP *utils.ServicePort = testSPs["HTTP-80-reg-bc"]
+	)
+
+	// Add Hooks
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockHealthChecks.UpdateHook = mock.UpdateHealthCheckHook
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaHealthChecks.UpdateHook = mock.UpdateAlphaHealthCheckHook
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockBetaHealthChecks.UpdateHook = mock.UpdateBetaHealthCheckHook
+
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+
+	_, err := healthChecks.SyncServicePort(defaultSP, nil)
+	if err != nil {
+		t.Fatalf("unexpected err while syncing healthcheck, err %v", err)
+	}
+
+	outputDefaultHC := getSingletonHealthcheck(t, fakeGCE)
+
+	if outputDefaultHC.Description != translator.DescriptionForDefaultHealthChecks {
+		t.Fatalf("incorrect Description, is: \"%v\", want: \"%v\"",
+			outputDefaultHC.Description, translator.DescriptionForDefaultHealthChecks)
+	}
+
+	_, err = healthChecks.SyncServicePort(backendConfigSP, nil)
+	if err != nil {
+		t.Fatalf("unexpected err while syncing healthcheck, err %v", err)
+	}
+
+	outputBCHC := getSingletonHealthcheck(t, fakeGCE)
+
+	// Verify that BackendConfig overwrites only the specified values.
+	// In particular, with the flag set to false, the Description does not get overwritten.
+	outputDefaultHC.HttpHealthCheck.RequestPath = *backendConfigSP.BackendConfig.Spec.HealthCheck.RequestPath
+	if !reflect.DeepEqual(outputBCHC, outputDefaultHC) {
+		t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(outputBCHC), pretty.Sprint(outputDefaultHC))
+	}
+
+	// Modify the flag and see what happens.
+	flags.F.EnableUpdateCustomHealthCheckDescription = true
+
+	_, err = healthChecks.SyncServicePort(backendConfigSP, nil)
+	if err != nil {
+		t.Fatalf("unexpected err while syncing healthcheck, err %v", err)
+	}
+
+	outputBCHCWithFlag := getSingletonHealthcheck(t, fakeGCE)
+
+	if outputBCHCWithFlag.Description != translator.DescriptionForHealthChecksFromBackendConfig {
+		t.Fatalf("incorrect Description, is: \"%v\", want: \"%v\"",
+			outputBCHCWithFlag.Description, translator.DescriptionForHealthChecksFromBackendConfig)
+	}
+
+	// Verify that only the Description is modified after rollout.
+	filter := func(hc *compute.HealthCheck) {
+		hc.Description = ""
+	}
+	filter(outputBCHC)
+	filter(outputBCHCWithFlag)
+	if !reflect.DeepEqual(outputBCHC, outputBCHCWithFlag) {
+		t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(outputBCHC), pretty.Sprint(outputBCHCWithFlag))
+	}
+}
+
 func TestVersion(t *testing.T) {
 	t.Parallel()
 	testCases := []struct {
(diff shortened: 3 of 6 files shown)


DamianSawicki:
> Safeguarding the main change from https://github.com/kubernetes/ingress-gce/pull/2008 with a flag.

Actually, it may be a good idea to safeguard other changes as well: this way we can test the behaviour during feature rollout by modifying the flag during a test. -> Changing PR to draft.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +429,105 @@ func TestHealthCheckUpdate(t *testing.T) {
...
+	// Verify the health check exists
+	var computeHCs []*compute.HealthCheck
+	computeHCs, _ = fakeGCE.Compute().HealthChecks().List(context.Background(), nil)
+	if len(computeHCs) != 1 {
+		t.Fatalf("Got %d healthchecks, want 1\n%s", len(computeHCs), pretty.Sprint(computeHCs))
+	}
+	hc := computeHCs[0]
+
+	// Retrieve health check
+	outputDefaultHC, err := healthChecks.Get(hc.Name, meta.VersionGA, meta.Global)
```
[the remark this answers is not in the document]
DamianSawicki: The reason for this redundancy was that `Get` returns `translator.HealthCheck`, while `List` returns `compute.HealthCheck`s.

Done.

pkg/healthchecks/healthcheck.go
```diff
@@ -72,7 +73,7 @@ func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCh
 	if c.Port != nil && old.Port != new.Port {
 		changes.add("Port", strconv.FormatInt(old.Port, 10), strconv.FormatInt(new.Port, 10))
 	}
-	if old.Description != new.Description {
+	if flags.F.EnableRestoreHCDefaultsOnCustomHCRemoval && old.Description != new.Description {
```
kl52752, line 76: I think that you should store this flag asa a bool somewhere in controller instead of using it directly. It will be easier to test and we could run test in parallel without worried about flakes.

DamianSawicki: Done. As we discussed offline, the relevant test remains not parallelizable because of the pre-existing flag EnableBackendConfigHealthCheck.

pkg/controller/controller.go
```diff
@@ -106,15 +106,15 @@ type LoadBalancerController struct {
 // NewLoadBalancerController creates a controller for gce loadbalancers.
 func NewLoadBalancerController(
 	ctx *context.ControllerContext,
-	stopCh chan struct{}) *LoadBalancerController {
+	stopCh chan struct{}, restoreHCDefaultsOnCustomHCRemoval bool) *LoadBalancerController {
```
kl52752, line 109: move restoreHCDefaultsOnCustomHCRemoval to next line

pkg/healthchecks/healthcheck.go
```diff
@@ -35,8 +35,9 @@ func (c *fieldDiffs) add(field, oldv, newv string) {
 }
 func (c *fieldDiffs) String() string { return strings.Join(c.f, ", ") }
 func (c *fieldDiffs) hasDiff() bool  { return len(c.f) > 0 }
+func (c *fieldDiffs) size() int64    { return int64(len(c.f)) }
 
-func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCheckConfig) *fieldDiffs {
+func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCheckConfig, restoreHCDefaultsOnCustomHCRemoval bool) *fieldDiffs {
```
kl52752, line 40: you can shorten the name like "restoreDefaults"

pkg/healthchecks/healthchecks.go
```diff
@@ -42,14 +42,16 @@ type HealthChecks struct {
 	path string
 	// This is a workaround which allows us to not have to maintain
 	// a separate health checker for the default backend.
-	defaultBackendSvc types.NamespacedName
+	defaultBackendSvc                  types.NamespacedName
+	restoreHCDefaultsOnCustomHCRemoval bool
```
kl52752, line 46: add comment

pkg/healthchecks/healthchecks_test.go
```diff
@@ -149,7 +167,7 @@ func TestHealthCheckAdd(t *testing.T) {
 
 func TestHealthCheckAddExisting(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, flags.F.EnableRestoreHCDefaultsOnCustomHCRemoval)
```
kl52752, line 170: In fake health checks you should hardcoded this value so the test output will not depend on the flag value

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +447,98 @@ func TestHealthCheckUpdate(t *testing.T) {
 	}
 }
 
+// Test rollout of the flag EnableRestoreHCDefaultsOnCustomHCRemoval
+func TestFlagRolloutEnableRestoreHCDefaultsOnCustomHCRemoval(t *testing.T) {
```
kl52752, line 451: TestFlagRollout should tests the original flag and this test should be rename.
I think it might be beneficial to add test with original flag setup.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +447,98 @@ func TestHealthCheckUpdate(t *testing.T) {
...
+		backendConfigSP *utils.ServicePort = testSPs["HTTP-80-reg-bc"]
+	)
+
+	// Add Hooks
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockHealthChecks.UpdateHook = mock.UpdateHealthCheckHook
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaHealthChecks.UpdateHook = mock.UpdateAlphaHealthCheckHook
+	(fakeGCE.Compute().(*cloud.MockGCE)).MockBetaHealthChecks.UpdateHook = mock.UpdateBetaHealthCheckHook
+
+	// Start with restoreHCDefaultsOnCustomHCRemoval =  false.
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, false)
```
kl52752, line 472: healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, /*restoreHCDefaultsOnCustomHCRemoval = */ false)

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +447,98 @@ func TestHealthCheckUpdate(t *testing.T) {
...
+
+	_, err := healthChecks.SyncServicePort(defaultSP, nil)
+	if err != nil {
+		t.Fatalf("unexpected err while syncing healthcheck, err %v", err)
+	}
+
+	// Verify the health check exists
+	var computeHCs []*compute.HealthCheck
```
kl52752, line 480: Maybe it will be worth to extract some parts to function and reuse it?
This way comments will not be needed.

pkg/controller/controller.go
DamianSawicki: Done.

pkg/healthchecks/healthcheck.go
DamianSawicki: Done.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Set to `false` throughout.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: > TestFlagRollout should tests the original flag and this test should be rename.

Done.

> I think it might be beneficial to add test with original flag setup.

This is definitely a good idea, but may require more work, so perhaps it would make sense to merge this PR without it to have the [already merged feature](https://github.com/kubernetes/ingress-gce/pull/2008/) guarded with a flag defaulting to false, and work on extra tests in a separate PR.

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +447,94 @@ func TestHealthCheckUpdate(t *testing.T) {
...
+// This emulates the rollout of setting the flag EnableRestoreHCDefaultsOnCustomHCRemoval to true.
+func TestSwitchRestoreHCDefaults(t *testing.T) {
+	// No parallel() because we modify the value of the flag EnableBackendConfigHealthCheck.
+	oldEnableBC := flags.F.EnableBackendConfigHealthCheck
+	flags.F.EnableBackendConfigHealthCheck = true
+	defer func() {
+		flags.F.EnableBackendConfigHealthCheck = oldEnableBC
+	}()
+
+	retrieveHealthCheck := func(t *testing.T, c *gce.Cloud) *compute.HealthCheck {
```
kl52752, line 460: move this function outside the test case

pkg/healthchecks/healthchecks_test.go
kl52752: maybe you forgot to push this change?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -740,11 +861,14 @@ func TestCalculateDiff(t *testing.T) {
...
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
-			diffs := calculateDiff(tc.old, tc.new, tc.c)
+			diffs := calculateDiff(tc.old, tc.new, tc.c, tc.restoreHCDefaults)
 			t.Logf("\nold=%+v\nnew=%+v\ndiffs = %s", tc.old, tc.new, diffs)
 			if diffs.hasDiff() != tc.hasDiff {
 				t.Errorf("calculateDiff(%+v, %+v, %+v) = %s; hasDiff = %t, want %t", tc.old, tc.new, tc.c, diffs, diffs.hasDiff(), tc.hasDiff)
 			}
+			if tc.diffSize != nil && diffs.size() != *tc.diffSize {
+				t.Errorf("calculateDiff(%+v, %+v, %+v) = %s; size = %v, want %v", tc.old, tc.new, tc.c, diffs, diffs.size(), *tc.diffSize)
```
kl52752, line 870: add tc.restoreHCDefaults to function args

pkg/backends/integration_test.go
```diff
@@ -42,7 +43,7 @@ type Jig struct {
 }
 
 func newTestJig(fakeGCE *gce.Cloud) *Jig {
-	fakeHealthChecks := healthchecks.NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	fakeHealthChecks := healthchecks.NewHealthChecker(fakeGCE, "/", defaultBackendSvc, flags.F.EnableRestoreHCDefaultsOnCustomHCRemoval)
```
kl52752, line 46: do you need this flag for tests, maybe you can hardcode this value?

pkg/backends/syncer_test.go
```diff
@@ -137,7 +138,7 @@ var (
 )
 
 func newTestSyncer(fakeGCE *gce.Cloud) *backendSyncer {
-	fakeHealthChecks := healthchecks.NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	fakeHealthChecks := healthchecks.NewHealthChecker(fakeGCE, "/", defaultBackendSvc, flags.F.EnableRestoreHCDefaultsOnCustomHCRemoval)
```
kl52752, line 141: as above

pkg/healthchecks/healthchecks_test.go

pkg/backends/integration_test.go
DamianSawicki: Done.

pkg/backends/syncer_test.go
DamianSawicki: Done.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Oh, actually I applied your suggestion everywhere except for this specific line because here there's already a comment above.

pkg/healthchecks/healthcheck.go
```diff
@@ -72,7 +73,8 @@ func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCh
 	if c.Port != nil && old.Port != new.Port {
 		changes.add("Port", strconv.FormatInt(old.Port, 10), strconv.FormatInt(new.Port, 10))
 	}
-	if old.Description != new.Description {
+	if updateHCDescription && old.Description != new.Description {
```
bowei, line 76: My preference is to not violate the contract of calculateDiff() and keep it pure.

If we want to ignore the description, we should zero the field out before the call to diff.

Also, `updateHCDescription` is somewhat strange out of context of your change for this function.

pkg/healthchecks/healthchecks.go
```diff
@@ -43,13 +43,17 @@ type HealthChecks struct {
 	// This is a workaround which allows us to not have to maintain
 	// a separate health checker for the default backend.
 	defaultBackendSvc types.NamespacedName
+	// Flag for updating the health check Description when
+	// the health check is configured using a BackendConfig CRD.
+	updateHCDescription bool
```
bowei, line 48: See my comment about referring to the global flag directly.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -109,9 +110,26 @@ func init() {
 	}
 }
 
+func copyViaJSON(dst, src interface{}) error {
```
bowei, line 113: This seems to be cut and paste from somewhere?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -429,6 +447,94 @@ func TestHealthCheckUpdate(t *testing.T) {
 	}
 }
 
+func retrieveHealthCheck(t *testing.T, c *gce.Cloud) *compute.HealthCheck {
```
bowei, line 450: let's call this getSingletonHealthcheck()

pkg/healthchecks/healthcheck.go
DamianSawicki: Thanks for your insights, fixed.

cmd/glbc/main.go
```diff
@@ -271,7 +271,7 @@ func makeLeaderElectionConfig(ctx *ingctx.ControllerContext, client clientset.In
 func runControllers(ctx *ingctx.ControllerContext) {
 	stopCh := make(chan struct{})
 	ctx.Init()
-	lbc := controller.NewLoadBalancerController(ctx, stopCh)
+	lbc := controller.NewLoadBalancerController(ctx, stopCh, flags.F.EnableUpdateCustomHealthCheckDescription)
```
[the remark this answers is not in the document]
panslava: Sorry, I don't really agree with this

Storing flag globally makes it really harder to write tests, and run them in parallel -- you will need to override global variable for that, which will change behaviour of all parallel tests

If the reason is not to change a lot of public signature -- we could have struct params, so it is easier to extend them. But even for this, I don't see that much reason for now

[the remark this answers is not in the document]
panslava: We were storing some flags globally for RBS, and found out it was bad approach and changed it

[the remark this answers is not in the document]
DamianSawicki: For the record, both tests using the new flag (one new and one pre-existing) need to use the flag `EnableBackendConfigHealthCheck`, so they are not parallelizable anyway.

pkg/healthchecks/healthchecks.go
DamianSawicki: Already done, but see the arguments  from @panslava in the thread you are referring to.

cmd/glbc/main.go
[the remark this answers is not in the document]
DamianSawicki: I'm starting manual tests.

DamianSawicki:
Manual tests results as expected, ready to merge.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: No complaints about my fix -> resolving.

PR #2020, opened 2023-03-16
Add report utils for gke-diagnoser to support json output

songrx1997:
Add report structs for gke-diagnoser CLI
Add json utils for the output report

### DIFF
--- a/cmd/gke-diagnoser/app/util/report.go
+++ b/cmd/gke-diagnoser/app/util/report.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"encoding/json"
+)
+
+const (
+	// Passed is the constant value for the Result field of a passed check
+	Passed string = "PASSED"
+	// Failed is the constant value for the Result field of a failed check
+	Failed string = "FAILED"
+	// Skipped is the constant value for the Result field of a skipped check
+	Skipped string = "SKIPPED"
+)
+
+const (
+	//JSONOutput is the constant value for output type JSON
+	JSONOutput string = "json"
+)
+
+// Report represents the final output of the analyzer
+type Report struct {
+	Resources []*Resource `json:"resources"`
+}
+
+// Resource represents the a resource of the cluster and all the checks done on it
+type Resource struct {
+	Kind      string   `json:"kind"`
+	Namespace string   `json:"namespace"`
+	Name      string   `json:"name"`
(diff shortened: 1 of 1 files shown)


cmd/gke-diagnoser/app/util/report.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// Report represents the final output of the analyzer
+type Report struct {
+	Resource []*Resource `json:"ingress,omitempty"`
+}
+
+// Resource represents the a resource of the cluster and all the checks done on it
+type Resource struct {
+	Type      string   `json:"type"`
```
bowei, line 37: This is called Kind in Kubernetes not Type

cmd/gke-diagnoser/app/util/report.go
```diff
@@ -0,0 +1,64 @@
...
+	Namespace string   `json:"namespace"`
+	Name      string   `json:"name"`
+	Checks    []*Check `json:"checks,omitempty"`
+}
+
+// Check represents the result of a check
+type Check struct {
+	Name string `json:"name"`
+	Msg  string `json:"msg"`
+	Res  string `json:"res"`
```
bowei, line 47: Please don't use abbreviations unless there is a good reason. This makes it hard to read and understand.

Message

Resource

cmd/gke-diagnoser/app/util/report.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func JsonReport(report Report) string {
```
bowei, line 50: report should be a pointer (report *Report)

cmd/gke-diagnoser/app/util/report.go
```diff
@@ -0,0 +1,64 @@
...
+	jsonRaw, err := json.MarshalIndent(report, "", "  ")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error Marshalling JSON: %s\n", jsonRaw)
```
bowei, line 53: Is this how we are handling errors?

cmd/gke-diagnoser/app/util/report.go
```diff
@@ -0,0 +1,64 @@
...
+		os.Exit(1)
+	}
+	return string(jsonRaw)
+}
+
+// SupportedOutputs returns a string list of output formats supposed by this package
+func SupportedOutputs() []string {
```
bowei, line 60: How is this used?

cmd/gke-diagnoser/app/util/report.go
songrx1997: Updated, thanks!

cmd/gke-diagnoser/app/util/report.go
songrx1997: Updated, thanks!

cmd/gke-diagnoser/app/util/report.go
songrx1997: Updated, thanks!

cmd/gke-diagnoser/app/util/report.go
songrx1997: This will be used by a flag checker to check whether a format is supported.

cmd/gke-diagnoser/app/util/report.go
songrx1997: Changed this to return an error and let the caller handle it.

swetharepakula (commented):
/approve
/lgtm

swetharepakula:
/hold cancel

As the original comments have been addressed, removing the hold

PR #2025, opened 2023-03-17
Add CheckServiceExistence Rule for gke-ingress-checker

songrx1997:
Rename util package in gke-diagnoser to report
Add CheckServiceExistence rule for gke-ingress-checker
Add unit test for CheckServiceExistence rule

### DIFF
--- a/cmd/gke-diagnoser/app/ingress/rule.go
+++ b/cmd/gke-diagnoser/app/ingress/rule.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingress
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


bowei:
I thought we were going to rename the cmd name? The directory is still named gke-diagnoser. Let's not forget to fix this.

cmd/gke-diagnoser/app/ingress/rule.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+func CheckServiceExistence(ns string, svcName string, client clientset.Interface) (*corev1.Service, string, string) {
+	svc, err := client.CoreV1().Services(ns).Get(context.TODO(), svcName, metav1.GetOptions{})
+	if err != nil {
```
bowei, line 30: You need to check that the error is actually a not found error.
If there was a different kind of error, you need to surface that to the user and also print a different message.

songrx1997: Updated, thanks!

songrx1997:
> I thought we were going to rename the cmd name? The directory is still named gke-diagnoser. Let's not forget to fix this.

Will rename the directory after current PRs merged

cmd/gke-diagnoser/app/ingress/rule_test.go
```diff
@@ -0,0 +1,78 @@
...
+			name:      "service-1",
+			expect:    "PASSED",
+		},
+		{
+			desc:      "correct namespace and wrong name",
+			namespace: "namespace1",
+			name:      "service-2",
+			expect:    "FAILED",
+		},
+		{
```
swetharepakula, line 60: duplicate testcase

songrx1997: Removed, thanks!

swetharepakula (commented):
/lgtm
/approve

/hold cancel
The rename will be handled in a separate PR

PR #2027, opened 2023-03-19
Fix get node using GetByKey instead

sawsa307:
Include Namespace in node key.


pkg/neg/syncers/utils.go
```diff
@@ -385,7 +385,7 @@ func validatePod(pod *apiv1.Pod, nodeLister cache.Indexer) bool {
 		klog.V(2).Info("Pod %s/%s is a terminal pod with status %v, skipping", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name, phase)
 		return false
 	}
-	obj, exists, err := nodeLister.GetByKey(pod.Spec.NodeName)
+	obj, exists, err := nodeLister.GetByKey(fmt.Sprintf("%s/%s", pod.ObjectMeta.Namespace, pod.Spec.NodeName))
```
bowei, line 388: Nodes don't have namespaces. What is going on here?

PR #2029, opened 2023-03-21
run presubmits with custom image

aojea:
See if we can detect problems before merging


gauravkghildiyal:
@aojea This is a longshot and I haven't tried it out but I believe we need to add two spaces for each line in `CUSTOM_INGRESS_YAML`

The reason why I believe this might work is because one of the places where this is consumed is [here](https://github.com/kubernetes/kubernetes/blob/45b96eae989efef9380fadc1e46957dc6174f351/cluster/gce/util.sh#L1179-L1180)
```yaml
CUSTOM_INGRESS_YAML: |
${CUSTOM_INGRESS_YAML//\'/\'\'}
```

So without those two spaces, the above piece of yaml would be incorrect and the above variable would not get propagated properly.

aojea:
/test pull-ingress-gce-e2e

@gauravkghildiyal why don't you give it a shot? I closed your PR, but you can reopen or create a new one

gauravkghildiyal:
@aojea Yes I'll give it a shot if this does not work (hopefully I won't have to ask again for ok-to-test label if I reopen the old PR)

gauravkghildiyal:
Good I think that worked.

Just that now the image seems incorrect right?

### Correct image tag:
image

### What we seem to be using:
image

gauravkghildiyal:
This is weird. Image `gcr.io/kubernetes-ingress/glbc:228fe2418-20230323T004249` also exists.

aojea:
ok, finally this went through https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/ingress-gce/2029/pull-ingress-gce-e2e/1638838371042201600

aojea:
/test pull-ingress-gce-e2e


let's see if the image we are building is incorrect, it passed with the nightly one

aojea:
ok, confirmedv the image built with ko is not ok

aojea:
/test pull-ingress-gce-e2e
--


<br class="Apple-interchange-newline">

aojea:
/test pull-ingress-gce-e2e


this should be the good one

aojea:
the e2e teakes too much, I will try to reduce the list

gauravkghildiyal:
Looks good to me
(don't have membership yet for an actual LGTM)

aojea:
it works and runs in 23 minutes

image

bowei:
/lgtm
/approve

Thanks Antonio!

hack/run-e2e-gce.sh
```diff
@@ -58,22 +53,113 @@ if [[ -z "${IMAGE_TAG:-}" ]]; then
   IMAGE_TAG=$(git rev-parse --short HEAD)-$(date +%Y%m%dT%H%M%S)
 fi
 
-export GCE_GLBC_IMAGE=$(go run github.com/google/ko@v0.12.0 build --tags ${IMAGE_TAG} --base-import-paths --push=true ./cmd/glbc/)
+KO_GLBC_IMAGE=$(go run github.com/google/ko@v0.12.0 build --tags ${IMAGE_TAG} --base-import-paths --push=true ./cmd/glbc/)
```
BenTheElder, line 56: so this pushes to a registry? what GCs these?

ameukam: This is pushed to the GCR registry in the boskos project. you can see the GCS for GCR created in https://github.com/kubernetes/k8s.io/pull/5015/files.

ameukam: I think we should migrate to regional ARs for each boskos project.

aojea: pushes to boskos created project, I copied from @justinsb , that is why I asked, I think this is eventually garbage collected

PR #2030, opened 2023-03-21
test pull-ingress-gce-e2e

gauravkghildiyal:
This is just a test pull request to investigate failing jobs.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	_ "sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/glbc/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	_ "sync"
```
aojea, line 25: this should not do much, should it?
I wonder if this is a permissions problem?

aojea:
to trigger the job you just add a coment like this

/test pull-ingress-gce-e2e

aojea:
/close

it was fixed by removing the endpointslice flag from the job

PR #2031, opened 2023-03-22
Add EndpointAnnotationMap to NEG syncer and update tests

songrx1997:
Add EndpointAnnotationMap
Update function signatures for EndpointAnnotationMap
Update unit tests related to EndpointAnnotationMap

### DIFF
--- a/pkg/neg/syncers/endpoints_calculator_test.go
+++ b/pkg/neg/syncers/endpoints_calculator_test.go
@@ -106,7 +106,7 @@ func TestLocalGetEndpointSet(t *testing.T) {
 	ec := NewLocalL4ILBEndpointsCalculator(nodeLister, zoneGetter, svcKey, klog.TODO())
 	for _, tc := range testCases {
 		createNodes(t, tc.nodeNames, tc.nodeLabelsMap, tc.nodeReadyStatusMap, transactionSyncer.nodeLister)
-		retSet, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
+		retSet, _, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
 		if err != nil {
 			t.Errorf("For case %q, expect nil error, but got %v.", tc.desc, err)
 		}
@@ -198,7 +198,7 @@ func TestClusterGetEndpointSet(t *testing.T) {
 	ec := NewClusterL4ILBEndpointsCalculator(nodeLister, zoneGetter, svcKey, klog.TODO())
 	for _, tc := range testCases {
 		createNodes(t, tc.nodeNames, tc.nodeLabelsMap, tc.nodeReadyStatusMap, transactionSyncer.nodeLister)
-		retSet, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
+		retSet, _, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
 		if err != nil {
 			t.Errorf("For case %q, expect nil error, but got %v.", tc.desc, err)
 		}
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -218,13 +218,14 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 }
 
 // toZoneNetworkEndpointMap translates addresses in endpoints object into zone and endpoints map, and also return the count for duplicated endpoints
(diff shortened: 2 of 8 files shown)


songrx1997:
CC @gauravkghildiyal

pkg/neg/types/types.go
```diff
@@ -375,3 +375,6 @@ type Label struct {
 	ShortKey          string
 	MaxLabelSizeBytes int
 }
+
+// EndpointAnnotationMap is a map from network endpoint to endpoint annotations.
```
bowei, line 379: Use the types as your documentation

// EndpointLabelMap is a map of key, label values.
type EndpointLabelMap map[string]string

// AnnotationsMap is a map of NetworkEndpoint to EndpointLabelMaps.

pkg/neg/types/interfaces.go
```diff
@@ -83,7 +83,7 @@ type NegSyncerManager interface {
 type NetworkEndpointsCalculator interface {
 	// CalculateEndpoints computes the NEG endpoints based on service endpoints and the current NEG state and returns a
 	// map of zone name to network endpoint set
-	CalculateEndpoints(eds []EndpointsData, currentMap map[string]NetworkEndpointSet) (map[string]NetworkEndpointSet, EndpointPodMap, int, error)
+	CalculateEndpoints(eds []EndpointsData, currentMap map[string]NetworkEndpointSet) (map[string]NetworkEndpointSet, EndpointPodMap, EndpointAnnotationMap, int, error)
```
bowei, line 86: If you find a functions return values are getting long, please don't just add to the issue and make it worse.

Create a struct for the return value.

```
type CalculateEndpointsResult struct {
  // EndpointSet is a map of zone name to endpoint set.
  EndpointSet map[string]NetworkEndpointSet

  // EndpointsToPod is a map from network endpoint to the Pod (ns, name).
  EndpointsToPod EndpointPodMap

  // what is the int return value?
}
```

pkg/neg/syncers/utils.go
```diff
@@ -439,7 +440,7 @@ func retrieveExistingZoneNetworkEndpointMap(negName string, zoneGetter negtypes.
 
 // makeEndpointBatch return a batch of endpoint from the input and remove the endpoints from input set
 // The return map has the encoded endpoint as key and GCE network endpoint object as value
-func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.NetworkEndpointType) (map[negtypes.NetworkEndpoint]*composite.NetworkEndpoint, error) {
+func makeEndpointBatch(endpoints negtypes.NetworkEndpointSet, negType negtypes.NetworkEndpointType, endpointAnnotationMap negtypes.EndpointAnnotationMap) (map[negtypes.NetworkEndpoint]*composite.NetworkEndpoint, error) {
```
bowei, line 443: Try to order the func arguments so they make sense intuitively.

Composite fields like maps etc should be before singular field like negType in this case.

bowei: Not sure why we are adding the parameter but never using it. Is this PR complete?

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -106,7 +106,7 @@ func TestLocalGetEndpointSet(t *testing.T) {
 	ec := NewLocalL4ILBEndpointsCalculator(nodeLister, zoneGetter, svcKey, klog.TODO())
 	for _, tc := range testCases {
 		createNodes(t, tc.nodeNames, tc.nodeLabelsMap, tc.nodeReadyStatusMap, transactionSyncer.nodeLister)
-		retSet, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
+		retSet, _, _, _, err := ec.CalculateEndpoints(tc.endpointsData, nil)
```
bowei, line 109: This looks like we are not testing the added functionality and ignore the returned value?

pkg/neg/syncers/utils.go
```diff
@@ -218,13 +218,14 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 }
 
 // toZoneNetworkEndpointMap translates addresses in endpoints object into zone and endpoints map, and also return the count for duplicated endpoints
-func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, lpConfig negtypes.PodLabelPropagationConfig) (map[string]negtypes.NetworkEndpointSet, negtypes.EndpointPodMap, int, error) {
+func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.ZoneGetter, servicePortName string, networkEndpointType negtypes.NetworkEndpointType, lpConfig negtypes.PodLabelPropagationConfig) (map[string]negtypes.NetworkEndpointSet, negtypes.EndpointPodMap, negtypes.EndpointAnnotationMap, int, error) {
```
bowei, line 221: See comment above. A function that returns a ton of values is a sign that some refactoring needs to be done. Please see my comment about the struct.

PR #1991, opened 2023-03-09
Bump k8s-cloud-provider to v1.22.0

spencerhance:
This is needed to pull in https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/79

PR #2033, opened 2023-03-22
Replace legacy-cloud-providers/gce with cloud-provider-gcp

spencerhance:
We should do this migration anyway since legacy-cloud-providers is deprecated in favor of cloud-provider-gcp.

This should be a no-op but will allow us to unblock https://github.com/kubernetes/ingress-gce/pull/1991


aojea:
/ok-to-test
/lgtm

@bowei for final approval

spencerhance:
Thanks!  Just updated with the gofmt fixes

PR #2036, opened 2023-03-23
Add Observe() func to internal GCERateLimiter

spencerhance:
It appears that there's yet another rate limiter that has to be updated due to https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/77

In my testing, this is the last patch needed to build ingress with the newest k8s-cloud-provider

/assign @aojea

### DIFF
--- a/pkg/ratelimit/ratelimit.go
+++ b/pkg/ratelimit/ratelimit.go
@@ -110,6 +110,9 @@ func (grl *GCERateLimiter) Accept(ctx context.Context, key *cloud.RateLimitKey) 
 	return err
 }
 
+// Observe is a no-op func to satisfy cloud.RateLimiter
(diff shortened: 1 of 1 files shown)


pkg/ratelimit/ratelimit.go
```diff
@@ -110,6 +110,9 @@ func (grl *GCERateLimiter) Accept(ctx context.Context, key *cloud.RateLimitKey)
 	return err
 }
 
+// Observe is a no-op func to satisfy cloud.RateLimiter
+func (*GCERateLimiter) Observe(context.Context, error, *cloud.RateLimitKey) {}
```
aojea, line 114:
```suggestion
func (grl *GCERateLimiter) Observe(context.Context, error, *cloud.RateLimitKey) {}
```

should we or is too much notpick?

PR #2037, opened 2023-03-24
Change e2e-test.yaml to support ARM

code-elinka:
I tried to run e2e tests on ARM and it does not work because I left this file unchanged in previous PRs


gauravkghildiyal:
Looks good

aojea:
where do we run these tests?
internal ci?

PR #2038, opened 2023-03-24
Set the k8s-cloud-provider API domain

spencerhance:
* Parse the GA basePath (which comes from gce.conf) to set the domain
* This is useful for custom on-prem environments


pkg/utils/utils_test.go
```diff
@@ -1535,3 +1535,43 @@ func TestAddIPToLBStatus(t *testing.T) {
...
+			basePath: "https://www.compute.googleapis.com/compute/v1",
+			want:     "https://www.compute.googleapis.com",
+		},
+		{
+			desc:     "arbitary path",
+			basePath: "mycompute.mydomain.com/mypath/compute/v1",
+			want:     "mycompute.mydomain.com/mypath",
+		},
+		{
+			desc:     "arbitary path with trailing /",
```
bowei, line 1559: spelling arbitary

pkg/utils/utils.go
```diff
@@ -868,3 +868,12 @@ func AddIPToLBStatus(status *api_v1.LoadBalancerStatus, ips ...string) *api_v1.L
 	}
 	return status
 }
+
+// GetDomainFromGABasePath takes a GA base path of the form <path>/compute/v1 and returns the path.
+func GetDomainFromGABasePath(basePath string) string {
```
bowei, line 873: it's unfortunate that we have to do this just because the code is in package main.

PR #2039, opened 2023-03-25
Update error state set and reset

sawsa307:
Create NegSyncError error type. Errors that need to be recorded in neg sync metrics or errors that used to set error state should be wrapped in NegSyncError.


sawsa307:
cc @gauravkghildiyal

pkg/neg/controller.go
```diff
@@ -128,6 +128,7 @@ func NewController(
 	runL4Controller bool,
 	enableNonGcpMode bool,
 	enableAsm bool,
+	enableDegradedMode bool,
```
bowei, line 131: can we just refer to the global flag directly? It avoids messing with the public interface...

sawsa307: Updated. Now the controller directly refer the global flag, and the manager would save the value for syncer creations. Thanks!

sawsa307:
cc @songrx1997

pkg/neg/controller.go
```diff
@@ -168,6 +168,7 @@ func NewController(
 		svcNegInformer.GetIndexer(),
 		syncerMetrics,
 		enableNonGcpMode,
+		flags.F.EnableDegradedMode,
```
bowei, line 171: Do we need to expose this flag everywhere? why not just refer to the global directly? I think I've put this comment on your PR before?

sawsa307: I leave it as a field in manager.go since we use this flag every time when we create a transactionSyncer. I'll can remove this field completely in manager and transactionSyncer, and only use global flag when needed.

bowei: Do we expect the value to change after startup? Why do we need to cache a copy of it in other places in the code?

bowei: Do we expect the value to change after startup? Why do we need to cache a copy of it in other places in the code?

bowei: Do we expect the value to change after startup? Why do we need to cache a copy of it in other places in the code?

bowei:
If you have changes that are just refactoring or reformatting, I suggest sending those as changes (e.g. PRs) by themselves so they are very easy to approve.

pkg/neg/controller.go
sawsa307: You're right, the value won't change. I just get rid of storing this value.

pkg/neg/syncers/transaction.go
```diff
@@ -376,43 +393,36 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
-	if _, ok := err.(utilerrors.Aggregate); ok {
-		return ""
-	}
-	if result, contains := negtypes.ErrorStateResult[err]; contains {
-		return string(result)
+// parseEndpointBatchErr checks the error from endpoint batch response
+// and determine if the syncer is in error state
+// it returns true and an error-state error if the syncer should be in error state
+// otherwise, it returns false and the error unmodified
+func parseEndpointBatchErr(err error, operation transactionOp) error {
```
sawsa307, line 400: do checkEndpointBatchErr()

pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -79,15 +79,15 @@ func (sm *SyncerMetrics) Run(stopCh <-chan struct{}) {
 func (sm *SyncerMetrics) export() {
 }
 
-// UpdateSyncer update the status of corresponding syncer based on the syncResult.
-func (sm *SyncerMetrics) UpdateSyncer(key negtypes.NegSyncerKey, syncResult *negtypes.NegSyncResult) {
+func (sm *SyncerMetrics) UpdateSyncerStatusInMetrics(key negtypes.NegSyncerKey, err error) {
```
swetharepakula, line 82: add back the comment

pkg/neg/syncers/transaction.go
```diff
@@ -321,18 +321,35 @@ func (s *transactionSyncer) getEndpointsCalculation(
 	return targetMap, endpointPodMap, nil
 }
 
+// updateErrorState checks if the error is an error-state error and update error state
+func (s *transactionSyncer) updateErrorState(err error) {
```
swetharepakula, line 325: as discussed offline, this may be better as a utility function to get the classification of the error. So perhaps rename to 
`classifyError(err error) negtypes.NegSyncError`. 

Then using that output, we can separately update the error state as well as use the same error to publish error metrics.

pkg/neg/syncers/transaction.go
swetharepakula: nit: move this function closer to where it will be used

pkg/neg/syncers/transaction.go
```diff
@@ -376,43 +393,36 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
 	apiErr, ok := err.(*googleapi.Error)
 	if !ok {
-		s.logger.Info("Detected error when parsing batch response error", "operation", operation, "error", err)
 		return negtypes.ErrInvalidAPIResponse
 	}
 	errCode := apiErr.Code
 	if errCode == http.StatusBadRequest {
-		s.logger.Info("Detected error when sending endpoint batch information", "operation", operation, "errorCode", errCode)
 		if operation == attachOp {
 			return negtypes.ErrInvalidEPAttach
```
swetharepakula, line 411: since now this is the error that will be bubbled up in the event, we should not lose the context of the original error which likely will give us information such as which ip failed etc and for what reason.

pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -106,3 +106,14 @@ func (sm *SyncerMetrics) SetSyncerEPMetrics(key negtypes.NegSyncerKey, endpointS
 	}
 	sm.syncerEPSStateMap[key] = endpointStat.EndpointSliceStateCount
 }
+
+func getReason(err error) negtypes.Reason {
```
swetharepakula, line 110: with the convert or classifyError function, this function might become redundant. There is likely some consolidation that can happen here.

pkg/neg/syncers/transaction.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
```diff
@@ -101,10 +102,10 @@ type transactionSyncer struct {
 	// 3. Attach/Detach EP fails due to incorrect batch information
 	// 4. Endpoint count from EPS or calculated endpoint list is 0
 	// Need to grab syncLock first for any reads or writes based on this value
-	errorState string
+	isErrorState bool
```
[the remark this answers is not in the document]
sawsa307: Updated. Thanks!

pkg/neg/metrics/neg_metrics_collector.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
sawsa307: Updated. Thanks!

pkg/neg/metrics/neg_metrics_collector.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go

pkg/neg/syncers/transaction.go
```diff
@@ -490,12 +461,16 @@ func (s *transactionSyncer) operationInternal(operation transactionOp, zone stri
...
-		endpointBatchErr := s.ValidateEndpointBatch(err, operation)
-		if endpointBatchErr != nil {
+		err := checkEndpointBatchErr(err, operation)
+		s.syncMetricsCollector.UpdateSyncerStatusInMetrics(s.NegSyncerKey, err)
+		syncErr := negtypes.ClassifyError(err)
+		if syncErr != nil && syncErr.IsErrorState {
+			s.logger.V(3).Info("Updating error state", "error state", syncErr.Reason)
 			s.syncLock.Lock()
-			s.setErrorState(s.getErrorStateReason(endpointBatchErr))
+			s.setErrorState()
```
swetharepakula, line 473: do we have a unit test for this logic? especially when we have multiple operationInternal (1 attach, 1 detach)?

pkg/neg/syncers/transaction.go
```diff
@@ -490,12 +461,16 @@ func (s *transactionSyncer) operationInternal(operation transactionOp, zone stri
 
 	if err == nil {
 		s.recordEvent(apiv1.EventTypeNormal, operation.String(), fmt.Sprintf("%s %d network endpoint(s) (NEG %q in zone %q)", operation.String(), len(networkEndpointMap), s.NegSyncerKey.NegName, zone))
+		s.syncMetricsCollector.UpdateSyncerStatusInMetrics(s.NegSyncerKey, nil)
 	} else {
 		s.recordEvent(apiv1.EventTypeWarning, operation.String()+"Failed", fmt.Sprintf("Failed to %s %d network endpoint(s) (NEG %q in zone %q): %v", operation.String(), len(networkEndpointMap), s.NegSyncerKey.NegName, zone, err))
-		endpointBatchErr := s.ValidateEndpointBatch(err, operation)
-		if endpointBatchErr != nil {
+		err := checkEndpointBatchErr(err, operation)
```
swetharepakula, line 467: can you add a comment about why this will work with multiple operationInternal? As in only the error ones will try to set the error, versus a successful run will not touch the error state.

pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -79,15 +81,22 @@ func (sm *SyncerMetrics) Run(stopCh <-chan struct{}) {
...
 
-// UpdateSyncer update the status of corresponding syncer based on the syncResult.
-func (sm *SyncerMetrics) UpdateSyncer(key negtypes.NegSyncerKey, syncResult *negtypes.NegSyncResult) {
+func (sm *SyncerMetrics) UpdateSyncerStatusInMetrics(key negtypes.NegSyncerKey, err error) {
+	reason := negtypes.ReasonSuccess
+	if err != nil {
+		if syncErr := negtypes.ClassifyError(err); syncErr != nil {
+			reason = syncErr.Reason
+		} else {
+			reason = negtypes.ReasonOtherError // if the error is not an NegSyncError, it means we are not tracking its reason
```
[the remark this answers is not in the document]
sawsa307: Yes, I can make classify error to always return a *NegSyncError. If the given error is type NegSyncError, we would return it by unwrapping it. Otherwise, we would create a new NegSyncError with ReasonOtherError.

pkg/neg/syncers/transaction.go
```diff
@@ -226,15 +230,15 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 	if s.needInit || s.isZoneChange() {
 		if err := s.ensureNetworkEndpointGroups(); err != nil {
-			return err
+			return fmt.Errorf("%v: %w", err, negtypes.ErrNegNotFound)
```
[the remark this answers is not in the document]
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
```diff
@@ -326,18 +330,18 @@ func (s *transactionSyncer) getEndpointsCalculation(
 }
 
 // syncLock must already be acquired before execution
-func (s *transactionSyncer) inErrorState() bool {
-	return s.errorState != ""
+func (s *transactionSyncer) isErrorState() bool {
```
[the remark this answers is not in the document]
sawsa307: The error state indicator is inErrorState as a boolean, so we can have them both with the same name.

pkg/neg/syncers/transaction.go
```diff
@@ -490,12 +461,16 @@ func (s *transactionSyncer) operationInternal(operation transactionOp, zone stri
 
 	if err == nil {
 		s.recordEvent(apiv1.EventTypeNormal, operation.String(), fmt.Sprintf("%s %d network endpoint(s) (NEG %q in zone %q)", operation.String(), len(networkEndpointMap), s.NegSyncerKey.NegName, zone))
+		s.syncMetricsCollector.UpdateSyncerStatusInMetrics(s.NegSyncerKey, nil)
 	} else {
 		s.recordEvent(apiv1.EventTypeWarning, operation.String()+"Failed", fmt.Sprintf("Failed to %s %d network endpoint(s) (NEG %q in zone %q): %v", operation.String(), len(networkEndpointMap), s.NegSyncerKey.NegName, zone, err))
-		endpointBatchErr := s.ValidateEndpointBatch(err, operation)
-		if endpointBatchErr != nil {
+		err := checkEndpointBatchErr(err, operation)
+		s.syncMetricsCollector.UpdateSyncerStatusInMetrics(s.NegSyncerKey, err)
```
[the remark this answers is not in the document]
sawsa307: Sounds good. Let me change that. Thank you!

pkg/neg/syncers/transaction.go
sawsa307: No we don't. I'll try to add one.

swetharepakula (commented):
mostly small comments left.

pkg/neg/syncers/transaction.go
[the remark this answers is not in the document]
sawsa307: we cannot have field and function with the same name. It would raise an error

pkg/neg/syncers/utils.go
```diff
@@ -342,21 +342,23 @@ func toZoneNetworkEndpointMap(eds []negtypes.EndpointsData, zoneGetter negtypes.
...
 					return ZoneNetworkEndpointMapResult{
-						NetworkEndpointSet: nil,
-						EndpointPodMap:     nil,
-						DupCount:           dupCount,
-					}, getPodErr // when degraded mode is enabled, we want to trigger degraded mode so return the error
+							NetworkEndpointSet: nil,
+							EndpointPodMap:     nil,
+							DupCount:           dupCount,
+						}, // when degraded mode is enabled, we want to trigger degraded mode so return the error
+						fmt.Errorf("%w: Unexpected error when getting pod for Endpoint %q in Endpoints %s/%s", getPodErr, endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name)
```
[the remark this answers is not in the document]
sawsa307: What is the reason we are doing this differently from https://github.com/kubernetes/ingress-gce/pull/2039#discussion_r1165821304?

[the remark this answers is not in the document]
sawsa307: > It would actually be good if we can get some more information from the zone error itself such as which zone was not found or which node name.

We need nodeName to get zone, which an endpoint might not have one, and we won't be able to add zone until we get it from zoneGetter. I can try to add the error received from zoneGetter if we need more context.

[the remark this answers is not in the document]
swetharepakula: Errors should always look like `new/ wrapped Msg: original error`. In all the other cases, the original error is coming after the colon. However in this case we are doing `original error: new msg`.

For the node name and zone related errors, we should included those in the error that is returned from the getZone function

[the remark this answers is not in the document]
sawsa307: Updated both. Thank you!

pkg/neg/syncers/transaction.go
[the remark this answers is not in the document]
swetharepakula: sorry, that is correct. Lets rename the variable to errorState instead and leave the function as inErrorState.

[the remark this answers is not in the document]
sawsa307: Got it. Updated.

pkg/neg/syncers/utils.go
```diff
@@ -413,10 +415,10 @@ func getEndpointZone(
 	}
 	zone, err := zoneGetter.GetZoneForNode(*endpointAddress.NodeName)
 	if err != nil {
-		return zone, negtypes.ErrEPNodeNotFound
+		return zone, fmt.Errorf("%w: %v", negtypes.ErrEPNodeNotFound, err)
 	}
 	if zone == "" {
-		return zone, negtypes.ErrEPZoneMissing
+		return zone, fmt.Errorf("%w: zone=%v", negtypes.ErrEPZoneMissing, zone)
```
swetharepakula, line 421: zone will always be "" in this case.
it would be better to say which node the zone was missing from.

sawsa307: Updated!

swetharepakula (commented):
/lgtm
/approve

PR #2041, opened 2023-03-27
don't log all nodenames

aojea:
This is a problem at large scale, it also make the browser to crash when accessing the log entries


code-elinka:
Thank you! I did not find anything else in `Fatalf` or `Warningf` etc
/lgtm
/approve

PR #2043, opened 2023-03-27
Use Patch when syncing GCP BackendService. Linking backend groups still uses Update

gauravkghildiyal:
/assign @swetharepakula


pkg/backends/syncer.go
```diff
@@ -305,6 +305,9 @@ func ensureProtocol(be *composite.BackendService, p utils.ServicePort) (needsUpd
 	if be.Protocol == string(p.Protocol) {
 		return false
 	}
+
+	fmt.Printf("protocol needs change; got be = %+v, want protocol = %v\n", be, p.Protocol)
```
swetharepakula, line 309: Should this be a log line instead?

gauravkghildiyal: Deleted. That log line shouldn't exist. I'm not really sure how it got there. I don't remember adding it.

swetharepakula (commented):
/lgtm
/approve

PR #2045, opened 2023-03-29
Refactor code related to PodLabelPropagationConfig

songrx1997:
1. Remove `PodLabelPropagationConfig` from `EndpointsCalculator` related code since we will no longer calculate label propagation in `EndpointsCalculator`.
2.  Add `PodLabelPropagationConfig` to NEG `TransactionSyncer`

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -161,7 +161,7 @@ func NewTransactionSyncer(
 	return syncer
 }
 
-func GetEndpointsCalculator(nodeLister, podLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger, lpConfig negtypes.PodLabelPropagationConfig) negtypes.NetworkEndpointsCalculator {
+func GetEndpointsCalculator(nodeLister, podLister cache.Indexer, zoneGetter negtypes.ZoneGetter, syncerKey negtypes.NegSyncerKey, mode negtypes.EndpointsCalculatorMode, logger klog.Logger) negtypes.NetworkEndpointsCalculator {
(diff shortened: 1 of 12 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -108,6 +108,9 @@ type transactionSyncer struct {
 
 	// enableDegradedMode indicates whether we do endpoint calculation using degraded mode procedures
 	enableDegradedMode bool
+
+	// podLabelPropagationConfig configures the pod label to be propagated to NEG endpoints
+	podLabelPropagationConfig negtypes.PodLabelPropagationConfig
```
bowei, line 113: Is there a reason why we need to put all the types (e.g. `PodLabelPropagationConfig`) into the `types` package? this this an anti-pattern unless there is an explicit circular dependency in the packages that you are trying to avoid.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1432,12 +1432,13 @@ func newTestTransactionSyncer(fakeGCE negtypes.NetworkEndpointGroupCloud, negTyp
 		testContext.SvcNegInformer.GetIndexer(),
 		reflector,
 		GetEndpointsCalculator(testContext.NodeInformer.GetIndexer(), testContext.PodInformer.GetIndexer(), fakeZoneGetter,
-			svcPort, mode, klog.TODO(), negtypes.PodLabelPropagationConfig{}),
+			svcPort, mode, klog.TODO()),
```
bowei, line 1435: Can you remove this newline? It's somewhat random.

pkg/neg/syncers/transaction.go
bowei: Suggest https://github.com/bowei/ingress-gce/commit/601803a11e74d0e3cc5ccd57f5aa1c15f6ebfdf0

songrx1997: Added labels package and moved label propagation related types to it.

songrx1997:
cc @gauravkghildiyal

PR #2047, opened 2023-03-29
Update k8s-cloud-provider to v1.23.0 for new IPv6 field in NEG API

gauravkghildiyal:


vendor/cloud.google.com/go/compute/metadata/metadata.go
```diff
@@ -70,7 +70,9 @@ func newDefaultHTTPClient() *http.Client {
 				Timeout:   2 * time.Second,
 				KeepAlive: 30 * time.Second,
 			}).Dial,
+			IdleConnTimeout: 60 * time.Second,
 		},
+		Timeout: 5 * time.Second,
```
aojea, line 75: this is an important change, 5 seconds is ok for all APIs?

aojea: oh, is only metadata, right?
I assume that is ok then

aojea:
/lgtm
/approve

Thanks

PR #2049, opened 2023-03-30
Degraded mode correctness metrics

sawsa307:
Add metrics that reports the number of endpoints differ between current endpoint calculation and degraded mode calculation. It is not emitted when current endpoint calculation returns any errors because the returned map will be empty, while in degraded mode calculation, errors are handled and a non-empty map is returned.

### DIFF
--- a/pkg/neg/metrics/metrics.go
+++ b/pkg/neg/metrics/metrics.go
@@ -24,21 +24,25 @@ import (
 )
 
 const (
-	negControllerSubsystem   = "neg_controller"
-	syncerLatencyKey         = "syncer_sync_duration_seconds"
-	managerProcessLatencyKey = "manager_process_duration_seconds"
-	initLatencyKey           = "neg_initialization_duration_seconds"
-	negOpLatencyKey          = "neg_operation_duration_seconds"
-	negOpEndpointsKey        = "neg_operation_endpoints"
-	lastSyncTimestampKey     = "sync_timestamp"
-	syncerStalenessKey       = "syncer_staleness"
-	epsStalenessKey          = "endpointslice_staleness"
+	negControllerSubsystem     = "neg_controller"
+	syncerLatencyKey           = "syncer_sync_duration_seconds"
+	managerProcessLatencyKey   = "manager_process_duration_seconds"
+	initLatencyKey             = "neg_initialization_duration_seconds"
+	negOpLatencyKey            = "neg_operation_duration_seconds"
+	negOpEndpointsKey          = "neg_operation_endpoints"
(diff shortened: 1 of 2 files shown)


sawsa307:
cc @gauravkghildiyal

pkg/neg/metrics/metrics.go
```diff
@@ -149,6 +158,17 @@ var (
 			Buckets: prometheus.ExponentialBuckets(1, 2, 14),
 		},
 	)
+
+	DegradeModeCorrectness = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      degradedModeCorrectnessKey,
+			Help:      "Number of endpoints differed between current endpoint calculation and degraded mode calculation",
+			// custom buckets - [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, +Inf]
```
bowei, line 167: Let's make this more fine grained. Can we in increase +Inf range to 20k endpoints, set number of buckets to 20.

pkg/neg/syncers/transaction.go
```diff
@@ -280,8 +280,9 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			if len(notInDegraded) == 0 && len(onlyInDegraded) == 0 {
 				s.resetErrorState()
 			}
+		} else {
+			computeDegradedModeCorrectness(notInDegraded, onlyInDegraded, string(s.NegSyncerKey.NegType))
```
bowei, line 284: Should we always publish instead of only publish when > 0?

As a histogram, each data point corresponds to ? What is the units/denominator of the metric?

sawsa307: > Should we always publish instead of only publish when > 0?
We are trying to collect metrics when the normal calculation doesn't run into error/yields non-empty result, so we can make a meaningful comparison. 
We end up in this line with both=0. When we are not in error state, and normal calculation doesn't run into error and produce endpoints, and it is possible that it is the same as the degraded mode result. 
We are missing one condition, when we are in error state and the normal calculation doesn't run into error, so I'll change the condition here.

> As a histogram, each data point corresponds to ? What is the units/denominator of the metric?
Each data point corresponds to the difference between normal and degraded mode endpoint calculations WHEN the normal calculation doesn't run into error. The unit is #endpoint, and the denominator is total count of endpoints. We are not recording the denominator here because all we care is if the difference is 0.

pkg/neg/metrics/metrics.go
sawsa307: Updated. Thanks!

PR #2050, opened 2023-03-30
refactor: keep attach/detch functions synchronous and move async behaviour to caller

gauravkghildiyal:
Caller side concurrency is generally better.

/assign @swetharepakula 
/cc @sawsa307 
/cc @songrx1997


gauravkghildiyal:
I'll send this out in the future depending on what the code looks then

PR #2052, opened 2023-03-31
Update set error state for CalculateEndpoints()

sawsa307:
Remove set error state from calculateEndpoints and instead set it during syncInternal, otherwise, the error state will be overwritten and error state won't be found.


swetharepakula (commented):
/approve
/lgtm

PR #2053, opened 2023-03-31
[Cherry pick #2051] Fix syncLock deadlock

sawsa307:
#2051
Deferring lock create deadlock with commitTransaction.


swetharepakula (commented):
/lgtm
/approve

PR #2056, opened 2023-03-31
[Cherry-pick #1991, #2043 to 1.22] Use Patch when syncing GCP BackendService. Linking backend groups still uses Update

gauravkghildiyal:
#1991
#2043

/assign @swetharepakula


gauravkghildiyal:
We will not be releasing this in 1.22 since it has some breaking changes that require more investigation.

PR #2057, opened 2023-04-03
Revert "Use Patch when syncing GCP BackendService. Linking backend groups still uses Update"

gauravkghildiyal:
Reverts kubernetes/ingress-gce#2043

This is currently breaking our e2e tests. We'll look at other possible alternatives or try to fix this.

/assign @swetharepakula


swetharepakula (commented):
/ok-to-test
/approve
/lgtm

PR #2063, opened 2023-04-04
Fix a case when regional instance group linker has to add a backend t…

mmamczur:
…o the list of already linked backends.

The previous code would set the backends to only the new instance groups and later cause the controller to flip new/old backends on each refresh.


panslava:
/lgtm 

thanks!

PR #2064, opened 2023-04-04
[Cherry-pick #2063, #2067 to 1.22] Fix a case when regional instance group l…

mmamczur:
…inker has to add a backend to the list of already linked backends.

The previous code would set the backends to only the new instance groups and later cause the controller to flip new/old backends on each refresh.

### DIFF
--- a/pkg/backends/ig_linker.go
+++ b/pkg/backends/ig_linker.go
@@ -96,7 +96,7 @@ func (igl *instanceGroupLinker) Link(sp utils.ServicePort, groups []GroupKey) er
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


panslava:
@mmamczur can you add to this PR also commit that fixes deletion of IGs?

/assign @cezarygerard

Cezary could you please take a look (after another commit is added), I think this is important change

mmamczur:
> @mmamczur can you add to this PR also commit that fixes deletion of IGs? (from here #2067)
> 
> /assign @cezarygerard
> 
> Cezary could you please take a look (after another commit is added), I think this is important change

Added

PR #2066, opened 2023-04-04
Degraded mode e2e

sawsa307:
Create e2e test for degraded mode.
When an endpointslice contains endpoint with missing or empty nodeName, this should be filled during degraded mode. When an endpointslice contains endpoint with invalid pod information, or the endpoint has an IP that doesn't correspond to any podIP, this endpoint should be filtered out.

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -24,12 +24,14 @@ import (
 
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/e2e"
 	"k8s.io/ingress-gce/pkg/fuzz"
 	"k8s.io/ingress-gce/pkg/neg/types/shared"
+	"k8s.io/ingress-gce/pkg/utils/common"
 )
 
 func TestNEG(t *testing.T) {
@@ -47,7 +49,7 @@ func testNEGOS(t *testing.T, os e2e.OS) {
 		serviceName1       = "neg-service1"
 		serviceName2       = "neg-service2"
 		ingressName        = "neg-ingress1"
-		replicas           = int32(2)
+		replicas           = 2
 	)
 	port80 := networkingv1.ServiceBackendPort{Number: 80}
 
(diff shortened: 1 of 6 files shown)


pkg/e2e/fixtures.go
```diff
@@ -576,3 +578,131 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+		return s.f.Clientset.DiscoveryV1().EndpointSlices(s.Namespace).Create(context.TODO(), endpointSlice, metav1.CreateOptions{})
+	}
+	if !reflect.DeepEqual(existingEPS.Endpoints, endpointSlice.Endpoints) ||
+		!reflect.DeepEqual(existingEPS.Ports, endpointSlice.Ports) {
+		return s.f.Clientset.DiscoveryV1().EndpointSlices(s.Namespace).Update(context.TODO(), endpointSlice, metav1.UpdateOptions{})
+	}
+	return existingEPS, nil
+}
+
+func EnsurePod(s *Sandbox, name string) (*v1.Pod, error) {
```
bowei, line 648: ensure echo headers pod

pkg/e2e/fixtures.go
```diff
@@ -576,3 +578,131 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
 func DeleteServiceAttachment(s *Sandbox, saName string) error {
 	return s.f.SAClient.Delete(s.Namespace, saName)
 }
+
+func EnsureEndpointSlice(s *Sandbox, svc *v1.Service, EPSName, managedByLabel string, endpointCount int, modify func(endpointslice *discoveryv1.EndpointSlice)) (*discoveryv1.EndpointSlice, error) {
```
bowei, line 582: document what this is doing

bowei, line 582: write a unit test for this

pkg/e2e/helpers.go
```diff
@@ -1122,3 +1123,45 @@ func Truncate(key string) string {
 	}
 	return key
 }
+
+func WaitForEndpointSlice(s *Sandbox, name string) (*discoveryv1.EndpointSlice, error) {
```
bowei, line 1127: Document all of these methods that you are adding especially since they are public

pkg/e2e/helpers.go
```diff
@@ -1122,3 +1123,45 @@ func Truncate(key string) string {
...
+		if err := CheckPod(pod); err != nil {
+			klog.Infof("WaitForPodStable(%s/%s) = %v", s.Namespace, name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+	return pod, returnErr
+}
+
+func CheckPod(pod *v1.Pod) error {
```
bowei, line 1159: Documentation??

cmd/e2e-test/neg_test.go
```diff
@@ -875,3 +876,137 @@ func TestLabelPropagation(t *testing.T) {
...
+				endpointslice.Endpoints[0].Addresses[0] = "1.1.1.1"
+			},
+			expectedCount: 3,
+		},
+	} {
+		tc := tc // Capture tc as we are running this in parallel.
+		Framework.RunWithSandbox(tc.desc, t, func(t *testing.T, s *e2e.Sandbox) {
+			t.Parallel()
+			ctx := context.Background()
+
```
bowei, line 937: Put the major steps as comments.

// Setup standalone NEG.

//...

sawsa307: Updated.

pkg/e2e/fixtures.go
sawsa307: Updated.

pkg/e2e/fixtures.go
sawsa307: Added. Thanks!

pkg/e2e/helpers.go
sawsa307: Added. Thanks!

pkg/e2e/helpers.go
sawsa307: Added.

pkg/e2e/fixtures.go
sawsa307: Added `TestEnsureCustomEndpointSlice` in `fixtures_test.go`. Thanks!

pkg/e2e/fixtures_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	// Create pods directly since the fake deployment does not create pods.
+	// We also need to provide podIP in the pod config since podIP won't be
+	// generating in fake Pod creation.
+	for i := 1; i <= endpointCount; i++ {
+		podName := fmt.Sprintf("pod-%d", i)
+		pod, err := newTestPodWithIP(s, podName, testEPSName, fmt.Sprintf("10.100.1.%d", i))
+		if err != nil || pod == nil {
+			t.Fatalf("Error when newTestPodWithIP(%s, %s)", podName, testEPSName)
```
swetharepakula, line 74: track the pod ips here

pkg/e2e/fixtures_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	podIPs := []string{}
+	for _, pod := range pods.Items {
+		for key, value := range testPodLabels {
+			if pod.ObjectMeta.Labels[key] != value {
+				continue
+			}
+		}
+		podIPs = append(podIPs, pod.Status.PodIP)
+	}
```
swetharepakula, line 97: track the pod ips at creation, so you don't have to go through the list a second time

pkg/e2e/fixtures_test.go
```diff
@@ -0,0 +1,172 @@
...
+		Spec: podSpec,
+	}
+
+	currentPod, err := s.f.Clientset.CoreV1().Pods(s.Namespace).Get(context.TODO(), pod.ObjectMeta.Name, metav1.GetOptions{})
+	if err != nil || currentPod == nil {
+		return s.f.Clientset.CoreV1().Pods(s.Namespace).Create(context.TODO(), pod, metav1.CreateOptions{})
+	}
+	if !reflect.DeepEqual(currentPod.Spec, pod.Spec) {
+		return s.f.Clientset.CoreV1().Pods(s.Namespace).Update(context.TODO(), pod, metav1.UpdateOptions{})
+	}
```
swetharepakula, line 170: Should we simplify this to always create a new pod? Is there a need to update an existing pod?

pkg/e2e/fixtures_test.go
sawsa307: Updated. Thanks!

pkg/e2e/fixtures_test.go
sawsa307: Updated. Since the function is for pod creation, we shouldn't do pod update. There is no need for update either.

pkg/e2e/fixtures_test.go
sawsa307: Updated. Thanks!

pkg/e2e/adapter/ingress.go
```diff
@@ -26,7 +26,7 @@ const (
 
 // IngressCRUD wraps basic CRUD to allow use of old and new APIs.
 type IngressCRUD struct {
-	C *kubernetes.Clientset
```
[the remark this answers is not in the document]
sawsa307: There is a change in the type for [framework.ClientSet](https://github.com/kubernetes/ingress-gce/blob/525de9e782f47ef43d002e920f8741b9232b6b30/pkg/e2e/framework.go#L142), which is why we also need to update here.
The change for framework.ClientSet is because we need to use NewSimpleClientset() to initiate the fake ClientSet in sandbox.

cmd/e2e-test/neg_test.go
```diff
@@ -875,3 +876,152 @@ func TestLabelPropagation(t *testing.T) {
...
 	}
 }
+
+func TestDegradedMode(t *testing.T) {
+	t.Parallel()
+
+	const (
+		epsName           = "custom-endpointslice"
+		svcName           = "service-1"
+		replicas          = int32(2)
```
bowei, line 886: In go numeric consts don't have a type -- you shouldn't need to do this
https://go.dev/blog/constants

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
 func DeleteServiceAttachment(s *Sandbox, saName string) error {
 	return s.f.SAClient.Delete(s.Namespace, saName)
 }
+
+// EnsureCustomEndpointSlice ensures that a custom endpoint slice with the given modification is set up.
+func EnsureCustomEndpointSlice(s *Sandbox, svc *v1.Service, name string, addressType discoveryv1.AddressType, endpointCount int, modify func(endpointslice *discoveryv1.EndpointSlice)) (*discoveryv1.EndpointSlice, error) {
```
bowei, line 582: You need to document what each of these parameters do, otherwise it's going to be impossible to reuse this function.

For example, endpointCount seems to be a parameter for testing versus something used to construct the custom endpoint slice. Mixing checking with fixture setup is very confusing.

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+	tcp := v1.ProtocolTCP
+	endpointSlice := &discoveryv1.EndpointSlice{
+		AddressType: addressType,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      svc.Name + "-" + name,
```
bowei, line 587: I find it pretty confusing that `name` is not the actual name of the resource created...

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+			Namespace: s.Namespace,
+			Labels: map[string]string{
+				discoveryv1.LabelServiceName: svc.Name,
+				discoveryv1.LabelManagedBy:   "foo", // This is a custom endpoint slice.
+			},
+		},
+		Ports: []discoveryv1.EndpointPort{
+			{
+				Name:     &svc.Spec.Ports[0].Name,
```
bowei, line 596: You seem to have very specific requirements for the value of `svc`. You should at least document this in the function docstring. Better is to check that this is true and emit and error if this isn't correct OR structure the code so that we don't have to know about this requirement.

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+				Port:     &svc.Spec.Ports[0].TargetPort.IntVal,
+				Protocol: &tcp,
+			},
+			{
+				Name:     &svc.Spec.Ports[1].Name,
+				Port:     &svc.Spec.Ports[1].TargetPort.IntVal,
+				Protocol: &tcp,
+			},
+		},
+		Endpoints: []discoveryv1.Endpoint{},
```
bowei, line 606: This is a zero value so can be omitted.

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+			TargetRef: &v1.ObjectReference{
+				Kind:      "Pod",
+				Namespace: s.Namespace,
+				Name:      pod.Name,
+				UID:       pod.ObjectMeta.UID,
+			},
+		}
+		endpointSlice.Endpoints = append(endpointSlice.Endpoints, endpoint)
+	}
+	if len(endpointSlice.Endpoints) != endpointCount {
```
bowei, line 630: I don't understand this -- we seem to be assuming we'll get a certain number of pods.

It's probably better to pass in the Deployment resource and dynamically add the deployments Pods to the EPS. This makes the function very fragile and hard to use

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+	modify(endpointSlice)
+	// Create the custom endpoint slice if it is doesn't exist.
+	existingEPS, err := s.f.Clientset.DiscoveryV1().EndpointSlices(s.Namespace).Get(context.TODO(), endpointSlice.ObjectMeta.Name, metav1.GetOptions{})
+	if err != nil || existingEPS == nil {
+		return s.f.Clientset.DiscoveryV1().EndpointSlices(s.Namespace).Create(context.TODO(), endpointSlice, metav1.CreateOptions{})
+	}
+
+	// Address type should immutable after creation.
+	if existingEPS.AddressType != endpointSlice.AddressType {
+		return nil, fmt.Errorf("An endpointSlice with a different addressType already exist.")
```
bowei, line 643: you should always include as much detail in your error to make it easy for people to debug. 

return nil, fmt.Errorf("endpointSlice %s:%s addressType cannot be modified (existing=%q, new=%q)", existingEPS.Namespace, existingEPS.Name, existingEPS.AddressType, endpointSlice.AddressType)

pkg/e2e/fixtures_test.go
```diff
@@ -0,0 +1,147 @@
...
+	svcAnnotations := map[string]string{annotations.NEGAnnotationKey: annotation.String()}
+	svc, err := EnsureEchoService(s, testServiceName, svcAnnotations, v1.ServiceTypeClusterIP, replicas)
+	if err != nil {
+		t.Fatalf("Failed to create echo service, err: %v", err)
+	}
+
+	// Create pods directly since the fake deployment does not create pods.
+	// We also need to provide podIP in the pod config since podIP won't be
+	// generating in fake Pod creation.
+	podIPs := []string{}
```
bowei, line 65: var podIPs []string

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
...
+		Endpoints: []discoveryv1.Endpoint{},
+	}
+
+	// Add endpoints to endpoint slice and only using IPs from newly created pods.
+	pods, err := s.f.Clientset.CoreV1().Pods(s.Namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, pod := range pods.Items {
+		if pod.ObjectMeta.Labels["app"] != name {
```
bowei, line 615: you should document this is what you are looking for when you are creating the slice

pkg/e2e/fixtures.go
```diff
@@ -576,3 +577,74 @@ func EnsureServiceAttachment(s *Sandbox, saName, svcName, subnetName string) (*s
 func DeleteServiceAttachment(s *Sandbox, saName string) error {
 	return s.f.SAClient.Delete(s.Namespace, saName)
 }
+
+// EnsureCustomEndpointSlice ensures that a custom endpoint slice with the given modification is set up.
+func EnsureCustomEndpointSlice(s *Sandbox, svc *v1.Service, name string, addressType discoveryv1.AddressType, endpointCount int, modify func(endpointslice *discoveryv1.EndpointSlice)) (*discoveryv1.EndpointSlice, error) {
+	tcp := v1.ProtocolTCP
+	endpointSlice := &discoveryv1.EndpointSlice{
+		AddressType: addressType,
```
bowei, line 585: Can AddressType come from service? Do we test for different AddressTypes? Can we leave this out and assume one kind?

pkg/e2e/fixtures.go
sawsa307: Updated.

pkg/e2e/fixtures_test.go
sawsa307: Updated.

pkg/e2e/fixtures.go
sawsa307: Updated. Thanks!

pkg/e2e/fixtures.go
sawsa307: Updated. Now we take a list of pods and use that to create endpoints.

pkg/e2e/fixtures.go
sawsa307: Moved to `TestDegradedMode` and added comments.

pkg/e2e/fixtures.go
sawsa307: Updated. Now we iterate the service ports to get ports for endpoint slices.

pkg/e2e/fixtures.go
sawsa307: Updated. Now we pass in a list of pods and use them to generate endpoints directly.

cmd/e2e-test/neg_test.go
sawsa307: Updated. Thanks!

pkg/e2e/fixtures.go
sawsa307: Updated. Now we get address type from svc.Spec.IPFamilies. 
We only test for ipv4 service here because we don't have access to dual stack NEG API. Testing for dual stack service will be written in g3.

cmd/e2e-test/neg_test.go
```diff
@@ -875,3 +876,168 @@ func TestLabelPropagation(t *testing.T) {
...
+		endpointAttach    = int32(2)
```
bowei, line 887: these are still type int32? Is this needed?

cmd/e2e-test/neg_test.go
```diff
@@ -875,3 +876,168 @@ func TestLabelPropagation(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("Error: e2e.WaitForNegs service %s/%s neg port/name %s/%s", svcName, s.Namespace, port, negName)
+				}
+			}
+
+			// To get endpoint IPs in custom endpoint slice, create a new
+			// deployment with labels not matched to the service so the desired
+			// number of podIPs are created and not used by the existing
+			// endpoint slice.
+			if err := e2e.EnsureEchoDeployment(s, epsName, endpointAttach, e2e.NoopModify); err != nil {
```
bowei, line 976: Can we do this setup before running through all of the test cases?
E.g. do we need to redploy new Pods every time through the test case or just recreate the endpointslices?

E.g. setup the Pods before we start.
Then we can make the expected not just be a count but more explicitly the Pods we expect to show up in the slice based on what has been created.

bowei: Let's pull the setup pods out of the for loop for the test cases

pkg/e2e/adapter/ingress.go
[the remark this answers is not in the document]
swetharepakula: Okay, that makes sense. Thanks for the clarification!

cmd/e2e-test/neg_test.go
sawsa307: Done. Thank you!

cmd/e2e-test/neg_test.go
sawsa307: Updated.

pkg/e2e/helpers.go
```diff
@@ -1122,3 +1123,26 @@ func Truncate(key string) string {
...
+			for _, networkEndpoint := range neg.Endpoints {
+				endpoints = append(endpoints, networkEndpoint.NetworkEndpoint.IpAddress)
+			}
+		}
+		for _, addr := range expectToContainEndpoint {
+			if !slices.Contains(endpoints, addr) {
+				return false, nil
+			}
+		}
+		return true, nil
```
swetharepakula, line 1146: we should also that the number of endpoints are the same. It could be that there are more endpoints in the NEG than we expect.

cmd/e2e-test/neg_test.go
```diff
@@ -875,3 +877,196 @@ func TestLabelPropagation(t *testing.T) {
...
+				}
+
+				// Get podIPs from service deployment.
+				podList, err := e2e.ListPods(s)
+				if err != nil {
+					t.Fatalf("Error: e2e.ListPods() = %v, want nil", err)
+				}
+				var epInNEGFromSvcDeployment []string
+				for _, pod := range podList.Items {
+					if pod.ObjectMeta.Labels["app"] != svcName {
```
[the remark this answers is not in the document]
sawsa307: Yes it should fail since we only created pods with echo service deployment.

pkg/e2e/helpers.go
sawsa307: Updated. Thanks!

swetharepakula (commented):
/lgtm
/approve

PR #2067, opened 2023-04-05
Support IG backends removal in regional_ig_linker.go.

mmamczur:
Tried to make it so that whatever exists in the backends list is preserved instead of overwriting with 'wantIGs'.

### DIFF
--- a/pkg/backends/ig_linker_test.go
+++ b/pkg/backends/ig_linker_test.go
@@ -183,62 +183,71 @@ func TestBackendsForIG(t *testing.T) {
 	}
 }
 
-func TestGetIGLinksToAdd(t *testing.T) {
+func TestGetIGLinksToAddAndRemove(t *testing.T) {
 	url := "https://googleapis.com/v1/compute/projects/my-project/global/backendServices/%s"
 	link := "projects/my-project/global/backendServices/%s"
 	for _, tc := range []struct {
 		name           string
 		igLinks        []string
 		currentIGLinks []string
-		want           []string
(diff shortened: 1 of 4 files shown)


pkg/backends/ig_linker_test.go
```diff
@@ -251,15 +260,22 @@ func TestGetIGLinksToAdd(t *testing.T) {
...
+				t.Fatalf("getInstanceGroupsToAddAndRemove(_,_): err:%v ", err)
 			}
 			sort.Slice(toAdd, func(i, j int) bool {
 				return toAdd[i] <= toAdd[j]
 			})
-			if !reflect.DeepEqual(toAdd, tc.want) {
-				t.Fatalf("getInstanceGroupsToAdd(_,_) error. Got:%v, Want:%v", toAdd, tc.want)
+			sort.Slice(tc.wantRemove, func(i, j int) bool {
+				return tc.wantRemove[i] <= tc.wantRemove[j]
+			})
```
panslava, line 272: we can just do sort.Strings

panslava: and on the previous lines too?

mmamczur: done

panslava:
code looks good to me, but still want @swetharepakula to take a look, I am not sure, if we should remove instance groups, and why it was not done before (in ig_linker.go as well)

swetharepakula (commented):
/lgtm
/approve

PR #2068, opened 2023-04-05
JSONify healthcheck Description for BackendConfig

DamianSawicki:
Replace a plain string with a JSON string as the health check Description when the health check is configured with a BackendConfig CRD. The longer plan is to similarly JSONify all health check Descriptions, irrespective of the health check configuration.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -328,12 +328,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	// In NonGCP mode, use the zone specified in gce.conf directly.
 	// This overrides the zone/fault-domain label on nodes for NEG controller.
 	if flags.F.EnableNonGCPMode {
-		zone, err := ctx.Cloud.GetZone(context.Background())
-		if err != nil {
-			klog.Errorf("Failed to retrieve zone information from Cloud provider: %v; Please check if local-zone is specified in gce.conf.", err)
-		} else {
-			zoneGetter = negtypes.NewSimpleZoneGetter(zone.FailureDomain)
-		}
+		zoneGetter = negtypes.NewSimpleZoneGetter(ctx.Cloud.LocalZone())
 	}
 
 	enableAsm := false
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"k8s.io/ingress-gce/pkg/translator"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/healthcheck"
 	"k8s.io/klog/v2"
 )
 
@@ -45,13 +46,27 @@ type HealthChecks struct {
 	defaultBackendSvc types.NamespacedName
 	recorderGetter    RecorderGetter
 	serviceGetter     ServiceGetter
+	clusterInfo       healthcheck.ClusterInfo
 }
 
 // NewHealthChecker creates a new health checker.
 // cloud: the cloud object implementing SingleHealthCheck.
 // defaultHealthCheckPath: is the HTTP path to use for health checks.
 func NewHealthChecker(cloud HealthCheckProvider, healthCheckPath string, defaultBackendSvc types.NamespacedName, recorderGetter RecorderGetter, serviceGetter ServiceGetter) *HealthChecks {
-	return &HealthChecks{cloud, healthCheckPath, defaultBackendSvc, recorderGetter, serviceGetter}
+	ci := generateClusterInfo(cloud.(*gce.Cloud))
+	return &HealthChecks{cloud, healthCheckPath, defaultBackendSvc, recorderGetter, serviceGetter, ci}
+}
+
+func generateClusterInfo(gceCloud *gce.Cloud) healthcheck.ClusterInfo {
+	var location string
+	regionalCluster := gceCloud.Regional()
+	if regionalCluster {
+		location = gceCloud.Region()
+	} else {
+		location = gceCloud.LocalZone()
+	}
+	name := flags.F.GKEClusterName
+	return healthcheck.ClusterInfo{Name: name, Location: location, Regional: regionalCluster}
 }
 
 // new returns a *HealthCheck with default settings and specified port/protocol
@@ -70,6 +85,7 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
 	hc.Service = h.getService(sp)
+	hc.SetHealthcheckInfo(h.generateHealthcheckInfo(sp, hc.ForILB))
 	return hc
 }
 
@@ -92,6 +108,19 @@ func (h *HealthChecks) mainService(sp utils.ServicePort) types.NamespacedName {
 		service = sp.ID.Service
 	}
 	return service
+}
+
+func (h *HealthChecks) generateHealthcheckInfo(sp utils.ServicePort, iLB bool) healthcheck.HealthcheckInfo {
+	serviceInfo := healthcheck.ServiceInfo(h.defaultBackendSvc)
+	if sp.ID.Service.Name != "" {
+		serviceInfo = healthcheck.ServiceInfo(sp.ID.Service)
+	}
+
+	return healthcheck.HealthcheckInfo{
+		ClusterInfo:       h.clusterInfo,
+		ServiceInfo:       serviceInfo,
+		HealthcheckConfig: healthcheck.DefaultHC,
+	}
 }
 
 // SyncServicePort implements HealthChecker.
--- a/pkg/healthchecks/healthchecks_test.go
+++ b/pkg/healthchecks/healthchecks_test.go
@@ -18,6 +18,7 @@ package healthchecks
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -44,6 +45,7 @@ import (
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"k8s.io/ingress-gce/pkg/translator"
 	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/healthcheck"
 	namer_util "k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog/v2"
 )
@@ -444,18 +446,28 @@ func getSingletonHealthcheck(t *testing.T, c *gce.Cloud) *compute.HealthCheck {
 
 // Test changing the value of the flag EnableUpdateCustomHealthCheckDescription from false to true.
 func TestRolloutUpdateCustomHCDescription(t *testing.T) {
-	// No parallel() because we modify the value of the flags EnableBackendConfigHealthCheck and EnableUpdateCustomHealthCheckDescription.
+	// No parallel() because we modify the value of the flags:
+	// - EnableBackendConfigHealthCheck,
+	// - EnableUpdateCustomHealthCheckDescription,
+	// - GKEClusterName.
+
+	testClusterValues := gce.DefaultTestClusterValues()
+
 	oldEnableBC := flags.F.EnableBackendConfigHealthCheck
 	oldUpdateDescription := flags.F.EnableUpdateCustomHealthCheckDescription
+	oldGKEClusterName := flags.F.GKEClusterName
 	flags.F.EnableBackendConfigHealthCheck = true
 	// Start with EnableUpdateCustomHealthCheckDescription = false.
 	flags.F.EnableUpdateCustomHealthCheckDescription = false
+	flags.F.GKEClusterName = testClusterValues.ClusterName
 	defer func() {
 		flags.F.EnableBackendConfigHealthCheck = oldEnableBC
 		flags.F.EnableUpdateCustomHealthCheckDescription = oldUpdateDescription
+		flags.F.GKEClusterName = oldGKEClusterName
 	}()
 
-	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
+	testClusterValues.Regional = true
+	fakeGCE := gce.NewFakeGCECloud(testClusterValues)
 
 	var (
 		defaultSP       *utils.ServicePort = testSPs["HTTP-80-reg-nil"]
@@ -506,9 +518,20 @@ func TestRolloutUpdateCustomHCDescription(t *testing.T) {
 
 	outputBCHCWithFlag := getSingletonHealthcheck(t, fakeGCE)
 
-	if outputBCHCWithFlag.Description != translator.DescriptionForHealthChecksFromBackendConfig {
+	wantDesc := healthcheck.HealthcheckDesc{
+		K8sCluster:  fmt.Sprintf("/locations/%s/clusters/%s", gce.DefaultTestClusterValues().Region, gce.DefaultTestClusterValues().ClusterName),
+		K8sResource: fmt.Sprintf("/namespaces/%s/services/%s", defaultBackendSvc.Namespace, defaultBackendSvc.Name),
+		Config:      "BackendConfig",
+	}
+	bytes, err := json.MarshalIndent(wantDesc, "", "    ")
+	if err != nil {
+		t.Fatalf("Error while generating the wantDesc JSON.")
+	}
+	wantDescription := string(bytes)
+
+	if outputBCHCWithFlag.Description != wantDescription {
 		t.Fatalf("incorrect Description, is: \"%v\", want: \"%v\"",
-			outputBCHCWithFlag.Description, translator.DescriptionForHealthChecksFromBackendConfig)
+			outputBCHCWithFlag.Description, wantDescription)
 	}
 
 	// Verify that only the Description is modified after rollout.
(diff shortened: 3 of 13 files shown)


pkg/backends/syncer.go
```diff
@@ -289,7 +289,11 @@ func (s *backendSyncer) ensureHealthCheck(sp utils.ServicePort) (string, error)
 			return "", fmt.Errorf("Error getting prober: %w", err)
 		}
 	}
-	return s.healthChecker.SyncServicePort(&sp, probe)
+	i, err := s.cloud.ExtractClusterInfo()
```
kl52752, line 292: ingress controller is one per cluster there is no need to extract cluster info every sync. Use `flags.F.GKEClusterName` and `flags.F.GKEClusterType` instead. The UHC is already in some project so cluster name will be sufficient.
Build this info  one and pass it to controller at startup to get Region and zone use `controller.cloud.Region()` `zone, err := controller.cloud.GetZone(ctx)`

pkg/healthchecks/healthchecks.go
```diff
@@ -67,12 +68,14 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Name = sp.BackendName()
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
+	info := descutils.HealthcheckInfo{ClusterInfo: i, ServiceInfo: descutils.ServiceInfo(h.defaultBackendSvc), HealthcheckType: descutils.DefaultHC} // IngressType: ???
```
kl52752, line 71: please break lines.
Do not leave undefined comments if there is sth to discuss create TODO

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+	DefaultHC        HealthcheckType = 1
+	ReadinessProbeHC HealthcheckType = 2
+	BackendConfigHC  HealthcheckType = 3
+)
+
+type HealthcheckInfo struct {
+	ClusterInfo
+	ServiceInfo
+	HealthcheckType
+	// IngressType string
```
kl52752, line 47: Add to do if needed

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+}
+
+type HealthcheckDesc struct {
+	K8sCluster  string
+	K8sResource string
+	// K8ResourceDependency string
```
kl52752, line 53: add todo if needed

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+	Config string
+}
+
+func NewClusterInfo(name, location string, regional bool) ClusterInfo {
```
kl52752, line 57: this is a public struct I don't think new function is needed here

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+		locType = "locations"
+	}
+	return fmt.Sprintf("/%s/%s/clusters/%s", locType, i.location, i.name)
+}
+
+func NewServiceInfo(namespace, resourceName string) ServiceInfo {
+	return ServiceInfo{namespace, resourceName}
+}
+
+func GenerateK8sResourceDescriptionFromServiceInfo(i ServiceInfo) string {
```
kl52752, line 74: FromServiceInfo is redundant IMO

pkg/utils/descutils/utils.go
```diff
@@ -33,3 +90,25 @@ func GenerateClusterLink(name, location string, regional bool) string {
 func GenerateK8sResourceLink(namespace, resourceType, resourceName string) string {
 	return fmt.Sprintf("/namespaces/%s/%s/%s", namespace, resourceType, resourceName)
 }
+
+func GenerateHealthcheckDescriptionFromHealthcheckInfo(i HealthcheckInfo) string {
```
kl52752, line 94: FromHealthCheckInfo is redundant

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+	K8sResource string
+	// K8ResourceDependency string
+	Config string
+}
+
+func NewClusterInfo(name, location string, regional bool) ClusterInfo {
+	return ClusterInfo{name, location, regional}
+}
+
+func GenerateClusterDescriptionFromClusterInfo(i ClusterInfo) string {
```
kl52752, line 61: this function can be member of ClusterInfo and there is no point to export it it is used only in GenerateHealthcheckDescriptionFromHealthcheckInfo

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+
+type ClusterInfo struct {
+	name     string
+	location string
+	regional bool
+}
+
+type ServiceInfo types.NamespacedName
+
+type HealthcheckType int
```
kl52752, line 35: use string then you will not need if statement and convert it to string

pkg/utils/descutils/utils.go
```diff
@@ -33,3 +90,25 @@ func GenerateClusterLink(name, location string, regional bool) string {
 func GenerateK8sResourceLink(namespace, resourceType, resourceName string) string {
 	return fmt.Sprintf("/namespaces/%s/%s/%s", namespace, resourceType, resourceName)
 }
+
+func GenerateHealthcheckDescriptionFromHealthcheckInfo(i HealthcheckInfo) string {
+	desc := HealthcheckDesc{}
+	desc.K8sCluster = GenerateClusterDescriptionFromClusterInfo(i.ClusterInfo)
+	desc.K8sResource = GenerateK8sResourceDescriptionFromServiceInfo(i.ServiceInfo)
+	// desc.K8ResourceDependency = i.IngressType
+	switch i.HealthcheckType {
```
kl52752, line 99: see comment about type.
Without this switch you can initialize description at once with proper values.

pkg/utils/descutils/utils.go
DamianSawicki: Ok. This requires renaming the struct fields to capital letters, but it's fine.

pkg/healthchecks/healthchecks.go

pkg/utils/descutils/utils.go
DamianSawicki: No longer needed :)

pkg/utils/descutils/utils.go
DamianSawicki: No longer needed :)

pkg/utils/descutils/utils.go
DamianSawicki: Turned into a member function and renamed.

pkg/utils/descutils/utils.go
DamianSawicki: Turned into a member function and renamed.

pkg/utils/descutils/utils.go
DamianSawicki: Done.

pkg/utils/descutils/utils.go
DamianSawicki: Great idea, thanks!

pkg/utils/descutils/utils.go
DamianSawicki: Cool, done.

pkg/utils/descutils/utils.go
DamianSawicki: Done.

pkg/backends/syncer.go
DamianSawicki: Done, although without using the function `GetZone(ctx)`. This function returns two types: value and error, despite the fact that it is clear from its trivial implementation that it never fails. Furthermore, the value type is a struct consisting of a zone and a region, despite the fact that everywhere the function is called, only the zone is utilised.

I suggest to delete `GetZone(ctx)` and replace calls to it by calls to `Zone()`.

DamianSawicki:
No e2e manual tests done yet, but unit tests (existing ones with modifications and some new ones) pass.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1325,15 +1337,40 @@ func TestSyncServicePort(t *testing.T) {
...
 
 	for _, tc := range cases {
+		tc := *tc
 		tc.updateHCDescription = true
 		tc.desc = tc.desc + " with updateHCDescription"
-		cases = append(cases, tc)
+		copyOfWant := *tc.wantComputeHC
+		if tc.sp.BackendConfig != nil {
+			var wantLocation, wantIngressType string
+			if tc.regional {
```
kl52752, line 1346: tc.regional refers to HealthCheck not the cluster.
I'm surprised that this test is passing. Probably moc sets zone and region to the same values.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1325,15 +1337,40 @@ func TestSyncServicePort(t *testing.T) {
...
+		tc.wantComputeHC = &copyOfWant
+		cases = append(cases, &tc)
 	}
 
 	for _, tc := range cases {
 		t.Run(tc.desc, func(t *testing.T) {
 			flags.F.EnableUpdateCustomHealthCheckDescription = tc.updateHCDescription
-			fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
+			testClusterValues := gce.DefaultTestClusterValues()
+			testClusterValues.Regional = tc.regional
```
kl52752, line 1372: make sure that Regiona and Zone in fake GCE are different values.

DamianSawicki: They are:
```
		Region:            "us-central1",
		ZoneName:          "us-central1-b",
```

pkg/translator/healthchecks.go
```diff
@@ -90,6 +91,19 @@ type HealthCheck struct {
 	// compute struct back.
 	computealpha.HTTPHealthCheck
 	computealpha.HealthCheck
+	healthcheckInfo descutils.HealthcheckInfo
+}
+
+func (hc *HealthCheck) UpdateHealthcheckConfig(c descutils.HealthcheckConfig) {
+	hc.healthcheckInfo.HealthcheckConfig = c
+	hc.SetHealthcheckInfo(hc.healthcheckInfo) // This is not a no-op.
```
kl52752, line 99: it doesn't really make sense you already updated the healthcheckInfo all you need to do is update the description if needed

pkg/healthchecks/healthchecks_test.go
kl52752: faceCloud sets managedZones [link](https://screenshot.googleplex.com/6ZkYuzYWPxDBBSt)
and in your code you use cloud.Zone() which uses localZone fields [link](https://screenshot.googleplex.com/6ZkYuzYWPxDBBSt) and it looks like it is not set here.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: > I'm surprised that this test is passing. Probably moc sets zone and region to the same values.

They are not:
```
		Region:            "us-central1",
		ZoneName:          "us-central1-b",
```

> tc.regional refers to HealthCheck not the cluster.

I added an extra parameter `tc.regionalCluster` to decouple these.

pkg/backends/syncer.go
kl52752: yes your are right, there is no point to check the error

pkg/healthchecks/healthchecks_test.go
DamianSawicki: After my changes fakeCloud sets both managedZones and localZone ([link](https://screenshot.googleplex.com/4yF2wUzyVta8BKt)). It is visible in your screenshots (btw. you entered the same link in both places).

pkg/healthchecks/healthchecks.go
```diff
@@ -43,17 +44,31 @@ type HealthChecks struct {
...
 	defaultBackendSvc types.NamespacedName
+	clusterInfo       descutils.ClusterInfo
 }
 
 // NewHealthChecker creates a new health checker.
 // cloud: the cloud object implementing SingleHealthCheck.
 // defaultHealthCheckPath: is the HTTP path to use for health checks.
 func NewHealthChecker(cloud HealthCheckProvider, healthCheckPath string, defaultBackendSvc types.NamespacedName) *HealthChecks {
-	return &HealthChecks{cloud, healthCheckPath, defaultBackendSvc}
+	var location string
```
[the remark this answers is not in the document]
DamianSawicki: I think it's already done. I mean, I extracted it from `NewHealthChecker` to `generateClusterInfo` after your verbal comments.

pkg/translator/healthchecks.go
DamianSawicki: Changed.

pkg/backends/syncer.go
DamianSawicki: Ok, the function `GetZones(ctx)` is a member function of the `Zones` interface (`vendor/k8s.io/cloud-provider/cloud.go`), so I won't delete it, but I replaced the calls `GetZones(ctx)` with `Zones()` as planned.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1325,15 +1354,40 @@ func TestSyncServicePort(t *testing.T) {
...
+				wantLocation = gce.DefaultTestClusterValues().Region
+			} else {
+				wantLocation = gce.DefaultTestClusterValues().ZoneName
+			}
+			if tc.sp.L7ILBEnabled {
+				wantIngressType = "Ingress Internal Load Balancer"
+			} else {
+				wantIngressType = "Ingress External Load Balancer"
+			}
+			copyOfWant.Description = "" +
```
swetharepakula, line 1373: same comment as above, use json marshal and unmarshal

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
...
+}
+
+type HealthcheckDesc struct {
+	K8sCluster            string
+	K8sResource           string
+	K8sResourceDependency IngressType
+	Config                HealthcheckConfig
+}
+
+func (i *ClusterInfo) generateClusterDescription() string {
```
swetharepakula, line 64: So the original idea was that this should be the selflink for the cluster resource. If using `locations` will this selflink be accurate?

Since the two functions are almost exactly the same and supposedly being used for the same thing, we should consolidate.

swetharepakula (commented):
I think we can do some consolidation from what is already done for the service attachment description.

cmd/glbc/main.go
```diff
@@ -307,12 +307,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	// In NonGCP mode, use the zone specified in gce.conf directly.
 	// This overrides the zone/fault-domain label on nodes for NEG controller.
 	if flags.F.EnableNonGCPMode {
-		zone, err := ctx.Cloud.GetZone(context.Background())
-		if err != nil {
-			klog.Errorf("Failed to retrieve zone information from Cloud provider: %v; Please check if local-zone is specified in gce.conf.", err)
-		} else {
```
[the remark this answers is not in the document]
DamianSawicki: `GetZone()` always returns nil as the error. `Zone()` just returns the desired value.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -503,9 +505,17 @@ func TestRolloutUpdateCustomHCDescription(t *testing.T) {
 
 	outputBCHCWithFlag := getSingletonHealthcheck(t, fakeGCE)
 
-	if outputBCHCWithFlag.Description != translator.DescriptionForHealthChecksFromBackendConfig {
+	wantDescription := "" +
+		fmt.Sprintf("{\n") +
```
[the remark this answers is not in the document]
DamianSawicki: I already use (un)marshal for generating these, so I think it would be better to be explicit (like elsewhere in the file) and also not to use the same way for testing.

pkg/utils/descutils/utils.go
```diff
@@ -16,7 +16,64 @@ limitations under the License.
 
 package descutils
 
-import "fmt"
+import (
```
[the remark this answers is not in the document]
DamianSawicki: Yeah, I had an impression that this is suboptimal, thanks for the suggestion!

pkg/healthchecks/healthchecks.go
```diff
@@ -43,13 +44,30 @@ type HealthChecks struct {
...
+	if regionalCluster {
+		location = gceCloud.Region()
+	} else {
+		location = gceCloud.Zone()
+	}
+	name, err := gceCloud.ClusterID.GetID()
+	if err != nil {
+		klog.Errorf("Failed to obtain cluster name, %+v", err)
+	}
+	return descutils.ClusterInfo{Name: name, Location: location, Regional: regionalCluster}
```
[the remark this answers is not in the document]
DamianSawicki: Well-spotted, thank you very much! It seems from [this line](https://github.com/kubernetes/ingress-gce/blob/5454f34be171d7ff924ebbd03883a5bdba0b7a2b/vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_loadbalancer.go#L152) that indeed the ID and name are distinct. Having seen [spaces](https://github.com/kubernetes/ingress-gce/blob/5454f34be171d7ff924ebbd03883a5bdba0b7a2b/vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_fake.go#L53) in the `ClusterName` test data, I inferred that the relevant field must be `ClusterID`.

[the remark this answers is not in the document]

pkg/healthchecks/healthchecks_test.go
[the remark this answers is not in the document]
DamianSawicki: @kl52752 seconded your point, and I changed it as suggested.

pkg/healthchecks/healthchecks_test.go

DamianSawicki:
The description looks good in manual tests.

DamianSawicki:
`K8sResourceDependency` removed as requested by @bowei. Tested as described in the bug-tracking system.

/unassign kl52752
/assign swetharepakula
/cc bowei

DamianSawicki:
I'll change description field names to start with lower-case letters.

DamianSawicki:
> I'll change description field names to start with lower-case letters.

Done.

PR #2072, opened 2023-04-07
Emit event on Description-only healthcheck update

DamianSawicki:
Since https://github.com/kubernetes/ingress-gce/pull/2008, the health check description is modified if the service health check is configured with a BackendConfig CRD. The purpose of the present PR is to emit an event on the service on this occasion.

The changes are guarded with the flag `flags.F.EnableUpdateCustomHealthCheckDescription` from https://github.com/kubernetes/ingress-gce/pull/2018/.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -114,7 +114,7 @@ func NewLoadBalancerController(
 		Interface: ctx.KubeClient.CoreV1().Events(""),
 	})
 
-	healthChecker := healthchecks.NewHealthChecker(ctx.Cloud, ctx.HealthCheckPath, ctx.DefaultBackendSvcPort.ID.Service)
+	healthChecker := healthchecks.NewHealthChecker(ctx.Cloud, ctx.HealthCheckPath, ctx.DefaultBackendSvcPort.ID.Service, ctx, ctx.Translator)
 	backendPool := backends.NewPool(ctx.Cloud, ctx.ClusterNamer)
 
 	lbc := LoadBalancerController{
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -43,13 +43,15 @@ type HealthChecks struct {
 	// This is a workaround which allows us to not have to maintain
 	// a separate health checker for the default backend.
 	defaultBackendSvc types.NamespacedName
+	recorderGetter    RecorderGetter
+	serviceGetter     ServiceGetter
 }
 
 // NewHealthChecker creates a new health checker.
 // cloud: the cloud object implementing SingleHealthCheck.
 // defaultHealthCheckPath: is the HTTP path to use for health checks.
-func NewHealthChecker(cloud HealthCheckProvider, healthCheckPath string, defaultBackendSvc types.NamespacedName) *HealthChecks {
-	return &HealthChecks{cloud, healthCheckPath, defaultBackendSvc}
+func NewHealthChecker(cloud HealthCheckProvider, healthCheckPath string, defaultBackendSvc types.NamespacedName, recorderGetter RecorderGetter, serviceGetter ServiceGetter) *HealthChecks {
+	return &HealthChecks{cloud, healthCheckPath, defaultBackendSvc, recorderGetter, serviceGetter}
 }
 
 // new returns a *HealthCheck with default settings and specified port/protocol
@@ -67,7 +69,29 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Name = sp.BackendName()
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
+	hc.Service = h.getService(sp)
 	return hc
+}
+
+func (h *HealthChecks) getService(sp utils.ServicePort) *v1.Service {
+	if !flags.F.EnableUpdateCustomHealthCheckDescription {
+		return nil
+	}
+	namespacedName := h.mainService(sp)
+	var err error
+	service, err := h.serviceGetter.GetService(namespacedName.Namespace, namespacedName.Name)
+	if err != nil {
+		klog.Warningf("Service %s/%s needed for emitting an event not found (we'll log instead): %v.", namespacedName.Namespace, namespacedName.Name, err)
+	}
+	return service
+}
+
+func (h *HealthChecks) mainService(sp utils.ServicePort) types.NamespacedName {
+	service := h.defaultBackendSvc
+	if sp.ID.Service.Name != "" {
+		service = sp.ID.Service
+	}
+	return service
 }
 
 // SyncServicePort implements HealthChecker.
@@ -142,6 +166,15 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 	changes := calculateDiff(filter(existingHC), filter(hc), bchcc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
+		if flags.F.EnableUpdateCustomHealthCheckDescription && changes.size() == 1 && changes.has("Description") {
+			message := fmt.Sprintf("Healthcheck will be updated and the only field updated is Description.\nOld: %+v\nNew: %+v\nDiff: %+v", existingHC, hc, changes)
+			if hc.Service != nil {
+				h.recorderGetter.Recorder(hc.Service.Namespace).Event(
+					hc.Service, v1.EventTypeNormal, "HealthcheckDescriptionUpdate", message)
+			} else {
+				klog.Info(message)
+			}
+		}
 		err := h.update(hc)
 		if err != nil {
 			klog.Errorf("Health check %q update error: %v", existingHC.Name, err)
--- a/pkg/healthchecks/healthchecks_test.go
+++ b/pkg/healthchecks/healthchecks_test.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -111,7 +113,7 @@ func init() {
 
 func TestHealthCheckAdd(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	sp := &utils.ServicePort{NodePort: 80, Protocol: annotations.ProtocolHTTP, NEGEnabled: false, BackendNamer: testNamer}
 	_, err := healthChecks.SyncServicePort(sp, nil)
@@ -149,7 +151,7 @@ func TestHealthCheckAdd(t *testing.T) {
 
 func TestHealthCheckAddExisting(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	// HTTP
 	// Manually insert a health check
@@ -221,7 +223,7 @@ func TestHealthCheckAddExisting(t *testing.T) {
 
 func TestHealthCheckDelete(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	// Create HTTP HC for 1234
 	hc := translator.DefaultHealthCheck(1234, annotations.ProtocolHTTP)
@@ -251,7 +253,7 @@ func TestHealthCheckDelete(t *testing.T) {
 
 func TestHTTP2HealthCheckDelete(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	// Create HTTP2 HC for 1234
 	hc := translator.DefaultHealthCheck(1234, annotations.ProtocolHTTP2)
@@ -278,7 +280,7 @@ func TestHTTP2HealthCheckDelete(t *testing.T) {
 
 func TestRegionalHealthCheckDelete(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	hc := healthChecks.new(
 		utils.ServicePort{
@@ -330,7 +332,7 @@ func TestHealthCheckUpdate(t *testing.T) {
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaHealthChecks.UpdateHook = mock.UpdateAlphaHealthCheckHook
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockBetaHealthChecks.UpdateHook = mock.UpdateBetaHealthCheckHook
 
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter())
 
 	// HTTP
 	// Manually insert a health check
@@ -465,7 +467,8 @@ func TestRolloutUpdateCustomHCDescription(t *testing.T) {
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockAlphaHealthChecks.UpdateHook = mock.UpdateAlphaHealthCheckHook
 	(fakeGCE.Compute().(*cloud.MockGCE)).MockBetaHealthChecks.UpdateHook = mock.UpdateBetaHealthCheckHook
 
-	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc)
+	fakeSingletonRecorderGetter := NewFakeSingletonRecorderGetter(1)
+	healthChecks := NewHealthChecker(fakeGCE, "/", defaultBackendSvc, fakeSingletonRecorderGetter, NewFakeServiceGetter())
 
 	_, err := healthChecks.SyncServicePort(defaultSP, nil)
 	if err != nil {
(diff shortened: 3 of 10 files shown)


pkg/healthchecks/healthchecks_test.go
```diff
@@ -517,6 +520,11 @@ func TestRolloutUpdateCustomHCDescription(t *testing.T) {
 	if !reflect.DeepEqual(outputBCHC, outputBCHCWithFlag) {
 		t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(outputBCHC), pretty.Sprint(outputBCHCWithFlag))
 	}
+
+	output := <-fakeRecorder.Events
+	if !strings.HasPrefix(output, "Warning HealthcheckDescriptionUpdate") {
+		t.Fatalf("Incorrect event emitted on healthcheck update: %s.", output)
+	}
```
aojea, line 527: this can block forever, use select with another input time.After to fail the test if it timesout

vendor/k8s.io/cloud-provider-gcp/providers/gce/gce.go
```diff
@@ -43,7 +43,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
aojea, line 46: vendor files should not be modified, maybe your IDE did this?

pkg/translator/healthchecks.go
```diff
@@ -338,3 +341,32 @@ func ApplyProbeSettingsToHC(p *v1.Probe, hc *HealthCheck) {
 
 	hc.Description = DescriptionForHealthChecksFromReadinessProbe
 }
+
+// GetObjectKind implements runtime.Object
```
aojea, line 345: is this change related?

pkg/healthchecks/healthchecks_test.go
DamianSawicki: I tried this (with modified settings or in another test) and the test timed out after 30 seconds.

pkg/healthchecks/healthchecks.go
```diff
@@ -142,6 +144,11 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 	changes := calculateDiff(filter(existingHC), filter(hc), bchcc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
+		if flags.F.EnableUpdateCustomHealthCheckDescription && changes.size() == 1 && changes.has("Description") {
+			message := fmt.Sprintf("Healthcheck will be updated and the only field updated is Description.\nOld: %+v\nNew: %+v\nDiff: %+v", existingHC, hc, changes)
```
aojea, line 148: this seems a very large message from an event, can we just send the diff or do we miss some important information?

aojea:
/ok-to-test

looks good overall, some comments that needs to be addressed

vendor/k8s.io/cloud-provider-gcp/providers/gce/gce.go
DamianSawicki: Right, it was my IDE. Thanks!

pkg/translator/healthchecks.go
DamianSawicki: Yes. `EventRecorder.Event` requires the first argument to be `runtime.Object`, and this interface contains two methods: `GetObjectKind() schema.ObjectKind` and `DeepCopyObject() Object`.

pkg/healthchecks/healthchecks_test.go

pkg/healthchecks/healthchecks.go
aojea: the message contains the New and Old too, can we send only the changes? do we need the others?

```
%+v\nNew: %+v\nDiff: %+v", existingHC, hc, changes)
```

pkg/healthchecks/healthchecks_test.go
```diff
@@ -517,6 +521,16 @@ func TestRolloutUpdateCustomHCDescription(t *testing.T) {
 	if !reflect.DeepEqual(outputBCHC, outputBCHCWithFlag) {
 		t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(outputBCHC), pretty.Sprint(outputBCHCWithFlag))
 	}
+
+	select {
+	case output := <-fakeRecorder.Events:
+		if !strings.HasPrefix(output, "Warning HealthcheckDescriptionUpdate") {
+			t.Fatalf("Incorrect event emitted on healthcheck update: %s.", output)
+		}
+	case <-time.After(2 * time.Second):
```
aojea, line 530: give it some room, CIs can be very slow, 10 seconds is ok, we just want to fail and now why in case it is blocked, updated the t.Fatalf to indicate that we timed out

pkg/translator/healthchecks.go
aojea: the object in this case is the Service or just pass the Namespace, doing this to the healtcheck object seems very brittle

aojea: I think you can get it from `defaultBackendSvc types.NamespacedName`

pkg/healthchecks/healthchecks.go
DamianSawicki: Yes, I'm trying to say that we need the others. The `changes` aka "diff" inside the `if` consists only of the modified Description (`changes.size() == 1 && changes.has("Description")`). However, we want to include more information in the message as a precaution.

A background fact that I should probably mention is that `calculateDiff()` does not return the full diff. It checks selected fields only (recently `Description` was added as one of them), sometimes just 2 of them.

pkg/translator/healthchecks.go
DamianSawicki: This again requires implementing the above methods either for `string` (in the case of namespace) or for `types.NamespacedName` (in the case of service), and when trying to do that, I'm getting the following error `cannot define new methods on non-local type "k8s.io/apimachinery/pkg/types".NamespacedName`. I guess I can write my own alias type for one of these and then implement the methods for the alias. @aojea, do you think it is a good idea?

pkg/healthchecks/healthchecks_test.go

pkg/translator/healthchecks.go
aojea: no sorry, my point is that we should not modify this type, and use one of the Kubernetes types associated to the healtcheck, it can be the Service or the Namespace, ideally the Service

DamianSawicki: Thanks for the clarification! Done, please have a look.

aojea: you forgot to remove this code

pkg/healthchecks/healthchecks.go
```diff
@@ -142,6 +160,11 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
...
+			h.eventRecorder.Event(
+				hc.Service, "Warning", "HealthcheckDescriptionUpdate", message)
```
aojea, line 166: hmm, I have to apologize, but now that I look it carefully , it seems Service is not the right object to pass, in this case this method is trying to sync `bchcc *backendconfigv1.HealthCheckConfig` , I think this object is implementing `runtime.Object`, isnt't it?, that will be the best thing since we can do this only with 4-5 lines of code

Sorry again, I've should have checked this better before doing any suggestion

pkg/backends/backends.go
```diff
@@ -41,7 +42,7 @@ type Backends struct {
 }
 
 // Backends is a Pool.
-var _ Pool = (*Backends)(nil)
+var _ interfaces.Pool = (*Backends)(nil)
```
kl52752, line 45: why this change was needed?

kl52752: I think you need to rebase

pkg/healthchecks/healthchecks.go
kl52752: @aojea we want to emit the even on the Service because UHC is created per Service and one BackendConfig can be assign to multiple Services but all Customer cares is a configuration of HealthCheck for Service.

pkg/backends/interfaces/interfaces.go
```diff
@@ -77,6 +77,21 @@ type NEGGetter interface {
 	GetNetworkEndpointGroup(name string, zone string, version meta.Version) (*composite.NetworkEndpointGroup, error)
 }
 
+// ServiceGetter is an interface to retrieve Kubernetes Services.
+type ServiceGetter interface {
+	GetService(namespace, name string) (*api_v1.Service, error)
+}
+
+func NewEmptyServiceGetter() ServiceGetter {
```
kl52752, line 85: NewFake

pkg/backends/interfaces/interfaces.go
```diff
@@ -77,6 +77,21 @@ type NEGGetter interface {
...
+	return &fakeServiceGetter{}
+}
+
+type fakeServiceGetter struct{}
+
+func (fsg *fakeServiceGetter) GetService(namespace, name string) (*api_v1.Service, error) {
```
kl52752, line 91: I think it will be better to return some default Service here. I don't think is good to return the empty object. Normally when you get nil error you expect that the Object is properly constructed. This may lead to some crashes I think.

pkg/controller/controller.go
```diff
@@ -114,7 +115,8 @@ func NewLoadBalancerController(
 		Interface: ctx.KubeClient.CoreV1().Events(""),
 	})
 
-	healthChecker := healthchecks.NewHealthChecker(ctx.Cloud, ctx.HealthCheckPath, ctx.DefaultBackendSvcPort.ID.Service)
+	eventRecorder := ctx.Recorder(ctx.Namespace)
```
kl52752, line 118: the namespace for recorder needs to be the namespace of the Service you wants to emit the event for. And health checker is created one per controller you need to postpone this function to the place where you have your Service. 
You probably need context there.

pkg/backends/interfaces/interfaces.go
DamianSawicki: Actually, according to the comments ,`api_v1.Service` has only one non-optional field of type `metav1.TypeMeta`, which in turn has only optional fields. So it seems that the "empty object" is a valid object.

kl52752: you can at least fill name and namespace to match function params.
Where is it used?

pkg/healthchecks/healthchecks.go
aojea: yeah, Damien commented offline, sounds good

aojea: yeah, Damien commented offline, sounds good

pkg/backends/interfaces/interfaces.go
aojea: this should be enough
```
	svc := &v1.Service{
			Name:      name,
			Namespace: namespace,
		},
	}
```

aojea:
just please don't use calls to the apiserver on the getter, remember that you have informers that already have Listers and Getter to a local cache with all the objects information. The other alternative is to plumb down the information on initialization, I'm not familiar with the code, so I can't suggest which one is the best option

pkg/controller/controller.go
DamianSawicki: Done.

pkg/backends/interfaces/interfaces.go
DamianSawicki: It's just used as one of the Event parameters.

pkg/backends/interfaces/interfaces.go
DamianSawicki: Done.

pkg/backends/backends.go
DamianSawicki: Discussed offline and fixed.

pkg/translator/healthchecks.go
DamianSawicki: Done.

pkg/backends/interfaces/interfaces.go
DamianSawicki: @aojea Yeah, I missed your comment earlier, but this is exactly what we ended up with.

pkg/healthchecks/healthchecks.go
```diff
@@ -67,9 +69,23 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Name = sp.BackendName()
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
+	service := h.mainService(sp)
```
DamianSawicki, line 72: I'll flag-gate the following lines.

pkg/healthchecks/healthchecks.go
```diff
@@ -142,6 +160,11 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 	changes := calculateDiff(filter(existingHC), filter(hc), bchcc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
+		if flags.F.EnableUpdateCustomHealthCheckDescription && changes.size() == 1 && changes.has("Description") {
+			message := fmt.Sprintf("Healthcheck will be updated and the only field updated is Description.\nOld: %+v\nNew: %+v\nDiff: %+v", existingHC, hc, changes)
+			h.recorderGetter.Recorder(hc.Service.Namespace).Event(
```
cezarygerard, line 165: hc.Service Can be null here,

if it error occurred while assigning the hc.Service in pkg/healthchecks/healthchecks.go
```
if flags.F.EnableUpdateCustomHealthCheckDescription {
		service := h.mainService(sp)
		var err error
		hc.Service, err = h.serviceGetter.GetService(service.Namespace, service.Name)
		if err != nil {
			klog.Errorf("Service %s/%s not found: %v.", service.Namespace, service.Name, err)
		}
	}
```

pkg/healthchecks/healthchecks.go
```diff
@@ -67,9 +69,25 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Name = sp.BackendName()
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
+	if flags.F.EnableUpdateCustomHealthCheckDescription {
```
cezarygerard, line 72: you should not be using flags this deep down the logic

convert them into fields somewhere in context or close to main

it makes it much more testable

pkg/healthchecks/healthchecks.go
```diff
@@ -67,9 +69,25 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
...
+		service := h.mainService(sp)
+		var err error
+		hc.Service, err = h.serviceGetter.GetService(service.Namespace, service.Name)
+		if err != nil {
+			klog.Errorf("Service %s/%s not found: %v.", service.Namespace, service.Name, err)
```
cezarygerard, line 77: 1. why erorr?
we keep going so it should be a Warning

2. Add one more sentence to this log, what are the consequences (not big, we keep going)

pkg/healthchecks/healthchecks.go
```diff
@@ -67,9 +69,25 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	hc.Name = sp.BackendName()
 	hc.Port = sp.NodePort
 	hc.RequestPath = h.pathFromSvcPort(sp)
+	if flags.F.EnableUpdateCustomHealthCheckDescription {
+		service := h.mainService(sp)
```
cezarygerard, line 73: service is a confusing name here
It suggests it's v1.Service
but it is just name and namespace

see my next comment

pkg/healthchecks/healthchecks.go
```diff
@@ -67,9 +69,25 @@ func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
...
+		var err error
+		hc.Service, err = h.serviceGetter.GetService(service.Namespace, service.Name)
+		if err != nil {
+			klog.Errorf("Service %s/%s not found: %v.", service.Namespace, service.Name, err)
+		}
+	}
```
cezarygerard, line 79: this whole part can be well hidden:
see this:
```
// new returns a *HealthCheck with default settings and specified port/protocol
	if sp.NEGEnabled && !sp.L7ILBEnabled {
		hc = translator.DefaultNEGHealthCheck(sp.Protocol)
	} else if sp.L7ILBEnabled {
		hc = translator.DefaultILBHealthCheck(sp.Protocol)
	} else {
		hc = translator.DefaultHealthCheck(sp.NodePort, sp.Protocol)
	}
	// port is the key for retrieving existing health-check
	// TODO: rename backend-service and health-check to not use port as key
	hc.Name = sp.BackendName()
	hc.Port = sp.NodePort
	hc.RequestPath = h.pathFromSvcPort(sp)
	hc.Service = h.getService(sp)

	return hc
}

func (h *HealthChecks) getService(sp utils.ServicePort) *v1.Service {
	if flags.F.EnableUpdateCustomHealthCheckDescription {
		namespacedName := h.mainService(sp)
		var err error
		service, err := h.serviceGetter.GetService(namespacedName.Namespace, namespacedName.Name)
		if err != nil {
			klog.Errorf("Service %s/%s not found: %v.", namespacedName.Namespace, namespacedName.Name, err)
		}
		return service
	}
	return nil
}

func (h *HealthChecks) mainService(sp utils.ServicePort) types.NamespacedName {
	service := h.defaultBackendSvc
	if sp.ID.Service.Name != "" {
		service = sp.ID.Service
	}
	return service
}
```

pkg/healthchecks/healthchecks.go
```diff
@@ -142,6 +160,11 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 	changes := calculateDiff(filter(existingHC), filter(hc), bchcc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
+		if flags.F.EnableUpdateCustomHealthCheckDescription && changes.size() == 1 && changes.has("Description") {
+			message := fmt.Sprintf("Healthcheck will be updated and the only field updated is Description.\nOld: %+v\nNew: %+v\nDiff: %+v", existingHC, hc, changes)
+			h.recorderGetter.Recorder(hc.Service.Namespace).Event(
+				hc.Service, "Warning", "HealthcheckDescriptionUpdate", message)
```
cezarygerard, line 166: this is Info type of message, not Warning

be cautious what we log to customer projects ;-)

cezarygerard: also use const from this package
vendor/k8s.io/api/core/v1/types.go

as we do in other places
```
// Valid values for event types (new types could be added in future)
const (
	// Information only and will not cause any problems
	EventTypeNormal string = "Normal"
	// These events are to warn that something might go wrong
	EventTypeWarning string = "Warning"
)
```

pkg/healthchecks/interfaces.go
```diff
@@ -56,3 +58,57 @@ type HealthChecker interface {
...
+func (sfrg *singletonFakeRecorderGetter) FakeRecorder() *record.FakeRecorder {
+	if sfrg.recorder == nil {
+		panic("singletonFakeRecorderGetter not initialised: recorder is nil.")
+	}
+	return sfrg.recorder
+}
+
+func NewFakeSingletonRecorderGetter(bufferSize int) *singletonFakeRecorderGetter {
+	return &singletonFakeRecorderGetter{recorder: record.NewFakeRecorder(bufferSize)}
+}
```
cezarygerard, line 114: all fakes should be moved to _test.go file

pkg/healthchecks/healthchecks.go
DamianSawicki: Great catch!

pkg/healthchecks/healthchecks.go
DamianSawicki: Much better, I somehow missed it!

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go
DamianSawicki: We had this discussion already with kl52752, panslava, and bowei: https://github.com/kubernetes/ingress-gce/pull/2018#discussion_r1145665678. This is the same flag as there and we're not adding any new non-parallelizable tests in this PR, so I'll stick to the approach bowei suggested in https://github.com/kubernetes/ingress-gce/pull/2018 and keep the flag deep down.

pkg/healthchecks/interfaces.go
DamianSawicki: Moved to a new `flakes.go` file because somehow when I moved it to `healthchecks_test.go` it was not visible in other packages.

cezarygerard: yeah... so it has to be compiled into final binary

I have tried to use build constraint 
```
//go:build test
```
or 
```
//go:build testing
```
but it does not work


golang keeps amazing me :-D

cezarygerard (approved):

pkg/healthchecks/interfaces.go
DamianSawicki: https://stackoverflow.com/a/22277807

PR #2073, opened 2023-04-11
Define the interface for NEG dual stack migration handler

gauravkghildiyal:
/assign @swetharepakula 

## Notes for the reviewer
This change includes 3 commits. The last commit is just moving code from one location to another -- that piece of code makes a bulk of this PR and reviewing that again can be skipped.


pkg/neg/syncers/transaction.go
```diff
@@ -459,10 +452,13 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
 			}
 
 			if operation == attachOp {
-				s.attachNetworkEndpoints(zone, batch)
+				go s.attachNetworkEndpoints(zone, batch)
 			}
 			if operation == detachOp {
-				s.detachNetworkEndpoints(zone, batch)
+				if zone == migrationZone {
```
swetharepakula, line 458: add a comment about why we are pausing

pkg/neg/syncers/transaction.go
```diff
@@ -478,22 +474,32 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
...
 
-// detachNetworkEndpoints creates go routine to run operations for detaching network endpoints
-func (s *transactionSyncer) detachNetworkEndpoints(zone string, networkEndpointMap map[negtypes.NetworkEndpoint]*composite.NetworkEndpoint) {
+// detachNetworkEndpoints runs operation for detaching network endpoints.
+func (s *transactionSyncer) detachNetworkEndpoints(zone string, networkEndpointMap map[negtypes.NetworkEndpoint]*composite.NetworkEndpoint, hasMigrationDetachments bool) {
 	s.logger.V(2).Info("Detaching endpoints from NEG.", "countOfEndpointsBeingDetached", len(networkEndpointMap), "negSyncerKey", s.NegSyncerKey.String(), "zone", zone)
-	go s.operationInternal(detachOp, zone, networkEndpointMap)
+	err := s.operationInternal(detachOp, zone, networkEndpointMap)
+
+	if hasMigrationDetachments {
```
swetharepakula, line 491: add comment as to why the migration is continuing

gauravkghildiyal: Done!

pkg/neg/syncers/transaction.go
gauravkghildiyal: Done!

pkg/neg/syncers/dualstack/migrator.go
```diff
@@ -0,0 +1,120 @@
...
+// TODO(gauravkghildiyal): Add details about the heuristics as we go on
+// implementing.
+type Migrator struct {
+	// Setting this to false will make all exported functions a no-op.
+	enableDualStack bool
+}
+
+func NewMigrator(enableDualStack bool) *Migrator {
+	return &Migrator{enableDualStack: enableDualStack}
+}
```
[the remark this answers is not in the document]
gauravkghildiyal: Hmmmmm! That feels a bit hacky personally and introduces another layer of sophistication where one is not required. Eventually we would not need this interface and the cleanup is slightly more involved for when we don't want this flag.

Having said that, if we still feel a strong need for this, I'd prefer doing this later on in a separate PR -- that would greatly help me avoid tedious individual patches to multiple pending commits.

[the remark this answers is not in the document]
swetharepakula: That is reasonable. My thought is this helps make the functions more readable and would also simplify the cleanup in the future. But I am okay with this approach and adding in an interface later if we find that it is better.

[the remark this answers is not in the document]
gauravkghildiyal: Perfect! Thanks for accommodating.

swetharepakula (commented):
/lgtm
/approve

PR #2044, opened 2023-03-28
Refactor syncInternalImpl and toZoneNetworkEndpointMap

sawsa307:
Refactor syncInternalImpl, toZoneNetworkEndpointMap, and toZoneNetworkEndpointMapDegradedMode.
1. Refactor syncInternalImpl. 
- Create convertUntypedToEPS for converting endpointslice, and computeEPSStaleness for compute EPS staleness metrics.
- Create getEndpointsCalculation() to account for endpoint calculation for error state and enableDegradedMode.

2. Refactor toZoneNetworkEndpointMap. 

- Reduce the number of return values by wrapping them in a struct ZoneNetworkEndpointMapResult
- Create helper function getEndpointPod(), getEndpointZone() to simplify endpoint node, pod, and zone checking,
- Refactor endpoint node, pod, and zone checking

3. Refactor toZoneNetworkEndpointMapDegradedMode. 

- Reduce the number of return values by wrapping them in a struct ZoneNetworkEndpointMapResult
- Refactor endpoint node, pod, and zone checking
- Refactor and remove validateAndAddEndpoints since most of the code about validating(https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/utils.go#L337-L357) does not need to be in the for loop
- Update unit tests for validateAndAddEndpoints since this function no longer exists, move the checks to TestToZoneNetworkEndpointMapDegradedMode

PR #2075, opened 2023-04-13
Update endpoint checking return

sawsa307:
Update validatePod to return error so we can get more context on the failing reason.
Change errors from get zone to ErrEPZoneMissing, and raise NodeNotFound error when we validate endpoint's node resource.
Addressing comment https://github.com/kubernetes/ingress-gce/pull/2044#discussion_r1158984260.


swetharepakula (commented):
/lgtm
/approve

PR #2076, opened 2023-04-13
Add label propagation metrics

songrx1997:
/assign @swetharepakula

### DIFF
--- a/pkg/neg/metrics/label_propagation_metrics.go
+++ b/pkg/neg/metrics/label_propagation_metrics.go
@@ -50,6 +50,8 @@ var (
 			Subsystem: negControllerSubsystem,
 			Name:      labelNumber,
 			Help:      "The number of labels per endpoint",
+			// custom buckets - [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
 		},
 	)
 
@@ -57,7 +59,9 @@ var (
 		prometheus.HistogramOpts{
 			Subsystem: negControllerSubsystem,
 			Name:      annotationSize,
-			Help:      "The size of endpoint annotations per endpoint",
+			Help:      "The size in byte of endpoint annotations per endpoint",
+			// custom buckets - [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, +Inf]
+			Buckets: prometheus.ExponentialBuckets(1, 2, 13),
 		},
 	)
 
@@ -81,4 +85,15 @@ type LabelPropagationStats struct {
 type LabelPropagationMetrics struct {
 	EndpointsWithAnnotation int
 	NumberOfEndpoints       int
+}
+
+// PublishLabelPropagationError publishes error occured during label propagation.
+func PublishLabelPropagationError(errType string) {
+	LabelPropagationError.WithLabelValues(errType).Inc()
+}
+
+// PublishAnnotationMetrics publishes collected metrics for endpoint annotations.
+func PublishAnnotationMetrics(annotationSize int, labelNumber int) {
+	AnnotationSize.Observe(float64(annotationSize))
+	LabelNumber.Observe(float64(labelNumber))
 }
--- a/pkg/neg/metrics/label_propagation_metrics.go
+++ b/pkg/neg/metrics/label_propagation_metrics.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	labelNumber       = "label_number_per_endpoint"
+	annotationSize    = "annotation_size_per_endpoint"
+	labelErrorNumber  = "label_propagation_error_count"
+	numberOfEndpoints = "number_of_endpoints"
+)
+
+var (
+	labelPropagationErrorLabels = []string{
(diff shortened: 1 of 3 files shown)


pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+*/
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	ratioWithAnnotation       = "ratio_endpoints_with_annotation"
```
swetharepakula, line 24: instead of ratio, lets say percentage

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+			Name:      averageAnnotationSize,
+			Help:      "The average size of endpoint annotations",
+		},
+	)
+
+	TotalLabelPropagationError = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      totalLabelError,
+			Help:      "The total number of errors of label propagation",
```
swetharepakula, line 70: nit: the total number of errors occurred for label propagation

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+	)
+
+	LabelTruncationErrorCount = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      labelTruncationErrorCount,
+			Help:      "The number of label truncation",
```
swetharepakula, line 78: the number of labels truncated

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	)
+
+	LabelFailureCount = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      labelFailureCount,
+			Help:      "The number of label truncation failures",
+		},
+	)
+)
```
swetharepakula, line 89: can these two metrics be combined and use a label to track truncation failures, truncated, and all labels added?

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+
+var (
+	EndpointsWithAnnotations = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      endpointWithAnnotation,
+			Help:      "The number of endpoints with annotations",
+		},
+	)
```
swetharepakula, line 40: this one lets name `number_of_endpoints` and then you would have a label that say with annotations.

This way we track total and number with annotations with a single metric

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+
+	RatioWithAnnotation = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      ratioWithAnnotation,
+			Help:      "The ratio between the number of endpoints with annotation and the total number",
+		},
+	)
```
swetharepakula, line 48: we don't need to have this metric specifically. with the one above, we can calculate the ratio after the fact.  For metrics, generally we try to stay away from doing any calculations when emitting/publishing them.

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+
+	AverageLabelNumber = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      averageLabelNumber,
+			Help:      "The average number of labels per endpoint",
+		},
+	)
```
swetharepakula, line 56: This one should be a distribution and we just emit the number of labels per endpoint. That way prometheus will do the calculation.

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,89 @@
...
+
+	AverageAnnotationSize = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      averageAnnotationSize,
+			Help:      "The average size of endpoint annotations",
+		},
+	)
```
swetharepakula, line 64: same comment as above, change this to a distribution

pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -30,13 +30,36 @@ type SyncerMetricsCollector interface {
...
+
+type LabelPropagationMetrics struct {
+	TotalLabelError        int
+	LabelTruncationCount   int
+	LabelFailureCount      int
+	EndpointWithAnnotation int
+	RatioWithAnnotation    float64
+	AverageLabelNum        float64
+	AverageAnnotationSize  float64
+}
```
swetharepakula, line 52: the labels should go in the same file as the metrics that will use them

pkg/neg/metrics/neg_metrics_collector.go
```diff
@@ -106,3 +129,42 @@ func (sm *SyncerMetrics) SetSyncerEPMetrics(key negtypes.NegSyncerKey, endpointS
 	}
 	sm.syncerEPSStateMap[key] = endpointStat.EndpointSliceStateCount
 }
+
+// computeLabelMetrics aggregates label propagation metrics.
+func (sm *SyncerMetrics) computeLabelMetrics() LabelPropagationMetrics {
```
swetharepakula, line 134: based on the suggestions I made on the metrics, the only one you have to aggregate and emit periodically are the total Endpoints and endpoints with annotation. The rest you should be able to emit during the sync itself.

pkg/neg/metrics/label_propagation_metrics.go
```diff
@@ -0,0 +1,107 @@
...
+
+	AnnotationSize = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      annotationSize,
+			Help:      "The size of endpoint annotations per endpoint",
+		},
+	)
```
swetharepakula, line 68: we should add custom buckets. Otherwise the max bucket is 10 and anything over is not recorded.

For the annotation we should indicate that this is in bytes

swetharepakula (commented):
/lgtm
/approve

PR #2078, opened 2023-04-17
Add network package containing the NetworkInfo struct

mmamczur:
This struct holds the networking URLs that are now used when creating GCE resources (this commit contains changes for NEGs).
Currently only the default network information is always populated but it will be different for multi-networked services.

Why adding to PortInfo/ServicePort:
1. The controller creates syncers per service port. On update a diff is made with already used Ports vs the ones from the update. Adding the network to PortInfo will make the diff recognize any network change, so it will be possible to shut down the syncers for old network. Updating network for a running syncer would be difficult
2. there can be only one network set for the service so it could make sense to keep it on a service level but this introduced more complicated logic of handling syncers. It is possible that changing network won't be supported but still I find this solution cleaner.

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/ingress-gce/pkg/neg/readiness"
 	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/network"
 	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/endpointslices"
@@ -67,6 +68,7 @@ type Controller struct {
 	namer        negtypes.NetworkEndpointGroupNamer
 	l4Namer      namer2.L4ResourcesNamer
 	zoneGetter   negtypes.ZoneGetter
+	cloud        negtypes.NetworkEndpointGroupCloud
 
 	hasSynced                   func() bool
 	ingressLister               cache.Indexer
@@ -196,6 +198,7 @@ func NewController(
 		gcPeriod:              gcPeriod,
 		recorder:              recorder,
 		zoneGetter:            zoneGetter,
+		cloud:                 cloud,
 		namer:                 namer,
 		l4Namer:               l4Namer,
 		defaultBackendService: defaultBackendService,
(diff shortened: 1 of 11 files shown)


pkg/network/network.go
```diff
@@ -0,0 +1,51 @@
...
+package network
+
+import (
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// ServiceNetwork determines the network data to be used for the LB resources.
+// This function currently returns only the default network but will provide
+// secondary networks information for multi-networked services in the future.
+func ServiceNetwork(_ *apiv1.Service, cloudProvider cloudNetworkProvider) (*NetworkInfo, error) {
```
cezarygerard, line 26: ServiceNetwork => NewNetworkInfo?

pkg/neg/controller.go
```diff
@@ -444,21 +447,25 @@ func (c *Controller) processService(key string) error {
 	}
 	negUsage := usageMetrics.NegServiceState{}
 	svcPortInfoMap := make(negtypes.PortInfoMap)
-	if err := c.mergeDefaultBackendServicePortInfoMap(key, service, svcPortInfoMap); err != nil {
+	serviceNetwork, err := network.ServiceNetwork(service, c.cloud)
```
cezarygerard, line 450: networkInfo

pkg/neg/controller_test.go
```diff
@@ -1030,6 +1010,7 @@ func TestMergeCSMPortInfoMap(t *testing.T) {
 	defer controller.stop()
 	n1s1 := newTestServiceCus(t, controller, "namespace1", "service1", []int32{80, 90})
 	n2s1 := newTestServiceCus(t, controller, "namespace2", "service1", []int32{90})
+	defaultNetwork := &network.NetworkInfo{K8sNetwork: "default"}
```
cezarygerard, line 1013: you can declare this defaultNetwork var at the very top of the file

in the 
```
var (
...
)
```
section

pkg/neg/manager_test.go
```diff
@@ -114,6 +115,10 @@ func TestEnsureAndStopSyncer(t *testing.T) {
 	portName0 := ""
 	portName1 := "http"
 	portName2 := "bar"
+	defaultNetwork := &network.NetworkInfo{
+		NetworkURL:    "defaultNetwork",
+		SubnetworkURL: "defaultSubnetwork",
+	}
```
cezarygerard, line 121: same here
declare 
defaultNetwork := &network.NetworkInfo{... 
as a global var at the beginning of the file

pkg/neg/syncers/transaction.go
```diff
@@ -393,6 +399,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 			s.recorder,
 			s.NegSyncerKey.GetAPIVersion(),
 			s.customName,
+			s.networkInfo,
```
cezarygerard, line 402: why ensureNetworkEndpointGroup is not part of transactionSyncer 😲 ?

NWM...

pkg/neg/syncers/utils.go
```diff
@@ -224,8 +225,8 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 		if networkEndpointType != negtypes.NonGCPPrivateEndpointType &&
 			// Only perform the following checks when the NEGs are not Non-GCP NEGs.
 			// Non-GCP NEGs do not have associated network and subnetwork.
-			(!utils.EqualResourceIDs(neg.Network, cloud.NetworkURL()) ||
-				!utils.EqualResourceIDs(neg.Subnetwork, cloud.SubnetworkURL())) {
+			(!utils.EqualResourceIDs(neg.Network, networkInfo.NetworkURL) ||
+				!utils.EqualResourceIDs(neg.Subnetwork, networkInfo.SubnetworkURL)) {
```
cezarygerard, line 229: this if should be a method

		if isNonGCPNEG(networkEndpointType, neg, networkInfo) {
...

ehhh...

pkg/neg/syncers/utils_test.go
```diff
@@ -616,77 +619,77 @@ func TestEnsureNetworkEndpointGroup(t *testing.T) {
...
-			false,
-		)
+			if !reflect.DeepEqual(*actualNegDesc, expectedNegDesc) {
+				t.Errorf("Unexpected neg description: %s, expected %s", neg.Description, expectedNegDesc.String())
+			}
 
-		if err != nil {
-			t.Errorf("Unexpected error when called with duplicated NEG: %v", err)
-		}
+			// Call ensureNetworkEndpointGroup with the same NEG.
```
cezarygerard, line 686: why you decided do inline the args?

pkg/neg/syncers/transaction.go
cezarygerard: Michał, this is just random comment I had reading the code. 
you can skip it

pkg/neg/syncers/utils.go
cezarygerard: Michał, this is just random comment I had reading the code. 
you can skip it

cezarygerard:
overall I like the idea of NetworkInfo struct

pkg/network/network.go
mmamczur: I could rename it but the main point of this func is to figure out the network values for the service, with funcs named like NewAbc I soft of expect them to not do that much besides creating the struct

with multinetwork all logic to get the GCE VPC and subnetwork will go here

pkg/neg/controller.go

pkg/neg/manager_test.go
mmamczur: good idea, done

pkg/neg/syncers/utils_test.go
mmamczur: this wasn't intended actually, thanks for catching

when you refactor in intellij with "change signature" (it's really helpful) it does this, not sure why, reverted this

pkg/network/network.go
cezarygerard: fair enough, let see how it evolves and decide later

PR #2079, opened 2023-04-17
Fix wait.PollImmediate in TestEnableDegradedMode

sawsa307:
Error state should be checked with syncLock, and we should return false and nil if we are not in the desired state.
We should also check errorState within waitPoll because seeing update in NEG does not guarantee we have also set the syncer's error state.


swetharepakula (commented):
/lgtm
/approve
/ok-to-test

PR #2080, opened 2023-04-17
Use time.Duration.String func to print duration

cezarygerard:
in the debug log after metrics export


mmamczur:
looks ok, let's have that merged so that there is no confusion with microseconds labelled as ms

PR #2081, opened 2023-04-19
Read THC annotation and configure UHC accordingly

DamianSawicki:
Configure UHC with THC defaults when the THC annotation is present for the service. Flag-gated with `flags.F.EnableTransparentHealthChecks`.

### DIFF
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -314,7 +314,7 @@ var (
 	ErrBackendConfigInvalidJSON       = errors.New("BackendConfig annotation is invalid json")
 	ErrBackendConfigAnnotationMissing = errors.New("BackendConfig annotation is missing")
 	ErrNEGAnnotationInvalid           = errors.New("NEG annotation is invalid.")
-	ErrTHCAnnotationInvalid           = errors.New("the annotation for Transparent Health Checker is invalid")
+	ErrTHCAnnotationInvalid           = errors.New("THC annotation is invalid")
 )
 
 // NEGAnnotation returns true if NEG annotation is found.
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -65,6 +65,9 @@ const (
 	// The valid values are "Standard" and "Premium" (default, if unspecified).
 	NetworkTierAnnotationKey = "cloud.google.com/network-tier"
 
+	// THCAnnotationKey is the boolean annotation key to enable Transparent Health Checks.
+	THCAnnotationKey = "networking.gke.io/transparent-health-checker"
+
 	// ProtocolHTTP protocol for a service
 	ProtocolHTTP AppProtocol = "HTTP"
 	// ProtocolHTTPS protocol for a service
@@ -136,6 +139,12 @@ type NegAnnotation struct {
 	// ExposedPorts maps ServicePort to attributes of the NEG that should be
 	// associated with the ServicePort.
 	ExposedPorts map[int32]NegAttributes `json:"exposed_ports,omitempty"`
+}
+
+// THCAnnotation is the format of the annotation associated with the THCAnnotationKey key.
+type THCAnnotation struct {
+	// "enabled" indicates whether to enable the Transparent Health Checks feature.
+	Enabled bool `json:"enabled,omitempty"`
 }
 
 // NegAttributes houses the attributes of the NEGs that are associated with the
@@ -305,6 +314,7 @@ var (
 	ErrBackendConfigInvalidJSON       = errors.New("BackendConfig annotation is invalid json")
 	ErrBackendConfigAnnotationMissing = errors.New("BackendConfig annotation is missing")
 	ErrNEGAnnotationInvalid           = errors.New("NEG annotation is invalid.")
+	ErrTHCAnnotationInvalid           = errors.New("the annotation for Transparent Health Checker is invalid")
 )
 
 // NEGAnnotation returns true if NEG annotation is found.
@@ -322,6 +332,21 @@ func (svc *Service) NEGAnnotation() (*NegAnnotation, bool, error) {
 	}
 
 	return &res, true, nil
+}
+
+// ShouldEnableTHC returns true if a THC annotation is found and its value is true.
+func (svc *Service) ShouldEnableTHC() (bool, error) {
+	var res THCAnnotation
+	annotation, ok := svc.v[THCAnnotationKey]
+	if !ok {
+		return false, nil
+	}
+
+	if err := json.Unmarshal([]byte(annotation), &res); err != nil {
+		return false, ErrTHCAnnotationInvalid
+	}
+
+	return res.Enabled, nil
 }
 
 func (svc *Service) NEGStatus() (*NegStatus, bool, error) {
--- a/pkg/controller/translator/translator.go
+++ b/pkg/controller/translator/translator.go
@@ -66,7 +66,9 @@ func NewTranslator(serviceInformer cache.SharedIndexInformer,
 	nodeInformer cache.SharedIndexInformer,
 	podInformer cache.SharedIndexInformer,
 	endpointSliceInformer cache.SharedIndexInformer,
-	kubeClient kubernetes.Interface) *Translator {
+	kubeClient kubernetes.Interface,
+	enableTHC bool,
+	recorderGetter healthchecks.RecorderGetter) *Translator {
 	return &Translator{
 		serviceInformer,
 		backendConfigInformer,
@@ -74,6 +76,8 @@ func NewTranslator(serviceInformer cache.SharedIndexInformer,
 		podInformer,
 		endpointSliceInformer,
 		kubeClient,
+		enableTHC,
+		recorderGetter,
 	}
 }
 
@@ -85,6 +89,8 @@ type Translator struct {
 	PodInformer           cache.SharedIndexInformer
 	EndpointSliceInformer cache.SharedIndexInformer
 	KubeClient            kubernetes.Interface
+	enableTHC             bool
+	recorderGetter        healthchecks.RecorderGetter
 }
 
 func (t *Translator) getCachedService(id utils.ServicePortID) (*api_v1.Service, error) {
@@ -202,22 +208,32 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 
 // manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
 // Transparent Health Checks should be enabled.
-func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service, recorderGetter healthchecks.RecorderGetter) {
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service) {
+	THCEnabled := false
+	defer func() {
+		klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, THCEnabled)
+		sp.THCEnabled = THCEnabled
+	}()
+
+	if !t.enableTHC {
+		return
+	}
+
+	if flags.F.EnableBackendConfigHealthCheck && sp.BackendConfig != nil && sp.BackendConfig.Spec.HealthCheck != nil {
+		return
+	}
+
 	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
 	if err != nil {
 		message := fmt.Sprintf("Parsing THC annotation failed: %+v.", err)
-		if flags.F.EnableTransparentHealthChecks {
-			recorderGetter.Recorder(sp.ID.Service.Namespace).Event(svc, api_v1.EventTypeWarning, "THCAnnotationParsingFailed", message)
-		}
+		t.recorderGetter.Recorder(sp.ID.Service.Namespace).Event(svc, api_v1.EventTypeWarning, "THCAnnotationParsingFailed", message)
 		klog.Warning(message)
 	}
-	klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, THCEnabled)
-	sp.THCEnabled = THCEnabled
 }
 
 // getServicePort looks in the svc store for a matching service:port,
 // and returns the nodeport.
-func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePortParams, namer namer_util.BackendNamer, recorderGetter healthchecks.RecorderGetter) (*utils.ServicePort, error) {
+func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePortParams, namer namer_util.BackendNamer) (*utils.ServicePort, error) {
 	svc, err := t.getCachedService(id)
 	if err != nil {
 		return nil, err
@@ -259,13 +275,13 @@ func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePo
 		return svcPort, err
 	}
 
-	t.manageEnableTHC(svcPort, svc, recorderGetter)
+	t.manageEnableTHC(svcPort, svc)
 
 	return svcPort, nil
 }
 
 // TranslateIngress converts an Ingress into our internal UrlMap representation.
-func (t *Translator) TranslateIngress(ing *v1.Ingress, systemDefaultBackend utils.ServicePortID, namer namer_util.BackendNamer, recorderGetter healthchecks.RecorderGetter) (*utils.GCEURLMap, []error) {
+func (t *Translator) TranslateIngress(ing *v1.Ingress, systemDefaultBackend utils.ServicePortID, namer namer_util.BackendNamer) (*utils.GCEURLMap, []error) {
 	var errs []error
 	urlMap := utils.NewGCEURLMap()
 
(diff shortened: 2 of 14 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -85,12 +88,16 @@ func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (
 	if bchcc != nil {
 		klog.V(2).Infof("ServicePort %v has BackendConfig healthcheck override", sp.ID)
 	}
-	return h.sync(hc, bchcc)
+	thcEnabled := flags.F.EnableTransparentHealthChecks && sp.THCEnabled
```
kl52752, line 91: Is it needed to check the flag here? I thought that sp/THCEnabled already checked the flag (with the annotation check).

pkg/controller/translator/translator.go
```diff
@@ -192,6 +192,18 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 	return nil
 }
 
+// manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// Transparent Health Checks should be enabled.
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service) error {
```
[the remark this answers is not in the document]
DamianSawicki: Discussed offline, won't change.

pkg/controller/translator/translator.go
```diff
@@ -236,6 +248,10 @@ func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePo
 		return svcPort, err
 	}
 
+	if err := t.manageEnableTHC(svcPort, svc); err != nil {
```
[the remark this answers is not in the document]
DamianSawicki: > I don't think we want to return the error for this. I think that the only case THC can fail is when we cannot parse the annotation.

Makes sense. We probably do not want to fail the entire `getServicePort` because of an incorrect THC annotation, right?

> Maybe it would be better to set THC to falce, log the error and emit the event to User with error. WDYT?

Logging certainly makes sense and if this function won't be called too frequently, then emitting an event is a great idea.

pkg/healthchecks/healthchecks.go
DamianSawicki: Good point!

pkg/annotations/service.go
```diff
@@ -324,6 +335,29 @@ func (svc *Service) NEGAnnotation() (*NegAnnotation, bool, error) {
...
+	}
+
+	if err := json.Unmarshal([]byte(annotation), &res); err != nil {
+		return false, ErrTHCAnnotationInvalid
+	}
+
+	return res.Enabled, nil
+}
+
+func (svc *Service) ShouldEnableTHC() (bool, error) {
```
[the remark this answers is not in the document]
DamianSawicki: I'm removing the global flag and adding a bool field to the `struct healthchecks.HealthChecks`.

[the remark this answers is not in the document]
DamianSawicki: Done (the second part, so the first part is void).

pkg/controller/translator/translator.go
```diff
@@ -192,6 +192,18 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
 }
 
+// manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// Transparent Health Checks should be enabled.
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service) error {
+	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
+	if err != nil {
+		return err
+	}
+	klog.Infof("Is THC enabled for %v? %v", svc.Name, THCEnabled)
```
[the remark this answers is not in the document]
DamianSawicki: Done.

pkg/healthchecks/healthchecks.go
```diff
@@ -135,15 +138,22 @@ func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (
 		bchcc = sp.BackendConfig.Spec.HealthCheck
 		klog.V(2).Infof("ServicePort (%v) has BackendConfig health check override (%+s)", sp.ID, formatBackendConfigHC(bchcc))
 	}
+	thcEnabled := false
 	if bchcc != nil {
 		klog.V(2).Infof("ServicePort %v has BackendConfig healthcheck override", sp.ID)
+	} else {
+		thcEnabled = h.thcFlagEnabled && sp.THCEnabled
```
kl52752, line 145: we should use sp.THCEnable - this flag should include bchcc

pkg/controller/translator/translator.go
```diff
@@ -199,9 +200,24 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 	return nil
 }
 
+// manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// Transparent Health Checks should be enabled.
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service, recorderGetter healthchecks.RecorderGetter) {
+	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
```
kl52752, line 206: In this function you should check if  BackendConfig for hc is set and include it in  sp.THCEnabled

pkg/healthchecks/healthchecks.go
```diff
@@ -125,6 +126,8 @@ func (h *HealthChecks) generateHealthcheckInfo(sp utils.ServicePort, iLB bool) h
 
 // SyncServicePort implements HealthChecker.
 func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (string, error) {
+	klog.Infof("SyncServicePort: ID=%v, NodePort=%v, Port=%v, PortName=%v, THCEnabled=%v.", sp.ID, sp.NodePort, sp.Port, sp.PortName, sp.THCEnabled)
```
kl52752, line 129: this log is inconsistent with line 145

pkg/healthchecks/healthchecks.go
kl52752: why do you check this flag here? I think that you should only check for THC annotation when flag is enabled.

pkg/controller/translator/translator.go
```diff
@@ -199,9 +200,24 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 	return nil
 }
 
+// manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// Transparent Health Checks should be enabled.
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service, recorderGetter healthchecks.RecorderGetter) {
```
kl52752, line 205: either this function should be guarded with feature flag or should do early exit when flag is turned off

pkg/controller/translator/translator.go
```diff
@@ -243,11 +259,13 @@ func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePo
...
 	}
 
+	t.manageEnableTHC(svcPort, svc, recorderGetter)
+
 	return svcPort, nil
 }
 
 // TranslateIngress converts an Ingress into our internal UrlMap representation.
-func (t *Translator) TranslateIngress(ing *v1.Ingress, systemDefaultBackend utils.ServicePortID, namer namer_util.BackendNamer) (*utils.GCEURLMap, []error) {
+func (t *Translator) TranslateIngress(ing *v1.Ingress, systemDefaultBackend utils.ServicePortID, namer namer_util.BackendNamer, recorderGetter healthchecks.RecorderGetter) (*utils.GCEURLMap, []error) {
```
kl52752, line 268: Why not store recorder in translator and then you don't have to pass it everywhere in tests.

pkg/healthchecks/healthchecks.go
```diff
@@ -182,6 +192,14 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 	if bchcc != nil {
 		// BackendConfig healthcheck settings always take precedence.
 		hc.UpdateFromBackendConfig(bchcc)
+		if thcEnabled {
+			klog.Warningf("BackendConfig exists and thcEnabled simultaneously for %v. Ignoring transparent health check.", hc.Name)
+		}
+	} else {
```
kl52752, line 198: I don't like this nested ifs. IMO if THC is enabled and BC is not set you should create hc *translator.HealthCheck with THC params and omit this else.

pkg/translator/healthchecks_test.go
```diff
@@ -84,3 +86,27 @@ func TestMerge(t *testing.T) {
...
+func TestApplyTHCSettings(t *testing.T) {
+	hc := &HealthCheck{HealthCheck: computealpha.HealthCheck{Type: "HTTP"}}
+	wantHC := &HealthCheck{
+		HealthCheck: computealpha.HealthCheck{
+			CheckIntervalSec:   5,
+			TimeoutSec:         5,
+			UnhealthyThreshold: 10,
+			HealthyThreshold:   1,
+			Type:               "HTTP",
+			Description:        "Kubernetes L7 transparent health check.",
```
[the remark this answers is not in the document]
DamianSawicki: Yes, so far I just rebased the PR after the JSONification was merged, and was going to JSONify THC soon.

pkg/healthchecks/healthchecks.go
DamianSawicki: I see that "ApplyTHCSettings" is not a good name for a function that overwrites everything. And I agree that a function that overwrites everything should really be a constructor, not a setter.

I've added these ifs in case somebody misuses the function, passing both non-nil bchcc and thcEnabled true. Perhaps this check can just be done at the beginning of the function, and then we get a smooth flow afterwards.

pkg/controller/translator/translator.go
DamianSawicki: Sounds nice, I'll check this possibility. Although this will need to store a RecorderGetter rather than a Recorder because Recorder depends on the namespace and I see many namespaces (`default` and `kube-system`) used with one translator in `TestTranslateIngress`.

pkg/controller/translator/translator.go
DamianSawicki: Done.

pkg/controller/translator/translator.go
DamianSawicki: Done.

pkg/healthchecks/healthchecks.go
DamianSawicki: Modified, PTAL.

pkg/healthchecks/healthchecks.go
DamianSawicki: > we should use sp.THCEnable - this flag should include bchcc

Done.

> why do you check this flag here? I think that you should only check for THC annotation when flag is enabled.

Had a different approach as explained offline, but changed my mind in the end and done.

pkg/healthchecks/healthchecks.go
DamianSawicki: Moved the flag check to `translator` in the end, where the annotation is checked -- as suggested earlier by kl52752.

pkg/healthchecks/healthchecks.go
```diff
@@ -125,7 +129,17 @@ func (h *HealthChecks) generateHealthcheckInfo(sp utils.ServicePort, iLB bool) h
 
 // SyncServicePort implements HealthChecker.
 func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (string, error) {
+	klog.Infof("SyncServicePort: sp.ID=%v, sp.NodePort=%v, sp.Port=%v, sp.PortName=%v, sp.THCEnabled=%v, h.thcFlagEnabled=%v.", sp.ID, sp.NodePort, sp.Port, sp.PortName, sp.THCEnabled, h.thcFlagEnabled)
+	if !h.thcFlagEnabled && sp.THCEnabled {
```
kl52752, line 133: right now after changes in `manageEnableTHC` this state is impossible right?

DamianSawicki: Correct. I added this sanity check because after the last changes, the field `HealthChecks.thcFlagEnabled` isn't used anywhere in the `healthchecks` package, which I find weird.

kl52752: Yes so the check is fine but maybe you can move it to `manageEnableTHC` and emit the event for Customer with error (to avoid Customer Isses?)

DamianSawicki: The check here is for developer errors, not configuration errors. I can add some configuration checks to `manageEnableTHC()`.

pkg/healthchecks/healthchecks.go
```diff
@@ -47,14 +47,15 @@ type HealthChecks struct {
 	recorderGetter    RecorderGetter
 	serviceGetter     ServiceGetter
 	clusterInfo       healthcheck.ClusterInfo
+	thcFlagEnabled    bool
```
swetharepakula, line 50: nit, this can be reduced to `thcEnabled`

pkg/controller/translator/translator.go
```diff
@@ -199,6 +206,38 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 	return nil
 }
 
+// manageEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// Transparent Health Checks should be enabled.
+func (t *Translator) manageEnableTHC(sp *utils.ServicePort, svc *api_v1.Service) {
```
swetharepakula, line 211: for the name perhaps `setEnableTHC` instead of manage. It isn't always clear what manage means.

pkg/controller/translator/translator.go
```diff
@@ -199,6 +206,38 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
+		return
+	}
+
+	if flags.F.EnableBackendConfigHealthCheck && sp.BackendConfig != nil && sp.BackendConfig.Spec.HealthCheck != nil {
+		return
+	}
+
+	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
+	if err != nil {
+		message := fmt.Sprintf("Parsing THC annotation failed: %+v.", err)
```
swetharepakula, line 228: Confirming that we want to an error event both on the Service and in the Ingress, since every error in TranslateIngress will be emitted as an error event on the Ingress resource as well.

pkg/controller/translator/translator_test.go
```diff
@@ -1161,3 +1166,136 @@ func TestSetTrafficScaling(t *testing.T) {
...
+		sp          *utils.ServicePort
+		svc         *apiv1.Service
+		enableTHC   bool
+		want        *utils.ServicePort
+		wantEvent   bool
+		eventPrefix string
+	}
+
+	const (
+		thcLabel = "networking.gke.io/transparent-health-checker"
```
swetharepakula, line 1196: you can use the constant defined in the annotations package directly

pkg/controller/translator/translator_test.go
```diff
@@ -1161,3 +1166,136 @@ func TestSetTrafficScaling(t *testing.T) {
...
+			name:        "annotation NEG disabled",
+			sp:          &utils.ServicePort{THCEnabled: true},
+			svc:         newService(map[string]string{thcLabel: thcValue}),
+			enableTHC:   true,
+			want:        &utils.ServicePort{THCEnabled: false},
+			wantEvent:   true,
+			eventPrefix: "Warning THCAnnotationWithoutNEG THC annotation present, but NEG is disabled. Will not enable Transparent Health Checks.",
+		},
+		{
+			name:      "wrong annotation flag disabled",
```
swetharepakula, line 1239: nit: "invalid" instead of "wrong"

pkg/controller/translator/translator_test.go
```diff
@@ -1161,3 +1166,136 @@ func TestSetTrafficScaling(t *testing.T) {
...
+			sp:        &utils.ServicePort{NEGEnabled: true, THCEnabled: true},
+			svc:       newService(map[string]string{thcLabel: "random text"}),
+			enableTHC: false,
+			want:      &utils.ServicePort{NEGEnabled: true, THCEnabled: false},
+		},
+		{
+			name:        "wrong annotation",
```
swetharepakula, line 1246: same comment as above

swetharepakula (commented):
Mostly nit s and small suggestions

pkg/controller/translator/translator.go
DamianSawicki: Thanks for letting me know. I would prefer not to skip a Warning log and not to skip a service event. So because of the coupling of logs with Ingress events, I'd keep both events. CC: @kl52752.

pkg/controller/translator/translator_test.go
DamianSawicki: Fixed. But shouldn't we keep the code and tests independent so that an accidental change in the code is caught by the tests?

pkg/controller/translator/translator_test.go
DamianSawicki: Done.

PR #2083, opened 2023-04-20
Label propagation e2e test

songrx1997:
/assign @swetharepakula

### DIFF
--- a/cmd/e2e-test/neg_test.go
+++ b/cmd/e2e-test/neg_test.go
@@ -778,3 +778,99 @@ func TestNegDisruptive(t *testing.T) {
 		}
 	})
 }
+
+func TestLabelPropagation(t *testing.T) {
+	t.Parallel()
+
+	svcName := "label-propagation-service"
+
+	for _, tc := range []struct {
+		desc              string
+		podLabels         map[string]string
+		expectAnnotations map[string]string
+	}{
+		{
+			desc:              "no labels on the pod",
+			podLabels:         map[string]string{},
+			expectAnnotations: nil,
+		},
+		{
+			desc: "tlsMode with value istio",
+			podLabels: map[string]string{
+				"security.istio.io/tlsMode": "istio",
+			},
+			expectAnnotations: map[string]string{
+				"itls": "istio",
+			},
+		},
+		{
+			desc: "tlsMode with value disabled",
+			podLabels: map[string]string{
+				"security.istio.io/tlsMode": "disabled",
+			},
+			expectAnnotations: map[string]string{
+				"itls": "disabled",
+			},
+		},
+		{
+			desc: "truncated label value",
+			podLabels: map[string]string{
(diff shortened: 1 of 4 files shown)


cmd/e2e-test/neg_test.go
```diff
@@ -778,3 +778,99 @@ func TestNegDisruptive(t *testing.T) {
 		}
 	})
 }
+
+func TestNEGAnnotations(t *testing.T) {
```
swetharepakula, line 782: we should name it differently so it doesn't come under the regex of the previous NEG tests.

`TestLabelPropagation`

cmd/e2e-test/neg_test.go
```diff
@@ -778,3 +778,99 @@ func TestNegDisruptive(t *testing.T) {
...
+			desc: "tlsMode with value disabled",
+			podLabels: map[string]string{
+				"security.istio.io/tlsMode": "disabled",
+			},
+			expectAnnotations: map[string]string{
+				"security.istio.io/tlsMode": "disabled",
+			},
+		},
+		{
+			desc: "Invalid label value",
```
swetharepakula, line 816: nit: truncated label value

cmd/e2e-test/neg_test.go
```diff
@@ -778,3 +778,99 @@ func TestNegDisruptive(t *testing.T) {
...
+
+	svcName := "service-1"
+
+	for _, tc := range []struct {
+		desc              string
+		podLabels         map[string]string
+		expectAnnotations map[string]string
+	}{
+		{
+			desc:              "No required label in pod",
```
swetharepakula, line 793: nit: no labels on the pod

pkg/e2e/fixtures.go
```diff
@@ -171,16 +175,17 @@ func DeleteService(s *Sandbox, svcName string) error {
...
 // EnsureEchoDeployment ensures that the Echo deployment with the given description is set up
 func EnsureEchoDeployment(s *Sandbox, name string, numReplicas int32, modify func(deployment *apps.Deployment)) error {
-	return ensureEchoDeployment(s, name, numReplicas, modify, Linux)
+	return ensureEchoDeployment(s, name, numReplicas, modify, Linux, map[string]string{})
 }
 
-func ensureEchoDeployment(s *Sandbox, name string, numReplicas int32, modify func(deployment *apps.Deployment), os OS) error {
+func ensureEchoDeployment(s *Sandbox, name string, numReplicas int32, modify func(deployment *apps.Deployment), os OS, podLabels map[string]string) error {
 	image := echoheadersImage
+	podLabels["app"] = name
```
swetharepakula, line 188: the only issue with is that you are editing the podLabels that are being provided to the ensureEchoDeployment since maps are pointers.

You probably have to make a duplicate one and add in the labels because we don't know what validations could be happening on the podLabels outside of this function.

swetharepakula (commented):
/lgtm
/approve

PR #2086, opened 2023-04-26
use cloudbuild to build container images

aojea:
This is the canonical and supported way to build images in kubernetes

https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing#image-pushing-jobs


aojea:
reference https://github.com/kubernetes/test-infra/pull/29371

PR #2088, opened 2023-04-26
Add firewall rule for Transparent Health Checks

DamianSawicki:
The purpose is to create a firewall rule enabling traffic to port 7877 serving transparent health checks.

~Child PR dependent on https://github.com/kubernetes/ingress-gce/pull/2081, please review the commit titled "Add firewall rule for Transparent Health Checks".~

### DIFF
--- a/pkg/firewalls/controller.go
+++ b/pkg/firewalls/controller.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/ingress-gce/pkg/controller/translator"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
+	pkg_translator "k8s.io/ingress-gce/pkg/translator"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
 )
@@ -255,10 +256,8 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 
 	var additionalPorts []string
-	if flags.F.EnableBackendConfigHealthCheck {
-		hcPorts := fwc.getCustomHealthCheckPorts(gceSvcPorts)
-		additionalPorts = append(additionalPorts, hcPorts...)
-	}
+	hcPorts := fwc.getCustomHealthCheckPorts(gceSvcPorts)
+	additionalPorts = append(additionalPorts, hcPorts...)
 	additionalPorts = append(additionalPorts, negPorts...)
 
 	// Ensure firewall rule for the cluster and pass any NEG endpoint ports.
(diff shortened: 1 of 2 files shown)


pkg/translator/healthchecks.go
```diff
@@ -59,6 +59,16 @@ const (
 	// defaultNEGTimeout defines the timeout of each probe for NEG
 	defaultNEGTimeout = 15 * time.Second
 
+	// Similar defaults as above, but for Transparent Health Checks.
+	defaultTHCCheckIntervalSec   = 5
+	defaultTHCTimeoutSec         = 5
+	defaultTHCUnhealthyThreshold = 10
```
kl52752, line 65: this one is equal to defaultUnhealthyThreshold so I don't think that there ia need to create new default

pkg/translator/healthchecks.go
```diff
@@ -59,6 +59,16 @@ const (
 	// defaultNEGTimeout defines the timeout of each probe for NEG
 	defaultNEGTimeout = 15 * time.Second
 
+	// Similar defaults as above, but for Transparent Health Checks.
+	defaultTHCCheckIntervalSec   = 5
+	defaultTHCTimeoutSec         = 5
```
kl52752, line 64: maybe it will be better to use the notation as above
defaultTHCCheckInterval   = 5 * time.Second

DamianSawicki: Done.

pkg/translator/healthchecks.go
DamianSawicki: I made it explicit that defaultTHCUnhealthyThreshold equals defaultUnhealthyThreshold.

pkg/firewalls/controller.go
```diff
@@ -301,11 +300,16 @@ func (fwc *FirewallController) ilbFirewallSrcRange(gceIngresses []*v1.Ingress) (
 func (fwc *FirewallController) getCustomHealthCheckPorts(svcPorts []utils.ServicePort) []string {
 	var result []string
 
-	for _, svcPort := range svcPorts {
-		if svcPort.BackendConfig != nil && svcPort.BackendConfig.Spec.HealthCheck != nil && svcPort.BackendConfig.Spec.HealthCheck.Port != nil {
-			result = append(result, strconv.FormatInt(*svcPort.BackendConfig.Spec.HealthCheck.Port, 10))
+	if flags.F.EnableBackendConfigHealthCheck {
```
swetharepakula, line 303: we likely can get rid of this flag. It has been in the code base for a while. But not a blocker for this PR

pkg/firewalls/controller.go
```diff
@@ -255,10 +256,8 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 
 	var additionalPorts []string
-	if flags.F.EnableBackendConfigHealthCheck {
```
swetharepakula, line 258: why is the flag being removed here?

swetharepakula (commented):
This looks good to me. Just one question about removing the flag in part of the code.

pkg/firewalls/controller.go
DamianSawicki: It is rather moved than removed. It's moved to the second line of the function `getCustomHealthCheckPorts`. I moved it because after this change the content of `hcPorts` depends on two flags, `EnableBackendConfigHealthCheck` and `EnableTransparentHealthChecks`, so `hcPorts` can be non-empty even if `EnableBackendConfigHealthCheck` is false.

pkg/firewalls/controller.go
DamianSawicki: Added to my backlog.

PR #2089, opened 2023-04-26
Add validation for ipv6 only service

sawsa307:
Return error when service uses ipv6 only. Remove this validation when single stack ipv6 endpoint is supported.

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -445,6 +445,10 @@ func (c *Controller) processService(key string) error {
 	if service == nil {
 		return fmt.Errorf("cannot convert to Service (%T)", obj)
 	}
+	// TODO(cheungdavid): Remove this validation when single stack ipv6 endpoint is supported
+	if service.Spec.Type != apiv1.ServiceTypeLoadBalancer && isSingleStackIPv6Service(service) {
+		return fmt.Errorf("NEG is not supported for ipv6 only service (%T)", service)
+	}
 	negUsage := usageMetrics.NegServiceState{}
 	svcPortInfoMap := make(negtypes.PortInfoMap)
 	networkInfo, err := network.ServiceNetwork(service, c.cloud)
@@ -847,4 +851,15 @@ func gatherPortMappingFromService(svc *apiv1.Service) negtypes.SvcPortTupleSet {
 		})
 	}
 	return svcPortTupleSet
+}
+
+// isSingleStackIPv6Service returns true if the given service is a single stack ipv6 service
+func isSingleStackIPv6Service(service *apiv1.Service) bool {
+	if service.Spec.IPFamilyPolicy != nil && *service.Spec.IPFamilyPolicy != apiv1.IPFamilyPolicySingleStack {
+		return false
+	}
+	if len(service.Spec.IPFamilies) == 1 && service.Spec.IPFamilies[0] == apiv1.IPv6Protocol {
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


sawsa307:
cc @gauravkghildiyal

pkg/neg/controller.go
```diff
@@ -445,6 +445,10 @@ func (c *Controller) processService(key string) error {
 	if service == nil {
 		return fmt.Errorf("cannot convert to Service (%T)", obj)
 	}
+	// TODO(cheungdavid): Remove this validation when single stack ipv6 endpoint is supported
```
swetharepakula, line 448: we should add a test for this

pkg/neg/controller.go
```diff
@@ -445,6 +445,10 @@ func (c *Controller) processService(key string) error {
...
+	if len(service.Spec.IPFamilies) == 1 && service.Spec.IPFamilies[0] == apiv1.IPv6Protocol {
+		return fmt.Errorf("NEG is not supported for ipv6 only service (%T)", service)
```
swetharepakula, line 450: this should only be for L7 load balancing

pkg/neg/controller.go
sawsa307: Added TestServiceIPFamilies.

pkg/neg/controller.go
```diff
@@ -445,6 +445,10 @@ func (c *Controller) processService(key string) error {
 	if service == nil {
 		return fmt.Errorf("cannot convert to Service (%T)", obj)
 	}
+	// TODO(cheungdavid): Remove this validation when single stack ipv6 endpoint is supported
+	if service.Spec.Type == apiv1.ServiceTypeClusterIP && len(service.Spec.IPFamilies) == 1 && service.Spec.IPFamilies[0] == apiv1.IPv6Protocol {
```
swetharepakula, line 449: instead we should look for not type load balancer. Negs can be added for NodePort services as well.

pkg/neg/controller.go
swetharepakula: Instead of enumerating through the different types, we should just check if if ipv6 and !service.spec.type != LB

sawsa307: Updated. Thanks!

swetharepakula: Looking at the docs for dual stack: https://kubernetes.io/docs/concepts/services-networking/dual-stack/

We need to check 2 things, 1) is the ipfamilypolicy ipv6 and 2) if the first element in the ipfamilies is ipv6.

Might be easier to create a helper function that determines if it is a single stack ipv6 service.

sawsa307: > if the first element in the ipfamilies is ipv6
I think we should check if ipv6 is the only element since the order doesn't matter.

sawsa307: Created. Please take a look.

pkg/neg/controller.go
```diff
@@ -848,3 +852,14 @@ func gatherPortMappingFromService(svc *apiv1.Service) negtypes.SvcPortTupleSet {
 	}
 	return svcPortTupleSet
 }
+
+// isSingleStackIPv6Service returns true if the given service is a single stack ipv6 service
+func isSingleStackIPv6Service(service *apiv1.Service) bool {
+	if service.Spec.IPFamilyPolicy != nil && *service.Spec.IPFamilyPolicy == apiv1.IPFamilyPolicyRequireDualStack {
```
swetharepakula, line 858: IPFamilyPolicyPreferDualStack is also a possible option.

This check should instead be `!= apiv1.IPFamilyPolicySingleStack`

swetharepakula (commented):
/lgtm
/approve
/ok-to-test

PR #2090, opened 2023-04-27
Implement heuristics for detaching migration endpoints

gauravkghildiyal:
/assign @swetharepakula

### DIFF
--- a/pkg/neg/syncers/dualstack/migrator.go
+++ b/pkg/neg/syncers/dualstack/migrator.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dualstack
 
 import (
+	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/dualstack/migrator.go
```diff
@@ -73,10 +92,13 @@ type syncable interface {
 
 func NewMigrator(enableDualStackNEG bool, syncer syncable, logger klog.Logger) *Migrator {
 	return &Migrator{
-		enableDualStack:       enableDualStackNEG,
-		syncer:                syncer,
-		migrationWaitDuration: defaultMigrationWaitDuration,
-		logger:                logger.WithName("DualStackMigrator"),
+		enableDualStack:                       enableDualStackNEG,
+		syncer:                                syncer,
+		migrationWaitDuration:                 defaultMigrationWaitDuration,
```
swetharepakula, line 97: i am wondering if we should be using a fake clock here to make the testing easier.

gauravkghildiyal: Yeah I just thought today that it may actually be useful. I plan on doing that in another PR which I am preparing. But that is only useful for the Pause/Continue and we should still be able to review this PR as it is _mostly_ (not completely) independent of that.

gauravkghildiyal: tl;dr I agree yes completely.

swetharepakula: a different PR to change this sounds good to me

swetharepakula (commented):
/lgtm
/approve

PR #2092, opened 2023-04-28
Use Alpha NEG API when DualStackNEG feature is enabled

gauravkghildiyal:
This is required to support new IPv6 fields within that are currently only present in Alpha APIs

/assign @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #2097, opened 2023-05-02
Track GCE/K8s server error

sawsa307:
Define `NegControllerErrorCount` that tracks both internal sync error and API server errors(GCE/K8s).

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -149,10 +149,12 @@ func NewController(
 	err := scheme.AddToScheme(negScheme)
 	if err != nil {
 		logger.Error(err, "Errored adding default scheme to event recorder")
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 	}
 	err = svcnegv1beta1.AddToScheme(negScheme)
 	if err != nil {
 		logger.Error(err, "Errored adding NEG CRD scheme to event recorder")
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 	}
 	recorder := eventBroadcaster.NewRecorder(negScheme,
 		apiv1.EventSource{Component: "neg-controller"})
@@ -401,6 +403,7 @@ func (c *Controller) processEndpoint(key string) {
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		c.logger.Error(err, "Failed to split endpoint namespaced key", "key", key)
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 		return
 	}
 	c.manager.Sync(namespace, name)
@@ -416,6 +419,7 @@ func (c *Controller) serviceWorker() {
 			defer c.serviceQueue.Done(key)
 			err := c.processService(key.(string))
 			c.handleErr(err, key)
+			metrics.PublishNegControllerErrorCountMetrics(err, false)
 		}()
 	}
 }
@@ -713,6 +717,7 @@ func (c *Controller) handleErr(err error, key interface{}) {
 	c.logger.Error(nil, msg)
 	if service, exists, err := c.serviceLister.GetByKey(key.(string)); err != nil {
 		c.logger.Error(err, "Failed to retrieve service from store", "service", key.(string))
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 	} else if exists {
 		c.recorder.Eventf(service.(*apiv1.Service), apiv1.EventTypeWarning, "ProcessServiceFailed", msg)
 	}
@@ -735,6 +740,7 @@ func (c *Controller) enqueueEndpointSlice(obj interface{}) {
 	key, err := endpointslices.EndpointSlicesServiceKey(endpointSlice)
 	if err != nil {
 		c.logger.Error(err, "Failed to find a service label inside endpoint slice", "endpointSlice", klog.KObj(endpointSlice))
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 		return
 	}
 	c.logger.V(3).Info("Adding EndpointSlice to endpointQueue for processing", "endpointSlice", key)
@@ -745,6 +751,7 @@ func (c *Controller) enqueueNode(obj interface{}) {
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
 		c.logger.Error(err, "Failed to generate node key")
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 		return
 	}
 	c.logger.V(3).Info("Adding Node to nodeQueue for processing", "node", key)
@@ -755,6 +762,7 @@ func (c *Controller) enqueueService(obj interface{}) {
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
 		c.logger.Error(err, "Failed to generate service key")
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
 		return
 	}
 	c.logger.V(3).Info("Adding Service to serviceQueue for processing", "service", key)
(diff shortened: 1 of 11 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -233,9 +253,35 @@ func PublishDegradedModeCorrectnessMetrics(count int, endpointType string, negTy
...
+		if utils.IsK8sServerError(err) {
+			return k8sServerError
+		}
+		return OtherError
+	}
+
+	// Here we track internal sync errors.
+	syncErr := negtypes.ClassifyError(err)
+	return string(syncErr.Reason)
+}
```
sawsa307, line 287: We could make this part cleaner by doing the following: 
`func getErrorType(err error) string {
    if utils.IsGCEServerError(err) {
        return gceServerError
    }
    if utils.IsK8sServerError(err) {
        return k8sServerError
    }
    syncErr := negtypes.ClassifyError(err)
    return string(syncErr.Reason)
}`
but it might raise confusion since K8s/GCE errors(that are not server error) are not sync errors either.

pkg/neg/metrics/metrics.go
```diff
@@ -49,6 +52,10 @@ const (
 	ipv6EndpointType      = "IPv6"
 	dualStackEndpointType = "DualStack"
 	migrationEndpointType = "Migration"
+
+	gceServerError    = "GCE_server_err"
+	k8sServerError    = "K8s_server_err"
+	internalSyncError = "internal_sync_err"
```
swetharepakula, line 58: should this be other error instead?

We also need a total

pkg/neg/controller.go
```diff
@@ -678,7 +678,9 @@ func (c *Controller) syncNegStatusAnnotation(namespace, name string, portMap neg
 			newSvcObjectMeta := service.ObjectMeta.DeepCopy()
 			delete(newSvcObjectMeta.Annotations, annotations.NEGStatusKey)
 			c.logger.V(2).Info("Removing NEG status annotation from service", "service", klog.KRef(namespace, name))
-			return patch.PatchServiceObjectMetadata(c.client.CoreV1(), service, *newSvcObjectMeta)
+			err := patch.PatchServiceObjectMetadata(c.client.CoreV1(), service, *newSvcObjectMeta)
```
swetharepakula, line 681: instead of individually doing this, would it be easier to publish the error count metric in the service worker function, where the service is processed?

swetharepakula (commented):
Some of these errors are returned at a central point, should we just emit the metric then?

pkg/neg/metrics/metrics.go
```diff
@@ -211,6 +218,17 @@ var (
 		},
 		[]string{"endpoint_type"},
 	)
+
+	// NegControllerErrorCount tracks the count of server errors(GCE/K8s) and
+	// all errors from NEG controller.
+	NegControllerErrorCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      "neg_controller_error_count",
```
swetharepakula, line 227: negControllerSubsystem already has neg_controller.

The Name should just be `error_count`

pkg/neg/syncers/transaction.go
```diff
@@ -573,6 +573,7 @@ func (s *transactionSyncer) commitTransaction(err error, networkEndpointMap map[
 		// This is to prevent if the NEG object is deleted or misconfigured by user
 		s.needInit = true
 		needRetry = true
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
```
swetharepakula, line 576: isn't this error already being counted? This would be counted where commitTransaction occurs?

swetharepakula: Perhaps we should count the metrics at the end of the sync instead?

swetharepakula (commented):
The errors in transaction.go that result in a failed sync should not be an ignored error

pkg/neg/controller.go
sawsa307: Updated. Thanks!

pkg/neg/metrics/metrics.go
sawsa307: Updated. Thanks!

pkg/neg/syncers/transaction.go
sawsa307: > Perhaps we should count the metrics at the end of the sync instead?
End of sync metrics are collected in syncer.go where we called s.core.sync(). Here we are collecting errors from each goroutine spawned by attach/detach calls.

pkg/neg/metrics/metrics.go
sawsa307: Updated. Thanks!

sawsa307:
> The errors in transaction.go that result in a failed sync should not be an ignored error

Understood. So the definition for `ignored_error` are errors that are not being handled is that correct?

pkg/neg/metrics/metrics.go
```diff
@@ -49,6 +50,12 @@ const (
 	ipv6EndpointType      = "IPv6"
 	dualStackEndpointType = "DualStack"
 	migrationEndpointType = "Migration"
+
+	gceServerError          = "GCE_server_error"
+	k8sServerError          = "K8s_server_error"
+	ignoredControllerError  = "ignored_controller_error"
+	otherControllerError    = "other_controller_error"
+	totalNegControllerError = "total_neg_controller_error"
```
swetharepakula, line 58: nit: remove controller

pkg/neg/syncers/dualstack/migrator.go
```diff
@@ -186,6 +187,7 @@ func (d *Migrator) Continue(err error) {
 	}
 
 	if err != nil {
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
```
swetharepakula, line 190: this error will be passed into commitTransaction as well so it will be tracked there. It does not need to be tracked here.

pkg/neg/syncers/transaction.go
swetharepakula: I was thinking instead of operating on a error that is passed in, to do it where it occurs. However in this case, this is where the error is handled so I am okay with this.

swetharepakula: This shouldn't be counted as ignored. An error here causes triggers a retry.

sawsa307: Updated. Thanks!

pkg/neg/syncers/dualstack/migrator.go
sawsa307: Updated. Thanks!

pkg/neg/metrics/metrics.go
sawsa307: Updated. Thanks!

pkg/neg/controller.go
```diff
@@ -713,6 +717,7 @@ func (c *Controller) handleErr(err error, key interface{}) {
 	c.logger.Error(nil, msg)
 	if service, exists, err := c.serviceLister.GetByKey(key.(string)); err != nil {
 		c.logger.Error(err, "Failed to retrieve service from store", "service", key.(string))
+		metrics.PublishNegControllerErrorCountMetrics(err, true)
```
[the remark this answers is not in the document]
sawsa307: The two errors are different.
This is the error from getting service. In the serviceWorker, we handle the error from processService()

pkg/utils/utils.go
```diff
@@ -268,6 +268,40 @@ func GetErrorType(err error) string {
...
+	if err == nil {
+		return false
+	}
+	var gerr *googleapi.Error
+	if !errors.As(err, &gerr) {
+		return false
+	}
+	for {
+		if apiErr, ok := err.(*googleapi.Error); ok {
+			return apiErr.Code >= http.StatusInternalServerError
```
bowei, line 282: don't we want to check in the range [500, 599] not >= 500

pkg/utils/utils.go
```diff
@@ -268,6 +268,40 @@ func GetErrorType(err error) string {
...
+	if err == nil {
+		return false
+	}
+	var k8serr *k8serrors.StatusError
+	if !errors.As(err, &k8serr) {
+		return false
+	}
+	for {
+		if apiErr, ok := err.(*k8serrors.StatusError); ok {
+			return apiErr.ErrStatus.Code >= http.StatusInternalServerError
```
bowei, line 299: don't we want to check in the range [500, 599] not >= 500

PR #2098, opened 2023-05-02
Add flags for gke-ingress-checker

songrx1997:


cmd/gke-diagnoser/app/command/root.go
```diff
@@ -21,21 +21,54 @@ import (
...
+			fmt.Println(err)
+			os.Exit(1)
+		}
 		fmt.Println("Starting check-gke-ingress")
 	},
 }
 
+func init() {
+	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "kubeconfig file to use for Kubernetes config")
+	rootCmd.PersistentFlags().StringVarP(&kubecontext, "context", "c", "", "context to use for Kubernetes config")
```
swetharepakula, line 52: Do we need both of these?

cmd/gke-diagnoser/app/command/root.go
```diff
@@ -21,21 +21,54 @@ import (
...
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "", "only check resources from this namespace")
+	rootCmd.PersistentFlags().StringVarP(&outputFormat,
+		"output", "o", report.JSONOutput,
+		fmt.Sprintf("output format for check results (supports: %v)", report.SupportedOutputs()))
```
swetharepakula, line 56: indicate that if not provided the output is printed to stdout

cmd/gke-diagnoser/app/command/root.go
songrx1997: Yes, `kubeconfig` is the path to the config file to be used, and `kubecontext` is the context name.
Since we are going to support user cases that customer use this CLI alone without gcpdiag, so we should have both to support user to choose the kubeconfig file and the context.

cmd/gke-diagnoser/app/command/root.go
```diff
@@ -21,21 +21,54 @@ import (
...
+		}
 		fmt.Println("Starting check-gke-ingress")
 	},
 }
 
+func init() {
+	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "path to the kubeconfig file for Kubernetes config")
+	rootCmd.PersistentFlags().StringVarP(&kubecontext, "context", "c", "", "context to use for Kubernetes config")
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "", "only check resources from this namespace")
+	rootCmd.PersistentFlags().StringVarP(&outputFormat,
```
swetharepakula, line 54: since we only support json, we should only include this flag when we support other outputs.

songrx1997: Removed, thanks!

cmd/gke-diagnoser/app/command/root.go
```diff
@@ -23,15 +23,32 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	kubeconfig   string
+	kubecontext  string
+	namespace    string
+	outputFormat string
```
swetharepakula, line 30: nit: remove outputFormat as it is not used anymore

swetharepakula (commented):
/lgtm
/approve

PR #2099, opened 2023-05-05
Remove the flag EnableBackendConfigHealthCheck

DamianSawicki:
After this PR is merged, the next steps (with some soak time in between as appropriate) are:
 * remove the flag `--enable-backendconfig-healthcheck` from config files (internal),
 * remove the flag `--enable-backendconfig-healthcheck` from `pkg/flags/flags.go`.

~This is a child PR dependent on https://github.com/kubernetes/ingress-gce/pull/2088. Please review only the last commit `Remove the flag EnableBackendConfigHealthCheck`.~

### DIFF
--- a/docs/deploy/resources/glbc.yaml
+++ b/docs/deploy/resources/glbc.yaml
@@ -63,7 +63,6 @@ spec:
         - --gce-ratelimit=ga.BackendServices.Get,qps,1.8,1
         - --gce-ratelimit=ga.HealthChecks.Get,qps,1.8,1
         - --gce-ratelimit=alpha.HealthChecks.Get,qps,1.8,1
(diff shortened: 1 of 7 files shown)


kl52752:
\lgtm

aojea:
/hold

last time we removed a flag we broke the deployments because the manifests were still using the flag

did we update all the manifests that have this flag set?

/cc @swetharepakula @bowei

DamianSawicki:
We haven't removed the CLI flag yet. Just the `if`s in the code.

aojea:
> We haven't removed the CLI flag yet. Just the `if`s in the code.

/hold cancel

thanks for clarifying, sorry for the confusion

PR #2103, opened 2023-05-09
Rename syncer and endpoint metrics to match Monarch

sawsa307:
Rename metrics name and key to match monarch definition.

### DIFF
--- a/pkg/neg/metrics/endpoint_metrics.go
+++ b/pkg/neg/metrics/endpoint_metrics.go
@@ -20,32 +20,24 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const (
-	endpointState      = "endpoint_state"
-	endpointSliceState = "endpoint_slice_state"
-
-	endpointStateKey      = "neg_sync_endpoint_state"
-	endpointSliceStateKey = "neg_sync_endpoint_slice_state"
-)
-
 var (
 	// syncerEndpointState tracks the count of endpoints in different states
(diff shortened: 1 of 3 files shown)


pkg/neg/metrics/endpoint_metrics.go
```diff
@@ -24,8 +24,8 @@ const (
 	endpointState      = "endpoint_state"
 	endpointSliceState = "endpoint_slice_state"
```
gauravkghildiyal, line 25: I think we also have to change the metric labels?

pkg/neg/metrics/endpoint_metrics.go
```diff
@@ -24,8 +24,8 @@ const (
...
 
-	endpointStateKey      = "neg_sync_endpoint_state"
-	endpointSliceStateKey = "neg_sync_endpoint_slice_state"
+	endpointStateKey      = "syncer_endpoint_state"
```
gauravkghildiyal, line 27: Since we are anyway making some changes here, another minor thing that I wanted to bring up: Can we not declare these constants since they aren't used anywhere and only make the naming more confusing and really affect readability?

I understand that there is a precedence here but I'm expecting to change that as well in future PRs

pkg/neg/metrics/endpoint_metrics.go
sawsa307: Updated. Thanks!

swetharepakula (commented):
/lgtm
/approve

PR #2108, opened 2023-05-10
Build binaries for docker images in cloudbuild.yaml

code-elinka:
If I run an image: gcr.io/k8s-ingress-image-push/ingress-gce-e2e-test
I see an error: 

```
==============================================================================
E2E TEST

./run.sh: line 113: /e2e-test: No such file or directory
```

It means it doesn't have binaries that were supposed to be copied. 
This change adds a line that builds those binaries (make all-build).

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,15 +5,35 @@ options:
   machineType: E2_HIGHCPU_8
 steps:
 - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20230206-8160eea68e
-  entrypoint: make
+  id: build-binaries
+  entrypoint: bash
   env:
-    - DOCKER_CLI_EXPERIMENTAL=enabled
-    - REGISTRY="gcr.io/k8s-ingress-image-push"
-    - VERSION=$_PULL_BASE_REF
-    - VERBOSE=1
-    - HOME=/root
-    - USER=root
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -5,15 +5,22 @@ options:
...
     - VERSION=$_PULL_BASE_REF
+    - ALL_ARCH="amd64 arm64"
     - VERBOSE=1
     - HOME=/root
     - USER=root
-  args: ['all-push','ALL_ARCH=amd64 arm64']
+  args:
+  - -c
+  - |
+    echo "build binaries"
```
aojea, line 20: what about using two different steps, one for each command?
https://cloud.google.com/build/docs/build-config-file-schema#yaml_8

aojea:
/hold

I don't have the same problem

```
 docker run gcr.io/k8s-ingress-image-push/ingress-gce-e2e-test:master

--- BEGIN ---
Warning: could not get Compute project name from the metadata server (attempt 1)
Warning: could not get Compute project name from the metadata server (attempt 2)
Warning: could not get Compute project name from the metadata server (attempt 3)
```

code-elinka:
The actual problem is that binaries that are built are not multiarch but amd in multiarch wrapping https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-ingress-gce-push-image/1656111566392659968#1:build-log.txt%3A1360 

@aojea PTAL

PR #2110, opened 2023-05-10
Remove verbose arg from cloudbuild.yaml

code-elinka:
The last cloudbuild.yaml change was merged with VERBOSE variable and it leads to postsubmit jobs to fail. This change removes duplicate mentionings of VERBOSE and uses direct list of architectures instead of overhead with ALL_ARCH.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -10,14 +10,13 @@ steps:
   env:
   - DOCKER_CLI_EXPERIMENTAL=enabled
   - VERSION=$_PULL_BASE_REF
-  - ALL_ARCH="amd64 arm64"
   - VERBOSE=1
   - HOME=/root
   - USER=root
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -17,7 +17,7 @@ steps:
   args: 
   - -c
   - |
-    VERBOSE="${VERBOSE}" make all-build
+    make all-build
```
aojea, line 20: remove the ALL_ARCH variable too and set it explicitly

```
make all-push ALL_ARCH="amd64 arm64"
```

there is no benefit in adding more indirections

code-elinka: done

aojea:
this is a try and see what happens thing 😄 

/lgtm
/approve

PR #2111, opened 2023-05-10
Emit Events on THC (mis)configuration

DamianSawicki:
The purpose of the PR is to emit Events on Service/Ingress objects when the UHC health check for the Service has been successfully configured with the Transparent Health Checks (THC) feature or such a configuration has been attempted.

### DIFF
--- a/pkg/controller/translator/translator.go
+++ b/pkg/controller/translator/translator.go
@@ -206,50 +206,67 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
 	return nil
 }
 
-// setEnableTHC sets the THCEnabled for the service port as true or false depending on whether
+// setThcOptInOnSvc sets the THCOptInOnSvc for the service port as true or false depending on whether
 // Transparent Health Checks should be enabled.
-func (t *Translator) setEnableTHC(sp *utils.ServicePort, svc *api_v1.Service) {
-	THCEnabled := false
-	defer func() {
-		klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, THCEnabled)
-		sp.THCEnabled = THCEnabled
-	}()
-
-	if !t.enableTHC {
-		return
-	}
-
-	if sp.BackendConfig != nil && sp.BackendConfig.Spec.HealthCheck != nil {
-		return
-	}
-
-	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
+func (t *Translator) setThcOptInOnSvc(sp *utils.ServicePort, svc *api_v1.Service) (flagWarning bool) {
+	thcOptIn, err := annotations.FromService(svc).IsThcAnnotated()
 	if err != nil {
-		message := fmt.Sprintf("Parsing THC annotation failed: %+v.", err)
-		t.recorderGetter.Recorder(sp.ID.Service.Namespace).Event(svc, api_v1.EventTypeWarning, "THCAnnotationParsingFailed", message)
-		klog.Warning(message)
+		klog.Warningf("Parsing THC annotation failed: %+v.", err)
+	}
+	if !thcOptIn {
+		sp.THCConfiguration.THCOptInOnSvc = false
+		klog.Infof("THCOptInOnSvc=%v for the sevice %v with service port (%v, %v).", sp.THCConfiguration.THCOptInOnSvc, svc.Name, sp.Port, sp.PortName)
+		return
 	}
 
-	if THCEnabled && !sp.NEGEnabled {
-		message := "THC annotation present, but NEG is disabled. Will not enable Transparent Health Checks."
-		t.recorderGetter.Recorder(sp.ID.Service.Namespace).Event(svc, api_v1.EventTypeWarning, "THCAnnotationWithoutNEG", message)
-		klog.Warning(message)
-		THCEnabled = false
+	// Feature flag for Transparent Health Checks not set.
+	if !t.enableTHC {
+		sp.THCConfiguration.THCEvents.THCAnnotationWithoutFlag = true
+		flagWarning = true
+
+		sp.THCConfiguration.THCOptInOnSvc = false
+		klog.Infof("THCOptInOnSvc=%v for the sevice %v with service port (%v, %v).", sp.THCConfiguration.THCOptInOnSvc, svc.Name, sp.Port, sp.PortName)
+		return
 	}
+
+	// There is a BackendConfig detailing the health check configuration for the service.
+	if sp.BackendConfig != nil && sp.BackendConfig.Spec.HealthCheck != nil {
+		sp.THCConfiguration.THCEvents.BackendConfigOverridesTHC = true
+
+		sp.THCConfiguration.THCOptInOnSvc = false
+		klog.Infof("THCOptInOnSvc=%v for the sevice %v with service port (%v, %v).", sp.THCConfiguration.THCOptInOnSvc, svc.Name, sp.Port, sp.PortName)
+		return
+	}
+
+	// THC works only with NEGs (not Instance Groups) and this is not a Service with NEG enabled.
+	if !sp.NEGEnabled {
+		sp.THCConfiguration.THCEvents.THCAnnotationWithoutNEG = true
+
+		sp.THCConfiguration.THCOptInOnSvc = false
+		klog.Infof("THCOptInOnSvc=%v for the sevice %v with service port (%v, %v).", sp.THCConfiguration.THCOptInOnSvc, svc.Name, sp.Port, sp.PortName)
+		return
+	}
+
+	sp.THCConfiguration.THCEvents.THCConfigured = true
+
+	sp.THCConfiguration.THCOptInOnSvc = true
+	klog.Infof("THCOptInOnSvc=%v for the sevice %v with service port (%v, %v).", sp.THCConfiguration.THCOptInOnSvc, svc.Name, sp.Port, sp.PortName)
+	return
 }
 
 // getServicePort looks in the svc store for a matching service:port,
 // and returns the nodeport.
-func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePortParams, namer namer_util.BackendNamer) (*utils.ServicePort, error) {
+// The returned bool indicates whether there is a warning being returned or not.
+func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePortParams, namer namer_util.BackendNamer) (*utils.ServicePort, error, bool) {
 	svc, err := t.getCachedService(id)
 	if err != nil {
-		return nil, err
+		return nil, err, false
 	}
 
 	port := ServicePort(*svc, id.Port)
 	if port == nil {
 		// This is a fatal error.
-		return nil, errors.ErrSvcPortNotFound{ServicePortID: id}
+		return nil, errors.ErrSvcPortNotFound{ServicePortID: id}, false
 	}
 
 	// We periodically add information to the ServicePort to ensure that we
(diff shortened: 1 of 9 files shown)


pkg/controller/translator/translator.go
```diff
@@ -209,33 +209,37 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
 
-	if !t.enableTHC {
-		return
-	}
-
-	if sp.BackendConfig != nil && sp.BackendConfig.Spec.HealthCheck != nil {
-		return
-	}
-
 	THCEnabled, err := annotations.FromService(svc).ShouldEnableTHC()
```
kl52752, line 218: move this line at the beginning of the function. I'm not sure if this is buggy or just confusing.  And if you are not doing early returns looks like you don't need those defer at all.

pkg/utils/serviceport.go
```diff
@@ -54,6 +64,7 @@ type ServicePort struct {
 	L4RBSEnabled   bool
 	L7ILBEnabled   bool
 	THCEnabled     bool
+	THCEvents      THCEvents
```
kl52752, line 67: instead of THCEnabled and THCEvents can you do THCConfiguration and put it there?

pkg/controller/translator/translator.go
DamianSawicki: Yeah, without early returns `defer` is not really needed. I haven't removed it yet because if we are asked to introduce another flag, then `defer` will be useful and the flag will be checked between `defer` and reading the annotation.

pkg/utils/serviceport.go
DamianSawicki: Done.

pkg/controller/translator/translator.go
DamianSawicki: Done.

pkg/controller/translator/translator.go
```diff
@@ -206,50 +206,56 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
-	THCEnabled := false
-	defer func() {
-		klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, THCEnabled)
-		sp.THCEnabled = THCEnabled
-	}()
+func (t *Translator) setThcOptInOnSvc(sp *utils.ServicePort, svc *api_v1.Service) (flagWarning bool) {
+	thcOptIn, err := annotations.FromService(svc).IsThcAnnotated()
+	if err != nil {
+		klog.Warningf("Parsing THC annotation failed: %+v.", err)
+	}
```
DamianSawicki, line 215: return here

aojea:
I have some concerns about the supportability of this approach in the future because we have to modify several functions signatures just to cascade down and up the information, I can see two possible problems:
- problem to scale for new features and we keep adding new parameters and return values, maybe it can be solved using an object that we can add new fields without modifying the methods?
- risk of breaking the feature, this can be alleviated adding unit test that covers the feature end to end

This is basically  using events to validate the Ingress object against a flag, I spent some time with @DamianSawicki trying to look for another approaches but we couldn't come up with good alternatives, maybe a more modular approach adding a new validation method  and validate the Ingress and all the Services there before calling `TranslateIngress()` ? Since everything is from the cache I can't think this can be a performance problem, but I'm not very familiar with this controller

@DamianSawicki @swetharepakula @kl52752 from the functional pov this lgtm, I defer to you how to proceed, I'm fine with whatever you decide

DamianSawicki:
Thanks again @aojea for looking at it with me!

> problem to scale for new features and we keep adding new parameters and return values, maybe it can be solved using an object that we can add new fields without modifying the methods?

You’re right, we need to be frugal with new parameters and return values. [I’ve decreased](https://github.com/kubernetes/ingress-gce/compare/a0420f5251bc6ffb10d9d7c0c9cdbca6fb36dfd8..a2f94d50b362c17c3c0711bbc780c77ec8629583) the number of parameters of `healthChecks.sync` by using the type `utils.THCConfiguration` instead of the types of its two fields. After this change, the present PR does not increase the number of parameters of any function.

The PR does increase the number of return values of `translator.(*Translator).TranslateIngress` and a non-exported method called by `TranslateIngress`, namely `(*Translator).getServicePort`. The extra return value indicates a warning condition because I prefer not to put a warning in a slice of errors (because then we couldn’t simply test the emptiness of the slice). This solution should scale well: with more features, the warning return value can become more complex, but `TranslateIngress` should always return 1) a positive-scenario value, 2) an error value, 3) a warning value. If you prefer, I can put the warning information in the error slice or replace the error slice with a struct with separate fields for errors and warnings.

> I spent some time with [@DamianSawicki](https://github.com/DamianSawicki) trying to look for another approaches but we couldn't come up with good alternatives, maybe a more modular approach adding a new validation method and validate the Ingress and all the Services there before calling TranslateIngress() ?

I had another look at the code and it seems that `TranslateIngress` and `getServicePort` already serve for validation purposes: `getServicePort` can return `errors.ErrBadSvcType`, `errors.ErrSvcAppProtosParsing`, `errors.ErrSvcBackendConfig`, or `errors.ErrBackendConfigValidation`. So I guess it makes sense to keep validation inside `TranslateIngress`.

> risk of breaking the feature, this can be alleviated adding unit test that covers the feature end to end

The risk is not severe as the feature is only about giving feedback to the User.

Still, the PR has a unit test `translator.TestSetThcOptInOnSvc` for `translator.setThcOptInOnSvc`, which contains the core logic. There is also `healthchecks.TestNotifyAboutTHC` testing Event emission on Services. However, in our discussions with @aojea we focused on event emissions on Ingresses, and while the above-mentioned `translator.TestSetThcOptInOnSvc` tests the core logic, I do not see an easy way to inject a fake Recorder to `controller.LoadBalancerController` to really check Event emission in `controller.(*LoadBalancerController).sync`. What seems feasible is testing `TranslateIngress`, which is the step right at the top of `controller.(*LoadBalancerController).sync` in the call stack, so it would cover the forwarding `setThcOptInOnSvc -> getServicePort` and the aggregation `getServicePort -> TranslateIngress`.

pkg/controller/translator/translator_test.go
```diff
@@ -1167,7 +1185,7 @@ func TestSetTrafficScaling(t *testing.T) {
 	}
 }
 
-func TestSetEnableTHC(t *testing.T) {
+func TestSetThcOptInOnSvc(t *testing.T) {
 	// No t.Parallel()
```
swetharepakula, line 1189: nit remove comment or add why we can't run this in parallel

pkg/controller/translator/translator_test.go
```diff
@@ -1194,101 +1212,122 @@ func TestSetEnableTHC(t *testing.T) {
...
-			svc:       newService(map[string]string{thcLabel: "random text"}),
-			enableTHC: false,
-			want:      &utils.ServicePort{NEGEnabled: true, THCEnabled: false},
+			name:        "invalid annotation flag disabled",
+			sp:          &utils.ServicePort{NEGEnabled: true, THCConfiguration: utils.THCConfiguration{THCOptInOnSvc: true}},
+			svc:         newService(map[string]string{thcLabel: "random text"}),
+			enableTHC:   false,
+			want:        &utils.ServicePort{NEGEnabled: true, THCConfiguration: utils.THCConfiguration{THCOptInOnSvc: false}},
+			wantEvent:   true,
+			eventPrefix: "Warning THCAnnotationParsingFailed Parsing THC annotation failed",
```
swetharepakula, line 1264: I don't see eventPrefix being used anymore. Should this be removed after the refactor?

pkg/healthchecks/healthchecks.go
```diff
@@ -148,17 +149,52 @@ func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (
...
+		klog.Warningf("BackendConfig exists and thcOptIn true simultaneously for %v. Ignoring transparent health check.", hc.Name)
+		thcConf.THCOptInOnSvc = false
 	}
+	createOrUpdate := false
+
+	defer func() {
+		if createOrUpdate {
+			h.notifyAboutTHC(hc, thcConf.THCEvents)
+		}
+	}()
```
swetharepakula, line 198: instead of the defer, since we know when we call create and when we call update (2 places). It might be better to just call notify then having a defer function.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -535,6 +535,67 @@ func getSingletonHealthcheck(t *testing.T, c *gce.Cloud) *compute.HealthCheck {
...
+	}
+
+	fakeRecorder := fakeSingletonRecorderGetter.FakeRecorder()
+	for _, tc := range testCases {
+		healthChecks.notifyAboutTHC(hc, tc.events)
+		for _, wantText := range tc.wantTexts {
+			select {
+			case output := <-fakeRecorder.Events:
+				if output != wantText {
+					t.Fatalf("Incorrect event emitted on healthcheck update: %s.", output)
```
swetharepakula, line 585: t.Errorf instead so the test keeps running.

swetharepakula (commented):
This PR LGTM. Mostly nits and suggestions

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

pkg/healthchecks/healthchecks.go
DamianSawicki: Done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

pkg/controller/translator/translator_test.go
DamianSawicki: Well spotted, done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

pkg/controller/translator/translator_test.go
DamianSawicki: Done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

pkg/controller/translator/translator.go
```diff
@@ -206,50 +206,67 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
+
+	sp.THCConfiguration.THCOptInOnSvc = true
+	klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, sp.THCConfiguration.THCOptInOnSvc)
+	return
 }
 
 // getServicePort looks in the svc store for a matching service:port,
 // and returns the nodeport.
-func (t *Translator) getServicePort(id utils.ServicePortID, params *getServicePortParams, namer namer_util.BackendNamer) (*utils.ServicePort, error) {
+// The returned bool is for warnings (there is currently only one type of warnings possible, so a bool type suffices).
```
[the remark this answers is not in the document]
DamianSawicki: Done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

pkg/controller/translator/translator.go
```diff
@@ -206,50 +206,67 @@ func (t *Translator) maybeEnableBackendConfig(sp *utils.ServicePort, svc *api_v1
...
-		sp.THCEnabled = THCEnabled
-	}()
+func (t *Translator) setThcOptInOnSvc(sp *utils.ServicePort, svc *api_v1.Service) (flagWarning bool) {
+	thcOptIn, err := annotations.FromService(svc).IsThcAnnotated()
+	if err != nil {
+		klog.Warningf("Parsing THC annotation failed: %+v.", err)
+	}
+	if !thcOptIn {
+		sp.THCConfiguration.THCOptInOnSvc = false
+		klog.Infof("Is THC enabled for the sevice %v with service port (%v, %v)? %v", svc.Name, sp.Port, sp.PortName, sp.THCConfiguration.THCOptInOnSvc)
```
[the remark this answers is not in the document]
DamianSawicki: Done (https://github.com/kubernetes/ingress-gce/compare/a2f94d50b362c17c3c0711bbc780c77ec8629583..2e27a1b6c5082e5a68bb1b253fbc16fbe3699b86).

PR #2112, opened 2023-05-10
Add CheckHealthCheckConfig rule to check-gke-ingress

songrx1997:
/assign @swetharepakula

### DIFF
--- a/cmd/check-gke-ingress/app/ingress/rule.go
+++ b/cmd/check-gke-ingress/app/ingress/rule.go
@@ -57,16 +57,6 @@ func CheckBackendConfigAnnotation(svc *corev1.Service) (*annotations.BackendConf
 	return beConfigs, report.Passed, fmt.Sprintf("BackendConfig annotation is valid in service %s/%s", svc.Namespace, svc.Name)
 }
 
-func getBackendConfigAnnotation(svc *corev1.Service) (string, bool) {
-	for _, bcKey := range []string{annotations.BackendConfigKey, annotations.BetaBackendConfigKey} {
-		val, ok := svc.Annotations[bcKey]
-		if ok {
-			return val, ok
-		}
-	}
-	return "", false
-}
-
 func CheckBackendConfigExistence(ns string, beConfigName string, svcName string, client beconfigclient.Interface) (*beconfigv1.BackendConfig, string, string) {
 	beConfig, err := client.CloudV1().BackendConfigs(ns).Get(context.TODO(), beConfigName, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/check-gke-ingress/app/ingress/rule.go
```diff
@@ -77,3 +67,23 @@ func CheckBackendConfigExistence(ns string, beConfigName string, svcName string,
 	}
 	return beConfig, report.Passed, fmt.Sprintf("BackendConfig %s/%s in service %s/%s found", ns, beConfigName, ns, svcName)
 }
+
+func CheckHealthCheckConfig(beConfig *beconfigv1.BackendConfig, svcName string) (string, string) {
```
swetharepakula, line 71: should we rename this check to specific to the timeout or interval. This will make it clearer when we read the report as well.

cmd/check-gke-ingress/app/ingress/rule.go
```diff
@@ -77,3 +67,23 @@ func CheckBackendConfigExistence(ns string, beConfigName string, svcName string,
...
+	if beConfig.Spec.HealthCheck == nil {
+		return report.Skipped, fmt.Sprintf("BackendConfig %s/%s in service %s/%s  does not have healthcheck specified", beConfig.Namespace, beConfig.Name, beConfig.Namespace, svcName)
+	}
+	if *beConfig.Spec.HealthCheck.TimeoutSec > *beConfig.Spec.HealthCheck.CheckIntervalSec {
```
swetharepakula, line 75: We should check if they are nil first, since these could be invalid configs.

cmd/check-gke-ingress/app/ingress/rule_test.go
```diff
@@ -186,3 +188,49 @@ func TestCheckBackendConfigExistence(t *testing.T) {
...
+
+func TestCheckHealthCheckConfig(t *testing.T) {
+
+	for _, tc := range []struct {
+		desc             string
+		timeoutSec       int64
+		checkIntervalSec int64
+		expect           string
+	}{
+		{
```
swetharepakula, line 200: We should test what happens when hc config is nil, and when the fields are nil as well.

cmd/check-gke-ingress/app/ingress/rule.go
songrx1997: Updated, thanks!

cmd/check-gke-ingress/app/ingress/rule.go
songrx1997: updated, thanks!

cmd/check-gke-ingress/app/ingress/rule_test.go
songrx1997: Added, thanks!

cmd/check-gke-ingress/app/ingress/rule_test.go
```diff
@@ -186,3 +188,80 @@ func TestCheckBackendConfigExistence(t *testing.T) {
...
+			},
+			expect: report.Failed,
+		},
+		{
+			desc:   "No healthCheck specified",
+			spec:   beconfigv1.BackendConfigSpec{},
+			expect: report.Skipped,
+		},
+		{
+			desc: "TimeoutSec not specidfied",
```
swetharepakula, line 237: nit: typo specified

swetharepakula (commented):
one last nit, otherwise this LGTM

swetharepakula (commented):
/lgtm
/approve

PR #2113, opened 2023-05-11
Change e2e-test dockerfile run cmds

code-elinka:
Install gcloud cli in e2e-test image
based on architecture to support multiarch.

### DIFF
--- a/Dockerfile.e2e-test
+++ b/Dockerfile.e2e-test
@@ -16,22 +16,28 @@
 # and replace with a new one if needed
 FROM debian:11
 
+#ARG_ARCH & ARG_BIN are replaces with sed in build/rules.mk
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
-ARG CLOUD_SDK_VERSION="416.0.0-0"
+
+ARG CLOUD_SDK_VERSION="416.0.0"
 ENV CLOUD_SDK_VERSION=$CLOUD_SDK_VERSION
+ENV PATH /google-cloud-sdk/bin:$PATH
 
 COPY cmd/e2e-test/run.sh /run.sh
+
+# Get architecture for gcloud cli installation
+RUN if [ "ARG_ARCH" = "amd64" ]; \
+  then echo -n "x86_64" > /tmp/arch; \
+  else echo -n "arm" > /tmp/arch; fi;
 
 # Download Google Cloud GPG and install google-cloud-cli
 # See recommended instructions:
 # cloud.google.com/sdk/docs/install#installation_instructions
-RUN apt-get update && \
+RUN ARCH=`cat /tmp/arch` && \
+  apt-get update && \
   apt-get install -y curl python apt-transport-https ca-certificates gnupg && \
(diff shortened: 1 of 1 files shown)


code-elinka:
Dockerfile can be built on arm machine: 
```
$ docker build -f Dockerfile .
[+] Building 0.1s (7/7) FINISHED                                                                                        
 => [internal] load build definition from Dockerfile                                                               0.1s
 => => transferring dockerfile: 1.69kB                                                                             0.0s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 2B                                                                                    0.0s
 => [internal] load metadata for docker.io/library/debian:11                                                       0.0s
 => [1/3] FROM docker.io/library/debian:11                                                                         0.0s
 => CACHED [2/3] RUN if [ `uname -m` = 'x86_64' ];   then echo -n "x86_64" > /tmp/arch;   else echo -n "arm" > /t  0.0s
 => CACHED [3/3] RUN ARCH=`cat /tmp/arch` &&   apt-get update &&   apt-get install -y curl python apt-transport-h  0.0s
 => exporting to image                                                                                             0.0s
 => => exporting layers                                                                                            0.0s
 => => writing image sha256:d2e047ebbd1e1ff8a6c5afb27cf3db8a8835d1a4ae10ad1ba141763ae06198c3                       0.0s

$ docker run -it --entrypoint bash d2e047ebbd1e
root@207c3630d888:/# 
```

Dockerfile.e2e-test
```diff
@@ -17,21 +17,26 @@
...
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
-ARG CLOUD_SDK_VERSION="416.0.0-0"
+
+ARG CLOUD_SDK_VERSION="416.0.0"
 ENV CLOUD_SDK_VERSION=$CLOUD_SDK_VERSION
+ENV PATH /google-cloud-sdk/bin:$PATH
 
 COPY cmd/e2e-test/run.sh /run.sh
 
+# Get architecture for gcloud cli installation
```
cezarygerard, line 27: debian 11 has `arch` command just for that

```
$ docker run -t -d debian:11
d6ec301cd8af5a89d5f826820e79e6ac5ead81a4e2baf41dce27a48453257325
$ docker ps
CONTAINER ID   IMAGE                  COMMAND                  CREATED         STATUS         PORTS                       NAMES
d6ec301cd8af   debian:11              "bash"                   3 seconds ago   Up 2 seconds                               naughty_brown

$ docker exec -it d6ec301cd8af arch
x86_64

```

aojea: this is copied from here https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/emulators/Dockerfile , I think that because the arch command doesn't return just `arm`

is a tomato tomatoe thing, this is consistent with the official dockerfile seems less risky

code-elinka: done

aojea:
@code-elinka can you check locally it works and paste the results? (not in the arm machine ;)

PR #2114, opened 2023-05-11
 Use gcloud base image for e2e-test container

code-elinka:
The google-cloud-cli:debian_component_based image has all needed binaries and certificates preinstalled, in addition to that it works with arm64 that makes it possible to run e2e test on different architectures.

### DIFF
--- a/Dockerfile.e2e-test
+++ b/Dockerfile.e2e-test
@@ -14,30 +14,12 @@
 
 # Check End of Life Date wiki.debian.org/DebianReleases
 # and replace with a new one if needed
-FROM debian:11
+# debian_component_based google-cloud-cli image supports arm64/amd64 architectures
+# and has curl, gcloud, and ca-certificated preinstalled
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/run.sh
```diff
@@ -16,7 +16,20 @@
 
 # run.sh manages the settings required for running containerized in a
 # Kubernetes cluster.
-echo '--- BEGIN ---'
+
+echo "Downloading test dependencies.."
+CLOUDSDK_VERSION=${CLOUD_SDK_VERSION:-"416.0.0-0"}
+apt-get update
+apt-get install -y curl python apt-transport-https ca-certificates gnupg
```
aojea, line 23: this is making a lot of assumptions, as that the script will run in a debian-like system

it also will install the packages any time

/hold

code-elinka: After offline discussion we decided to drop this change and proceed with a new base layer image that has curl, ca-certs, and gcloud cli and supports both arm and amd architectures

PR #2121, opened 2023-05-17
Use hack script for e2e test image building

code-elinka:
The e2e image needs to be built with docker buildx because it may be built in a cross-compiled way
and it has not only pre-copied binaries but also
some dependencies (sh script, curl, gcloud-cli).

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,25 +5,11 @@ options:
   machineType: E2_HIGHCPU_8
 steps:
 - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20230206-8160eea68e
-  id: build-binaries
-  entrypoint: bash
-  env:
-  - DOCKER_CLI_EXPERIMENTAL=enabled
-  - VERSION=$_PULL_BASE_REF
-  - VERBOSE=1
-  - HOME=/root
-  - USER=root
-  args: 
-  - -c
-  - |
-    make all-build ALL_ARCH="amd64 arm64"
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -5,21 +5,26 @@ options:
...
   - VERSION=$_PULL_BASE_REF
   - VERBOSE=1
   - HOME=/root
   - USER=root
-  args: 
+  args:
   - -c
   - |
-    make all-build ALL_ARCH="amd64 arm64"
+    # Build and push every image except e2e-test
```
aojea, line 20: explain a bit more the context, expand on why we need to do this

cloudbuild.yaml
```diff
@@ -5,21 +5,26 @@ options:
...
   - USER=root
-  args: 
+  args:
   - -c
   - |
-    make all-build ALL_ARCH="amd64 arm64"
+    # Build and push every image except e2e-test 
+    BINARIES="glbc 404-server 404-server-with-metrics echo fuzzer"
+    make all-push ALL_ARCH="amd64 arm64" CONTAINER_BINARIES=${BINARIES}
+substitutions:
```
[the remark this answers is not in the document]
code-elinka: lol, thank you!

cloudbuild.yaml
code-elinka: Sure. Added the comment into the e2e-test image building step. Tell me if I need to change it somehow

code-elinka:
> /hold Cloudbuild treats `"e2e-test"` not like `Dockerfile.e2e-test` but `Dockerfile."e2e-test"`... ooh

/unhold

PR #2123, opened 2023-05-18
Use only one buildx command for e2e-test

code-elinka:
1) Docker buildx cmd can build a multiarch image
in one run. Use it instead of many commands
2) The hack/push-multiarch.sh script is used for e2e build
via docker buildx (/root/.docker/cli-plugins/docker-buildx).
It requires to specify HOME=/root so buildx works as expected.
See:
[github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing
](https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing#simple-build-example)

### DIFF
--- a/hack/push-multiarch.sh
+++ b/hack/push-multiarch.sh
@@ -17,6 +17,10 @@ set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
+
+# docker buildx is in /root/.docker/cli-plugins/docker-buildx
+HOME=/root
+
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd ${REPO_ROOT}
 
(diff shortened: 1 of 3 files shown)


hack/push-multiarch.sh
```diff
@@ -17,6 +17,10 @@ set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
+
+# docker buildx is in /root/.docker/cli-plugins/docker-buildx
```
aojea, line 21: if you see the logs failures https://storage.googleapis.com/kubernetes-jenkins/logs/post-ingress-gce-push-image/1658960735641800704/build-log.txt the error seem to come from `docker buildx build --platform=linux/$arch`

the command has a --push flag
```
    docker buildx build --platform=linux/$arch \
            -f ".dockerfile-${arch}.${binary}" \
            -t "${IMAGE_NAME}:${VERSION}" --push
```

this script does not look right to me, with buildx you don't need to iterate and create the manifest it is all done in the same command, this should be something like

```
docker buildx build --push \
--platform linux/amd64,linux/arm64 \
--tag "${IMAGE_NAME}:${VERSION}" .
```

and remove the for loop entirely

aojea: it seems that other cloudbuilds https://github.com/kubernetes/kubernetes/blob/master/build/pause/cloudbuild.yaml use a previous step ` gcloud auth configure-docker \` to authenticate it first

code-elinka: /hold 

> this script does not look right to me, with buildx you don't need to iterate and create the manifest it is all done in the same command, this should be something like
> and remove the for loop entirely

I will need to get architecture into dockerfile so it copies right binaries. Or copy both lol and run a correct one later

code-elinka: Because the second command requires us to have only one dockerfile (for now we create a new dockerfile for every binary and every arch based on original ones. That's why we use `sed`

code-elinka: /unhold Builds correctly

PR #2124, opened 2023-05-18
Introduce dynamic throttling for NEG requests

alexkats:
The purpose of this whole PR is to introduce the dynamic throttling mechanism for NEG requests (List/Attach/DetachNetworkEndpoints) that should be more flexible than static QPS rate limiter

### DIFF
--- a/pkg/neg/syncers/backoff.go
+++ b/pkg/neg/syncers/backoff.go
@@ -26,17 +26,19 @@ import (
 
 var ErrRetriesExceeded = fmt.Errorf("maximum retry exceeded")
 
-// backoffHandler handles delays for back off retry
-type backoffHandler interface {
-	// NextRetryDelay returns the delay for next retry or error if maximum number of retries exceeded.
-	NextRetryDelay() (time.Duration, error)
-	// ResetRetryDelay resets the retry delay
-	ResetRetryDelay()
+// BackoffHandler handles delays for back off retry
+type BackoffHandler interface {
+	// NextDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextDelay() (time.Duration, error)
+	// ResetDelay resets the retry delay
+	ResetDelay()
+	// DecreaseDelay returns the decreased delay for next retry
+	DecreaseDelay() time.Duration
 }
 
-// exponentialBackOffHandler is a backoff handler that returns retry delays semi-exponentially with random jitter within boundary.
-// exponentialBackOffHandler returns ErrRetriesExceeded when maximum number of retries has reached.
-type exponentialBackOffHandler struct {
+// exponentialBackoffHandler is a backoff handler that returns retry delays semi-exponentially with random jitter within boundary.
+// exponentialBackoffHandler returns ErrRetriesExceeded when maximum number of retries has reached.
+type exponentialBackoffHandler struct {
 	lock           sync.Mutex
 	lastRetryDelay time.Duration
 	retryCount     int
@@ -45,8 +47,8 @@ type exponentialBackOffHandler struct {
 	maxRetryDelay  time.Duration
 }
 
-func NewExponentialBackendOffHandler(maxRetries int, minRetryDelay, maxRetryDelay time.Duration) *exponentialBackOffHandler {
-	return &exponentialBackOffHandler{
+func NewExponentialBackoffHandler(maxRetries int, minRetryDelay, maxRetryDelay time.Duration) BackoffHandler {
+	return &exponentialBackoffHandler{
 		lastRetryDelay: time.Duration(0),
 		retryCount:     0,
 		maxRetries:     maxRetries,
@@ -55,8 +57,8 @@ func NewExponentialBackendOffHandler(maxRetries int, minRetryDelay, maxRetryDela
 	}
 }
 
-// NextRetryDelay returns the next back off delay for retry.
-func (handler *exponentialBackOffHandler) NextRetryDelay() (time.Duration, error) {
+// NextDelay returns the next back off delay for retry.
+func (handler *exponentialBackoffHandler) NextDelay() (time.Duration, error) {
 	handler.lock.Lock()
 	defer handler.lock.Unlock()
 	handler.retryCount += 1
@@ -72,10 +74,26 @@ func (handler *exponentialBackOffHandler) NextRetryDelay() (time.Duration, error
 	return handler.lastRetryDelay, nil
 }
 
-// ResetRetryDelay resets the retry delay.
-func (handler *exponentialBackOffHandler) ResetRetryDelay() {
+// ResetDelay resets the retry delay.
+func (handler *exponentialBackoffHandler) ResetDelay() {
 	handler.lock.Lock()
 	defer handler.lock.Unlock()
 	handler.retryCount = 0
 	handler.lastRetryDelay = time.Duration(0)
+}
+
+// DecreaseDelay returns the decreased delay for next retry.
+func (handler *exponentialBackoffHandler) DecreaseDelay() time.Duration {
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+	handler.retryCount = 0
+	if handler.lastRetryDelay <= handler.minRetryDelay {
+		handler.lastRetryDelay = time.Duration(0)
+	} else {
+		handler.lastRetryDelay = handler.lastRetryDelay*2 - wait.Jitter(handler.lastRetryDelay, 0.5)
+		if handler.lastRetryDelay < handler.minRetryDelay {
+			handler.lastRetryDelay = handler.minRetryDelay
+		}
+	}
+	return handler.lastRetryDelay
 }
--- a/pkg/neg/types/sync_errors_test.go
+++ b/pkg/neg/types/sync_errors_test.go
@@ -51,6 +51,16 @@ func TestClassifyError(t *testing.T) {
 			expectSyncErr: ErrEPNodeMissing,
 		},
 		{
+			desc:          "an error that wraps a NegSyncError and another error",
+			err:           fmt.Errorf("%w: %w", ErrEPNodeMissing, errors.New("additional context")),
+			expectSyncErr: ErrEPNodeMissing,
+		},
+		{
+			desc:          "an error that wraps another error and a NegSyncError",
+			err:           fmt.Errorf("%w: %w", errors.New("additional context"), ErrEPNodeMissing),
+			expectSyncErr: ErrEPNodeMissing,
+		},
+		{
 			desc: "an error that wraps a non NegSyncError",
 			err:  nonNegSyncErrorWrapped,
 			expectSyncErr: NegSyncError{
--- a/pkg/ratelimit/metrics/metrics.go
+++ b/pkg/ratelimit/metrics/metrics.go
@@ -21,38 +21,101 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/ingress-gce/pkg/utils"
 )
 
 var register sync.Once
 
-const (
-	rateLimitSubsystem  = "rate_limit"
-	rateLimitLatencyKey = "rate_limit_delay_seconds"
-)
+const negControllerSubsystem = "neg_controller"
 
 var (
-	rateLimitLatencyMetricsLabels = []string{
-		"key", // rate limiter key
+	metricsLabels = []string{
+		"api_key", // rate limiter key (e.g. ga.Addresses.Get)
 	}
 
 	RateLimitLatency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Subsystem: rateLimitSubsystem,
-			Name:      rateLimitLatencyKey,
+			Subsystem: negControllerSubsystem,
+			Name:      "rate_limit_delay_seconds",
 			Help:      "Latency of the RateLimiter Accept Operation",
 			// custom buckets = [0.5s, 1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048s(~34min), 4096s(~68min), +Inf]
 			Buckets: prometheus.ExponentialBuckets(0.5, 2, 14),
 		},
-		rateLimitLatencyMetricsLabels,
+		metricsLabels,
+	)
+
+	StrategyLatency = prometheus.NewHistogramVec(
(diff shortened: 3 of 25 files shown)


pkg/neg/syncers/backoff.go
```diff
@@ -26,17 +26,19 @@ import (
...
-	NextRetryDelay() (time.Duration, error)
-	// ResetRetryDelay resets the retry delay
-	ResetRetryDelay()
+// BackoffHandler handles delays for back off retry
+type BackoffHandler interface {
+	// NextDelay returns the delay for next retry or error if maximum number of retries exceeded.
+	NextDelay() (time.Duration, error)
+	// ResetDelay resets the retry delay
+	ResetDelay()
+	// DecreaseDelay returns the decreased delay for next retry
```
bowei, line 35: This is kind of confusing -- does this have a side effect and also returns a value?

When will this be called etc vs ResetDelay? I think the other methods are more intuitive and their documentation is sufficient, but DecreaseDelay is somewhat out of ordinary understanding.

aojea: I remember to make the same comment, I  can't think on having a backoff handler that decrease, I can see how can be useful to have Reset()  method to start over, but allowing to go up and down is indeed confusing

pkg/ratelimit/ratelimit_test.go
```diff
@@ -71,3 +148,119 @@ func TestRateLimitScale(t *testing.T) {
...
+	}()
+	blocked = true
+	acceptAndVerify()
+	blocked = false
+	// Needed if the call to Accept() wasn't blocked, otherwise test would
+	// finish without the check for blocked request in the goroutine
+	wg.Wait()
+}
+
+func TestStrategyRateLimiterGroupBlock(t *testing.T) {
```
swetharepakula, line 208: see if we can implement this without using a time.Sleep

pkg/ratelimit/metrics/metrics.go
```diff
@@ -43,16 +50,100 @@ var (
 			// custom buckets = [0.5s, 1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s(~4min), 512s(~8min), 1024s(~17min), 2048s(~34min), 4096s(~68min), +Inf]
 			Buckets: prometheus.ExponentialBuckets(0.5, 2, 14),
 		},
-		rateLimitLatencyMetricsLabels,
+		rateLimitMetricsLabels,
+	)
+
+	StrategyLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: rateLimitSubsystem,
```
swetharepakula, line 58: use the neg controller subsystem here as the rate limiter is for the neg controller.
The end metric name will be `neg_controller_rate_limit_delay_seconds` with the proposed change.


Same comment for the other metrics

pkg/ratelimit/metrics/metrics.go
```diff
@@ -21,17 +21,24 @@ import (
...
+	strategyUsedDelayKey    = "strategy_used_delay_seconds"
+	strategyLockLatencyKey  = "strategy_lock_delay_seconds"
+	quotaExceededCounterKey = "quota_exceeded_counter"
+	otherErrorsCounterKey   = "other_errors_counter"
+	noErrorsCounterKey      = "no_errors_counter"
 )
 
 var (
-	rateLimitLatencyMetricsLabels = []string{
+	rateLimitMetricsLabels = []string{
```
swetharepakula, line 41: what are the possible values for this?

pkg/ratelimit/metrics/metrics.go
```diff
@@ -21,17 +21,24 @@ import (
...
-	rateLimitSubsystem  = "rate_limit"
-	rateLimitLatencyKey = "rate_limit_delay_seconds"
+	rateLimitSubsystem      = "rate_limit"
+	rateLimitLatencyKey     = "rate_limit_delay_seconds"
+	strategyLatencyKey      = "strategy_delay_seconds"
+	strategyUsedDelayKey    = "strategy_used_delay_seconds"
+	strategyLockLatencyKey  = "strategy_lock_delay_seconds"
+	quotaExceededCounterKey = "quota_exceeded_counter"
+	otherErrorsCounterKey   = "other_errors_counter"
+	noErrorsCounterKey      = "no_errors_counter"
```
swetharepakula, line 37: unless the keys will be reused, we can directly add them into the metric definition. This makes it easier when people come to this file to get metric names etc

pkg/ratelimit/metrics/metrics.go
```diff
@@ -43,16 +50,100 @@ var (
...
+		},
+		rateLimitMetricsLabels,
+	)
+
+	StrategyLockLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: rateLimitSubsystem,
+			Name:      strategyLockLatencyKey,
+			Help:      "Latency of the lock acquisition for the StrategyRateLimiter",
+			Buckets:   append(prometheus.ExponentialBuckets(1e-9, 10, 8), prometheus.ExponentialBuckets(0.1, 12, 14)...),
```
swetharepakula, line 83: Add a comment with what the buckets are

swetharepakula (commented):
Generally, the logic looks good. Main comment is to comment both the code and the tests

pkg/neg/types/sync_errors_test.go
```diff
@@ -50,6 +50,11 @@ func TestClassifyError(t *testing.T) {
 			err:           fmt.Errorf("%w: %v", ErrEPNodeMissing, errors.New("additional context")),
 			expectSyncErr: ErrEPNodeMissing,
 		},
+		{
+			desc:          "an error that wraps a NegSyncError and another error",
+			err:           fmt.Errorf("%w: %w", ErrEPNodeMissing, errors.New("additional context")),
```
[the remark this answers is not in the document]
alexkats: tested it using go1.20, and everything worked both by using `go test` and `make test`

pkg/throttling/strategy.go
```diff
@@ -0,0 +1,255 @@
...
+	strategy.clockCtx, strategy.clockCancelCtx = context.WithCancel(context.Background())
+	decreaseDelayTimeout := strategy.noRequestsTimeoutBeforeDecreasingDelay
+	resetDelayTimeout := strategy.noRequestsTimeoutBeforeResettingDelayAfterDecreasingDelay
+	if strategy.currentDelay*3 > (decreaseDelayTimeout + resetDelayTimeout) {
+		resetDelayTimeout = strategy.currentDelay*3 - decreaseDelayTimeout
+	}
+	go func() {
+		if strategy.wait(decreaseDelayTimeout, strategy.decreaseDelay) {
+			strategy.wait(resetDelayTimeout, strategy.resetDelay)
+		}
```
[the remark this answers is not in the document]
alexkats: Added a comment for this in the code as well. `resetDelayTimeout` holds the time to wait before resetting the delay
after the `decreaseDelayTimeout` has passed. This condition is to prevent resetting the delay if it will be less than the current delay multiplied by 3

pkg/throttling/strategy.go
```diff
@@ -0,0 +1,255 @@
...
+	strategy.currentDelay = strategy.backoff.DecreaseDelay()
+}
+
+// Delay returns the delay for next request, taking into account the last request timestamp.
+func (strategy *defaultStrategy) Delay() time.Duration {
+	strategy.lock.Lock()
+	defer strategy.lock.Unlock()
+	elapsed := strategy.clock.Now().Sub(strategy.lastRequestTimestamp)
+	delay := strategy.currentDelay - elapsed
+	if delay < 0 || !strategy.delayed {
```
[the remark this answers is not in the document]
alexkats: It will make code a bit more messy, cause we still need to update the last request timestamp, even if the request is not delayed

pkg/throttling/strategy_test.go
```diff
@@ -0,0 +1,323 @@
...
+	strategy, err := NewTwoWayStrategy(testMinDelay, testMaxDelay, testNoRequestsTimeoutBeforeDecreasingDelay, testNoRequestsTimeoutBeforeResettingDelay, fakeClock)
+	verifyErrorIsNil(t, err)
+	verifyExactDelay(t, strategy, 0, "")
+	quotaErr := &googleapi.Error{Code: 429}
+
+	for i := 0; i < 100; i++ {
+		strategy.Observe(quotaErr)
+	}
+
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
alexkats: Added the comment in the code, basically it's needed trigger goroutine switch and process the queue of events in the fake clock

pkg/throttling/strategy_test.go
```diff
@@ -0,0 +1,323 @@
...
+	fakeClock.Step(testNoRequestsTimeoutBeforeDecreasingDelay)
+	time.Sleep(time.Second)
+	fakeClock.Step(testNoRequestsTimeoutBeforeResettingDelay - testNoRequestsTimeoutBeforeDecreasingDelay)
+	time.Sleep(time.Second)
+	// it's needed to check the delay without elapsed time from the last request
```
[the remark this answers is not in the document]
alexkats: I'd say it's the part of the same case

pkg/throttling/strategy_test.go
```diff
@@ -0,0 +1,323 @@
...
+	fakeClock := clocktesting.NewFakeClock(time.Now())
+	strategy, err := NewDynamicStrategy(testMinDelay, testMaxDelay, testErrorsBeforeIncreasingDelay, testSuccessesBeforeDecreasingDelay, testSuccessesBeforeResettingDelay, testNoRequestsTimeoutBeforeDecreasingDelay, testNoRequestsTimeoutBeforeResettingDelay, fakeClock)
+	verifyErrorIsNil(t, err)
+	quotaErr := &googleapi.Error{Code: 429}
+
+	for _, tc := range []struct {
+		desc string
+		errs []error
+		// -2 = reset delay, -1 = decrease delay, 0 = no change, 1 = increase delay
+		expectedDeltaDelays []int
```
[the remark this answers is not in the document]
alexkats: Agreed, introduced an enum here

pkg/throttling/strategy_test.go
```diff
@@ -0,0 +1,323 @@
...
+			errs:                []error{quotaErr, quotaErr, quotaErr, quotaErr, nil, nil, nil, nil, quotaErr, quotaErr},
+			expectedDeltaDelays: []int{1, 1, 0, 1, 0, -1, 0, 0, 0, 1},
+		},
+		{
+			desc:                "1 error, 5 successes and 1 error",
+			errs:                []error{quotaErr, nil, nil, nil, nil, nil, quotaErr},
+			expectedDeltaDelays: []int{1, 0, -1, 0, 0, 0, 1},
+		},
+	} {
+		strategy.(*dynamicStrategy).resetDelay()
```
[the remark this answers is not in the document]
alexkats: Great, thanks for pointing out, didn't know about this thing

pkg/ratelimit/ratelimit.go
```diff
@@ -49,41 +80,62 @@ func init() {
...
+				return nil, err
+			}
+			rateLimitImpls[key] = impl
+			klog.Infof("Configured rate limiting for: %v, scale = %f", key, flags.F.GCERateLimitScale)
 		}
-		rateLimitImpls[key] = impl
-		klog.Infof("Configured rate limiting for: %v, scale = %f", key, flags.F.GCERateLimitScale)
 	}
-	if len(rateLimitImpls) == 0 {
+	if len(rateLimitImpls) == 0 && len(strategyRLs) == 0 {
```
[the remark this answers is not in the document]
alexkats: Decided not to change the current behavior for now, but it should be considered in the future

pkg/ratelimit/ratelimit_test.go
```diff
@@ -71,3 +148,119 @@ func TestRateLimitScale(t *testing.T) {
...
+	}
+	rlk := &cloud.RateLimitKey{
+		ProjectID: "",
+		Version:   "ga",
+		Service:   "test",
+		Operation: "test",
+	}
+	acceptAndVerify := func() {
+		go func() {
+			time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
alexkats: As in a previous similar comment, we need it for processing events of the fake clock

pkg/ratelimit/ratelimit_test.go
alexkats: As in a previous similar comment, we need it for processing events of the fake clock

pkg/ratelimit/metrics/metrics.go
alexkats: It's a `RateLimitKey`, smth like `ga.Addresses.Get`, added the example in the comment, and renamed the `key` to `api_key`

pkg/neg/syncers/backoff.go
alexkats: The intended usage for it is inside throttling strategy (which is introduced in later commits). The dynamic strategy tries to adjust the delay based on the previous request results, and that's the place where all three methods are used, including `DecreaseDelay`

pkg/neg/types/sync_errors_test.go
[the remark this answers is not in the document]
swetharepakula: confirmed that this works starting 1.20. 1.19 this is not supported

pkg/ratelimit/metrics/metrics.go
```diff
@@ -21,38 +21,121 @@ import (
...
+		metricsLabels,
+	)
+
+	NoErrorsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: negControllerSubsystem,
+			Name:      "no_errors_counter",
+			Help:      "Number of successful requests in the rate limiter",
+		},
+		metricsLabels,
```
swetharepakula, line 104: Could we combine this into a single metric and then using the labels you can differentiate no errors, other errors, and quota exceeded errors?

swetharepakula (commented):
one comment about the metrics, otherwise LGTM

swetharepakula (commented):
one comment about the metrics, otherwise LGTM

pkg/ratelimit/metrics/metrics.go
alexkats: Makes sense, thanks, converted to one metric

swetharepakula (commented):
Thanks Alex!

/lgtm
/approve

PR #2126, opened 2023-05-18
Change logic of deleting RBS service

panslava:
- Delete only if service has finalizer or rbs forwarding rule, meaning it was provisioned by RBS controller at some point
- Don't delete RBS annotation on garbage collection

/assign @cezarygerard @aojea

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -134,14 +134,15 @@ func (lc *L4NetLBController) needsAddition(newSvc, oldSvc *v1.Service) bool {
 
 // needsDeletion return true if svc required deleting RBS based NetLB
 func (lc *L4NetLBController) needsDeletion(svc *v1.Service) bool {
-	// Check if service has RBS related fields/resources
-	// this check differs from isRBSBasedService() func by checking also non load balancer type services
-	if !(lc.hasRBSAnnotation(svc) || utils.HasL4NetLBFinalizerV2(svc) || lc.hasRBSForwardingRule(svc)) {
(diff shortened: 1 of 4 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -621,14 +622,14 @@ func (lc *L4NetLBController) garbageCollectRBSNetLB(key string, svc *v1.Service)
 
 	// IMPORTANT: Remove LB annotations from the Service LAST, cause changing service fields are emitted as service update to other controllers
 	if lc.enableDualStack {
-		if err := deleteL4RBSDualStackAnnotations(lc.ctx, svc); err != nil {
```
aojea, line 624: This is the same error we found yesterday, the last thing to be removed should be the finalizer, also why do we need to remove the annotations if it is going to be deleted?

aojea: should not we just delete the finalizer and bail out?

panslava: Annotation deletion covers the case of garbage collection on switching service types (to clusterIP or ILB), when service not really completely deleted

panslava: But we should not remove RBS annotation (as it was set up by user) so this PR removes deleting RBS annotation

aojea: >  switching service types (

ok, but then we should have 2 flows , one for deletion and other for mutation.
The deletion just removes the finalizer
The mutation just removes the annotations added by the controller but keeps the finalizer

IIUIC this still conflates both

PR #2127, opened 2023-05-19
Add CheckL7ILBFrontendConfig to check-gke-ingress

songrx1997:
/assign @swetharepakula

### DIFF
--- a/cmd/check-gke-ingress/app/ingress/rule.go
+++ b/cmd/check-gke-ingress/app/ingress/rule.go
@@ -144,6 +144,18 @@ func CheckAppProtocolAnnotation(svc *corev1.Service) (string, string) {
 	return report.Passed, fmt.Sprintf("AppProtocol annotation is valid in service %s/%s", svc.Namespace, svc.Name)
 }
 
+// CheckL7ILBFrontendConfig checks whether an internal ingress has a
+// frontendConfig. It will fail if an internal ingress has a frontendConfig.
+func CheckL7ILBFrontendConfig(ing *networkingv1.Ingress) (string, string) {
+	if !isL7ILB(ing) {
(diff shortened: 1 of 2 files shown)


cmd/check-gke-ingress/app/ingress/rule.go
```diff
@@ -113,6 +113,18 @@ func CheckFrontendConfigExistence(namespace, name string, client feconfigclient.
 	return feConfig, report.Passed, fmt.Sprintf("FrontendConfig %s/%s found", namespace, name)
 }
 
+// CheckL7ILBFrontendConfig checks whether an internal ingress has a
```
swetharepakula, line 116: Add that it will fail if an internal ingress has a frontend config.

swetharepakula (commented):
one nit, otherwise LGTM

cmd/check-gke-ingress/app/ingress/rule.go
songrx1997: Updated

cmd/check-gke-ingress/app/ingress/rule.go
```diff
@@ -123,3 +135,25 @@ func getBackendConfigAnnotation(svc *corev1.Service) (string, bool) {
 	}
 	return "", false
 }
+
+// isL7ILB wether an ingress is for internal load balancing.
```
swetharepakula, line 139: nit: whether

swetharepakula (commented):
one more nit. Also looks like there is a conflict, this might need to be rebased.

cmd/check-gke-ingress/app/ingress/rule.go
songrx1997: updated.

swetharepakula (commented):
/lgtm
/approve

PR #2131, opened 2023-05-19
Reset L4 DualStack Gauge metrics before exporting

panslava:
Now, if service does not exist anymore, it will not be zeroed in metrics


pkg/metrics/metrics.go
```diff
@@ -626,12 +626,30 @@ func (im *ControllerMetrics) computeL4ILBMetrics() map[feature]int {
...
+					IPFamilies:     ipFamilies,
+					IPFamilyPolicy: ipFamilyPolicy,
+					Status:         status,
+				}] = 0
+			}
+		}
+	}
+	return zeroCounts
+}
+
```
cezarygerard, line 646: move it to the bottom of the file, or to another file "l4_dualstack_metrics.gom" (together with all dualstack stuff)

pkg/metrics/l4metrics_test.go
```diff
@@ -467,8 +467,12 @@ func TestComputeL4NetLBDualStackMetrics(t *testing.T) {
 			for i, serviceState := range tc.serviceStates {
 				newMetrics.SetL4NetLBDualStackService(fmt.Sprint(i), serviceState)
 			}
+			expectedCount := zeroL4DualStackCounts()
```
cezarygerard, line 470: maybe add short comment how you construct the expected count

cezarygerard, line 470: this funny  code will work either:

```
			t.Parallel()
			newMetrics := FakeControllerMetrics()
			for i, serviceState := range tc.serviceStates {
				newMetrics.SetL4ILBDualStackService(fmt.Sprint(i), serviceState)
			}
			for key, _ := range zeroL4DualStackCounts() {
				tc.expectL4ILBDualStackCount[key] = tc.expectL4ILBDualStackCount[key]
			}
			got := newMetrics.computeL4ILBDualStackMetrics()
			if diff := cmp.Diff(tc.expectL4ILBDualStackCount, got); diff != "" {
				t.Fatalf("Got diff for L4 ILB Dual-Stack service counts (-want +got):\n%s", diff)
			}
		})
```

which is better?

panslava: yeh, good point, it saves one line, but I prefer not to change tc, though I understand, this code looks a bit strange maybe
i will add comment

panslava: done

pkg/metrics/metrics.go
panslava: moved to another file

panslava:
/hold

I will squash commits after lgtm

cezarygerard:
just use  Reset()  method on gave metrics :-D

panslava:
@cezarygerard I changed, now using Reset

PR #2132, opened 2023-05-19
Remove running tests with verbose mode

panslava:
We have too many logs in all tests

/assign @aojea @swetharepakula @gauravkghildiyal

### DIFF
--- a/build/test.sh
+++ b/build/test.sh
@@ -23,7 +23,7 @@ export CGO_ENABLED=0
 TARGETS=$(for d in "$@"; do echo ./$d/...; done)
 
 echo "Running tests:"
-go test -v -installsuffix "static" ${TARGETS}
+go test -installsuffix "static" ${TARGETS}
 echo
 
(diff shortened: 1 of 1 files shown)


aojea:
that is on purpose, otherwise how do you debug the failures?

panslava:
@aojea how it works, at least for me, without -v it will print logs for the tests that failed, but not for all of the tests

image

aojea:
your tests show (cached), you need to use `-count 1` to not run the cached test

aojea:
indeed `-v` logs everything
```
        -v
            Verbose output: log all tests as they are run. Also print all
            text from Log and Logf calls even if the test succeeds.
```
so I think is a matter of taste then https://grep.app/search?q=go%20test%20-v&filter[lang][0]=Go, 

I like to run verbose on CI but I can't see any other reason to not do it, up to you then

aojea:
/lgtm

it is ok

PR #2136, opened 2023-05-22
Reset gauges when collecting service metrics

cezarygerard:
otherwise deleted services will appear in metrics forever it their label combination was unique


code-elinka:
@cezarygerard 
just a nitpick to replace `otherwise deleted services will appear in metrics forever it their label combination was unique` with `otherwise deleted services will appear in metrics forever IF their label combination was unique`. 

Looks good to me.

mmamczur:
weren't we concerned before that this can make the metrics show 0 if collected at a bad moment?

PR #2138, opened 2023-05-22
Endpoint State metrics Cleanup

sawsa307:
1. Fix metrics emit issue. Metrics should be emitted when error is nil.
2. Remove unneeded parameter in computeEndpointStateMetrics.
3. Add endpoint and endpoint slice count to the emit metrics log.


swetharepakula (commented):
/lgtm
/approve

PR #2139, opened 2023-05-23
Use multiarch echo image for e2e tests

code-elinka:
To run some of the e2e tests we need an echo
image, therefore we want both of them be
multiarch.
NB! The Windows image doesn't support multiarch.


gauravkghildiyal:
LGTM

PR #2140, opened 2023-05-23
Pin golang patch version in Makefile

code-elinka:
Patches of golang could differ and
make the debugging process harder.
To demolish that problem we should
pin not only major.minor versions
but also patch one.

@cezarygerard I decided to proceed with patch, not sha because it's more readable and can be used in any basic image.  
image
see: https://hub.docker.com/_/golang

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ARCH ?= amd64
 ALL_ARCH := amd64
 
 # Image to use for building.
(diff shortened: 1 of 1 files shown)


gauravkghildiyal:
Thanks @code-elinka!
LGTM

swetharepakula (commented):
/lgtm
/approve

PR #2144, opened 2023-05-24
Push arch specific images

code-elinka:
We have 3 registries for images:
* multiarch
* amd64
* arm64
 Even we can keep everything in the first one,
we still want to support last two registries
for users who may need that.

### DIFF
--- a/hack/push-multiarch.sh
+++ b/hack/push-multiarch.sh
@@ -42,6 +42,9 @@ echo "setting up docker buildx.."
 docker buildx install
 docker buildx create --use
 
+# Download crane cli
+curl -sL "https://github.com/google/go-containerregistry/releases/download/v0.15.2/go-containerregistry_$(uname -s)_$(uname -m).tar.gz" | tar xvzf - krane
+
 for binary in ${BINARIES}
 do
     # "arm64 amd64" ---> "linux/arm64,linux/amd64" 
@@ -54,4 +57,12 @@ do
         --tag  ${MULTIARCH_IMAGE} \
         -f Dockerfile.${binary} .
     echo "done, pushed $MULTIARCH_IMAGE image"
+
+    # Tag arch specific images for the legacy registries
+    for arch in ${ALL_ARCH}
+    do
+        # krane is a variation of crane that supports k8s auth
+        ./krane copy --platform linux/${arch} ${MULTIARCH_IMAGE} ${REGISTRY}/ingress-gce-${binary}-${arch}:${VERSION}
+    done
+    echo "images are copied to arch specific registries"
 done


hack/push-multiarch.sh
```diff
@@ -54,4 +54,15 @@ do
...
     echo "done, pushed $MULTIARCH_IMAGE image"
+
+    # Download crane cli
+    curl -sL "https://github.com/google/go-containerregistry/releases/download/v0.15.2/go-containerregistry_$(uname -s)_$(uname -m).tar.gz"
+    tar -zxvf go-containerregistry.tar.gz -C /usr/local/bin/ crane
+    
+    # Tag arch specific images for the legacy registries
+    for arch in ${ALL_ARCH}
+    do
+        crane copy --platform linux/${arch} ${MULTIARCH_IMAGE} ${REGISTRY}/ingress-gce-${binary}-${arch}:${VERSION}
```
aojea, line 65: use `krane` https://github.com/google/go-containerregistry/blob/main/cmd/krane/README.md

hack/push-multiarch.sh
```diff
@@ -54,4 +54,15 @@ do
         --tag  ${MULTIARCH_IMAGE} \
         -f Dockerfile.${binary} .
     echo "done, pushed $MULTIARCH_IMAGE image"
+
+    # Download crane cli
+    curl -sL "https://github.com/google/go-containerregistry/releases/download/v0.15.2/go-containerregistry_$(uname -s)_$(uname -m).tar.gz"
+    tar -zxvf go-containerregistry.tar.gz -C /usr/local/bin/ crane
```
aojea, line 60:
```suggestion
curl -sL "https://github.com/google/go-containerregistry/releases/download/v0.15.2/go-containerregistry_$(uname -s)_$(uname -m).tar.gz" | tar xvzf - krane
```

hack/push-multiarch.sh
aojea:
```suggestion
        ./krane copy --platform linux/${arch} ${MULTIARCH_IMAGE} ${REGISTRY}/ingress-gce-${binary}-${arch}:${VERSION}
```

code-elinka: done, thank you

hack/push-multiarch.sh
code-elinka: done, thank you

hack/push-multiarch.sh
```diff
@@ -54,4 +54,15 @@ do
         --tag  ${MULTIARCH_IMAGE} \
         -f Dockerfile.${binary} .
     echo "done, pushed $MULTIARCH_IMAGE image"
+
+    # Download crane cli
+    curl -sL "https://github.com/google/go-containerregistry/releases/download/v0.15.2/go-containerregistry_$(uname -s)_$(uname -m).tar.gz" | tar xvzf - krane
```
aojea, line 59: if you have multiple binaries this will download krane multiple times, you can put it at the beginning of the script

PR #2145, opened 2023-05-24
Delete RBS annotation explicitly in preventTargetPoolRaceWithRBSOnCreation

panslava:
As we removed deleting RBS annotation in garbageCollectRBSNetLB we now need to remove annotation explicitly, otherwise service will not be handled by legacy controller

/assign @cezarygerard

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -316,7 +316,7 @@ func (lc *L4NetLBController) preventTargetPoolRaceWithRBSOnCreation(service *v1.
 			"Failed to clean RBS resources for load balancer with target pool, err: %v", result.Error)
 		return result.Error
 	}
-	return nil
+	return lc.deleteRBSAnnotation(service)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -316,7 +316,7 @@ func (lc *L4NetLBController) preventTargetPoolRaceWithRBSOnCreation(service *v1.
 			"Failed to clean RBS resources for load balancer with target pool, err: %v", result.Error)
 		return result.Error
 	}
-	return nil
+	return lc.deleteRBSAnnotation(service)
```
aojea, line 319: this seems to mutate the object, is it considering that the object belong to the cache or the passed service is a copy?

panslava: Good question. Can I ask you, what problem appears when we mutate cache object?

aojea: https://github.com/kubernetes/kubernetes/issues/116694

aojea: https://github.com/kubernetes/kubernetes/issues/89157

PR #2152, opened 2023-05-26
Deep copy service after getting from informer

### LINKED ISSUE #2135: L4 controller references cache objects
It seems that the controller build new objects that reference the existing object of the cache, I don't understand well enough to know if this is on. purpose, but it is risky since the cache is shared and I find very brittle if this new object works on a reference to the cache, maybe is safer to deepcopy the Service

https://github.com/kubernetes/ingress-gce/blob/0a822cc11ab9de1273f02fdeb1ad3768cd3cf8bd/pkg/l4lb/l4controller.go#L230-L238

https://github.com/kubernetes/ingress-gce/blob/0a822cc11ab9de1273f02fdeb1ad3768cd3cf8bd/pkg/l4lb/l4netlbcontroller.go#L581-L588

/kind bug
/assign @cezarygerard @panslava

panslava:
Fixes #2135 

/assign @aojea @cezarygerard

### DIFF
--- a/pkg/common/typed/service.go
+++ b/pkg/common/typed/service.go
@@ -43,7 +43,8 @@ func (s *ServiceStore) Get(c *api_v1.Service) (*api_v1.Service, bool, error) {
 	if item == nil {
 		return nil, exists, err
 	}
-	return item.(*api_v1.Service), exists, err
+	svc := item.(*api_v1.Service)
+	return svc.DeepCopy(), exists, err
 }
 
 // GetByKey implements Store.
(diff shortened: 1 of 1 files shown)


aojea:
I don't know if this is the right fix

cezarygerard:
lgtm

aojea:
oh, when we were talking about deepcopy I thought that we were talking about the controller not this PR

I don't think this is a good solution, is very hidden in the code and you are going to propagate this change to all the users of this wrapper without them knowing

PR #2157, opened 2023-05-31
Reserve Static IPv6 address before syncing L4 ILB

panslava:
This will prevent losing address while recreating forwarding rule

Tested running locally with real cluster

### DIFF
--- a/pkg/loadbalancers/address_manager.go
+++ b/pkg/loadbalancers/address_manager.go
@@ -18,12 +18,12 @@ package loadbalancers
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 
+	compute "google.golang.org/api/compute/v1"
 	"k8s.io/cloud-provider-gcp/providers/gce"
 	"k8s.io/ingress-gce/pkg/utils"
-
-	compute "google.golang.org/api/compute/v1"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"k8s.io/klog/v2"
@@ -38,6 +38,21 @@ const (
 	IPAddrUnmanaged
 )
 
+// IPVersion represents compute.Address IpVersion field
+type IPVersion = string
+
+const (
+	IPv4Version IPVersion = "IPV4"
+	IPv6Version IPVersion = "IPV6"
+	// IPv6LBAddressPrefixLength used for reserving IPv6 addresses.
+	// Google Cloud reserves not a single IPv6 address, but a /96 range.
+	// At the moment, no other ranges are supported
+	IPv6LBAddressPrefixLength = 96
+	// IPv6EndpointTypeNetLB is a value on Address.Ipv6EndpointType used specify
+	// that IPv6 address will be used for NetLB. Required for new IPv6 NetLB address creation.
+	IPv6EndpointTypeNetLB = "NETLB"
+)
+
 // Original file in https://github.com/kubernetes/legacy-cloud-providers/blob/6aa80146c33550e908aed072618bd7f9998837f6/gce/gce_address_manager.go
 type addressManager struct {
 	logPrefix   string
@@ -50,9 +65,17 @@ type addressManager struct {
 	subnetURL   string
 	tryRelease  bool
 	networkTier cloud.NetworkTier
+	ipVersion   IPVersion
 }
 
-func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetURL, name, targetIP string, addressType cloud.LbScheme, networkTier cloud.NetworkTier) *addressManager {
+func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetURL, name, targetIP string, addressType cloud.LbScheme, networkTier cloud.NetworkTier, ipVersion IPVersion) *addressManager {
+	if targetIP != "" {
+		// Store address in normalized format.
+		// This is required for IPv6 addresses, to be able to filter by exact address,
+		// because filtering is happening by regexp and google cloud stores address in the shortest form.
+		targetIP = net.ParseIP(targetIP).String()
+	}
+	// We don't have verification that targetIP matches IPVersion, we will rely on Google cloud API verification
 	return &addressManager{
 		svc:         svc,
 		logPrefix:   fmt.Sprintf("AddressManager(%q)", name),
@@ -64,6 +87,7 @@ func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetU
 		tryRelease:  true,
 		subnetURL:   subnetURL,
 		networkTier: networkTier,
+		ipVersion:   ipVersion,
 	}
 }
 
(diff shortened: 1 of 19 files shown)


panslava:
CI will fail on vendor folder verification for now

I am waiting for vendor (GoogleCloudPlatform) to have a tag that includes addresses mocks with IPv6 https://github.com/GoogleCloudPlatform/k8s-cloud-provider/pull/123

pkg/loadbalancers/address_manager.go
```diff
@@ -64,6 +73,7 @@ func newAddressManager(svc gce.CloudAddressService, serviceName, region, subnetU
 		tryRelease:  true,
 		subnetURL:   subnetURL,
 		networkTier: networkTier,
+		ipVersion:   ipVersion,
```
aojea, line 76: since the addressManager now is family aware should you validate the targetIP passed?

pkg/loadbalancers/address_manager.go
```diff
@@ -38,6 +37,15 @@ const (
 	IPAddrUnmanaged
 )
 
+// IPVersion represents compute.Address IpVersion field
+type IPVersion = string
+
+const (
+	IPv4Version               IPVersion = "IPV4"
+	IPv6Version               IPVersion = "IPV6"
+	IPv6LBAddressPrefixLength           = 96
```
aojea, line 46: explain this magic number

pkg/loadbalancers/forwarding_rules_ipv6.go
```diff
@@ -226,3 +223,17 @@ func EqualIPv6ForwardingRules(fr1, fr2 *composite.ForwardingRule) (bool, error)
...
+		return ""
+	}
+	if requestedSubnet != ipv6FwdRule.Subnetwork {
+		// reset ip address since subnet is being changed.
+		return ""
+	}
+
+	// Google Cloud creates ipv6 forwarding rules with IPAddress in CIDR form,
+	// but to create static address you need to specify address without range
+	return strings.Split(ipv6FwdRule.IPAddress, "/")[0]
```
aojea, line 238: is this safe?

pkg/loadbalancers/address_manager.go
panslava: what do you think about relying on validations on just google cloud api side (they validate that address is "proper")? Or you think it is better to replicate some validation here too?

pkg/loadbalancers/forwarding_rules_ipv6.go
panslava: I think so.
in the worst weird case ipv6FwdRule.IPAddress will be empty string (it is not a pointer) -- which is fine, we will reserve new address

pkg/loadbalancers/address_manager.go
panslava: Added comment

pkg/loadbalancers/address_manager.go
aojea: then just add a comment explaining that assumption, I think that is legit

panslava: added a comment

cezarygerard:
Is it even supported on GCP public APIs?

panslava:
@cezarygerard yes, it was added recently, and now all of these code api calls are supported (I verified)

cezarygerard:
are you going to implement the same for L4NET LB?

panslava:
@cezarygerard yes https://github.com/kubernetes/ingress-gce/pull/2165

pkg/loadbalancers/l4.go
```diff
@@ -437,14 +468,14 @@ func (l4 *L4) provideIPv4HealthChecks(nodeNames []string, result *L4ILBSyncResul
 	return hcResult.HCLink
 }
 
-func (l4 *L4) ensureDualStackResources(result *L4ILBSyncResult, nodeNames []string, options gce.ILBOptions, bs *composite.BackendService, existingFR *composite.ForwardingRule, subnetworkURL, ipToUse string) {
+func (l4 *L4) ensureDualStackResources(result *L4ILBSyncResult, nodeNames []string, options gce.ILBOptions, bs *composite.BackendService, existingIPv4FwdRule, existingIPv6FwdRule *composite.ForwardingRule, subnetworkURL, ipv4ToUse, ipv6ToUse string) {
```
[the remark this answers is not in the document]
panslava: I will be honest, this is more a heritage of previous versions of the code, they were passing those forwarding rules down to forwarding rule ensuring

So I did the same for ipv6, to be consistent

But it also saves api calls :)

pkg/loadbalancers/l4ipv6.go
```diff
@@ -181,3 +182,15 @@ func (l4 *L4) deleteIPv6NodesFirewall() error {
 
 	return l4.deleteFirewall(ipv6FirewallName)
 }
+
+func (l4 *L4) getOldIPv6ForwardingRule(existingBS *composite.BackendService) (*composite.ForwardingRule, error) {
```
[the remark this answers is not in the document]
panslava: There is a test TestDualStackInternalLoadBalancerModifyProtocol which verifies the whole flow where this functions is useful, that we reserve IP address when protocol is changed

Though, if you think it would be useful, I can add a test specifically for this function, but I would say this is more a private helper function, that just abstracts functionality for readability

pkg/loadbalancers/l4.go
```diff
@@ -309,56 +309,87 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+			ipv6ToUse, _, err = ipv6AddrMgr.HoldAddress()
+			if err != nil {
+				result.Error = fmt.Errorf("EnsureInternalLoadBalancer error: ipv6AddrMgr.HoldAddress() returned error %w", err)
+				return result
+			}
+			klog.V(2).Infof("EnsureInternalLoadBalancer(%v): reserved IPv6 address %q", nm, ipv6ToUse)
+			defer func() {
+				// Release the address that was reserved, in all cases. If the forwarding rule was successfully created,
+				// the ephemeral IP is not needed anymore. If it was not created, the address should be released to prevent leaks.
+				if err := ipv6AddrMgr.ReleaseAddress(); err != nil {
```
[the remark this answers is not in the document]
panslava: ReleaseAddress() deletes only controller managed address. Controller created address will have the same name as forwarding rule name, so controller will try to delete address with this name. If it was custom static address owned by customer -- controller will not delete it (nit: as long as customer didn't use forwarding rule name for their address)

pkg/loadbalancers/l4ipv6.go
```diff
@@ -181,3 +182,15 @@ func (l4 *L4) deleteIPv6NodesFirewall() error {
 
 	return l4.deleteFirewall(ipv6FirewallName)
 }
+
```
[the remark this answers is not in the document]
panslava: done

pkg/loadbalancers/l4.go
```diff
@@ -309,56 +309,87 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+					klog.Errorf("EnsureInternalLoadBalancer: failed to release IPv6 address reservation, possibly causing an orphan: %v", err)
```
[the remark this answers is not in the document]
panslava: Not sure..... I am not 100% sure if this does not return false positive errors, which we don't want to return to customers and scare them

pkg/loadbalancers/l4.go
```diff
@@ -309,56 +309,87 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
 				return result
 			}
-			klog.V(2).Infof("EnsureInternalLoadBalancer(%v): reserved IP %q", nm, ipv4ToUse)
+			klog.V(2).Infof("EnsureInternalLoadBalancer(%v): reserved IPv4 address %q", nm, ipv4ToUse)
 			defer func() {
 				// Release the address that was reserved, in all cases. If the forwarding rule was successfully created,
 				// the ephemeral IP is not needed anymore. If it was not created, the address should be released to prevent leaks.
 				if err := addrMgr.ReleaseAddress(); err != nil {
-					klog.Errorf("EnsureInternalLoadBalancer: failed to release address reservation, possibly causing an orphan: %v", err)
+					klog.Errorf("EnsureInternalLoadBalancer: failed to release IPv4 address reservation, possibly causing an orphan: %v", err)
```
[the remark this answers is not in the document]
panslava: Not sure..... I am not 100% sure if this does not return false positive errors, which we don't want to return to customers and scare them

pkg/loadbalancers/l4ipv6.go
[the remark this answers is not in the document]
cezarygerard: leave it as is

PR #2161, opened 2023-06-06
Update logging in endpoint calculation.

sawsa307:
Update logging in endpoint calculation.
image
image

The way to trigger this is by constructing a custom endpoint slice so that we contains two endpoints from EPS that share the same pod but contains different endpointIP. The reason we are seeing this is because for this endpoint, the IP we use for network endpoint(and used for this check) is different from the IP from the endpoint in EPS. 
image
The first IP we have is the endpoint IP from EPS, and the later one is the one we used for network endpoint.

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -349,7 +349,7 @@ func mergeWithGlobalCounts(localEPCount, globalEPCount, globalEPSCount negtypes.
 	}
 }
 
-// getEndpointZone use an endpoint's node information to get its corresponding zone
+// getEndpointZone use an endpoint's nodeName to get its corresponding zone
 func getEndpointZone(endpointAddress negtypes.AddressData, zoneGetter negtypes.ZoneGetter) (string, negtypes.StateCountMap, error) {
 	count := make(negtypes.StateCountMap)
 	if endpointAddress.NodeName == nil || len(*endpointAddress.NodeName) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -461,7 +461,7 @@ func toZoneNetworkEndpointMapDegradedMode(eds []negtypes.EndpointsData, zoneGett
 			// endpoint address should match to the IP of its pod
 			checkIPErr := podContainsEndpointAddress(networkEndpoint, pod)
 			if checkIPErr != nil {
-				klog.Errorf("Endpoint %q in Endpoints %s/%s has IP(s) not match to its pod %s: %v, skipping", endpointAddress.Addresses, ed.Meta.Namespace, ed.Meta.Name, pod.Name, checkIPErr)
+				klog.Errorf("Endpoint %q in Endpoints %s/%s has IP(s) not match to its pod %s: %v, skipping", networkEndpoint.IP, ed.Meta.Namespace, ed.Meta.Name, pod.Name, checkIPErr)
```
swetharepakula, line 464: nit: Endpoint %q in Endpoints %s/%s has at least one IP that not match to its pod %s: %v, skipping

sawsa307: Updated. Thanks!

pkg/neg/syncers/utils.go
```diff
@@ -591,7 +591,7 @@ func podContainsEndpointAddress(networkEndpoint negtypes.NetworkEndpoint, pod *a
 		}
 	}
 	if matching != len(endpointIPs) {
-		return fmt.Errorf("%w: endpoint %v has IP(s) not match to its pod's IP(s)", negtypes.ErrEPIPNotFromPod, endpointIPs)
+		return fmt.Errorf("%w: endpoint has IP(s) %v not match to its pod's IP(s)", negtypes.ErrEPIPNotFromPod, endpointIPs)
```
swetharepakula, line 594: reword: endpoint has at least one IP %v that  does not match its pod's IP(s)

swetharepakula (commented):
/lgtm
/approve

PR #2163, opened 2023-06-07
Fix firewalls for L4 ILBs.

mmamczur:
There was an omission in the code for multinet and some firewall rules were created in the incorrect VPC.

### DIFF
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -266,6 +266,7 @@ func (l4netlb *L4NetLB) ensureIPv4NodesFirewall(nodeNames []string, ipAddress st
 	}
 
 	// Add firewall rule for L4 External LoadBalancer traffic to nodes
+	defaultNetwork := network.DefaultNetwork(l4netlb.cloud)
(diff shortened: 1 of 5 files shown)


pkg/loadbalancers/l4netlb.go
```diff
@@ -266,6 +266,7 @@ func (l4netlb *L4NetLB) ensureIPv4NodesFirewall(nodeNames []string, ipAddress st
 	}
 
 	// Add firewall rule for L4 External LoadBalancer traffic to nodes
+	defaultNetwork := network.DefaultNetwork(l4netlb.cloud)
```
panslava, line 269: I think it should be not default, but also as done in L4, we should store network in l4netlb struct, and pass it here

Otherwise, I guess it will fail, if nodes are not in default network

(same for l4netlbipv6)

panslava (changes requested):

mmamczur:
I'll store this in l4netlb later when I do multinet for netlb.

Default here means cluster network.
This change will mimic the behavior we had before

panslava:
oh okay, thought default is "default" network in google cloud

thanks for the fix

/lgtm

but let's track, if that will be safe

PR #2166, opened 2023-06-08
Add support for Network Tier annotation for IPv6 NetLB

panslava:


pkg/loadbalancers/forwarding_rules_ipv6.go
```diff
@@ -175,14 +187,25 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 				klog.Errorf("ensureIPv6ForwardingRule: failed to release address reservation, possibly causing an orphan: %v", err)
 			}
 		}()
+	} else if existingIPv6FwdRule != nil && existingIPv6FwdRule.NetworkTier != netTier.ToGCEValue() {
+		err := l4netlb.forwardingRules.Delete(existingIPv6FwdRule.Name)
```
mmamczur, line 191: here the existingIPv6FwdRule is deleted if the net tier doesn't match. I assume the `existingIPv6FwdRule` variable stays non nil. Then below at line 203 there is a check for net tier. won't it trigger if you delete here because of the mismatch?

mmamczur: oh didn't notice it was there before, yea, makes sense to deal with it separately then

mmamczur:
looks good

cezarygerard:
ok

PR #2173, opened 2023-06-20
update k8s/gce server request metrics to count server error

sawsa307:
- Update k8s/gce server request metrics to count server error
- For NEG controller error count metrics, count server error with ignore=True as server error instead of ignored error. Ignore=True means we are okay with these errors, but we do care if the failure is due to server error.


swetharepakula (commented):
/lgtm
/approve

PR #2174, opened 2023-06-20
Add support for Custom Subnet for IPv6 NetLB

panslava:
- Add new annotation networking.gke.io/load-balancer-subnet which adds support to specify custom subnet for external IPv6 forwarding rule
- Use new annotation instead of old "networking.gke.io/internal-load-balancer-subnet" for ILB, but fallback to old annotation

/assign cezarygerard

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -115,7 +115,7 @@ func (l4 *L4) getILBOptions() gce.ILBOptions {
 	}
 
 	return gce.ILBOptions{AllowGlobalAccess: gce.GetLoadBalancerAnnotationAllowGlobalAccess(l4.Service),
-		SubnetName: gce.GetLoadBalancerAnnotationSubnet(l4.Service)}
+		SubnetName: annotations.FromService(l4.Service).GetInternalLoadBalancerAnnotationSubnet()}
 }
 
 // EnsureInternalLoadBalancerDeleted performs a cleanup of all GCE resources for the given loadbalancer service.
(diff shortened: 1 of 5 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -115,7 +115,7 @@ func (l4 *L4) getILBOptions() gce.ILBOptions {
 	}
 
 	return gce.ILBOptions{AllowGlobalAccess: gce.GetLoadBalancerAnnotationAllowGlobalAccess(l4.Service),
```
cezarygerard, line 117: please extend the
func TestEnsureInternalLoadBalancerCustomSubnet(t *testing.T) {
in pkg/loadbalancers/l4_test.go

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1009,6 +1078,20 @@ func verifyNetLBForwardingRule(l4netlb *L4NetLB, frName string, backendServiceLi
 	if err == nil || addr != nil {
 		return fmt.Errorf("expected error when looking up ephemeral address, got %v", addr)
 	}
+
+	// IPv4 Forwarding Rule should not have subnet
+	expectedSubnet := ""
+	// IPv6 Forwarding Rule should have subnet
+	if fwdRule.IpVersion == IPVersionIPv6 {
+		expectedSubnet, err = l4netlb.IPv6ForwardingRuleSubnet()
```
cezarygerard, line 1086: technially you unit test l4netlb.IPv6ForwardingRuleSubnet() by asserting on result from l4netlb.IPv6ForwardingRuleSubnet()

1. can you ad a unit test in pkg/loadbalancers/forwarding_rules_ipv6_test.go for IPv6ForwardingRuleSubnet() 
or
2. pass subnet name as string to new "verify", func and just use this new one in TestEnsureIPv6ExternalLoadBalancerCustomSubnet

I think #2 is better, the TestEnsureIPv6ExternalLoadBalancerCustomSubnet would just test 1 thing, and fail for 1 reason

cezarygerard:
lgtm overall

please see my comments

pkg/loadbalancers/l4netlb_test.go
panslava: thanks, went with path 2, take a look :)

pkg/loadbalancers/l4.go
panslava: extended

PR #2177, opened 2023-06-26
Neg count metrics

swetharepakula:
Add a new managed neg count metric that will track the count of negs by zone and endpoint type. This is different from the neg count in the usage metrics which only counts syncers. The usage metrics can be an undercount as it does not account for regional clusters where a neg is created per zone. Syncers' locations can point to regions for regional clusters which does not give an accurate picture of the distribution of negs across zones.

A followup commit is also included that cleans up the syncerState metric so that the different states no longer need to be hardcoded.

/assign @gauravkghildiyal 
/cc @sawsa307

### DIFF
--- a/pkg/neg/metrics/metricscollector/metrics.go
+++ b/pkg/neg/metrics/metricscollector/metrics.go
@@ -155,21 +155,6 @@ type syncerState struct {
 	inErrorState   bool
 }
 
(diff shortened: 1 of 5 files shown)


sawsa307:
LGTM, thank you for making this change!

gauravkghildiyal:
Thanks!

/lgtm

PR #2179, opened 2023-06-28
Add support to enable single ingress check

songrx1997:
/assign @swetharepakula

### DIFF
--- a/cmd/check-gke-ingress/app/command/root.go
+++ b/cmd/check-gke-ingress/app/command/root.go
@@ -48,7 +48,14 @@ var rootCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "Error connecting to Kubernetes: %v", err)
 			os.Exit(1)
 		}
-		output := ingress.CheckAllIngresses(namespace, client, beconfigClient, feConfigClient)
+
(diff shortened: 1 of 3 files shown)


cmd/check-gke-ingress/app/command/root.go
```diff
@@ -48,7 +48,14 @@ var rootCmd = &cobra.Command{
...
 		}
-		output := ingress.CheckAllIngresses(namespace, client, beconfigClient, feConfigClient)
+
+		var output report.Report
+		if len(args) == 0 {
+			output = ingress.Check("", namespace, client, beconfigClient, feConfigClient)
+		} else {
+			output = ingress.Check(args[0], namespace, client, beconfigClient, feConfigClient)
+		}
+
```
swetharepakula, line 58: This check is repeated again in ingress.Check when you verify whether the ingress name is empty or not. Split into ingress.Check and ingress.CheckAllIngresses.

Then rename the the existing CheckAllIngress to something like RunChecks

songrx1997: Done, thanks!

swetharepakula (commented):
/lgtm
/approve

PR #2181, opened 2023-06-29
Update healthcheck Description on BackendConfig removal

DamianSawicki:
In https://github.com/kubernetes/ingress-gce/pull/2008 and https://github.com/kubernetes/ingress-gce/pull/2068, we guarantee that the health check Description is updated when a BackendConfig is added to the Service (behind the flag --enable-update-hc-description). When a BackendConfig is removed, typically a health check update is triggered, which includes resetting the Description. However, under some circumstances the health check parameters update won't happen on BackendConfig removal, but it is still desirable to reset the Description to reflect the lack of BackendConfig. This reset is added in the present PR.

The changes are behind the abovementioned flag --enable-update-hc-description. This is a temporary fix and corresponding code will be removed with the subsequent planned changes.

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -17,6 +17,7 @@ limitations under the License.
 package healthchecks
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
@@ -152,9 +153,9 @@ func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (
 	return h.sync(hc, bchcc, sp.THCConfiguration)
 }
 
-// notifyAboutTHC emits Events about successful or attempted THC configuration.
+// emitTHCEvents emits Events about successful or attempted THC configuration.
 // Currently called on creation or update of a health check.
-func (h *HealthChecks) notifyAboutTHC(hc *translator.HealthCheck, thcEvents utils.THCEvents) {
+func (h *HealthChecks) emitTHCEvents(hc *translator.HealthCheck, thcEvents utils.THCEvents) {
 	if thcEvents.THCConfigured {
 		message := "Transparent Health Check successfully configured."
 		h.recorderGetter.Recorder(hc.Service.Namespace).Event(
@@ -184,9 +185,9 @@ func (h *HealthChecks) notifyAboutTHC(hc *translator.HealthCheck, thcEvents util
 
 // sync retrieves a health check based on port, checks type and settings and updates/creates if necessary.
 // sync is only called by the backends.Add func - it's not a pool like other resources.
-// We assume that bchcc cannot be non-nil and thcOptIn be true simultaneously.
-func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.HealthCheckConfig, thcConf utils.THCConfiguration) (string, error) {
-	if bchcc != nil && thcConf.THCOptInOnSvc {
+// We assume that backendConfigHCConfig cannot be non-nil and thcOptIn be true simultaneously.
+func (h *HealthChecks) sync(hc *translator.HealthCheck, backendConfigHCConfig *backendconfigv1.HealthCheckConfig, thcConf utils.THCConfiguration) (string, error) {
+	if backendConfigHCConfig != nil && thcConf.THCOptInOnSvc {
 		klog.Warningf("BackendConfig exists and thcOptIn true simultaneously for %v. Ignoring transparent health check.", hc.Name)
 		thcConf.THCOptInOnSvc = false
 	}
@@ -201,12 +202,12 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 
 	existingHC, err := h.Get(hc.Name, hc.Version(), scope)
 	if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
-		klog.V(2).Infof("Health check %q does not exist, creating (hc=%+v, bchcc=%+v)", hc.Name, hc, bchcc)
-		if err = h.create(hc, bchcc); err != nil {
+		klog.V(2).Infof("Health check %q does not exist, creating (hc=%+v, backendConfigHCConfig=%+v)", hc.Name, hc, backendConfigHCConfig)
+		if err = h.create(hc, backendConfigHCConfig); err != nil {
 			klog.Errorf("Health check %q creation error: %v", hc.Name, err)
 			return "", err
 		}
-		h.notifyAboutTHC(hc, thcConf.THCEvents)
+		h.emitTHCEvents(hc, thcConf.THCEvents)
 		// TODO(bowei) -- we don't need to fetch the self-link here as it is
 		// returned as part of the GCE call.
 		selfLink, err := h.getHealthCheckLink(hc.Name, hc.Version(), scope)
@@ -228,9 +229,9 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		klog.V(3).Infof("Resulting HC = %+v", hc)
 
 		// Then, BackendConfig will override any fields that are explicitly set.
-		if bchcc != nil {
+		if backendConfigHCConfig != nil {
 			// BackendConfig healthcheck settings always take precedence.
-			hc.UpdateFromBackendConfig(bchcc)
+			hc.UpdateFromBackendConfig(backendConfigHCConfig)
 		}
 	}
 
@@ -242,7 +243,7 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		return &ans
 	}
 
-	changes := calculateDiff(filter(existingHC), filter(hc), bchcc, thcConf.THCOptInOnSvc)
+	changes := calculateDiff(filter(existingHC), filter(hc), backendConfigHCConfig, thcConf.THCOptInOnSvc)
 	if changes.hasDiff() {
 		klog.V(2).Infof("Health check %q needs update (%s)", existingHC.Name, changes)
 		if flags.F.EnableUpdateCustomHealthCheckDescription && changes.size() == 1 && changes.has("Description") {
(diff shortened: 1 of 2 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -262,6 +263,28 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		return existingHC.SelfLink, err
 	}
 
+	// The contents of the following 'if' are temporary.
```
aojea, line 266: you mention that this happen in removal, in line 204 you have
`	if utils.IsHTTPErrorCode(err, http.StatusNotFound) {` that returns, will this code be reached in the situation you are describing?

pkg/healthchecks/healthchecks.go
```diff
@@ -262,6 +263,28 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		return existingHC.SelfLink, err
 	}
```
aojea, line 265: add more explanations to the comment, why is temporary and what fixes, if there is an open issue in github add a link referencing it

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1743,6 +1789,23 @@ func TestSyncServicePort(t *testing.T) {
 				t.Errorf("hcs.SyncServicePort(tc.sp, tc.probe) = %q, _; want = %q", gotSelfLink, tc.wantSelfLink)
 			}
 			verify()
+
+			if tc.expectedEvent > 0 {
+				fakeRecorder := fakeSingletonRecorderGetter.FakeRecorder()
+				for i := 0; i < tc.expectedEvent; i++ {
+					select {
+					case output := <-fakeRecorder.Events:
+						t.Logf("Event = %s", output)
```
aojea, line 1798: don't you have to compare the expected even with the received one?

pkg/healthchecks/healthchecks.go
DamianSawicki: Line 204 is about non-existent healthcheck and here we are talking about the removal of a BackendConfig.

pkg/healthchecks/healthchecks.go
DamianSawicki: There's only Google-internal documentation, how should I proceed in this case? It's a matter of a few weeks, it's not going to hang for long.

DamianSawicki: Ok, I just referenced the PR as the description has quite some details.

pkg/healthchecks/healthchecks_test.go

pkg/healthchecks/healthchecks.go
```diff
@@ -262,6 +263,28 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
...
 
+	// The contents of the following 'if' are temporary, see https://github.com/kubernetes/ingress-gce/pull/2181 for details.
+	if flags.F.EnableUpdateCustomHealthCheckDescription {
+		desc := &healthcheck.HealthcheckDesc{}
+		err := json.Unmarshal([]byte(existingHC.Description), desc)
```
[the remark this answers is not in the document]
DamianSawicki: Error means that the Description is not a JSON. We are in the process of switching from plain strings to JSON, depending on health check types and flags enabled, so it's not really an error to see a non-jsonified Description.

pkg/healthchecks/healthchecks.go
```diff
@@ -262,6 +263,28 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
 		return existingHC.SelfLink, err
 	}
 
+	// The contents of the following 'if' are temporary, see https://github.com/kubernetes/ingress-gce/pull/2181 for details.
```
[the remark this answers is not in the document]
DamianSawicki: Discussed this with Antonio already, there are Google-internal docs with details.

pkg/healthchecks/healthchecks.go
```diff
@@ -262,6 +263,28 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
...
+	if flags.F.EnableUpdateCustomHealthCheckDescription {
+		desc := &healthcheck.HealthcheckDesc{}
+		err := json.Unmarshal([]byte(existingHC.Description), desc)
+		if err == nil && desc.Config == healthcheck.BackendConfigHC && bchcc == nil {
```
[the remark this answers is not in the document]
DamianSawicki: BC stands for BackendConfig
HC must stand for Health Check
C is probably Config

Renaming this is a topic for a refactor, not a quick bug fix.

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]
swetharepakula: I second Cezary's comment. Since we cannot link the internal issue, it would be good to have some description in the code.

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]
swetharepakula: for debugging purposes it would be good to have a log that says that this was not a JSON description

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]
swetharepakula: It would be a good idea to describe what this trying to do as in we are only updating the HC description and nothing else

swetharepakula (commented):
The changes look good to me. I do agree with the existing feedback that we should have a more detailed comment on what this is fixing. Though this will be changed in the future, we will still be supporting this version of the code for a long time and may need the comment to understand past context.

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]
DamianSawicki: Done.

pkg/healthchecks/healthchecks.go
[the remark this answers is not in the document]

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1743,6 +1792,25 @@ func TestSyncServicePort(t *testing.T) {
 				t.Errorf("hcs.SyncServicePort(tc.sp, tc.probe) = %q, _; want = %q", gotSelfLink, tc.wantSelfLink)
 			}
 			verify()
+
+			if tc.expectedEvent > 0 {
```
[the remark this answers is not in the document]
DamianSawicki: Done.

pkg/healthchecks/healthchecks.go
```diff
@@ -258,10 +259,39 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
...
-		h.notifyAboutTHC(hc, thcConf.THCEvents)
+		h.emitTHCEvents(hc, thcConf.THCEvents)
 		return existingHC.SelfLink, err
 	}
 
+	// The content of the following 'if' guarantees that when BackendConfig is removed, the health check Description is updated accordingly
+	// even if changes.hasDiff() above is false. No other health check field is modified. The purpose is for the Description to accurately
+	// reflect the existence of a backendconfigv1.HealthCheckConfig for the service. This is temporary, see
+	// https://github.com/kubernetes/ingress-gce/pull/2181 for details.
+	if flags.F.EnableUpdateCustomHealthCheckDescription {
```
kl52752, line 270: I don't like the idea to postpone the update if it will not happened earlier. I think it would be better to put all logic related to checking against backend config description to a function fg `isDescriptionUpdateNeeded` and add condition to  line 247
```
if changes.hasDiff() || isDescriptionUpdateNeeded() {}
```
WDYT?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1747,5 +1747,164 @@ func TestSyncServicePort(t *testing.T) {
...
+		wantEventPrefix     string
+	}
+	fixture := syncSPFixture{}
+
+	var cases []*tc
+
+	shallowCopy := func(x *compute.HealthCheck) *compute.HealthCheck { y := *x; return &y }
+
+	// Don't recalculate health check on BackendConfig removal (legacy behaviour, to be changed), but update Description.
+	chc := fixture.hc()
```
kl52752, line 1778: instead of using ficture.hc() and overriding almost everything can you create new function `healthCheckFromBackendConfig` and reuse it below? (I don't see any difference here)

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1747,5 +1747,164 @@ func TestSyncServicePort(t *testing.T) {
...
+
+			verify := func() {
+				t.Helper()
+				var computeHCs []*compute.HealthCheck
+				computeHCs, _ = fakeGCE.Compute().HealthChecks().List(context.Background(), nil)
+				if len(computeHCs) != 1 {
+					t.Fatalf("Got %d healthchecks, want 1\n%s", len(computeHCs), pretty.Sprint(computeHCs))
+				}
+
+				// Filter out SelfLink because it is hard to deal with in the mock and
```
kl52752, line 1852: And description if flag is disabled

pkg/healthchecks/healthchecks.go
```diff
@@ -258,10 +259,39 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, bchcc *backendconfigv1.H
...
+		desc := &healthcheck.HealthcheckDesc{}
+		err := json.Unmarshal([]byte(existingHC.Description), desc)
+		if err != nil {
+			klog.V(3).Info("Description for healthcheck %s is not a JSON (probably a plain-text description): %s.", existingHC.Name, existingHC.Description)
```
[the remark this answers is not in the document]
DamianSawicki: > Also you can mention in this log that the update is omitted because the description is not a JSON.

Done.

> Can you do early return here and remove else.

I would prefer not to to avoid copy-pasting the following lines.
```
	klog.V(2).Infof("Health check %q already exists and needs no update", hc.Name)
	return existingHC.SelfLink, nil
```

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Good idea, done.

pkg/healthchecks/healthchecks_test.go

pkg/healthchecks/healthchecks.go
DamianSawicki: Discussed outside GitHub, done.

PR #2182, opened 2023-06-29
Use field selector to get ASM configmap

linxiulei:
Doing so avoids caching unneeded configmaps in memory.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -14,10 +14,12 @@ limitations under the License.
 package context
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


aojea:
/ok-to-test
/lgtm
/assign @bowei 

Thanks

swetharepakula:
Out of curiosity, did we see that there was a large configmap memory usage after changing to just the ASM namespace?

linxiulei:
> Out of curiosity, did we see that there was a large configmap memory usage after changing to just the ASM namespace?

No, not in practice. This PR is more of a cleanup than an optimization.

PR #2187, opened 2023-07-03
Cherry-pick #2186: Include proxy-only subnet with purpose=REGIONAL_MANAGED_PROXY when generating firewall rules for Ingress

gauravkghildiyal:
/assign @swetharepakula


cezarygerard:
/lgtm

we need the same for branch 1.23

PR #2192, opened 2023-07-06
Recalculate health check params on BC/THC removal

DamianSawicki:
The purpose of the present PR is to skip the function `mergeUserSettings()` when the removal of a _BackendConfig_ or _TransparentHealthCheck_ annotation is detected (or the removal of a health check specification from a _BackendConfig_). The rationale is that by removing custom annotations, the user should expect falling back to the default (or _Readiness Probe_) health check parameters and invoking `mergeUserSettings()` effectively preserves the custom setup.


pkg/flags/flags.go
```diff
@@ -295,6 +298,12 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.DisableFWEnforcement, "disable-fw-enforcement", false, "Disable Ingress contrller to enforce the firewall ruless. If set to true, Ingress Controller stops creating GCE firewall rules. We can only enable this if enable-firewall-cr sets to true.")
 }
 
+func Validate() {
+	if F.EnableRecalculateUHCOnBCRemoval && !F.EnableUpdateCustomHealthCheckDescription {
+		klog.Fatalf("The flag --enable-recalculate-uhc-on-backendconfig-removal cannot be used without --enable-update-hc-description.")
```
kl52752, line 303: Add this info to flag description in line 285

pkg/healthchecks/healthchecks.go
```diff
@@ -183,6 +196,36 @@ func (h *HealthChecks) emitTHCEvents(hc *translator.HealthCheck, thcEvents utils
...
+}
+
+func (h *HealthChecks) shouldRecalculateOnTHCRemoval(hc *translator.HealthCheck, thcOptIn bool) bool {
+	// This is not behind a feature flag because it should work in particular at the time of disabling the flag EnableTransparentHealthChecks.
+	hcDesc := &healthcheck.HealthcheckDesc{}
+	if err := json.Unmarshal([]byte(hc.Description), hcDesc); err != nil {
+		klog.V(3).Infof("Health check description is not JSONified: %s", hc.Description)
+		return false
+	}
+	// The existing HC is configured via Transparent Health Checks, but they are currently disabled.
```
kl52752, line 222: please update the comment that the THC annotation was removed. 
We don't have scenarios for disabling feature flag in running clusters I don't think we will support that.

pkg/healthchecks/healthchecks.go
```diff
@@ -218,10 +261,13 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, backendConfigHCConfig *b
 		return "", err
 	}
 
+	// Do not merge the existing settings and perform the full diff in calculateDiff.
+	var recalculate bool = thcConf.THCOptInOnSvc || h.shouldRecalculateOnBCRemoval(existingHC, backendConfigHCConfig) || h.shouldRecalculateOnTHCRemoval(existingHC, thcConf.THCOptInOnSvc)
```
kl52752, line 265: I think it should be called reconfigureHC and why not using ":=" instead of "var"?

extract those ors to new function like `shouldReconfigureHC`

pkg/healthchecks/healthchecks.go
```diff
@@ -218,10 +261,13 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, backendConfigHCConfig *b
...
+
 	// First, merge in the configuration from the existing healthcheck to cover
```
kl52752, line 267: put this comment under if condition

pkg/healthchecks/healthchecks.go
```diff
@@ -218,10 +261,13 @@ func (h *HealthChecks) sync(hc *translator.HealthCheck, backendConfigHCConfig *b
...
 	// the case where the user has changed healthcheck settings outside of
 	// GKE.
-	if !thcConf.THCOptInOnSvc {
+	if !recalculate {
```
kl52752, line 270: you included` hc.UpdateFromBackendConfig(backendConfigHCConfig)` in the if. This line applies changes form BC to hc. Please make sure that hc contains BC changes when needed. I'm thinking on one scenario when you swich from THC to BC.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1691,6 +1692,34 @@ func TestSyncServicePort(t *testing.T) {
...
+			tc := *tc
+			tc.recalc = true
+			tc.desc = tc.desc + " and recalc"
+			copyOfWant := *tc.wantComputeHC
+			tc.wantComputeHC = &copyOfWant
+			cases = append(cases, &tc)
+		}
+	}
+
+	/*
```
kl52752, line 1706: why this is commented out?

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1825,6 +1862,91 @@ func TestBackendConfigRemoval(t *testing.T) {
 		wantEventPrefix:     "Normal HealthcheckDescriptionUpdate Healthcheck will be updated and the only field updated is Description",
 	})
 
+	// Recalculate health check on BackendConfig removal.
+	chc = fixture.hc()
+	chc.CheckIntervalSec = 17
+	chc.Description = (&healthcheck.HealthcheckInfo{HealthcheckConfig: healthcheck.BackendConfigHC}).GenerateHealthcheckDescription()
+	wantCHC = fixture.hc()
+	cases = append(cases, &tc{
+		desc:                "recalculate on backendconfig removal",
```
kl52752, line 1871: What is expected config on BC removal? It's not clear from reading this test I would like to see clear traversal like from Backend Config based hc to Readiness probe c
please change description to reflect that and applied to others as well
example: "update hc from backendConfig based to default"

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1825,6 +1862,91 @@ func TestBackendConfigRemoval(t *testing.T) {
...
+	wantCHC = fixture.neg()
+	cases = append(cases, &tc{
+		desc:          "recalculate on thc flag disablement",
+		setup:         fixture.setupExistingHCFunc(chc),
+		sp:            testSPs["HTTP-80-neg-nil-thc"],
+		wantComputeHC: wantCHC,
+		enableTHC:     false,
+	})
+
+	// Recalculate readiness probe health check on THC removal.
```
kl52752, line 1928: Add test case to switch from THC to BC and the other round

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1912,5 +2034,130 @@ func TestBackendConfigRemoval(t *testing.T) {
...
+			healthcheckerFlags = append(healthcheckerFlags, HealthcheckFlags{EnableTHC: thc, EnableRecalculationOnBackendConfigRemoval: recalc})
+		}
+	}
+
+	healthcheckBefore := map[string]string{
+		"default": translator.DescriptionForDefaultHealthChecks,
+		"ILB":     translator.DescriptionForDefaultILBHealthChecks,
+		"NEG":     translator.DescriptionForDefaultNEGHealthChecks,
+		"RP":      translator.DescriptionForHealthChecksFromReadinessProbe,
+		"BC":      (&healthcheck.HealthcheckInfo{HealthcheckConfig: healthcheck.BackendConfigHC}).GenerateHealthcheckDescription(),
```
kl52752, line 2121: BC is taking priority over THC so THC removal should be NoOp for BC based hc

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1912,5 +2034,130 @@ func TestBackendConfigRemoval(t *testing.T) {
 	}
 }
 
+func replaceDescription(hc *translator.HealthCheck, new string) *translator.HealthCheck {
+	hc.Description = new
+	return hc
+}
+
+func TestShouldRecalculateOnBCRemoval(t *testing.T) {
```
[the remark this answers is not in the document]
DamianSawicki: What do you mean? There's a separate test `TestShouldRecalculateOnTHCRemoval`.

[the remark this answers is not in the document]
kl52752: yes right - resolved

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Sure. This test is probably overly complicated, but it tests a very simple thing: that `shouldRecalculateOnTHCRemoval` returns `true` only for `beforeKey == "THC" && afterKey == "NO-THC"`, and `false` otherwise. It does not look at annotations for _before_, so the case you are describing is `beforeKey == BC && afterKey == NO-THC`, in which case `false` should be returned. It is correct.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1912,5 +2034,130 @@ func TestBackendConfigRemoval(t *testing.T) {
...
+	healthcheckBefore := map[string]string{
+		"default": translator.DescriptionForDefaultHealthChecks,
+		"ILB":     translator.DescriptionForDefaultILBHealthChecks,
+		"NEG":     translator.DescriptionForDefaultNEGHealthChecks,
+		"RP":      translator.DescriptionForHealthChecksFromReadinessProbe,
+		"BC":      (&healthcheck.HealthcheckInfo{HealthcheckConfig: healthcheck.BackendConfigHC}).GenerateHealthcheckDescription(),
+		"THC":     (&healthcheck.HealthcheckInfo{HealthcheckConfig: healthcheck.TransparentHC}).GenerateHealthcheckDescription(),
+	}
+
+	thcAfter := //[]bool{true, false}
```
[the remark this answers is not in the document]

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1912,5 +2034,130 @@ func TestBackendConfigRemoval(t *testing.T) {
...
+		"no-BC": nil,
+		"BC":    {},
+	}
+
+	type testCase struct {
+		desc          string
+		healthchecker *HealthChecks
+		hc            *translator.HealthCheck
+		bchcc         *backendconfigv1.HealthCheckConfig
+		want          bool
```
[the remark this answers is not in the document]
DamianSawicki: > what does "want" stands for?

It stands for the return value of the tested function, like in the pattern got-want. Don't you think this name makes sense?

I renamed the tested functions and the test function in v2, please let me know if you still recommend renaming `want` to something else.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1912,5 +2034,130 @@ func TestBackendConfigRemoval(t *testing.T) {
...
+
+	testClusterValues := gce.DefaultTestClusterValues()
+	fakeGCE := gce.NewFakeGCECloud(testClusterValues)
+
+	healthcheckers := map[string]*HealthChecks{
+		"recalc-disabled": NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter(), HealthcheckFlags{}),
+		"recalc-enabled":  NewHealthChecker(fakeGCE, "/", defaultBackendSvc, NewFakeRecorderGetter(0), NewFakeServiceGetter(), HealthcheckFlags{EnableRecalculationOnBackendConfigRemoval: true}),
+	}
+
+	healthcheckBefore := map[string]string{
```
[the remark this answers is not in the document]

pkg/healthchecks/healthchecks_test.go

pkg/healthchecks/healthchecks_test.go

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Can't remember exactly, but it's possible that I realised that the bug fixed in https://github.com/kubernetes/ingress-gce/pull/2181 existed while writing this. The behaviour described in the commented out test case is incorrect, and I will remove this case. Btw. now we have similar (but correct) cases in `TestCustomHealthcheckRemoval`.

DamianSawicki: Deleted.

pkg/healthchecks/healthchecks.go
DamianSawicki: Greatly spotted! Fixed.

pkg/healthchecks/healthchecks.go

pkg/healthchecks/healthchecks.go

>  why not using ":=" instead of "var"?

No particular reason. Is one preferred over the other?

pkg/healthchecks/healthchecks.go

pkg/flags/flags.go
DamianSawicki: Done.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -1855,20 +2072,17 @@ func TestBackendConfigRemoval(t *testing.T) {
...
-						hc.Description = ""
-					}
 					return hc
 				}
 				gotHC := filter(*computeHCs[0])
 				wantHC := filter(*tc.wantComputeHC)
 
 				if !reflect.DeepEqual(gotHC, wantHC) {
-					t.Fatalf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(gotHC), pretty.Sprint(wantHC))
+					t.Errorf("Compute healthcheck is:\n%s, want:\n%s", pretty.Sprint(gotHC), pretty.Sprint(wantHC))
```
kl52752, line 2085: why changing that?

kl52752 (commented):
Overall LGTM, waiting to @swetharepakula to look at.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: I changed it during debugging to be able to see more and remembered to revert it... but then I forgot in the end.

pkg/healthchecks/healthchecks.go
```diff
@@ -183,6 +196,40 @@ func (h *HealthChecks) emitTHCEvents(hc *translator.HealthCheck, thcEvents utils
...
+
+func (h *HealthChecks) isTHCRemoved(hc *translator.HealthCheck, thcOptIn bool) bool {
+	// This is not behind a feature flag because it should work in particular at the time of disabling the flag EnableTransparentHealthChecks.
+	hcDesc := &healthcheck.HealthcheckDesc{}
+	if err := json.Unmarshal([]byte(hc.Description), hcDesc); err != nil {
+		klog.V(3).Infof("Health check description is not JSONified: %s", hc.Description)
+		return false
+	}
+	// The existing HC is configured via Transparent Health Checks, but the annotation has been removed.
+	if hcDesc.Config == healthcheck.TransparentHC && !thcOptIn {
```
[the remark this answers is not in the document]
DamianSawicki: Nice! Fixed.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Fixed.

pkg/healthchecks/healthcheck.go
```diff
@@ -57,28 +57,28 @@ func calculateDiff(old, new *translator.HealthCheck, c *backendconfigv1.HealthCh
 
 	// TODO(bowei): why don't we check Port, timeout etc.
 
-	if c == nil && !thcEnabled {
+	if c == nil && !fullDiff {
 		return &changes
 	}
 
 	// This code assumes that the changes wrt to `c` has been applied to `new`.
```
swetharepakula, line 64: To make these checks a little simpler can we not check if fullDIff is false and return early?

pkg/healthchecks/healthchecks.go
```diff
@@ -183,6 +196,37 @@ func (h *HealthChecks) emitTHCEvents(hc *translator.HealthCheck, thcEvents utils
...
+	hcDesc := &healthcheck.HealthcheckDesc{}
+	if err := json.Unmarshal([]byte(hc.Description), hcDesc); err != nil {
+		klog.V(3).Infof("Health check description is not JSONified: %s", hc.Description)
+		return false
+	}
+	// The existing HC is configured via Transparent Health Checks, but the annotation has been removed.
+	return hcDesc.Config == healthcheck.TransparentHC && !thcOptIn
+}
+
+func (h *HealthChecks) shouldRecalculateHC(existingHC *translator.HealthCheck, backendConfigHCConfig *backendconfigv1.HealthCheckConfig, thcConf utils.THCConfiguration) bool {
```
swetharepakula, line 226: nit: i would unmarshal the healthcheck once in here and pass it to the various validation functions.

pkg/healthchecks/healthchecks.go
```diff
@@ -183,6 +196,37 @@ func (h *HealthChecks) emitTHCEvents(hc *translator.HealthCheck, thcEvents utils
 
 }
 
+func (h *HealthChecks) isBackendConfigRemoved(hc *translator.HealthCheck, bchcc *backendconfigv1.HealthCheckConfig) bool {
+	if !h.healthcheckFlags.EnableRecalculationOnBackendConfigRemoval {
```
swetharepakula, line 200: nit: based on the name, this is a little weird that we don't consider the BackendConfig removed based on a flag.

Instead I would think we would check the flag at where this would be called, and only if enabled we would check if the backend config was removed.

Either a rename or slight change in the code to make it match.

swetharepakula (commented):
Mostly nits. Otherwise LGTM

pkg/healthchecks/healthcheck.go
DamianSawicki: Well, I also think that calculateDiff is overly complicated and these conditions in practise duplicate the functionality of `mergeUserSettings`. Eventually, there should be no early returns, just comparisons `old.X != new.X`. Having said that, health checks is a very sensitive issue, and a refactor here should be a separate and well-tested PR, perhaps additionally consulted. For now, I'm just renaming `fullDiff -> fullDiffOnRecalculation` to make clear what we do.

PR #2194, opened 2023-07-11
Flag for THC port

DamianSawicki:
Integer flag to overwrite the default THC port.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -77,6 +77,7 @@ var (
 		GKEClusterType                   string
 		HealthCheckPath                  string
 		HealthzPort                      int
+		THCPort                          int
 		InCluster                        bool
 		IngressClass                     string
 		KubeConfigFile                   string
@@ -283,6 +284,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableTransparentHealthChecks, "enable-transparent-health-checks", false, "Enable Transparent Health Checks.")
 	flag.BoolVar(&F.EnableUpdateCustomHealthCheckDescription, "enable-update-hc-description", false, "Update health check Description when it is customized with BackendConfig CRD.")
 	flag.BoolVar(&F.EnableRecalculateUHCOnBCRemoval, "enable-recalculate-uhc-on-backendconfig-removal", false, "Recalculate health check parameters when BackendConfig is removed from service. This flag cannot be used without --enable-update-hc-description.")
+	flag.IntVar(&F.THCPort, "transparent-health-checks-port", 7877, "The port for Transparent Health Checks. It must be aligned with Transparent Health Check controller server. This flag only works when --enable-transparent-health-checks is enabled.")
 	flag.BoolVar(&F.EnablePinhole, "enable-pinhole", false, "Enable Pinhole firewall feature")
 	flag.BoolVar(&F.EnableL4ILBDualStack, "enable-l4ilb-dual-stack", false, "Enable Dual-Stack handling for L4 Internal Load Balancers")
 	flag.BoolVar(&F.EnableL4NetLBDualStack, "enable-l4netlb-dual-stack", false, "Enable Dual-Stack handling for L4 External Load Balancers")
@@ -301,6 +303,11 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 func Validate() {
 	if F.EnableRecalculateUHCOnBCRemoval && !F.EnableUpdateCustomHealthCheckDescription {
 		klog.Fatalf("The flag --enable-recalculate-uhc-on-backendconfig-removal cannot be used without --enable-update-hc-description.")
+	}
+
+	// There is no information available whether --transparent-health-checks-port is set, but it is certainly set if F.THCPort stores a non-default value.
+	if F.THCPort != 7877 && !F.EnableTransparentHealthChecks {
+		klog.Fatalf("The flag --transparent-health-checks-port cannot be used without --enable-transparent-health-checks.")
 	}
 }
 
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -21,14 +21,12 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
-	"time"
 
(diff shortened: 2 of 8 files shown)


pkg/flags/flags.go
```diff
@@ -283,6 +284,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableTransparentHealthChecks, "enable-transparent-health-checks", false, "Enable Transparent Health Checks.")
 	flag.BoolVar(&F.EnableUpdateCustomHealthCheckDescription, "enable-update-hc-description", false, "Update health check Description when it is customized with BackendConfig CRD.")
 	flag.BoolVar(&F.EnableRecalculateUHCOnBCRemoval, "enable-recalculate-uhc-on-backendconfig-removal", false, "Recalculate health check parameters when BackendConfig is removed from service. This flag cannot be used without --enable-update-hc-description.")
+	flag.IntVar(&F.THCPort, "transparent-health-checks-port", 7877, "The port for Transparent Health Checks.")
```
kl52752, line 287: please add the information that it is used only when thc feature is enabled

kl52752: and maybe some info that it cannot be set to 0 otherwise we will fallback to default

pkg/healthchecks/healthchecks.go
```diff
@@ -54,13 +58,17 @@ type HealthChecks struct {
...
 	EnableRecalculationOnBackendConfigRemoval bool
+	THCPort                                   int64
 }
 
 // NewHealthChecker creates a new health checker.
 // cloud: the cloud object implementing SingleHealthCheck.
 // defaultHealthCheckPath: is the HTTP path to use for health checks.
 func NewHealthChecker(cloud HealthCheckProvider, healthCheckPath string, defaultBackendSvc types.NamespacedName, recorderGetter RecorderGetter, serviceGetter ServiceGetter, flags HealthcheckFlags) *HealthChecks {
 	ci := generateClusterInfo(cloud.(*gce.Cloud))
+	if flags.THCPort == 0 {
```
kl52752, line 69: I don't like this check 0 this is a misconfiguration we should check on system startup and then print the error and either fallback to default or diable the feature.

pkg/translator/healthchecks_test.go
```diff
@@ -97,7 +97,7 @@ func TestOverwriteWithTHC(t *testing.T) {
 			UnhealthyThreshold: 10,
 			HealthyThreshold:   1,
 			Type:               "HTTP",
-			Description:        (&healthcheck.HealthcheckInfo{HealthcheckConfig: healthcheck.TransparentHC}).GenerateHealthcheckDescription(),
+			Description:        "Kubernetes L7 transparent health check.",
```
kl52752, line 100: lets include the port in description

pkg/healthchecks/healthchecks.go
DamianSawicki: I was actually considering adding a comment here. This check is only for tests, so that in some unrelated tests we can just write `healthchecks.HealthcheckFlags{}` in a call to `NewHealthChecker`. In non-test code, we have
```
	healthchecks.HealthcheckFlags{
		EnableTHC: flags.F.EnableTransparentHealthChecks,
		EnableRecalculationOnBackendConfigRemoval: flags.F.EnableRecalculateUHCOnBCRemoval,
		THCPort: int64(flags.F.THCPort),
	}
```

There are probably many invalid values that a user can pass as `--transparent-health-checks-port`, so checking particularly for `0` probably does not make sense as a part of flag validation. As I said, here it is for golang defaults rather than for user mistake.

I suggest that I add a comment in the code explaining this condition. WDYT?

pkg/flags/flags.go
DamianSawicki: > please add the information that it is used only when thc feature is enabled

Will do. Should I fail the application if thc port is set but thc is not enabled like I did for healthcheck recalculation? This way there's less room for users to misconfigure stuff.

pkg/translator/healthchecks_test.go
DamianSawicki: Sorry, there should be no change here. I must have done a mistake while merging. As for the JSON description format, port was not included in the design.

kl52752: I think that we can leverage the Json format and pass thing that are useful for debugging.

pkg/healthchecks/healthchecks.go
DamianSawicki: > but this function is used in no-test code also so you should not include this validation just for test and use it everywhere. It is inconsistent with the state.

Removed.

> As for many invalid values maybe instead of validation we should say that this port needs to align with transparent custom health check server.

Done.

pkg/healthchecks/healthchecks_test.go
```diff
@@ -467,12 +467,14 @@ func TestHealthCheckUpdate(t *testing.T) {
 func TestEnableTHC(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
 
+	const thcPort int32 = 5678
```
kl52752, line 470: it doesn't m,ake sense to do this int32 and the cast to 64 right?

pkg/healthchecks/healthchecks.go
```diff
@@ -54,6 +54,7 @@ type HealthChecks struct {
 type HealthcheckFlags struct {
 	EnableTHC                                 bool
 	EnableRecalculationOnBackendConfigRemoval bool
+	THCPort                                   int64
```
kl52752, line 57: why did yoy make it int64? I think that u16 will be sufficient but u32 is fine too

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Sounds right 😅

pkg/healthchecks/healthchecks.go
DamianSawicki: For consistency with the type of `translator.HealthCheck.HTTPHealthCheck.Port`.

pkg/healthchecks/healthchecks_test.go
DamianSawicki: Fixed.

PR #2198, opened 2023-07-24
Call init function in metrics.go

DamianSawicki:
As far as I can tell:
 * `func init()` in `metrics.go` is never called,
 * certain metrics are not exported:

1. `number_of_ingresses`
2. `number_of_l4_ilbs`
3. `number_of_l4_netlbs`
4. `number_of_negs`
5. `number_of_service_attachments`
6. `number_of_service_ports`
7. `number_of_services`
8. `component_version`

Calling `init()` seems to fix the above problem.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -121,7 +121,7 @@ func (netlbCount *netLBFeatureCount) record() {
 }
 
 // init registers ingress usage metrics.
-func init() {
+func Init() {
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -121,7 +121,7 @@ func (netlbCount *netLBFeatureCount) record() {
 }
 
 // init registers ingress usage metrics.
-func init() {
```
aojea, line 124: init() runs as soons as the module is loaded, is a golang thing

DamianSawicki: Thanks @aojea! I've already worked this out :) We should not log with `V(3)` in the `init()` function because it's not visible even when the controller is run with a higher `-v` level.

PR #2201, opened 2023-07-26
Metrics for multinetworking

mmamczur:
This adds multinetworking latency specific metrics. Also it adds generic L4 metrics `l4_ilbs` and `l4_netlbs` with multinet label. These metrics are meant to be extended with other labels to make cross querying possible.


pkg/loadbalancers/l4.go
```diff
@@ -76,6 +76,24 @@ type L4ILBSyncResult struct {
 	StartTime             time.Time
 }
 
+func NewL4ILBSyncResult(svc *corev1.Service, startTime time.Time) *L4ILBSyncResult {
+	result := &L4ILBSyncResult{
+		Annotations:           make(map[string]string),
+		StartTime:             startTime,
+		SyncType:              SyncTypeCreate,
+		DualStackMetricsState: metrics.InitServiceDualStackMetricsState(svc, &startTime),
```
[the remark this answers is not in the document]
code-elinka: I see it was done on [line 312](https://github.com/kubernetes/ingress-gce/pull/2201/files#diff-c28bf6e0840e72d4c64115b6258b48b0965c32daa537a67b2e67ffc06a881bdbL312). But I don't understand why Dualstack metrics are not kept separately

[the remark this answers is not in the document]
mmamczur: this is just the state object. it is not really used if dual stack is not enabled - see here: https://github.com/kubernetes/ingress-gce/blob/master/pkg/l4lb/l4controller.go#L531

pkg/metrics/types.go
```diff
@@ -107,6 +119,8 @@ type L4NetLBServiceState struct {
 	IsManagedIP bool
 	// IsPremiumTier specifies if network tier for forwarding rule is premium.
 	IsPremiumTier bool
+	// IsMultinet
```
code-elinka, line 122: nit: elaborate a little like it's done with others

code-elinka:
/assign @panslava 
Slavik, can we ask you to review this change? We need to finish this and another PR for metrics based on this one before freeze on 11th of Aug. 
Feel free to ping me if you need to review something as well :)

pkg/metrics/l4_metrics.go
```diff
@@ -80,6 +102,14 @@ func newL4ontrollerMetrics(l4NetLBProvisionDeadline time.Duration) *l4Controller
 	}
 }
 
+func InitL4ILBServiceState(startTime *time.Time) L4ILBServiceState {
```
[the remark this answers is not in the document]
panslava: I am also a bit scared, that we have 2 mutexes now, separate for l4 metrics, and separate for metrics, but l4 metrics are always called from "metrics"

Maybe it will work fine, but I think it is better to double check if we need this, maybe we could've went with what I did in l4_dualstack_metrics

sorry for just opinion comment

pkg/metrics/l4_metrics.go
```diff
@@ -17,13 +17,19 @@ limitations under the License.
 package metrics
```
panslava, line 17: Maybe I am not really aware of idea of previous PR, but metrics that are stored in this file are a bit confusing

Hard to say by the names, what are the difference, and also we are not being consistent with naming, somewhere we say `l4ILBCount` but function that computes was renamed to computeL4ILB*Feature*Metrics

It was maybe fine when it was only 2 simple counts, which had 1:1 relation with all logic, like metric <-> state <-> compute functions

Now state stores labels also for multinet, but metrics are actually different

idk what to do to be honest, I have too little brain power to keep all the names in bio-memory cache... maybe could we move all multinet metrics to the separate file, like I did in dualstack? Or maybe at least make naming consistent. Or maybe you have other idea how to improve it

pkg/metrics/l4_metrics.go
```diff
@@ -125,8 +160,8 @@ func (im *ControllerMetrics) DeleteL4NetLBService(svcKey string) {
 	delete(im.l4ControllerMetrics.l4NetLBServiceMap, svcKey)
 }
 
-// computeL4ILBMetrics aggregates L4 ILB metrics in the cache.
-func (l4 *l4ControllerMetrics) computeL4ILBMetrics() map[feature]int {
+// computeL4ILBFeatureMetrics aggregates L4 ILB metrics in the cache.
+func (l4 *l4ControllerMetrics) computeL4ILBFeatureMetrics() map[feature]int {
```
panslava, line 164: I think we should not use `l4` as `l4ControllerMetrics`, I think it is L4 or L4Controller

maybe let's do `l4cm`

pkg/metrics/l4_metrics.go
panslava: Also, maybe that was my problem that I started naming everything we had in DualStack with "DualStack" in it

Because in dualstack, I basically changed some not the best past decisions, which were hard to extend with what I find better version of storing services state. So let's say if L4DualStackServiceState would be just called "BetterL4ServiceState" or even "ServiceState" you could've just embed L4MultiNetServiceLabels into it (as I did embed L4DualStackServiceLabels) and then you would not need to copy a lot of code, like identifying "status" labels, and you could still have separate metric for it.

And sorry, I still don't know what to do with that, I don't think I have the power to say "refactor all of this"...

panslava: And I feel you took some not the best patterns from past, like storing 

InSuccess
IsUserError

I think they are too intercepting (for example InSuccess: true IsUserError: true does not makes sense), and that's why just having label status as I have in dualstack is better

panslava: I think the best way would be is to rename DualStackStates and etc and embed your implementation into those

panslava (changes requested):
Sorry, my comments are mostly very vague and not simply fixable... But I am a bit scared that this could make our code base harder to understand, which we should try not to do

I would be happy if we could improve something, we can also discuss together, I am not hard on my opinions

pkg/metrics/l4_metrics.go
mmamczur: lot of comments to address

so in general we have a mess in the metrics area
there are these old metrics named `number_of_l4_*`. These work in a particular way where there is just one label named feature. For example Error is a feature there. So effectively you can't filter to know how many services using feature X are in error. And you need this for SLO purposes. You know that because you created a separate metric for dual stack.

So, we decided to create a new metric that will be generic and that you could extend. The naming is a problem there because now we have 2 sets of metrics that do the same but differently. That's why I renamed parts of old functions to say 'feature'.

Another option was to do the same as dual stack - have a separate metric - but in the team we decided against doing this.

As for having bools InSuccess IsUserError - I reused the existing service map. We could refactor this but I'd prefer not to do it now.

panslava: I am not sure, if you could really have general metrics that you will keep extending, with fields like multinet, then with strong session affinity, then with something else. Would not cardinality become a problem?

Maybe we should rename old metrics to legacy_metrics ? at least on the variable names level, not the metric itself

pkg/l4lb/l4controller.go
```diff
@@ -237,9 +237,16 @@ func (l4c *L4Controller) processServiceCreateOrUpdate(service *v1.Service) *load
 	if err != nil {
 		return &loadbalancers.L4ILBSyncResult{Error: err}
 	}
-	network, err := network.ServiceNetwork(service, l4c.networkLister, l4c.gkeNetworkParamSetLister, l4c.ctx.Cloud, klog.TODO())
+	serviceNetwork, err := network.ServiceNetwork(service, l4c.networkLister, l4c.gkeNetworkParamSetLister, l4c.ctx.Cloud, klog.TODO())
 	if err != nil {
-		return &loadbalancers.L4ILBSyncResult{Error: err}
+		result := loadbalancers.NewL4ILBSyncResult(service, startTime)
```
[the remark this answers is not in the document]
mmamczur: I agree, it's bad

Also I realized later that resolving the network should happen later in `EnsureLB()`. I even tried to refactor that (unfortunately it's not so trivial because of how all these structs are set up and also because ILB and NetLB controllers are structured). Maybe I'll come back to that and attempt to unify but for now I just put the metric stuff here.

pkg/l4lb/l4controller.go
```diff
@@ -237,9 +237,16 @@ func (l4c *L4Controller) processServiceCreateOrUpdate(service *v1.Service) *load
...
+		result.Error = err
+		result.SyncType = loadbalancers.SyncTypeCreate
```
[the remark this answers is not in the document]
mmamczur: it will, removing

pkg/metrics/l4_metrics.go
panslava: What would you say if you extend L4DualStackServiceState with MultiNet and maybe rename L4DualStackServiceState to just L4ServiceState? I mean only keeping the state of the service in the same struct, not use the same metric

pkg/metrics/l4_metrics.go
```diff
@@ -39,6 +45,22 @@ var (
 		},
 		[]string{label},
 	)
+
+	l4ILBs = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "l4_ilbs",
+			Help: "Metric about ILBs",
```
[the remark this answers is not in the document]
mmamczur: modifying the help msg. It's meant to be extended

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -456,7 +456,12 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.L4
 	networkInfo, err := network.ServiceNetwork(service, lc.networkLister, lc.gkeNetworkParamSetLister, lc.ctx.Cloud, klog.TODO())
 	if err != nil {
 		klog.Errorf("Failed to get network for service %s/%s, err: %v", service.Namespace, service.Name, err)
-		return &loadbalancers.L4NetLBSyncResult{Error: err}
+		result := loadbalancers.NewL4SyncResult(loadbalancers.SyncTypeCreate, service)
```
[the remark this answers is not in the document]
mmamczur: done

pkg/metrics/l4_metrics.go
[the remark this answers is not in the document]
mmamczur: actually it looks like you found a bug where I was using the wrong mutex. I just moved the function to be in the L4 type so I can't confuse both. Export is called from the general metrics controller but never holds 2 mutexes

The thing is that I took away the L4 metrics to a separate file because the original one has just became really big

pkg/metrics/l4_metrics.go
mmamczur: > Would not cardinality become a problem?

it will at some point.

we can rename to legacy - it's actually a good idea, I'll do it

mmamczur: > What would you say if you extend L4DualStackServiceState with MultiNet and maybe rename L4DualStackServiceState to just L4ServiceState? 
We could do it but then when I look at L4DualStackServiceState it's a struct that inherits from `L4DualStackServiceLabels`. I could add other Labels type but then it also has `Status`. So when you inherit from both it will complain about `Status` field references ambiguous.

TBH I'm not sure how to resolve all this without redoing the whole PR all over again

pkg/metrics/types.go
mmamczur: done

pkg/l4lb/l4controller.go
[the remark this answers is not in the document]
mmamczur: I've set the user error in dualstackmetricsstate

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -456,7 +456,12 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service) *loadbalancers.L4
...
+		result.MetricsState.IsUserError = utils.IsUserError(err)
+		result.MetricsState.IsMultinet = network.HasMultiNetSelector(service)
```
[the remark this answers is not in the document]
mmamczur: added

pkg/loadbalancers/l4.go
```diff
@@ -305,19 +325,8 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
-	}
-
-	// If service already has an IP assigned, treat it as an update instead of a new Loadbalancer.
-	// This will also cover cases where an external LB is updated to an ILB, which is technically a create for ILB.
-	// But this is still the easiest way to identify create vs update in the common case.
-	if len(svc.Status.LoadBalancer.Ingress) > 0 {
-		result.SyncType = SyncTypeUpdate
-	}
+	result := NewL4ILBSyncResult(svc, startTime)
+	result.MetricsState.IsMultinet = !l4.network.IsDefault
```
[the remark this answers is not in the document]
mmamczur: isDefault means the network is the cluster network - essentially it's the network that would be used before any multinet stuff came.
We could call it isClusterNetwork I guess but not in this PR

pkg/loadbalancers/l4netlb.go
```diff
@@ -179,6 +179,7 @@ func (l4netlb *L4NetLB) checkStrongSessionAffinityRequirements() *utils.UserErro
 // This function does not link instances to Backend Service.
 func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service) *L4NetLBSyncResult {
 	result := NewL4SyncResult(SyncTypeCreate, svc)
+	result.MetricsState.IsMultinet = !l4netlb.networkInfo.IsDefault
```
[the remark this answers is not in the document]
mmamczur: moved

pkg/metrics/l4_metrics.go
panslava: @mmamczur why Status of DualStack is not enough? Could not you just use it? It shows the status of service "Error, UserError, Success, PersistentError"

panslava: I think the main problem is that those variables in Statuses and Metrics have "dualstack" prefix in name, if they would not, would be more elegant to just use them

mmamczur: no no, the status is good. I'll try to explain, the whole DualStack (that we'd rename) looks like this:

```
type L4DualStackServiceState struct {
	L4DualStackServiceLabels
	// FirstSyncErrorTime specifies the time timestamp when the service sync ended up with error for the first time.
	FirstSyncErrorTime *time.Time
}
```
and then we have labels
```
type L4DualStackServiceLabels struct {

	...

	Status L4ServiceStatus
}
```
The labels struct has to stay like this or otherwise dual stack metric would be calculated wrong. But it has the 
`Status` field because the metric has it.
Now how would I extend `L4DualStackServiceState` with another labels struct? it also needs `Status` and I can't have status in both because the compiler will complain (unless I address each status separately).

1. let's replace the booleans with `ServiceState` - some refactor needed
2. let's convert booleans to State only when calculating the metric
3. lets rename `L4DualStackServiceState` to `L4ServiceState`. Remove status from any 'labels' struct and then, only when calculating metrics use some other type like:
```
struct {
  L4DualStackServiceLabels
  Status L4ServiceStatus 
}
```
You'd have to manually set Status each time.
4. ???

If we rename and reuse existing DualStack stuff, do we put it in l4_metrics.go? Do we revert the change that introduced this file?  (https://github.com/kubernetes/ingress-gce/pull/2199)

panslava: Sorry, there is one thing I don't understand -- why do you need different Status? Why DualStack Status is not enough, and you just use it in new l4 metrics?

mmamczur: Dual stack status is enough and good. The problem is how these structured are composed, with Status being embedded in `L4DualStackServiceLabels`.

panslava: Can we move status on upper level then (remove from `L4DualStackServiceLabels`)

Or we could also rename `L4DualStackServiceLabels` ti `L4ServiceLabels` and put there multinet labels too

mmamczur: we can't since metric calculation depends on Labels being the key in the map with counters. If we remove status from there it won't be counted properly

mmamczur:
obsoleted by #2223

PR #2208, opened 2023-07-28
Update GKENetworkParamSet to use v1 version of the CRD

mmamczur:
This is what is installed in multinet clusters.

Also upgrades cloud-provider-gcp dependency to a newer version.


aojea:
/test pull-ingress-gce-e2e 

should be fixed by https://github.com/kubernetes/kubernetes/pull/119671

PR #2216, opened 2023-07-31
Fix error message, arguments were misplaced, restructure unit tests

panslava:
What a miss :D

/assign @code-elinka @aojea

### DIFF
--- a/pkg/loadbalancers/l4_test.go
+++ b/pkg/loadbalancers/l4_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -50,6 +51,8 @@ const (
 // TODO Uncomment after https://github.com/kubernetes/kubernetes/pull/87667 is available in vendor.
 // eventMsgFirewallChange = "XPN Firewall change required by network admin"
 )
+
+var noInternalIPv6InSubnetError = regexp.MustCompile("subnet [a-z]([-a-z0-9]*[a-z0-9])? does not have internal IPv6 ranges, required for an internal IPv6 Service. You can specify an internal IPv6 subnet using the \"networking.gke.io/load-balancer-subnet\" annotation on the Service")
 
 func getFakeGCECloud(vals gce.TestClusterValues) *gce.Cloud {
 	fakeGCE := gce.NewFakeGCECloud(vals)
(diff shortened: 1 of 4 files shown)


pkg/loadbalancers/l4_test.go
```diff
@@ -1651,6 +1654,63 @@ func TestEnsureInternalDualStackLoadBalancer(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+
+			ipFamilies := []v1.IPFamily{v1.IPv4Protocol, v1.IPv6Protocol}
+			svc := test.NewL4ILBDualStackService(8080, v1.ProtocolTCP, ipFamilies, v1.ServiceExternalTrafficPolicyTypeCluster)
```
code-elinka, line 1684: Could you please remind why it is only ExternalTrafficPolicy=Cluster, not Local? Probably it jumed out of my head.

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -614,22 +617,6 @@ func TestEnsureIPv6ExternalLoadBalancerCustomSubnet(t *testing.T) {
 	svc.Annotations = result.Annotations
 	assertDualStackNetLBResourcesWithCustomIPv6Subnet(t, l4NetLB, nodeNames, "another-subnet")
 
-	// change to internal IPv6 subnet and expect error
```
code-elinka, line 617: Could you please add a comment about the logic of https://github.com/kubernetes/ingress-gce/pull/2216/files#diff-43bde103b548631af39e388b8ebe203e970ed5f4957ea6be8d710746230fb3d1L569 ? It seems to be long

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -647,6 +634,63 @@ func TestEnsureIPv6ExternalLoadBalancerCustomSubnet(t *testing.T) {
 	assertDualStackNetLBResourcesDeleted(t, l4NetLB)
 }
 
+func TestDualStackNetLBBadCustomSubnet(t *testing.T) {
```
code-elinka, line 637: Do we have a test with wrong subnet name only for dualstack?

code-elinka:
Logic of the test looks good. I would also move new tests closer to original subnet tests so it's easier to catch that some of the cases are out of standard test.

pkg/loadbalancers/l4_test.go
panslava: it just does not matter for this test case, we can not really test every possible combination, and etp=Local does not change logic in this case, 

we could omit that, but NewL4ILBDualStackService requires to specify

pkg/loadbalancers/l4netlb_test.go
panslava: added a comment

panslava:
applied comments, moved TestDualStackILBBadSubnet close to TestEnsureInternalLoadBalancerCustomSubnet

ready for re-review

pkg/loadbalancers/l4netlb_test.go
code-elinka: Okay, I got it now, yeah, my question was if we want to test that behaviour for ipv4 only as well. Thank you for explaining it

pkg/loadbalancers/l4_test.go
code-elinka: got it!

PR #2218, opened 2023-08-04
Refactor multinetwork resolution.

mmamczur:
Push network resolution down to `EnsureLB` functions so that the sync result can be returned consistently. Add Resolver interface to avoid passing down network and gkenetworkparamset listers and allow to use fake resolver in tests.

### DIFF
--- a/pkg/healthchecksl4/healthchecksl4.go
+++ b/pkg/healthchecksl4/healthchecksl4.go
@@ -63,16 +63,14 @@ type l4HealthChecks struct {
 	hcProvider          healthChecksProvider
 	cloud               *gce.Cloud
 	recorder            record.EventRecorder
-	network             network.NetworkInfo
 }
 
-func NewL4HealthChecks(cloud *gce.Cloud, recorder record.EventRecorder, network network.NetworkInfo) *l4HealthChecks {
+func NewL4HealthChecks(cloud *gce.Cloud, recorder record.EventRecorder) *l4HealthChecks {
 	return &l4HealthChecks{
 		sharedResourcesLock: sharedLock,
 		cloud:               cloud,
 		recorder:            recorder,
 		hcProvider:          healthchecksprovider.NewHealthChecks(cloud, meta.VersionGA),
-		network:             network,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -548,7 +546,7 @@ func (lc *L4NetLBController) emitEnsuredDualStackEvent(service *v1.Service) {
 		"Successfully ensured %v External LoadBalancer resources", strings.Join(ipFamilies, " "))
 }
 
-func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, networkInfo *network.NetworkInfo) error {
+func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, linkWithNEGs bool) error {
```
cezarygerard, line 549: please convert the second argument to typed enum

type LinkType int64

const (
	IstanceGroupLink LinkType = 0
	NegLink LinkType = 1
)

pkg/neg/controller.go
```diff
@@ -208,30 +207,29 @@ func NewController(
...
+		recorder:              recorder,
+		zoneGetter:            zoneGetter,
+		cloud:                 cloud,
+		namer:                 namer,
+		l4Namer:               l4Namer,
+		defaultBackendService: defaultBackendService,
+		hasSynced:             hasSynced,
+		ingressLister:         ingressInformer.GetIndexer(),
+		serviceLister:         serviceInformer.GetIndexer(),
+		networkResolver:       network.NewNetworksResolver(networkIndexer, gkeNetworkParamSetIndexer, cloud),
```
cezarygerard, line 223: put the logger into this constructor method, not as the ServiceNetwork method arg

pkg/healthchecksl4/healthchecksl4.go
```diff
@@ -63,16 +63,14 @@ type l4HealthChecks struct {
 	hcProvider          healthChecksProvider
 	cloud               *gce.Cloud
 	recorder            record.EventRecorder
-	network             network.NetworkInfo
 }
 
-func NewL4HealthChecks(cloud *gce.Cloud, recorder record.EventRecorder, network network.NetworkInfo) *l4HealthChecks {
+func NewL4HealthChecks(cloud *gce.Cloud, recorder record.EventRecorder) *l4HealthChecks {
```
cezarygerard, line 68: why not pass network resolver to NewL4HealthChecks and have it as field in l4HealthChecks?

the L4HealthChecks interface methods will be much cleaner

mmamczur: then I resolve the network twice which requires retrieving and checking K8s resources

pkg/neg/controller.go

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -486,32 +487,15 @@ func TestProcessServiceCreate(t *testing.T) {
 }
 
 func TestProcessMultinetServiceCreate(t *testing.T) {
+	flags.F.EnableMultiNetworking = true
```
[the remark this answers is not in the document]
mmamczur: done

pkg/neg/controller.go
cezarygerard: it's a bummer that golang does not have  proper enums
but this reads better `err = controller.ensureBackendLinking(svc, instanceGroupLink)` than this `err = controller.ensureBackendLinking(svc, false)`

mmamczur: I'm guessing this comment was meant to be above. We can leave that with the enum, it's ok. Alternatively have another function altoghether, but there is some common setup that would be duplicated then.

PR #2229, opened 2023-08-14
Remove unused flag --enable-backendconfig-healthcheck

DamianSawicki:
The flag `--enable-backendconfig-healthcheck` was made void in https://github.com/kubernetes/ingress-gce/pull/2099, and the present PR removes it altogether.


gauravkghildiyal:
As per the discussion, let's put this on hold.

/hold

PR #2232, opened 2023-08-16
Refactor ServerErrror helper functions

swetharepakula:
* Remove infinite for loop since errors.As finds the relevant error
 * Add tests to validate the functions work for multi-wrapped errors

/assign @bowei 
/cc @gauravkghildiyal

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -304,12 +304,7 @@ func IsGCEServerError(err error) bool {
 	if !errors.As(err, &gerr) {
 		return false
 	}
-	for {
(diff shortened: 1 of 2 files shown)


gauravkghildiyal:
Thanks for making this change.

This LGTM.

bowei:
Did we test the previous behavior to see what happens in the original case?

swetharepakula:
Yes, we did, anytime there is a "%w:%w" error, the original functions result in an infinite loop

bowei:
How does the controller work then? Wouldn't all of the syncs get stuck?

PR #2225, opened 2023-08-10
[Cherry-pick #2134]  Delete finalizer last in L4 RBS controller

cezarygerard:
Move deleting the finalizer to the very last part of the sync

Otherwise deleting annotations may result in error and retry

to reviewers:
wait for internal approval on this cherry pick

PR #2134, opened 2023-05-22
Delete finalizer last in L4 RBS controller

panslava:
/assign @cezarygerard @aojea

PR #2234, opened 2023-08-16
Revert "[Cherry-pick #2134]  Delete finalizer last in L4 RBS controller"

aojea:
Reverts kubernetes/ingress-gce#2225

It does not have a approval and I think the problem is the annotation removal, if we are going to delete the Service we don't need to remove the annotations first AFAIK

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -346,7 +346,6 @@ func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service) *lo
 			return result
 		}
 	}
-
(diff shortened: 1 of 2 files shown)


panslava:
We need to remove annotations when it is not actually a whole service deletion — but transition to different type (netlb to ilb or to clusterip or whatever else)

aojea:
> We need to remove annotations when it is not actually a whole service deletion — but transition to different type (netlb to ilb or to clusterip or whatever else)

then remove only when DeletionTimestamp is not nil

panslava:
Do I get it right, that the main reason of this revert is because no approval, otherwise, we could've left it as is?

aojea:
> Do I get it right, that the main reason of this revert is because no approval, otherwise, we could've left it as is?

@panslava no approval and https://github.com/kubernetes/ingress-gce/pull/2236 ;)

PR #2236, opened 2023-08-17
l4 controller deletion vs mutation

aojea:
The l4 controller can stop processing a Service in two cases, when it has been deleted (it has to remove the finalizer) or it has being mutated and is no longer managed by the controller.

If the Service is being mutated the controller removes the custom annotations and status before removing the finalizer.

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -334,29 +334,35 @@ func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service) *lo
 		l4c.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancerFailed", "Error deleting load balancer: %v", result.Error)
 		return result
 	}
-	// Reset the loadbalancer status first, before resetting annotations.
-	// Other controllers(like service-controller) will process the service update if annotations change, but will ignore a service status change.
-	// Following this order avoids a race condition when a service is changed from LoadBalancer type Internal to External.
-	if err := updateServiceStatus(l4c.ctx, svc, &v1.LoadBalancerStatus{}); err != nil {
-		l4c.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
-			"Error resetting load balancer status to empty: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -334,29 +334,35 @@ func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service) *lo
...
-	} else {
-		if err := updateL4ResourcesAnnotations(l4c.ctx, svc, nil); err != nil {
-			l4c.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "DeleteLoadBalancer",
-				"Error resetting resource annotations for load balancer: %v", err)
-			result.Error = fmt.Errorf("failed to reset resource annotations, err: %w", err)
-			return result
+
+		// Also remove any ILB annotations from the service metadata
+		if l4c.enableDualStack {
+			if err := updateL4DualStackResourcesAnnotations(l4c.ctx, svc, nil); err != nil {
```
aojea, line 353: I still don't have clear why the annotations are removed @panslava , are these annotations added by the end user or by the controller?
if the former, the users must be the one updating them in case he wants to mutate the service

panslava: After we ensure service -- we add annotations to it, in L4 I think all annotations just saying name of resources (like "forwarding-rule: aaaa-l4-forwarding-rule")

When service is migrated to some other type, those LB resources are deleted by our controller, so we should also remove annotations, automatically by our controller

aojea: thanks for educating me 

😄

aojea:
error looks legit

> I0817 11:24:58.079736   16816 event.go:298] Event(v1.ObjectReference{Kind:"Service", Namespace:"default", Name:"ilbtest", UID:"", APIVersion:"v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'SyncLoadBalancerFailed' Error syncing load balancer: IP configuration error: "" googleapi: Error 409: IP_IN_USE_BY_ANOTHER_RESOURCE - IP '1.1.1.1' is already being used by another resource.

cezarygerard:
looks good

we nee the same in l4netlbcontroller

PR #2261, opened 2023-09-06
Skip user errors from retrying to provision L4 service

cezarygerard:


aojea:
it makes sense to not retry errors we know we'll not be able to reconcile, I don't know where is the best way to implement it though

pkg/l4lb/l4lbcommon.go
```diff
@@ -111,3 +111,10 @@ func isHealthCheckDeleted(cloud *gce.Cloud, hcName string) bool {
 	_, err := composite.GetHealthCheck(cloud, meta.GlobalKey(hcName), meta.VersionGA)
 	return utils.IsNotFoundError(err)
 }
+
+func skipUserError(err error) error {
```
mmamczur, line 115: if we skip, will this still be emitted as event for the service? so that the users get the error message

code-elinka:
> it makes sense to not retry errors we know we'll not be able to reconcile, I don't know where is the best way to implement it though

@aojea Maybe we can use `t.queue.Forget(key)` for it

cezarygerard:
L4 controller has accces to the following queue interface
```
type TaskQueue interface {
	Run()
	Enqueue(objs ...interface{})
	Shutdown()
	Len() int
	NumRequeues(obj interface{}) int
}
```

there is no access to underlying queue
image

pkg/l4lb/l4lbcommon.go
cezarygerard: yes, we emit the error imperatively

PR #2272, opened 2023-09-25
[1.19] Update go version and some dependencies

songrx1997:
Upgrade go version to 1.20
Upgrade github.com/prometheus/client_golang to v1.11.1
Upgrade k8s.io/client-go to v0.27.0

/assign @swetharepakula


swetharepakula:
Do you mind making these into 3 commits: 
(1) Removal of the v1beta1 adapter
(2) go.mod and go.sum changes
(3) vendor update

songrx1997:
> Do you mind making these into 3 commits: (1) Removal of the v1beta1 adapter (2) go.mod and go.sum changes (3) vendor update

Hi @swetharepakula, I merged some of those commits as your suggestion, but didn't find a good way to merge the rest. Do you think it's ok for you to review?

pkg/apis/backendconfig/v1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
swetharepakula, line 1: will these changes be overridden when we regenerate these files?

swetharepakula: Had an offline conversation with @songrx1997, who re-ran the generation scripts and confirmed that these would not get overwritten.

PR #2273, opened 2023-09-25
[1.20] Update go version and some dependendies

songrx1997:
Upgrade go version to 1.20
Upgrade github.com/prometheus/client_golang to v1.11.1
Upgrade k8s.io/client-go to v0.27.0

/assign @swetharepakula


swetharepakula:
/ok-to-test

Same comment as #2272. Can you break this into 3 commits to make it easier to review. Thanks!

songrx1997:
The lint check in test failed. We are going to bypass this check to update the version. 

Run several e2e tests locally as sanity check:

Screenshot 2023-09-28 at 11 53 22 PM

Also all other checks and tests passed in  `pull-ingress-gce-test`.

@bowei Could you review this?

swetharepakula:
/lgtm
/approve

@bowei, can you review and merge. We should bypass the golinter in this case since this is a cherry-pick to update just the go modules with no functionality change and the golinter is passing on master and our current release versions.

aojea:
the errors look legit to me, why can't we fix the errors?

aojea:
/hold

those errors does not seem right

aojea:
/hold cancel

Talked with @songrx1997 offline, verified that are only linter problem, the job logs are noisy and I was worried we may merge something that can cause problems later

bowei:
There seems to be some issues with the linter, merging.

PR #2280, opened 2023-10-03
Trim unneeded fields in informer cache

linxiulei:
Trim unneeded fields for memory efficiency. Lint check is added to prevent using those fields in the future changes.

Fixes #2279

PR #2281, opened 2023-10-03
Use latest golangci-lint

linxiulei:
https://github.com/kubernetes/ingress-gce/pull/2280 needs latest golangci-lint, which seems to work for tests only in a separate PR.

### DIFF
--- a/hack/verify-lint.sh
+++ b/hack/verify-lint.sh
@@ -29,7 +29,7 @@ export PATH=$GOBIN:$PATH
 # Install golangci-lint
 echo "Installing golangci-lint"
 echo
-go install github.com/golangci/golangci-lint/cmd/golangci-lint > /dev/null
+go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest > /dev/null
 cd "../.."
 
 export GOLANGCI_LINT_CACHE=$PWD/.cache


aojea:
/assign @swetharepakula 
/ok-to-test

I think this repo is under code freeze these days, Swetha will know better

swetharepakula:
Yes we are under a code freeze for the week.

/hold We are under code freeze until Friday Oct 6th

hack/verify-lint.sh
```diff
@@ -29,7 +29,7 @@ export PATH=$GOBIN:$PATH
 # Install golangci-lint
 echo "Installing golangci-lint"
 echo
-go install github.com/golangci/golangci-lint/cmd/golangci-lint > /dev/null
+go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest > /dev/null
```
swetharepakula, line 32: We should pin this to a specific version and it should match what we have in [tools go.mod](https://github.com/kubernetes/ingress-gce/blob/master/hack/tools/go.mod)

linxiulei: ah, actually you pointed out why pull-ingress-gce-test was failed for #2280. Changing go.mod is the correct way. We can discard this PR. Thanks!

PR #2284, opened 2023-10-12
Do not compare connection tracking setting in a Backend Service if st…

mmamczur:
…rong session affinity is not used.

We should not set connection tracking at all if strong session affinity is not used. Also don't compare these fields if SSA is not needed to avoid problem with GCE returning defaults we don't expect.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -288,7 +288,7 @@ func (b *Backends) DeleteSignedUrlKey(be *composite.BackendService, keyName stri
 }
 
 // EnsureL4BackendService creates or updates the backend service with the given name.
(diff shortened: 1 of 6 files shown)


aojea:
maybe too late, but I wonder if we should add feature flags for these news development to avoid being exercised by default, or are we considering this is feature complete?

aojea:
> maybe too late, but I wonder if we should add feature flags for these news development to avoid being exercised by default, or are we considering this is feature complete?

Ok, my bad, we have a flag, is just that the flag is not gating the new code, we should fix that, this comparions are being executed despite the flag is not set

mmamczur:
> > maybe too late, but I wonder if we should add feature flags for these news development to avoid being exercised by default, or are we considering this is feature complete?
> 
> Ok, my bad, we have a flag, is just that the flag is not gating the new code, we should fix that, this comparions are being executed despite the flag is not set

so this actually does introduce dependency on the flag.

There is a feature flag for SSA (currently disabled). The bool introduced here can only be true if the flag is true (and also the service must be SSA service for it to be true)

PR #2286, opened 2023-10-17
Remove unused old ingress-class flag

panslava:


pkg/utils/utils_test.go
```diff
@@ -718,28 +692,12 @@ func TestIsGCEIngress(t *testing.T) {
 			},
 			expected: false,
 		},
-		{
-			// Annotation supercedes spec.ingressClassName
-			desc: "Set by flag with matching class, and ingressClassName set",
```
swetharepakula, line 723: We should keep this test for the annotation superseding ingressClassName aspect. We can remove the ingressClassFlag piece.

swetharepakula (commented):
Overall LGTM. One comment about one of the tests.

pkg/utils/utils.go
```diff
@@ -424,9 +423,6 @@ func IGLinks(igs []*compute.InstanceGroup) (igLinks []string) {
 // controller.
 func IsGCEIngress(ing *networkingv1.Ingress) bool {
 	class := annotations.FromIngress(ing).IngressClass()
-	if flags.F.IngressClass != "" && class == flags.F.IngressClass {
```
aojea, line 427: Are you sure there is no ingress object around with this annotation?

panslava: which annotation?

pkg/utils/utils_test.go
panslava: thanks, done

pkg/utils/utils.go
aojea: > 			desc: "Annotation supersedes spec.ingressClassName",

panslava: Sorry, I don't think I understand the question completely

Before this pr, we had a flag, that enabled only single Ingress Class handling (so you pass in the flag --ingress-class="my-own-ingress" and now we would only handle "my-own-ingress" ingresses). But we are not using this flag and functionality at the moment, this flag is empty for us, and this should not change users behaviour (besides if someone self-hosted ingress-gce with this custom flag)

spencerhance: > (besides if someone self-hosted ingress-gce with this custom flag)

I think this is the issue.  It will make our lives easier to leave this around until we deprecate it separately.

PR #2290, opened 2023-10-18
When updating a BackendService copy the backends from the existing re…

mmamczur:
…source to avoid detachment.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -353,6 +353,8 @@ func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinit
 	klog.V(2).Infof("EnsureL4BackendService: updating backend service %v", name)
 	// Set fingerprint for optimistic locking
 	expectedBS.Fingerprint = bs.Fingerprint
+	// Copy backends to avoid detaching them during update. This could be replaced with a patch call in the future.
+	expectedBS.Backends = bs.Backends
 	if err := composite.UpdateBackendService(b.cloud, key, expectedBS); err != nil {
 		return nil, err
 	}
--- a/pkg/backends/backends_test.go
+++ b/pkg/backends/backends_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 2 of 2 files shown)


pkg/backends/backends.go
```diff
@@ -353,6 +353,8 @@ func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinit
 	klog.V(2).Infof("EnsureL4BackendService: updating backend service %v", name)
 	// Set fingerprint for optimistic locking
 	expectedBS.Fingerprint = bs.Fingerprint
+	// Copy backends to avoid detaching them during update. This could be replaced with a patch call in the future.
+	expectedBS.Backends = bs.Backends
```
aojea, line 357: why are these things not added on line 312 when we build the desired state?

adding it here gets the risk of introducing a new change in the middle that carries the same bug

aojea:
/lgtm
to the best of my knowledge this looks correct

/assign @bowei

pkg/backends/backends.go
bowei, line 357: Why do we do this after line 345 where we check if it's equal? Shouldn't we do it before the equality check?

pkg/backends/backends_test.go
```diff
@@ -111,6 +112,66 @@ func TestEnsureL4BackendService(t *testing.T) {
 	}
 }
 
+func TestEnsureL4BackendServiceDoesNotDetachBackends(t *testing.T) {
+	serviceName := "test-service"
```
bowei, line 116: use const for constants

pkg/backends/backends_test.go
```diff
@@ -111,6 +112,66 @@ func TestEnsureL4BackendService(t *testing.T) {
...
+	}
+	if diff := cmp.Diff(backend, existingBS.Backends[0]); diff != "" {
+		t.Errorf("BackendService.Backends were changed, expected no change: %s", diff)
+	}
+	description, err := utils.MakeL4LBServiceDescription(namespacedName.String(), "", meta.VersionGA, false, utils.ILB)
+	if err != nil {
+		t.Errorf("utils.MakeL4LBServiceDescription() failed %v", err)
+	}
+	if bs.Description != description {
+		t.Errorf("BackendService.Description was not populated correctly, want=%q, got=%q", description, bs.Description)
```
bowei, line 171: got before want

mmamczur:
check #2301 for a follow up to the comments from here

PR #2294, opened 2023-10-20
POC of L7 Regional External

panslava:


panslava:
This whole functionality should be in master now

PR #2295, opened 2023-10-23
Introduce panic recovery for metrics export

alexkats:


mmamczur:
looks ok I guess.

Have you tried to trigger a panic and see if it does not crash the whole binary?

pkg/metrics/metrics.go
```diff
@@ -286,6 +296,12 @@ func (im *ControllerMetrics) DeleteService(serviceKey string) {
 
 // export computes and exports ingress usage metrics.
 func (im *ControllerMetrics) export() {
+	defer func() {
+		if r := recover(); r != nil {
+			klog.Errorf("failed to export metrics: %v", r)
+			MetricExportFailureCount.WithLabelValues("main").Inc()
```
panslava, line 302: What if this triggers a panic? :D

alexkats: Well, then we have a problem with exporting metrics as a whole, and it's a more serious issue. The idea of recovery here is to avoid panic when we have issue while calculating metrics, but we still expect that the export itself is working fine.

alexkats:
> looks ok I guess.
> 
> Have you tried to trigger a panic and see if it does not crash the whole binary?

Yes, tested this change and recovery works as expected

pkg/metrics/metrics.go
```diff
@@ -286,6 +296,12 @@ func (im *ControllerMetrics) DeleteService(serviceKey string) {
 
 // export computes and exports ingress usage metrics.
 func (im *ControllerMetrics) export() {
+	defer func() {
+		if r := recover(); r != nil {
```
cezarygerard, line 300: this recover only cathes panics from gorutine that started it? 
In this case the goroutine started in `func (im *ControllerMetrics) Run(stopCh <-chan struct{})`  in this file

right?

alexkats: Basically yes, it catches panics from the goroutine where `recover` happens and all subgoroutines as well. So in this case it's only on `export` function with all the subfunctions

PR #2299, opened 2023-10-24
Add gce-regional ingress class annotation

panslava:
- Add "gce-regional" ingress class annotation
- Add GceL7XLBRegionalIngressClass that checks if ingress has
      gce-regional annotation class, and if regional external ingress is
      enabled
- Should be no-op


swetharepakula (commented):
/lgtm

PR #2300, opened 2023-10-24
Add L7XLBRegionalEnabled to translator and service port 

panslava:
Just passing flag for now, should be no-op changes

### DIFF
--- a/pkg/controller/translator/translator.go
+++ b/pkg/controller/translator/translator.go
@@ -57,7 +57,8 @@ const (
 
 // getServicePortParams allows for passing parameters to getServicePort()
 type getServicePortParams struct {
-	isL7ILB bool
+	isL7ILB         bool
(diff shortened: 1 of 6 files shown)


pkg/controller/translator/translator.go
```diff
@@ -313,6 +319,9 @@ func (t *Translator) TranslateIngress(ing *v1.Ingress, systemDefaultBackend util
 
 	params := &getServicePortParams{}
 	params.isL7ILB = utils.IsGCEL7ILBIngress(ing)
+	if t.enableL7XLBRegional {
```
swetharepakula, line 322: should there be a corresponding test change with this?

panslava: Sure, added a test for this function with gce-regional, also moved this part to separate func

PR #2301, opened 2023-10-25
Post review fixes to PR #2290.

mmamczur:


pkg/backends/backends_test.go
```diff
@@ -172,10 +170,10 @@ func TestEnsureL4BackendServiceDoesNotDetachBackends(t *testing.T) {
 	}
 }
 
-// TestBackendSvcEqual checks that backendSvcEqual() and
-// connectionTrackingPolicyEqual() (as a part ofit  backendSvcEqual)
+// TestBackendSvcL4RelevantPropertiesEqual checks that backendSvcL4RelevantPropertiesEqual() and
```
bowei, line 173: We should have unit test for (if it doesn't exist):

- Ignores Backends
- Any other attribute that is being ignored.

mmamczur: I added a test case with BackendService popualted with all the attributes the function doesn't care about.

pkg/backends/backends_test.go
```diff
@@ -323,12 +322,184 @@ func TestBackendSvcEqual(t *testing.T) {
...
+			},
+			wantEqual: true,
+		},
 	} {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
 			t.Parallel()
-			if res := backendSvcEqual(tc.oldBackendService, tc.newBackendService) == tc.wantEqual; !res {
-				t.Errorf("backendSvcEqual() returned %v, expected %v. Diff(oldScv, newSvc): %s",
+			if res := backendSvcL4RelevantPropertiesEqual(tc.oldBackendService, tc.newBackendService) == tc.wantEqual; !res {
```
cezarygerard, line 501: `if backendSvcL4RelevantPropertiesEqual(tc.oldBackendService, tc.newBackendService) != tc.wantEqual;`

it's simpler

cezarygerard (commented):
lgtm overall

PR #2304, opened 2023-10-26
Simplify if order, no logic change, no-op change

panslava:
/cc @bowei

### DIFF
--- a/pkg/healthchecks/healthchecks.go
+++ b/pkg/healthchecks/healthchecks.go
@@ -88,10 +88,10 @@ func generateClusterInfo(gceCloud *gce.Cloud) healthcheck.ClusterInfo {
 // new returns a *HealthCheck with default settings and specified port/protocol
 func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	var hc *translator.HealthCheck
-	if sp.NEGEnabled && !sp.L7ILBEnabled {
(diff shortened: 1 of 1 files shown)


pkg/healthchecks/healthchecks.go
```diff
@@ -88,10 +88,10 @@ func generateClusterInfo(gceCloud *gce.Cloud) healthcheck.ClusterInfo {
 // new returns a *HealthCheck with default settings and specified port/protocol
 func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	var hc *translator.HealthCheck
-	if sp.NEGEnabled && !sp.L7ILBEnabled {
-		hc = translator.DefaultNEGHealthCheck(sp.Protocol)
-	} else if sp.L7ILBEnabled {
+	if sp.L7ILBEnabled {
```
bowei, line 91: Since we are going to be adding more cases,

suggest

switch {
case sp.L7ILBEnabled:...
...
}

panslava: Being honest, personally I don't like switch statements, because I always confused do we need to put breaks (I just google, in golang you don't need), and somehow because they are rarely used, it makes it's harder for me to read

Also for me switch is more like simple 1:1 mapping (though it would still work for this case), but we want to return NEGHealthCheck only if it is not ILB, and this would be confusing to read and understand in switch statement, and could result in a bug if we change order in switch (which is a bit unexpected to me)

PR #2306, opened 2023-10-26
Add and pass enableL7XLBRegional to NEG controller

panslava:
- No-op change for now
- We will use it later to process negs for gce-regional ingresses

/cc @bowei

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -104,6 +104,10 @@ type Controller struct {
 	// runL4 indicates whether to run NEG controller that processes L4 services
 	runL4 bool
 
+	// enableIngressRegionalExternal indicates where NEG controller should process
+	// gce-regional-external ingresses
+	enableIngressRegionalExternal bool
+
 	logger klog.Logger
 }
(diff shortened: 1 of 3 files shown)


pkg/neg/controller.go
```diff
@@ -104,6 +104,10 @@ type Controller struct {
 	// runL4 indicates whether to run NEG controller that processes L4 services
 	runL4 bool
 
+	// enableL7XLBRegional indicates where NEG controller should process
+	// gce-regional ingresses
+	enableL7XLBRegional bool
```
spencerhance, line 109: Can we just use the same flag name everywhere?

The controller flag is `EnableIngressRegionalExternal`, but it gets renamed here when you are plumbing it through.  Might as well just use `EnableIngressRegionalExternal` here instead.

spencerhance (changes requested):

pkg/neg/controller.go
panslava: sure, renamed (but keeping private)

PR #2316, opened 2023-10-31
Move setting healthcheck ForNEG, ForILB to upper functions

panslava:
- This will reduce getRegional to do only 1 simple thing -- getting
  health check
- This will allow us to set ForRegionalXLB later
- Should not change any behaviour


mmamczur:
I'm not sure what is the reason to do this change but it looks ok

PR #2322, opened 2023-11-03
Fix vulnerability in release-1.20

sawsa307:
Fix vulnerability in release-1.20.
* Bump go version to 1.20.10.
* Bump golang.org/x/net version to v0.13.0.


bowei:
merging manually as it fails lint check

PR #2323, opened 2023-11-03
Fix vulnerability in release-1.23

sawsa307:
Fix vulnerability in release-1.23.
* Bump go version to 1.20.10.
* Bump golang.org/x/net version to v0.13.0.


go.mod
```diff
@@ -27,6 +27,8 @@ require (
 	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d
 )
 
+require k8s.io/cloud-provider-gcp/crd v0.0.0-20230411231432-6ab330834e4e
```
swetharepakula, line 30: why did this move out here?

swetharepakula (commented):
One question about the go.mod but otherwise LGTM

/ok-to-test

go.mod
sawsa307: That's some automatic formatting did by `go mod tidy`.

swetharepakula: thanks for clarifying!

PR #2326, opened 2023-11-04
Add IsL7XLBRegional to translator

panslava:
- For IsL7XLBRegional construct EXTERNAL_MANAGED forwarding rules
- Add utils.IsGCEL7XLBRegionalIngress for every place of NewTranslator

/assign @spencerhance

### DIFF
--- a/pkg/translator/translator.go
+++ b/pkg/translator/translator.go
@@ -80,13 +80,19 @@ func NewEnv(ing *v1.Ingress, client kubernetes.Interface, vip, net, subnet strin
 type Translator struct {
 	// IsL7ILB is true if the Ingress will be translated into an L7 ILB (as opposed to an XLB).
 	IsL7ILB bool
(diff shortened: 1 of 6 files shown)


pkg/translator/translator.go
```diff
@@ -282,6 +288,9 @@ func (t *Translator) ToCompositeForwardingRule(env *Env, protocol namer.NamerPro
 		} else {
 			fr.Subnetwork = env.Subnetwork
 		}
+	} else if t.IsL7XLBRegional {
+		fr.LoadBalancingScheme = "EXTERNAL_MANAGED"
+		fr.NetworkTier = cloud.NetworkTierStandard.ToGCEValue()
```
spencerhance, line 293: nit:  add comment here:

// Regional External LB only supports Standard Network Tier.

panslava: Thanks, done!

PR #2331, opened 2023-11-06
Add syncing default backends for L7 Regional XLB

panslava:
/assign @spencerhance


pkg/neg/controller_test.go
```diff
@@ -861,6 +861,106 @@ func TestDefaultBackendServicePortInfoMapForL7ILB(t *testing.T) {
 	}
 }
 
+func TestDefaultBackendServicePortInfoMapForL7RXLB(t *testing.T) {
```
panslava, line 864: this is almost complete copy of test TestDefaultBackendServicePortInfoMapForL7ILB

PR #2339, opened 2023-11-08
Add simple e2e creation/deletion test for gce-regional-external ingress

panslava:
- Extend e2e helpers to support verifications for external regional LBs
- New test is almost a full copy of ilb_test.go


panslava:

(Swetha not necessary, but maybe worth for the first test)

PR #2341, opened 2023-11-10
New flag for degraded mode metrics collection, and add logging for the selection of endpoint calculation.

sawsa307:
* To avoid changes in code logic, we don't handle degradedMode calculation error outside, but only collect metrics when degraddedModeErr is nil.
* Log if we are using degraded mode endpoint calculation.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -123,6 +123,7 @@ var (
 		EnableMultiNetworking                    bool
 		MaxIGSize                                int
 		EnableDegradedMode                       bool
+		EnableDegradedModeMetrics                bool
 		EnableDualStackNEG                       bool
 		EnableFirewallCR                         bool
 		DisableFWEnforcement                     bool
@@ -298,7 +299,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
 	flag.DurationVar(&F.NegMetricsExportInterval, "neg-metrics-export-interval", 5*time.Second, `Period for calculating and exporting internal neg controller metrics, not usage.`)
-	flag.BoolVar(&F.EnableDegradedMode, "enable-degraded-mode", false, `Enable endpoint calculation using degraded mode procedures`)
+	flag.BoolVar(&F.EnableDegradedMode, "enable-degraded-mode", false, `Enable degraded mode endpoint calculation and use results when error state is triggered. enabledDegradedMode also enables degrade mode correctness metrics with or without enabledDegradedModeMetrics.`)
+	flag.BoolVar(&F.EnableDegradedModeMetrics, "enable-degraded-mode-metrics", false, `Enable metrics collection for degraded mode, but uses normal mode calculation result when error state is triggered.`)
 	flag.BoolVar(&F.EnableNEGLabelPropagation, "enable-label-propagation", false, "Enable NEG endpoint label propagation")
 	flag.BoolVar(&F.EnableDualStackNEG, "enable-dual-stack-neg", false, `Enable support for Dual-Stack NEGs within the NEG Controller`)
 	flag.BoolVar(&F.EnableFirewallCR, "enable-firewall-cr", false, "Enable generating firewall CR")
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -296,9 +296,12 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		}
 	}
 
-	if !s.enableDegradedMode && err != nil {
-		return err
-	} else if s.enableDegradedMode {
+	if !s.enableDegradedMode {
+		if err != nil {
+			return err
+		}
+		s.logger.Info("Using normal mode endpoint calculation")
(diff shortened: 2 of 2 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -277,22 +277,26 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	endpointsData := negtypes.EndpointsDataFromEndpointSlices(endpointSlices)
 	targetMap, endpointPodMap, err = s.getEndpointsCalculation(endpointsData, currentMap)
 
+	var notInDegraded, onlyInDegraded map[string]negtypes.NetworkEndpointSet
```
swetharepakula, line 280: let's still flag gate this with --enabled-degraded-mode-metrics or something like that

pkg/neg/syncers/transaction.go
```diff
@@ -286,6 +286,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		computeDegradedModeCorrectness(notInDegraded, onlyInDegraded, string(s.NegSyncerKey.NegType), s.logger)
 	}
 
+	var useDegradedModeResult bool
```
swetharepakula, line 289: why do we need another variable. Shouldn't this just be the enableDegradedMode bool?

sawsa307: No it is not the same. 
If we are in error-state, useDegradedModeResult=enableDegradedMode
But if we are not in error-state, and when we enabledDegradedMode, we won't the results from degraded mode https://github.com/kubernetes/ingress-gce/blob/1f544e95e319be1d4280e47b71a585d74121e846/pkg/neg/syncers/transaction.go#L299-L302.

swetharepakula: Since this is purely to figure out what to output in the logs.

We should just log the decision. So instead of setting the bool, log that we are using a particular calculation where we use it. This eliminates the need for this variable

pkg/neg/syncers/transaction.go
```diff
@@ -277,22 +281,30 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	endpointsData := negtypes.EndpointsDataFromEndpointSlices(endpointSlices)
 	targetMap, endpointPodMap, err = s.getEndpointsCalculation(endpointsData, currentMap)
 
+	var degradedTargetMap, notInDegraded, onlyInDegraded map[string]negtypes.NetworkEndpointSet
+	var degradedPodMap negtypes.EndpointPodMap
+	var degradedModeErr error
+	if s.enableDegradedModeMetrics || s.enableDegradedMode {
```
swetharepakula, line 287: if we flag gate like this, we need to update the `enableDegradedMode` flag to include that enableDegradedMode also enables the metrics without or without the flag.

pkg/neg/syncers/transaction.go
```diff
@@ -314,6 +316,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			}
 		}
 	}
+	s.logger.Info("Selected endpoint calculation", "useDegradedModeResult", useDegradedModeResult)
```
swetharepakula: move the log into the different conditions that select the calculation and get rid of the useDegradedModeResult flag.

Since this flag is used only in logs, it is a little confusing and may be misleading for future developers.

swetharepakula: saw the next commit. Squash this and the next commit

pkg/neg/syncers/transaction.go
sawsa307: Updated. I think gated this way, we won't need to worry about forgetting to turn on both flags.

swetharepakula (commented):
a few nits, otherwise LGTM

pkg/neg/syncers/transaction.go
sawsa307: Oh sorry I missed this part. Squashed.

sawsa307:
> a few nits, otherwise LGTM

Updated. Please take another look!

pkg/flags/flags.go
```diff
@@ -298,7 +299,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.IntVar(&F.MaxIGSize, "max-ig-size", 1000, "Max number of instances in Instance Group")
 	flag.DurationVar(&F.MetricsExportInterval, "metrics-export-interval", 10*time.Minute, `Period for calculating and exporting metrics related to state of managed objects.`)
 	flag.DurationVar(&F.NegMetricsExportInterval, "neg-metrics-export-interval", 5*time.Second, `Period for calculating and exporting internal neg controller metrics, not usage.`)
-	flag.BoolVar(&F.EnableDegradedMode, "enable-degraded-mode", false, `Enable endpoint calculation using degraded mode procedures`)
+	flag.BoolVar(&F.EnableDegradedMode, "enable-degraded-mode", false, `Enable endpoint calculation using degraded mode, and use degraded mode result when error state is triggered. enableDegradedMode also enables the metrics without or without enableDegradedModeMetrics.`)
```
swetharepakula, line 302: Reword:
Enable endpoint calculation and use results when error state is triggered. enabledDegradedMode also enables degrade mode correctness metrics with or without enabledDegradedModeMetrics.

swetharepakula:
Can we squash the last two commits? Mostly because log line that is incorrect is fixed in the last commit, so might be easier to squash both.

sawsa307:
> Can we squash the last two commits? Mostly because log line that is incorrect is fixed in the last commit, so might be easier to squash both.

Done!

swetharepakula (commented):
/lgtm
/approve

PR #2356, opened 2023-11-15
Add TestRegionalXLBILBTransition e2e test

panslava:
- This test verifies transition between default backend, RXLB, ILB

This test is almost a full copy of TestILBUpdate

/assign @spencerhance

### DIFF
--- a/cmd/e2e-test/regional_xlb_test.go
+++ b/cmd/e2e-test/regional_xlb_test.go
@@ -316,3 +316,183 @@ func TestRXLBILBShared(t *testing.T) {
 		})
 	}
 }
+
+// Test Updating RXLB and transitioning between ILB/RXLB
+func TestRegionalXLBILBTransition(t *testing.T) {
+	t.Parallel()
+
+	// These names are useful when reading the debug logs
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/regional_xlb_test.go
```diff
@@ -126,3 +127,171 @@ func TestRegionalXLB(t *testing.T) {
...
+			}
+
+			// Verify everything works
+			ing, err = e2e.WaitForIngress(s, tc.ingUpdate, nil, nil)
+			if err != nil {
+				t.Fatalf("error waiting for Ingress to stabilize: %v", err)
+			}
+			t.Logf("GCLB resources created (%s/%s)", s.Namespace, tc.ingUpdate.Name)
+
+			// Perform whitebox testing.
```
spencerhance, line 278: Can we also verify the GC of the other LB here?

panslava: Added verification for GC of both LBs

spencerhance: Thanks!

PR #2358, opened 2023-11-16
Remove credential validation for IAP

swetharepakula:
IAP API allows users to configure IAP without providing credentials and use the default Google OAuth. In order to support in the ingress controller, the API validation needs to be loosened.

If customers try to switch to using default Oauth they will receive an error like the following:

```
 Warning  Sync    1s (x3 over 7s)  loadbalancer-controller  Error syncing to GCP: error running backend syncing routine: Errored updating IAP Settings for service default/whereami: Cannot switch to default OAuth once IAP credentials have been set
```


swetharepakula:
/hold

Found a bug. Will need to update the PR.

pkg/backends/features/iap.go
```diff
@@ -39,9 +39,10 @@ func EnsureIAP(sp utils.ServicePort, be *composite.BackendService) bool {
 	beTemp.Iap.Oauth2ClientSecretSha256 = fmt.Sprintf("%x", sha256.Sum256([]byte(beTemp.Iap.Oauth2ClientSecret)))
 	if be.Iap == nil || beTemp.Iap.Enabled != be.Iap.Enabled || beTemp.Iap.Oauth2ClientId != be.Iap.Oauth2ClientId || beTemp.Iap.Oauth2ClientSecretSha256 != be.Iap.Oauth2ClientSecretSha256 {
 		applyIAPSettings(sp, be)
-		klog.V(2).Infof("Updated IAP settings for service %v/%v.", sp.ID.Service.Namespace, sp.ID.Service.Name)
+		klog.Infof("Updated IAP settings for service %v/%v.", sp.ID.Service.Namespace, sp.ID.Service.Name)
```
bowei, line 42: %s/%s

Should we log some non-security related parts of the IAP settings.

Suggest:

```
  diff := diffIAP(be, beTemp)
  if diff == "" {
    klog.Infof("Detected no change ...")
    return false
  } else {
    ...
  }
}

func diffIAP(oldIAP, newIAP) string {
  if oldIAP == nil {
    return "nil policy to new policy"
 
  } 
  var diffs []string
  
  if oldIAP.Enabled != newIAP.Enabled {
    diffs = append(diffs, fmt.Sprintf("enabled (%t to %t)", oldIAP.Enabled, newIAP.Enabled)
  }
  ...
  return strings.Join(diffs, ", ")
}
```

swetharepakula: I have done something similar, please take a look.

swetharepakula:
/hold cancel

The IAP API does not support switching from set credentials to the default. Adding a commit to publish an error event in that case.

This PR is ready to be reviewed and merged.

pkg/backends/features/iap_test.go
```diff
@@ -165,15 +225,61 @@ func TestEnsureIAP(t *testing.T) {
...
+			expectErr:      true,
+		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			result := EnsureIAP(tc.sp, tc.be)
+			result, err := EnsureIAP(tc.sp, tc.be, klog.TODO())
+			if tc.expectErr && err == nil {
+				t.Errorf("%v: got nil but expected an error", tc.desc)
```
gauravkghildiyal, line 271: nit: Seems like we have been following a weird pattern. I guess since we already have the description in the test name (`t.Run()`), maybe we don't actually need it here (and this should rather be the name of the function under test.)

gauravkghildiyal (commented):
Thanks this is perfect. Just had that question about IAP field null check which we can discuss independently. Feel free to ignore the nitpicks.

/lgtm
/hold (Waiting for completion of our discussion)

pkg/backends/features/iap.go
```diff
@@ -18,31 +18,40 @@ package features
...
 	}
 	beTemp := &composite.BackendService{}
 	applyIAPSettings(sp, beTemp)
-	// We need to compare the SHA256 of the client secret instead of the client secret itself
-	// since that field is redacted when getting a BackendService.
-	beTemp.Iap.Oauth2ClientSecretSha256 = fmt.Sprintf("%x", sha256.Sum256([]byte(beTemp.Iap.Oauth2ClientSecret)))
-	if be.Iap == nil || beTemp.Iap.Enabled != be.Iap.Enabled || beTemp.Iap.Oauth2ClientId != be.Iap.Oauth2ClientId || beTemp.Iap.Oauth2ClientSecretSha256 != be.Iap.Oauth2ClientSecretSha256 {
+
+	if err := switchingToDefault(beTemp, be); err != nil {
+		logger.Error(err, "Errored updating IAP settings", "service namespace", sp.ID.Service.Namespace, "service name", sp.ID.Service.Name)
```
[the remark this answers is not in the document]
swetharepakula: I agree. thanks for catching!

pkg/backends/features/iap_test.go
swetharepakula: I agree. Here is what a failure looks like with the current code:

```
FAIL: TestEnsureIAP (0.00s)
    --- FAIL: TestEnsureIAP/credentials_removed,_update_will_error (0.00s)
        iap_test.go:271: credentials removed, update will error: got nil but expected an error
 ```
 
 Removing so it looks like this now:
 ```
 --- FAIL: TestEnsureIAP (0.00s)
    --- FAIL: TestEnsureIAP/credentials_removed,_update_will_error (0.00s)
        iap_test.go:271: EnsureIAP: got nil but expected an error
 ```

swetharepakula:
/hold cancel

Discussed nits and other questions offline. Nits have been addressed and new comments added to address the remaining questions.

gauravkghildiyal:
Thanks!

/lgtm
/unhold

PR #2359, opened 2023-11-16
Make PublishNegControllerErrorCountMetrics run in its own goroutine

sawsa307:
* Put PublishNegControllerErrorCountMetrics in its own goroutine because it is running on critial path in syncer.go.

### DIFF
--- a/pkg/neg/syncers/syncer.go
+++ b/pkg/neg/syncers/syncer.go
@@ -91,7 +91,7 @@ func (s *syncer) Start() error {
 			retryCh := make(<-chan time.Time)
 			err := s.core.sync()
 			if err != nil {
-				metrics.PublishNegControllerErrorCountMetrics(err, false)
+				go metrics.PublishNegControllerErrorCountMetrics(err, false)
 				delay, retryErr := time.Duration(0), error(nil)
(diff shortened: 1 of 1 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -325,11 +325,13 @@ func PublishDegradedModeCorrectnessMetrics(count int, endpointType string, negTy
 // PublishNegControllerErrorCountMetrics publishes collected metrics
 // for neg controller errors.
 func PublishNegControllerErrorCountMetrics(err error, isIgnored bool) {
-	if err == nil {
-		return
-	}
-	NegControllerErrorCount.WithLabelValues(totalNegError).Inc()
-	NegControllerErrorCount.WithLabelValues(getErrorLabel(err, isIgnored)).Inc()
+	go func() {
```
swetharepakula, line 328: instead of making a go func here, you should make the caller run this as its own go routine:
```
go metrics.PublishNegControllerErrorCountMetrics(err, false) 
```

https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/syncer.go#L94

swetharepakula (commented):
/lgtm

PR #2361, opened 2023-11-17
Add HTTPs redirects e2e test for RXLB Ingress

panslava:
- Add test implementation, almost a full copy of the TestHttpsRedirects
- Add to regional utils handling https redirect map, which was not expected for ILB (which was the only user of regional utils)

/assign @spencerhance

### DIFF
--- a/pkg/fuzz/gcp.go
+++ b/pkg/fuzz/gcp.go
@@ -426,10 +426,11 @@ func hasBetaResource(resourceType string, validators []FeatureValidator) bool {
 }
 
 type GCLBForVIPParams struct {
-	VIP        string
-	Region     string
-	Network    string
-	Validators []FeatureValidator
+	VIP              string
+	Region           string
+	Network          string
+	Validators       []FeatureValidator
+	IsForRegionalXLB bool
 }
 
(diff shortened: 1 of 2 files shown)


pkg/fuzz/gcp.go
```diff
@@ -793,6 +794,15 @@ func RegionalGCLBForVIP(ctx context.Context, c cloud.Cloud, gclb *GCLB, params *
 			if urlMapKey == nil {
 				urlMapKey = urlMapResID.Key
 			}
+			// Ignore redirect urlmaps since they will not have backends, but add them to the gclb map
```
spencerhance, line 797: Let's discuss this offline - I don't think we should be using the redirect map with regional XLB.

panslava: Redirect map will be created only if user specified HttpsRedirectConfig in FrontendConfig (to redirect Http to https)

spencerhance: Ah I thought RXLB didn't require this anymore, but I still see it being suggested: https://cloud.google.com/load-balancing/docs/https/setting-up-reg-http-https-redirect

nevermind

PR #2374, opened 2023-11-25
Create firewall if it doesn't exist during periodic resync.

sawsa307:
GetFirewall could fail for reasons other than notFound, and treating them all as notFound could lead to error during CreateFirewall(StatusConflict).

### DIFF
--- a/pkg/firewalls/fakes.go
+++ b/pkg/firewalls/fakes.go
@@ -30,6 +30,7 @@ type fakeFirewallsProvider struct {
 	networkURL       string
 	onXPN            bool
 	fwReadOnly       bool
+	getFirewallHook  func(name string) (*compute.Firewall, error)
 }
 
 // NewFakeFirewallsProvider creates a fake for firewall rules.
@@ -44,6 +45,10 @@ func NewFakeFirewallsProvider(onXPN bool, fwReadOnly bool) *fakeFirewallsProvide
 }
 
 func (ff *fakeFirewallsProvider) GetFirewall(name string) (*compute.Firewall, error) {
+	if ff.getFirewallHook != nil {
+		return ff.getFirewallHook(name)
+	}
+
 	rule, exists := ff.fw[name]
 	if exists {
 		return rule, nil
--- a/pkg/firewalls/firewalls.go
+++ b/pkg/firewalls/firewalls.go
@@ -69,16 +69,20 @@ func NewFirewallPool(cloud Firewall, namer *namer_util.Namer, l7SrcRanges []stri
 func (fr *FirewallRules) Sync(nodeNames, additionalPorts, additionalRanges []string, allowNodePort bool) error {
 	fr.logger.V(4).Info("Sync", "nodeNames", nodeNames)
 	name := fr.namer.FirewallRule()
-	existingFirewall, _ := fr.cloud.GetFirewall(name)
 
 	expectedFirewall, err := fr.buildExpectedFW(nodeNames, additionalPorts, additionalRanges, allowNodePort)
 	if err != nil {
 		return err
 	}
 
-	if existingFirewall == nil {
-		fr.logger.V(3).Info("Creating firewall rule", "firewallRuleName", name)
-		return fr.createFirewall(expectedFirewall)
+	existingFirewall, err := fr.cloud.GetFirewall(name)
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			fr.logger.V(3).Info("Firewall not found, creating firewall rule", "firewallRuleName", name)
+			return fr.createFirewall(expectedFirewall)
+		}
+		fr.logger.Error(err, "Failed to get firewall", "firewallRuleName", name)
+		return err
 	}
 
 	// Early return if an update is not required.
--- a/pkg/firewalls/firewalls_test.go
+++ b/pkg/firewalls/firewalls_test.go
@@ -18,16 +18,19 @@ package firewalls
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"k8s.io/klog/v2"
 	"strings"
 	"testing"
+
+	"k8s.io/klog/v2"
 
 	"google.golang.org/api/compute/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	firewallclient "k8s.io/cloud-provider-gcp/crd/client/gcpfirewall/clientset/versioned/fake"
 	"k8s.io/cloud-provider-gcp/providers/gce"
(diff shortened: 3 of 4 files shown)


pkg/firewalls/firewalls.go
```diff
@@ -66,16 +66,19 @@ func NewFirewallPool(cloud Firewall, namer *namer_util.Namer, l7SrcRanges []stri
...
-	if existingFirewall == nil {
-		klog.V(3).Infof("Creating firewall rule %q", name)
-		return fr.createFirewall(expectedFirewall)
+	existingFirewall, err := fr.cloud.GetFirewall(name)
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			klog.V(3).Infof("Creating firewall rule %q", name)
+			return fr.createFirewall(expectedFirewall)
+		}
+		return err
```
bowei, line 81: klog.Errorf("GetFirewall(%q) = %v", name, err)

pkg/firewalls/firewalls.go
```diff
@@ -66,16 +66,19 @@ func NewFirewallPool(cloud Firewall, namer *namer_util.Namer, l7SrcRanges []stri
...
 		return err
 	}
 
-	if existingFirewall == nil {
-		klog.V(3).Infof("Creating firewall rule %q", name)
-		return fr.createFirewall(expectedFirewall)
+	existingFirewall, err := fr.cloud.GetFirewall(name)
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			klog.V(3).Infof("Creating firewall rule %q", name)
```
bowei, line 78: Firewall not found, creating firewall %q

pkg/firewalls/firewalls.go
```diff
@@ -66,16 +66,19 @@ func NewFirewallPool(cloud Firewall, namer *namer_util.Namer, l7SrcRanges []stri
...
 
 	expectedFirewall, err := fr.buildExpectedFW(nodeNames, additionalPorts, additionalRanges, allowNodePort)
 	if err != nil {
 		return err
 	}
 
-	if existingFirewall == nil {
-		klog.V(3).Infof("Creating firewall rule %q", name)
-		return fr.createFirewall(expectedFirewall)
+	existingFirewall, err := fr.cloud.GetFirewall(name)
```
bowei, line 75: Is existingFirewall used anywhere?

bowei:
Is there some test for this change?

pkg/firewalls/firewalls.go
sawsa307: Yes it is used at line 85 when we compare existing and expected firewalls.

pkg/firewalls/firewalls.go
sawsa307: Added.

pkg/firewalls/firewalls.go
aojea: should you not keep the same order, first try to get the firwall and it does not existi create and if it does exist reconcile it?

sawsa307: Actually we are preserving the order here.
In the original code, though we called GetFirewall first, we did not check the result or handle the error after we buildExpectedFW. The first return is due to buildExpectedFW error in the original code, then we get check the result from Get and issue create or continue with reconciliation. So it makes no difference whether we do
```
existingFirewall, _ := fr.cloud.GetFirewall(name)
expectedFirewall, err := fr.buildExpectedFW(nodeNames, additionalPorts, additionalRanges, allowNodePort)
if err != nil {
	return err
}
if existingFirewall == nil {
	klog.V(3).Infof("Creating firewall rule %q", name)
	return fr.createFirewall(expectedFirewall)
}
...
```
or 
```
expectedFirewall, err := fr.buildExpectedFW(nodeNames, additionalPorts, additionalRanges, allowNodePort)
if err != nil {
	return err
}
existingFirewall, _ := fr.cloud.GetFirewall(name)
if existingFirewall == nil {
	klog.V(3).Infof("Creating firewall rule %q", name)
	return fr.createFirewall(expectedFirewall)
}
...
```

pkg/firewalls/firewalls.go
```diff
@@ -66,16 +66,20 @@ func NewFirewallPool(cloud Firewall, namer *namer_util.Namer, l7SrcRanges []stri
...
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			klog.V(3).Infof("Firewall not found, creating firewall %q", name)
+			return fr.createFirewall(expectedFirewall)
```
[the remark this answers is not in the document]
sawsa307: Previously we also created firewall, but the issue is we rely on the check existingFirewall==nil, and it can cause issue because we if existingFirewall==nil, it could due to
1. the firewall does not exist
2. Other reasons like server error etc. 

And we should only create the firewall if it doesn't exist. 
The reason for this change is we have a customer observed statusConflict error during createFirewall, and it is caused by GetFirewall returning an error that is not StatusNotFound

sawsa307:
> Is there some test for this change?

Added a test to mock behavior of GetFirewall returning ServerError instead of StatusNotFound

pkg/firewalls/fakes.go
```diff
@@ -30,25 +30,30 @@ type fakeFirewallsProvider struct {
...
+	return ff
 }
 
 func (ff *fakeFirewallsProvider) GetFirewall(name string) (*compute.Firewall, error) {
-	rule, exists := ff.fw[name]
-	if exists {
-		return rule, nil
-	}
-	return nil, test.FakeGoogleAPINotFoundErr()
+	return ff.GetFirewallHook(name)
```
swetharepakula, line 56: Instead of mocking out this function as the next person using this mock may not realize that they need to define a hook to get a mocked Firewall provider, perhaps we introduce an override concept? I am a little wary to completely add a hook especially if the existing test code was built on the assumption that this mock doesn't use hooks

You can define a hook, and the function will use the hook if it exists, otherwise it will do what its existing behavior is. This also gives a way to clear the hook if needed and go back to original behavior

sawsa307: Got it, it makes sense.
Updated, PTAL!

pkg/firewalls/fakes.go
```diff
@@ -30,6 +30,7 @@ type fakeFirewallsProvider struct {
 	networkURL       string
 	onXPN            bool
 	fwReadOnly       bool
+	GetFirewallHook  func(name string) (*compute.Firewall, error)
```
swetharepakula, line 33: nit, this should be private

sawsa307: yes thanks for catching this!

swetharepakula (commented):
/lgtm
/approve

PR #2378, opened 2023-11-29
Minor fixes and updates to logging related to multinet

mmamczur:
Fix some missing values

also add information about the network used when creating NEG

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -352,12 +352,12 @@ func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinit
 
 	// Create backend service if none was found
 	if bs == nil {
-		klog.V(2).Infof("EnsureL4BackendService(%v, _, %v, _, %v, %v ...): creating backend service %v", name, protocol, scheme, namespacedName)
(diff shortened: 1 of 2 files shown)


pkg/backends/backends.go
```diff
@@ -352,35 +352,35 @@ func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinit
 
 	// Create backend service if none was found
 	if bs == nil {
-		klog.V(2).Infof("EnsureL4BackendService(%v, _, %v, _, %v, %v ...): creating backend service %v", name, protocol, scheme, namespacedName)
+		klog.V(2).Infof("EnsureL4BackendService(%v, _, %v, _, %v, %v ...): creating backend service %v", name, protocol, scheme, namespacedName, namespacedName.String())
```
cezarygerard, line 355: klog.V(2).Infof("EnsureL4BackendService(%v, _, %v, _, %v, %v ...): creating backend service", name, protocol, scheme, namespacedName)

originally backend name was the only logged var, now it's a first arg

[same applies to the rest of changes in this file]

mmamczur: ouh, I somehow this was supposed to be the service name. changed

PR #2383, opened 2023-11-30
Add a refactored ZoneGetter package and filter nodes with providerID

songrx1997:
/assign @swetharepakula @bowei

### DIFF
--- a/pkg/utils/zonegetter/zone_getter_test.go
+++ b/pkg/utils/zonegetter/zone_getter_test.go
@@ -1,0 +1,348 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package zonegetter
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	informerv1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+func TestList(t *testing.T) {
+	zoneGetter := NewFakeZoneGetter()
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ReadyNodeWithProviderID",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project/us-central1-a/bar-node",
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionTrue,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "UnReadyNodeWithProviderID",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project/us-central1-b/bar-node",
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionFalse,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ReadyNodeWithoutProviderID",
+		},
+		Spec: apiv1.NodeSpec{},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionTrue,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "UnReadyNodeWithoutProviderID",
+		},
+		Spec: apiv1.NodeSpec{},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionFalse,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ReadyNodeInvalidProviderID",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://us-central1-c/bar-node",
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionTrue,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "UpgradingNodeWithProviderID",
+			Labels: map[string]string{
+				"operation.gke.io/type": "drain",
+			},
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project/us-central1-f/bar-node",
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionFalse,
+				},
+			},
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ReadyNodeWithEmptyZone",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project//bar-node",
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionTrue,
+				},
+			},
+		},
+	})
+
+	for _, tc := range []struct {
+		desc      string
+		filter    Filter
+		expectLen int
+	}{
+		{
+			desc:      "List with AllNodesFilter",
+			filter:    AllNodesFilter,
+			expectLen: 3,
+		},
+		{
+			desc:      "List with CandidateNodesFilter",
+			filter:    CandidateNodesFilter,
+			expectLen: 1,
+		},
+		{
+			desc:      "List with CandidateAndUnreadyNodesFilter",
+			filter:    CandidateAndUnreadyNodesFilter,
+			expectLen: 2,
+		},
+	} {
+		zones, _ := zoneGetter.List(tc.filter, klog.TODO())
+		if len(zones) != tc.expectLen {
+			t.Errorf("For test case %q, got %d zones, want %d,", tc.desc, len(zones), tc.expectLen)
+		}
+		for _, zone := range zones {
+			if zone == "" {
+				t.Errorf("For test case %q, got an empty zone,", tc.desc)
+			}
+		}
+	}
+}
+
+func TestZoneForNode(t *testing.T) {
+	zoneGetter := NewFakeZoneGetter()
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "NodeWithValidProviderID",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project/us-central1-a/bar-node",
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "NodeWithInvalidProviderID",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://us-central1-a/bar-node",
+		},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "NodeWithNoProviderID",
+		},
+		Spec: apiv1.NodeSpec{},
+	})
+	zoneGetter.nodeInformer.GetIndexer().Add(&apiv1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "NodeWithEmptyZone",
+		},
+		Spec: apiv1.NodeSpec{
+			ProviderID: "gce://foo-project//bar-node",
+		},
+	})
+
+	for _, tc := range []struct {
+		desc       string
+		nodeName   string
+		expectZone string
+		expectErr  error
+	}{
+		{
+			desc:       "Node not found",
+			nodeName:   "fooNode",
+			expectZone: "",
+			expectErr:  ErrNodeNotFound,
+		},
+		{
+			desc:       "Node with valid provider ID",
+			nodeName:   "NodeWithValidProviderID",
+			expectZone: "us-central1-a",
+			expectErr:  nil,
+		},
+		{
+			desc:       "Node with invalid provider ID",
+			nodeName:   "NodeWithInvalidProviderID",
+			expectZone: "",
+			expectErr:  ErrSplitProviderID,
+		},
+		{
+			desc:       "Node with no provider ID",
+			nodeName:   "NodeWithNoProviderID",
+			expectZone: "",
+			expectErr:  ErrProviderIDNotFound,
+		},
+		{
+			desc:       "Node with empty zone in providerID",
+			nodeName:   "NodeWithEmptyZone",
+			expectZone: "",
+			expectErr:  ErrSplitProviderID,
+		},
+	} {
+		zone, err := zoneGetter.ZoneForNode(tc.nodeName, klog.TODO())
+		if zone != tc.expectZone {
+			t.Errorf("For test case %q, got zone: %s, want: %s,", tc.desc, zone, tc.expectZone)
+		}
+		if !errors.Is(err, tc.expectErr) {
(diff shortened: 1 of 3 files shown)


pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+// ZoneLister is a interface to manages lookups for GCE instances to zones.
+type ZoneLister interface {
+	ListZones(mode Mode) ([]string, error)
+	GetZoneForNode(name string) (string, error)
+}
+
+// ZoneGetter implements ZoneLister Interface and manages lookups for
+// GCE instances to zones.
+type ZoneGetter struct {
+	NodeInformer cache.SharedIndexInformer
```
bowei, line 54: this should be package private

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// GetZoneForNode returns the zone for a given node by looking up providerID.
+func (z *ZoneGetter) GetZoneForNode(name string) (string, error) {
```
bowei, line 58: Get is redundant

ZoneForNode

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+	zone, err := getZone(node)
+	if err != nil {
+		return "", err
+	}
+	return zone, nil
+}
+
+// ListZones returns a list of zones containing nodes that satisfy the given
+// node filtering mode.
+func (z *ZoneGetter) ListZones(mode Mode) ([]string, error) {
```
bowei, line 76: ListZones => List

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return filtered, nil
+}
+
+func getZone(node *api_v1.Node) (string, error) {
+	zone, err := getZoneByProviderID(node.Spec.ProviderID)
+	if err != nil {
+		klog.Errorf("Failed to get zone information from node %q: %v", node.Name, err)
+		return "", err
+	}
```
bowei, line 125: if zone == "" { return "", fmt.Errorf("node %s has an empty zone", node.Name) }

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+	return zone, nil
+}
+
+// getZoneByProviderID gets zone information from node provider id.
+// A providerID is build out of '${ProviderName}://${project-id}/${zone}/${instance-name}'
+func getZoneByProviderID(providerID string) (string, error) {
+	var providerIDRE = regexp.MustCompile(`^` + "gce" + `://([^/]+)/([^/]+)/([^/]+)$`)
+	matches := providerIDRE.FindStringSubmatch(providerID)
+	if len(matches) != 4 {
+		return "", fmt.Errorf("%w: providerID %q is not in valid format", ErrSplitProviderID, providerID)
```
bowei, line 135: providerID %q is not valid

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+		klog.Errorf("Failed to get zone information from node %q: %v", node.Name, err)
+		return "", err
+	}
+	return zone, nil
+}
+
+// getZoneByProviderID gets zone information from node provider id.
+// A providerID is build out of '${ProviderName}://${project-id}/${zone}/${instance-name}'
+func getZoneByProviderID(providerID string) (string, error) {
+	var providerIDRE = regexp.MustCompile(`^` + "gce" + `://([^/]+)/([^/]+)/([^/]+)$`)
```
bowei, line 132: you should move this outside at the package level rather than MustCompile everytime this func is called

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+		if predicate(nodes[i]) && nodes[i].Spec.ProviderID != "" {
+			filtered = append(filtered, nodes[i])
+		}
+	}
+
+	return filtered, nil
+}
+
+func getZone(node *api_v1.Node) (string, error) {
+	zone, err := getZoneByProviderID(node.Spec.ProviderID)
```
bowei, line 121: seems a little redundant to have getZone and getZoneByProviderID, can we just merge

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+func listNodesWithMode(nodeLister listers.NodeLister, mode Mode) ([]*api_v1.Node, error) {
+	var predicate utils.NodeConditionPredicate
+	switch mode {
+	case AllNodesMode:
+		predicate = utils.AllNodesPredicate
+	case CandidateNodesMode:
+		predicate = utils.CandidateNodesPredicate
+	case CandidateNodeIncludeUnreadyExcludeUpgradingMode:
+		predicate = utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
+	}
```
bowei, line 102: default:
  return nil, <err>

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,138 @@
...
+
+	nodes, err := nodeLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	var filtered []*api_v1.Node
+	for i := range nodes {
+		// Filter nodes with predicate and exclude nodes without providerID
+		if predicate(nodes[i]) && nodes[i].Spec.ProviderID != "" {
```
bowei, line 112: We should put the state machine in a doc.go or markdown file in this package describing when provider ID will be ""

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter_test.go
```diff
@@ -0,0 +1,245 @@
...
+			mode:      AllNodesMode,
+			expectLen: 2,
+		},
+		{
+			desc:      "List with CandidateNodesPredicate",
+			mode:      CandidateNodesMode,
+			expectLen: 1,
+		},
+		{
+			desc:      "List with CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes",
```
swetharepakula, line 121: should we add nodes that are upgrading?

pkg/zonegetter/zone_getter_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestList(t *testing.T) {
```
swetharepakula, line 27: can we add a test case that shows that we handle empty zones properly in list?

pkg/zonegetter/node-state.md
```diff
@@ -0,0 +1,5 @@
+| Node State     | ProviderID | Consider it valid? |
```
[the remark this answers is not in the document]
songrx1997: Done

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// ZoneGetter implements ZoneLister Interface and manages lookups for
+// GCE instances to zones.
+type ZoneGetter struct {
+	nodeInformer cache.SharedIndexInformer
+}
+
+// ZoneForNode returns the zone for a given node by looking up providerID.
+func (z *ZoneGetter) ZoneForNode(name string) (string, error) {
```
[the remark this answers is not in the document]
songrx1997: Done

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+	return zones.List(), nil
+}
+
+// ListWithPredicate gets nodes that matches node filter mode.
+func listNodesWithMode(nodeLister listers.NodeLister, mode Mode) ([]*api_v1.Node, error) {
+	var filtered []*api_v1.Node
+
+	var predicate utils.NodeConditionPredicate
```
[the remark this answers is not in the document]
songrx1997: Done

pkg/zonegetter/zone_getter_test.go
songrx1997: Done

pkg/zonegetter/zone_getter_test.go
songrx1997: Done

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+var ErrProviderIDNotFound = errors.New("providerID not found")
+var ErrSplitProviderID = errors.New("error splitting providerID")
+var ErrNodeNotFound = errors.New("node not found")
+
+// providerIDRE is the regex to process providerID.
+var providerIDRE = regexp.MustCompile(`^` + "gce" + `://([^/]+)/([^/]+)/([^/]+)$`)
+
+// ZoneLister is a interface to manages lookups for GCE instances to zones.
```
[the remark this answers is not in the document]
songrx1997: We also have a lot of tests depend on the [FakeZoneGetter](https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/types/fakes.go#L52). If we remove the interface, we will need to implement the FakeZoneGetter in a different way. What do you think?

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,190 @@
...
+
+// NewZoneGetter initialize a ZoneGetter in normal mode.
+func NewZoneGetter(nodeInformer cache.SharedIndexInformer) *ZoneGetter {
+	return &ZoneGetter{
+		mode:         Normal,
+		nodeInformer: nodeInformer,
+	}
+}
+
+// NewZoneGetter initialize a fake ZoneGetter in normal mode.
```
swetharepakula, line 186: move this into the test file

pkg/zonegetter/zone_getter_test.go
```diff
@@ -0,0 +1,333 @@
...
+		},
+		{
+			desc:      "List with CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes",
+			filter:    CandidateNodeIncludeUnreadyExcludeUpgradingFilter,
+			expectLen: 2,
+		},
+	} {
+		zones, _ := zoneGetter.List(tc.filter, klog.TODO())
+		if len(zones) != tc.expectLen {
+			t.Errorf("For test case %q, got %d zones, want %d,", tc.desc, len(zones), tc.expectLen)
```
swetharepakula, line 166: can we add an expectation that no empty zones are in the list

pkg/zonegetter/zone_getter_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"errors"
+	"reflect"
+	"testing"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+func TestList(t *testing.T) {
```
swetharepakula, line 29: we need tests for the simple zone getter mode as well

swetharepakula: thought i deleted the comment. There is a different simple zone getter test which covers it.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,190 @@
...
+		predicate = utils.CandidateNodesPredicate
+	case CandidateNodeIncludeUnreadyExcludeUpgradingFilter:
+		predicate = utils.CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes
+	default:
+		return filtered, nil
+	}
+
+	nodes, err := nodeLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
songrx1997: The err will be logged in the caller List(). Since this function is a internal helper function, it will be duplicate if we log it twice.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,190 @@
...
+
+type Mode int
+type Filter string
+
+const (
+	Normal Mode = iota
+	Simple
+)
+
+const (
```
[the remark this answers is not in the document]
songrx1997: If we make these enums, it will be unclear which filter we are using in the logs.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,190 @@
...
+	AllNodesFilter                                    = "AllNodesFilter"
+	CandidateNodesFilter                              = "CandidateNodesFilter"
+	CandidateNodeIncludeUnreadyExcludeUpgradingFilter = "CandidateNodeIncludeUnreadyExcludeUpgradingFilter"
+)
+
+var ErrProviderIDNotFound = errors.New("providerID not found")
+var ErrSplitProviderID = errors.New("error splitting providerID")
+var ErrNodeNotFound = errors.New("node not found")
```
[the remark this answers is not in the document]
songrx1997: My plan is handle this outside like this https://github.com/kubernetes/ingress-gce/commit/5bf9c82d0f6900ccf38a3835219ef685daa39a9a#diff-f64364884edd3d19cda733b2ca7bda358bfba8fca128782330fe53d1fde6f08aL371-R385

We can classify errors for metrics outside with our current metrics.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,190 @@
...
+	nodes, err := nodeLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range nodes {
+		// Filter nodes with predicate and exclude nodes without providerID
+		if predicate(nodes[i]) {
+			filtered = append(filtered, nodes[i])
+		}
```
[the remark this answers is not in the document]
songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/zonegetter/zone_getter_test.go
songrx1997: Done

pkg/zonegetter/node-state.md
```diff
@@ -0,0 +1,5 @@
+| Node State     | ProviderID | Controller Processes/Uses the Node |
```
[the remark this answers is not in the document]
songrx1997: Done

pkg/zonegetter/zone_getter.go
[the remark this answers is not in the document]
bowei: you can always add the String() method for logging purposes, but usually it's ok to keep as string const to skip the step

```
type enumType int

func (e enumType) String() string {
  switch e {
  case ...: return "enumType1"
  }
  return "invalid enumType"
}

```

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,184 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+type Mode int
+type Filter string
+
+const (
+	Normal Mode = iota
```
bowei, line 38: This needs docstring to explain what these mean

bowei: Also, we should probably change the names as Normal and Simple are almost meaningless.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,184 @@
...
+	Simple
+)
+
+const (
+	AllNodesFilter                                    = "AllNodesFilter"
+	CandidateNodesFilter                              = "CandidateNodesFilter"
+	CandidateNodeIncludeUnreadyExcludeUpgradingFilter = "CandidateNodeIncludeUnreadyExcludeUpgradingFilter"
```
bowei, line 45: is there not a better name?

pkg/zonegetter/zone_getter.go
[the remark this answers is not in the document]
swetharepakula: for now, since there is only one caller, that's okay. in the future it might not be so. But we can add it then

pkg/zonegetter/zone_getter.go
swetharepakula: Perhaps SingleZone, or since there is only one usecase for this (Non-GCP). We could have a Non-GCP and GCP modes?

songrx1997: Done

pkg/zonegetter/zone_getter.go
songrx1997: Renamed it since `candidateNodes` has already excluded upgrading nodes.

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+
+	for _, node := range nodes {
+		// Filter nodes with predicate and exclude nodes without providerID
+		if predicate(node) {
+			filtered = append(filtered, node)
+		} else if len(nodes) <= 50 {
+			logger.Info("Filtered out node when listing node zones", "node", node.Name, "filter", filter)
+		}
```
swetharepakula, line 151: perhaps we do a single line after the loop instead?

pkg/zonegetter/zone_getter.go
```diff
@@ -0,0 +1,187 @@
...
+	nodeLister := z.nodeInformer.GetIndexer()
+	zones := sets.String{}
+	nodes, err := listNodesWithFilter(listers.NewNodeLister(nodeLister), filter, logger)
+	if err != nil {
+		logger.Error(err, "Failed to list nodes")
+		return zones.List(), err
+	}
+	for _, n := range nodes {
+		zone, err := getZone(n)
+		if err != nil {
```
swetharepakula, line 115: should this check for the empty zone error?

swetharepakula: read the code wrong, ignore the statement

swetharepakula (commented):
can you move the zone getter package to the pkg/utils.

We should reduce the number of high level packages.

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -0,0 +1,348 @@
...
+			filter:    AllNodesFilter,
+			expectLen: 3,
+		},
+		{
+			desc:      "List with CandidateNodesPredicate",
+			filter:    CandidateNodesFilter,
+			expectLen: 1,
+		},
+		{
+			desc:      "List with CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes",
```
swetharepakula, line 163: nit: update the predicate with the new name

pkg/zonegetter/zone_getter.go
songrx1997: Done

pkg/utils/zonegetter/zone_getter_test.go
songrx1997: Done

swetharepakula (commented):
/lgtm
/approve

PR #2387, opened 2023-12-07
Update cloud-provider-gcp, k8s-cloud-provider. Fix broken build from packages update

panslava:
7 commits (all have more detailed description in commit message):

- Update cloud-provider-gcp to v0.28.2, as dependency updates k8s-cloud-provider to v1.25.0

- Update composite types ./hack/update-codegen.sh

- Change ConfigMapsLeasesResourceLock to LeasesResourceLock

- Change mock hook signatures to have ...options in the end

- Add sprintf to properly print with new klog

- Fix functionality for new k8s-cloud-provider ResourceID.Equal 

/assign @bowei @swetharepakula


panslava:
Partial resource links are no longer equal to the full one, because APIGroup is empty for partial 
image

pkg/loadbalancers/forwarding_rules_test.go
```diff
@@ -319,8 +319,8 @@ func TestForwardingRulesEqual(t *testing.T) {
 				IPProtocol:          "TCP",
 				LoadBalancingScheme: string(cloud.SchemeInternal),
 				BackendService:      "http://www.googleapis.com/projects/test/regions/us-central1/backendServices/bs1",
-				Network:             "projects/test-poject/global/networks/test-vpc",
-				Subnetwork:          "projects/test-poject/regions/us-central1/subnetworks/default-subnet",
+				Network:             "https://www.googleapis.com/compute/v1/projects/test-poject/global/networks/test-vpc",
```
swetharepakula, line 322: I thought by converting to ParseResourceURL and only validating the fields we care about, we no longer were dependent on this being full or partial.

Can we fix how we equate so we don't introduce a dependency on the full URL

pkg/utils/utils_test.go
```diff
@@ -175,7 +175,7 @@ func TestEqualResourceIDs(t *testing.T) {
 		"partial vs full": {
 			a:    "https://www.googleapis.com/compute/beta/projects/project-id/zones/us-central1-a/instanceGroups/example-group",
 			b:    "projects/project-id/zones/us-central1-a/instanceGroups/example-group",
-			want: true,
+			want: false,
```
swetharepakula, line 178: Do we have an idea of which controllers may have been dependent on this behavior?

swetharepakula (commented):
I think we need to be a little careful about the ResourceID.

Do we know that none of the controllers are assuming that partial and full URLs are equated to be the same. Can we investigate and see if we can keep that the same. We have had issues with full URLs, because GCE has changed the domain and versioning a few times. Based on these unit tests, it looks we are still able to handle those cases, but I am wondering if we are introducing some new bug with this change.

panslava:
@swetharepakula sure, I agree we don't want to break anyone. I added a new function utils.EqualCloudResourceIDs which behaves as the old .Equal from k8s-cloud-provider. I also replaced every usage of .Equal to this new function

Ready for review

PR #2391, opened 2023-12-12
Extend metrics with Regional XLB Ingress feature

panslava:
/assign @spencerhance


spencerhance:
@panslava it looks like some unit tests are failing: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/ingress-gce/2391/pull-ingress-gce-test/1734658738456891392

panslava:
@spencerhance oops, my miss, fixed

PR #2400, opened 2023-12-18
Change the subsystem name for the rate limiter metrics

alexkats:


swetharepakula:
/ok-to-test
/hold

There are downstream systems that are dependent on this. We may need to duplicate the metrics before we can completely remove the old subsystem name.

PR #2401, opened 2023-12-18
Change instanceGroup manager logic to use zone information from GCE when removing nodes

songrx1997:
/assign @swetharepakula

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -187,27 +187,31 @@ func (m *manager) DeleteInstanceGroup(name string) error {
 }
 
 // listIGInstances lists all instances of provided instance group name in all zones.
-func (m *manager) listIGInstances(name string) (sets.String, error) {
+// The return format will be a set of nodes in the instance group and
+// a map from node name to zone.
+func (m *manager) listIGInstances(name string) (sets.String, map[string]string, error) {
 	nodeNames := sets.NewString()
+	nodeZoneMap := make(map[string]string)
 	zones, err := m.ListZones(utils.AllNodesPredicate)
 	if err != nil {
-		return nodeNames, err
+		return nodeNames, nodeZoneMap, err
 	}
 
 	for _, zone := range zones {
 		instances, err := m.cloud.ListInstancesInInstanceGroup(name, zone, allInstances)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/instancegroups/manager.go
```diff
@@ -297,10 +298,27 @@ func (m *manager) add(groupName string, names []string) error {
 }
 
 // Remove removes the given instances from the appropriately zoned Instance Group.
-func (m *manager) remove(groupName string, names []string) error {
+func (m *manager) remove(groupName string, names []string, nameZoneMap map[string]string) error {
 	events.GlobalEventf(m.recorder, core.EventTypeNormal, events.RemoveNodes, "Removing %s from InstanceGroup %q", events.TruncatedStringList(names), groupName)
 	var errs []error
-	for zone, nodeNames := range m.splitNodesByZone(names) {
```
swetharepakula, line 303: can we use a single loop to do everything?

You could still use the zone, nodeNames list and just add all the node names into the map as the first step?

swetharepakula: misread the diff. This commend can be ignored.

pkg/instancegroups/manager.go
```diff
@@ -187,27 +187,28 @@ func (m *manager) DeleteInstanceGroup(name string) error {
 }
 
 // listIGInstances lists all instances of provided instance group name in all zones.
-func (m *manager) listIGInstances(name string) (sets.String, error) {
-	nodeNames := sets.NewString()
+// The return format will be a map from node name to zone.
+func (m *manager) listIGInstances(name string) (map[string]string, error) {
```
swetharepakula, line 191: can you return the string set here?

pkg/instancegroups/manager.go
```diff
@@ -297,10 +298,27 @@ func (m *manager) add(groupName string, names []string) error {
 }
 
 // Remove removes the given instances from the appropriately zoned Instance Group.
-func (m *manager) remove(groupName string, names []string) error {
+func (m *manager) remove(groupName string, names []string, nameZoneMap map[string]string) error {
```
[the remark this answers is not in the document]
songrx1997: Done

swetharepakula (commented):
/lgtm
/approve

PR #2402, opened 2023-12-19
Add some validations for L4 multinetwork services

mmamczur:
- only services in network type L3 are supported.
- multinet services with externalTrafficPolicy other than Local are not supported for now.

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -61,9 +61,7 @@ func NewNetworksResolver(networkLister, gkeNetworkParamSetLister cache.Indexer, 
 	}
 }
 
-// ServiceNetwork determines the network data to be used for the LB resources.
-// This function currently returns only the default network but will provide
-// secondary networks information for multi-networked services in the future.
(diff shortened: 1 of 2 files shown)


pkg/network/network.go
```diff
@@ -73,6 +71,12 @@ func (nr *NetworksResolver) ServiceNetwork(service *apiv1.Service) (*NetworkInfo
 	if !ok || networkName == "" || networkName == networkv1.DefaultPodNetworkName {
 		return DefaultNetwork(nr.cloudProvider), nil
 	}
+
+	// TODO: remove this check once DPv2 supports externalTrafficPolicy=Cluster services.
```
alexkats, line 75: Shouldn't we use cilium here instead of DPv2?

alexkats: It'd good to point to some issue in cilium, if there's any

mmamczur: this is a google specific feature of multi networking, not available in vanilla cilium.

alexkats: right, sorry for the nit picking

PR #2406, opened 2023-12-21
Replicate dependencies for neg usage metrics migration.

sawsa307:
* NEG usage metrics will be pulled from pkg/metrics to pkg/neg/metrics, and emit via NEG metrics collector.
* In this PR, only dependencies are replicated. No behavior change is introduced.
* Existing types and constants will be cleaned up in #2407.

/assign @swetharepakula

### DIFF
--- a/pkg/neg/metrics/metricscollector/features.go
+++ b/pkg/neg/metrics/metricscollector/features.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metricscollector
+
+type feature string
+
+func (f feature) String() string {
+	return string(f)
+}
+
+const (
+	neg            = feature("NEG")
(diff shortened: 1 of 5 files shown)


pkg/neg/metrics/metricscollector/types.go
```diff
@@ -0,0 +1,55 @@
...
+// that is used for computing neg usage metrics.
+type NegMetricsCollector interface {
+	// SetNegService adds/updates neg state for given service key.
+	SetNegService(svcKey string, negState NegServiceState)
+	// DeleteNegService removes the given service key.
+	DeleteNegService(svcKey string)
+}
+
+// NegServiceState contains all the neg usage associated with one service
+type NegServiceState struct {
```
swetharepakula, line 29: let's move the types to metrics.go where the other types are.

OR 

We should move the existing types here as a followup commit

pkg/neg/metrics/metricscollector/metrics_collector.go
```diff
@@ -48,55 +48,6 @@ func RegisterMetrics() {
 	})
 }
 
-type SyncerMetricsCollector interface {
```
swetharepakula, line 51: leave the SyncerMetricsCollector interface and the SyncerMetrics object here as that is core to this file.

pkg/neg/metrics/metricscollector/features.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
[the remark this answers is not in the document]
sawsa307: Updated!

pkg/neg/metrics/metricscollector/metrics_collector.go
sawsa307: Updated!

swetharepakula (commented):
/lgtm
/approve

PR #2407, opened 2023-12-21
Update NEG usage metrics emit.

sawsa307:
* Update NEG usage metrics so it is now emitted via NEG metrics collector.

/assign @swetharepakula


pkg/metrics/features.go
```diff
@@ -101,18 +101,6 @@ const (
 	sslPolicy      = feature("SSLPolicy")
```
swetharepakula, line 101: line 81, the feature "NEG" does that need to be removed?

sawsa307: No we need that for servicePort metrics at line 286

swetharepakula: or is this only used for Ingress metrics?

sawsa307: It is used for both Ingress and NEG metrics.
In ingress, we used it as the label for features on a service port.
In NEG, we used it as the label for neg usage metrics.

swetharepakula: clarified offline. NEG is used by multiple metrics, so we will duplicate this one.

swetharepakula (commented):
/lgtm
/approve

sawsa307:
image

Local run result to verify metrics are transported correctly.

PR #2412, opened 2023-12-21
Introduce enable-ig-controller flag, and flag-gate IG controller.

sawsa307:
* The IG controller will run if enable-ig-controller is enabled.
* Setting its default value as true so we won't break any existing tests.
* Flag-gate IG controller with enable-ig-controller.

/assign @swetharepakula

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -88,6 +88,7 @@ var (
 		RunIngressController             bool
 		RunL4Controller                  bool
 		RunL4NetLBController             bool
+		EnableIGController               bool
 		Version                          bool
 		WatchNamespace                   string
 		LeaderElection                   LeaderElectionConfiguration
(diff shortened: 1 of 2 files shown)


pkg/flags/flags.go
```diff
@@ -276,6 +277,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.RunIngressController, "run-ingress-controller", true, `Optional, whether or not to run IngressController as part of glbc. If set to false, ingress resources will not be processed. Only the L4 Service controller will be run, if that flag is set to true.`)
 	flag.BoolVar(&F.RunL4Controller, "run-l4-controller", false, `Optional, whether or not to run L4 Service Controller as part of glbc. If set to true, services of Type:LoadBalancer with Internal annotation will be processed by this controller.`)
 	flag.BoolVar(&F.RunL4NetLBController, "run-l4-netlb-controller", false, `Optional, f enabled then the L4NetLbController will be run.`)
+	flag.BoolVar(&F.EnableIGController, "enable-ig-controller", true, `Optional, f enabled then the IG controller will be run.`)
```
swetharepakula, line 280: typo if

sawsa307: Updated. PTAL!

swetharepakula (commented):
/lgtm
/approve

PR #2413, opened 2023-12-22
Flag-gate the ingress and firewall controllers.

sawsa307:
* Flag-gate the ingress controller and the firewall controller using run-ingress-controller flag.
* This flag's default value is true, and we have always turned on this flag, so the flag-gate should be a no-op and not breaking any existing tests or features.
* Update the flag usage message.

/assign @swetharepakula

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -294,21 +294,29 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	stopCh := make(chan struct{})
 	exitCh := make(chan error)
 	ctx.Init()
-	lbc := controller.NewLoadBalancerController(ctx, stopCh, exitCh)
-	if ctx.EnableASMConfigMap {
-		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
-			// We want to trigger a restart, don't have to clean up all the resources.
-			if err := lbc.Stop(false); err != nil {
-				klog.Errorf("Failed to stop the load balancer controller: %v", err)
-			}
-		})
-	}
+	go app.RunSIGTERMHandler(stopCh, exitCh)
 
-	if !flags.F.EnableFirewallCR && flags.F.DisableFWEnforcement {
-		klog.Fatalf("We can only disable the ingress controller FW enforcement when enabling the FW CR")
-	}
+	var lbc *controller.LoadBalancerController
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -327,7 +327,7 @@ func (lbc *LoadBalancerController) Run() {
 		lbc.exitCh <- lbc.Stop(flags.F.DeleteAllOnQuit)
 	}()
 	klog.Infof("Starting loadbalancer controller")
-	lbc.ingQueue.Run()
+	go lbc.ingQueue.Run()
```
swetharepakula, line 330: should this change be part of SIGTERM Handler change?

cmd/glbc/main.go
```diff
@@ -294,21 +294,29 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
-	lbc := controller.NewLoadBalancerController(ctx, stopCh, exitCh)
-	if ctx.EnableASMConfigMap {
-		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
-			// We want to trigger a restart, don't have to clean up all the resources.
-			if err := lbc.Stop(false); err != nil {
-				klog.Errorf("Failed to stop the load balancer controller: %v", err)
-			}
-		})
-	}
+	go app.RunSIGTERMHandler(stopCh, exitCh)
```
swetharepakula, line 297: if ingress controller is not running, the exitCh will never close/get an error so the process will be stuck forever

pkg/controller/controller.go
sawsa307: Yes that makes sense. I'll update this in the other PR.

cmd/glbc/main.go
sawsa307: I was thinking to have exitCh pass to ingress controller when running in one pod. When running with two pods, exitCh will be used for ingress controller and neg controller respectively. But I think a catch-all exitCh makes more sense. I'll update the approach correspondingly.

sawsa307: Discussed offline, RunSIGTERMHandler will still be responsible to catch TERM signal, but it would only use exit channel if ingress controller is enabled.

sawsa307:
Combined the change to #2414.

PR #2414, opened 2023-12-22
Pull out SIGTERMHandler, flag-gate ingress and firewall controller, and run L7 NEG as default

sawsa307:
* Pull out SIGTERM Handler so it is independent from the ingress controller and only closes closeCh when TERM signal is received.
* Run ingress controller in its own goroutine.
* Add a waitgroup to ensure cleanups are properly executed after close.
* Exit when cleanups are finished or 30 seconds timeout is reached. Instead of waiting on cleanup indefenitely, this avoids possible hanging of the program.
* Flag-gate ingress and firewall controller using runIngressController flag.

/assign @swetharepakula

### DIFF
--- a/cmd/glbc/app/handlers.go
+++ b/cmd/glbc/app/handlers.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -45,7 +46,7 @@ func RunHTTPServer(healthChecker func() context.HealthCheckResults) {
 	klog.Fatal(http.ListenAndServe(fmt.Sprintf(":%v", flags.F.HealthzPort), nil))
 }
 
-func RunSIGTERMHandler(closeStopCh func()) {
+func RunSIGTERMHandler(closeStopCh func(), wg *sync.WaitGroup) {
 	// Multiple SIGTERMs will get dropped
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, syscall.SIGTERM)
@@ -54,6 +55,10 @@ func RunSIGTERMHandler(closeStopCh func()) {
 	klog.Infof("Received SIGTERM, shutting down")
 
 	closeStopCh()
+	// Wait until all controllers are done with cleanup.
+	klog.Info("Starting to wait for controller cleanup")
+	wg.Wait()
+	klog.Info("Finished waiting for controller cleanup")
 	os.Exit(0)
 }
 
--- a/cmd/glbc/app/handlers.go
+++ b/cmd/glbc/app/handlers.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/klog/v2"
 
 	"k8s.io/ingress-gce/pkg/context"
-	"k8s.io/ingress-gce/pkg/controller"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/version"
 )
@@ -46,7 +45,7 @@ func RunHTTPServer(healthChecker func() context.HealthCheckResults) {
 	klog.Fatal(http.ListenAndServe(fmt.Sprintf(":%v", flags.F.HealthzPort), nil))
 }
 
-func RunSIGTERMHandler(lbc *controller.LoadBalancerController) {
+func RunSIGTERMHandler(closeStopCh func()) {
 	// Multiple SIGTERMs will get dropped
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, syscall.SIGTERM)
@@ -54,7 +53,7 @@ func RunSIGTERMHandler(lbc *controller.LoadBalancerController) {
 	<-signalChan
 	klog.Infof("Received SIGTERM, shutting down")
 
-	lbc.Stop()
+	closeStopCh()
 	os.Exit(0)
 }
 
--- a/cmd/glbc/app/handlers.go
+++ b/cmd/glbc/app/handlers.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"os/signal"
 	"strings"
-	"sync"
 	"syscall"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -46,7 +45,7 @@ func RunHTTPServer(healthChecker func() context.HealthCheckResults) {
 	klog.Fatal(http.ListenAndServe(fmt.Sprintf(":%v", flags.F.HealthzPort), nil))
 }
 
-func RunSIGTERMHandler(closeStopCh func(), wg *sync.WaitGroup) {
+func RunSIGTERMHandler(closeStopCh func()) {
 	// Multiple SIGTERMs will get dropped
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, syscall.SIGTERM)
@@ -55,11 +54,6 @@ func RunSIGTERMHandler(closeStopCh func(), wg *sync.WaitGroup) {
 	klog.Infof("Received SIGTERM, shutting down")
 
 	closeStopCh()
-	// Wait until all controllers are done with cleanup.
-	klog.Info("Starting to wait for controller cleanup")
-	wg.Wait()
-	klog.Info("Finished waiting for controller cleanup")
-	os.Exit(0)
 }
 
 func healthCheckHandler(checker func() context.HealthCheckResults) http.HandlerFunc {
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"sync"
 	"time"
 
 	flag "github.com/spf13/pflag"
@@ -295,12 +296,21 @@ func makeLeaderElectionConfig(ctx *ingctx.ControllerContext, client clientset.In
 
 func runControllers(ctx *ingctx.ControllerContext) {
 	stopCh := make(chan struct{})
+	var once sync.Once
+	// This ensures that stopCh is only closed once.
+	// Right now, we have two callers.
+	// One is triggered when the ASM configmap changes, and the other one is
+	// triggered by the SIGTERM handler.
+	closeStopCh := func() {
+		once.Do(func() { close(stopCh) })
+	}
+
 	ctx.Init()
 	lbc := controller.NewLoadBalancerController(ctx, stopCh, klog.TODO())
 	if ctx.EnableASMConfigMap {
 		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
 			// We want to trigger a restart.
-			lbc.Stop()
+			closeStopCh()
 		})
 	}
 
@@ -331,7 +341,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	if flags.F.EnableNEGController {
 		runNEGController(ctx, stopCh)
 	}
-	go app.RunSIGTERMHandler(lbc)
+	go app.RunSIGTERMHandler(closeStopCh)
 
 	go fwc.Run()
 	klog.V(0).Infof("firewall controller started")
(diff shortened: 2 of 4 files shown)


cmd/glbc/main.go
```diff
@@ -292,8 +292,9 @@ func makeLeaderElectionConfig(ctx *ingctx.ControllerContext, client clientset.In
 
 func runControllers(ctx *ingctx.ControllerContext) {
 	stopCh := make(chan struct{})
+	exitCh := make(chan error)
 	ctx.Init()
-	lbc := controller.NewLoadBalancerController(ctx, stopCh)
+	lbc := controller.NewLoadBalancerController(ctx, stopCh, exitCh)
```
swetharepakula, line 297: is the load balancer controller the only controller that has cleanup?

cmd/glbc/app/handlers.go
```diff
@@ -55,8 +54,10 @@ func RunSIGTERMHandler(lbc *controller.LoadBalancerController, deleteAll bool) {
 	klog.Infof("Received SIGTERM, shutting down")
 
 	// TODO: Better retries than relying on restartPolicy.
+	close(stopCh)
+
 	exitCode := 0
-	if err := lbc.Stop(deleteAll); err != nil {
+	if err := <-exitCh; err != nil {
```
swetharepakula, line 60: should we instead have a catch all that calculates the status code from the cleanup of all the controllers that are running?

The reason being, that exitCh is still linked to the ingress controller, so if the ingress controller isn't running what would the exit code be?

swetharepakula: Should this be a catch all stop function that is generated from runControllers which is based on all the controllers that are run?

cmd/glbc/main.go
```diff
@@ -295,21 +295,28 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
+	if flags.F.RunIngressController {
+		lbc = controller.NewLoadBalancerController(ctx, stopCh, exitCh, klog.TODO())
+		if ctx.EnableASMConfigMap {
+			ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
+				// We want to trigger a restart, don't have to clean up all the resources.
+				if err := lbc.Stop(false); err != nil {
+					klog.Errorf("Failed to stop the load balancer controller: %v", err)
+				}
+			})
+		}
```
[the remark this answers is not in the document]
sawsa307: Updated!

swetharepakula (commented):
Posting my comments, but based on offline discussion with David, the best approach might be to remove the DeleteAllOnQuit flag and feature.

This is a feature that was made for dev purposes only and is not being used. Supporting this adds complexity to the code base and increases surface area for potential bugs unnecessarily. 

Instead we will remove this feature which should remove this unnecessary complexity from the code.

pkg/neg/controller.go
```diff
@@ -234,52 +233,49 @@ func NewController(
...
-				pod := obj.(*apiv1.Pod)
-				negController.reflector.SyncPod(pod)
-			},
-			UpdateFunc: func(old, cur interface{}) {
-				pod := cur.(*apiv1.Pod)
-				negController.reflector.SyncPod(pod)
-			},
-		})
-	}
+	ingressInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
sawsa307: Created as a separate commit.

cmd/glbc/main.go
```diff
@@ -297,22 +297,30 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
 	if ctx.EnableASMConfigMap {
 		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
 			// We want to trigger a restart, don't have to clean up all the resources.
-			if err := lbc.Stop(false); err != nil {
-				klog.Errorf("Failed to stop the load balancer controller: %v", err)
+			close(stopCh)
+			if flags.F.RunIngressController {
+				if err := <-exitCh; err != nil {
+					klog.Errorf("Failed to stop the load balancer controller: %v", err)
+				}
```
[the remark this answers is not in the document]
sawsa307: We need exitCh since in lbc.Run() we registered `lbc.exitCh <-lbc.Run()`. Without this, lbc would stuck on the exitCh.

However, based on the newer change, this no longer applies.

cmd/glbc/app/handlers.go
```diff
@@ -57,9 +57,22 @@ func RunSIGTERMHandler(stopCh chan struct{}, exitCh <-chan error) {
...
+	// ingress controller's stop() and etermine the exit code.
+	// For other controllers, no errors are returned from stop/cleanup
+	// functions, and they do not contribute to the determination of the exit
+	// code.
+	// Here we flag-gate exitCh since the ingress controller is the only
+	// controller that uses this channel to communciate its exit error.
+	//
+	// TODO(cheungdavid): Add a waitGroup here to make sure graceful exits for
+	// other controllers.
+	if flags.F.RunIngressController {
```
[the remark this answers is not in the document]
sawsa307: No longer applies.

cmd/glbc/main.go
```diff
@@ -300,7 +300,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	if ctx.EnableASMConfigMap {
 		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
 			// We want to trigger a restart, don't have to clean up all the resources.
-			lbc.Stop()
+			close(stopCh)
```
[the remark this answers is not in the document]
sawsa307: Actually we should closeCh because that is what we do in the code currently.
Right now, calling lbc.Stop() close stopCh and run cleanup for ingress. It does not do anything related to signalChan.

cmd/glbc/main.go
```diff
@@ -329,9 +332,15 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	}
 
 	if flags.F.EnableNEGController {
-		runNEGController(ctx, stopCh)
+		wg.Add(1)
```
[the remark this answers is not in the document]
sawsa307: We didn't wait for all controllers previously, but it makes sense to wait for them all.

cmd/glbc/main.go
```diff
@@ -300,9 +300,7 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	if ctx.EnableASMConfigMap {
 		ctx.ASMConfigController.RegisterInformer(ctx.ConfigMapInformer, func() {
 			// We want to trigger a restart, don't have to clean up all the resources.
-			if err := lbc.Stop(false); err != nil {
-				klog.Errorf("Failed to stop the load balancer controller: %v", err)
-			}
+			lbc.Stop()
```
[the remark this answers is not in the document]
sawsa307: Updated.

sawsa307:
/hold
Wait for #2450

cmd/glbc/main.go
```diff
@@ -299,22 +299,26 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	ctx.Init()
 
 	var wg sync.WaitGroup
-	wg.Add(1)
-	lbc := controller.NewLoadBalancerController(ctx, stopCh, klog.TODO())
+	var lbc *controller.LoadBalancerController
+	var fwc *firewalls.FirewallController
```
swetharepakula, line 303: Since nothing else is using these variables, combine the creation of the controllers to running them so we don't need to have the variables like this.

If something is using these controllers, we should cleanup and ensure that they are not used by anything else

cmd/glbc/main.go
```diff
@@ -299,22 +299,26 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	ctx.Init()
```
swetharepakula, line 299: update the commit message to remove the line about the waitgroup as it is added in the previous commit.

cmd/glbc/main.go
```diff
@@ -345,18 +356,21 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
 		l4netlbController := l4lb.NewL4NetLBController(ctx, stopCh)
 
+		go runWithWgDone(l4netlbController.Run, &wg)
 		klog.V(0).Infof("L4NetLB controller started")
-		go l4netlbController.Run()
 	}
-	go lbc.Run()
+
+	go runWithWgDone(lbc.Run, &wg)
 	// Keep the program running until TERM signal.
```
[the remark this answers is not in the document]
sawsa307: No longer applies as we will use select channel

cmd/glbc/main.go
```diff
@@ -424,11 +438,17 @@ func runNEGController(ctx *ingctx.ControllerContext, stopCh chan struct{}) {
...
 
 	ctx.AddHealthCheck("neg-controller", negController.IsHealthy)
+	return negController
+}
 
-	go negController.Run(stopCh)
-	klog.V(0).Infof("negController started")
+// runWithWgDone is a convenience wrapper that runs the given function
+// with a deferred wg.Done()
+func runWithWgDone(f func(), wg *sync.WaitGroup) {
```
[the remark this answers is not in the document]
sawsa307: Make sense, this would greatly reduce the complexity.

But i think we also need to make IG controller and NetLB controller before SIGTERM Handler. Otherwise, we may encounter situation where we do wg.Add(1) with counter equals 0 after wg.Wait(), which is specifically mentioned to avoid in https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/sync/waitgroup.go;l=34.

cmd/glbc/main.go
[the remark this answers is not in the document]
sawsa307: In addition to that, since the two closing logic and independent of each other(one depends on configmap change and the other depends on TERM signal), we could possibly encounter the situation of "closing a closed channel".

Thus, I wrapped close(stopCh) with sync.Once to make sure we will only close it once. sync.Once uses sync.mutex mechanism and it also makes code simpler then managing locks ourselves.

sawsa307:
Create #2453 and #2454 to split the PR.

cmd/glbc/main.go
```diff
@@ -374,12 +374,25 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
-		if ctx.EnableASMConfigMap {
-			return
-		}
+	klog.Warning("Handled quit, awaiting pod deletion.")
+
+	doneCh := make(chan struct{})
+	go func() {
+		defer close(doneCh)
+		// Wait until related controllers are done with cleanup.
+		// If Ingress controller is not running, the program will exit immediately
```
swetharepakula, line 383: does this comment still apply?

sawsa307: No it should not, I'll remove this.

swetharepakula:
LGTM, just needs to be rebased with the other PRs before this can merge.

cmd/glbc/main.go
```diff
@@ -373,12 +373,25 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	go runWithWg(lbc.Run, &wg)
 	// Keep the program running until TERM signal.
 	<-stopCh
-	for {
-		klog.Warning("Handled quit, awaiting pod deletion.")
-		time.Sleep(30 * time.Second)
-		if ctx.EnableASMConfigMap {
-			return
-		}
+	klog.Warning("Handled quit, awaiting pod deletion.")
```
swetharepakula, line 376: change to say shutdown has been triggered.

cmd/glbc/main.go
```diff
@@ -373,12 +373,25 @@ func runControllers(ctx *ingctx.ControllerContext) {
...
+
+	doneCh := make(chan struct{})
+	go func() {
+		defer close(doneCh)
+		// Wait until related controllers are done with cleanup.
+		// If Ingress controller is not running, the program will exit immediately
+		// after stopCh is closed without waitGroup. It allows controllers to finish
+		// their cleanups.
+		wg.Wait()
```
swetharepakula, line 385: add logs before and after the wait to indicate started to wait for cleanup

swetharepakula (commented):
couple nits on logging, otherwise LGTM.

Thanks!

cmd/glbc/main.go
sawsa307: Updated.

cmd/glbc/main.go
sawsa307: Added.

sawsa307:
> couple nits on logging, otherwise LGTM.
> 
> Thanks!

Comments are addressed. PTAL!

swetharepakula (commented):
/lgtm
/approve

PR #2420, opened 2024-01-03
Revert "[Cherry pick #2359 -> 1.26] Make PublishNegControllerErrorCountMetrics run in its own goroutine"

swetharepakula:
Reverts kubernetes/ingress-gce#2403

Temporarily reverting the bug fixes on 1.26

PR #2423, opened 2024-01-03
Revert "Revert "[Cherry pick #2359 -> 1.26] Make PublishNegControllerErrorCountMetrics run in its own goroutine""

swetharepakula:
Reverts kubernetes/ingress-gce#2420

Since 1.26.2 was cut without any bug fixes, we can now undo the temporary reverts.

/assign @gauravkghildiyal


swetharepakula:
Error is not reproducible locally

/retest

PR #2425, opened 2024-01-03
Introduce new ZoneGetter and refactor related packages and tests

songrx1997:
/assign @swetharepakula


pkg/utils/zonegetter/fake.go
```diff
@@ -0,0 +1,252 @@
+/*
```
swetharepakula, line 1: can you move this into its own commit?

pkg/utils/zonegetter/fake.go
```diff
@@ -0,0 +1,252 @@
...
+				},
+			},
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func InitializeFakeZoneGetter(zoneGetter *ZoneGetter) {
```
swetharepakula, line 80: I wonder instead of hardcoding this, we should have a utils function that returns a list of nodes and this would sync the node informer in the fake to that?

Or perhaps we should make it so that this uses the same node informer that already exists, so it should mean as long as the node informer is maintained in the test, we don't need to do anything for the fake zone getter.

Wdyt?

pkg/utils/zonegetter/fake.go
songrx1997: Done

pkg/utils/zonegetter/fake.go
```diff
@@ -0,0 +1,252 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	informerv1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/ingress-gce/pkg/utils"
+)
+
+// NewFakeZoneGetter initialize a fake ZoneGetter in normal mode.
+func NewFakeZoneGetter() *ZoneGetter {
```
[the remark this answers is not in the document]
songrx1997: Removed this since the `NewZoneGetter` did this and can be used for tests with a fake nodeInformer.
Added another function `FakeNodeInformer` to create a fake nodeInformer for tests.

pkg/utils/zonegetter/fake.go
songrx1997: Change this to take a nodeInformer and add nodes to it.

pkg/utils/zonegetter/fake.go
```diff
@@ -0,0 +1,278 @@
...
+		if nodeZone == zone {
+			if err := nodeIndexer.Delete(node); err != nil {
+				t.Errorf("Failed to delete node %q in zone %q, err - %v", node.Name, zone, err)
+			}
+		}
+	}
+}
+
+// Fake add nodes adds fake nodes to the ZoneGetter in the provided zone.
+func FakeAddNodes(zoneGetter *ZoneGetter, newZone string, instances ...string) error {
```
swetharepakula, line 55: maybe AddFakeNodes

pkg/neg/manager_test.go
```diff
@@ -1417,6 +1421,11 @@ func TestSyncNodesConditions(t *testing.T) {
 		addZones      map[string][]string
 		deleteZones   []string
 	}{
+		{
+			desc:       "vm ip port neg, zones are the same",
+			expectSync: false,
+			negType:    negtypes.VmIpPortEndpointType,
+		},
```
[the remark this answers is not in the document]
songrx1997: Maybe this got moved when debugging the tests, reverted the move.

pkg/utils/zonegetter/fake.go
songrx1997: Done

pkg/utils/zonegetter/fake.go
```diff
@@ -0,0 +1,278 @@
...
+				},
+			},
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func FakeInitializeNodeInformer(nodeInformer cache.SharedIndexInformer) {
```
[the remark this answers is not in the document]
songrx1997: Changed name a added the comment. The nodes added to the informer is only for NEG tests. For the tests in other packages, only fake informer is required.

PR #2431, opened 2024-01-12
Fix NEG degraded mode for L4 externalTrafficPolicy=Local services.

mmamczur:
Whenever the NEG controller would go into the degraded mode for L4 LB service with externalTrafficPolicy=Local, the controller would ignore endpoint slice data for NEG endpoints calculation. This would result in detachment of all endpoints and broken LB.


sawsa307:
Looks good to me. Thanks for catching this!

PR #2432, opened 2024-01-15
IG Manager: return the instance group name once even if it existis in…

mmamczur:
… many nodes.

The manager sync function assumes the names of the groups are listed without a zone and will attempt to do the add/remove operations in all zones. Before this change, if the group existed in more than one zone it would be returned many times (list will contain the same name a few times). This leads to unnecessary iterations and instances listing - everything would happen N^2 for N=number of zones.

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -675,7 +675,7 @@ func TestMCIngressIG(t *testing.T) {
 	if err != nil {
 		t.Errorf("lbc.instancePool.List() = _, %v, want nil", err)
 	}
-	var wantInstanceGroups []string
+	wantInstanceGroups := []string{}
 	if diff := cmp.Diff(wantInstanceGroups, instanceGroups); diff != "" {
 		t.Errorf("lbc.instancePool.List()() mismatch (-want +got):\n%s", diff)
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/controller_test.go
```diff
@@ -675,7 +675,7 @@ func TestMCIngressIG(t *testing.T) {
 	if err != nil {
 		t.Errorf("lbc.instancePool.List() = _, %v, want nil", err)
 	}
-	var wantInstanceGroups []string
+	wantInstanceGroups := []string{}
```
code-elinka, line 678: why did you change it?

pkg/instancegroups/manager.go
```diff
@@ -226,6 +226,8 @@ func (m *manager) Get(name, zone string) (*compute.InstanceGroup, error) {
 }
 
 // List lists the names of all Instance Groups belonging to this cluster.
+// It will return only the names of the groups without the zone. If w group with the same name exists in more than one
```
code-elinka, line 229:
```suggestion
// It will return only the names of the groups without the zone. If a group with the same name exists in more than one
```

code-elinka:
Looks good

code-elinka:
nit: it is better to have a short commit message header 
Screenshot from 2024-01-16 10-04-40
Screenshot from 2024-01-16 10-04-34

see guidelines for commit messages: https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

mmamczur:
> nit: it is better to have a short commit message header

I changed it (but only in the commit message and not PR title)

PR #2433, opened 2024-01-16
Add support for sslPolicy for regional XLB ingress

panslava:
- Regional HTTPS Proxy requires patch method, instead of setSslPolicy

/assign @spencerhance

### DIFF
--- a/pkg/loadbalancers/target_proxies.go
+++ b/pkg/loadbalancers/target_proxies.go
@@ -17,13 +17,16 @@ limitations under the License.
 package loadbalancers
 
 import (
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/cloud-provider-gcp/providers/gce"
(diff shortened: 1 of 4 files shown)


pkg/loadbalancers/target_proxies.go
```diff
@@ -220,10 +225,31 @@ func (l7 *L7) ensureSslPolicy(env *translator.Env, currentProxy *composite.Targe
...
+// method.
+func ensureRegionalSslPolicy(cloud cloud.Cloud, key *meta.Key, currentProxy *composite.TargetHttpsProxy, policyLink string) error {
+	patchProxy := &computebeta.TargetHttpsProxy{
+		Fingerprint: currentProxy.Fingerprint,
+		SslPolicy:   policyLink,
+	}
+	if policyLink == "" {
+		patchProxy.NullFields = []string{"SslPolicy"}
+	}
+	return cloud.BetaRegionTargetHttpsProxies().Patch(context.Background(), key, patchProxy)
```
spencerhance, line 254: Can we not use the composite library here?  (I recognize that this might require updating k8s-cloud-provider)

PR #2435, opened 2024-01-17
[Cherry pick #2431 -> 1.27] Fix NEG degraded mode for L4 externalTrafficPolicy=Local services.

mmamczur:
Whenever the NEG controller would go into the degraded mode for L4 LB service with externalTrafficPolicy=Local, the controller would ignore endpoint slice data for NEG endpoints calculation. This would result in detachment of all endpoints and broken LB.

(cherry picked from commit 6ed11af3d7e2388449fac50bfc649c8354cecd28)


swetharepakula (commented):
/lgtm
/approve

PR #2430, opened 2024-01-12
Fix log params for new klog

panslava:
/assign @gauravkghildiyal @swetharepakula

PR #2335, opened 2023-11-07
Add contextual logging for Backend pools. 

sawsa307:
* BackendSyncer works like the manager for backends, while backend.go
  contains functions working on individual backends.
* We will do context embedding in backendSyncer when we process each
  backend.
Exmaple:
```
I0131 20:02:35.890028 1904521 syncer.go:62] Sync: backend {ID:default/neg-demo-svc/&ServiceB(14 results) [2237/2826]
80,} NodePort:0 Port:80 PortName:http Protocol:HTTP TargetPort:{Type:0 IntVal:9376 StrVal:} NEGEnabled:true VMIPNEGE
nabled:false L4RBSEnabled:false L7ILBEnabled:false L7XLBRegionalEnabled:false THCConfiguration:{THCOptInOnSvc:false
THCEvents:{THCConfigured:false BackendConfigOverridesTHC:false THCAnnotationWithoutFlag:false THCAnnotationWithoutNE
G:false}} BackendConfig:<nil> BackendNamer:0xc00043d940 MaxRatePerEndpoint:<nil> CapacityScaler:<nil>}
I0201 00:12:43.701529 2206655 gen.go:4354] "IngressController/SyncBackends: Getting ga BackendService" ingressKey="d
efault/neg-demo-ing" backendServiceName="k8s1-89f6839e-default-neg-demo-svc-80-b0c7e439" backendVersion="ga" backend
Scope="global" port=0 name="k8s1-89f6839e-default-neg-demo-svc-80-b0c7e439"
```

PR #2317, opened 2023-10-31
Add contextual logging L7 Health Checker.

sawsa307:
* Add contextual logging for L7 Health Checker functions.
* Functions will take the logger from higher level so contexts can be preserved and embedded. 
  Exmaple:
```
I0131 20:02:32.530192 1904521 healthchecks.go:146] "SyncServicePort" servicePortID="default/neg-demo-svc/&ServiceBackendPort{Name:,Number:80,}" nodePort=0 portNumber=80 portName
="http" THCOptInOnSvc=false enableTHC=false                                                                                                                                      
I0131 20:02:32.530259 1904521 healthchecks.go:307] "DefaultNEGHealthCheck()" servicePortID="default/neg-demo-svc/&ServiceBackendPort{Name:,Number:80,}" protocol="HTTP"          
I0131 20:02:32.530292 1904521 healthchecks.go:553] "Getting Health Check" servicePortID="default/neg-demo-svc/&ServiceBackendPort{Name:,Number:80,}" healthCheckName="k8s1-89f683
9e-default-neg-demo-svc-80-b0c7e439" healthCheckVersion="beta" healthCheckScope="global"
```

PR #2308, opened 2023-10-26
Add contextual logging for ingress controller.

sawsa307:
* Initialize a logger instance in cmd/glbc/main.go.
* All ingress controller log messages will be prefixed by `IngressController`. 
  Example:
```
I0111 19:55:55.582574  873925 controller.go:170] "IngressController: Ingress added, enqueuing" ingressKey="default/neg-demo-ing" 
I0111 19:55:55.583044  873925 controller.go:585] "IngressController/preSyncGC: Running preSyncGC" ingressKey="default/neg-demo-ing"
I0111 19:55:55.583458  873925 controller.go:406] "IngressController/SyncBackends: Skip syncing instance groups" ingressKey="default/neg-demo-ing"    
```
* Logging for sub-components like HealthChecker will be added in separate PRs.

PR #2436, opened 2024-01-17
Add context logging for all controllers

alexkats:
This PR adds context logging, with these changes we'll have only one root logger, and all other loggers are children, which gives us the ability to track the context of the source of a log line. It also changes the way of logging, we have a pair of key/values instead of formatting string. It gives the ability to introduce a logger with some common key/value pairs and not repeat them every time in the following log lines. As an example, when we process some service and have 20 log lines, we can create a new logger with `serviceKey` and it will be logged in all of those 20 lines. This PR aims to do only the first iteration of introducing context loggers everywhere, and in order not to make it even bigger, all other improvements (especially with more complex contexts and loggers with specified key/value pair) will be in the follow-up PRs.


pkg/backends/backends.go
```diff
@@ -175,22 +175,22 @@ func (b *Backends) Get(name string, version meta.Version, scope meta.KeyType) (*
 
 // Delete implements Pool.
 func (b *Backends) Delete(name string, version meta.Version, scope meta.KeyType) error {
-	klog.V(2).Infof("Deleting backend service %v", name)
+	b.logger.V(2).Info("Deleting backend service", "backendService", name)
```
sawsa307, line 178: We can do `logger := logger.WithValues("backendServiceName", name)` and use this logger within Delete(), so this context is included in all logs and can be easily queried with backend name.
Same can be applied for Create() and Get().

We can also do similar thing to key and version below so we don't need to replicate this in all log lines.

And we should try to pass the logger with context to helper functions so this context is preserved.

pkg/backends/backends.go
```diff
@@ -175,22 +175,22 @@ func (b *Backends) Get(name string, version meta.Version, scope meta.KeyType) (*
...
 	key, err := composite.CreateKey(b.cloud, name, scope)
 	if err != nil {
 		return err
 	}
-	err = composite.DeleteBackendService(b.cloud, key, version, klog.TODO())
+	err = composite.DeleteBackendService(b.cloud, key, version, b.logger)
 	if err != nil {
 		if utils.IsHTTPErrorCode(err, http.StatusNotFound) || utils.IsInUsedByError(err) {
-			klog.Infof("DeleteBackendService(_, %v, %v) = %v; ignorable error", key, version, err)
+			b.logger.Info("DeleteBackendService; ignorable error", "key", key, "version", version, "err", err)
```
sawsa307, line 187: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments

Can we make it clear on what are key and version for? Like do `backendKey` and `backendVersion`

pkg/backends/backends.go
```diff
@@ -307,40 +307,39 @@ func (b *Backends) DeleteSignedUrlKey(be *composite.BackendService, keyName stri
 // TODO(code-elinka): refactor the list of arguments (there are too many now)
 func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinity, scheme string, namespacedName types.NamespacedName, network network.NetworkInfo, connectionTrackingPolicy *composite.BackendServiceConnectionTrackingPolicy) (*composite.BackendService, error) {
 	start := time.Now()
-	klog.V(2).Infof("EnsureL4BackendService started, backend service name: %v, service: %v, scheme: %v, protocol: %v, sessionAffinity: %v, network: %v, subnetwork: %v ", name, namespacedName, scheme, protocol, sessionAffinity, network.NetworkURL, network.SubnetworkURL)
+	b.logger.V(2).Info("EnsureL4BackendService started", "backendService", name, "serviceKey", namespacedName, "scheme", scheme, "protocol", protocol, "sessionAffinity", sessionAffinity, "network", network.NetworkURL, "subnetwork", network.SubnetworkURL)
```
sawsa307, line 310: logger.WithValues(...)

pkg/backends/backends.go
```diff
@@ -307,40 +307,39 @@ func (b *Backends) DeleteSignedUrlKey(be *composite.BackendService, keyName stri
...
 	}
 	// We need this configuration only for Strong Session Affinity feature
 	if b.useConnectionTrackingPolicy {
-		klog.V(2).Infof("EnsureL4BackendService %v,  service: %v, using connection tracking policy: %+v", name, namespacedName, connectionTrackingPolicy)
+		b.logger.V(2).Info(fmt.Sprintf("EnsureL4BackendService: using connection tracking policy: %+v", connectionTrackingPolicy), "backendService", name, "serviceKey", namespacedName)
 		expectedBS.ConnectionTrackingPolicy = connectionTrackingPolicy
 	}
 	if !network.IsDefault {
-		klog.V(2).Infof("EnsureL4BackendService %v,  service: %v, using non-default network: %+v", name, namespacedName, network)
+		b.logger.V(2).Info(fmt.Sprintf("EnsureL4BackendService: using non-default network: %+v", network), "backendService", name, "serviceKey", namespacedName)
```
sawsa307, line 342: Is there a reason we need to do Sprintf here? I think the intention of using context logging is to have a general message and key/value pairs as contexts. If we want to show the content for `network`, we can do something like #2430

pkg/backends/backends.go
```diff
@@ -352,37 +351,37 @@ func (b *Backends) EnsureL4BackendService(name, hcLink, protocol, sessionAffinit
 
 	// Create backend service if none was found
 	if bs == nil {
-		klog.V(2).Infof("EnsureL4BackendService(%v, _, %v, _, %v, %v ...): creating backend service %v", name, protocol, scheme, namespacedName, name)
-		err := composite.CreateBackendService(b.cloud, key, expectedBS, klog.TODO())
+		b.logger.V(2).Info("EnsureL4BackendService: creating backend service", "backendService", name, "protocol", protocol, "scheme", scheme, "serviceKey", namespacedName)
```
sawsa307, line 354: It would be great if we can consistently use `backendService` and `backendName`. I would prefer `backendName` or `backendServiceName`.

pkg/backends/syncer.go
```diff
@@ -59,7 +63,7 @@ func (s *backendSyncer) Init(pp ProbeProvider) {
 // Sync implements Syncer.
 func (s *backendSyncer) Sync(svcPorts []utils.ServicePort) error {
 	for _, sp := range svcPorts {
-		klog.V(3).Infof("Sync: backend %+v", sp)
+		s.logger.V(3).Info(fmt.Sprintf("Sync: backend %+v", sp))
```
sawsa307, line 66: spLogger := s.logger.WithValues("servicePort", sp)
spLogger.V(3).Info("Sync Backends")

pkg/backends/syncer.go
```diff
@@ -91,7 +95,7 @@ func (s *backendSyncer) ensureBackendService(sp utils.ServicePort) error {
 			return getErr
 		}
 		// Only create the backend service if the error was 404.
-		klog.V(2).Infof("Creating backend service for port %v named %v", sp.NodePort, beName)
```
sawsa307, line 94: We should only use rootLogger/s.logger in root functions like Sync(), and for all helper functions within Sync(), it should use the logger passed from Sync(). This way, the context we embedded in Sync() is preserved.

See comments https://github.com/kubernetes/ingress-gce/pull/2308#pullrequestreview-1725836671.

pkg/backends/syncer.go
```diff
@@ -148,27 +152,27 @@ func (s *backendSyncer) ensureBackendSignedUrlKeys(sp utils.ServicePort, be *com
 	newSignedUrlKeys := []*composite.SignedUrlKey{}
 	if sp.BackendConfig != nil && sp.BackendConfig.Spec.Cdn != nil && sp.BackendConfig.Spec.Cdn.SignedUrlKeys != nil {
 		for _, key := range sp.BackendConfig.Spec.Cdn.SignedUrlKeys {
-			klog.V(5).Infof("Search for SignedUrlKey %q in backend %q", key.KeyName, be.Name)
```
sawsa307, line 151: We should use a named logger here since this context is shared within this iteration of for loop.
```
logger.WithName("signedUrlKey", key.Name, ...)

File: pkg/healthchecks/healthchecks.go [hunk 9]
```diff
@@ -88,11 +91,11 @@ func generateClusterInfo(gceCloud *gce.Cloud) healthcheck.ClusterInfo {
 func (h *HealthChecks) new(sp utils.ServicePort) *translator.HealthCheck {
 	var hc *translator.HealthCheck
```
sawsa307, line 92: Pass the logger from SyncServicePort()

pkg/healthchecks/healthchecks.go
```diff
@@ -117,7 +120,7 @@ func (h *HealthChecks) getService(sp utils.ServicePort) *v1.Service {
 	var err error
 	service, err := h.serviceGetter.GetService(namespacedName.Namespace, namespacedName.Name)
 	if err != nil {
-		klog.Warningf("Service %s/%s needed for emitting an event not found (we'll log instead): %v.", namespacedName.Namespace, namespacedName.Name, err)
```
sawsa307, line 120: Pass the logger from new()

pkg/healthchecks/healthchecks.go
```diff
@@ -141,29 +144,29 @@ func (h *HealthChecks) generateServiceInfo(sp utils.ServicePort) healthcheck.Ser
 
 // SyncServicePort implements HealthChecker.
 func (h *HealthChecks) SyncServicePort(sp *utils.ServicePort, probe *v1.Probe) (string, error) {
-	klog.Infof("SyncServicePort: sp.ID=%v, sp.NodePort=%v, sp.Port=%v, sp.PortName=%v, sp.THCConfiguration.THCOptInOnSvc=%v, h.thcEnabled=%v.", sp.ID, sp.NodePort, sp.Port, sp.PortName, sp.THCConfiguration.THCOptInOnSvc, h.healthcheckFlags.EnableTHC)
+	h.logger.Info("SyncServicePort", "spId", sp.ID, "spNodePort", sp.NodePort, "spPort", sp.Port, "spPortName", sp.PortName, "spTHCOptInOnSvc", sp.THCConfiguration.THCOptInOnSvc, "thcEnabled", h.healthcheckFlags.EnableTHC)
```
sawsa307, line 147: logger.WithValues(...)

pkg/healthchecks/healthchecks.go
```diff
@@ -141,29 +144,29 @@ func (h *HealthChecks) generateServiceInfo(sp utils.ServicePort) healthcheck.Ser
...
-		klog.Warningf("THC flag disabled for HealthChecks, but ServicePort %v has Transparent Health Checks enabled. Disabling.", sp.ID)
+		h.logger.Info("THC flag disabled for HealthChecks, but ServicePort has Transparent Health Checks enabled. Disabling.", "spId", sp.ID)
 		sp.THCConfiguration.THCOptInOnSvc = false
 	}
 	defer func() { sp.THCConfiguration.THCEvents = utils.THCEvents{} }()
 
 	hc := h.new(*sp)
 	if sp.THCConfiguration.THCOptInOnSvc {
-		klog.V(2).Infof("ServicePort %v has Transparent Health Checks enabled", sp.ID)
+		h.logger.V(2).Info("ServicePort has Transparent Health Checks enabled", "spId", sp.ID)
```
sawsa307, line 156: Can we do `servicePortID`?

pkg/healthchecks/healthchecks.go
```diff
@@ -437,35 +440,35 @@ func (h *HealthChecks) updateRegional(hc *translator.HealthCheck) error {
...
-		klog.V(2).Infof("Updating alpha health check with protocol %v", hc.Type)
+		h.logger.V(2).Info("Updating alpha health check", "protocol", hc.Type)
 		alpha, err := hc.ToAlphaComputeHealthCheck()
 		if err != nil {
 			return err
 		}
 		return h.cloud.UpdateAlphaHealthCheck(alpha)
 	case meta.VersionBeta:
-		klog.V(2).Infof("Updating beta health check with protocol %v", hc.Type)
+		h.logger.V(2).Info("Updating beta health check with protocol", "protocol", hc.Type)
```
sawsa307, line 464: "Updating beta health check", "protocol", hc.Type

pkg/healthchecks/healthchecks.go
```diff
@@ -437,35 +440,35 @@ func (h *HealthChecks) updateRegional(hc *translator.HealthCheck) error {
...
 		beta, err := hc.ToBetaComputeHealthCheck()
 		if err != nil {
 			return err
 		}
 		return h.cloud.UpdateBetaHealthCheck(beta)
 	case meta.VersionGA:
-		klog.V(2).Infof("Updating health check %q for port %v with protocol %v", hc.Name, hc.Port, hc.Type)
+		h.logger.V(2).Info("Updating health check for port with protocol", "healthCheckName", hc.Name, "port", hc.Port, "protocol", hc.Type)
```
sawsa307, line 471: "Updating health check for port", "protocol", hc.Type

pkg/healthchecks/healthchecks.go
```diff
@@ -437,35 +440,35 @@ func (h *HealthChecks) updateRegional(hc *translator.HealthCheck) error {
 	compositeType.Version = meta.VersionGA
 	compositeType.Region = key.Region
 
-	return composite.UpdateHealthCheck(cloud, key, compositeType, klog.TODO())
+	return composite.UpdateHealthCheck(cloud, key, compositeType, h.logger)
 }
 
 func (h *HealthChecks) update(hc *translator.HealthCheck) error {
```
sawsa307, line 446: We can do logger.WithValues(..) here and included everything like protocol, name, healthCheck, etc.
It doesn't hurt to add more contexts to logs that don't have them previously.

sawsa307 (commented):
Add some comments in pkg/healthshecks and pkg/backends. Most comments apply for other places too. PTAL!

alexkats:
Waiting for #2317 and #2335 to go in
/hold

cmd/glbc/main.go
```diff
@@ -81,16 +81,17 @@ func main() {
 		os.Exit(0)
 	}
 
-	klog.V(0).Infof("Starting GLBC image: %q, cluster name %q", version.Version, flags.F.ClusterName)
-	klog.V(0).Infof("Latest commit hash: %q", version.GitCommit)
+	rootLogger := klog.TODO()
```
mmamczur, line 84: maybe we should use `klog.Background()` here? I tried finding a recommended way of creating a new root logger but I have no idea

pkg/loadbalancers/l4.go
```diff
@@ -109,18 +111,21 @@ type L4ILBParams struct {
 }
 
 // NewL4Handler creates a new L4Handler for the given L4 service.
-func NewL4Handler(params *L4ILBParams) *L4 {
+func NewL4Handler(params *L4ILBParams, logger klog.Logger) *L4 {
+	logger = logger.WithName("L4Handler")
```
mmamczur, line 115: maybe we should also do `.WithValues("service", svcKey)` here so that all lines produced by this handler have service details?
we could do this as a follow up not to make things too complicated

pkg/l4lb/l4controller.go
```diff
@@ -96,34 +99,35 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}) *L
...
+	l4c.svcQueue = utils.NewPeriodicTaskQueueWithMultipleWorkers("l4", "services", l4c.numWorkers, l4c.syncWrapper, logger)
 
 	if ctx.NetworkInformer != nil {
 		l4c.networkLister = ctx.NetworkInformer.GetIndexer()
 	}
 	if ctx.GKENetworkParamsInformer != nil {
 		l4c.gkeNetworkParamSetLister = ctx.GKENetworkParamsInformer.GetIndexer()
 	}
-	l4c.networkResolver = network.NewNetworksResolver(l4c.networkLister, l4c.gkeNetworkParamSetLister, ctx.Cloud, ctx.EnableMultinetworking, klog.TODO())
+	l4c.networkResolver = network.NewNetworksResolver(l4c.networkLister, l4c.gkeNetworkParamSetLister, ctx.Cloud, ctx.EnableMultinetworking, logger)
```
mmamczur, line 117: I think I messed up the Resolver. Probably it should not keep the logger reference but instead accept a param when resolving the service network so that the lines from there have all context. I'll have to redo it later

pkg/loadbalancers/l4.go
mmamczur: I believe if we do this right then we can get rid of the "serviceKey" key that is added to many log lines in the handlers

cmd/glbc/main.go
alexkats: I don't know as well, but `TODO()` technically just invokes `Background()`

pkg/loadbalancers/l4.go
alexkats: I agree, but I'd rather postpone it for a follow-up. It feels like we already have a lot of changes, so I'd suggest to have straightforward change from `klog...` to context logger, and then have another PR with all the improvements

cezarygerard:
Alex, add more description to this PR

- what it gives us
- why you chose to only 'replace' existing loggers in this PR (rather than utilize klog features)

PR #2437, opened 2024-01-17
Introduce enable-l4-neg flag and use it to flag-gate L4 NEGs.

sawsa307:
* NEG controller should not have logic to understand which controllers are running.
* Introduce enable-l4-neg to flag-gate L4 NEG processing. Its value will be determined from the manifest level.


sawsa307:
/hold
Wait for upstream change to go in first.

swetharepakula (commented):
/hold cancel

/lgtm
/approve

PR #2438, opened 2024-01-18
Trigger node process when providerID populated

songrx1997:
/assign @swetharepakula

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -319,6 +319,11 @@ func NewController(
 				logger.Info("Node has changed, enqueueing", "node", currentNode.Name)
 				negController.enqueueNode(currentNode)
 			}
+			// Trigger a sync when node provider ID changed.
+			if oldNode.Spec.ProviderID != currentNode.Spec.ProviderID {
+				logger.Info("Node provider ID changed, enqueueing", "node", currentNode.Name, "old providerID", oldNode.Spec.ProviderID, "new providerID", currentNode.Spec.ProviderID)
+				negController.enqueueNode(currentNode)
+			}
 		},
 	})
 
--- a/pkg/neg/controller_test.go
+++ b/pkg/neg/controller_test.go
@@ -458,6 +458,53 @@ func TestEnqueueNodeWithILBSubsetting(t *testing.T) {
 	ensureNodeEnqueue(t, nodeKey, controller)
 }
 
+func TestEnqueueNodeWhenProviderIDPopulated(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/neg/controller.go
```diff
@@ -319,6 +319,11 @@ func NewController(
 				logger.Info("Node has changed, enqueueing", "node", currentNode.Name)
 				negController.enqueueNode(currentNode)
 			}
+			// Trigger a sync when node providerID is populated.
+			if oldNode.Spec.ProviderID != currentNode.Spec.ProviderID {
+				logger.Info("Node was populated with provider id, enqueueing", "node", "providerID", currentNode.Name, currentNode.Spec.ProviderID)
```
swetharepakula, line 324: nit: Node provider ID changed

The parameters should be in the format "key1", "value1", "key2","value2"

pkg/neg/controller_test.go
```diff
@@ -458,6 +458,53 @@ func TestEnqueueNodeWithILBSubsetting(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatalf("Failed to create test node, error - %v", err)
+	}
+	nodeKey, err := cache.DeletionHandlingMetaNamespaceKeyFunc(node)
+	if err != nil {
+		t.Fatalf("Failed to create node key for test node %v, error - %v", node, err)
+	}
+	// sleep for the Informer store to pick this up.
+	time.Sleep(5 * time.Second)
```
swetharepakula, line 491: does this need to be 5 seconds?

pkg/neg/controller.go
songrx1997: Done

pkg/neg/controller_test.go
songrx1997: Change this to 1s.

pkg/neg/controller.go
```diff
@@ -319,6 +319,11 @@ func NewController(
 				logger.Info("Node has changed, enqueueing", "node", currentNode.Name)
 				negController.enqueueNode(currentNode)
 			}
+			// Trigger a sync when node provider ID changed.
+			if oldNode.Spec.ProviderID != currentNode.Spec.ProviderID {
+				logger.Info("Node provider ID changed, enqueueing", "node", currentNode.Name, "providerID", currentNode.Spec.ProviderID)
```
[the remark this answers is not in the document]
songrx1997: The providerID can only be changed once from "" to a real string. The old one should always be a empty string here. What do you think?

swetharepakula (commented):
/lgtm
/approve

PR #2440, opened 2024-01-19
Fix Regional XLB ingress by setting Network. 

panslava:
Before LB was created only in default network

/assign @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #2441, opened 2024-01-22
Introduce leader election phase flag.

sawsa307:
* NEG Leader Election lock can only be one of the three defined phases. Default value is phase 0.

/assign @swetharepakula


sawsa307:
Closed after offline discussion. Going with a different approach.

PR #2450, opened 2024-01-25
Make firewall controller waits on stopCh.

sawsa307:
* Firewall controller should waits on stopCh so its shutdown() can be properly executed.


swetharepakula (commented):
/lgtm

PR #2451, opened 2024-01-29
Add lock availability metrics.

sawsa307:
* Add lock availability metrics.
* NEG controller leader election will run separately if leader election is enabled.
* For now, NEG controller leader election only does metrics collection.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -464,3 +464,16 @@ func runWithWg(runFunc func(), wg *sync.WaitGroup) {
 	defer wg.Done()
 	runFunc()
 }
+
+func collectLockAvailabilityMetrics(lockName, clusterType string, stopCh <-chan struct{}) {
+	for {
+		select {
+		case <-stopCh:
+			klog.Infof("StopCh is closed. Stop collecting metrics for %s.", lockName)
+			return
+		case <-time.Tick(time.Second):
+			app.PublishLockAvailabilityMetrics(lockName, clusterType)
+			klog.Infof("Exported %s availability metrics.", lockName)
+		}
+	}
+}
--- a/cmd/glbc/app/metrics.go
+++ b/cmd/glbc/app/metrics.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/klog/v2"
+)
+
+func init() {
+	klog.V(3).Infof("Registering lock availability metrics %v", LockAvailability)
+	prometheus.MustRegister(LockAvailability)
+}
+
+var (
+	// LockAvailability tracks the how long has the container been running with the corresponding resource lock
+	LockAvailability = prometheus.NewCounterVec(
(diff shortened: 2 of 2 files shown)


pkg/neg/metrics/metrics.go
```diff
@@ -375,6 +375,8 @@ func PublishK8sRequestCountMetrics(start time.Time, requestType string, err erro
 	K8sRequestCount.WithLabelValues(requestType, result).Inc()
 }
 
+func PublishNegLockAvailabilityMetrics() {}
```
swetharepakula, line 378: this doesn't do anything yet?

cmd/glbc/main.go
```diff
@@ -431,3 +432,11 @@ func runNEGController(ctx *ingctx.ControllerContext, stopCh <-chan struct{}) {
 	go negController.Run()
 	klog.V(0).Infof("negController started")
 }
+
+func collectNEGLockAvailabilityMetrics() {
+	for {
+		negmetrics.PublishNegLockAvailabilityMetrics()
```
swetharepakula, line 480: if this is going to be a counter, lets just emit the metric her and not bring in a dependency on pkg/neg/metrics into main

pkg/neg/metrics/metrics.go
sawsa307: Yes I'll put the actual implementation in #2452. Here it is a draft of how it will work.

cmd/glbc/main.go
sawsa307: I want to put this metrics under pkg/neg since it is a NEG related metrics, and we don't have a metrics.go in cmd/.

cmd/glbc/main.go
```diff
@@ -294,6 +300,42 @@ func makeLeaderElectionConfig(ctx *ingctx.ControllerContext, client clientset.In
...
+		Lock:          negLock,
+		LeaseDuration: flags.F.LeaderElection.LeaseDuration.Duration,
+		RenewDeadline: flags.F.LeaderElection.RenewDeadline.Duration,
+		RetryPeriod:   flags.F.LeaderElection.RetryPeriod.Duration,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(context.Context) {
+				run()
+			},
+			OnStoppedLeading: func() {
+				klog.Warning("Stopped NEG controller")
```
swetharepakula, line 333: nit, this should just be stopped emitting metrics for now?

cmd/glbc/main.go
swetharepakula: Maybe make a general metric that takes in a key and emits on that?

That way other locks can re-use the function. Perhaps we can put it in cmd/app

cmd/glbc/main.go
```diff
@@ -431,3 +431,11 @@ func runNEGController(ctx *ingctx.ControllerContext, stopCh <-chan struct{}) {
 	go negController.Run()
 	klog.V(0).Infof("negController started")
 }
+
+func collectLockAvailabilityMetrics(lockName, clusterType string) {
```
swetharepakula, line 478: We should be able to close this based on the stop chan as well.

cmd/glbc/main.go
```diff
@@ -465,3 +465,13 @@ func runWithWg(runFunc func(), wg *sync.WaitGroup) {
 	defer wg.Done()
 	runFunc()
 }
+
+func collectLockAvailabilityMetrics(lockName, clusterType string, stopCh <-chan struct{}) {
+	select {
```
swetharepakula, line 470: you need to have a for outside of this, otherwise the metric will stop emitting after one second.

cmd/glbc/main.go
```diff
@@ -465,3 +465,15 @@ func runWithWg(runFunc func(), wg *sync.WaitGroup) {
 	defer wg.Done()
 	runFunc()
 }
+
+func collectLockAvailabilityMetrics(lockName, clusterType string, stopCh <-chan struct{}) {
+	for {
+		select {
+		case <-stopCh:
+			return
```
swetharepakula, line 473: nit can you add a log line that says that the stop chan closed

sawsa307: Added.

swetharepakula (commented):
/lgtm
/approve

PR #2455, opened 2024-01-30
Make all controllers follow the same Run() pattern.

sawsa307:
* We require all controllers follow the same pattern for their Run functions as in they cannot take any parameters. NEG controller and PSC controller are updated to accomodate this.


swetharepakula (commented):
/lgtm
/approve

PR #2456, opened 2024-02-01
Update logic on EnableNegController.

sawsa307:
/assign @swetharepakula

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -147,17 +147,16 @@ func main() {
 			klog.Fatalf("Failed to create Firewall client: %v", err)
 		}
 	}
-	var svcNegClient svcnegclient.Interface
+
 	if flags.F.EnableNEGController {
 		negCRDMeta := svcneg.CRDMeta()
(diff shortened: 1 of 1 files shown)


pkg/context/context.go
```diff
@@ -379,7 +381,7 @@ func (ctx *ControllerContext) Start(stopCh chan struct{}) {
 	if ctx.EnableASMConfigMap && ctx.ConfigMapInformer != nil {
 		go ctx.ConfigMapInformer.Run(stopCh)
 	}
-	if ctx.SvcNegInformer != nil {
+	if ctx.SvcNegInformer != nil && flags.F.EnableNEGController {
```
swetharepakula, line 384: is there anything else that is using the SvcNegInformer?

swetharepakula: I believe ingress also uses the svcneginformer

sawsa307: Right I missed that. It is actually being used by ingress, L4 ILB/NetLB and NEG controllers. Let me update the flag.

swetharepakula (commented):
/lgtm
/approve

PR #2457, opened 2024-02-03
Update NEG controller run behavior based on GateNEGByLock flag.

sawsa307:
* If EnableNEGControllerLeaderElection is true, then we will not only collect metrics on NEG lock, but also run NEG Controller with it.
* NEG Controller will also listen to stopCh, and terminate if stopCh is closed.

/assign @swetharepakula

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -68,6 +68,8 @@ import (
 	"k8s.io/ingress-gce/pkg/version"
 )
 
+const negLockName = "ingress-gce-neg-lock"
+
 func main() {
 	flags.Register()
 	rand.Seed(time.Now().UTC().UnixNano())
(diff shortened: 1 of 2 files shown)


cmd/glbc/main.go
```diff
@@ -394,6 +410,16 @@ func runControllers(ctx *ingctx.ControllerContext) {
 	}
 }
 
+func runNEGController(ctx *ingctx.ControllerContext, option runOption) {
```
sawsa307, line 413: We will add leader election logic in this function in the next commit.

cmd/glbc/main.go
```diff
@@ -473,3 +499,13 @@ func collectLockAvailabilityMetrics(lockName, clusterType string) {
 		time.Sleep(1 * time.Second)
 	}
 }
+
+func getHostID() (string, error) {
```
sawsa307, line 562: Share host id among two leader elections.

cmd/glbc/main.go
```diff
@@ -409,10 +415,66 @@ func runControllers(ctx *ingctx.ControllerContext, option runOption) {
...
 	negController := createNEGController(ctx, option.stopCh)
-	go runWithWg(negController.Run, option.wg)
-	klog.V(0).Infof("negController started")
+	if !option.leaderElect || !flags.F.GateNEGByLock {
+		go runWithWg(negController.Run, option.wg)
+		klog.V(0).Infof("negController started")
+	}
+	if !option.leaderElect {
+		return
+	}
```
swetharepakula, line 428: shouldn't we just always return after line 424?

cmd/glbc/main.go
```diff
@@ -409,10 +415,66 @@ func runControllers(ctx *ingctx.ControllerContext, option runOption) {
...
+	klog.Infof("Start NEG leader election with %s", negLockName)
```
swetharepakula, line 429: nit: reword, "Attempting to grab NEG lock"

cmd/glbc/main.go
```diff
@@ -409,10 +415,66 @@ func runControllers(ctx *ingctx.ControllerContext, option runOption) {
...
 }
 
-func runNEGController(ctx *ingctx.ControllerContext, option runOption) {
+// runNEGController does not need to be blocking because runController is the
+// main thread.
+func runNEGController(ctx *ingctx.ControllerContext, id string, option runOption) {
 	negController := createNEGController(ctx, option.stopCh)
-	go runWithWg(negController.Run, option.wg)
-	klog.V(0).Infof("negController started")
+	if !option.leaderElect || !flags.F.GateNEGByLock {
```
swetharepakula, line 422: actually here we do not need the flags.F.GateNEGByLock.

if no leaderElection, we should just run the neg controller. The gate is only necessary when leader elect is true.

sawsa307: We should run the NEG controller if we have leader election but not gated by flag as well. This corresponds to leader election=true & GateByFlag=false.
At this time we run NEG controller along with other controller, and only spawn leader election to collect metrics on the lock.

swetharepakula (commented):
/lgtm
/approve

PR #2458, opened 2024-02-06
Update project type to be gce-project.

sawsa307:
* Update project type to `gce-project` as `ingress-project` does not exist in community clusters.

/assign @aojea 
/assign @swetharepakula

### DIFF
--- a/hack/boskos.sh
+++ b/hack/boskos.sh
@@ -22,7 +22,7 @@ set -o xtrace
 # acquires a project from boskos
 acquire_project() {
     local project=""
-    local project_type="ingress-project"
+    local project_type="gce-project"
 
     boskos_response=$(curl -X POST "http://boskos.test-pods.svc.cluster.local/acquire?type=${project_type}&state=free&dest=busy&owner=${JOB_NAME}")
 


ameukam:
For reference the pool that will be used: https://github.com/kubernetes/k8s.io/blob/5eb099ddf8c9d7413b75119d043d3588711be831/infra/gcp/terraform/k8s-infra-prow-build/prow-build/resources/test-pods/boskos-resources-configmap.yaml#L5-L127

hack/run-e2e-gce.sh
```diff
@@ -157,7 +157,6 @@ EOF
   --env=GCE_ALPHA_FEATURES=NetworkEndpointGroup \
   --env=KUBE_GCE_ENABLE_IP_ALIASES=true \
   --extract=ci/latest \
-  --gcp-project="${GCP_PROJECT}" \
```
aojea, line 160: you need this

aojea:
I think we just need only the change in `hack/boskos.sh`

sawsa307:
Updated. PTAL!

swetharepakula (commented):
/lgtm
/approve

PR #2465, opened 2024-02-12
Introduce separate backend naming scheme for RXLB Ingress

panslava:
- This is required to share the same service between RXLB and ILB. Frontend resources already have separate names (they depend on the ingress)

/assign @spencerhance

### DIFF
--- a/pkg/utils/serviceport.go
+++ b/pkg/utils/serviceport.go
@@ -88,7 +88,7 @@ func (sp *ServicePort) BackendName() string {
 	if sp.L7XLBRegionalEnabled {
 		return sp.BackendNamer.RXLBBackendName(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.Port)
 	} else if sp.NEGEnabled || sp.VMIPNEGEnabled || sp.L4RBSEnabled {
-		// L4, and ILB, GXLB Ingresses are using NEG Name for all backend resources.
+		// L4 ILB and RBS (with NEGs), Ingress ILB and GXLB are using NEG Name for all backend resources.
 		return sp.NEGName()
 	}
 	return sp.BackendNamer.IGBackend(sp.NodePort)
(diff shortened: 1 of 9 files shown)


spencerhance:
generally LGTM but would like another pair of eyes due to inherent risk in modifying namers

panslava:
there might be a bug

/hold

pkg/utils/serviceport.go
```diff
@@ -85,13 +85,22 @@ func (sp *ServicePort) GetDescription() Description {
 
 // BackendName returns the name of the backend which would be used for this ServicePort.
 func (sp *ServicePort) BackendName() string {
-	if sp.NEGEnabled {
-		return sp.BackendNamer.NEG(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.Port)
-	} else if sp.VMIPNEGEnabled || sp.L4RBSEnabled {
+	if sp.L7XLBRegionalEnabled {
```
swetharepakula, line 88: can you add a unit test to test this condition?

pkg/utils/serviceport.go
```diff
@@ -85,13 +85,22 @@ func (sp *ServicePort) GetDescription() Description {
...
+		return sp.BackendNamer.RXLBBackendName(sp.ID.Service.Namespace, sp.ID.Service.Name, sp.Port)
+	} else if sp.NEGEnabled || sp.VMIPNEGEnabled || sp.L4RBSEnabled {
+		// L4, and ILB, GXLB Ingresses are using NEG Name for all backend resources.
+		return sp.NEGName()
+	}
+	return sp.BackendNamer.IGBackend(sp.NodePort)
+}
+
+func (sp *ServicePort) NEGName() string {
```
swetharepakula, line 97: can we add unit tests for NegName since it is now exposed to be used by itself?

swetharepakula (commented):
This does maintain the status quo for not L4 RXLB. 

I think we need a couple more unit tests that test the conditional logic in the two naming functions (BackendName & NEGName).

pkg/utils/serviceport.go
panslava: Added unit tests for NEGName

pkg/utils/serviceport.go
panslava: Added unit tests for BackendName

swetharepakula (commented):
Thanks Slavik!

/lgtm
/approve

PR #2467, opened 2024-02-15
Remove useless big log

panslava:

printing all the ingresses is overkill

### DIFF
--- a/pkg/firewalls/controller.go
+++ b/pkg/firewalls/controller.go
@@ -267,7 +267,7 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 	if fwc.enableIngressRegionalExternal {
 		rxlbRange, err := fwc.rxlbFirewallSrcRange(gceIngresses)
-		klog.Infof("fwc.rxlbFirewallSrcRange(%+v) = %v, %v", gceIngresses, rxlbRange, err)
+		klog.Infof("fwc.rxlbFirewallSrcRange() = %v, %v", rxlbRange, err)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/firewalls/controller.go
```diff
@@ -267,7 +267,7 @@ func (fwc *FirewallController) sync(key string) error {
 	}
 	if fwc.enableIngressRegionalExternal {
 		rxlbRange, err := fwc.rxlbFirewallSrcRange(gceIngresses)
-		klog.Infof("fwc.rxlbFirewallSrcRange(%+v) = %v, %v", gceIngresses, rxlbRange, err)
+		klog.Infof("fwc.rxlbFirewallSrcRange() = %v, %v", rxlbRange, err)
```
swetharepakula, line 270: is there anything from the ingress object that would be useful to log?

spencerhance: We should probably log the namespace/name right?

panslava: just which namespace/name, as here is the list of all gceIngresses, that's why this log gets very big, when you have multiple ingresses

panslava: we basically currently print alll gce ingresses in cluster, which is kind of unneeded... but I am not sure what else can we print

swetharepakula: Chatted with Slavik, looks like this line was merged accidentally and was a debug line.

The ingresses are actually used just to determine if at least one has RXLB enabled. Otherwise nothing on the ingress is used. Instead of logging this, we should just log which ingress that triggered that condition.

swetharepakula (commented):
/lgtm

PR #2469, opened 2024-02-21
Reset error state even when degraded mode is not enabled.

sawsa307:
* Make sure error state is reset when there is no change in normal and degraded mode calculation.
* This allows errorState to be cleared when enableDegradedMode is false.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -313,13 +313,17 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			s.logger.Info("Using degraded mode endpoint calculation")
 			targetMap = degradedTargetMap
 			endpointPodMap = degradedPodMap
-			if len(notInDegraded) == 0 && len(onlyInDegraded) == 0 {
-				s.logger.Info("Resetting error state")
-				s.resetErrorState()
-			}
 		} else {
 			s.logger.Info("Using normal mode endpoint calculation")
 		}
+	}
+	// When the flags are not enabled, error state should be reset when no
+	// error occurs in the sync.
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -294,6 +294,10 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			if err == nil {
 				computeDegradedModeCorrectness(notInDegraded, onlyInDegraded, string(s.NegSyncerKey.NegType), s.logger)
 			}
+			if len(notInDegraded) == 0 && len(onlyInDegraded) == 0 {
```
swetharepakula, line 297: should this move to right before s.logStats?

Is there any subtle difference that occurs that we process the degraded mode and calculations after we reset the ErrorState versus before we resetErrorState at the end

pkg/neg/syncers/transaction.go
```diff
@@ -313,14 +313,18 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
-				s.logger.Info("Resetting error state")
-				s.resetErrorState()
-			}
 		} else {
 			s.logger.Info("Using normal mode endpoint calculation")
 		}
 	}
+	// Make sure error state does not reset if degraded mode runs into error,
+	// since both results would be empty.
+	if (s.enableDegradedModeMetrics || s.enableDegradedMode) && degradedModeErr == nil {
```
swetharepakula, line 322: If degradedModeErr is not equal to nil, will we even get here?

Shouldn't the only check here be if we are in error state or not?

sawsa307:
@swetharepakula Addressed the comment. Please take a look!

swetharepakula (commented):
/lgtm
/approve

PR #2470, opened 2024-02-21
Add garbage collection when transitioning between ILB and RXLB

panslava:
1. Check if forwarding rule exists with previous lbScheme
2. If yes -- start garbage collection for the ingress


panslava:

Ready for review, appreciate in advance!

PR #2472, opened 2024-02-29
Don't return error if ingress already do not exist

panslava:


pkg/e2e/helpers.go
```diff
@@ -273,7 +273,7 @@ func performWhiteboxTests(s *Sandbox, ing *networkingv1.Ingress, fc *frontendcon
 // resources associated with it to be deleted.
 func WaitForIngressDeletion(ctx context.Context, g *fuzz.GCLB, s *Sandbox, ing *networkingv1.Ingress, options *fuzz.GCLBDeleteOptions) error {
 	crud := adapter.IngressCRUD{C: s.f.Clientset}
-	if err := crud.Delete(ing.Namespace, ing.Name); err != nil {
+	if err := crud.Delete(ing.Namespace, ing.Name); err != nil && !strings.Contains(err.Error(), "not found") {
```
spencerhance, line 276: let's use `if apierrors.IsNotFound(err)` rather than a substring check.

panslava: thanks, done

PR #2483, opened 2024-03-12
Add metric with timestamp of last sync item by l4 controllers

cezarygerard:
The metric was proposed by our SRE when we discussed  improvements in the component.

This metric gives better observability than l4_failed_healthcheck_count / l4_controller_healthcheck

### DIFF
--- a/pkg/l4lb/metrics/metrics.go
+++ b/pkg/l4lb/metrics/metrics.go
@@ -37,6 +37,7 @@ const (
 	L4netlbLegacyToRBSMigrationPreventedMetricName = "l4_netlb_legacy_to_rbs_migration_prevented_count"
 	l4failedHealthCheckName                        = "l4_failed_healthcheck_count"
 	l4ControllerHealthCheckName                    = "l4_controller_healthcheck"
+	l4LastSyncTimeName                             = "l4_last_sync_time"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/l4lb/metrics/metrics.go
```diff
@@ -259,3 +269,9 @@ func IncreaseL4NetLBLegacyToRBSMigrationAttempts() {
 func IncreaseL4NetLBTargetPoolRaceWithRBS() {
 	l4NetLBLegacyToRBSPrevented.WithLabelValues("race").Inc()
 }
+
+// PublishL4controllerLastProcessedItemTime records  L4 controller last processed item time
+func PublishL4controllerLastProcessedItemTime(controllerName string) {
+	l4LastProcessedItemTime.WithLabelValues(controllerName).SetToCurrentTime()
+
```
alexkats, line 276: nit: empty line

pkg/l4lb/metrics/metrics.go
```diff
@@ -144,6 +145,13 @@ var (
 		},
 		[]string{"type"}, // currently, can be migration or race
 	)
+	l4LastProcessedItemTime = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: l4LastProcessedItemTimeName,
+			Help: "Timestamp of last processed item by controller",
```
alexkats, line 151: Should we probably indicate that it's a start sync timestamp rather than the timestamp when the item was processed?

pkg/l4lb/metrics/metrics.go
cezarygerard: done

pkg/l4lb/metrics/metrics.go
cezarygerard: good point, done

alexkats: seems like the change wasn't pushed

PR #2500, opened 2024-03-19
Update google.golang.org/protobuf v1.31.1 -> v1.33.0

gauravkghildiyal:
/assign @swetharepakula


sawsa307:
Shall we do bump on 1.26 versions? I don't think we are using 1.27 in any GKE versions.

swetharepakula (commented):
/lgtm
/approve

PR #2505, opened 2024-03-21
Enhance context logging

alexkats:
/cc @mmamczur 
/cc @cezarygerard

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -201,31 +201,32 @@ func (l4c *L4Controller) shutdown() {
 // If the service has either the v1 finalizer or the forwarding rule created by v1 implementation(service controller),
 // the subsetting controller will not process it. Processing it will fail forwarding rule creation with the same IP anyway.
 // This check prevents processing of v1-implemented services whose finalizer field got wiped out.
-func (l4c *L4Controller) shouldProcessService(service *v1.Service) bool {
+func (l4c *L4Controller) shouldProcessService(service *v1.Service, svcLogger klog.Logger) bool {
 	// Ignore services with LoadBalancerClass set. LoadBalancerClass can't be updated (see the field API doc) so we don't need to worry about cleaning up services that changed the class.
 	if service.Spec.LoadBalancerClass != nil {
-		l4c.logger.Info("Ignoring service managed by another controller", "serviceKey", klog.KRef(service.Namespace, service.Name), "serviceLoadBalancerClass", *service.Spec.LoadBalancerClass)
+		svcLogger.Info("Ignoring service managed by another controller", "serviceLoadBalancerClass", *service.Spec.LoadBalancerClass)
 		return false
 	}
 	// skip services that are being handled by the legacy service controller.
 	if utils.IsLegacyL4ILBService(service) {
-		l4c.logger.Info("Ignoring update for service managed by service controller", "serviceKey", klog.KRef(service.Namespace, service.Name))
+		svcLogger.Info("Ignoring update for service managed by service controller")
 		return false
 	}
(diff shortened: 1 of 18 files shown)


cezarygerard:
looks good overall

I would not use svcLogger in places when we just use one logger

alexkats:
> I would not use svcLogger in places when we just use one logger

I started with just `logger` everywhere, but then I realized that I want to understand what's the context of the logger, especially for some newly added log lines. This name helps to understand that we have a service context inside (`serviceKey` specifically). In the worst case, the keys are deduplicated inside, but I wanted to avoid such situation as much as possible.

pkg/l4lb/l4controller.go
```diff
@@ -201,54 +201,55 @@ func (l4c *L4Controller) shutdown() {
...
 		return false
 	}
 	// skip services that are being handled by the legacy service controller.
 	if utils.IsLegacyL4ILBService(service) {
-		l4c.logger.Info("Ignoring update for service managed by service controller", "serviceKey", klog.KRef(service.Namespace, service.Name))
+		svcLogger.Info("Ignoring update for service managed by service controller")
 		return false
 	}
 	frName := utils.LegacyForwardingRuleName(service)
+	fwLogger := svcLogger.WithValues("forwardingRule", frName)
```
sawsa307, line 216: Can we do frLogger instead of fwLogger? Since fw usually refers to firewall

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -216,9 +215,9 @@ func (l4 *L4) ensureIPv4ForwardingRule(bsLink string, options gce.ILBOptions, ex
 	version := meta.VersionGA
 	frName := l4.GetFRName()
 
-	l4.logger.V(2).Info("Ensuring internal forwarding rule for L4 ILB Service", "forwardingRuleName", frName, "serviceKey", klog.KRef(l4.Service.Namespace, l4.Service.Name), "backendServiceLink", bsLink)
```
sawsa307, line 219: Can we create a new logger at the beginning of this function and embed contexts on forwarding rules? Maybe a new frLogger with the forwarding rule name.

pkg/loadbalancers/l4ipv6.go
```diff
@@ -127,9 +126,9 @@ func (l4 *L4) ensureIPv6NodesFirewall(ipAddress string, nodeNames []string, resu
 	portRanges := utils.GetServicePortRanges(svcPorts)
 	protocol := utils.GetProtocol(svcPorts)
 
-	l4.logger.V(2).Info("Ensuring IPv6 nodes firewall for L4 ILB Service", "firewallName", firewallName, "serviceKey", klog.KRef(l4.Service.Namespace, l4.Service.Name), "ipAddress", ipAddress, "protocol", protocol, "len(nodeNames)", len(nodeNames), "portRanges", portRanges)
```
sawsa307, line 130: Similarly, since this function is to sync/ensure a specific firewall, can we create a new logger and embed the firewall information so we don't need to repeat it in the subsequent log lines?

sawsa307 (commented):
Added a few comments. Generally LGTM!

pkg/l4lb/l4controller.go
alexkats: Yeah, changed, thanks

pkg/loadbalancers/forwarding_rules.go
alexkats: Done

pkg/loadbalancers/l4ipv6.go
alexkats: Done

cezarygerard:
/lgtm

great job, thank you!

PR #2507, opened 2024-03-22
Fix comparison for partial and full resource urls

panslava:
some partial link do not even have project set...

### DIFF
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -121,6 +121,16 @@ func TestEqualResourcePaths(t *testing.T) {
 			b:    "zones/us-central1-a/instanceGroups/example-group",
 			want: true,
 		},
+		"partial without project id vs full": {
+			a:    "https://www.googleapis.com/compute/v1/projects/project-name/regions/us-central1/sslPolicies/example-policy",
+			b:    "regions/us-central1/sslPolicies/example-policy",
+			want: true,
+		},
+		"partial without project id vs partial": {
+			a:    "projects/project-id/zones/us-central1-a/instanceGroups/example-group",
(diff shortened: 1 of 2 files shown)


pkg/utils/utils_test.go
```diff
@@ -179,6 +179,11 @@ func TestEqualResourceIDs(t *testing.T) {
 			b:    "projects/project-id/zones/us-central1-a/instanceGroups/example-group",
 			want: true,
 		},
+		"partial vs full 2": {
```
swetharepakula, line 182: nit: indicate partial without project ID

pkg/utils/utils.go
```diff
@@ -393,6 +393,20 @@ func EqualResourceIDs(a, b string) bool {
 		return false
 	}
 
+	if aId.APIGroup == "" {
```
swetharepakula, line 396: update the comment to indicate that if project or API group is missing, that those are ignored

swetharepakula (commented):
Generally this looks safe, but we should ensure that this is the expectation at all the call sites. We are adding some unique behavior under the hood of a utils function which may be unexpected. 

There also may be upcoming features where we do need to verify project information

swetharepakula:
There also seems to be other ResourceURLs functions in utils: https://github.com/kubernetes/ingress-gce/blob/5284ec45ac7ddb87497290eed1001e25e34a07c8/pkg/utils/utils.go#L339-L382

Do any of those do the same thing as what you are looking fo?

panslava:
Thank you for review @swetharepakula 

Talked offline, there is a better function that suits this case "EqualResourcePaths"

Decided to use it, and just added unit tests for utils

swetharepakula (commented):
/lgtm

PR #2519, opened 2024-04-08
Do wg.Add before spawning a goroutine.

sawsa307:
* wg.Add(1) when the counter is zero must execute before wg.Wait(). This is to ensure wg.Wait() does wait instead of finishing before the counter is incremented.

/assign @swetharepakula
/assign @gauravkghildiyal


swetharepakula (commented):
/lgtm
/approve

PR #2520, opened 2024-04-09
Update type metrics

mmamczur:
Add update opeation type to the L4 latency metrics.

This will allow to separate SLO for updates caused by resource changes from the periodic resync operations.


alexkats:
I guess we don't need #2517 then, right?

mmamczur:
> I guess we don't need #2517 then, right?

yeah I guess it was connected and merged with this one automatically. Interesting I didn't know it works like this but makes sense

PR #2526, opened 2024-04-11
Add a flag a logic to override k8scp compute api endpoint

panslava:
- Should not change any behaviour when flag is not set

This is mostly rollforward of https://github.com/kubernetes/ingress-gce/pull/2060 but with changing to override endpoint conditionally, and using flag as a value


swetharepakula (commented):
/lgtm
/approve

cmd/glbc/main.go
```diff
@@ -213,6 +214,16 @@ func main() {
 	kubeSystemUID := kubeSystemNS.GetUID()
 
 	cloud := app.NewGCEClient(rootLogger)
+
+	if flags.F.OverrideComputeAPIEndpoint != "" {
+		// Globally set the domain for all urls generated by GoogleCloudPlatform/k8s-cloud-provider.
+		// The cloud object is configured by the gce.conf file and parsed in app.NewGCEClient().
+		// basePath will be of the form <URL>/v1
+		domain := utils.GetDomainFromGABasePath(flags.F.OverrideComputeAPIEndpoint)
+		klog.Infof("Overriding k8s-cloud-provider API Domain to %q", domain)
```
alexkats, line 223: Please use context logger, `rootLogger` in this case

panslava: thank you, fixed!

PR #2527, opened 2024-04-16
L7 NEG endpoint calculation updates for MSC.

sawsa307:
Skip endpoints from non-default subnet in calculations, and include it in the validateEndpoint so we won't trigger degraded mode when we filter out endpoints.

/assign @swetharepakula 
/assign @mmamczur


cmd/glbc/main.go
```diff
@@ -213,6 +214,16 @@ func main() {
 	kubeSystemUID := kubeSystemNS.GetUID()
 
 	cloud := app.NewGCEClient(rootLogger)
+	var defaultSubnet string
+	// TODO(cheungdavid): Also gate by L4 and IG msc flags.
+	if flags.F.EnableIngressMultiSubnetCluster || flags.F.EnableNEGMultiSubnetCluster {
+		defaultSubnetURL := cloud.SubnetworkURL()
+		defaultSubnet, err = utils.KeyName(defaultSubnetURL)
```
mmamczur, line 221: so this extracts the last part of the subnet name, right?

from `https://www.googleapis.com/compute/v1/projects/proj/regions/us-central1/subnetworks/defaultSubnet` gets `defaultSubnet`

the labels will container something more like the full URL (Basant should confirm) so may be easier to keep the full value. Also, you have access to `cloud` in most controllers via the ControllerContext

sawsa307: > so this extracts the last part of the subnet name, right?

Yes that's right.

> the labels will container something more like the full URL (Basant should confirm) so may be easier to keep the full value. Also, you have access to cloud in most controllers via the ControllerContext

Yes actually keep the whole url makes more sense. It would be easier for comparison as well.

cmd/glbc/main.go
```diff
@@ -559,6 +559,7 @@ func createNEGController(ctx *ingctx.ControllerContext, stopCh <-chan struct{},
 		lpConfig,
 		flags.F.EnableMultiNetworking,
 		ctx.EnableIngressRegionalExternal,
+		ctx.Cloud.SubnetworkURL(),
```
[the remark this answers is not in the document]
sawsa307: That makes sense. I can move the parsing of the value in the controller, instead of parsing in main and pass it down.

pkg/neg/controller_test.go
```diff
@@ -57,6 +57,8 @@ const (
 	testServiceName         = "test-Name"
 	testNamedPort           = "named-Port"
 	testNamedPortWithNumber = "80"
+
+	defaultSubnetURL = "https://www.googleapis.com/compute/v1/projects/proj/regions/us-central1/subnetworks/default"
```
swetharepakula, line 61: nit: defaultTestSubnetURL

pkg/neg/manager.go
```diff
@@ -115,6 +115,13 @@ type syncerManager struct {
 
 	// lpConfig configures the pod label to be propagated to NEG endpoints.
 	lpConfig podlabels.PodLabelPropagationConfig
+
+	// Whether the NEG controller should process Multi-Subnet Cluster related
+	// changes
+	EnableMultiSubnetCluster bool
```
swetharepakula, line 121: does this need to be exported?

pkg/neg/manager.go
```diff
@@ -133,34 +140,38 @@ func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer,
 	enableDualStackNEG bool,
 	numGCWorkers int,
 	lpConfig podlabels.PodLabelPropagationConfig,
+	EnableMultiSubnetCluster bool,
```
swetharepakula, line 143: nit: enableMultiSubnetCluster

pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -42,22 +42,26 @@ import (
...
-	svcId           string
-	logger          klog.Logger
-	networkInfo     *network.NetworkInfo
+	nodeLister               listers.NodeLister
+	zoneGetter               *zonegetter.ZoneGetter
+	subsetSizeLimit          int
+	svcId                    string
+	logger                   klog.Logger
+	networkInfo              *network.NetworkInfo
+	EnableMultiSubnetCluster bool
```
swetharepakula, line 51: does this need to be exported?

swetharepakula: same comment applies to all the structs we are adding this to

pkg/neg/manager.go
swetharepakula: same applies to all the function signatures we have added this to

pkg/neg/syncers/utils_test.go
```diff
@@ -46,6 +46,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	defaultSubnetURL    = "https://www.googleapis.com/compute/v1/projects/proj/regions/us-central1/subnetworks/default"
```
swetharepakula, line 50: can these be variables instead?

swetharepakula (commented):
mostly nits, otherwise LGTM

mmamczur:
after the changes to zone getter where it knows the flag value and the default subnet - what is the intent of this change? I see the flag and defaultSubnet values being pumped down to the calculators but nothing seems to be acting on these values

pkg/neg/syncers/endpoints_calculator.go
sawsa307: Updated

pkg/neg/controller_test.go
sawsa307: Updated

pkg/neg/manager.go
sawsa307: Updated

pkg/neg/manager.go
sawsa307: No it is not needed, updated

pkg/neg/syncers/utils_test.go
sawsa307: Updated

sawsa307:
> after the changes to zone getter where it knows the flag value and the default subnet - what is the intent of this change? I see the flag and defaultSubnet values being pumped down to the calculators but nothing seems to be acting on these values

Yes the endpoint calculator change should not be needed, thanks for catching that!
This PR is needed so pod not from default subnet are marked as ready without blocking.

mmamczur:
> > after the changes to zone getter where it knows the flag value and the default subnet - what is the intent of this change? I see the flag and defaultSubnet values being pumped down to the calculators but nothing seems to be acting on these values
> 
> Yes the endpoint calculator change should not be needed, thanks for catching that! This PR is needed so pod not from default subnet are marked as ready without blocking.

Hmm I still don't understand
Yo uadd 2 new fields to the `Controller` and `transactionSyncer` structs. Also set's these values. But then nothing seems to use the fields. No "ifs" for `enableMultiSubnetCluster` etc. 
Is there an intended follow up?

sawsa307:
> > > after the changes to zone getter where it knows the flag value and the default subnet - what is the intent of this change? I see the flag and defaultSubnet values being pumped down to the calculators but nothing seems to be acting on these values
> > 
> > 
> > Yes the endpoint calculator change should not be needed, thanks for catching that! This PR is needed so pod not from default subnet are marked as ready without blocking.
> 
> Hmm I still don't understand Yo uadd 2 new fields to the `Controller` and `transactionSyncer` structs. Also set's these values. But then nothing seems to use the fields. No "ifs" for `enableMultiSubnetCluster` etc. Is there an intended follow up?

Yes it will be introduced in a follow up PR

pkg/utils/zonegetter/fake.go
```diff
@@ -97,15 +97,11 @@ func AddFakeNode(zoneGetter *ZoneGetter, node *apiv1.Node) error {
...
-func PopulateFakeNodeInformer(nodeInformer cache.SharedIndexInformer) {
-	defaultSubnetTestLabel := map[string]string{
-		utils.LabelNodeSubnet: defaultTestSubnet,
-	}
+func PopulateFakeNodeInformer(nodeInformer cache.SharedIndexInformer, addMSCNodes bool) {
 	// Ready nodes.
 	if err := nodeInformer.GetIndexer().Add(&apiv1.Node{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   "instance1",
-			Labels: defaultSubnetTestLabel,
```
sawsa307, line 108: Remove subnet label from fake nodes since they are considered as nodes without MSC feature.

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -536,6 +553,210 @@ func TestValidateEndpoints(t *testing.T) {
...
+				t.Errorf("With multi-subnet cluster enabled, received error when calculating endpoint: %v", err)
+			}
+			endpointData, endpointPodMap = tc.mutation(endpointData, endpointPodMap)
+			if got := tc.ecMSC.ValidateEndpoints(endpointData, endpointPodMap, endpointsExcludedInCalculation); !errors.Is(got, tc.expect) {
+				t.Errorf("With multi-subnet cluster enabled, ValidateEndpoints() = %v, expected %v", got, tc.expect)
+			}
+		})
+	}
+
+	// Run tests with multi-subnet cluster enabled.
```
swetharepakula, line 570: i think this line can be removed

sawsa307: Removed, thanks!

swetharepakula (commented):
/lgtm
/approve

PR #2529, opened 2024-04-16
Add feature flag for Multi-Subnet Cluster.

sawsa307:
/assign @swetharepakula 
/cc @mmamczur

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -130,6 +130,7 @@ var (
 		EnableIngressGlobalExternal              bool
 		OverrideComputeAPIEndpoint               string
 		EnableIGMultiSubnetCluster               bool
+		EnableMultiSubnetCluster                 bool
 	}{
 		GCERateLimitScale: 1.0,
 	}
@@ -308,6 +309,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableIngressGlobalExternal, "enable-ingress-global-external", true, "Enable L7 Ingress Global External. Should be disabled when Regional External is enabled.")
 	flag.StringVar(&F.OverrideComputeAPIEndpoint, "override-compute-api-endpoint", "", "Override endpoint that is used to communicate to GCP compute APIs.")
 	flag.BoolVar(&F.EnableIGMultiSubnetCluster, "enable-ig-multi-subnet-cluster", false, "Enable Multi Subnet support for the controllers that use Instance Group backends.")
+	flag.BoolVar(&F.EnableMultiSubnetCluster, "enable-multi-subnet-cluster", false, "Enable Multi Subnet support for all controllers that are running.")
 }
 
 func Validate() {


pkg/controller/controller.go
```diff
@@ -106,7 +106,10 @@ type LoadBalancerController struct {
 
 	ZoneGetter *zonegetter.ZoneGetter
 
-	logger klog.Logger
+	// Whether the Ingress controller should process Multi-Subnet Cluster
+	// related changes
+	enableIngressMultiSubnetCluster bool
```
swetharepakula, line 111: rename to enableMultiSubnetCluster

We are already inside the ingress controller so don't need to add Ingress to the name

pkg/flags/flags.go
```diff
@@ -130,6 +130,8 @@ var (
 		EnableIngressGlobalExternal              bool
 		OverrideComputeAPIEndpoint               string
 		EnableIGMultiSubnetCluster               bool
+		EnableIngressMultiSubnetCluster          bool
+		EnableNEGMultiSubnetCluster              bool
```
swetharepakula, line 134: do we need two? Can we do an EnableMultiSubnetCluster that either names on all clusters or disables?

cc/ @mmamczur 

Nit: can we break out the high level flag creation into its own commit

pkg/neg/controller.go
```diff
@@ -106,6 +106,10 @@ type Controller struct {
 	// gce-regional-external ingresses
 	enableIngressRegionalExternal bool
 
+	// Whether the NEG controller should process Multi-Subnet Cluster related
+	// changes
+	enableNEGMultiSubnetCluster bool
```
swetharepakula, line 111: same comment, this should just be enableMultiSubnetCluster

sawsa307:
Updated, PTAL! @swetharepakula

pkg/flags/flags.go
sawsa307: Currently we have a separate flag in IG controller. I'll leave the cleanup to @mmamczur

mmamczur:
for me, one shared flag works well. However, I remember we wanted  to have a split, specifically for the NEG controller - so I guess you decide. 
I'm ok using this shared flag in my changes.

swetharepakula:
Even when we split, the flag will operate the same. The split just means we can run the neg controller by itself or combined.

pkg/flags/flags.go
```diff
@@ -308,6 +309,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableIngressGlobalExternal, "enable-ingress-global-external", true, "Enable L7 Ingress Global External. Should be disabled when Regional External is enabled.")
 	flag.StringVar(&F.OverrideComputeAPIEndpoint, "override-compute-api-endpoint", "", "Override endpoint that is used to communicate to GCP compute APIs.")
 	flag.BoolVar(&F.EnableIGMultiSubnetCluster, "enable-ig-multi-subnet-cluster", false, "Enable Multi Subnet support for the controllers that use Instance Group backends.")
+	flag.BoolVar(&F.EnableMultiSubnetCluster, "enable-multi-subnet-cluster", false, "Enable Multi Subnet support for all controllers(Ingress/NEG/L4/IG).")
```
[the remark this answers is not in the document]
sawsa307: I think we can scope this PR to only introduce flag. I'll add it to controllers when it is needed, given we have move most logic into zoneGetter

PR #2533, opened 2024-04-24
Multi-subnet clusters: only add nodes from the default network to the shared instance group

mmamczur:


pkg/utils/utils.go
```diff
@@ -512,6 +530,26 @@ var (
 	}
 )
 
+func CandidateNodesInDefaultSubnetPredicate(defaultSubnet string) func(node *api_v1.Node, logger klog.Logger) bool {
+	return func(node *api_v1.Node, logger klog.Logger) bool {
+		predicate := nodePredicateInternal(node, false, false, logger)
```
sawsa307, line 535: can we do `if !CandidateNodesPredicate(node, logger)` here?

mmamczur: changed

pkg/utils/utils.go
```diff
@@ -512,6 +530,25 @@ var (
...
+		}
+		gkeSubnetworkLabel, hasGKESubnetworkLabel := node.Labels[LabelGKESubnetworkName]
+		// if there is no label but the podCIDR is populated assume default network
+		if !hasGKESubnetworkLabel && node.Spec.PodCIDR != "" {
+			return true
+		}
+		if hasGKESubnetworkLabel && EqualResourceIDs(gkeSubnetworkLabel, defaultSubnet) {
+			return true
+		}
+		logger.Info("Skipping node for the instance group since it is from a non-default network", "node", node.Name, "nodeSubnetwork", gkeSubnetworkLabel)
```
sawsa307, line 546: Wondering if we can have separate logs for `Node without PodCIDR` and `Node from non-default subnet`? Right now they would both emit this same log line and can be confusing during debugging

sawsa307: I just created a PR #2534 to generalize the way we do node checking. If we use the same helper functions, it is also easier to grep for logs.
 Let me know if the PR makes sense to you, thanks!

pkg/utils/utils.go
```diff
@@ -481,6 +484,21 @@ func GetReadyNodeNames(lister listers.NodeLister, logger klog.Logger) ([]string,
 	return nodeNames, nil
 }
 
+// GetReadyNodeNamesInDefaultSubnet returns names of schedulable, ready nodes from the node lister
```
cezarygerard, line 487: is it possible to move
GetReadyNodeNamesInDefaultSubnet
and
GetReadyNodeNames

to pkg/instancegroups/nodegetter.go

maybe in separate PR

cezarygerard:
lgtm overall

PR #2534, opened 2024-04-25
Exclude non-default subnet nodes from zoneGetter

sawsa307:
Exclude non-default subnet nodes from zoneGetter.

### DIFF
--- a/pkg/backends/ig_linker_test.go
+++ b/pkg/backends/ig_linker_test.go
@@ -38,7 +38,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const defaultZone = "zone-a"
+const (
+	defaultTestZone      = "zone-a"
+	defaultTestSubnetURL = "https://www.googleapis.com/compute/v1/projects/proj/regions/us-central1/subnetworks/default"
+)
 
 func newTestIGLinker(fakeGCE *gce.Cloud, fakeInstancePool instancegroups.Manager) *instanceGroupLinker {
 	fakeBackendPool := NewPool(fakeGCE, defaultNamer)
@@ -56,8 +59,8 @@ func TestLink(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
 
 	nodeInformer := zonegetter.FakeNodeInformer()
-	fakeZoneGetter := zonegetter.NewZoneGetter(nodeInformer)
-	zonegetter.AddFakeNodes(fakeZoneGetter, defaultZone, "test-instance")
+	fakeZoneGetter := zonegetter.NewZoneGetter(nodeInformer, defaultTestSubnetURL)
+	zonegetter.AddFakeNodes(fakeZoneGetter, defaultTestZone, "test-instance")
 
 	fakeNodePool := instancegroups.NewManager(&instancegroups.ManagerConfig{
 		Cloud:      fakeIGs,
@@ -79,7 +82,7 @@ func TestLink(t *testing.T) {
 	// Mimic the syncer creating the backend.
 	linker.backendPool.Create(sp, "fake-health-check-link", klog.TODO())
 
-	if err := linker.Link(sp, []GroupKey{{Zone: defaultZone}}); err != nil {
+	if err := linker.Link(sp, []GroupKey{{Zone: defaultTestZone}}); err != nil {
 		t.Fatalf("%v", err)
 	}
 
@@ -98,8 +101,8 @@ func TestLinkWithCreationModeError(t *testing.T) {
 	fakeGCE := gce.NewFakeGCECloud(gce.DefaultTestClusterValues())
 
 	nodeInformer := zonegetter.FakeNodeInformer()
-	fakeZoneGetter := zonegetter.NewZoneGetter(nodeInformer)
-	zonegetter.AddFakeNodes(fakeZoneGetter, defaultZone, "test-instance")
+	fakeZoneGetter := zonegetter.NewZoneGetter(nodeInformer, defaultTestSubnetURL)
+	zonegetter.AddFakeNodes(fakeZoneGetter, defaultTestZone, "test-instance")
 
 	fakeNodePool := instancegroups.NewManager(&instancegroups.ManagerConfig{
 		Cloud:      fakeIGs,
@@ -135,7 +138,7 @@ func TestLinkWithCreationModeError(t *testing.T) {
 		// Mimic the syncer creating the backend.
 		linker.backendPool.Create(sp, "fake-health-check-link", klog.TODO())
 
-		if err := linker.Link(sp, []GroupKey{{Zone: defaultZone}}); err != nil {
+		if err := linker.Link(sp, []GroupKey{{Zone: defaultTestZone}}); err != nil {
 			t.Fatalf("%v", err)
 		}
 
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -238,7 +238,8 @@ func NewControllerContext(
 		context.EnableIngressRegionalExternal,
 		logger,
 	)
-	context.ZoneGetter = zonegetter.NewZoneGetter(context.NodeInformer)
+	// The subnet specified in gce.conf is considered as the default subnet.
+	context.ZoneGetter = zonegetter.NewZoneGetter(context.NodeInformer, context.Cloud.SubnetworkURL())
 	context.InstancePool = instancegroups.NewManager(&instancegroups.ManagerConfig{
 		Cloud:      context.Cloud,
 		Namer:      context.ClusterNamer,
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -310,8 +310,8 @@ func NewController(
 			vmIpCandidateNodeCheck := zonegetter.CandidateAndUnreadyNodesFilter
 			vmIpPortCandidateNodeCheck := zonegetter.CandidateNodesFilter
 
-			if zoneGetter.CheckNodeWithPredicate(oldNode, vmIpCandidateNodeCheck, logger) != zoneGetter.CheckNodeWithPredicate(currentNode, vmIpCandidateNodeCheck, logger) ||
-				zoneGetter.CheckNodeWithPredicate(oldNode, vmIpPortCandidateNodeCheck, logger) != zoneGetter.CheckNodeWithPredicate(currentNode, vmIpPortCandidateNodeCheck, logger) {
+			if zoneGetter.IsNodeSelectedByFilter(oldNode, vmIpCandidateNodeCheck, logger) != zoneGetter.IsNodeSelectedByFilter(currentNode, vmIpCandidateNodeCheck, logger) ||
+				zoneGetter.IsNodeSelectedByFilter(oldNode, vmIpPortCandidateNodeCheck, logger) != zoneGetter.IsNodeSelectedByFilter(currentNode, vmIpPortCandidateNodeCheck, logger) {
 				logger.Info("Node has changed, enqueueing", "node", currentNode.Name)
 				negController.enqueueNode(currentNode)
 			}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -92,6 +92,9 @@ const (
 	// be removed in 1.18.
 	LabelAlphaNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
 	DualStackSubnetStackType          = "IPV4_IPV6"
+
+	// LabelNodeSubnet specifies the subnet name of this node.
+	LabelNodeSubnet = "cloud.google.com/gke-np-subnet"
 )
 
 var networkTierErrorRegexp = regexp.MustCompile(`The network tier of external IP is STANDARD|PREMIUM, that of Address must be the same.`)
--- a/pkg/utils/zonegetter/zone_getter.go
+++ b/pkg/utils/zonegetter/zone_getter.go
@@ -87,6 +87,9 @@ type ZoneGetter struct {
 	// singleStoredZone is the single stored zone in the zoneGetter
 	// It is only used in Non-GCP mode.
 	singleStoredZone string
+
+	// The subnetURL of the cluster's default subnet.
+	defaultSubnetURL string
 }
 
 // ZoneForNode returns the zone for a given node by looking up providerID.
@@ -280,9 +283,10 @@ func NewNonGCPZoneGetter(zone string) *ZoneGetter {
 }
 
 // NewZoneGetter initialize a ZoneGetter in GCP mode.
-func NewZoneGetter(nodeInformer cache.SharedIndexInformer) *ZoneGetter {
+func NewZoneGetter(nodeInformer cache.SharedIndexInformer, defaultSubnetURL string) *ZoneGetter {
 	return &ZoneGetter{
-		mode:         GCP,
-		nodeInformer: nodeInformer,
+		mode:             GCP,
+		nodeInformer:     nodeInformer,
+		defaultSubnetURL: defaultSubnetURL,
 	}
 }
--- a/pkg/utils/zonegetter/zone_getter.go
+++ b/pkg/utils/zonegetter/zone_getter.go
@@ -27,6 +27,7 @@ import (
 
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
 )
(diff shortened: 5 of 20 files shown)


pkg/utils/utils.go
```diff
@@ -956,3 +968,38 @@ func AddIngressUserAgent(config *restclient.Config) *restclient.Config {
...
+
+// CheckNodeSubnet checks if the node is in the default subnet.
+func CheckNodeSubnet(node *api_v1.Node, defaultSubnetURL string, logger klog.Logger) error {
+	nodeLogger := logger.WithValues("nodeName", node.Name)
+	subnetURL, err := getNodeSubnetURL(node, logger)
+	if err != nil {
+		nodeLogger.Error(err, "Failed to get subnetURL from node")
+		return err
+	}
+	if subnetURL != "" && subnetURL != defaultSubnetURL {
```
mmamczur, line 980: I'd use something like `EqualResourceIDs(subnetURL, defaultSubnetURL)`

I wouldn't be surprised if we end up in these labels containing URL for the same subnet but with a slightly different format

mmamczur:
seems to contain somewhat similar pieces to #2533 We'll have to converge probably

Add some unit tests for this, looks ok other than that

sawsa307:
> defaultSubnetURL

Added unit tests. PTAL!

pkg/utils/utils.go
```diff
@@ -517,6 +522,14 @@ var (
 	}
 )
 
+// GetPredicateInDefaultSubnet returns the predicate that also filters out nodes in hte non-default subnet
+func GetPredicateInDefaultSubnet(defaultSubnetURL string, predicateFunc func(*api_v1.Node, klog.Logger) bool) func(*api_v1.Node, klog.Logger) bool {
+	return func(node *api_v1.Node, logger klog.Logger) bool {
+		err := CheckNodeSubnet(node, defaultSubnetURL, logger)
```
mmamczur, line 528: maybe we should output an info log here so that if some customer comes complaining we can at least find why nodes are not added

sawsa307: I don't think we need to put this logic here since it is just a wrapper around `CandidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes`, `CandidateNodesPredicate`, and `AllNodesPredicate`, and we didn't put any logs there.

> we can at least find why nodes are not added

It should be available in `CheckNodeSubnet`, where we log each decision we made.

pkg/utils/utils.go
```diff
@@ -956,3 +969,38 @@ func AddIngressUserAgent(config *restclient.Config) *restclient.Config {
...
+	}
+	return nil
+}
+
+// getNodeSubnetURL returns the subnet URL on the node if it exists.
+//
+// The subnet label is guaranteed to be populated when PodCIDR is set.
+// For a node without the label, this should only be a transient state, and it
+// should only happen on existing nodes.
+func getNodeSubnetURL(node *api_v1.Node, nodeLogger klog.Logger) (string, error) {
```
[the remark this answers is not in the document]
sawsa307: I don't think there is any issue of doing this check as a prerequisite. In general, given they are populated together, we should only expect either
1. Label ready, PodCIDR ready
2. Label not ready, PodCIDR not ready 

The following condition should be consider as bugs:
3. Label ready, PodCIDR not ready
4. Label not ready, PodCIDR ready

And doing this check actually adds another layer of protection that we don't run into condition3, while I don't think there is anything we can do for condition4.

pkg/utils/utils.go
```diff
@@ -956,3 +969,43 @@ func AddIngressUserAgent(config *restclient.Config) *restclient.Config {
...
+	subnetURL, err := getNodeSubnetURL(node, logger)
+	if err != nil {
+		nodeLogger.Error(err, "Failed to get subnetURL from node")
+		return err
+	}
+	if subnetURL == "" {
+		nodeLogger.Info("Node does not have subnet label, assumed to be in the default subnet", "subnetURL", subnetURL, "defaultSubnetURL", defaultSubnetURL)
+		return nil
+	}
+	sameSubnet := EqualResourceIDs(subnetURL, defaultSubnetURL)
```
sawsa307, line 985: We should be good on this one, since EqualResourceIDs uses ParseResourceURL, which accept inputs in `projects/<proj>/zones/<zone>/<res>/<name>` and `[https://www.googleapis.com/<apigroup>/<ver>]/projects/<proj>/zones/<zone>/<res>/<name>` format

mmamczur:
looks ok, let me know if you want this merged (not sure if anyone else should take a look) and I can add lgtm

pkg/utils/utils.go
```diff
@@ -517,6 +521,13 @@ var (
 	}
 )
 
+// GetPredicateInDefaultSubnet returns the predicate that also filters out nodes in hte non-default subnet
+func GetPredicateInDefaultSubnet(defaultSubnetURL string, predicateFunc func(*api_v1.Node, klog.Logger) bool) func(*api_v1.Node, klog.Logger) bool {
+	return func(node *api_v1.Node, logger klog.Logger) bool {
+		return predicateFunc(node, logger) && isNodeInDefaultSubnet(node, defaultSubnetURL, logger)
+	}
+}
```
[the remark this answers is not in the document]
sawsa307: Yes, but one reason of doing this is, this should only be a temporary change that we will have for phase0.

For phase1, we will still be getting all nodes from all subnets, and this is the cleanest way I can think of without involving a large code change.

pkg/utils/utils.go
```diff
@@ -517,6 +521,13 @@ var (
 	}
 )
 
+// GetPredicateInDefaultSubnet returns the predicate that also filters out nodes in hte non-default subnet
```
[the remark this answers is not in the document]
sawsa307: Updated.

pkg/utils/utils.go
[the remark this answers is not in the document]
sawsa307: If we want to push the change into the predicate, the predicate interface will need to take in an additional parameter `Subnet`, and we will be reverting all these changes in phase1.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -285,8 +297,9 @@ func NewNonGCPZoneGetter(zone string) *ZoneGetter {
 // NewZoneGetter initialize a ZoneGetter in GCP mode.
 func NewZoneGetter(nodeInformer cache.SharedIndexInformer, defaultSubnetURL string) *ZoneGetter {
 	return &ZoneGetter{
-		mode:             GCP,
-		nodeInformer:     nodeInformer,
+		mode:         GCP,
+		nodeInformer: nodeInformer,
+		// TODO(cheungdavid): Add enableMultiSubnetCluster from flags once it is available.
```
swetharepakula, line 302: shouldn't this already be available?

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -145,7 +186,65 @@ func TestListZones(t *testing.T) {
...
+		},
+		Status: apiv1.NodeStatus{
+			Conditions: []apiv1.NodeCondition{
+				{
+					Type:   apiv1.NodeReady,
+					Status: apiv1.ConditionTrue,
+				},
+			},
+		},
+	}) // Invalid node with valid zone.
```
swetharepakula, line 247: same as above comment, expand on what makes this an invalid node

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -145,7 +186,65 @@ func TestListZones(t *testing.T) {
...
```
swetharepakula, line 208: can you expand why this is an invalid node

pkg/context/context.go
```diff
@@ -238,7 +238,7 @@ func NewControllerContext(
 		context.EnableIngressRegionalExternal,
 		logger,
 	)
-	context.ZoneGetter = zonegetter.NewZoneGetter(context.NodeInformer)
+	context.ZoneGetter = zonegetter.NewZoneGetter(context.NodeInformer, context.Cloud.SubnetworkURL())
```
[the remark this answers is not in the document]
sawsa307: Added

pkg/backends/ig_linker_test.go
```diff
@@ -38,7 +38,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const defaultZone = "zone-a"
+const (
+	defaultZone      = "zone-a"
```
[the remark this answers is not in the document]
sawsa307: Added

pkg/utils/utils_test.go
```diff
@@ -1763,3 +1769,77 @@ func TestGetDomainFromGABasePath(t *testing.T) {
...
+					Labels: map[string]string{
+						LabelNodeSubnet: defaultSubnet,
+					},
+				},
+				Spec: api_v1.NodeSpec{},
+			},
+			want: false,
+		},
+		{
+			desc: "Node without Subnet Label",
```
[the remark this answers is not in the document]
sawsa307: Added

pkg/utils/zonegetter/zone_getter.go
sawsa307: Yes thanks for catching this!

pkg/utils/zonegetter/zone_getter_test.go
sawsa307: Added.

pkg/utils/zonegetter/zone_getter_test.go
sawsa307: Added.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -51,21 +51,23 @@ const (
 
 // nodeConditionPredicate is a function that indicates whether the given node's conditions meet
 // some set of criteria defined by the function.
```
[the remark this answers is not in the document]
sawsa307: Yes we need the bool and the subnetURL to determine if we need to filter and what is the default. The decision/check is done within nodePredicateInternal.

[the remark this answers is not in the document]
sawsa307: I moved one step further to resolved the discussion we had in https://github.com/kubernetes/ingress-gce/pull/2547#discussion_r1595911411

go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/ingress-gce
 
-go 1.22.0
+go 1.22
```
swetharepakula: Was this meant to be part of this PR?

pkg/utils/zonegetter/zone_getter.go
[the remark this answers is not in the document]
swetharepakula: I am wondering should we make these zone getter private helper functions? Then you only need variable. The point was we didn't want to have to keep sending the defaultSubnetURL everywhere. Now the zone getter takes care of all of that, caller just needs to know if they care about the defaultSubnet or no.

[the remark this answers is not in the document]
sawsa307: Updated.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -100,34 +85,54 @@ func (z *ZoneGetter) ZoneForNode(name string, logger klog.Logger) (string, error
...
+		nodeLogger.Error(err, "Failed to get zone from the providerID")
 		return "", err
 	}
 	return zone, nil
 }
 
 // ListNodes returns a list of nodes that satisfy the given node filtering mode.
 func (z *ZoneGetter) ListNodes(filter Filter, logger klog.Logger) ([]*api_v1.Node, error) {
-	nodeLister := z.nodeInformer.GetIndexer()
 	logger.Info("Listing nodes", "filter", filter)
```
swetharepakula, line 112: can you add the filter to the logger

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -100,34 +85,54 @@ func (z *ZoneGetter) ZoneForNode(name string, logger klog.Logger) (string, error
 		return z.singleStoredZone, nil
 	}
 
-	nodeLister := z.nodeInformer.GetIndexer()
-	node, err := listers.NewNodeLister(nodeLister).Get(name)
+	nodeLogger := logger.WithValues("nodeName", name)
```
swetharepakula, line 88: can you add the filter as well?

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -638,65 +956,87 @@ func TestGetNodeConditionPredicate(t *testing.T) {
...
 		}
-		unreadyAccept := unreadyPred(&test.node, klog.TODO())
+		unreadyAccept := zoneGetter.candidateNodesPredicateIncludeUnreadyExcludeUpgradingNodes(&test.node, klog.TODO())
 		if unreadyAccept != test.expectAcceptByUnreadyNodePredicate {
 			t.Errorf("Test failed for unreadyNodesPredicate in case %s, got %v, want %v", test.name, unreadyAccept, test.expectAcceptByUnreadyNodePredicate)
 		}
 	}
 }
 
-func TestGetPredicate(t *testing.T) {
```
swetharepakula, line 653: do we have coverage on IsNodeSelectedByFilter to ensure it does this conversion between the string to the filter correctly?

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -153,77 +161,51 @@ func (z *ZoneGetter) ListZones(filter Filter, logger klog.Logger) ([]string, err
...
-		return "", fmt.Errorf("%w: node %s does not have providerID", ErrProviderIDNotFound, node.Name)
-	}
-	matches := providerIDRE.FindStringSubmatch(node.Spec.ProviderID)
-	if len(matches) != 4 {
-		return "", fmt.Errorf("%w: providerID %q of node %s is not valid", ErrSplitProviderID, node.Spec.ProviderID, node.Name)
-	}
-	if matches[2] == "" {
-		return "", fmt.Errorf("%w: node %s has an empty zone", ErrSplitProviderID, node.Name)
+func (z *ZoneGetter) nodePredicateInternal(node *api_v1.Node, includeUnreadyNodes, excludeUpgradingNodes bool, nodeLogger klog.Logger) bool {
+	if z.onlyUseDefaultSubnetNodes && !isNodeInDefaultSubnet(node, z.defaultSubnetURL, nodeLogger) {
```
mmamczur, line 203: I think this will make it more difficult to later separate the logic for instance groups that will always be in the mode of 'default subnet only', there is no way to pass in the call context (if it is IGs or other code)

it is ok for now but we must make sure we change it before we start flipping that flag back once we implement musti-subnet fully

mmamczur:
looks ok for me in general

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -100,34 +85,54 @@ func (z *ZoneGetter) ZoneForNode(name string, logger klog.Logger) (string, error
...
 	return zone, nil
 }
 
 // ListNodes returns a list of nodes that satisfy the given node filtering mode.
 func (z *ZoneGetter) ListNodes(filter Filter, logger klog.Logger) ([]*api_v1.Node, error) {
-	nodeLister := z.nodeInformer.GetIndexer()
 	logger.Info("Listing nodes", "filter", filter)
-	nodes, err := listNodesWithFilter(listers.NewNodeLister(nodeLister), filter, logger)
+
+	nodeLister := listers.NewNodeLister(z.nodeInformer.GetIndexer())
```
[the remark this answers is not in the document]
sawsa307: Yes we can get rid of `GetIndexer()`, but we still need to keep `listers.NewNodeLister` part because `listers.NewNodeLister` returns a lister which is immutable, and we won't be able to add any fake nodes with it, while we can mutate the node indexer.

pkg/utils/zonegetter/zone_getter.go
sawsa307: Yes there will be more updates to the zoneGetter for phase1. At that time, nodePredicateInternal will take in another parameter `onlyIncludeDefaultSubnetNodes`, similar to `includeUnreadyNodes` and `excludeUpgradingNodes`. If this is set to true, zoneGetter will then lookup the default subnet URL from its field and do the filter correspondingly.

sawsa307 (commented):
Addressed all comments, PTAL!

pkg/neg/controller.go
```diff
@@ -327,6 +327,8 @@ func NewController(
 		negController.enableASM = enableAsm
 		negController.asmServiceNEGSkipNamespaces = asmServiceNEGSkipNamespaces
 	}
+	_, err = zoneGetter.ListZones(zonegetter.AllNodesFilter, klog.TODO())
```
swetharepakula, line 330: Is this needed? Perhaps change got added onto the wrong commit?

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -266,12 +245,53 @@ func nodePredicateInternal(node *api_v1.Node, includeUnreadyNodes, excludeUpgrad
...
+// subnet.
+func isNodeInDefaultSubnet(node *api_v1.Node, defaultSubnetURL string, nodeAndFilterLogger klog.Logger) bool {
+	if node.Spec.PodCIDR == "" {
+		nodeAndFilterLogger.Info("Node does not have PodCIDR set")
+		return false
+	}
 
+	nodeSubnet := node.Labels[utils.LabelNodeSubnet]
+	if nodeSubnet == "" {
+		nodeAndFilterLogger.Info("Node does not have subnet label, assumed to be in the default subnet")
```
mmamczur, line 269: now it seems the label will exist but will contain an empty string for the default network. The logic would work here but this line would be logged most of the times and be misleading

sawsa307: Updated log message

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -266,12 +245,57 @@ func nodePredicateInternal(node *api_v1.Node, includeUnreadyNodes, excludeUpgrad
...
+		return false
+	}
+
+	nodeSubnet, exist := node.Labels[utils.LabelNodeSubnet]
+	if !exist {
+		nodeAndFilterLogger.Info("Node does not have subnet label, assumed to be in the default subnet")
+		return true
+	}
+	if nodeSubnet == "" {
+		nodeAndFilterLogger.Info("Node has empty subnet label, assumed to be in the default subnet")
```
swetharepakula, line 273: nit: reword to 
"Node has empty value for <key>, assumed to be in the default subnet"

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -266,12 +245,57 @@ func nodePredicateInternal(node *api_v1.Node, includeUnreadyNodes, excludeUpgrad
...
+// subnet.
+func isNodeInDefaultSubnet(node *api_v1.Node, defaultSubnetURL string, nodeAndFilterLogger klog.Logger) bool {
+	if node.Spec.PodCIDR == "" {
+		nodeAndFilterLogger.Info("Node does not have PodCIDR set")
+		return false
+	}
+
+	nodeSubnet, exist := node.Labels[utils.LabelNodeSubnet]
+	if !exist {
+		nodeAndFilterLogger.Info("Node does not have subnet label, assumed to be in the default subnet")
```
swetharepakula, line 269: nit: reword to
"Node does not have label key <>, assumed to be in the default subnet"

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -266,12 +245,57 @@ func nodePredicateInternal(node *api_v1.Node, includeUnreadyNodes, excludeUpgrad
...
+		return true
+	}
+	if nodeSubnet == "" {
+		nodeAndFilterLogger.Info("Node has empty subnet label, assumed to be in the default subnet")
+		return true
+	}
+
```
swetharepakula, line 276: can you add a log mentioning that we got an non empty label?

pkg/utils/zonegetter/zone_getter.go
sawsa307: Updated

pkg/utils/zonegetter/zone_getter.go
sawsa307: Updated.

pkg/utils/zonegetter/zone_getter.go
sawsa307: Added

swetharepakula (commented):
/lgtm
/approve

PR #2535, opened 2024-04-25
Update zone getter to exclude non-default subnet nodes

sawsa307:
/assign @swetharepakula 
/assign @mmamczur


sawsa307:
Closed since merged with #2534

PR #2552, opened 2024-05-15
Only emit errors for non-LoadBalancer IPv6 services that need NEGs

sawsa307:


swetharepakula (commented):
/lgtm
/approve

/cc @aojea

PR #2553, opened 2024-05-16
Create fake zone getter

sawsa307:
/assign @swetharepakula


sawsa307:
Since the feature flag is not always-on, we need to make sure we have test coverage for both flag is on and off. Closing this as no longer needed.

PR #2557, opened 2024-05-21
Refactor EnsureL4BackendService to use params struct as an argument

cezarygerard:
Instead of lengthy list of args


code-elinka:
/approve 
lgtm

PR #2558, opened 2024-05-21
Make KubeClient QPS configurable

swetharepakula:


cmd/glbc/app/clients.go
```diff
@@ -133,3 +140,32 @@ func generateConfigReaderFunc(config []byte) readerFunc {
...
+// addKubeClientQPSSettings updates the provided config with the QPS settings
+// determined by the KubeClientQPS and KubeClientBurst flags
+func addKubeClientQPSSettings(config *rest.Config, logger klog.Logger) {
+	if flags.F.KubeClientQPS == 0 {
+		logger.Info("Setting Kubernetes Client to default QPS")
+	} else {
+		logger.Info("Setting Kubernetes Client to %d", flags.F.KubeClientQPS)
+	}
+
+	if flags.F.KubeClientQPS == 0 {
```
gauravkghildiyal, line 163: typo: `KubeClientBurst`

cmd/glbc/app/clients.go
```diff
@@ -133,3 +140,32 @@ func generateConfigReaderFunc(config []byte) readerFunc {
...
+	config.UserAgent = ingressUserAgent()
+}
+
+// addKubeClientQPSSettings updates the provided config with the QPS settings
+// determined by the KubeClientQPS and KubeClientBurst flags
+func addKubeClientQPSSettings(config *rest.Config, logger klog.Logger) {
+	if flags.F.KubeClientQPS == 0 {
+		logger.Info("Setting Kubernetes Client to default QPS")
+	} else {
+		logger.Info("Setting Kubernetes Client to %d", flags.F.KubeClientQPS)
```
gauravkghildiyal, line 160: Does the log need to mention "Setting Kubernetes Client QPS to %d"? (emphasis on "QPS", similar doubt regard the other comment for Burst)

gauravkghildiyal (commented):
(Sorry for the delay, I wasn't sure if you want me to wait for the review after you unhold, but looks like it's quite simple so far)

cmd/glbc/app/clients.go
```diff
@@ -148,3 +150,22 @@ func ingressUserAgent() string {
 func addIngressUserAgent(config *rest.Config) {
 	config.UserAgent = ingressUserAgent()
 }
+
+// addKubeClientQPSSettings updates the provided config with the QPS settings
+// determined by the KubeClientQPS and KubeClientBurst flags
+func addKubeClientQPSSettings(config *rest.Config, logger klog.Logger) {
+	if flags.F.KubeClientQPS == 0 {
+		logger.Info("Setting Kubernetes Client to default QPS")
```
sawsa307, line 158: I don't think we are actually "setting the number to default" here.
What we are doing is we set it as 0, so in the package which creates the client, it detects the zero value and overwrite it with the default value.
```
if rateLimiter == nil {
	qps := config.QPS
	if config.QPS == 0.0 {
		qps = DefaultQPS
	}
	burst := config.Burst
	if config.Burst == 0 {
		burst = DefaultBurst
	}
	if qps > 0 {
		rateLimiter = flowcontrol.NewTokenBucketRateLimiter(qps, burst)
	}
}
```
I would prefer we state the fact we set it as 0 so default value will be used.

swetharepakula:
/unhold
The PR is now ready for review as I have added a commit to separate out the events kube client

gauravkghildiyal:
/lgtm

Thanks!

PR #2560, opened 2024-05-22
L7 readiness reflector update for MSC.

sawsa307:
/assign @swetharepakula

### DIFF
--- a/pkg/neg/readiness/reflector.go
+++ b/pkg/neg/readiness/reflector.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/ingress-gce/pkg/neg/metrics"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
 	"k8s.io/ingress-gce/pkg/neg/types/shared"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 )
@@ -74,10 +75,16 @@ type readinessReflector struct {
 
 	queue workqueue.RateLimitingInterface
 
+	zoneGetter *zonegetter.ZoneGetter
+
+	// If enabled, the reflector will mark pods that are from the non-default
+	// subnet as ready without processing.
+	enableMultiSubnetCluster bool
+
 	logger klog.Logger
 }
 
-func NewReadinessReflector(kubeClient kubernetes.Interface, podLister cache.Indexer, negCloud negtypes.NetworkEndpointGroupCloud, lookup NegLookup, enableDualStackNEG bool, logger klog.Logger) Reflector {
+func NewReadinessReflector(kubeClient kubernetes.Interface, podLister cache.Indexer, negCloud negtypes.NetworkEndpointGroupCloud, lookup NegLookup, zoneGetter *zonegetter.ZoneGetter, enableDualStackNEG, enableMultiSubnetCluster bool, logger klog.Logger) Reflector {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartLogging(klog.Infof)
 	broadcaster.StartRecordingToSink(&unversionedcore.EventSinkImpl{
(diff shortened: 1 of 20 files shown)


pkg/neg/readiness/reflector_test.go
```diff
@@ -271,24 +334,26 @@ func TestSyncPod(t *testing.T) {
...
-			pod.CreationTimestamp = tc.expectPod.CreationTimestamp
-			if !reflect.DeepEqual(pod, tc.expectPod) {
-				t.Errorf("For test case %q, expect pod to be %v, but got %v", tc.desc, tc.expectPod, pod)
-			}
-		}
 
+			if tc.expectExists {
+				pod, err := fakeContext.KubeClient.CoreV1().Pods(testServiceNamespace).Get(context.TODO(), tc.expectPod.Name, metav1.GetOptions{})
+				if err != nil {
+					t.Errorf("For test case %q, expect err to be nil, but got %v", tc.desc, err)
```
swetharepakula, line 402: nit: reword this error or reword the error on 343 so we are able to distinguish which one is a problem when it fails.

pkg/neg/readiness/reflector.go
```diff
@@ -74,10 +75,16 @@ type readinessReflector struct {
 
 	queue workqueue.RateLimitingInterface
 
+	zoneGetter *zonegetter.ZoneGetter
+
+	// Whether the NEG controller should process Multi-Subnet Cluster related
+	// changes
+	enableMultiSubnetCluster bool
```
swetharepakula, line 82: nit: in the comment, we should indicate what it means for the readiness reflector, so maybe something like if enabled the reflector will mark pods that are from the non-default subnet as ready without processing.

pkg/neg/readiness/reflector_test.go
```diff
@@ -354,6 +407,133 @@ func TestSyncPod(t *testing.T) {
...
+
+			err = testReadinessReflectorMSC.syncPod(tc.inputKey, tc.inputNeg, tc.inputBackendService)
+			if err != nil {
+				t.Errorf("For test case %q with multi-subnet cluster enabled, expect err to be nil, but got %v", tc.desc, err)
+			}
+
+			if tc.expectExists {
+				pod, err := fakeContext.KubeClient.CoreV1().Pods(testServiceNamespace).Get(context.TODO(), tc.expectPod.Name, metav1.GetOptions{})
+				if err != nil {
+					t.Errorf("For test case %q with multi-subnet cluster enabled,, expect err to be nil, but got %v", tc.desc, err)
```
swetharepakula, line 419: nit: extra comma in the error message

pkg/neg/readiness/reflector_test.go
```diff
@@ -65,10 +84,14 @@ func TestSyncPod(t *testing.T) {
 	nodeLister := fakeContext.NodeInformer.GetIndexer()
 	nodeName := "instance1"
 
-	testReadinessReflector := newTestReadinessReflector(fakeContext)
+	testReadinessReflector := newTestReadinessReflector(fakeContext, false)
 	testReadinessReflector.clock = fakeClock
 	testlookUp := testReadinessReflector.lookup.(*fakeLookUp)
 
+	testReadinessReflectorMSC := newTestReadinessReflector(fakeContext, true)
```
swetharepakula, line 91: instead of duplicating this since in most cases they will behave the same, perhaps we can simplify by moving the reflector creation to later in the test and loop over true and false for the `enableMultiSubnetCluster`

Then we can add a different test that is specific to multiple subnets in which you can do the dual reflectors and verify the outputs, but its more to simplify the logic in this test

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -252,32 +252,43 @@ func (z *ZoneGetter) nodePredicateInternal(node *api_v1.Node, includeUnreadyNode
 	return true
 }
 
+// ZoneForNode returns if the given node is in default subnet.
+func (z *ZoneGetter) IsDefaultSubnetNode(name string, logger klog.Logger) bool {
```
swetharepakula, line 256: nit: rename `IsNodeInDefaultSubnet` ?

pkg/neg/readiness/reflector_test.go
sawsa307: Sounds good, updated

pkg/neg/readiness/reflector_test.go
sawsa307: Updated

pkg/neg/readiness/reflector.go
sawsa307: Updated.

pkg/utils/zonegetter/zone_getter.go
sawsa307: The subtle difference in naming is intended because the signatures are different, one takes node name while the other one takes Node. Having `isNodeInDefaultSubnet` to supply node as parameter, and `IsDefaultSubnetNode` with node name as parameters makes more sense to me.

swetharepakula: okay, we can leave it, but can you rename `name` to `nodeName` so it is clear what we need to pass along

sawsa307: Updated, thanks!

pkg/neg/readiness/reflector_test.go
```diff
@@ -338,22 +379,142 @@ func TestSyncPod(t *testing.T) {
...
+					if !reflect.DeepEqual(pod, tc.expectPod) {
+						t.Errorf("For test case %q with enableMultiSubnetCluster = %v, expect pod to be %v, but got %v", tc.desc, enableMultiSubnetCluster, tc.expectPod, pod)
+					}
+				}
+			}
+		})
+	}
+}
+
+func TestSyncPodMultipleSunbets(t *testing.T) {
```
swetharepakula, line 406: nit : TestSyncPodMultipleSubnets

swetharepakula (commented):
/lgtm
/approve

PR #2564, opened 2024-05-23
Version update of informerfirewall

maciejriedl:
Updates version of 'informerfirewall' from v1beta to v1.


cezarygerard:
I think you should bump dependencies in go.mod
the "k8s.io/cloud-provider-gcp/crd" 


and then run 


  go mod tidy
  go mod vendor

I don't think you should touch the vendor/modules.txt

PR #2565, opened 2024-05-23
L4 RBS External LB NEG support

mmamczur:
This PR contains the implementation of NEG variant of the L4 RBS LoadBalancer

PR #2566, opened 2024-05-23
NEG controller part - L4 RBS External LB NEG support

mmamczur:
This PR contains changes for the NEG controller to support L4 NetLB services.
The NEG controller will be triggered for NetLBs with the presence of the new netlb V3 finalizer

Also the PortInfo had to be extended to contain the type of L4 LB since the calculators need this info in order to pick the right subset sizes. NetLBs will use larger subsets than ILB. Also this allows to differentiate a NEG syncer for ILB vs one for NetLB, when the service moves between ILB and NetLB the syncer for the old service spec will be stopped and a new one created for the changed service.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -116,6 +116,8 @@ var (
 		EnableL4NetLBDualStack                   bool
 		EnableNEGController                      bool
 		EnableL4NEG                              bool
+		EnableL4NetLBNEG                         bool
+		EnableL4NetLBNEGDefault                  bool
 		GateNEGByLock                            bool
 		EnableMultipleIGs                        bool
 		EnableL4StrongSessionAffinity            bool
@@ -281,6 +283,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.RunL4NetLBController, "run-l4-netlb-controller", false, `Optional, if enabled then the L4NetLbController will be run.`)
 	flag.BoolVar(&F.EnableNEGController, "enable-neg-controller", true, `Optional, if enabled then the NEG controller will be run.`)
 	flag.BoolVar(&F.EnableL4NEG, "enable-l4-neg", false, `Optional, if enabled then the NEG controller will process L4 NEGs.`)
+	flag.BoolVar(&F.EnableL4NetLBNEG, "enable-l4-netlb-neg", false, `Optional, if enabled then the NetLB controller can create L4 NetLB services with NEG backends.`)
+	flag.BoolVar(&F.EnableL4NetLBNEGDefault, "enable-l4-netlb-neg-default", false, `Optional, if enabled then newly created L4 NetLB services will use NEG backends. Has effect only if '--enable-l4-netlb-neg' is set to true.`)
 	flag.BoolVar(&F.GateNEGByLock, "gate-neg-by-lock", false, "If enabled then the NEG controller will be run via leader election with NEG resource lock")
 	flag.BoolVar(&F.EnableIGController, "enable-ig-controller", true, `Optional, if enabled then the IG controller will be run.`)
 	flag.BoolVar(&F.EnablePSC, "enable-psc", false, "Enable PSC controller")
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -97,12 +97,15 @@ type Controller struct {
 	// syncerMetrics collects NEG controller metrics
 	syncerMetrics *syncMetrics.SyncerMetrics
 
-	// runL4 indicates whether to run NEG controller that processes L4 services
-	runL4 bool
+	// runL4ForILB indicates whether to run NEG controller that processes L4 ILB services
+	runL4ForILB bool
 
 	// enableIngressRegionalExternal indicates where NEG controller should process
 	// gce-regional-external ingresses
 	enableIngressRegionalExternal bool
+
+	// runL4ForNetLB indicates if the controller can create NEGs for L4 NetLB services.
+	runL4ForNetLB bool
 
 	stopCh <-chan struct{}
 	logger klog.Logger
@@ -140,6 +143,7 @@ func NewController(
 	lpConfig labels.PodLabelPropagationConfig,
 	enableMultiNetworking bool,
 	enableIngressRegionalExternal bool,
+	runL4ForNetLB bool,
 	stopCh <-chan struct{},
 	logger klog.Logger,
 ) *Controller {
@@ -231,8 +235,9 @@ func NewController(
 		syncTracker:                   utils.NewTimeTracker(),
 		reflector:                     reflector,
 		syncerMetrics:                 syncerMetrics,
-		runL4:                         runL4Controller,
+		runL4ForILB:                   runL4Controller,
 		enableIngressRegionalExternal: enableIngressRegionalExternal,
+		runL4ForNetLB:                 runL4ForNetLB,
 		stopCh:                        stopCh,
 		logger:                        logger,
 	}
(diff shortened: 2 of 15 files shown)


mmamczur:
/hold

let's not merge this before the next code freeze ends

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -37,7 +37,7 @@ import (
 	"k8s.io/klog/v2"
 )
 
-// TestLocalGetEndpointSet verifies the GetEndpointSet method implemented by the LocalL4ILBEndpointsCalculator.
+// TestLocalGetEndpointSet verifies the GetEndpointSet method implemented by the LocalL4EndpointsCalculator.
```
swetharepakula, line 40: Can you add tests to cover the logic of switching between netlb and ilb?

pkg/neg/controller_test.go
```diff
@@ -158,6 +158,7 @@ func newTestControllerWithParamsAndContext(kubeClient kubernetes.Interface, test
 		labels.PodLabelPropagationConfig{},
 		true,
 		false,
+		false,
```
swetharepakula, line 161: can you add tests where this is true?

swetharepakula (commented):
Though the core logic of selecting endpoints has not changed, I think we need more tests to cover this new knob.

pkg/neg/controller.go
```diff
@@ -600,7 +605,7 @@ func (c *Controller) mergeStandaloneNEGsPortInfo(service *apiv1.Service, name ty
 func (c *Controller) mergeVmIpNEGsPortInfo(service *apiv1.Service, name types.NamespacedName, portInfoMap negtypes.PortInfoMap, negUsage *metricscollector.NegServiceState, networkInfo *network.NetworkInfo) error {
 	wantsILB, _ := annotations.WantsL4ILB(service)
 	wantsNetLB, _ := annotations.WantsL4NetLB(service)
-	needsNEGForNetLB := wantsNetLB && !networkInfo.IsDefault && annotations.HasRBSAnnotation(service)
+	needsNEGForNetLB := wantsNetLB && (!networkInfo.IsDefault || utils.HasL4NetLBFinalizerV3(service) && c.enableL4NetLBNEGs) && annotations.HasRBSAnnotation(service)
```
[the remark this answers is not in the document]
mmamczur: I'v split it

pkg/neg/controller_test.go
mmamczur: added a test case to verify combinations of this being turned off etc

pkg/neg/syncers/endpoints_calculator_test.go
mmamczur: this is a good point, adding a test here won't help much. To sum up
The NetLB and ILB will use the same syncing logic but with different subset sizes.
The endpoint calculators are created together with the syncer and keep 'living' in the controller.
There is no mechanism to update the running syncers. What we can do here is to treat these syncers as different. When the service is moved ILB <-> NetLB just stop the old syncer and create a new one. I added the LB type to the syncer key, this will achieve just that.
I intended to add a test for this in the controller_test, it would work but there is a problem with the fake SvcNEG indexer and client. When a SvcNEG is created the indexer doesn't know. There is code that checks if an object extsts using the indexer and then uses the client to delete https://github.com/kubernetes/ingress-gce/blob/c6b0781404b8f221415786adb342d24f122c8f7c/pkg/neg/manager.go#L457-L475. Of course in this case it does not delete since it thinks the object does not exist.
Not sure how to resolve it. Maybe just e2e test that handles that case?

cezarygerard:
I have reviewed this changes in https://github.com/kubernetes/ingress-gce/pull/2565

it looks good to me

please merge these as soon as we have LGMT from Swetha and Gaurav

sawsa307:
Looks good to me, will leave the final approval to @swetharepakula 

/approve

pkg/neg/controller.go
```diff
@@ -597,13 +603,13 @@ func (c *Controller) mergeStandaloneNEGsPortInfo(service *apiv1.Service, name ty
 // mergeVmIpNEGsPortInfo merges the PortInfo for ILB and multinet NetLB services using GCE_VM_IP NEGs into portInfoMap
```
swetharepakula, line 603: nit: update the comments to include NEG default

pkg/neg/controller.go
```diff
@@ -503,7 +508,8 @@ func (c *Controller) processService(key string) error {
 		}
 	}
 
-	if c.runL4 {
+	// Create L4 PortInfo if subsetting is enable `runL4ForILB==true` or RBS NEG support is enabled `runL4ForNetLB==true`.
+	if c.runL4ForILB || c.runL4ForNetLB {
```
swetharepakula, line 512: we check for the variables again in the utils function, do we need the check here still?

pkg/neg/types/types_test.go
```diff
@@ -724,8 +724,8 @@ func TestEndpointsCalculatorMode(t *testing.T) {
 		portInfoMap PortInfoMap
 		expectMode  EndpointsCalculatorMode
 	}{
-		{"L4 Local Mode", NewPortInfoMapForVMIPNEG("testns", "testsvc", testContext.L4Namer, true, defaultNetwork), L4LocalMode},
-		{"L4 Cluster Mode", NewPortInfoMapForVMIPNEG("testns", "testsvc", testContext.L4Namer, false, defaultNetwork), L4ClusterMode},
+		{"L4 Local Mode", NewPortInfoMapForVMIPNEG("testns", "testsvc", testContext.L4Namer, true, defaultNetwork, L4InternalLB), L4LocalMode},
```
swetharepakula, line 727: should we add L4ExternalLB cases here too?

pkg/neg/syncers/endpoints_calculator_test.go
swetharepakula: Yes, unfortunately we have had to manage it by keep the K8s store and the indexer in sync which is not ideal.

Adding the LBType to the syncer is a good option

swetharepakula: I don't see the change to the syncer key. Did you mean the portinfo struct?

pkg/flags/flags.go
```diff
@@ -281,6 +283,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.RunL4NetLBController, "run-l4-netlb-controller", false, `Optional, if enabled then the L4NetLbController will be run.`)
 	flag.BoolVar(&F.EnableNEGController, "enable-neg-controller", true, `Optional, if enabled then the NEG controller will be run.`)
 	flag.BoolVar(&F.EnableL4NEG, "enable-l4-neg", false, `Optional, if enabled then the NEG controller will process L4 NEGs.`)
+	flag.BoolVar(&F.EnableL4NetLBNEG, "enable-l4-netlb-neg", false, `Optional, if enabled then the NetLB controller can create L4 NetLB services with NEG backends.`)
+	flag.BoolVar(&F.EnableL4NetLBNEGDefault, "enable-l4-netlb-neg-default", false, `Optional, if enabled then newly created L4 NetLB services will use NEG backends. Has effect only if '--enable-l4-netlb-neg' is set to true.`)
```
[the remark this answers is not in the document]
mmamczur: not really unfortunately

you can have subsetting turned on for the cluster which relates to ILB only

then there is the new RBS variant for NEGs which we want to gate separately. So, you can have a cluster without subsetting but with RBS services, that's why I separated these 2.

Now the logic is that NEG controller should handle L4 when subsetting is turned on OR when NetLB NEGs are used. I modified the inside of `mergeVmIpNEGsPortInfo` to make sure that these flags are respected (like subsetting turned off -> don't touch ILBs etc)

For the multinet case, this was probably wrong, however, multinet clusters will have use subsetting anyway so it's not a super serious issue. I can try to resolve this multinet case as well but best not to do it in this PR.

pkg/neg/controller.go
```diff
@@ -97,13 +97,16 @@ type Controller struct {
 	// syncerMetrics collects NEG controller metrics
 	syncerMetrics *syncMetrics.SyncerMetrics
 
-	// runL4 indicates whether to run NEG controller that processes L4 services
-	runL4 bool
+	// runL4ForILB indicates whether to run NEG controller that processes L4 ILB services
```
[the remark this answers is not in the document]
mmamczur: yes, exactly, I renamed this to reflect that it is about ILB

multinet should have had it's own setting, I'll have to try figure this out but it was wrong for some rare case when a multinet cluster would not use subsetting.

pkg/neg/controller.go
mmamczur: we can avoid some checks that may be unneeded if we keep it, but it will work if we remove this here. Do you have any preference?

pkg/neg/types/types_test.go
mmamczur: added

swetharepakula (commented):
/lgtm
/approve

PR #2575, opened 2024-05-29
Forward individual ports for NetLB with 5 or less service ports

TheBeeZee:
GCE forwarding rules have a limit of 5 forwarded ports. In the external L4 load balancer, we currently turn a set of exposed ports into a port range from minPort to maxPort and forward all ports in the range, which needlessly forwards traffic for ports that are not supposed to be exposed.

With this change, if 5 or less ports are exposed, we expose each distinct, otherwise we use the old mechanism of exposing a port range. This is similar to have the Internal Load Balancer is setup, except the ILB exposes all ports if more than 5 ports need to be exposed.

To avoid recreating forwarding rules when code logic is updated, we treat port ranges and distinct ports that fall into the same port range as equivalent.

### DIFF
--- a/pkg/loadbalancers/forwarding_rules.go
+++ b/pkg/loadbalancers/forwarding_rules.go
@@ -18,10 +18,11 @@ package loadbalancers
 
 import (
 	"fmt"
-	"k8s.io/klog/v2"
 	"net/http"
 	"strings"
 	"time"
+
+	"k8s.io/klog/v2"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -34,14 +35,15 @@ import (
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/events"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/translator"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 )
 
 const (
-	// maxL4ILBPorts is the maximum number of ports that can be specified in an L4 ILB Forwarding Rule
-	maxL4ILBPorts = 5
+	// maxForwardedPorts is the maximum number of ports that can be specified in an Forwarding Rule
+	maxForwardedPorts = 5
 	// addressAlreadyInUseMessageExternal is the error message string returned by the compute API
 	// when creating an external forwarding rule that uses a conflicting IP address.
 	addressAlreadyInUseMessageExternal = "Specified IP address is in-use and would result in a conflict."
@@ -247,7 +249,7 @@ func (l4 *L4) ensureIPv4ForwardingRule(bsLink string, options gce.ILBOptions, ex
 		AllowGlobalAccess:   options.AllowGlobalAccess,
 		Description:         frDesc,
 	}
-	if len(ports) > maxL4ILBPorts {
+	if len(ports) > maxForwardedPorts {
 		fr.Ports = nil
 		fr.AllPorts = true
 	}
@@ -349,8 +351,10 @@ func (l4netlb *L4NetLB) ensureIPv4ForwardingRule(bsLink string) (*composite.Forw
 		}()
 	}
 
-	portRange, protocol := utils.MinMaxPortRangeAndProtocol(l4netlb.Service.Spec.Ports)
-
+	svcPorts := l4netlb.Service.Spec.Ports
+	ports := utils.GetPorts(svcPorts)
+	portRange := utils.MinMaxPortRange(svcPorts)
+	protocol := utils.GetProtocol(svcPorts)
 	serviceKey := utils.ServiceKeyFunc(l4netlb.Service.Namespace, l4netlb.Service.Name)
 	frDesc, err := utils.MakeL4LBServiceDescription(serviceKey, ipToUse, version, false, utils.XLB)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/utils/utils_test.go
```diff
@@ -245,6 +247,423 @@ func TestEqualResourceIDs(t *testing.T) {
 	}
 }
 
+func TestEqualForwardingRules(t *testing.T) {
+	t.Parallel()
+
+	fwdRules := []*composite.ForwardingRule{
```
cezarygerard, line 253: please switch this slice to well named variables
or
create forwarding rules in each test case

I find it very difficult to understand the test cases using fwdRules[X]

pkg/utils/utils_test.go
```diff
@@ -245,6 +247,423 @@ func TestEqualResourceIDs(t *testing.T) {
...
+			Name:                "fwd-rule-bs-link2-premium-ntier",
+			IPAddress:           "10.0.0.0",
+			Ports:               []string{"1", "2", "3"},
+			IPProtocol:          "TCP",
+			LoadBalancingScheme: string(cloud.SchemeInternal),
+			BackendService:      "http://www.googleapis.com/projects/test/regions/us-central1/backendServices/bs1",
+		},
+	}
+
+	frPortRange1 := &composite.ForwardingRule{
```
cezarygerard, line 341: nit: rename to frPortRange2to3

pkg/utils/utils_test.go
```diff
@@ -245,6 +247,423 @@ func TestEqualResourceIDs(t *testing.T) {
...
+		Name:                "tcp-fwd-rule",
+		IPAddress:           "10.0.0.0",
+		PortRange:           "2-3",
+		IPProtocol:          "TCP",
+		LoadBalancingScheme: string(cloud.SchemeInternal),
+		BackendService:      "http://www.googleapis.com/projects/test/regions/us-central1/backendServices/bs1",
+	}
+	frPortRange2 := &composite.ForwardingRule{
```
cezarygerard, line 349: nit: rename to frPortRange1to2

pkg/utils/utils_test.go
```diff
@@ -245,6 +247,423 @@ func TestEqualResourceIDs(t *testing.T) {
...
+		Name:                "tcp-fwd-rule",
+		IPAddress:           "10.0.0.0",
+		PortRange:           "1-2",
+		IPProtocol:          "TCP",
+		LoadBalancingScheme: string(cloud.SchemeInternal),
+		BackendService:      "http://www.googleapis.com/projects/test/regions/us-central1/backendServices/bs1",
+	}
+	frPortRange3 := &composite.ForwardingRule{
```
cezarygerard, line 357: nit: rename to frPortRange1to3

pkg/utils/utils_test.go
cezarygerard: when I see
			oldFwdRule:  fwdRules[9],


it's not easy to fine the 10th fwd rule

pkg/utils/utils_test.go
```diff
@@ -245,6 +247,423 @@ func TestEqualResourceIDs(t *testing.T) {
...
+			expectEqual: false,
+		},
+		{
+			desc:        "same forwarding rule, different port ranges",
+			oldFwdRule:  frPortRange1,
+			newFwdRule:  frPortRange2,
+			expectEqual: false,
+		},
+		{
+			desc:        "same forwarding rule, ports vs port ranges",
```
cezarygerard, line 421: does this test case make sense? WDYT?
￼			oldFwdRule:    frPortRange3,
￼			newFwdRule:  fwdRules[9],

pkg/utils/utils_test.go
cezarygerard: No I see it was like that originally 🤦

so please just add your new fwd rules as well named variables

cezarygerard (commented):
I looks good overall

I will also try to find some more interesting test cases for port range vs individual ports equality

pkg/loadbalancers/forwarding_rules.go
```diff
@@ -361,20 +363,24 @@ func (l4netlb *L4NetLB) ensureIPv4ForwardingRule(bsLink string) (*composite.Forw
...
+	}
 
 	if existingFwdRule != nil {
 		if existingFwdRule.NetworkTier != fr.NetworkTier {
 			resource := fmt.Sprintf("Forwarding rule (%v)", frName)
 			networkTierMismatchError := utils.NewNetworkTierErr(resource, existingFwdRule.NetworkTier, fr.NetworkTier)
 			return nil, IPAddrUndefined, networkTierMismatchError
 		}
-		equal, err := Equal(existingFwdRule, fr)
+		equal, err := utils.EqualForwardingRules(existingFwdRule, fr)
```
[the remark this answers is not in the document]
TheBeeZee: This function is mostly the same between IPv4 and IPv6. By moving it out to a separate module it allows us to share the common code (or potentially the entire function).

pkg/utils/utils_test.go
TheBeeZee: Done.

pkg/utils/utils_test.go
TheBeeZee: I updated the tests to have the test data inline (where it makes sense).

pkg/utils/utils_test.go
TheBeeZee: It should make sense now that I have updated the test case data.

pkg/utils/utils.go
```diff
@@ -421,6 +422,61 @@ func EqualCloudResourceIDs(a, b *cloud.ResourceID) bool {
...
+	}
+	return equal && fr1.IPAddress == fr2.IPAddress, nil
+}
+
+// EqualForwardingRules returns true if forwarding rules fr1 and fr2 have equal protocol,
+// load balancing scheme, ports or port ranges, resource paths and network tier.
+func EqualIPv6ForwardingRules(fr1, fr2 *composite.ForwardingRule) (bool, error) {
+	return equalForwardingRulesImpl(fr1, fr2)
+}
+
```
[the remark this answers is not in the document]
TheBeeZee: Ports are compared in equalForwardingRulesImpl using the equalPorts helper.

I don't understand why EqualForwardingRules and EqualIPv6ForwardingRules have to be different, but that's the current implementation. The difference is that EqualIPv6ForwardingRules doesn't compare IPv6 addresses. I asked Slavik, who originally implemented it, and he doesn't recall why it was implemented like this.

I can obviously eliminate EqualIPv6ForwardingRules and just use EqualForwardingRules for both, but it seems like this is beyond the scope of this change and without knowing the rationale for the difference, I'd rather not make the change now.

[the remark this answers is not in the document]
cezarygerard: > I can obviously eliminate EqualIPv6ForwardingRules and just use EqualForwardingRules for both, but it seems like this is beyond the scope of this change and without knowing the rationale for the difference, I'd rather not make the change now.

agree

on ipv6 vs ipv4 euqals:
one reason may be that there is a few ways to expres the same ipv6 address? With "::" and using prefixes

cezarygerard:
How we can E2E test this?

pkg/utils/utils.go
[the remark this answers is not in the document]
TheBeeZee: We can fix that by converting the IP to a net.IP and then using net.IP.Equal() to compare them. But we should do that in a follow-up change.

PR #2578, opened 2024-06-03
Update service annotation + IPs in single API call

08volt:
In L4 (ILB and NetLB) controllers after successful sync we do 2 things -- updating .Status.Ingress.LoadBalancerIPs [https://github.com/kubernetes/ingress-gce/blob/05316396ac12cf8e538339bfdfaa30826b08664d/pkg/l4lb/l4controller.go#L281](https://www.google.com/url?q=https://github.com/kubernetes/ingress-gce/blob/05316396ac12cf8e538339bfdfaa30826b08664d/pkg/l4lb/l4controller.go%23L281&sa=D&source=buganizer&usg=AOvVaw1z9JzXQLW6ZYSnKdX-0uZG) and updating service annotations [https://github.com/kubernetes/ingress-gce/blob/05316396ac12cf8e538339bfdfaa30826b08664d/pkg/l4lb/l4controller.go#L290](https://www.google.com/url?q=https://github.com/kubernetes/ingress-gce/blob/05316396ac12cf8e538339bfdfaa30826b08664d/pkg/l4lb/l4controller.go%23L290&sa=D&source=buganizer&usg=AOvVaw36Ga_tyUdTMMcTqWFX6LVG) (links are for L4 ILB, same can be found in L4 NetLB)

Now, both of those are triggering separate API call, to just patch a service. Technically, we could update these things in a single API call

Having separate api calls is also bug-provoking, the order of calls matters, and currently, for example, if service got IPs updated, we can not be sure if Annotation were also updated. This already triggered some failures in e2e tests, but also even in implementation, when we have some tricky situations when webhook + migration + something else is involved

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"math/rand"
 	"reflect"
-	"strings"
 	"sync"
 	"time"
 
@@ -320,41 +319,14 @@ func (l4c *L4Controller) processServiceCreateOrUpdate(service *v1.Service, svcLo
 		syncResult.Error = err
 		return syncResult
 	}
-	err = updateServiceStatus(l4c.ctx, service, syncResult.Status, svcLogger)
+	err = updateServiceInformation(l4c.ctx, l4c.enableDualStack, service, syncResult.Status, syncResult.Annotations, svcLogger)
 	if err != nil {
 		l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncLoadBalancerFailed",
-			"Error updating load balancer status: %v", err)
+			"Error updating Service status and GCP resource annotations: %v", err)
 		syncResult.Error = err
 		return syncResult
 	}
-	if l4c.enableDualStack {
-		l4c.emitEnsuredDualStackEvent(service)
-		if err = updateL4DualStackResourcesAnnotations(l4c.ctx, service, syncResult.Annotations, svcLogger); err != nil {
-			l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncLoadBalancerFailed",
-				"Failed to update Dual Stack annotations for load balancer, err: %v", err)
-			syncResult.Error = fmt.Errorf("failed to set Dual Stack resource annotations, err: %w", err)
-			return syncResult
-		}
-	} else {
-		l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeNormal, "SyncLoadBalancerSuccessful",
-			"Successfully ensured load balancer resources")
-		if err = updateL4ResourcesAnnotations(l4c.ctx, service, syncResult.Annotations, svcLogger); err != nil {
-			l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeWarning, "SyncLoadBalancerFailed",
-				"Failed to update annotations for load balancer, err: %v", err)
-			syncResult.Error = fmt.Errorf("failed to set resource annotations, err: %w", err)
-			return syncResult
-		}
-	}
 	return syncResult
-}
-
-func (l4c *L4Controller) emitEnsuredDualStackEvent(service *v1.Service) {
-	var ipFamilies []string
-	for _, ipFamily := range service.Spec.IPFamilies {
-		ipFamilies = append(ipFamilies, string(ipFamily))
-	}
-	l4c.ctx.Recorder(service.Namespace).Eventf(service, v1.EventTypeNormal, "SyncLoadBalancerSuccessful",
-		"Successfully ensured %v load balancer resources", strings.Join(ipFamilies, " "))
 }
 
 func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service, svcLogger klog.Logger) *loadbalancers.L4ILBSyncResult {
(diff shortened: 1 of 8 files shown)


pkg/l4lb/l4lbcommon.go
```diff
@@ -66,46 +70,42 @@ func mergeAnnotations(existing, lbAnnotations map[string]string, keysToRemove []
...
-func updateServiceStatus(ctx *context.ControllerContext, svc *v1.Service, newStatus *v1.LoadBalancerStatus, svcLogger klog.Logger) error {
-	svcLogger.V(2).Info("Updating service status", "newStatus", fmt.Sprintf("%+v", newStatus))
-	if helpers.LoadBalancerStatusEqual(&svc.Status.LoadBalancer, newStatus) {
-		svcLogger.V(2).Info("New and old statuses are equal, skipping patch")
+func updateServiceInformation(ctx *context.ControllerContext, enableDualStack bool, svc *v1.Service, newStatus *v1.LoadBalancerStatus, newL4LBAnnotations map[string]string, svcLogger klog.Logger) error {
+	emitIpFamiliesStackEvent(ctx, svc)
+	newObjectMeta := computeNewAnnotationsIfNeeded(svc, newL4LBAnnotations, enableDualStack)
+
+	svcLogger.V(2).Info("Updating service information", "newStatus", fmt.Sprintf("%+v", newStatus), "newObjectMeta", newObjectMeta)
+	if helpers.LoadBalancerStatusEqual(&svc.Status.LoadBalancer, newStatus) && reflect.DeepEqual(svc.ObjectMeta.Annotations, newObjectMeta.Annotations) {
```
aojea, line 89: this is a very tricky change, you have ` reflect.DeepEqual(svc.ObjectMeta.Annotations, newObjectMeta.Annotations)`

you are passing **svc**, the entire object, and **newL4LBAnnotations** , here comparing the existing annotations with the new inside the function

It is mixing too many things in the same function.

Is newL4LBAnnotations the entire object or only the loadbalancer annotations, per example, what if user has annotation-foo unrelated to this loadbalancer

pkg/context/context.go
```diff
@@ -198,7 +198,7 @@ func NewControllerContext(
 		PodInformer:             podInformer,
 		NodeInformer:            nodeInformer,
 		SvcNegInformer:          informersvcneg.NewServiceNetworkEndpointGroupInformer(svcnegClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer()),
-		recorders:               map[string]record.EventRecorder{},
+		Recorders:               map[string]record.EventRecorder{},
```
cezarygerard, line 201: this Recorders change is an overkill, please find other way to unit test the change

cezarygerard: one way will be to drop assertions on events

it would be enough to read service annotation from the test fake

pkg/l4lb/l4controller.go
```diff
@@ -384,28 +356,12 @@ func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service, svc
 	// Reset the loadbalancer status first, before resetting annotations.
 	// Other controllers(like service-controller) will process the service update if annotations change, but will ignore a service status change.
 	// Following this order avoids a race condition when a service is changed from LoadBalancer type Internal to External.
-	if err := updateServiceStatus(l4c.ctx, svc, &v1.LoadBalancerStatus{}, svcLogger); err != nil {
+	if err := updateServiceInformation(l4c.ctx, l4c.enableDualStack, svc, &v1.LoadBalancerStatus{}, nil, svcLogger); err != nil {
```
cezarygerard, line 359: use named constant istead of nil here

pkg/l4lb/l4lbcommon.go
cezarygerard: Enrico,


Don't use deep equal here

in private func check of annotations require update - check if existing annotations contain all the new annotation

pkg/l4lb/l4lbcommon_test.go
```diff
@@ -141,3 +174,106 @@ func TestFinalizerWasRemovedUnexpectedly(t *testing.T) {
...
+				"networking.gke.io/load-balancer-type":     "Internal",
+				"service.kubernetes.io/firewall-rule-ipv6": "ok",
+			},
+		},
+	}
+
+	logger := klog.TODO()
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
```
cezarygerard, line 259: add t.Parallel()

pkg/l4lb/l4lbcommon_test.go
```diff
@@ -141,3 +174,106 @@ func TestFinalizerWasRemovedUnexpectedly(t *testing.T) {
...
+func TestUpdateServiceInformation(t *testing.T) {
+	testCases := []struct {
+		desc                string
+		enableDualStack     bool
+		newStatus           *v1.LoadBalancerStatus
+		newAnnotations      map[string]string
+		expectedAnnotations map[string]string
+	}{
+		{
+			desc:            "Add annotations",
```
cezarygerard, line 187: add case with overriding the annotation

pkg/l4lb/l4lbcommon_test.go
```diff
@@ -141,3 +174,106 @@ func TestFinalizerWasRemovedUnexpectedly(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdateServiceInformation(t *testing.T) {
+	testCases := []struct {
+		desc                string
+		enableDualStack     bool
+		newStatus           *v1.LoadBalancerStatus
```
cezarygerard, line 182: add fields for 
oldStatus
oldAnnotations

aojea:
FYI the kubernetes project is now trying to avoid this behavior for new APIs, to not modify metadata and status at the same time https://github.com/search?q=repo%3Akubernetes%2Fkubernetes%20ResetObjectMetaForStatus&type=code

it means this behavior is considered a bad practice

cezarygerard:
/hold
we should probably abandon this, 
after Antonio's last comment

08volt:
> /hold we should probably abandon this, after Antonio's last comment

PR #2582, opened 2024-06-05
Fix instance group sizes

cezarygerard:
the is an error in truncating to 1000 nodes before iterating over zones when adding/removing instances from instance group

plus unit tests for multizonal case

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -190,34 +190,6 @@ func (m *manager) DeleteInstanceGroup(name string) error {
 	return fmt.Errorf("%v", errs)
 }
 
-// listIGInstances lists all instances of provided instance group name in all zones.
-// The return format will be a set of nodes in the instance group and
-// a map from node name to zone.
-func (m *manager) listIGInstances(name string) (sets.String, map[string]string, error) {
-	nodeNames := sets.NewString()
-	nodeZoneMap := make(map[string]string)
-	zones, err := m.ZoneGetter.ListZones(zonegetter.AllNodesFilter, m.logger)
-	if err != nil {
-		return nodeNames, nodeZoneMap, err
-	}
-
-	for _, zone := range zones {
-		instances, err := m.cloud.ListInstancesInInstanceGroup(name, zone, allInstances)
-		if err != nil {
-			return nodeNames, nodeZoneMap, err
-		}
-		for _, ins := range instances {
-			name, err := utils.KeyName(ins.Instance)
-			if err != nil {
-				return nodeNames, nodeZoneMap, err
-			}
-			nodeNames.Insert(name)
-			nodeZoneMap[name] = zone
-		}
-	}
-	return nodeNames, nodeZoneMap, nil
-}
-
 // Get returns the Instance Group by name.
 func (m *manager) Get(name, zone string) (*compute.InstanceGroup, error) {
 	ig, err := m.cloud.GetInstanceGroup(name, zone)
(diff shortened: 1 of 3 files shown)


pkg/instancegroups/manager.go
```diff
@@ -359,46 +299,27 @@ func (m *manager) Sync(nodes []string) (err error) {
...
-	pool, err := m.List()
-	if err != nil {
-		m.logger.Error(err, "List error")
-		return err
-	}
+	zonedNodes := m.splitNodesByZone(nodes)
+	for zone, kubeNodesFromZone := range zonedNodes {
+		igName := m.namer.InstanceGroup()
+		if len(kubeNodesFromZone) > m.maxIGSize {
+			m.logger.Info(fmt.Sprintf("Total number of kubeNodes: %d, truncating to maximum Instance Group size = %d. zone: %s. First truncated instances: %v", len(kubeNodesFromZone), m.maxIGSize, zone, nodes[m.maxIGSize:]))
```
mmamczur, line 306: shouldn't 'First truncated instances' be from kubeNodesFrom zone instead of nodes?
also, before there was this function to truncate. This `nodes[m.maxIGSize:]` seems like it can output all of them

pkg/instancegroups/manager.go
```diff
@@ -359,46 +299,27 @@ func (m *manager) Sync(nodes []string) (err error) {
...
-	for _, igName := range pool {
-		// Keep the zone information for each node in this map.
-		// This will be used as a reference to get zone information
-		// when removing nodes.
-		gceNodes, gceNodeZoneMap, err := m.listIGInstances(igName)
+		gceNodes := sets.NewString()
+		instances, err := m.cloud.ListInstancesInInstanceGroup(igName, zone, allInstances)
 		if err != nil {
-			m.logger.Error(err, "listIGInstances error", "name", igName)
+			m.logger.Error(err, "TODO error", "name", igName)
```
mmamczur, line 315: TODO?

pkg/instancegroups/manager.go
```diff
@@ -359,46 +299,27 @@ func (m *manager) Sync(nodes []string) (err error) {
...
-				if len(nodes) <= maxLogsSampleSize {
-					return nodes
-				}
-				return nodes[:maxLogsSampleSize]
-			}
-
-			m.logger.Info(fmt.Sprintf("Total number of kubeNodes: %d, truncating to maximum Instance Group size = %d. Instance group name: %s. First truncated instances: %v", len(kubeNodesList), m.maxIGSize, igName, truncateForLogs(nodes[m.maxIGSize:])))
-			kubeNodes = sets.NewString(kubeNodesList[:m.maxIGSize]...)
+		for _, ins := range instances {
+			parts := strings.Split(ins.Instance, "/")
```
mmamczur, line 319: why not keep using that KeyName util?

or name some function so that it's clear that this gets the instance name from the url

pkg/instancegroups/manager.go
cezarygerard: fixed

pkg/instancegroups/manager.go
cezarygerard: done

pkg/instancegroups/manager.go
cezarygerard: I should us KeyName

pkg/instancegroups/manager.go
mmamczur: but do you intend to output all or a truncated list? this could log hundreds of node names

bowei:
Please fix your commit message (typos). Would be good to give a bit more context on the change:

> "the is an error in truncating to 1000 nodes before iterating over zones when adding/removing instances from instance group"

* What is the error in the original code path? What this bug being hit frequently in production but not noticed or an edge condition that we finally saw?
* Under what conditions will we encounter this error?
* How will the result change between the old behavior and the new behavior?
* What kind of testing did we add to make sure the behavior did not change other than what we want to change? 

Do we want to add some more unit tests? It looks like a medium sized change with potentially large impact.

mmamczur:
@bowei 
Adding some context on the bug

This is the code prior to this change:
```
		// Individual InstanceGroup has a limit for 1000 instances in it.
		// As a result, it's not possible to add more to it.
		if len(kubeNodes) > m.maxIGSize {
```
The intent seemed to have limit of 1k per zone, not in total. Looking at the comment and variable name that someone meant to have a limit per IG, not per zone.
Also, to add to that, there is code in the legacy ILB controller that also syncs the same IG (it's shared). If there is a cluster that for some reason uses both, both controllers will resize the instance groups, legacy making them 1k per zone while this one will make it 1k total across zones. Taking into account that the ingress-gce was initially based on the legacy one this was a bug.

The issue is not hit frequently in prod since you need a large cluster with ILB subsetting to see it. Conditions:
- cluster with above 1k nodes
- ILB subsetting service with etp: Local
- be unlucky so that all pods are on nodes NOT added to the IG

Result: no pods get traffic

This could still be a problem for with more than 1k nodes in one zone.

But even excluding the customer issue, the difference between legacy and ingress-gce + the intent from the comments warrant this fix

cezarygerard:
> What is the error in the original code path? What this bug being hit frequently in production but not noticed or an edge condition that we finally saw?

Original code path. Occurs only at certain scale, only noticeable with services with ExternalTrafficPolicy: Local

> Under what conditions will we encounter this error?

1K+ nodes, ExternalTrafficPolicy: Local, pods added only on trimmed nodes.

> How will the result change between the old behavior and the new behavior?

we will limit to 1k nodes on per zone basis, exactly as we do in cloud-controller-gcp code managing instance groups

> What kind of testing did we add to make sure the behavior did not change other than what we want to change?

- unit tests for IG management that only passes with the new code
- all the remaining unit tests are passing unchanged
- we have robous E2E tests that mixes all possible service types, employs all our controllers

cezarygerard:
> Please fix your commit message (typos). Would be good to give a bit more context on the change:

added more context

but where is a typo in the original message:
"the is an error in truncating to 1000 nodes before iterating over zones when adding/removing instances from instance group"

?

bowei:
1) I think my "typo" comment threw off the main point of my comment which is to take the above detailed explanations for the fixand put them in the commit message to describe what the change was. I'm not sure it's obvious what the context of the commit is apart from reading the discussion on Github.

2) If the fix references an already existing implementation, please include a link or reference to that implementation in the commit message.

3) How was this tested? Did we do an integration test up to the IG size limit where the controller will start behaving differently?

pkg/instancegroups/manager_test.go
```diff
@@ -161,6 +164,91 @@ func TestNodePoolSync(t *testing.T) {
...
+	testCases := []struct {
+		description    string
+		gceNodesZoneA  sets.String
+		gceNodesZoneB  sets.String
+		gceNodesZoneC  sets.String
+		kubeNodesZoneA sets.String
+		kubeNodesZoneB sets.String
+		kubeNodesZoneC sets.String
+	}{
+		{
```
bowei, line 180: Can we add test cases for 1 zone, 2 zones?

pkg/instancegroups/manager.go
```diff
@@ -288,64 +260,32 @@ func (m *manager) getInstanceReferences(zone string, nodeNames []string) (refs [
...
-		zone, ok := nodeZoneMap[name]
-		if !ok {
-			m.logger.Error(nil, "Failed to get zones for node, skipping", "name", name)
-			continue
-		}
-		if _, ok := nodesByZone[zone]; !ok {
-			nodesByZone[zone] = []string{}
-		}
-		nodesByZone[zone] = append(nodesByZone[zone], name)
-	}
```
swetharepakula, line 329: This code was added because of a bug found: https://github.com/kubernetes/ingress-gce/commit/080faa4d77b060d3ec99a48ddfa766ae676feb72

pkg/instancegroups/manager.go
```diff
@@ -288,64 +260,32 @@ func (m *manager) getInstanceReferences(zone string, nodeNames []string) (refs [
...
 	}
-
-	err := fmt.Errorf("RemoveInstances: %v", errs)
-	events.GlobalEventf(m.recorder, core.EventTypeWarning, events.RemoveNodes, "Error removing nodes %s from InstanceGroup %q: %v", events.TruncatedStringList(names), groupName, err)
-	return err
+	return nil
 }
 
 // Sync nodes with the instances in the instance group.
 func (m *manager) Sync(nodes []string) (err error) {
```
swetharepakula, line 287: Based on cloud-provider-gcp, the node selection is done by alphabetical order: https://github.com/kubernetes/cloud-provider-gcp/blob/fca628cb3bf9267def0abb509eaae87d2d4040f3/providers/gce/gce_loadbalancer_internal.go#L620-L622

In ingress-gce the nodes are not ordered: https://github.com/kubernetes/ingress-gce/blob/master/pkg/instancegroups/controller.go#L132

Will this be a problem?

pkg/instancegroups/manager_test.go
```diff
@@ -210,6 +213,151 @@ func (fakeIG *fakeIGAlreadyExists) AddInstancesToInstanceGroup(name, zone string
...
+			description:    "too many kube nodes in one 1 of 3 zone",
```
bowei, line 230: Can we have a test case that checks for alphabetical ordering to match the other implementation?

cezarygerard: we already have it,  the next test func verifies exactly that "TestInstanceTruncatingOrder":

pkg/instancegroups/manager_test.go
```diff
@@ -210,6 +213,151 @@ func (fakeIG *fakeIGAlreadyExists) AddInstancesToInstanceGroup(name, zone string
...
+			ports := []int64{80}
+			_, err := pool.EnsureInstanceGroupsAndPorts(igName, ports)
+			if err != nil {
+				t.Fatalf("pool.EnsureInstanceGroupsAndPorts(%s, %v) returned error %v, want nil", igName, ports, err)
+			}
+			allKubeNodes := append(tc.kubeNodesZoneA.List(), tc.kubeNodesZoneB.List()...)
+			allKubeNodes = append(allKubeNodes, tc.kubeNodesZoneC.List()...)
+
+			// When
+			// Execute manager's main instance group sync function
```
swetharepakula, line 279: nit: the comment seems incomplete

pkg/instancegroups/manager_test.go
```diff
@@ -210,6 +213,151 @@ func (fakeIG *fakeIGAlreadyExists) AddInstancesToInstanceGroup(name, zone string
...
+	}
+
+	// When
+	// Execute manager's main instance group sync function
+	err = pool.Sync(kubeNodesZoneA)
+	if err != nil {
+		t.Fatalf("pool.Sync(_) returned error %v, want nil", err)
+	}
+
+	//then
```
swetharepakula, line 340: nit: remove comment?

pkg/instancegroups/manager_test.go
```diff
@@ -210,6 +213,151 @@ func (fakeIG *fakeIGAlreadyExists) AddInstancesToInstanceGroup(name, zone string
...
+	manager := pool.(*manager)
+	zonegetter.AddFakeNodes(manager.ZoneGetter, testZoneA, gceNodesZoneA...)
+
+	ports := []int64{80}
+	_, err := pool.EnsureInstanceGroupsAndPorts(igName, ports)
+	if err != nil {
+		t.Fatalf("pool.EnsureInstanceGroupsAndPorts(%s, %v) returned error %v, want nil", igName, ports, err)
+	}
+
+	// When
```
swetharepakula, line 333: nit: comment seems incomplete

pkg/instancegroups/manager.go
swetharepakula: The code has changed to determining the zones and then traversing the zones, so when removing nodes we do not depend on the K8s node resource for the zone and therefore do not run into the bug from above.

pkg/instancegroups/manager.go
```diff
@@ -364,65 +300,51 @@ func (m *manager) Sync(nodes []string) (err error) {
...
 				return err
 			}
 		}
 
 		start = time.Now()
 		if len(addNodes) != 0 {
-			err = m.add(igName, addNodes)
-			m.logger.V(2).Info("Add finished", "name", igName, "err", err, "timeTaken", time.Now().Sub(start), "addNodes", addNodes)
+			err = m.add(igName, addNodes, zone)
+			m.logger.V(2).Info("Add finished", "name", igName, "err", err, "timeTaken", time.Now().Sub(start), "addNodes", events.TruncatedStringList(addNodes))
```
swetharepakula, line 347: can we add a metric here to track how often we are modifying the instance group?

At a cluster level, we can get some data on how much churn is occurring.

pkg/instancegroups/manager.go
swetharepakula: Can you add a comment with a link to the code in cloud-provider-gcp?

pkg/instancegroups/manager.go
cezarygerard: how you want to define metric?
- cumulative count for number of times we added(removed?) any number of nodes?
- gauges showing number nodes added/removed in latest sync?
- gauges showing nodes  nodes added/removed in latest sync if len(addNodes) != 0 ||len(removeNodes)
- all of them?

swetharepakula: For metrics I was thinking perhaps 2:

* A histogram that tracks how many we are adding or removing. This gives us some idea of the average churn? However, a cumulative counter though could work as well so we can see at any given time how many we are adding or removing by doing a diff.

* A counter that tracks the number of adds and removes we are attempting. This lets us know how many times our logic feels the need to make changes to the instance group.

Together I think that should give us some more insight into the management of the instance group

bowei:
Can the metrics change be done as a follow on? Or do we need to have this go in in one commit?

cezarygerard:
> Can the metrics change be done as a follow on? Or do we need to have this go in in one commit?

yes, please.

Added comment with link to cloud-provider-gcp's Legacy L4 ILB Controller code


Internally we have run manual tests to check consistency between legacy cloud-provider-gcp's Legacy L4 ILB Controller and this new logic and they look good

cezarygerard:
I have also added internal ticket to follow up on metrics, linking to this PR and Swetha's comments

pkg/instancegroups/manager.go
swetharepakula: Discussed offline, this will be added as a followup

swetharepakula (commented):
/lgtm

PR #2583, opened 2024-06-11
Update readiness check for non-default subnet nodes.

sawsa307:
* For pods in non-default subnet, it is possible for them to share the same labels with the pods in default subnets. As a result, the list of neg will not be empty in the NEG lookup, so they will not be marked as ready for not in default subnet.


swetharepakula (commented):
/lgtm
/approve

PR #2585, opened 2024-06-17
Add panic recovery with error messages in L4Controller

FelipeYepez:
- Add panic recovery to syncWrapper, AddFunc, and UpdateFunc
- Log detailed error messages
- Emit metric on panic for monitoring

/assign @mmamczur @cezarygerard

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -408,19 +408,19 @@ func (l4c *L4Controller) linkNEG(l4 *loadbalancers.L4, svcLogger klog.Logger) er
 	return l4c.NegLinker.Link(l4.ServicePort, groupKeys)
 }
 
-func (l4c *L4Controller) syncWrapper(key string) error {
+func (l4c *L4Controller) syncWrapper(key string) (err error) {
 	syncTrackingId := rand.Int31()
 	svcLogger := l4c.logger.WithValues("serviceKey", key, "syncId", syncTrackingId)
 
-	var syncErr error
 	defer func() {
 		if r := recover(); r != nil {
 			errMessage := fmt.Sprintf("Panic in L4 ILB sync worker goroutine: %v", r)
 			svcLogger.Error(nil, errMessage)
 			l4metrics.PublishL4ControllerPanicCount(l4ILBControllerName)
+			err = fmt.Errorf(errMessage)
 		}
 	}()
-	syncErr = l4c.sync(key, svcLogger)
+	syncErr := l4c.sync(key, svcLogger)
 	return skipUserError(syncErr, svcLogger)
 }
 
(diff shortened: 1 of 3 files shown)


aojea:
why there should be a panic in an event handler?

pkg/l4lb/l4controller.go
```diff
@@ -120,8 +120,15 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}, lo
 		AddFunc: func(obj interface{}) {
 			addSvc := obj.(*v1.Service)
 			svcKey := utils.ServiceKeyFunc(addSvc.Namespace, addSvc.Name)
-			needsILB, svcType := annotations.WantsL4ILB(addSvc)
 			svcLogger := logger.WithValues("serviceKey", svcKey)
+			defer func() {
```
aojea, line 124: @cezarygerard talked with @FelipeYepez offline, I don't think we should use recovers on handlers and be paranoid, handlers must only do basic filtering and enqueu the event

/hold

pkg/l4lb/l4controller.go
```diff
@@ -411,7 +425,17 @@ func (l4c *L4Controller) linkNEG(l4 *loadbalancers.L4, svcLogger klog.Logger) er
 func (l4c *L4Controller) syncWrapper(key string) error {
 	syncTrackingId := rand.Int31()
 	svcLogger := l4c.logger.WithValues("serviceKey", key, "syncId", syncTrackingId)
-	return skipUserError(l4c.sync(key, svcLogger), svcLogger)
+
+	var syncErr error
+	defer func() {
+		if r := recover(); r != nil {
+			errMessage := fmt.Sprintf("Panic in L4 ILB sync worker goroutine: %v", r)
```
aojea, line 432: I do not object to have this here, though , I don't think is a good practice

FelipeYepez: Thank you Antonio, after our conversation I'll leave the recover function for the syncWrapper.

pkg/l4lb/l4controller.go
FelipeYepez: Thank you Antonio for your insights, removed this recover functions from here

aojea:
/ok-to-test
/hold cancel
/lgtm

Thanks

pkg/l4lb/l4controller.go
```diff
@@ -411,7 +411,17 @@ func (l4c *L4Controller) linkNEG(l4 *loadbalancers.L4, svcLogger klog.Logger) er
...
+			svcLogger.Error(nil, errMessage)
+			l4metrics.PublishL4ControllerPanicCount(l4ILBControllerName)
+		}
+	}()
+	syncErr = l4c.sync(key, svcLogger)
+	return skipUserError(syncErr, svcLogger)
```
mmamczur, line 424: as discussed offline, do we know what the syncWrapper will return if recover happens?

aojea:
/hold

there are some comments form Michal, also I didn't approve it 🤔 , I was just waiting for someone else to do it

pkg/l4lb/l4controller.go
FelipeYepez: Thank you Michal. The recover function will now return an error that will requeue the task.

PR #2588, opened 2024-06-21
Remove deprecated metrics

FelipeYepez:
revert https://github.com/kubernetes/ingress-gce/pull/1943/


pkg/flags/flags.go
```diff
@@ -280,7 +279,6 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableL4NEG, "enable-l4-neg", false, `Optional, if enabled then the NEG controller will process L4 NEGs.`)
 	flag.BoolVar(&F.GateNEGByLock, "gate-neg-by-lock", false, "If enabled then the NEG controller will be run via leader election with NEG resource lock")
 	flag.BoolVar(&F.EnableIGController, "enable-ig-controller", true, `Optional, if enabled then the IG controller will be run.`)
-	flag.BoolVar(&F.EnableServiceMetrics, "enable-service-metrics", false, `Optional, if enabled then the service metrics controller will be run.`)
```
mmamczur, line 283: make sure this is not used in the new component configs. I can see already that our nightly image specifies that flag o need to remove it from there before you remove it in here.

mmamczur:
make sure you try running the controller with these changes. Keep it on for more than 10mins and see if nothing bad is happening

pkg/flags/flags.go
FelipeYepez: Thank you, removing it.

PR #2599, opened 2024-07-08
Skip marking NEG readiness gate as ready until nodeName is set for msc, and add new error state for Node without PodCIDR

sawsa307:
* If nodeName is empty on the Pod, it means the Pod hasn't been scheduled. We will mark it as unready, and when the Pod is scheduled, it will trigger another Pod update.
* If we ever encounter a persistent issue when checking node subnet during pod readiness, we mark the pod as ready if the timeout is reached.
* An endpoint corresponds to a node without PodCIDR should be considered as an error-state condition. A new error-state error is added to ensure we can correctly trigger this, and we will exclude this endpoint in degraded mode.

### DIFF
--- a/pkg/neg/readiness/reflector.go
+++ b/pkg/neg/readiness/reflector.go
@@ -204,13 +204,6 @@ func (r *readinessReflector) getExpectedNegCondition(pod *v1.Pod, neg, backendSe
 		return expectedCondition
 	}
 
-	if r.enableMultiSubnetCluster && !r.zoneGetter.IsDefaultSubnetNode(pod.Spec.NodeName, r.logger) {
-		expectedCondition.Status = v1.ConditionTrue
-		expectedCondition.Reason = negReadyReason
-		expectedCondition.Message = fmt.Sprintf("Pod belongs to a node in non-default subnet. Marking condition %q to True.", shared.NegReadinessGate)
-		return expectedCondition
-	}
-
 	negs := r.lookup.ReadinessGateEnabledNegs(pod.Namespace, pod.Labels)
 	// mark pod as ready if it belongs to no NEGs
 	if len(negs) == 0 {
@@ -222,10 +215,33 @@ func (r *readinessReflector) getExpectedNegCondition(pod *v1.Pod, neg, backendSe
 
 	// check if the pod has been waiting for the endpoint to show up as Healthy in NEG for too long
 	if r.clock.Now().After(pod.CreationTimestamp.Add(unreadyTimeout)) {
+		r.logger.Info(pod.Name, "now", r.clock.Now(), "pod time", pod.CreationTimestamp)
 		expectedCondition.Status = v1.ConditionTrue
 		expectedCondition.Reason = negReadyTimedOutReason
 		expectedCondition.Message = fmt.Sprintf("Timeout waiting for pod to become healthy in at least one of the NEG(s): %v. Marking condition %q to True.", negs, shared.NegReadinessGate)
 		return expectedCondition
+	}
+
+	if r.enableMultiSubnetCluster {
+		if pod.Spec.NodeName == "" {
+			r.logger.Error(nil, "Unable to determine the pod's node name.", "podNamespace", pod.Namespace, "podName", pod.Name)
+			expectedCondition.Reason = negNotReadyReason
+			expectedCondition.Message = "Unable to determine the pod's node name."
+			return expectedCondition
+		}
+		isInDefaultSubnet, err := r.zoneGetter.IsDefaultSubnetNode(pod.Spec.NodeName, r.logger)
+		if err != nil {
+			r.logger.Error(err, "Unable to determine if the pod is in the default subnet.", "podNamespace", pod.Namespace, "podName", pod.Name)
+			expectedCondition.Reason = negNotReadyReason
+			expectedCondition.Message = "Unable to determine if the pod is in the default subnet."
+			return expectedCondition
+		}
+		if !isInDefaultSubnet {
+			expectedCondition.Status = v1.ConditionTrue
+			expectedCondition.Reason = negReadyReason
+			expectedCondition.Message = fmt.Sprintf("Pod belongs to a node in non-default subnet. Marking condition %q to True.", shared.NegReadinessGate)
+			return expectedCondition
+		}
 	}
 
 	// do not patch condition status in this case to prevent race condition:
--- a/pkg/neg/readiness/reflector_test.go
+++ b/pkg/neg/readiness/reflector_test.go
@@ -78,16 +78,18 @@ func newTestReadinessReflector(testContext *negtypes.TestContext, enableMultiSub
 func TestSyncPod(t *testing.T) {
 	t.Parallel()
 	fakeContext := negtypes.NewTestContext()
-	fakeClock := clocktesting.NewFakeClock(time.Now())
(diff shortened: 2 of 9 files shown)


pkg/neg/readiness/reflector.go
```diff
@@ -204,11 +204,28 @@ func (r *readinessReflector) getExpectedNegCondition(pod *v1.Pod, neg, backendSe
...
+		// trigger another Pod update.
+		if pod.Spec.NodeName == "" {
+			expectedCondition.Message = "Waiting for Pod to have a non-empty node name."
+			expectedCondition.Reason = negNotReadyReason
+			return expectedCondition
+		}
+		isInDefaultSubnet, err := r.zoneGetter.IsDefaultSubnetNode(pod.Spec.NodeName, r.logger)
+		if err != nil {
+			r.logger.Error(err, "Failed to verify if the node is in the default subnet.")
+			expectedCondition.Message = "Failed to verify if the node is in the default subnet."
```
swetharepakula, line 219: if by chance the nodes are stuck, or a repeated error occurs here, we shouldn't get stuck.

We should make forward progress -- can we make it so at 15 minutes, this error is also ignored and the pod is marked as ready?

pkg/neg/readiness/reflector_test.go
```diff
@@ -455,9 +455,43 @@ func TestSyncPodMultipleSubnets(t *testing.T) {
...
-							Message: fmt.Sprintf("Pod belongs to a node in non-default subnet. Marking condition %q to True.", shared.NegReadinessGate),
+							Reason:  negNotReadyReason,
+							Message: fmt.Sprintf("Failed to verify if the node is in the default subnet."),
+						},
+					},
+				},
+			},
+		},
+		{
+			desc: "need to update pod: pod belongs to a node without nodeName",
```
swetharepakula, line 466: nit: maybe pod does not have nodeName specified?

swetharepakula (commented):
Since we are adding error cases, we need to make sure pod readiness does not get stuck if the error occurs.

For degraded mode, does the error trigger error state?

sawsa307:
> For degraded mode, does the error trigger error state?

Yes. During endpoint calculation, if we find an endpoint with nodeName == "" or nodeName = nil, we will return the error ErrEPNodeMissing. If we encounter an issue when trying to get pod from zoneGetter, we will return ErrNodeNotFound. And in both of these situation, if we are not currently in degraded mode, this error will trigger degraded mode. If we are in degraded mode, we will skip these endpoints in calculation. So the only concern would be on pod readiness side.

pkg/neg/readiness/reflector_test.go
sawsa307: Updated

pkg/neg/readiness/reflector.go
```diff
@@ -204,11 +204,37 @@ func (r *readinessReflector) getExpectedNegCondition(pod *v1.Pod, neg, backendSe
...
+		if pod.Spec.NodeName == "" {
+			expectedCondition.Message = "Waiting for Pod to have a non-empty node name."
+			expectedCondition.Reason = negNotReadyReason
+			return expectedCondition
+		}
+		isInDefaultSubnet, err := r.zoneGetter.IsDefaultSubnetNode(pod.Spec.NodeName, r.logger)
+		if err != nil {
+			// If by chance the nodes are stuck, or a repeated error occurs here,
+			// we unblock the pod when the timeout is reached.
+			if r.clock.Now().After(pod.CreationTimestamp.Add(unreadyTimeout)) {
```
swetharepakula, line 220: should we instead move the unready timeout check to above so it is taken care of in one place instead of adding another timeout check?

swetharepakula (commented):
We should try to simplify the readiness condition check. I think consolidating the conditions where it makes sense e.g. only check for the timeout once

pkg/neg/readiness/reflector.go
sawsa307: Updated.

pkg/neg/syncers/utils_test.go
```diff
@@ -2626,7 +2626,7 @@ func TestValidateEndpointFieldsMultipleSubnets(t *testing.T) {
 			},
 		},
 		{
-			desc: "include one endpoint that corresponds to a node without PodCIDR, endpoint should be excluded in both calculations",
+			desc: "include one endpoint that corresponds to a node without PodCIDR, endpoint should be excluded in degraded mode",
```
swetharepakula, line 2629: Offline we did discuss this as only for degraded mode, however I am now thinking we should exclude in both calculations.

I am not sure this state is even possible because if the pod has an endpoint it comes from the Node's PodCIDR.

pkg/neg/readiness/reflector_test.go
```diff
@@ -225,8 +225,9 @@ func TestSyncPod(t *testing.T) {
 		},
 		{
 			desc: "need to update pod: there is NEGs associated but pod is not healthy",
-			mutateState: func(testlookUp *fakeLookUp) {
+			mutateState: func(testlookUp *fakeLookUp, fakeClock *clocktesting.FakeClock) {
```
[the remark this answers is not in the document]
sawsa307: We need to pass in the clock because we need a more fine-grained control over the timeout. Previously we were using the same clock, and this creates an implicit dependency between each test case, which is not ideal(later test cases will always timeout because we have stepped the clock forward multiple times).

pkg/neg/syncers/utils_test.go
sawsa307: I have a different opinion here.

We explicitly exclude endpoints from non-default subnet because we explicitly raise an error for any nodes not from default subnet when we try to get its zone. This is not an error case on a node, but a condition we try to prevent so we won't include them in the endpoint calculation, so it makes sense we skip this endpoint.

However, for a node without PodCIDR, I think it is a valid error case. We depend on the `PodCIDR being set` as a signal to know if the node is ready by the node controller, and I agree that we should never run into cases that we have an endpoint but the PodCIDR is not set, but that's how we have been handle error cases currently, so it would be more consistent if we skip the node in degraded mode.

pkg/neg/readiness/reflector.go
```diff
@@ -222,12 +215,35 @@ func (r *readinessReflector) getExpectedNegCondition(pod *v1.Pod, neg, backendSe
...
+		r.logger.Info(pod.Name, "now", r.clock.Now(), "pod time", pod.CreationTimestamp)
 		expectedCondition.Status = v1.ConditionTrue
 		expectedCondition.Reason = negReadyTimedOutReason
 		expectedCondition.Message = fmt.Sprintf("Timeout waiting for pod to become healthy in at least one of the NEG(s): %v. Marking condition %q to True.", negs, shared.NegReadinessGate)
 		return expectedCondition
 	}
 
+	if r.enableMultiSubnetCluster {
+		if pod.Spec.NodeName == "" {
+			r.logger.Error(nil, "Unable to determine the pod's node name.")
```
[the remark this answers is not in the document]
sawsa307: Added

pkg/neg/readiness/reflector_test.go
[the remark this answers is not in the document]
swetharepakula: I am thinking we do not need to step forward in the clock at all. If we set a pod creationTS to something in the past, we will trigger the same situation. But this would simplify the number of knobs in the test by removing the lock.

swetharepakula (commented):
/lgtm
/approve

PR #2606, opened 2024-07-24
Replace obsolete test based on neg-status with one based on NEG CR for Standalone NEGs.

gauravkghildiyal:
Replace obsolete test based on neg-status with one based on NEG CR for Standalone NEGs.

- neg-status is not guaranteed to be accurate and a has race condition with the Node Becoming ready and the NEG Controller processing the Service.
- Make the test based on NEG CR the standard one and remove the old test based on NEG status (which is way past the supported version)

/assign @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #2610, opened 2024-07-26
Add instance group resize metrics

08volt:
Add instance group resize metrics.

- An histogram that tracks how many nodes we are adding or removing.
- A counter that tracks the number of adds and removes we are attempting.


08volt:
@cezarygerard

cezarygerard:
We have discussed with Enrico offline, hence the last patch

I will LGTM just after the test pass

great job, Thank you Enrico!

PR #2614, opened 2024-07-30
Update NEG Status update logic when there is conflict in NEG desc.

sawsa307:
* When there is conflict in NEG description, if the conflict occurs within the same cluster, skip update since that CR is owned by a different syncer. Otherwise, set initialize condition to false.
* Always update negObjRef in updateInitStatus to make sure the information is up-to-date.

/assign @swetharepakula

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -33,6 +33,7 @@ import (
 	discovery "k8s.io/api/discovery/v1"
 	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/network"
+	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/endpointslices"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -423,6 +424,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 
 	var errList []error
 	var negObjRefs []negv1beta1.NegObjectReference
+	updateNEGStatus := true
 	negsByLocation := make(map[string]int)
 	for _, zone := range zones {
 		var negObj negv1beta1.NegObjectReference
@@ -445,6 +447,12 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 		)
 		if err != nil {
 			errList = append(errList, err)
+			// Do not modify NEG Status if there is conflict within the same cluster
+			// and namespace because the CR is owned by a different syncer.
+			if errors.Is(err, utils.ErrNEGUsedByAnotherSyncer) {
+				updateNEGStatus = false
+				break
+			}
 		}
 
 		if s.svcNegClient != nil && err == nil {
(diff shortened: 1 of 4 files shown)


pkg/neg/syncers/transaction_test.go
```diff
@@ -1207,7 +1213,8 @@ func TestTransactionSyncerWithNegCR(t *testing.T) {
 			err = syncer.ensureNetworkEndpointGroups()
 			if !tc.expectErr && err != nil {
 				t.Errorf("Expected no error, but got: %v", err)
-			} else if tc.expectErr && err == nil {
+			}
```
swetharepakula, line 1216: Mostly curious why did this change? Are these not mutually exclusive conditions?

pkg/neg/syncers/transaction_test.go
```diff
@@ -1115,40 +1117,43 @@ func TestTransactionSyncerWithNegCR(t *testing.T) {
...
 				ClusterUID:  kubeSystemUID,
 				Namespace:   "namespace-2",
 				ServiceName: testServiceName,
 				Port:        "80",
 			}.String(),
-			crStatusPopulated: false,
-			expectErr:         true,
+			crStatusPopulated:     false,
+			expectErr:             true,
+			expectNoopOnNegStatus: true,
```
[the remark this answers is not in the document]
sawsa307: Updated

pkg/neg/syncers/transaction.go
```diff
@@ -453,7 +459,12 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 		}
 	}
 
-	s.updateInitStatus(negObjRefs, errList)
+	// Do not modify NEG Status if there is conflict within the same cluster
+	// and namespace because the CR is owned by a different syncer.
+	if !noopOnNEGStatus {
```
swetharepakula, line 464: nit: for readability can we reverse the boolean to something like if `updateNegStatus` or `updateInitStatus`.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1042,12 +1041,29 @@ func TestTransactionSyncerWithNegCR(t *testing.T) {
...
+		negExists bool
+		negDesc   string
+		// crStatusPopulated indicates if the NEG CR in this cluster has NEG
+		// status populated, including NEG Initialized and Sync condition, and
+		// the list of NEG references.
 		crStatusPopulated bool
 		customName        bool
 		expectErr         bool
+
+		// expectNoopOnNegStatus indicates if the NEG controller should do
```
swetharepakula, line 1054: Thanks for adding a detailed comment. Can you add what this is checking in terms of the test? Does it check init status, does it check status didn't change etc. That isn't quite clear still.

pkg/neg/syncers/transaction_test.go
```diff
@@ -1042,12 +1041,29 @@ func TestTransactionSyncerWithNegCR(t *testing.T) {
 	testNegType := negtypes.VmIpPortEndpointType
 
 	testCases := []struct {
-		desc              string
-		negExists         bool
-		negDesc           string
+		desc      string
+		negExists bool
+		negDesc   string
+		// crStatusPopulated indicates if the NEG CR in this cluster has NEG
```
swetharepakula, line 1047: Can you add that this means that the status was populated as part of the test setup and this is not an expectation on the result?

sawsa307: Updated the comment.

pkg/neg/syncers/transaction_test.go
sawsa307: Updated the comment.

pkg/neg/syncers/transaction.go
sawsa307: Updated the naming and the boolean.

PR #2615, opened 2024-07-30
Query NEGs from Svcneg when linking BackendServices.

sawsa307:
* Query NEGs from Svcneg Status, and only do GCE query as a fallback.

/assign @swetharepakula

### DIFF
--- a/pkg/backends/neg_linker.go
+++ b/pkg/backends/neg_linker.go
@@ -62,28 +62,10 @@ func NewNEGLinker(
 // Link implements Link.
 func (nl *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 	version := befeatures.VersionFromServicePort(&sp)
-	var negSelfLinks []string
-	var err error
-	for _, group := range groups {
-		// If the group key contains a name, then use that.
-		// Otherwise, get the name from svc port.
-		negName := group.Name
-		if negName == "" {
-			negName = sp.NEGName()
-		}
 
-		negUrl := ""
-		svcNegKey := fmt.Sprintf("%s/%s", sp.ID.Service.Namespace, negName)
-		negUrl, ok := getNegUrlFromSvcneg(svcNegKey, group.Zone, nl.svcNegLister, nl.logger)
-		if !ok {
-			nl.logger.V(4).Info("Falling back to use NEG API to retrieve NEG url for NEG", "negName", negName)
-			neg, err := nl.negGetter.GetNetworkEndpointGroup(negName, group.Zone, version, nl.logger)
-			if err != nil {
-				return err
-			}
-			negUrl = neg.SelfLink
-		}
-		negSelfLinks = append(negSelfLinks, negUrl)
+	negSelfLinks, err := nl.getNegSelfLinks(sp, groups)
+	if err != nil {
+		return err
 	}
 
 	beName := sp.BackendName()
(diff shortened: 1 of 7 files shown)


pkg/backends/neg_linker.go
```diff
@@ -118,6 +100,29 @@ func (nl *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 	return composite.UpdateBackendService(nl.cloud, key, backendService, nl.logger)
 }
 
+func (nl *negLinker) getNegSelfLinks(sp utils.ServicePort, groups []GroupKey) ([]string, error) {
+	version := befeatures.VersionFromServicePort(&sp)
+	negName := sp.NEGName()
```
gauravkghildiyal, line 105: Just confirming for my understanding: The name of the SvcNEG object will be based on the name of the NEG from the default subnet, but this SvcNEG object will contain both the (1.) NEGs from the default and (2.) non-default subnet. Is that correct?

pkg/backends/neg_linker_test.go
```diff
@@ -165,16 +193,132 @@ func TestLinkBackendServiceToNEG(t *testing.T) {
 
 				// mimic cluster node shrinks to one of the zone
 				shrinkZone := []GroupKey{zones[0]}
-				if err := linker.Link(svcPort, shrinkZone); err != nil {
-					t.Fatalf("Failed to link backend service to NEG for svcPort %v: %v", svcPort, err)
+				if err := linker.Link(tc.svcPort, shrinkZone); err != nil {
+					t.Fatalf("Failed to link backend service to NEG for svcPort %v when populateSvcNeg = %v: %v", tc.svcPort, populateSvcNeg, err)
 				}
 
 				validate()
```
gauravkghildiyal, line 200: I think this second `validate()` is trying to ensure that even if the Zones shrink, we don't remove the NEG from the BackendService.

With the new way of using SvcNEG for finding all the NEGs from all the zones, the previous "mimic'ing" part is not sufficiently "mimic'ing" the zone contraction. In addition to passing only zones[0], you may also need to update the existing SvcNEG with only one NegObjectReference, before calling this second `validate()`. 

(NOTE: I think it's also true that SvcNEGs would never be updated with such a zone contraction, but it's still nice to verify this behavior independently here)

gauravkghildiyal (commented):
Thanks @sawsa307. This all looks good to me.

Just one minor suggestion for test.

/lgtm
/hold (If you want approval from Swetha)

pkg/backends/neg_linker.go
sawsa307: Yes that's correct!

pkg/backends/neg_linker_test.go
sawsa307: Thank you so much Gaurav for pointing this out! It is a very subtle point that I haven't thought about.
We do need to update existing SvcNEG, but instead of remove one of the NegObjectReference, we should instead mark the other one as Inactive since this will match the expected behavior by NEG controller(Detailed in #2604, my apology that I haven't thought about this dependency).

In short, when NEG controller no longer observes nodes in a specific zone, it will mark the NEG ref in that zone as Inactive. This is to retain the NEG ref so we won't miss any NEGs during GC().

As a side effect, it would also introduce a small behavior change during zone contract. Previously/currently, we won't include NEGs from contracted zone in the Backend Service. With this PR, since we include NEGs in both Active and Inactive state, NEGs from contracted zone will present in the Backend Service. However, there shouldn't be any effect/issue since this situation is simply equivalent to the current behavior when there are nodes but no endpoints/pods in a specific zone.

gauravkghildiyal: Thanks.

_>Previously/currently, we won't include NEGs from contracted zone in the Backend Service._

Actually I think this is not the case, which means I don't think there's a behavior change -- even better :)

sawsa307: > I think this second validate() is trying to ensure that even if the Zones shrink, we don't remove the NEG from the BackendService.

When the zones shrink, one of the implication is: we will not have any Nodes in the contracted zone.
For the version of codes before this change, it implies that when we call `Link(sp utils.ServicePort, groups []GroupKey)` and iterate over `groups` to construct the updated list of NEGs, the contracted zone should not present in `groups`.

PR #2619, opened 2024-08-06
Remove NEGs in to-be-deleted state after merging Backends.

sawsa307:
/assign @swetharepakula

### DIFF
--- a/pkg/backends/neg_linker.go
+++ b/pkg/backends/neg_linker.go
@@ -82,13 +82,40 @@ func (nl *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 		return err
 	}
 
-	newBackends := backendsForNEGs(negSelfLinks, &sp)
-	// merge backends
+	newBackends := backendsForNEGs(negSelfLinks.negsToAdd, &sp)
+	// Historically, we merged the old backends with the new backends to ensure
+	// that we don't detach NEGs when zones contract. Given that now we primarily
+	// use SvcNEGs as the source-of-truth for calculating backends, and since
+	// they don't get affected by zone-contraction, it's possible we could have
+	// simply used the new backends in the BackendService.
+	// We decided not to do this because we don't know what other errors the
+	// previous decision to mergeBackends may have been hiding.
+	//
+	// Current backend calculation = oldBackends + backends from negsToAdd
+	// - backends from negsToRemove.
 	mergedBackend, err := mergeBackends(backendService.Backends, newBackends)
 	if err != nil {
 		nl.logger.Error(err, fmt.Sprintf("Failed to merge backends from %#v and %#v", backendService.Backends, newBackends))
 		nl.logger.Info("Fall back to ensure backend service with newBackends.")
 		mergedBackend = newBackends
+	}
+
+	// Filter out backends from to-be-deleted NEGs if EnableMultiSubnetClusterPhase1=true.
+	if nl.enableMultiSubnetClusterPhase1 {
+		filteredBackends := mergedBackend
+		if len(negSelfLinks.negsToRemove) != 0 {
+			nl.logger.Info("Removing NEGs in to-be-deleted state from merged backends", "negsToRemove", negSelfLinks.negsToRemove)
+			backendsToRemove := backendsForNEGs(negSelfLinks.negsToRemove, &sp)
+
+			// Remove backends belonging to to-be-deleted NEGs from mergedBackend.
+			filteredBackends, err = removeBackends(mergedBackend, backendsToRemove)
+			if err != nil {
+				nl.logger.Error(err, fmt.Sprintf("Failed to remove backends %#v from %#v", backendsToRemove, mergedBackend))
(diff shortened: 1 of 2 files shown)


pkg/backends/neg_linker.go
```diff
@@ -265,6 +265,9 @@ func getNegUrlsFromSvcneg(key string, svcNegLister cache.Indexer, logger klog.Lo
 	svcneg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
 
 	for _, negRef := range svcneg.Status.NetworkEndpointGroups {
+		if flags.F.EnableMultiSubnetClusterPhase1 && negRef.State == negv1beta1.ToBeDeletedState {
+			continue
```
swetharepakula, line 269: log the decision

pkg/backends/neg_linker_test.go
```diff
@@ -294,6 +295,17 @@ func TestGetNegSelfLinks(t *testing.T) {
 			},
 			expectedNegCount: 4,
 		},
+		{
+			desc:           "Get NEGs from SvcNeg, all NEGs are in active state, and containing NEGs from non-default subnet",
```
swetharepakula, line 299: nit: reword maybe "Get NEGs from SvcNEG, include only NEGs in active state from all subnets"

swetharepakula (commented):
2 nits, otherwise LGTM

pkg/backends/neg_linker_test.go
sawsa307: Updated, thanks!

pkg/backends/neg_linker.go
sawsa307: Added.

gauravkghildiyal:
From the PR description:

> Backend Service updates are performed by providing a new list of backends every time. Thus, if we skip adding NEGs in to-be-delated state, they will be detached from Backend Services when a sync triggers.

Do you actually desire this behaviour? Or are you changing this behavior? I assume it's the former, in which case, I don't think this is what the current behaviour is. 

Somewhat related to the discussion that we were having in the other PR: https://github.com/kubernetes/ingress-gce/pull/2615#discussion_r1762162681. Can you please double check?

gauravkghildiyal:
An improved unit test for various cases of the package level function `Link()` would help

pkg/backends/neg_linker.go
```diff
@@ -80,7 +84,7 @@ func (nl *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 		return err
 	}
 
-	newBackends := backendsForNEGs(negSelfLinks, &sp)
+	newBackends := backendsForNEGs(negSelfLinks.urlsToAdd, &sp)
```
gauravkghildiyal, line 87: Let's add a detailed comment here about what calculation we are doing and why we are doing that, with the hope that somewhere down the line, we may be able to simplify this.

Modify and add this based on your analysis, but this is my understanding:
- Historically, we merged the old backends with the new backends to ensure that we don't detach NEGs when zones contract.
- Given that now we primarily use SvcNEGs as the source-of-truth for calculating backends, and since they don't get affected by zone-contraction, it's possible we could have simply used the new backends in the BackendService. We decided not to do this because we don't know what other errors the previous decision to mergeBackends may have been hiding.
- Explain the new backendscalculation: newBackends = oldBackends + urlsToAdd - urlsToRemove (Rename the variables in the code if required)

pkg/backends/neg_linker.go
```diff
@@ -89,6 +93,12 @@ func (nl *negLinker) Link(sp utils.ServicePort, groups []GroupKey) error {
 		mergedBackend = newBackends
 	}
 
+	// Filter out backends from to-be-deleted NEGs if EnableMultiSubnetClusterPhase1=true.
+	if nl.enableMultiSubnetClusterPhase1 && len(negSelfLinks.urlsToRemove) != 0 {
+		nl.logger.Info("Removing NEGs in to-be-deleted state from merged backends", "negToRemove", negSelfLinks.urlsToRemove)
+		mergedBackend = removeBackendsFromLinks(mergedBackend, negSelfLinks.urlsToRemove)
```
gauravkghildiyal, line 99: If possible, let's use the diffBackends function and mergeBackends function to achieve this. `removeBackendsFromLinks` is doing simple URL String comparision, and we've historically had issues with naively comparing URLs. `diffBackends` has been updated to be resilient from previous known errors.

gauravkghildiyal (commented):
Almost there.

pkg/backends/neg_linker.go
sawsa307: Discussed offline. Since `diffBackend` is a function for reporting purpose, adding a new function `removeBackends` and ensure it behaves just as `mergeBackends` so we won't introduce bugs during comparison.

pkg/backends/neg_linker.go
sawsa307: Updated.

gauravkghildiyal:
Thanks

/lgtm

PR #2620, opened 2024-08-06
Use Premium Network Tier for RXLB

panslava:
- Apparently Premium tier became available for RXLB https://cloud.google.com/network-tiers/docs/overview#resources
- Verified with locally running controller, and it indeed works


swetharepakula (commented):
/lgtm
/approve

PR #2621, opened 2024-08-07
L4 weighted LB backend locality policy

FelipeYepez:
/assign @cezarygerard

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -35,9 +35,18 @@ const (
 	DefaultConnectionDrainingTimeoutSeconds = 30
 	defaultTrackingMode                     = "PER_CONNECTION"
 	PerSessionTrackingMode                  = "PER_SESSION" // the only one supported with strong session affinity
+)
 
-	WeightedLocalityLbPolicy = "WEIGHTED_MAGLEV"
-	DefaultLocalityLbPolicy  = ""
+// LocalityLBPolicyType is the type of locality lb policy the backend service should use.
+type LocalityLBPolicyType string
+
+const (
+	// LocalityLBPolicyDefault is the default locality lb policy for a backend service.
+	LocalityLBPolicyDefault LocalityLBPolicyType = ""
+	// LocalityLBPolicyWeightedMaglev is the locality lb policy for weighted load balancing by pods-per-node.
+	LocalityLBPolicyWeightedMaglev LocalityLBPolicyType = "WEIGHTED_MAGLEV"
+	// LocalityLBPolicyMaglev is the locality lb policy when weighted load balancing by pods-per-node is disabled.
+	LocalityLBPolicyMaglev LocalityLBPolicyType = "MAGLEV"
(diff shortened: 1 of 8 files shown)


pkg/backends/backends.go
```diff
@@ -73,17 +70,29 @@ func NewPoolWithConnectionTrackingPolicy(cloud *gce.Cloud, namer namer.BackendNa
 	}
 }
 
+// LocalityLbPolicyType is the type of locality lb policy the backend service should use.
+type LocalityLbPolicyType string
```
cezarygerard, line 74: nit: put this at the top of the file

pkg/backends/backends.go
```diff
@@ -73,17 +70,29 @@ func NewPoolWithConnectionTrackingPolicy(cloud *gce.Cloud, namer namer.BackendNa
...
+
+const (
```
cezarygerard, line 76: merge with other consts at the top of the file

pkg/backends/backends.go
```diff
@@ -414,23 +420,20 @@ func (b *Backends) EnsureL4BackendService(params L4BackendServiceParams, beLogge
...
-func backendSvcEqual(a, b *composite.BackendService, compareConnectionTracking, compareWeightedLB bool) bool {
+func backendSvcEqual(a, b *composite.BackendService, compareConnectionTracking bool) bool {
 	svcsEqual := a.Protocol == b.Protocol &&
 		a.Description == b.Description &&
 		a.SessionAffinity == b.SessionAffinity &&
 		a.LoadBalancingScheme == b.LoadBalancingScheme &&
 		utils.EqualStringSets(a.HealthChecks, b.HealthChecks) &&
-		a.Network == b.Network
+		a.Network == b.Network &&
+		a.LocalityLbPolicy == b.LocalityLbPolicy
```
cezarygerard, line 430: a.LocalityLbPolicy = ""
b.LocalityLbPolicy = "MAGLEV"

should return equal

cezarygerard: let's do

	svcsEqual = svcsEqual && (a.LocalityLbPolicy == b.LocalityLbPolicy) || (a.LocalityLbPolicy == "" && b.LocalityLbPolicy == "MAGLEV")


instead of
		a.LocalityLbPolicy == b.LocalityLbPolicy

pkg/backends/backends.go
FelipeYepez: Done

pkg/backends/backends.go
FelipeYepez: Fixed, thank you

pkg/backends/backends.go
FelipeYepez: The first set of constants and the second set are distinct concepts and keeping them separate enhances code organization and clarity. Merging the second set with the first would lose type safety and could lead to potential errors if values are accidentally used in the wrong context.

PR #2622, opened 2024-08-07
fix: GCE_VM_IP NEGs should be synced on each Node change.

gauravkghildiyal:
This reverts the behaviour to what existed prior to https://github.com/kubernetes/ingress-gce/pull/1754.

This is also logical, since for GCE_VM_IP NEGs, the endpoints are the nodes themselves. So similar to the behavior for GCE_VM_IP_PORT negs which get synced on EndpointSlice changes, it makes sense to sync GCE_VM_IP NEGs on Node changes.

/assign @swetharepakula 
/assign @mmamczur

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -318,22 +318,29 @@ func (manager *syncerManager) SyncNodes() {
 	isVmIpPortZoneChange := updateZoneMap(&manager.vmIpPortZoneMap, negtypes.NodeFilterForNetworkEndpointType(negtypes.VmIpPortEndpointType), manager.zoneGetter, manager.logger)
 
 	for key, syncer := range manager.syncerMap {
+		manager.logger.V(1).Info("SyncNodes: Evaluating sync decision for syncer", "negSyncerKey", key.String())
+
 		if syncer.IsStopped() {
+			manager.logger.V(1).Info("SyncNodes: Syncer is already stopped; not syncing.", "negSyncerKey", key.String())
 			continue
 		}
 
 		switch key.NegType {
 
 		case negtypes.VmIpEndpointType:
+			manager.logger.V(1).Info("SyncNodes: Triggering sync", "negSyncerKey", key.String(), "negSyncerType", key.NegType)
 			syncer.Sync()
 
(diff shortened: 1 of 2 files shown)


swetharepakula (commented):
/approve

As this affects L4 as well, will defer to @cezarygerard give the final lgtm

pkg/neg/manager.go
```diff
@@ -318,7 +315,6 @@ func (manager *syncerManager) SyncNodes() {
 	defer manager.mu.Unlock()
 
 	// When a zone change occurs (new zone is added or deleted), a sync should be triggered
-	isVmIpZoneChange := updateZoneMap(&manager.vmIpZoneMap, negtypes.NodeFilterForNetworkEndpointType(negtypes.VmIpEndpointType), manager.zoneGetter, manager.logger)
 	isVmIpPortZoneChange := updateZoneMap(&manager.vmIpPortZoneMap, negtypes.NodeFilterForNetworkEndpointType(negtypes.VmIpPortEndpointType), manager.zoneGetter, manager.logger)
```
swetharepakula, line 319: Can you add log lines here for all the decisions made as in do we sync, do we not sync. And some context as to why the decision was made?

gauravkghildiyal: Done

cezarygerard:
/assign @mmamczur 

PTAL as you did all the hard work in RCA-ing this bug

pkg/neg/manager.go
mmamczur: could we also verify that we don't end up with this log line dominating output?
example right now, this seems to be the log line dominating the logs `"Exported resource lock availability metrics" lockName="ingress-gce-neg-lock"` or `"NEGController/ReadinessReflector: Skip processing pod"`
when debugging you end up having to come up with filter for all of that stuff.

mmamczur:
looks ok if we're not worried about increased requests rate to GCE

/approve

pkg/neg/manager.go
gauravkghildiyal: I kinda understand the concern, but this one isn't as noisy as the other two:
- The first one is logged almost every second (no questions asked)
- The second one is logged on pod creation.

This one would get logged on each node creation / deletion. I expect this to be far lesser than the number of log creations.

PR #2633, opened 2024-08-16
Cluster with network id - fix

08volt:
Supporting cluster with network defined using network Id instead of network Name

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -23,6 +23,7 @@ import (
 	networkv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/network/v1"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	compute "google.golang.org/api/compute/v1"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/ingress-gce/pkg/utils"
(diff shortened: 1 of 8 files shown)


pkg/network/network.go
```diff
@@ -61,6 +97,21 @@ func NewNetworksResolver(networkLister, gkeNetworkParamSetLister cache.Indexer,
...
 
+func GetNetworkSelfLink(gceCloud *gce.Cloud) (string, error) {
+	if gceCloud == nil {
+		return "", fmt.Errorf("Netowrk resolver: provided cloud is nil")
+	}
+	url := gceCloud.NetworkURL()
+	lastIndex := strings.LastIndex(url, "/")
+	networkId := url[lastIndex+1:]
+
+	networkResource, err := gceCloud.GetNetwork(networkId)
```
mmamczur, line 108: can we try to make this fall back to the old behavior if this call fails for whatever reason and log an error?

but maybe we don't need this, up to you

pkg/network/network.go
```diff
@@ -45,13 +46,48 @@ type Resolver interface {
...
 	logger                   klog.Logger
 }
 
+type CloudNetworkProviderAdapterWithSelfLink struct {
+	cloud      *gce.Cloud
+	networkURL string
+}
+
+// CloudNetworkProviderAdapterWithSelfLink uses the network selfLink as networkUrl.
+func NewAdapterNetworkSelfLink(gceCloud *gce.Cloud) (CloudNetworkProvider, error) {
```
mmamczur, line 60: could you add a test to at least verify if this adapter correctly resolves the network? You can use fakeCloud for this probably

pkg/network/network.go
08volt: Yes, this is what it does if 
``` 
adapter, err := network.NewAdapterNetworkSelfLink(ctx.Cloud)
```
fails, adapter = ctx.Cloud, which is like it was in the past

mmamczur: ah true I haven't noticed this

pkg/network/network.go
08volt: fixed

swetharepakula:
What was the context behind needing this change? Could we consider using a utils function instead to convert the URL? I feel this adds unnecessary complexity.

mmamczur:
so there is this issue that when you create a cluster and specify an URL of the VPC to use. If in that scenario you give it an URL with network number instead of the typical network URL with the network name, the cluster will be created and will seemingly work. ingress-gce however, will have this numeric ID in the config. The problem is that if you pass the network URL in this form to some parts of GCE API, it will error on it. So you need to always use the network name, if you only have the number you actually have to do an API call to get the name.

PR #2635, opened 2024-08-19
Add initial cache sync check

08volt:
## Wait for cache sync before L4NetLBController and L4Controller run

This pull request introduces a delay mechanism to ensure cache synchronization before the L4NetLBController and L4Controller commence operation. This is accomplished by incorporating a 10-minute wait period with 5-second polling intervals, ensuring that all required informers have been duly updated.

## Testing:

* **Unit Tests:** Existing unit tests should continue to pass without modifications.

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -28,6 +28,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/ingress-gce/pkg/annotations"
@@ -79,6 +80,8 @@ type L4Controller struct {
 	sharedResourcesLock sync.Mutex
 	enableDualStack     bool
 
+	hasSynced func() bool
+
 	serviceVersions *serviceVersionsTracker
 
 	logger klog.Logger
@@ -103,6 +106,7 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}, lo
 		enableDualStack: ctx.EnableL4ILBDualStack,
 		serviceVersions: NewServiceVersionsTracker(),
 		logger:          logger,
+		hasSynced:       ctx.HasSynced,
 	}
 	l4c.backendPool = backends.NewPool(ctx.Cloud, l4c.namer)
 	l4c.NegLinker = backends.NewNEGLinker(l4c.backendPool, negtypes.NewAdapter(ctx.Cloud), ctx.Cloud, ctx.SvcNegInformer.GetIndexer(), logger)
@@ -193,7 +197,13 @@ func (l4c *L4Controller) checkHealth() error {
 
 func (l4c *L4Controller) Run() {
 	defer l4c.shutdown()
-	l4c.logger.Info("Running L4 Controller with worker goroutines", "numWorkers", l4c.numWorkers)
+
+	wait.PollUntil(5*time.Second, func() (bool, error) {
+		l4c.logger.V(2).Info("Waiting for initial cache sync before starting L4 Controller")
+		return l4c.hasSynced(), nil
+	}, l4c.stopCh)
+
+	l4c.logger.Info("Running L4 Controller", "numWorkers", l4c.numWorkers)
 	l4c.svcQueue.Run()
 	<-l4c.stopCh
 }
--- a/pkg/l4lb/l4controller_test.go
+++ b/pkg/l4lb/l4controller_test.go
@@ -457,7 +457,7 @@ func TestProcessUpdateClusterIPToILBService(t *testing.T) {
 func TestProcessMultipleServices(t *testing.T) {
 	backoff := retry.DefaultRetry
 	// Increase the duration since updates take longer on prow.
-	backoff.Duration = 1 * time.Second
+	backoff.Duration = 3 * time.Second
 	for _, onlyLocal := range []bool{true, false} {
 		t.Run(fmt.Sprintf("L4 with LocalMode=%v", onlyLocal), func(t *testing.T) {
 			l4c := newServiceController(t, newFakeGCE())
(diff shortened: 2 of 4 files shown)


cmd/glbc/main.go
```diff
@@ -277,6 +278,16 @@ func main() {
 	}
 	ctx.Init()
 
+	// Ensures all informers have populated the cache before initiating any controllers.
+	// This centralized approach eliminates the need for individual controllers to perform their own cache synchronization checks.
+	err = wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 5*time.Minute, true, func(context.Context) (bool, error) {
```
cezarygerard, line 283: 2 seconds / 1 minute?

cezarygerard:
lgtm overall

let's wait for @gauravkghildiyal and/ro @swetharepakula

cmd/glbc/main.go
08volt: done

cmd/glbc/main.go
```diff
@@ -277,6 +278,16 @@ func main() {
...
+	// Ensures all informers have populated the cache before initiating any controllers.
+	// This centralized approach eliminates the need for individual controllers to perform their own cache synchronization checks.
+	err = wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 1*time.Minute, true, func(context.Context) (bool, error) {
+		rootLogger.Info("Waiting for initial cache sync before running controllers")
+		return ctx.HasSynced(), nil
+	})
+	if err != nil {
+		klog.Fatalf("Initial sync failed: %v", err)
+	}
+
```
swetharepakula, line 290: The informers don't start populating until ctx.Start()  which is started in runControllers or runNEGController. Those functions also determine which informers are necessary based on the controllers that are enabled

We also don't want to start the informers until it is determined this binary wins leader election to ensure we don't cache K8s resources unnecessarily and increase our memory footprint

gauravkghildiyal: Great analysis!

(This makes me wonder what the e2e tests are verifying since they clearly passed.)

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -416,9 +416,13 @@ func (lc *L4NetLBController) checkHealth() error {
...
 func (lc *L4NetLBController) Run() {
-	defer lc.shutdown()
 	lc.logger.Info("Starting l4NetLBController")
-	lc.svcQueue.Run()
+	defer func() {
+		lc.logger.Info("Shutting down l4NetLBController")
+		lc.shutdown()
+	}()
+
+	go lc.svcQueue.Run()
```
gauravkghildiyal, line 425: nit: With reference to the PR description: _"Refactored to start the svcQueue in a separate goroutine."_

Please self-evaluate if we really gain anything out of this.

cmd/glbc/main.go
08volt: Thanks, I was not aware of that. I will fix this.

cezarygerard: Fair enough, we will follow the established patterns fro now.


My next point is for wider discussion, beyond this PR:

 This is a bit error prone to always remember to block on  hasSynced. The check was missing in both L4 controllers. It would be better to put it in context's.Start func, like:

```
func (ctx *ControllerContext) Start(stopCh <-chan struct{}) {
	go ctx.IngressInformer.Run(stopCh)
	go ctx.ServiceInformer.Run(stopCh)
	go ctx.EndpointSliceInformer.Run(stopCh)

	if ctx.FirewallInformer != nil {
		go ctx.FirewallInformer.Run(stopCh)
	}
...

     err = wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 1*time.Minute, true, func(context.Context) (bool, error) {
		rootLogger.Info("Waiting for initial cache sync before running controllers")
		return ctx.HasSynced(), nil
	})
	if err != nil {
		klog.Fatalf("Initial sync failed: %v", err)
	}
}
```

This would make controllers code simpler, test code simpler and we will never repeat  same bug again.
The only drawback may be slightly longer  startup time for the component

pkg/context/context.go
```diff
@@ -423,6 +425,14 @@ func (ctx *ControllerContext) Start(stopCh <-chan struct{}) {
 	}
 	// Export ingress usage metrics.
 	go ctx.ControllerMetrics.Run(stopCh)
+
+	err := wait.PollUntilContextTimeout(context2.Background(), 2*time.Second, 1*time.Minute, true, func(context2.Context) (bool, error) {
```
wojtek-t, line 429: FWIW, I've definitely seen clusters where PodInformer taking more than 1 minute to initialize...

08volt: fixed

cmd/glbc/main.go
08volt: fixed

cezarygerard:
@swetharepakula @gauravkghildiyal 

I see your points, let's not refactore anything here, agree that big changes in startup may be missed by current E2E tests.

Enrico has now implemented the minimal changes  in the 
- pkg/l4lb/l4lbcontroller.go
- pkg/l4lb/l4netlbcontroller.go

and the affected tests

these changes only follow the existing patterns, bare minimum to fix the issue.

Please review and LGTM

pkg/l4lb/l4controller.go
```diff
@@ -193,7 +198,16 @@ func (l4c *L4Controller) checkHealth() error {
 
 func (l4c *L4Controller) Run() {
 	defer l4c.shutdown()
-	l4c.logger.Info("Running L4 Controller with worker goroutines", "numWorkers", l4c.numWorkers)
+
+	err := wait.PollUntilContextTimeout(context2.Background(), 5*time.Second, 10*time.Minute, true, func(context2.Context) (bool, error) {
```
swetharepakula, line 202: instead of a timeout, we should use the stopchan. it is unlikely that 10 minutes is not enough but on the chance it is not, the controller will never succeed in running

pkg/l4lb/l4controller_test.go
```diff
@@ -515,7 +515,9 @@ func TestProcessServiceWithDelayedNEGAdd(t *testing.T) {
 	addILBService(l4c, newSvc)
 	l4c.svcQueue.Enqueue(newSvc)
 
-	if err := retry.OnError(retry.DefaultRetry, func(error) bool { return true }, func() error {
+	backoff := retry.DefaultRetry
+	backoff.Duration = 3 * time.Second
```
swetharepakula, line 519: We should further investigate why this was needed with the cache sync, but can be done outside of this PR

swetharepakula (commented):
/lgtm
/approve

pkg/l4lb/l4controller_test.go
cezarygerard: After the controller waits for informers, which takes now longer that original initial delay on the test.

Without this change tests if finished, running assertions before controller 

it's clear as day

PR #2645, opened 2024-08-26
Remove unused code, add unused linter check

panslava:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,6 +7,7 @@ linters:
   enable:
     - errcheck
     - forbidigo
+    - unused
 
 issues:
   exclude-rules:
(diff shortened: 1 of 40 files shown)


gauravkghildiyal:
Thanks @panslava -- Please double check that besides the `timeout` flag, we aren't removing any other flags (atleast I didn't spot any but worth checking). For `timeout`, I tried to verify that it isn't being used elsewhere in the binary invocation so hopefully we should be all good with removing it.

/lgtm
/approve
/hold (Unhold once verified)

panslava:
@gauravkghildiyal I rebased on master could you please lgtm again

I checked again, I don't think we remove any other flags than `serverTimeout`

/unhold

PR #2646, opened 2024-08-26
Determine zones by querying the SvcNEG CRs

swetharepakula:
Neg status annotation is a best effort annotation that is not always accurate in regards to the zones. Therefore switch to using the SvcNEG CRs to determine which zones NEGs are currently in.

/assign @gauravkghildiyal


gauravkghildiyal:
Thanks

/lgtm
/hold (for your verification, if any)

swetharepakula:
/unhold 
Verified that tests pass locally

PR #2648, opened 2024-08-27
force send backend service fields: localityLbPolicy

cezarygerard:
PR tries to force send localityLbPolicy 

but fails...


pkg/composite/gen.go
```diff
@@ -4719,6 +4719,7 @@ func (backendService *BackendService) ToAlpha() (*computealpha.BackendService, e
 			alpha.LogConfig.ForceSendFields = []string{"Enable", "SampleRate"}
 		}
 	}
+	alpha.ForceSendFields = []string{"localityLbPolicy"}
```
bowei, line 4722: NullFields?

PR #2653, opened 2024-08-29
Improve backendSvcEqual  code for  backend service creation

cezarygerard:
with new backendSvcEqual we never update the localityLbPolicy on existing services that do not use Weighted Load Balancers
also now this invariant is true: `backendSvcEqual(old, new) == backendSvcEqual(new, old)`

plus I have changed the naming of some variables as confusing names  may have contributed to the problem

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -344,7 +344,7 @@ func (b *Backends) EnsureL4BackendService(params L4BackendServiceParams, beLogge
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -79,6 +80,8 @@ type L4Controller struct {
 	sharedResourcesLock sync.Mutex
 	enableDualStack     bool
 
+	hasSynced func() bool
+
```
FelipeYepez, line 84: There are previous commits as part of this PR which should not be included

FelipeYepez (changes requested):

PR #2658, opened 2024-09-03
Instance Group Manager does not remove nodes with empty string in the zone field

08volt:
This PR updates the Instance group manager.

Bug: Load Balancer controller removes nodes when they have an empty string in the Zone field, but the nodes are actually already in the correct instance group.

Solution:
Avoid removing nodes which have an empty string in the Zone field.


08volt:
@cezarygerard

cezarygerard:
my lgtm was accidental, for now

pkg/instancegroups/manager.go
```diff
@@ -310,7 +310,17 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
 	// https://github.com/kubernetes/cloud-provider-gcp/blob/fca628cb3bf9267def0abb509eaae87d2d4040f3/providers/gce/gce_loadbalancer_internal.go#L606C1-L675C1
 	// the m.maxIGSize should be set to 1000 as is in the cloud-provider-gcp.
 	zonedNodes := m.splitNodesByZone(nodes, iglogger)
```
cezarygerard, line 312: the splitNodesByZone will not return empty zone

 m.ZoneGetter.ZoneForNode(name, logger) returns zone or error,
never empty string without error

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -58,7 +58,7 @@ var ErrNodePodCIDRNotSet = errors.New("Node does not have PodCIDR set")
 
 // providerIDRE is the regex to process providerID.
 // A providerID is build out of '${ProviderName}://${project-id}/${zone}/${instance-name}'
-var providerIDRE = regexp.MustCompile(`^` + "gce" + `://([^/]+)/([^/]+)/([^/]+)$`)
```
cezarygerard, line 61: why touch this?

pkg/instancegroups/manager.go
```diff
@@ -310,7 +310,17 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
...
+	iglogger.Info(fmt.Sprintf("Syncing nodes: %d nodes over %d zones", len(nodes), len(zonedNodes)))
+
+	emptyZoneNodesNames := sets.NewString(zonedNodes[""]...)
+	if len(emptyZoneNodesNames) > 0 {
+		iglogger.Info(fmt.Sprintf("%d nodes have empty zone: %v", len(emptyZoneNodesNames), emptyZoneNodesNames))
```
cezarygerard, line 317: iglogger.Info(fmt.Sprintf("%d nodes have empty zone: %v. They will not be removed from instance group as long as zone is missing", len(emptyZoneNodesNames), emptyZoneNodesNames))

pkg/instancegroups/manager.go
```diff
@@ -335,7 +345,7 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
 			gceNodes.Insert(instance)
 		}
 
-		removeNodes := gceNodes.Difference(kubeNodes).List()
+		removeNodes := gceNodes.Difference(kubeNodes).Difference(emptyZoneNodesNames).List() // Do not remove nodes which zone label still need to be assigned
```
cezarygerard, line 348: removeNodes := gceNodes.Difference(kubeNodes)

iglogger.V(2).Info("Nodes that are removal candidates", "removeNodes", events.TruncatedStringList(removeNodes))

removeNodes = removeNodes.Difference(emptyZoneNodesNames).List() // Do not remove nodes which zone label still need to be assigned

iglogger.V(2).Info("Removing nodes (after ignoring nodes without zone assigned)", "removeNodes", events.TruncatedStringList(removeNodes))

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -161,7 +161,7 @@ func (z *ZoneGetter) ListZones(filter Filter, logger klog.Logger) ([]string, err
 	zones := sets.String{}
 	for _, n := range nodes {
 		zone, err := getZone(n)
-		if err != nil {
+		if err != nil || zone == "" {
```
cezarygerard, line 164: add const here for ""

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +60,108 @@ func newNodePool(f Provider, maxIGSize int) Manager {
...
+		if err != nil {
+			t.Fatalf("fakeGCEInstanceGroups.ListInstancesInInstanceGroup(%s, %s, %s) returned error %v, want nil", ig.Name, defaultTestZone, allInstances, err)
+		}
+		instances, err := test.InstancesListToNameSet(instancesList)
+		if err != nil {
+			t.Fatalf("test.InstancesListToNameSet(%v) returned error %v, want nil", ig, err)
+		}
+
+		expectedInstancesSize := len(testCase.kubeNodes)
+		if len(testCase.kubeNodes) > maxIGSize {
```
mmamczur, line 138: do we need this mxIGSize check in a test that won't create so many nodes?

08volt: fixed, thanks!

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +60,103 @@ func newNodePool(f Provider, maxIGSize int) Manager {
 	return pool
 }
 
+func TestNodePoolSyncWithEmptyZone(t *testing.T) {
+	maxIGSize := 1000
+
+	names1001 := make([]string, maxIGSize+1)
+	for i := 1; i <= maxIGSize+1; i++ {
+		names1001[i-1] = fmt.Sprintf("n%d", i)
+	}
```
cezarygerard, line 69: not used in the test?

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +60,103 @@ func newNodePool(f Provider, maxIGSize int) Manager {
 	return pool
 }
 
+func TestNodePoolSyncWithEmptyZone(t *testing.T) {
```
mmamczur, line 63: this test is super difficult to follow, maybe because it was copied

the scenerio overall I understand should be that there is an instance group in zone X with some nodes,
then one of these nodes looses the zone assignment but we don't want it to be removed from the IG

1. this is a table test but has only 1 case. Do we plan to extend it?
2. it's hard to see that the kubeNodes and kubeZones have to match by index
3. test does a lot of operations that are not clear, counting API calls etc

could we maybe make it simpler, for example just explicitly if the node that was supposed to remain in the group is there instead of checking if removal API call was made? (unless removing is broken in that 'fake')

mmamczur: also the test name should reflect that this tests the scenario where the node lost the zone, it can't be empty initally

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +59,154 @@ func newNodePool(f Provider, maxIGSize int) Manager {
...
+					name: "n2",
+					zone: "",
+				},
+			},
+			wantedNodes: []string{"n1", "n2"},
+		},
+	}
+
+	for _, testCase := range testCases {
+		// create fake gce node pool with existing gceNodes
```
mmamczur, line 153: put the contents of the for block inside 
```
// test case code here
}
```
that way when you run the code these will be named test cases, if some fail you'll se which ones failed specifically etc

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +59,139 @@ func newNodePool(f Provider, maxIGSize int) Manager {
...
+		names = append(names, name)
+	}
+	return names
+}
+
+func TestNodePoolSyncWithEmptyZone(t *testing.T) {
+
+	testCases := []struct {
+		desc              string
+		gceNodes          []string            // all in defaultTestZone
```
mmamczur, line 74: can we rename this as discussed, so let's say this would be named like `instanceGroupMembers` or `instanceGroupVMs`

pkg/instancegroups/manager_test.go
```diff
@@ -59,6 +59,139 @@ func newNodePool(f Provider, maxIGSize int) Manager {
...
+		kubeNodesEachStep []map[string]string // map of node:zone during each "ensure instance group"
```
mmamczur, line 75: I think we don't need these steps, I think it would be enough to have the instanceGroupState and then the nodes in k8s to update and compare with what we want to be in the instance group

in all these cases it doesn't matter what happens in step 1 because wantNodes are checked at the end.

example
```
		{
			desc:     "Both nodes without zone during second update do not get deleted",
			instanceGroupMembers: []string{"n1", "n2"},
			kubeNodesWithZone: map[string]string{
					"n1": "",
					"n2": "",
			},
			wantedNodes: []string{"n1", "n2"},
		},
```

PR #2659, opened 2024-09-03
Increase number of buckets for DualStack Sync Latency metrics

08volt:
Update number of buckets from 12 to 15 to the following metrics:
- Latency of an L4 NetB DualStack Sync
- Latency of an L4 ILB DualStack Sync


08volt:
@cezarygerard

PR #2661, opened 2024-09-05
Modify the regional IG linker to not detach instance gropus from zone…

mmamczur:
…s that still have nodes.

There is an issue where after master upgrade all nodes are seen as unready by the master. This causes the controller to see this state, these nodes seem to be unfit candidates for LB backends and the controller removes the matching instance groups. Theis causes traffic outage in the LB.

This change prevents removing instance groups for zones where some nodes exist, even if these nodes are unready.

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -642,7 +642,14 @@ func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, linkType 
 		return lc.negLinker.Link(servicePort, groupKeys)
 	} else if linkType == instanceGroupLink {
 		svcLogger.V(2).Info("Linking backend service with Instance Groups for service (uses default network)")
-		return lc.igLinker.Link(servicePort, lc.ctx.Cloud.ProjectID(), zones)
+
+		// This lists zones from all nodes, even unready ones. If a zone would be removed from the service but is in this list it should stay linked.
+		// This is to prevent situations where a zone would be removed by nodes becoming temporarily unready causing an outage.
+		zonesThatShouldNotBeRemoved, err := lc.zoneGetter.ListZones(zonegetter.AllNodesFilter, svcLogger)
+		if err != nil {
+			return err
+		}
+		return lc.igLinker.Link(servicePort, lc.ctx.Cloud.ProjectID(), zones, zonesThatShouldNotBeRemoved)
 	} else {
 		return fmt.Errorf("cannot link backend service - invalid backend link type %d", linkType)
 	}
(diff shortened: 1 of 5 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -642,7 +642,14 @@ func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, linkType
 		return lc.negLinker.Link(servicePort, groupKeys)
```
cezarygerard, line 642: the entire func ensureBackendLinking func can be much simpler

check this:



```
func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, linkType backendLinkType, svcLogger klog.Logger) error {
	start := time.Now()

	svcLogger.V(2).Info("Linking backends to backend service for k8s service")
	defer func() {
		svcLogger.V(2).Info("Finished linking backends to backend service for k8s service", "timeTaken", time.Since(start))
	}()

	namespacedName := types.NamespacedName{Name: service.Name, Namespace: service.Namespace}
	portId := utils.ServicePortID{Service: namespacedName}
	servicePort := utils.ServicePort{
		ID:           portId,
		BackendNamer: lc.namer,
		L4RBSEnabled: true,
	}
	// NEG backends should only be used for multinetwork services on the non default network.
	if linkType == negLink {
		return lc.linkNegs(svcLogger, servicePort)
	} else if linkType == instanceGroupLink {
		return lc.linkInstanceGroups(svcLogger, servicePort)
	} else {
		return fmt.Errorf("cannot link backend service - invalid backend link type %d", linkType)
	}
}

func (lc *L4NetLBController) linkNegs(svcLogger klog.Logger, servicePort utils.ServicePort) error {
	zones, err := lc.zoneGetter.ListZones(zonegetter.CandidateNodesFilter, svcLogger)
	if err != nil {
		return err
	}
	svcLogger.V(2).Info("Linking backend service with NEGs for service")
	servicePort.VMIPNEGEnabled = true
	var groupKeys []backends.GroupKey
	for _, zone := range zones {
		groupKeys = append(groupKeys, backends.GroupKey{Zone: zone})
	}
	return lc.negLinker.Link(servicePort, groupKeys)
}

func (lc *L4NetLBController) linkInstanceGroups(svcLogger klog.Logger, servicePort utils.ServicePort) error {
	zones, err := lc.zoneGetter.ListZones(zonegetter.AllNodesFilter, svcLogger)
	if err != nil {
		return err
	}
	svcLogger.V(2).Info("Linking backend service with Instance Groups for service (uses default network)")
	return lc.igLinker.Link(servicePort, lc.ctx.Cloud.ProjectID(), zones)
}
```



I believe it achieves everything we need here

mmamczur: so what you do here is basically use AllNodesFilter for Instance Groups instead of the old one

the issue here is, I think, that the ensure IGs will create groups only for candidate nodes. So you may end up trying to link to a group that does not exist. 
You could maybe change the IG creation code as well but this is in the L4 RBS controller + the legacy ILB one + in Ingress
Maybe it would be fine if we created more groups just for L4RBS, it would be left empty in the worst case

PR #2663, opened 2024-09-06
Use all zones when creating the shared IG for an RBS service.

mmamczur:
With this change, the code to ensure IGs will use AllNodesFilter to determine which zones to create IGs in. The goal is to avoid detaching groups which become temporarily unready, for example after master upgrade. This change should be fine since creating and using an IG that could be empty is ok from the LB perspective. On the other hand removing a temporarily unready group can cause 10 minutes of no traffic.

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -84,8 +84,8 @@ func NewManager(config *ManagerConfig) Manager {
 // all of which have the exact same named ports.
 func (m *manager) EnsureInstanceGroupsAndPorts(name string, ports []int64, logger klog.Logger) (igs []*compute.InstanceGroup, err error) {
 	iglogger := logger.WithName("InstanceGroupsManager")
-	// Instance groups need to be created only in zones that have ready nodes.
-	zones, err := m.ZoneGetter.ListZones(zonegetter.CandidateNodesFilter, iglogger)
+	// Instance groups need to be created in all zones that nodes are in.
+	zones, err := m.ZoneGetter.ListZones(zonegetter.AllNodesFilter, iglogger)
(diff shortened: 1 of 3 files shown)


pkg/instancegroups/manager.go
```diff
@@ -85,7 +85,7 @@ func NewManager(config *ManagerConfig) Manager {
 func (m *manager) EnsureInstanceGroupsAndPorts(name string, ports []int64, logger klog.Logger) (igs []*compute.InstanceGroup, err error) {
 	iglogger := logger.WithName("InstanceGroupsManager")
 	// Instance groups need to be created only in zones that have ready nodes.
```
cezarygerard, line 87: please change the comment here

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -644,6 +639,10 @@ func (lc *L4NetLBController) ensureBackendLinking(service *v1.Service, linkType
 	// NEG backends should only be used for multinetwork services on the non default network.
 	if linkType == negLink {
 		svcLogger.V(2).Info("Linking backend service with NEGs for service")
+		zones, err := lc.zoneGetter.ListZones(zonegetter.CandidateNodesFilter, svcLogger)
```
cezarygerard, line 642: add a comment we only attach ready nodes to the NEG

pkg/instancegroups/manager.go
mmamczur: done

pkg/l4lb/l4netlbcontroller.go
mmamczur: added

PR #2665, opened 2024-09-06
ForwardingRule - Patch

08volt:
Patching forwarding rules instead of recreating them when updating only patchable fields.

Patchable fields:

AllowGlobalAccess
NetworkTier


08volt:
@cezarygerard

cezarygerard:
in pkg/loadbalancers/forwarding_rules_test.go

func TestL4CreateExternalForwardingRule(t *testing.T) {


you need to add recorder fake

PR #2667, opened 2024-09-06
Fix the race condition that causes syncer missing in the syncerMap

ruixiansong:
/assign @swetharepakula 
/cc @gauravkghildiyal

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -175,7 +175,6 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 
 	removes := currentPorts.Difference(newPorts)
 	adds := newPorts.Difference(currentPorts)
-	samePorts := newPorts.Difference(adds)
 	// There may be duplicate ports in adds and removes due to difference in readinessGate flag
 	// Service/Ingress config changes can cause readinessGate to be turn on or off for the same service port.
 	// By removing the duplicate ports in removes and adds, this prevents disruption of NEG syncer due to the config changes
@@ -199,32 +198,21 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 		}
 	}
 
-	for _, portInfo := range samePorts {
+	// Ensure a syncer is running for each port in newPorts.
+	for svcPort, portInfo := range newPorts {
+		syncerKey := manager.getSyncerKey(namespace, name, svcPort, portInfo)
+		syncer, ok := manager.syncerMap[syncerKey]
+		// To ensure that a NEG CR always exists during the lifecycle of a NEG, do not create a
+		// syncer for the NEG until the NEG CR is successfully created. This will reduce the
+		// possibility of invalid states and reduces complexity of garbage collection
 		// To reduce the possibility of NEGs being leaked, ensure a SvcNeg CR exists for every
 		// desired port.
 		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-			errList = append(errList, fmt.Errorf("failed to ensure svc neg cr %s/%s/%d for existing port: %w", namespace, portInfo.NegName, portInfo.PortTuple.Port, err))
+			errList = append(errList, fmt.Errorf("failed to ensure svc neg cr %s/%s/%d for port: %w ", namespace, portInfo.NegName, svcPort.ServicePort, err))
 			errorSyncers += 1
-		} else {
-			successfulSyncers += 1
+			continue
 		}
-	}
-
-	// Ensure a syncer is running for each port that is being added.
-	for svcPort, portInfo := range adds {
-		syncerKey := manager.getSyncerKey(namespace, name, svcPort, portInfo)
-		syncer, ok := manager.syncerMap[syncerKey]
 		if !ok {
-
-			// To ensure that a NEG CR always exists during the lifecycle of a NEG, do not create a
-			// syncer for the NEG until the NEG CR is successfully created. This will reduce the
-			// possibility of invalid states and reduces complexity of garbage collection
-			if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
-				errList = append(errList, fmt.Errorf("failed to ensure svc neg cr %s/%s/%d for new port: %w ", namespace, portInfo.NegName, svcPort.ServicePort, err))
-				errorSyncers += 1
-				continue
-			}
-
 			// determine the implementation that calculates NEG endpoints on each sync.
 			epc := negsyncer.GetEndpointsCalculator(
 				manager.podLister,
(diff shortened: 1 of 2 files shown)


pkg/neg/manager_test.go
```diff
@@ -330,6 +330,43 @@ func TestEnsureAndStopSyncer(t *testing.T) {
...
+	portInfo := types.PortInfo{PortTuple: negtypes.SvcPortTuple{Name: portName, Port: svcPort, TargetPort: targetPort}, NegName: negName}
+
+	portMap[negtypes.PortInfoMapKey{ServicePort: svcPort}] = portInfo
+
+	manager.serviceLister.Add(&v1.Service{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}})
+
+	svcKey := serviceKey{namespace: namespace, name: name}
+	negSyncerKey := manager.getSyncerKey(namespace, name, negtypes.PortInfoMapKey{ServicePort: svcPort}, portInfo)
+	manager.svcPortMap[svcKey] = portMap
+	manager.svcPortMap = map[serviceKey]negtypes.PortInfoMap{
```
sawsa307, line 352: I suppose this line is doing the same thing as the previous line?

sawsa307 (commented):
One nit, everything else looks good to me. Will leave final lgtm to @swetharepakula and @gauravkghildiyal 

/approve

pkg/neg/manager_test.go
ruixiansong: Thanks! Done!

gauravkghildiyal:
Thanks @ruixiansong 

It's possible that I am misunderstanding something, but to me it seems that this still has the same race condition:

- The first time around when the syncer fails to start, because we've added it to the syncerMap already, the second Ensure() corresponding to the `samePorts` will never get executed.

Assuming that we still have the race, we likely also need to improve our unit test to detect the problem?

ruixiansong:
> Thanks @ruixiansong
> 
> It's possible that I am misunderstanding something, but to me it seems that this still has the same race condition:
> 
> * The first time around when the syncer fails to start, because we've added it to the syncerMap already, the second Ensure() corresponding to the `samePorts` will never get executed.
> 
> Assuming that we still have the race, we likely also need to improve our unit test to detect the problem?

Thanks @gauravkghildiyal for reviewing this.

I'm assuming the syncer in the map will finally got GCed. Then it won't be in the syncerMap and then we can recreate it here.

Or we need to start the syncer here again before it got GCed?

gauravkghildiyal:
> > Thanks @ruixiansong
> > It's possible that I am misunderstanding something, but to me it seems that this still has the same race condition:
> > 
> > * The first time around when the syncer fails to start, because we've added it to the syncerMap already, the second Ensure() corresponding to the `samePorts` will never get executed.
> > 
> > Assuming that we still have the race, we likely also need to improve our unit test to detect the problem?
> 
> Thanks @gauravkghildiyal for reviewing this.
> 
> I'm assuming the syncer in the map will finally got GCed. Then it won't be in the syncerMap and then we can recreate it here.
> 
> Or we need to start the syncer here again before it got GCed?

Thanks for clarification. Yes that does seem to resolve the race.

Though doesn't the wait for the GC introduce the following degrading behaviour:

1. Service is created. Syncers are running well and healthy
2. Service is deleted. We allow sufficient time for the Service to be deleted (maybe like 1 minute)
3. After that 1 minute, we recreate the Service.
4. It's possible that the GC hasn't happened by this point which means the syncerMap still has the previous syncer. 
5. When we do EnsureSyncer() again, because syncerMap still has the previous syncer (which isn't GC'd yet), we never start it again. In addition, we also don't return any errors.
6. This means that for the recreated Service, it likely has to wait for the next periodic resync of the Service to properly start the syncers...which means a delay of 10 minutes at minimum.

Sorry for making this painful, but the changes here touch a critical lifecycle, and I want to make sure we have all bases covered. (So I may end up calling out problems which I may not have fully vetted)

pkg/neg/manager.go
```diff
@@ -199,7 +199,19 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
...
+			if err := manager.EnsureSyncer(namespace, name, svcPort, portInfo); err != nil {
+				errList = append(errList, err)
+				errorSyncers += 1
+			}
+			successfulSyncers += 1
+			continue
+		}
 		// To reduce the possibility of NEGs being leaked, ensure a SvcNeg CR exists for every
 		// desired port.
 		if err := manager.ensureSvcNegCR(key, portInfo); err != nil {
```
gauravkghildiyal, line 217: We're doing an ensureSvcNEGCR here, and we're also doing one within EnsureSyncer().

Doesn't this mean that we're doing the exact same thing for both `samePorts` and `addPorts`? Assuming `newPorts = samePorts + addPorts`, maybe we can refactor this to just process `newPorts`? (Let's double check)

ruixiansong:
> > > Thanks @ruixiansong
> > > It's possible that I am misunderstanding something, but to me it seems that this still has the same race condition:
> > > 
> > > * The first time around when the syncer fails to start, because we've added it to the syncerMap already, the second Ensure() corresponding to the `samePorts` will never get executed.
> > > 
> > > Assuming that we still have the race, we likely also need to improve our unit test to detect the problem?
> > 
> > 
> > Thanks @gauravkghildiyal for reviewing this.
> > I'm assuming the syncer in the map will finally got GCed. Then it won't be in the syncerMap and then we can recreate it here.
> > Or we need to start the syncer here again before it got GCed?
> 
> Thanks for clarification. Yes that does seem to resolve the race.
> 
> Though doesn't the wait for the GC introduce the following degrading behaviour:
> 
> 1. Service is created. Syncers are running well and healthy
> 2. Service is deleted. We allow sufficient time for the Service to be deleted (maybe like 1 minute)
> 3. After that 1 minute, we recreate the Service.
> 4. It's possible that the GC hasn't happened by this point which means the syncerMap still has the previous syncer.
> 5. When we do EnsureSyncer() again, because syncerMap still has the previous syncer (which isn't GC'd yet), we never start it again. In addition, we also don't return any errors.
> 6. This means that for the recreated Service, it likely has to wait for the next periodic resync of the Service to properly start the syncers...which means a delay of 10 minutes at minimum.
> 
> Sorry for making this painful, but the changes here touch a critical lifecycle, and I want to make sure we have all bases covered. (So I may end up calling out problems which I may not have fully vetted)

That makes sense, updated the EnsureSyncers() function to restart the stopped syncers to avoid the degrading behaviour.

pkg/neg/manager.go
ruixiansong: Yes that makes sense, I completely refactored this one and combined those with one loop for all newPorts.

gauravkghildiyal:
Thanks!

/lgtm

PR #2671, opened 2024-09-12
Check PodIP against PodCIDR only when hostNetwork=false.

sawsa307:
/assign @swetharepakula
/assign @gauravkghildiyal

### DIFF
--- a/pkg/neg/syncers/utils.go
+++ b/pkg/neg/syncers/utils.go
@@ -568,9 +568,11 @@ func validatePod(pod *apiv1.Pod, nodeLister, serviceLister cache.Indexer, networ
 		count[negtypes.OtherError]++
 		return count, negtypes.ErrEPNodeTypeAssertionFailed
 	}
-	if err = nodeContainsPodIP(node, networkEndpoint); err != nil {
-		count[negtypes.IPOutOfPodCIDR]++
-		return count, err
+	if !pod.Spec.HostNetwork {
+		if err = nodeContainsPodIP(node, networkEndpoint); err != nil {
+			count[negtypes.IPOutOfPodCIDR]++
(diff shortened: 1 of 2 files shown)


swetharepakula (commented):
can you add tests for this?

gauravkghildiyal:
Thanks

/lgtm

sawsa307:
Just added test cases to ensure we don't filter out pods if it uses hostNetwork. Please take a look!

pkg/neg/syncers/utils.go
```diff
@@ -568,7 +568,7 @@ func validatePod(pod *apiv1.Pod, nodeLister, serviceLister cache.Indexer, networ
 		count[negtypes.OtherError]++
 		return count, negtypes.ErrEPNodeTypeAssertionFailed
 	}
-	if err = nodeContainsPodIP(node, networkEndpoint); err != nil {
+	if err = nodeContainsPodIP(node, networkEndpoint); !pod.Spec.HostNetwork && err != nil {
```
aojea, line 571: I don't know how the golang compiler translate this, but is also easier to understand if you only execute `nodeContainsPodIP(node, networkEndpoint)` if `!pod.Spec.HostNetwork`

```suggestion
        if  !pod.Spec.HostNetwork {
            if err = nodeContainsPodIP(node, networkEndpoint); && err != nil {
        }
```

gauravkghildiyal: Seems + reads better

sawsa307: Yes I agree. This makes the check easier to read. Updated.

swetharepakula (commented):
Thanks!

/approve

PR #2675, opened 2024-09-16
Create Node Topology CRD client and informer.

sawsa307:
/assign @swetharepakula 
/assign @gauravkghildiyal


gauravkghildiyal:
Looks good, just one question: Who installs this CRD on the cluster? Is it safe to assume that if `EnableMultiSubnetClusterPhase1` flag is set, then the CRD will be available?

sawsa307:
> Who installs this CRD on the cluster?
It shall be installed by NodeIPAM controller.

> Is it safe to assume that if EnableMultiSubnetClusterPhase1 flag is set, then the CRD will be available?
Yes we will make sure the flag is enabled only after NodeIPAM controller has included the definition in the component manifest.

gauravkghildiyal:
Thanks

/lgtm
/approve

PR #2685, opened 2024-09-23
Allow externalTrafficPolicy: Cluster on multinetworking services

shouri007:
Removes the constraint that previously restricted externalTrafficPolicy to only `Local` on multinetworking services

### DIFF
--- a/pkg/network/network_test.go
+++ b/pkg/network/network_test.go
@@ -195,35 +195,6 @@ func TestServiceNetwork(t *testing.T) {
 			service:            serviceWithSecondaryNet,
 			wantErr:            "network.Spec.Type=Device is not supported for multinetwork LoadBalancer services, the only supported network type is L3",
 		},
-		{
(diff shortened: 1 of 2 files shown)


shouri007:
@mmamczur dpv2 component with the changes is rolled out to prod. We can merge this PR if these changes look good

pkg/network/network_test.go
```diff
@@ -195,35 +195,6 @@ func TestServiceNetwork(t *testing.T) {
 			service:            serviceWithSecondaryNet,
 			wantErr:            "network.Spec.Type=Device is not supported for multinetwork LoadBalancer services, the only supported network type is L3",
 		},
-		{
```
mmamczur, line 198: since this was a special case we should now change it to expect a valid result

mmamczur: ping

mmamczur:
ok let's merge this now since I don't want surprises later but I don't like how tests were handled here

/lgtm

PR #2689, opened 2024-09-26
Moving from string(zone) to endpointGroupInfo when mapping NEGs, and populate subnet values for nodes in zoneGetter and NEG syncers.

sawsa307:
* Use endpointGroupInfo instead of string(zone) when mapping NEGs since now endpoints shall be partitioned based on the zone and the subnet they are in.
* Nodes in non-default subnets will be filtered when enableMultiSubnetCluster=true and enableMultiSubnetClusterPhase1=false. When enableMultiSubnetClusterPhase1=true, we will include endpoints
  and nodes from non-default subnets.

### DIFF
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -56,6 +56,7 @@ func NewLocalL4EndpointsCalculator(nodeLister listers.NodeLister, zoneGetter *zo
 	if lbType == negtypes.L4ExternalLB {
 		subsetSize = maxSubsetSizeNetLBLocal
 	}
+
 	return &LocalL4EndpointsCalculator{
 		nodeLister:      nodeLister,
 		zoneGetter:      zoneGetter,
@@ -106,7 +107,7 @@ func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsDat
 				l.logger.Info("Node not connected to service network", "nodeName", node.Name, "network", l.networkInfo.K8sNetwork)
 				continue
 			}
-			zone, err := l.zoneGetter.ZoneForNode(node.Name, l.logger)
+			zone, _, err := l.zoneGetter.ZoneAndSubnetForNode(node.Name, l.logger)
 			if err != nil {
 				l.logger.Error(err, "Unable to find zone for node, skipping", "nodeName", node.Name)
 				metrics.PublishNegControllerErrorCountMetrics(err, true)
@@ -121,7 +122,10 @@ func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsDat
 	}
 	// Compute the networkEndpoints, with total endpoints count <= l.subsetSizeLimit
 	l.logger.V(2).Info("Got zoneNodeMap as input for service", "zoneNodeMap", nodeMapToString(zoneNodeMap), "serviceID", l.svcId)
+	// TODO(sawsa307): Make sure to include logic for subsetting endpoints in non-default subnets.
+	// Currently we only select endpoints from the default subnet.
 	subsetMap, err := getSubsetPerZone(zoneNodeMap, l.subsetSizeLimit, l.svcId, currentMap, l.logger, l.networkInfo)
+
 	return subsetMap, nil, 0, err
 }
 
@@ -182,7 +186,7 @@ func (l *ClusterL4EndpointsCalculator) CalculateEndpoints(_ []types.EndpointsDat
 			l.logger.Info("Node not connected to service network", "nodeName", node.Name, "network", l.networkInfo.K8sNetwork)
 			continue
 		}
-		zone, err := l.zoneGetter.ZoneForNode(node.Name, l.logger)
+		zone, _, err := l.zoneGetter.ZoneAndSubnetForNode(node.Name, l.logger)
 		if err != nil {
 			l.logger.Error(err, "Unable to find zone for node skipping", "nodeName", node.Name)
 			metrics.PublishNegControllerErrorCountMetrics(err, true)
@@ -192,6 +196,8 @@ func (l *ClusterL4EndpointsCalculator) CalculateEndpoints(_ []types.EndpointsDat
 	}
 	l.logger.V(2).Info("Got zoneNodeMap as input for service", "zoneNodeMap", nodeMapToString(zoneNodeMap), "serviceID", l.svcId)
 	// Compute the networkEndpoints, with total endpoints <= l.subsetSizeLimit.
+	// TODO(sawsa307): Make sure to include logic for subsetting endpoints in non-default subnets.
+	// Currently we only select endpoints from the default subnet.
 	subsetMap, err := getSubsetPerZone(zoneNodeMap, l.subsetSizeLimit, l.svcId, currentMap, l.logger, l.networkInfo)
 	return subsetMap, nil, 0, err
 }
--- a/pkg/neg/syncers/endpoints_calculator.go
+++ b/pkg/neg/syncers/endpoints_calculator.go
@@ -72,7 +72,7 @@ func (l *LocalL4EndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, currentMap map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
 	// List all nodes where the service endpoints are running. Get a subset of the desired count.
 	zoneNodeMap := make(map[string][]*v1.Node)
 	processedNodes := sets.String{}
@@ -125,7 +125,7 @@ func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsDat
 	return subsetMap, nil, 0, err
 }
 
-func (l *LocalL4EndpointsCalculator) CalculateEndpointsDegradedMode(eds []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, error) {
+func (l *LocalL4EndpointsCalculator) CalculateEndpointsDegradedMode(eds []types.EndpointsData, currentMap map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, error) {
 	// this should be the same as CalculateEndpoints for L4 ec
 	subsetMap, podMap, _, err := l.CalculateEndpoints(eds, currentMap)
 	return subsetMap, podMap, err
@@ -173,7 +173,7 @@ func (l *ClusterL4EndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *ClusterL4EndpointsCalculator) CalculateEndpoints(_ []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+func (l *ClusterL4EndpointsCalculator) CalculateEndpoints(_ []types.EndpointsData, currentMap map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
 	// In this mode, any of the cluster nodes can be part of the subset, whether or not a matching pod runs on it.
 	nodes, _ := l.zoneGetter.ListNodes(zonegetter.CandidateAndUnreadyNodesFilter, l.logger)
 	zoneNodeMap := make(map[string][]*v1.Node)
@@ -196,7 +196,7 @@ func (l *ClusterL4EndpointsCalculator) CalculateEndpoints(_ []types.EndpointsDat
 	return subsetMap, nil, 0, err
 }
 
-func (l *ClusterL4EndpointsCalculator) CalculateEndpointsDegradedMode(eps []types.EndpointsData, currentMap map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, error) {
+func (l *ClusterL4EndpointsCalculator) CalculateEndpointsDegradedMode(eps []types.EndpointsData, currentMap map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, error) {
 	// this should be the same as CalculateEndpoints for L4 ec
 	subsetMap, podMap, _, err := l.CalculateEndpoints(eps, currentMap)
 	return subsetMap, podMap, err
@@ -244,7 +244,7 @@ func (l *L7EndpointsCalculator) Mode() types.EndpointsCalculatorMode {
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
+func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, int, error) {
 	result, err := toZoneNetworkEndpointMap(eds, l.zoneGetter, l.podLister, l.servicePortName, l.networkEndpointType, l.enableDualStackNEG, l.enableMultiSubnetCluster, l.logger)
 	if err == nil { // If current calculation ends up in error, we trigger and emit metrics in degraded mode.
 		l.syncMetricsCollector.UpdateSyncerEPMetrics(l.syncerKey, result.EPCount, result.EPSCount)
@@ -253,7 +253,7 @@ func (l *L7EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsData, _ 
 }
 
 // CalculateEndpoints determines the endpoints in the NEGs based on the current service endpoints and the current NEGs.
-func (l *L7EndpointsCalculator) CalculateEndpointsDegradedMode(eds []types.EndpointsData, _ map[string]types.NetworkEndpointSet) (map[string]types.NetworkEndpointSet, types.EndpointPodMap, error) {
+func (l *L7EndpointsCalculator) CalculateEndpointsDegradedMode(eds []types.EndpointsData, _ map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet) (map[negtypes.EndpointGroupInfo]types.NetworkEndpointSet, types.EndpointPodMap, error) {
 	result := toZoneNetworkEndpointMapDegradedMode(eds, l.zoneGetter, l.podLister, l.nodeLister, l.serviceLister, l.servicePortName, l.networkEndpointType, l.enableDualStackNEG, l.enableMultiSubnetCluster, l.logger)
 	l.syncMetricsCollector.UpdateSyncerEPMetrics(l.syncerKey, result.EPCount, result.EPSCount)
 	return result.NetworkEndpointSet, result.EndpointPodMap, nil
--- a/pkg/neg/syncers/endpoints_calculator_test.go
+++ b/pkg/neg/syncers/endpoints_calculator_test.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/neg/metrics/metricscollector"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
 	"k8s.io/ingress-gce/pkg/network"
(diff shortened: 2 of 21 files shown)


sawsa307:
cc @mmamczur

pkg/neg/types/interfaces.go
```diff
@@ -81,11 +81,16 @@ type NegSyncerManager interface {
...
+	CalculateEndpointsDegradedMode(eds []EndpointsData, currentMap map[EndpointInfo]NetworkEndpointSet) (map[EndpointInfo]NetworkEndpointSet, EndpointPodMap, error)
 	// Mode indicates the mode that the EndpointsCalculator is operating in.
 	Mode() EndpointsCalculatorMode
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+type EndpointInfo struct {
+	Zone   string
+	Subnet string
```
mmamczur, line 95: so the empty Subnet value means the default subnet?
What should be the content of this field for the non default subnets?

pkg/neg/types/interfaces.go
```diff
@@ -81,11 +81,16 @@ type NegSyncerManager interface {
...
 	// CalculateEndpointsDegradedMode computes the NEG endpoints using degraded mode calculation
-	CalculateEndpointsDegradedMode(eds []EndpointsData, currentMap map[string]NetworkEndpointSet) (map[string]NetworkEndpointSet, EndpointPodMap, error)
+	CalculateEndpointsDegradedMode(eds []EndpointsData, currentMap map[EndpointInfo]NetworkEndpointSet) (map[EndpointInfo]NetworkEndpointSet, EndpointPodMap, error)
 	// Mode indicates the mode that the EndpointsCalculator is operating in.
 	Mode() EndpointsCalculatorMode
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+type EndpointInfo struct {
```
mmamczur, line 93: 'EndpointInfo' suggests it holds information about an endpoint - in our case a pod for l7 or node for l4, while the struct contents are data that map 1:1 to the NetworkEndpoint**Group**. Maybe it could be `EndpointGroupInfo` or `GroupZoneAndSubnet` or something like that

pkg/neg/types/interfaces.go
sawsa307: > so the empty Subnet value means the default subnet?

I'm open to suggestions here. For now I leave it as empty string since most of nodes in default subnets doesn't have this label. A different approach is we can manually pipe the default subnet value so it is not empty

gauravkghildiyal:
I haven't taken a look, but I thought this was a draft?

sawsa307:
> I haven't taken a look, but I thought this was a draft?

Yes it is still a draft. I converted to "ready to review" so we can get some result on the `pull-ingress-gce-e2e` and `pull-ingress-gce-test`.

gauravkghildiyal:
> > I haven't taken a look, but I thought this was a draft?
> 
> Yes it is still a draft. I converted to "ready to review" so we can get some result on the `pull-ingress-gce-e2e` and `pull-ingress-gce-test`.

Oh gotcha! Yep all I want is for you to save time on not trying to perfect it without deciding which path you want to take.

pkg/neg/types/interfaces.go
sawsa307: Sounds good, updated to `EndpointGroupInfo`.

pkg/neg/types/interfaces.go
swetharepakula: I think both situations need to be handled. We need to understand the empty string means the default subnet and handle that some nodes might have empty and some might have the subnet name.

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -62,8 +62,8 @@ func TestLocalGetEndpointSet(t *testing.T) {
 			endpointsData: negtypes.EndpointsDataFromEndpointSlices(getDefaultEndpointSlices()),
```
swetharepakula, line 62: do we have tests that cover both cases of nodes. Some with the empty string as the subnet and some with the default subnet name

swetharepakula (commented):
Please expand in the commit messages to provide more context such as "moving from string (zone) to the endpoint group info in preparation for creating NEGs per subnet".

The last commit (`Use defaultSubnet as subnet value for endpoints in default subnet.`), I don't understand what the goal of this commit is? It is just passing in the default subnet. Do we need this? In most of these places you will have to change the code again when we start pulling the info for the other subnets.

pkg/neg/syncers/transaction.go
```diff
@@ -498,12 +505,12 @@ func (s *transactionSyncer) syncNetworkEndpoints(addEndpoints, removeEndpoints m
 				go s.attachNetworkEndpoints(zone, batch)
```
mmamczur, line 505: so I guess here you also need to pass down the subnet

pkg/neg/controller.go
```diff
@@ -170,6 +171,15 @@ func NewController(
 	recorder := eventBroadcaster.NewRecorder(negScheme,
 		apiv1.EventSource{Component: "neg-controller"})
 
+	var defaultSubnet string
+	if flags.F.EnableMultiSubnetClusterPhase1 {
+		defaultSubnet, err = utils.KeyName(defaultSubnetURL)
```
mmamczur, line 176: here you have access to cloud, so you could just do `cloud.SubnetworkURL()` here or even in the manager, unless you prefer this for testing purposes

pkg/neg/syncers/transaction.go
```diff
@@ -678,7 +680,8 @@ func (s *transactionSyncer) commitPods(endpointMap map[string]negtypes.NetworkEn
 			}
 			zoneEndpointMap[endpoint] = podName
 		}
-		s.reflector.CommitPods(s.NegSyncerKey, s.NegSyncerKey.NegName, zone, zoneEndpointMap)
+		// Only zone is needed because NEGs from non-default subnet has different names.
```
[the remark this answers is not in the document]
sawsa307: Updated.

pkg/neg/syncers/endpoints_calculator_test.go
sawsa307: No that's not needed. We will make sure the subnet value is not empty. This enhancement is introduced in the last PR.

If a node/endpoint is in the default subnet, the subnet field in its `EndpointGroupInfo` should contain the value of the defaultSubnet.

sawsa307 (commented):
> Do we need this? In most of these places you will have to change the code again when we start pulling the info for the other subnets.

This is to make sure we have a uniform way to present endpoints in the default subnet, instead of having empty string and default subnet value representing the same thing.
There are only two places we generate `EndpointGroupInfo`. One is during endpoint calculation, which we can get the subnet value from the node. The other one is during `retrieveExistingZoneNetworkEndpointMap`, which should only contains endpoints from the default subnet. It will require update anyways because we don't currently create NEGs from non-default subnet.

pkg/neg/syncers/transaction.go
sawsa307: Added a todo, thanks

pkg/neg/controller.go
sawsa307: Updated, thanks

pkg/neg/types/interfaces.go
```diff
@@ -81,11 +81,16 @@ type NegSyncerManager interface {
...
 	// CalculateEndpointsDegradedMode computes the NEG endpoints using degraded mode calculation
-	CalculateEndpointsDegradedMode(eds []EndpointsData, currentMap map[string]NetworkEndpointSet) (map[string]NetworkEndpointSet, EndpointPodMap, error)
+	CalculateEndpointsDegradedMode(eds []EndpointsData, currentMap map[EndpointGroupInfo]NetworkEndpointSet) (map[EndpointGroupInfo]NetworkEndpointSet, EndpointPodMap, error)
 	// Mode indicates the mode that the EndpointsCalculator is operating in.
 	Mode() EndpointsCalculatorMode
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+type EndpointGroupInfo struct {
```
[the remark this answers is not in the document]
sawsa307: Done.

mmamczur:
lgtm from my side, we need this to be able to do endpoint calculator changes

pkg/neg/syncers/subsets.go
```diff
@@ -193,7 +193,7 @@ func getSubsetPerZone(nodesPerZone map[string][]*v1.Node, totalLimit int, svcID
 			} else {
 				ip = utils.GetNodePrimaryIP(node, logger)
 			}
-			result[zone.Name].Insert(negtypes.NetworkEndpoint{Node: node.Name, IP: ip})
+			result[negtypes.EndpointGroupInfo{Zone: zone.Name, Subnet: defaultSubnet}].Insert(negtypes.NetworkEndpoint{Node: node.Name, IP: ip})
```
gauravkghildiyal, line 196: There's a potential bug here -- this part seems to be assuming that the Node is in the defaultSubnet, and I don't think such a constraint is enforced anywhere.

If you've already discussed this with @mmamczur and fixing this is something outside the scope of this PR, let's atleast add a comment here so that it can be fixed shortly after by the L4 team in a followup PR.

pkg/neg/manager.go
```diff
@@ -227,6 +245,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				manager.syncerMetrics,
 				&portInfo.NetworkInfo,
 				portInfo.L4LBType,
+				manager.defaultSubnet,
```
gauravkghildiyal, line 248: Why do we need a separate `defaultSubnet` when we are already passing NetworkInfo? Why can't we resuse the networkInfo? (Similar comment for the TransactionSyncer construction)

pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -49,20 +49,24 @@ type LocalL4EndpointsCalculator struct {
 	svcId           string
 	logger          klog.Logger
 	networkInfo     *network.NetworkInfo
+	// The subnet specified in gce.conf is considered as the default subnet.
+	defaultSubnet string
 }
 
-func NewLocalL4EndpointsCalculator(nodeLister listers.NodeLister, zoneGetter *zonegetter.ZoneGetter, svcId string, logger klog.Logger, networkInfo *network.NetworkInfo, lbType types.L4LBType) *LocalL4EndpointsCalculator {
+func NewLocalL4EndpointsCalculator(nodeLister listers.NodeLister, zoneGetter *zonegetter.ZoneGetter, svcId string, logger klog.Logger, networkInfo *network.NetworkInfo, lbType types.L4LBType, defaultSubnet string) *LocalL4EndpointsCalculator {
```
gauravkghildiyal, line 56: Why do we need a separate defaultSubnet, when networkInfo has the same information?

pkg/neg/syncers/transaction.go
```diff
@@ -156,6 +159,15 @@ func NewTransactionSyncer(
...
 	logger := log.WithName("Syncer").WithValues("service", klog.KRef(negSyncerKey.Namespace, negSyncerKey.Name), "negName", negSyncerKey.NegName)
 
+	var defaultSubnet string
+	var err error
+	if flags.F.EnableMultiSubnetClusterPhase1 {
+		defaultSubnet, err = utils.KeyName(networkInfo.SubnetworkURL)
+		if err != nil {
+			logger.Error(err, "Errored getting default subnet from NetworkInfo")
+		}
+	}
```
gauravkghildiyal, line 169: What do we gain from pre-calculating the defaultSubnet during the TransactionSyncer object construction? Why can't we calculate it when we want? Doesn't the current code lead to a situation where you are logging an error, but still continuing assuming everything is just fine? (I know sometimes these things are unavoidable, but in this case this to me seems to hint towards the fact that perhaps this is not the right place to do this)

gauravkghildiyal (commented):
Thanks for addressing the comments @sawsa307!

Besides the two comments that I have, the rest seems reasonable to me. 

I will not block the PR on the following, but I somehow feel the naming `EndpointGroupInfo` might be overly vague. Given the idea with the struct is to define what is the granularity for each NEG (i.e. For each `EndpointGroupInfo`, we have a unique NEG), so how does something like `NEGLocation` sound? Anyone else has any thoughts here? There's also the option of `ZoneSubnet` -- I know we likely are avoiding this to be flexible for some potential future use case but perhaps this vagueness might not be fully justified here. 

Either way, we can likely do this renaming in a followup PR to unblock the rest of the feature development.

pkg/neg/syncers/endpoints_calculator.go
sawsa307: This is to ensure we only parse the information once in manager.go. The alternative I can think of is we parse the defaultSubnet from defaultSubnetURL in `NewTransactionSyncer` and `GetEndpointsCalculator`.

sawsa307: Just saw the other comment. Now it makes more sense to generate on the fly. Let me update this.

pkg/neg/syncers/transaction.go
sawsa307: Okay yes that makes sense. Generate on need sounds like a better solution. Let me update this, thank you!

sawsa307:
> Thanks for addressing the comments @sawsa307!

Besides the two comments that I have, the rest seems reasonable to me.

I will not block the PR on the following, but I somehow feel the naming EndpointGroupInfo might be overly vague. Given the idea with the struct is to define what is the granularity for each NEG (i.e. For each EndpointGroupInfo, we have a unique NEG), so how does something like NEGLocation sound? The idea being, for each location (NEGLocation), I have a NEG there. Anyone else has any thoughts here? There's also the option of ZoneSubnet -- I know we likely are avoiding this to be flexible for some potential future use case but perhaps this vagueness might not be fully justified here.

Either way, we can likely do this renaming in a followup PR to unblock the rest of the feature development.

Thank you so much Gaurav for reviewing this!
I'm fine with either naming(NEGLocation/EndpointGroupInfo). Definitely would like more opinions on this.
There is one another change included since the previous review. I discussed with @swetharepakula and we think we should try to simplify the logic in the zoneGetter regarding `defaultSubnet`, given the two msc phases have different behaviors towards endpoints in non-default subnets.
The change being made is: Now everything will based on the field `onlyIncludeDefaultSubnetNodes` in zoneGetter. If it is true, we will filter out all nodes in non-default subnets. Otherwise, we will populate the correct subnet value for the node.

pkg/neg/syncers/transaction.go
```diff
@@ -259,7 +259,16 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 	s.logger.V(2).Info("Sync NEG", "negSyncerKey", s.NegSyncerKey.String(), "endpointsCalculatorMode", s.endpointsCalculator.Mode())
 
-	currentMap, currentPodLabelMap, err := retrieveExistingZoneNetworkEndpointMap(s.NegSyncerKey.NegName, s.zoneGetter, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, s.logger)
+	var defaultSubnet string
+	var err error
+	if !flags.F.EnableMultiSubnetCluster {
```
gauravkghildiyal, line 264: If `flags.F.EnableMultiSubnetCluster = true`:

- defaultSubnet for currentMap = ""
- In targetMap, from zone getter, defaultSubnet received would depend on whether the Node has the label so both "" and "<SUBNET_NAME>" values are possible (depending on whether the Node had a label or not)

Is it guaranteed that for the phase when `flags.F.EnableMultiSubnetCluster = true`, then the Node Labels for subnet will never be set?

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -331,6 +339,26 @@ func getZone(node *api_v1.Node) (string, error) {
 	return matches[2], nil
 }
 
+// getSubnet gets subnet information from node's LabelNodeSubnet.
+// If a node doesn't have this label, or the label value is empty, it means
+// this node is in the defaultSubnet and we will parse the subnet from the
+// defaultSubnetURL.
+func getSubnet(node *api_v1.Node, defaultSubnetURL string) (string, error) {
```
gauravkghildiyal, line 346: This seems almost similar to `isNodeInDefaultSubnet`, is there no way to combine these two?

sawsa307: That's a good point. Refactored, thanks!

pkg/neg/syncers/transaction.go
sawsa307: Sorry for the confusion. Just updated the code again since the logic is very confusing.
Right now `defaultSubnet` should not be empty in any case. In the currentMap, we will use defaultSubnet = `the value of the default subnet`, and in targetMap and zoneGetter, `defaultSubnet` received will be based on the node label. If the label is non-empty, the value is the actual subnet value. If the label is empty or doesn't exist, we will return the actual value of the default subnet as well. Please let me know if there is more confusion on this, thank you!

gauravkghildiyal (commented):
Almost there.

---

From the commit message:

```
Populate subnet values for nodes in zoneGetter and NEG syncers.
* Nodes in non-default subnets will be filtered when
  enableMultiSubnetCluster=true. This behavior will be overriden by
  enableMultiSubnetClusterPhase1 flag, since we will include non-default
  subnet nodes then.
```

Does this mean when `enableMultiSubnetClusterPhase1=FALSE`, we will start considering nodes from non-default subnet?

pkg/neg/syncers/transaction.go
```diff
@@ -259,7 +259,12 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 	s.logger.V(2).Info("Sync NEG", "negSyncerKey", s.NegSyncerKey.String(), "endpointsCalculatorMode", s.endpointsCalculator.Mode())
 
-	currentMap, currentPodLabelMap, err := retrieveExistingZoneNetworkEndpointMap(s.NegSyncerKey.NegName, s.zoneGetter, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, s.logger)
+	defaultSubnet, err := utils.KeyName(s.networkInfo.SubnetworkURL)
+	if err != nil {
+		s.logger.Error(err, "Errored getting default subnet from NetworkInfo")
```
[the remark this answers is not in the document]
sawsa307: Updated.

pkg/neg/syncers/subsets.go
```diff
@@ -173,13 +173,20 @@ func getSubsetPerZone(nodesPerZone map[string][]*v1.Node, totalLimit int, svcID
 	// Sort zones in increasing order of node count.
 	zoneList := sortZones(nodesPerZone)
 
+	defaultSubnet, err := utils.KeyName(networkInfo.SubnetworkURL)
+	if err != nil {
+		logger.Error(err, "Errored getting default subnet from NetworkInfo")
```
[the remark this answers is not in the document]
sawsa307: Updated.

sawsa307:
> Almost there.
> 
> From the commit message:
> 
> ```
> Populate subnet values for nodes in zoneGetter and NEG syncers.
> * Nodes in non-default subnets will be filtered when
>   enableMultiSubnetCluster=true. This behavior will be overriden by
>   enableMultiSubnetClusterPhase1 flag, since we will include non-default
>   subnet nodes then.
> ```
> 
> Does this mean when `enableMultiSubnetClusterPhase1=FALSE`, we will start considering nodes from non-default subnet?

Updated the commit message so it correctly reflects the behavior.

gauravkghildiyal:
> > Almost there.
> > From the commit message:
> > ```
> > Populate subnet values for nodes in zoneGetter and NEG syncers.
> > * Nodes in non-default subnets will be filtered when
> >   enableMultiSubnetCluster=true. This behavior will be overriden by
> >   enableMultiSubnetClusterPhase1 flag, since we will include non-default
> >   subnet nodes then.
> > ```
> > 
> > Does this mean when `enableMultiSubnetClusterPhase1=FALSE`, we will start considering nodes from non-default subnet?
> 
> Updated the commit message so it correctly reflects the behavior.

My concern is whether the behavior correctly matches the commit message; isn't this contradictory (from zone getter)?

```
onlyIncludeDefaultSubnetNodes: flags.F.EnableMultiSubnetCluster && flags.F.EnableMultiSubnetClusterPhase1,
```

sawsa307:
> Body

Oh yes the second flag should be negated. Pushed up another commit.

gauravkghildiyal:
Thanks for the patience. This LGTM. Once the other related changes merge, I hope we can thoroughly test this out manually.

If people don't have a strong preference, I'd suggest we rename `EndpointGroupInfo` to `NEGLocation` in a followup PR -- I hope that should not be as intense as this given it's a simple rename.

/lgtm
/hold (In case you are waiting for additional reviews)

sawsa307:
Trying to squash and rebase the code based on swetharepakula@'s request.

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -199,62 +200,80 @@ func TestListNodesMultipleSubnets(t *testing.T) {
...
 				if zone != tc.expectZone {
 					t.Errorf("For test case %q with onlyIncludeDefaultSubnetNodes = %v , got zone: %s, want: %s,", tc.desc, enableMultiSubnetCluster, zone, tc.expectZone)
 				}
+
 				if !errors.Is(err, tc.expectErr) {
 					t.Errorf("For test case %q with onlyIncludeDefaultSubnetNodes = %v, got error: %s, want: %s,", tc.desc, enableMultiSubnetCluster, err, tc.expectErr)
 				}
+			}
 
+			for _, enableMultiSubnetClusterPhase1 := range []bool{true, false} {
```
swetharepakula, line 265: I think it is a great idea to re-use the test cases, but I would put these into its own t.Run as the earlier part of the test and this part are not connected.

pkg/neg/syncers/subsets.go
```diff
@@ -173,13 +173,21 @@ func getSubsetPerZone(nodesPerZone map[string][]*v1.Node, totalLimit int, svcID
 	// Sort zones in increasing order of node count.
 	zoneList := sortZones(nodesPerZone)
 
+	defaultSubnet, err := utils.KeyName(networkInfo.SubnetworkURL)
```
swetharepakula, line 176: made a comment on a previous commit. we should just store this and calculate it at initialization. if there is a problem better to throw an error at startup because it will alway cause an error

pkg/neg/syncers/transaction_test.go
```diff
@@ -1332,9 +1343,11 @@ func TestTransactionSyncerWithNegCR(t *testing.T) {
 // zone has the expected State.
 func TestUpdateInitStatusWithMultiSubnetCluster(t *testing.T) {
 	testNetwork := cloud.ResourcePath("network", &meta.Key{Name: "test-network"})
-	testSubnetwork := cloud.ResourcePath("subnetwork", &meta.Key{Name: "test-subnetwork"})
 	testNegType := negtypes.VmIpPortEndpointType
 	flags.F.EnableMultiSubnetClusterPhase1 = true
+	defer func() {
+		flags.F.EnableMultiSubnetClusterPhase1 = false
```
swetharepakula, line 1349: this should store the previous value and set it back then assuming that the flag will be false

PR #2690, opened 2024-09-27
Zonal Affinity for ILB

08volt:
This PR allows the user to use the Zonal Affinity feature when creating an ILB

It includes
- Flag: EnableZonalAffinity
- Annotation: networking.gke.io/zonal-affinity-spillover-ratio

( Using backendService API v0.Alpha, need to be updated as soon as the feature is released )


08volt:
@mmamczur @FelipeYepez

PR #2691, opened 2024-09-30
Add backend_type label to l4_netlbs_count and the L4_netlb latency metrics

mmamczur:
This extends the L4 NetLB metrics with `backend_type` field that will allow tracking the adoption and performance of NEG backed LBs vs IG ones.

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -533,6 +533,8 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
 		svcLogger.Info("Finished syncing L4 NetLB RBS service", "timeTaken", time.Since(startTime))
 	}()
 
+	usesNegBackends := lc.shouldUseNEGBackends(service)
+
 	l4NetLBParams := &loadbalancers.L4NetLBParams{
 		Service:                          service,
 		Cloud:                            lc.ctx.Cloud,
@@ -543,10 +545,10 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
 		NetworkResolver:                  lc.networkResolver,
 		EnableWeightedLB:                 lc.ctx.EnableWeightedL4NetLB,
 		DisableNodesFirewallProvisioning: lc.ctx.DisableL4LBFirewall,
+		UseNEGs:                          usesNegBackends,
 	}
 	l4netlb := loadbalancers.NewL4NetLB(l4NetLBParams, svcLogger)
 
-	usesNegBackends := lc.shouldUseNEGBackends(service)
 	finalizer := common.NetLBFinalizerV2
 	if usesNegBackends {
 		finalizer = common.NetLBFinalizerV3
@@ -819,5 +821,7 @@ func (lc *L4NetLBController) publishSyncMetrics(result *loadbalancers.L4NetLBSyn
 	l4metrics.PublishL4SyncDetails(l4NetLBControllerName, result.Error == nil, isResync, result.GCEResourceUpdate.WereAnyResourcesModified())
 
 	isWeightedLB := result.MetricsState.WeightedLBPodsPerNode
-	l4metrics.PublishNetLBSyncMetrics(result.Error == nil, result.SyncType, result.GCEResourceInError, utils.GetErrorType(result.Error), result.StartTime, isResync, isWeightedLB)
+	backendType := result.MetricsState.BackendType
+
+	l4metrics.PublishNetLBSyncMetrics(result.Error == nil, result.SyncType, result.GCEResourceInError, utils.GetErrorType(result.Error), result.StartTime, isResync, isWeightedLB, backendType)
 }
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -94,7 +94,7 @@ func NewL4ILBSyncResult(syncType string, startTime time.Time, svc *corev1.Servic
 		StartTime:   startTime,
 		SyncType:    syncType,
 		// Internal Load Balancer doesn't support strong session affinity (passing `false` all along)
-		MetricsState: metrics.InitServiceMetricsState(svc, &startTime, isMultinetService, enabledStrongSessionAffinity, isWeightedLBPodsPerNode),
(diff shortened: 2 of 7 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -539,10 +545,15 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
...
 		EnableWeightedLB:                 lc.ctx.EnableWeightedL4NetLB,
 		DisableNodesFirewallProvisioning: lc.ctx.DisableL4LBFirewall,
+		UseNEGs:                          usesNegBackends,
 	}
 	l4netlb := loadbalancers.NewL4NetLB(l4NetLBParams, svcLogger)
 
-	if err := common.EnsureServiceFinalizer(service, common.NetLBFinalizerV2, lc.ctx.KubeClient, svcLogger); err != nil {
+	finalizer := common.NetLBFinalizerV2
+	if usesNegBackends {
+		finalizer = common.NetLBFinalizerV3
```
FelipeYepez, line 554: This change would require us to update the old controller to verify that it detects it's an RBS if it has V3 finalizer:
https://github.com/kubernetes/cloud-provider-gcp/blob/master/providers/gce/gce_util.go#L409

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -796,5 +821,6 @@ func (lc *L4NetLBController) publishSyncMetrics(result *loadbalancers.L4NetLBSyn
 	l4metrics.PublishL4SyncDetails(l4NetLBControllerName, result.Error == nil, isResync, result.GCEResourceUpdate.WereAnyResourcesModified())
 
 	isWeightedLB := result.MetricsState.WeightedLBPodsPerNode
-	l4metrics.PublishNetLBSyncMetrics(result.Error == nil, result.SyncType, result.GCEResourceInError, utils.GetErrorType(result.Error), result.StartTime, isResync, isWeightedLB)
+
+	l4metrics.PublishNetLBSyncMetrics(result.Error == nil, result.SyncType, result.GCEResourceInError, utils.GetErrorType(result.Error), result.StartTime, isResync, isWeightedLB, result.MetricsState.BackendType)
```
FelipeYepez, line 825: Nit: save result.MetricsState.BackendType in a variable and use the variable as argument

pkg/loadbalancers/l4.go
```diff
@@ -94,7 +94,7 @@ func NewL4ILBSyncResult(syncType string, startTime time.Time, svc *corev1.Servic
 		StartTime:   startTime,
 		SyncType:    syncType,
 		// Internal Load Balancer doesn't support strong session affinity (passing `false` all along)
-		MetricsState: metrics.InitServiceMetricsState(svc, &startTime, isMultinetService, enabledStrongSessionAffinity, isWeightedLBPodsPerNode),
+		MetricsState: metrics.InitServiceMetricsState(svc, &startTime, isMultinetService, enabledStrongSessionAffinity, isWeightedLBPodsPerNode, metrics.L4BackendTypeNEG),
```
FelipeYepez, line 97: Consider creating a variable for metrics.L4BackendTypeNEG

pkg/metrics/l4_metrics.go
```diff
@@ -47,7 +48,7 @@ var (
 			Name: "l4_netlbs_count",
 			Help: "Metric containing the number of NetLBs that can be filtered by feature labels and status",
 		},
-		[]string{l4LabelStatus, l4LabelMultinet, l4LabelStrongSessionAffinity, l4LabelWeightedLBPodsPerNode},
+		[]string{l4LabelStatus, l4LabelMultinet, l4LabelStrongSessionAffinity, l4LabelWeightedLBPodsPerNode, l4LabelNetLBBackendType},
```
FelipeYepez, line 51: It's redundant to say NetLB when adding l4LabelNetLBBackendType as a NetLB metric label

pkg/utils/common/finalizer.go
```diff
@@ -125,3 +127,27 @@ func EnsureDeleteServiceFinalizer(service *corev1.Service, key string, kubeClien
...
+		return nil
+	}
+
+	// Make a copy of object metadata so we don't mutate the shared informer cache.
+	updatedObjectMeta := service.ObjectMeta.DeepCopy()
+	for _, key := range needToRemove {
+		updatedObjectMeta.Finalizers = slice.RemoveString(updatedObjectMeta.Finalizers, key, nil)
+	}
+
+	svcLogger.V(2).Info("Removing finalizers from service", "finalizerKeys", ensureRemoveKeys)
```
FelipeYepez, line 151: Replace ensureRemoveKeys with needToRemove

FelipeYepez: Instead of having 2 for loops you can combine both in one:
```suggestion
	updatedObjectMeta := service.ObjectMeta.DeepCopy()
	modified := false 

	for _, key := range ensureRemoveKeys {
		if HasGivenFinalizer(service.ObjectMeta, key) {
			updatedObjectMeta.Finalizers = slice.RemoveString(updatedObjectMeta.Finalizers, key, nil)
			svcLogger.V(2).Info("Removing finalizer from service", "finalizerKey", key)
			modified = true
		}
	}

	if !modified {
		return nil
	}

	return patch.PatchServiceObjectMetadata(kubeClient.CoreV1(), service, *updatedObjectMeta)
```

pkg/loadbalancers/l4.go
mmamczur: this a constant, I'd say we don't need a variable to hold it

pkg/utils/common/finalizer.go
mmamczur: this was actually part of the parent commit that was merged yesterday

pkg/l4lb/l4netlbcontroller.go
mmamczur: you are right, we should add it. IT shouldn't be a problem unless someone removes the rbs annotation so we should update CCM anyway

pkg/l4lb/l4netlbcontroller.go
mmamczur: done

mmamczur:
/hold
need to make sure to update monarch schema with the new field before merging

swetharepakula (commented):
/lgtm
/approve

PR #2692, opened 2024-09-30
Remove unused IngClassInformer and IngParamsInformer from context

panslava:
Looks a bit scary, but, I couldn't find any places those informers are used (beside at HasSynced())... Maybe someone has more context?

Also I am not sure if ingparam is still needed, and if this part of the code is needed

```
ingParamsCRDMeta := ingparams.CRDMeta()
		if _, err := crdHandler.EnsureCRD(ingParamsCRDMeta, false); err != nil {
			klog.Fatalf("Failed to ensure GCPIngressParams CRD: %v", err)
		}
```


gauravkghildiyal:
Thanks for the cleanup @panslava 

Swetha helped shed some light on this and informed that this thing with IngParams were not productionized. So we should be fine with removing this. Same thing about creating the CRD, don't think we need that either.

/lgtm
/hold (In case you feel comfortable with removing the EnsureCRD as well

panslava:
Thank you @gauravkghildiyal !

Removed all ingParams functionality, basically deleted the folder with CRD and removed ensuring from main

Could you take a look again, please :)

panslava:
/unhold 

@gauravkghildiyal if possible, needs new LGTM as new changes were introduced

PR #2694, opened 2024-10-01
Make NodeTopologyInformer only looks for cluster-wide CR

sawsa307:
/assign @gauravkghildiyal 
/assign @swetharepakula

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -230,7 +230,9 @@ func NewControllerContext(
 
 	if flags.F.EnableMultiSubnetClusterPhase1 {
 		if nodeTopologyClient != nil {
-			context.NodeTopologyInformer = informernodetopology.NewNodeTopologyInformer(nodeTopologyClient, config.ResyncPeriod, utils.NewNamespaceIndexer())
+			context.NodeTopologyInformer = informernodetopology.NewFilteredNodeTopologyInformer(nodeTopologyClient, config.ResyncPeriod, utils.NewNamespaceIndexer(), func(listOptions *metav1.ListOptions) {
+				listOptions.FieldSelector = fmt.Sprintf("metadata.name=%s", flags.F.NodeTopologyCRName)
+			})
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


sawsa307:
Discussed with @swetharepakula, we will make node topology CR name as a flag just in case we need to modify in the future.

pkg/context/context.go
```diff
@@ -230,7 +230,9 @@ func NewControllerContext(
 
 	if flags.F.EnableMultiSubnetClusterPhase1 {
 		if nodeTopologyClient != nil {
-			context.NodeTopologyInformer = informernodetopology.NewNodeTopologyInformer(nodeTopologyClient, config.ResyncPeriod, utils.NewNamespaceIndexer())
+			context.NodeTopologyInformer = informernodetopology.NewFilteredNodeTopologyInformer(nodeTopologyClient, config.ResyncPeriod, utils.NewNamespaceIndexer(), func(listOptions *metav1.ListOptions) {
+				listOptions.FieldSelector = fmt.Sprintf("metadata.name=%s", flags.F.NodeTopologyCRName)
```
mmamczur, line 234: so we'll have informer for one resource with this filter, right?

The only thing here to maybe figure out is to know what would happen if that CRD is not present in the cluster for whatever reason. Failing to start the controller is fine if we're ok with it but it's best to know what happens

gauravkghildiyal: Relevant reply from David regarding this dependency ordering: https://github.com/kubernetes/ingress-gce/pull/2675#issuecomment-2357534843

+1 to failing to start until CRD is not available (which may be possible when the cluster gets created for the first time  and NEG controller wins the race)

swetharepakula: One concern however is for any reason the component that ensures the CRD has to be rolled back, the NEG controller will also be stuck and require a flag change to start working (turning of MSC flags).

I think we need to think about have a fail-safe and make sure neg controller is able to do best effort and continue processing as if the flag was disabled (so only default subnet). We would also need a way to trigger allowing MSC again later when the CRD and CR is available

PR #2695, opened 2024-10-02
IPv6 NetLB - disallow using on standard tier network

08volt:
BUG FIX:

NetLB does not support IPv6 on standard tier network.
Whenever this config was used the real setup was silently using premium tier.


pkg/loadbalancers/forwarding_rules_ipv6.go
```diff
@@ -169,6 +169,12 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 	netTier, isFromAnnotation := utils.GetNetworkTier(l4netlb.Service)
 	frLogger.V(2).Info("network tier for service", "networkTier", netTier, "isFromAnnotation", isFromAnnotation)
 
+	// IPv6 address is not supported for External Regional Network Load Balancing with Standard network tier.
```
mmamczur, line 172: add a test case for this

mmamczur: could even be something like https://github.com/kubernetes/ingress-gce/blob/a849580c343cad39882dbd4c504a12c33313bb99/pkg/loadbalancers/l4netlb_test.go#L738

so also you'd verify if it's user error (I see from you code that it will be but let's have a test)

08volt: fixed, thanks

PR #2697, opened 2024-10-08
Cherry pick the NEG syncer fix for the race condition when rapidly delete and create services.

ruixiansong:
* Ensure syncer for all the ports in newPorts to avoid syncers failed to start never starting again.
* Remove the ensure svcneg CR logic since it is included in ensure syncer.
/assgin @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #2700, opened 2024-10-11
IPv6 NetLB - disallow using on standard tier network

08volt:
BUG FIX:

NetLB does not support IPv6 on standard tier network.
Whenever this config was used the real setup was silently using premium tier.

### DIFF
--- a/pkg/loadbalancers/l4netlb_test.go
+++ b/pkg/loadbalancers/l4netlb_test.go
@@ -442,17 +442,11 @@ func TestEnsureDualStackNetLBNetworkTierChange(t *testing.T) {
 	svc.Annotations[annotations.NetworkTierAnnotationKey] = "Standard"
 	l4NetLB := mustSetupNetLBTestHandler(t, svc, nodeNames)
 
-	// Ensure dualstack load balancer with Standard Network Tier and verify it synced successfully.
+	// Ensure dualstack load balancer with Standard Network Tier and verify it did not synced successfully.
 	result := l4NetLB.EnsureFrontend(nodeNames, svc)
-	if result.Error != nil {
-		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+	if _, ok := result.Error.(*utils.UnsupportedNetworkTierError); !ok {
+		t.Errorf("Expected error to be of type *utils.UnsupportedNetworkTierError, got %T", result.Error)
 	}
-	if len(result.Status.Ingress) == 0 {
-		t.Errorf("Got empty loadBalancer status using handler %v", l4NetLB)
-	}
-	l4NetLB.Service.Annotations = result.Annotations
-	l4NetLB.Service.Annotations[annotations.NetworkTierAnnotationKey] = "Standard"
-	assertDualStackNetLBResources(t, l4NetLB, nodeNames)
 
 	// Change network Tier to Premium, and trigger sync.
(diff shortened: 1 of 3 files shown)


pkg/loadbalancers/l4netlb_test.go
```diff
@@ -442,17 +442,11 @@ func TestEnsureDualStackNetLBNetworkTierChange(t *testing.T) {
...
-	// Ensure dualstack load balancer with Standard Network Tier and verify it synced successfully.
+	// Ensure dualstack load balancer with Standard Network Tier and verify it did not synced successfully.
 	result := l4NetLB.EnsureFrontend(nodeNames, svc)
-	if result.Error != nil {
-		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
-	}
-	if len(result.Status.Ingress) == 0 {
-		t.Errorf("Got empty loadBalancer status using handler %v", l4NetLB)
+	if result.Error == nil {
+		t.Errorf("LoadBalancer was provisioned with wrong Network Tier")
```
mmamczur, line 448: could we check here we get the expected error and explain that we wanted error X bug did not get it?

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -792,6 +786,71 @@ func TestDualStackNetLBBadCustomSubnet(t *testing.T) {
...
+			if tc.dualstack {
+				svc.Spec.IPFamilies = append(svc.Spec.IPFamilies, v1.IPv6Protocol)
+			}
+
+			result := l4NetLB.EnsureFrontend(nodeNames, svc)
+			if tc.returnError && result.Error == nil {
+				t.Fatalf("Expected an error to be returned when ensuring the external load balancer, but the call succeeded unexpectedly.")
+			}
+			if tc.returnError && !utils.IsUserError(result.Error) {
+				t.Fatalf("Expected to get user error if ensuring external IPv6 service, got %v", result.Error)
```
mmamczur, line 848: nit: `error WHEN ensuring`

but actually I'd drop the second part of this sentence, "Expected to get user error but got: %v`

also the same above, "expected an error but got none" or "expected an error but the call succeeded"

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -792,6 +786,71 @@ func TestDualStackNetLBBadCustomSubnet(t *testing.T) {
...
+		},
+		{
+			desc:        "Should not return error on ipv4 with Premium NetworkTier",
+			dualstack:   false,
+			networkTier: string(cloud.NetworkTierPremium),
+			returnError: false,
+		},
+		{
+			desc:        "Should return error on Dualstack with Standard NetworkTier",
+			dualstack:   false,
```
mmamczur, line 813: why is `dualstack: false` here and in the case below?

if the test passes with this value, the logic is wrong somewhere

mmamczur: also, you could consider making this a field to directly pass the IP families and test both when its IPv4_IPv6 and IPv6 only

cezarygerard:
what about existing services?

pkg/loadbalancers/l4netlb_test.go
08volt: Fixed

mmamczur:
> what about existing services?

apparently there are no services like this. but maybe we could confirm this with our queries

PR #2702, opened 2024-10-18
Create additional NEGs based on subnets in Node Topology CR.

sawsa307:
* Query Node Topology CR for the current set of NEGs in the cluster.
* When ensureNetworkEndpointGroups(), ensure NEGs are properly provisioned in the non-default subnets as well.

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -46,6 +46,7 @@ import (
 	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/common"
+	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/ingress-gce/pkg/utils/zonegetter"
 	"k8s.io/klog/v2"
@@ -64,6 +65,7 @@ func (k serviceKey) Key() string {
 // syncerManager contains all the active syncer goroutines and manage their lifecycle.
 type syncerManager struct {
 	namer      negtypes.NetworkEndpointGroupNamer
+	l4Namer    namer.L4ResourcesNamer
 	recorder   record.EventRecorder
 	cloud      negtypes.NetworkEndpointGroupCloud
 	zoneGetter *zonegetter.ZoneGetter
@@ -118,6 +120,7 @@ type syncerManager struct {
 }
 
 func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer,
+	l4Namer namer.L4ResourcesNamer,
 	recorder record.EventRecorder,
 	cloud negtypes.NetworkEndpointGroupCloud,
 	zoneGetter *zonegetter.ZoneGetter,
@@ -140,6 +143,7 @@ func newSyncerManager(namer negtypes.NetworkEndpointGroupNamer,
 
 	return &syncerManager{
 		namer:               namer,
+		l4Namer:             l4Namer,
 		recorder:            recorder,
 		cloud:               cloud,
 		zoneGetter:          zoneGetter,
@@ -228,6 +232,11 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				&portInfo.NetworkInfo,
 				portInfo.L4LBType,
 			)
+			nonDefaultSubnetNEGNamer := manager.namer
+			if syncerKey.NegType == negtypes.VmIpEndpointType {
+				nonDefaultSubnetNEGNamer = manager.l4Namer
+			}
+
 			syncer = negsyncer.NewTransactionSyncer(
 				syncerKey,
 				manager.recorder,
@@ -243,12 +252,12 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				string(manager.kubeSystemUID),
 				manager.svcNegClient,
 				manager.syncerMetrics,
-				!manager.namer.IsNEG(portInfo.NegName),
+				syncerKey.NegType == negtypes.VmIpPortEndpointType && !manager.namer.IsNEG(portInfo.NegName),
 				manager.logger,
 				manager.lpConfig,
 				manager.enableDualStackNEG,
 				portInfo.NetworkInfo,
-				manager.namer,
+				nonDefaultSubnetNEGNamer,
 			)
 			manager.syncerMap[syncerKey] = syncer
 		}
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -248,6 +248,7 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				manager.lpConfig,
 				manager.enableDualStackNEG,
 				portInfo.NetworkInfo,
+				manager.namer,
 			)
 			manager.syncerMap[syncerKey] = syncer
 		}
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -130,6 +130,8 @@ type transactionSyncer struct {
 	// networkInfo contains the network information to use in GCP resources (VPC URL, Subnetwork URL).
 	// and the k8s network name (can be used in endpoints calculation).
 	networkInfo network.NetworkInfo
+
+	namer negtypes.NetworkEndpointGroupNamer
 }
 
 func NewTransactionSyncer(
@@ -152,6 +154,7 @@ func NewTransactionSyncer(
 	lpConfig labels.PodLabelPropagationConfig,
 	enableDualStackNEG bool,
 	networkInfo network.NetworkInfo,
+	namer negtypes.NetworkEndpointGroupNamer,
 ) negtypes.NegSyncer {
 
 	logger := log.WithName("Syncer").WithValues("service", klog.KRef(negSyncerKey.Namespace, negSyncerKey.Name), "negName", negSyncerKey.NegName)
@@ -182,6 +185,7 @@ func NewTransactionSyncer(
 		enableDualStackNEG:        enableDualStackNEG,
 		podLabelPropagationConfig: lpConfig,
 		networkInfo:               networkInfo,
+		namer:                     namer,
 	}
 	// Syncer implements life cycle logic
 	syncer := newSyncer(negSyncerKey, serviceLister, recorder, ts, logger)
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"google.golang.org/api/googleapi"
 	apiv1 "k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -425,7 +426,6 @@ func (s *transactionSyncer) resetErrorState() {
 
 // ensureNetworkEndpointGroups ensures NEGs are created and configured correctly in the corresponding zones.
 func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
-	var err error
 	// NEGs should be created in zones with candidate nodes only.
 	zones, err := s.zoneGetter.ListZones(negtypes.NodeFilterForEndpointCalculatorMode(s.EpCalculatorMode), s.logger)
 	if err != nil {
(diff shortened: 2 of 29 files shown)


pkg/utils/zonegetter/zone_getter.go
```diff
@@ -177,6 +187,39 @@ func (z *ZoneGetter) ListZones(filter Filter, logger klog.Logger) ([]string, err
...
+			logger.Error(err, "Failed to parse defaultSubnetURL", "defaultSubnetURL", z.defaultSubnetURL)
+			return nil, err
+		}
+		defaultSubnetName := resourceID.Key.Name
+		defaultSubnetPath := cloud.RelativeResourceName(resourceID.ProjectID, resourceID.Resource, resourceID.Key)
+		return []nodetopologyv1.SubnetConfig{{Name: defaultSubnetName, SubnetPath: defaultSubnetPath}}, nil
+	}
+
+	n, exists, err := z.nodeTopologyInformer.GetIndexer().GetByKey(flags.F.NodeTopologyCRName)
+	if err != nil {
```
gauravkghildiyal, line 209: nit: Use declared variable `nodeTopologyCRName`

pkg/neg/syncers/transaction.go
```diff
@@ -467,6 +473,13 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 		}
 	}
 
+	if flags.F.EnableMultiSubnetClusterPhase1 {
+		negs, needToUpdate, errs := s.ensureNetworkEndpointGroupsFromNodeTopology(zones, negsByLocation)
+		negObjRefs = append(negObjRefs, negs...)
+		updateNEGStatus = updateNEGStatus && needToUpdate
```
gauravkghildiyal, line 479: Should this be an OR? `updateNEGStatus || needToUpdate`

pkg/neg/syncers/transaction.go
```diff
@@ -475,6 +488,88 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+			if len(s.NegSyncerKey.NegName) > 56 {
+				s.logger.Error(err, "Unable to create custom NEGs in non-default subnet", "customNegName", s.NegSyncerKey.Name)
+				errList = append(errList, ErrCustomNEGNameTooLong)
+				continue
+			}
+			negName = fmt.Sprintf("%s-%s", s.NegSyncerKey.Name, namer.SubnetHash(subnetConfig.Name))
+		}
+
+		// Use a networkInfo with non-default subnet as subnetURL.
+		networkInfoInNonDefaultSubnet := s.networkInfo
```
gauravkghildiyal, line 527: If we're deriving a networkInfo for the non-default subnet, also ensure that the "IsDefault" field is set to false. I know it is being used in your codepath, but better to have a consistent state of the NetworkInfo object.

pkg/neg/syncers/transaction.go
```diff
@@ -475,6 +488,88 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+		resourceID, err := cloud.ParseResourceURL(subnetConfig.SubnetPath)
+		if err != nil {
+			s.logger.Error(err, "Failed to parse subnet path", "subnetPath", subnetConfig.SubnetPath)
+			errList = append(errList, err)
+			continue
+		}
+
+		// Add compute and version GA prefix.
+		networkInfoInNonDefaultSubnet.SubnetworkURL = cloud.SelfLink(meta.VersionGA, resourceID.ProjectID, "subnetworks", resourceID.Key)
```
gauravkghildiyal, line 536: Can "subnetworks" be replaced with `resourceID.Resource`?

pkg/neg/syncers/utils.go
```diff
@@ -148,6 +150,7 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 			ServiceName: svcName,
 			Port:        port,
 		}
+		fmt.Printf("remove me: expectedDesc = %v \n", expectedDesc)
```
gauravkghildiyal, line 153: nit: cleanup

pkg/neg/syncers/transaction.go
```diff
@@ -475,6 +488,88 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	return utilerrors.NewAggregate(errList)
 }
 
+func (s *transactionSyncer) ensureNetworkEndpointGroupsFromNodeTopology(zones []string, negsByLocation map[string]int) ([]negv1beta1.NegObjectReference, bool, []error) {
```
gauravkghildiyal, line 491: Have we thought about refactoring this to the `ensureNetworkEndpointGroups` itself?

sawsa307 (commented):
Just addressed all the comments, @gauravkghildiyal could you take another look when you have time? Thank you!

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -177,6 +187,39 @@ func (z *ZoneGetter) ListZones(filter Filter, logger klog.Logger) ([]string, err
 	return zones.List(), nil
 }
 
+func (z *ZoneGetter) ListSubnets(logger klog.Logger) ([]nodetopologyv1.SubnetConfig, error) {
+	nodeTopologyCRName := flags.F.NodeTopologyCRName
+
+	nodeTopologyNotSynced := z.nodeTopologyInformer != nil && !z.nodeTopologyHasSynced()
```
gauravkghildiyal, line 193: It seems that this function is assuming that if it get's called, then `z.nodeTopologyInformer` will never be nil. So the "!= nil" check might be redundant?

Alternatively, we may want to check that `z.nodeTopologyInformer != nil` before executing the `GetIndexer()` part?

pkg/neg/syncers/transaction.go
gauravkghildiyal: Thanks for the clarification. You are right

gauravkghildiyal:
Also, is [this TODO](https://github.com/kubernetes/ingress-gce/blob/d712709781a63da1c8352c070cb6b4bb08835973/pkg/neg/syncers/utils.go#L699) something that we plan on fixing independently? I guess as long as the MSC flags are disabled, this shouldn't cause any problems but please do confirm. Thanks!

pkg/utils/zonegetter/zone_getter.go
sawsa307: Yes that sounds good, added nil check for nodeTopologyInformer.

sawsa307:
> Also, is [this TODO](https://github.com/kubernetes/ingress-gce/blob/d712709781a63da1c8352c070cb6b4bb08835973/pkg/neg/syncers/utils.go#L699) something that we plan on fixing independently? I guess as long as the MSC flags are disabled, this shouldn't cause any problems but please do confirm. Thanks!

Yes I'm planning to tackle those TODO in follow up PRs(like #2728), but if you would like, I can also include some of them in this PR as well.

pkg/utils/zonegetter/zone_getter.go
gauravkghildiyal: Hmm this seems pretty confusing to understand.

So we have `nodeTopologyNotSynced := z.nodeTopologyInformer != nil && !z.nodeTopologyHasSynced()`
And then we have another expression A=`z.nodeTopologyInformer == nil || nodeTopologyNotSynced`
Substituting value of nodeTopologyNotSynced in A, we have A=`z.nodeTopologyInformer == nil || (z.nodeTopologyInformer != nil && !z.nodeTopologyHasSynced())`
which simplifies to A=`z.nodeTopologyInformer == nil || !z.nodeTopologyHasSynced()`

But in the construction of `NewZoneGetter()`, this further simplifies to A=`!z.nodeTopologyHasSynced()`. 

Even logically thinking, the question that you're trying to ask yourself is that "can I make use of the NodeTopology client or not". The `nodeTopologyHasSynced()` within ZoneGetter should anyways be defined in such a way that it answers that question directly -- which seems to be what it's already doing.

I'd recomment thinking this through again and simplifying it in this manner if it's correct.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -109,7 +114,12 @@ func (z *ZoneGetter) ZoneAndSubnetForNode(name string, logger klog.Logger) (stri
 		nodeLogger.Error(err, "Failed to get subnet from node's LabelNodeSubnet")
 		return "", "", err
 	}
-	if z.onlyIncludeDefaultSubnetNodes {
+
+	nodeTopologyNotSynced := z.nodeTopologyInformer != nil && !z.nodeTopologyHasSynced()
+	if z.onlyIncludeDefaultSubnetNodes || z.nodeTopologyInformer == nil || nodeTopologyNotSynced {
+		if z.nodeTopologyInformer == nil || nodeTopologyNotSynced {
+			logger.Info("NodeTopologyInformer is not ready, fall back to use default subnet only")
```
gauravkghildiyal, line 121: I'd recommend

```
logger.Info("Falling back to only using default subnet", "z.onlyIncludeDefaultSubnetNodes", z.onlyIncludeDefaultSubnetNodes, <Rest of the variables influencing this decision>)
```

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -110,7 +114,11 @@ func (z *ZoneGetter) ZoneAndSubnetForNode(name string, logger klog.Logger) (stri
 		nodeLogger.Error(err, "Failed to get subnet from node's LabelNodeSubnet")
 		return "", "", err
 	}
-	if z.onlyIncludeDefaultSubnetNodes {
+
+	if z.onlyIncludeDefaultSubnetNodes || !z.nodeTopologyHasSynced() {
+		if !z.nodeTopologyHasSynced() {
+			logger.Info("Falling back to only using default subnet", "z.onlyIncludeDefaultSubnetNodes", z.onlyIncludeDefaultSubnetNodes)
```
swetharepakula, line 120: since the same log line is appearing in multiple places we should provide some context of where the log is coming from

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -178,6 +186,42 @@ func (z *ZoneGetter) ListZones(filter Filter, logger klog.Logger) ([]string, err
...
+	}
+	if !exists {
+		return nil, fmt.Errorf("node topology CR %s is not in store", nodeTopologyCRName)
+	}
+
+	nodeTopologyCR, ok := n.(*nodetopologyv1.NodeTopology)
+	if !ok {
+		return nil, fmt.Errorf("failed to cast %v to node topology type", n)
+	}
+	return nodeTopologyCR.Status.Subnets, nil
```
swetharepakula, line 222: will the subnets always include the default subnet?

pkg/neg/syncers/transaction.go
```diff
@@ -436,38 +439,87 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+
+	// Get default subnet from syncer's networkInfo.
+	var defaultSubnet string
+	defaultSubnet, err = utils.KeyName(s.networkInfo.SubnetworkURL)
+	if err != nil {
+		s.logger.Error(err, "Errored getting default subnet from NetworkInfo")
+		errList = append(errList, err)
+	}
+
+	subnetConfigs := []nodetopologyv1.SubnetConfig{{Name: defaultSubnet, SubnetPath: s.networkInfo.SubnetworkURL}}
```
swetharepakula, line 454: is this necessary, doesn't zoneGetter already take care of evaluating whether only the default subnet should be returned or all?

pkg/neg/syncers/transaction.go
```diff
@@ -436,38 +439,87 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+		if subnetConfig.Name != defaultSubnet {
+			// Determine the NEG name for the non-default subnet NEGs.
+			negName = s.namer.NonDefaultSubnetNEG(s.NegSyncerKey.Namespace, s.NegSyncerKey.Name, subnetConfig.Name, s.NegSyncerKey.PortTuple.Port)
+			if s.customName {
+				if len(s.NegSyncerKey.NegName) > negtypes.MaxDefaultSubnetNegNameLength {
+					s.logger.Error(ErrCustomNEGNameTooLong, "Unable to create custom NEGs in non-default subnet", "customNegName", s.NegSyncerKey.Name)
+					errList = append(errList, ErrCustomNEGNameTooLong)
+					continue
+				}
+				negName = fmt.Sprintf("%s-%s", s.NegSyncerKey.Name, namer.SubnetHash(subnetConfig.Name))
```
swetharepakula: can we move this logic to the neg namer

pkg/neg/syncers/transaction.go
```diff
@@ -134,7 +134,8 @@ type transactionSyncer struct {
 	// and the k8s network name (can be used in endpoints calculation).
 	networkInfo network.NetworkInfo
 
-	namer negtypes.NetworkEndpointGroupNamer
+	namer   negtypes.NetworkEndpointGroupNamer
+	l4Namer namer.L4ResourcesNamer
```
swetharepakula, line 138: do we need both. the namer should be L4 / L7 agnostic. The right namer should be set on transactionSyncer creation

pkg/neg/syncers/transaction.go
```diff
@@ -939,6 +939,23 @@ func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.Endp
...
+		negNamer = s.l4Namer
+	}
+	negName := negNamer.NonDefaultSubnetNEG(s.NegSyncerKey.Namespace, s.NegSyncerKey.Name, subnet, s.NegSyncerKey.PortTuple.Port)
+	if s.customName {
+		if len(s.NegSyncerKey.NegName) > negtypes.MaxDefaultSubnetNegNameLength {
+			s.logger.Error(ErrCustomNEGNameTooLong, "Unable to generate NEG name for custom NEGs in non-default subnet", "customNegName", s.NegSyncerKey.Name)
+			return "", ErrCustomNEGNameTooLong
+		}
+		negName = fmt.Sprintf("%s-%s", s.NegSyncerKey.Name, namer.SubnetHash(subnet))
+	}
```
swetharepakula, line 955: this should be moved to the namer. We should hav all the naming scheme information in one place.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -110,7 +114,11 @@ func (z *ZoneGetter) ZoneAndSubnetForNode(name string, logger klog.Logger) (stri
 		nodeLogger.Error(err, "Failed to get subnet from node's LabelNodeSubnet")
 		return "", "", err
 	}
-	if z.onlyIncludeDefaultSubnetNodes {
+
+	if z.onlyIncludeDefaultSubnetNodes || !z.nodeTopologyHasSynced() {
```
swetharepakula, line 118: I wonder if there is a better way to do this. Like one the zone getter gets into this mode it is always in this mode?

pkg/utils/zonegetter/zone_getter.go
sawsa307: Updated, thanks!

pkg/utils/zonegetter/zone_getter.go
sawsa307: Updated, now we will use nodeTopologyNotSynced as a dynamic expression to check if node topology informer is ready, and use onlyIncludeDefaultSubnetNodes as an static expression based on the flags.

pkg/neg/syncers/transaction.go
sawsa307: Updated, thanks!

pkg/neg/syncers/transaction.go
sawsa307: We need both since they have different interface and naming schema.
For L4 NEGs, the NEG name is determined by L4Backend(), while NEG() is only for L7 NEGs. As a result, I also separated the naming schema for L4 and L7 non-default subnet NEGs as L4NonDefaultSubnetNEG and NonDefaultSubnetNEG.

pkg/neg/syncers/transaction.go
sawsa307: Updated, thanks!

pkg/neg/syncers/transaction.go
sawsa307: Updated, now the zoneGetter will take care of the logic of listSubnets and also returning default subnet.

sawsa307:
@gauravkghildiyal @swetharepakula This PR is ready for another round of review, please take a look, thank you!

pkg/neg/syncers/transaction.go
gauravkghildiyal: Interestingly, because of this complication of having two separate L4 and L7 namers, even the existing code has a bug at: https://github.com/kubernetes/ingress-gce/blob/169b2c0d175c40566a921a630a43c6d8ec705db7/pkg/neg/manager.go#L246

where this value determines the value of `customName bool`.  The way our new logic is, this should lead to `customName` always being true for L4 NEGs and thus it would end up using the L7 custom namer for L4 NEGs.

Do we want to make L4 and L7 namers implement the following?

```
	NonDefaultSubnetNEG(namespace, name, subnetName string, port int32) string

	NonDefaultSubnetCustomNEG(customNEGName, subnetName string) (string, error)
```

- The `port` value will not be used for L4
- Because we will set customNEG for L4 to always be false, the L4 `NonDefaultSubnetCustomNEG` should also never be used.

gauravkghildiyal:
Github is just the worst at showing newly added comments so sharing the links:

- https://github.com/kubernetes/ingress-gce/pull/2702#discussion_r1847504199
- https://github.com/kubernetes/ingress-gce/pull/2702#discussion_r1847608077

sawsa307:
> Github is just the worst at showing newly added comments so sharing the links:
> 
> * [Create additional NEGs based on subnets in Node Topology CR. #2702 (comment)](https://github.com/kubernetes/ingress-gce/pull/2702#discussion_r1847504199)
> * [Create additional NEGs based on subnets in Node Topology CR. #2702 (comment)](https://github.com/kubernetes/ingress-gce/pull/2702#discussion_r1847608077)

Addressed both comments, please take a look! @gauravkghildiyal

pkg/neg/syncers/transaction.go
```diff
@@ -933,20 +933,18 @@ func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.Endp
...
 		}
 		return negName, nil
 	}
-
-	negName := negNamer.NonDefaultSubnetNEG(s.NegSyncerKey.Namespace, s.NegSyncerKey.Name, subnet, s.NegSyncerKey.PortTuple.Port)
-	return negName, nil
+	if s.NegType == negtypes.VmIpEndpointType {
+		return s.l4Namer.L4NonDefaultSubnetNEG(s.NegSyncerKey.Namespace, s.NegSyncerKey.Name, subnet), nil
+	} else {
+		return s.namer.NonDefaultSubnetNEG(s.NegSyncerKey.Namespace, s.NegSyncerKey.Name, subnet, s.NegSyncerKey.PortTuple.Port), nil
```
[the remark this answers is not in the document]
sawsa307: Just created another commit to consolidate the namer from syncer initialization https://github.com/kubernetes/ingress-gce/pull/2702/commits/21218630c257b798f9aea481693d41a28a92958d

pkg/neg/syncers/transaction.go
```diff
@@ -883,6 +928,19 @@ func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.Endp
 	}
 }
 
+// getNonDefaultSubnetName returns the name of the NEG based on the subnet name.
+func (s *transactionSyncer) getNonDefaultSubnetName(subnet string) (string, error) {
+	if s.customName {
```
gauravkghildiyal, line 933: Don't we still need to adjust the value stored in `s.customName` such that for L4 NEGs, this is always false?

pkg/utils/namer/interfaces.go
```diff
@@ -60,6 +62,9 @@ type BackendNamer interface {
 	// RXLBBackendName returns the Regional External Ingress backend name,
 	// based on the service namespace, name and target port.
 	RXLBBackendName(namespace, name string, port int32) string
+	// NonDefaultSubnetCustomNEG returns the gce neg name for custom NEGs created
+	// in non-default subnets.
+	NonDefaultSubnetCustomNEG(customNEGName, subnetName string) (string, error)
```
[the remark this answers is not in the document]
sawsa307: Updated, thanks!

pkg/utils/namer/l4_namer.go
```diff
@@ -79,6 +80,15 @@ func (namer *L4Namer) L4NonDefaultSubnetNEG(namespace, name, subnetName string)
 	}, "-")
 }
 
+// NonDefaultSubnetCustomNEG returns the gce neg name in the non-default subnet
+// when the NEG name is a custom one.
+func (n *L4Namer) NonDefaultSubnetCustomNEG(customNEGName, subnetName string) (string, error) {
```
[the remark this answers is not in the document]
sawsa307: Updated, thanks!

pkg/neg/manager.go
```diff
@@ -228,6 +232,11 @@ func (manager *syncerManager) EnsureSyncers(namespace, name string, newPorts neg
 				&portInfo.NetworkInfo,
 				portInfo.L4LBType,
 			)
+			negNamer := manager.namer
```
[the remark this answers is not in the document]
sawsa307: Updated, thanks!

sawsa307 (commented):
Addressed all the comments, thank you!

sawsa307:
Did a squash over the previous fix-up to make it cleaner, thanks!

gauravkghildiyal:
/lgtm

Thank you for the patience!

PR #2703, opened 2024-10-22
Implement namespaced informer

panslava:
- This informer satisfies interface of normal informer
- This informer will react to events happening only for objects in specific namespace
- This informer will List only objects in specific namespace

Not used yet, will be used in next PRs


panslava:
For reviewers -- I am open for suggestions of folder/file namings and structure, but that's the best I came up myself at the moment, I will keep multiproject stuff inside folder pkg/multiproject and split it by folder-packages inside

pkg/multiproject/namespacedinformer/namespacedindexer.go
```diff
@@ -0,0 +1,42 @@
...
+type namespacedIndexer struct {
+	cache.Indexer
+	namespace string
+}
+
+func (n *namespacedIndexer) List() []interface{} {
+	return namespaceFilteredList(n.Indexer.List(), n.namespace)
+}
+
+func (n *namespacedIndexer) ByIndex(indexName, indexKey string) ([]interface{}, error) {
```
gauravkghildiyal, line 17: nit: cache.Indexer.ByIndex seems to use `indexValue` instead of `indexKey`. It defines the concept of "indexed value". Should we use the same terminology?

pkg/multiproject/namespacedinformer/namespacedindexer.go
```diff
@@ -0,0 +1,42 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+// namespacedIndexer wraps an Indexer to provide namespace filtering.
+type namespacedIndexer struct {
+	cache.Indexer
+	namespace string
+}
+
+func (n *namespacedIndexer) List() []interface{} {
```
gauravkghildiyal, line 13: So it seems like this `List()` is actually implementing the function from the `Store` interface within the `cache.Indexer` interface. But on the other hand, we also have a different implementation of `namespacedstore`. I'm curious if this `List()` is actually needed? If it is, then it feels a bit weird that we re-implement one of the functions from the Store interface here while the rest are not re-implemented. Is this maybe revealing some slight flaw in how these structs are being defined?

panslava: You are right, I don't think `List` here is needed, removed

panslava:
Ready for re-review @gauravkghildiyal , thank you for the comments!

pkg/multiproject/namespacedinformer/namespacedstore.go
```diff
@@ -0,0 +1,47 @@
...
+		if key, err := cache.MetaNamespaceKeyFunc(item); err == nil {
+			if isObjectInNamespace(item, n.namespace) {
+				keys = append(keys, key)
+			}
+		}
+	}
+	return keys
+}
+
+func (n *namespacedStore) Get(obj interface{}) (item interface{}, exists bool, err error) {
```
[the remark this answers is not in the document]
gauravkghildiyal: Sorry didn't notice you are call GetByKey from it. Yes it absolutely makes sense to override it.

gauravkghildiyal (commented):
This mostly looks good to me. 

This may be just future foolproofing but wanted to see what your thoughts are about this:

If someone uses the `GetIndexer` function from our informer, they receive the custom `namespacedIndexer` implementation of the `Indexer` Interface. Given the fact that the `Indexer` interface also embeds the `Store` interface, a user is within their rights to use methods of the `Store` interface from an `Indexer` object. But in our case, I think we'd want to have those methods also be filtered (like we do in the case when someone explicitly fetches the `Store` using `GetStore()`.

Do we think this it's reasonable to have a solution for this peculiar behaviour right now? Taking a look at how the upstream implements Store and Indexer, it actually does this through a shared `cache` type which satisfies both interfaces: https://github.com/kubernetes/client-go/blob/c5146a9031d2c0dbe2a67dede62361ecf3d651bf/tools/cache/store.go#L282-L295. Borrowing the idea, we could also define a single struct which implements `Indexer` (which in turn would also implement `Store`)

panslava:
@gauravkghildiyal that's a very good thinking, thanks for digging into kubernetes code, I agree it will be better (and code actually cleaner) if we merge store and indexer into the same struct, I did that, and removed namespacedstore completely, now indexer and store is implemented inside namespacedcache struct and file

gauravkghildiyal:
Great, thanks!

/lgtm

PR #2705, opened 2024-10-22
Add Shared Context for multi-project controller

panslava:
- Not used yet
- Stores fields that are shared between projects and does not need to be created per project
- Will be used later to start controllers for multi-project scenarious


gauravkghildiyal:
Thanks!

While the change in isolation LGTM, it's possible that we may have to revisit something depending on the usage.

/lgtm

PR #2706, opened 2024-10-22
Fix port range to port comparison for forwarding rules

TheBeeZee:


pkg/loadbalancers/forwarding_rules.go
```diff
@@ -539,17 +539,16 @@ func Equal(fr1, fr2 *composite.ForwardingRule) (bool, error) {
...
+	// Existing forwarding rule contains a port range. To keep it that way,
+	// compare new list of ports as if it was a port range, too.
+	if len(newPorts) != 0 {
+		newPortRange = utils.MinMaxPortRange(newPorts)
 	}
-	if len(ports2) != 0 && portRange2 == "" {
-		portRange2 = utils.MinMaxPortRange(ports2)
-	}
-	return portRange1 == portRange2
+	return existingPortRange == newPortRange
```
FelipeYepez, line 551: Is there any reason why we are not comparing `utils.EqualStringSets(existingPorts, newPorts)`here?

TheBeeZee: At this point we are guaranteed that existingPorts is empty (see the check at the top of the function), which means that existing ports are specified as a port range, so we want to compare it as a port range. Remember, we don't want to auto-migrate users from port range to port list, because that would be disruptive to traffic.

If the existing port range covers the same ports as the new ports list, then the forwarding rule will continue working as-is without modifications.

FelipeYepez (changes requested):
/LGTM

pkg/loadbalancers/forwarding_rules.go
FelipeYepez: Thanks

pkg/loadbalancers/forwarding_rules.go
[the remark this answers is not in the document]
TheBeeZee: I have added a test case for exactly 5 ports as requested. The ExpectedPorts field is already checked to be empty - the checkForwardingRule() function compares both the port range and ports list.

I have also updated the variable names in Equal() to make it clear which one is the existing and new forwarding rule.

PR #2708, opened 2024-10-24
Fix port range to port comparison for forwarding rules

TheBeeZee:


FelipeYepez:
Thanks for implementing the changes requested in the other PR.

PR #2709, opened 2024-10-25
ILB - if wrong subnet specified mark as user error

anyush:
If non-existent subnetwork for internal load balancer specified using `networking.gke.io/internal-load-balancer-subnet` annotation mark resulting error as user one

### DIFF
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -677,11 +677,15 @@ func (l4 *L4) getServiceSubnetworkURL(options gce.ILBOptions) (string, error) {
 }
 
 func (l4 *L4) getSubnetworkURLByName(subnetName string) (string, error) {
-	subnetKey, err := l4.CreateKey(subnetName)
+	subnetwork, err := l4.cloud.GetSubnetwork(l4.cloud.Region(), subnetName)
 	if err != nil {
+		if utils.IsNotFoundError(err) {
+			return "", utils.NewInvalidSubnetConfigurationError(l4.cloud.ProjectID(), subnetName)
+		}
 		return "", err
 	}
-	return cloud.SelfLink(meta.VersionGA, l4.cloud.NetworkProjectID(), "subnetworks", subnetKey), nil
+
+	return subnetwork.SelfLink, nil
 }
 
 func (l4 *L4) hasAnnotation(annotationKey string) bool {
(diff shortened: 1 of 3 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -377,6 +377,18 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+	if options.SubnetName != "" {
+		_, err := l4.cloud.GetSubnetwork(l4.cloud.Region(), options.SubnetName)
+		if utils.IsNotFoundError(err) {
+			err = utils.NewInvalidSubnetConfigurationError(l4.cloud.ProjectID(), options.SubnetName)
+		}
+		if err != nil {
+			result.Error = err
+			return result
+		}
+	}
```
08volt, line 390: Can we use the SelfLink from `GetSubnetwork` directly instead of building it on line 392?  

Maybe refactor `getServiceSubnetworkURL` to handle the check internally and use the SelfLink field of the object returned from the API call

pkg/loadbalancers/l4_test.go
```diff
@@ -1315,6 +1329,32 @@ func TestDualStackILBBadCustomSubnet(t *testing.T) {
 	}
 }
 
+func TestInternalLBBadCustomSubnet(t *testing.T) {
+	t.Parallel()
+	nodeNames := []string{"test-node-1"}
+
+	t.Run("Specifying non-existing subnetwork should return error marked as user one", func(t *testing.T) {
```
08volt, line 1336: Create test cases ( see test line  924 ) with description and options so it will be easy to add new test cases in the future.

options can be
description (string)
isError (bool)
isUserError (bool)
expectedError (string)

I would have two test cases, one with a 'good' subnetwork name and one with 'bad' subnetwork name.

pkg/loadbalancers/l4.go
anyush: @08volt moved to `getSubnetworkURLByName` where SelfLink built previously

pkg/loadbalancers/l4_test.go
anyush: @08volt done. Slightly different structure: this test checks only specific error, so instead of `isError`, `isUserError` and `expectedError` added just `existingSubnetwork` (if `true` - no errors should be, otherwise - the same one for all cases) + `subnetName`

08volt: looks good!

PR #2710, opened 2024-10-25
[Cherry-pick #2708 -> 1.31] Fix port range to port comparison for forwarding rules

TheBeeZee:


mmamczur:
can we make sure the name of the PR and comming message follow the pattern `[Cherry-pick #2708 -> 1.31]`

Other than that lgtm

but I'm not sure how we plan to release, @swetharepakula will we do any more releases from ingress-gce release-1.31 branch?

FelipeYepez:
This fixes an issue present in release-1.31 so we need to cherry pick this

mmamczur:
> This fixes an issue present in release-1.31 so we need to cherry pick this

but if we cut a new branch 1.32 and use that then 1.31 may not be needed anymore. ingress-gce 1.32 will be used in gke 1.31.

FelipeYepez:
I see, thanks.

TheBeeZee:
> can we make sure the name of the PR and comming message follow the pattern `[Cherry-pick #2708 -> 1.31]`
> 
> Other than that lgtm
> 
> but I'm not sure how we plan to release, @swetharepakula will we do any more releases from ingress-gce release-1.31 branch?

I have updated the description. I'm ok with either cherrypicking into 1.31 or releasing as 1.32, whichever is the proper solution.

PR #2722, opened 2024-10-30
Refactor AddressManager into pkg/address

TortillaZHawaii:
This change shrinks the size of the `pkg/loadbalancers` by moving address related utilities to another package.


mmamczur:
looks good, let's merge this after the freeze

PR #2729, opened 2024-11-05
Add function that starts NEG controller per ProviderConfig

panslava:
- Will be used with ProviderConfig Controller, starting NEG controller as
  ProviderConfig EventHandler

based on other PR https://github.com/kubernetes/ingress-gce/pull/2755

### DIFF
--- a/pkg/multiproject/neg/neg.go
+++ b/pkg/multiproject/neg/neg.go
@@ -1,0 +1,231 @@
+package neg
+
+import (
+	"fmt"
+
+	networkclient "github.com/GoogleCloudPlatform/gke-networking-api/client/network/clientset/versioned"
+	informernetwork "github.com/GoogleCloudPlatform/gke-networking-api/client/network/informers/externalversions/network/v1"
+	nodetopologyclient "github.com/GoogleCloudPlatform/gke-networking-api/client/nodetopology/clientset/versioned"
+	informernodetopology "github.com/GoogleCloudPlatform/gke-networking-api/client/nodetopology/informers/externalversions/nodetopology/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/cloud-provider-gcp/providers/gce"
+	providerconfig "k8s.io/ingress-gce/pkg/apis/providerconfig/v1"
+	"k8s.io/ingress-gce/pkg/flags"
+	"k8s.io/ingress-gce/pkg/multiproject/filteredinformer"
+	multiprojectgce "k8s.io/ingress-gce/pkg/multiproject/gce"
+	"k8s.io/ingress-gce/pkg/neg"
+	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/network"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	informersvcneg "k8s.io/ingress-gce/pkg/svcneg/client/informers/externalversions/svcneg/v1beta1"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+	"k8s.io/klog/v2"
+)
+
+func StartNEGController(
+	informersFactory informers.SharedInformerFactory,
+	kubeClient kubernetes.Interface,
+	svcNegClient svcnegclient.Interface,
+	eventRecorderClient kubernetes.Interface,
+	networkClient networkclient.Interface,
+	nodeTopologyClient nodetopologyclient.Interface,
+	kubeSystemUID types.UID,
+	clusterNamer *namer.Namer,
+	l4Namer *namer.L4Namer,
+	lpConfig labels.PodLabelPropagationConfig,
+	defaultCloudConfig string,
+	globalStopCh <-chan struct{},
+	logger klog.Logger,
+	providerConfig *providerconfig.ProviderConfig,
+) (<-chan struct{}, error) {
+	cloud, err := multiprojectgce.NewGCEForProviderConfig(defaultCloudConfig, providerConfig, logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create GCE client for provider config %+v: %v", providerConfig, err)
+	}
+
+	providerConfigName := providerConfig.Name
+
+	// Using informer factory, create required namespaced informers for the NEG controller.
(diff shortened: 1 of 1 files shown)


pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get gkeNetworkParams informer: %v", err)
+	}
+	providerConfigFilteredGkeNetworkParamsInformer := filteredinformer.NewProviderConfigFilteredInformer(gkeNetworkParamsInformer.Informer(), providerConfigName)
+
+	nodeTopologyInformer, err := informersFactory.ForResource(nodeTopologyResource)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get nodeTopology informer: %v", err)
+	}
+	providerConfigFilteredNodeTopologyInformer := filteredinformer.NewProviderConfigFilteredInformer(nodeTopologyInformer.Informer(), providerConfigName)
```
gauravkghildiyal, line 84: Network, GKENetworkParamSet and NodeTopology are optional resources that may not always be present. This is more true for our case where these may be disabled. Let's not make these necessary prerequisite for starting the controller. You can borrow ideas from the existing code on how they are presently handled.

(It's very well possible we may need to extend this concept to make Ingress also optional but that would require some minor changes in the NEG Controller code and not worth doing right now)

pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,221 @@
...
+	nodeInformer := filteredinformer.NewProviderConfigFilteredInformer(informersFactory.Core().V1().Nodes().Informer(), providerConfigName)
+	endpointSliceInformer := filteredinformer.NewProviderConfigFilteredInformer(informersFactory.Discovery().V1().EndpointSlices().Informer(), providerConfigName)
+
+	svcNegInformer, err := informersFactory.ForResource(svcNegResource)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get svcNeg informer: %v", err)
+	}
+	providerConfigFilteredSvcNegInformer := filteredinformer.NewProviderConfigFilteredInformer(svcNegInformer.Informer(), providerConfigName)
+
+	networkInformer, err := informersFactory.ForResource(gkeNetworkParamsResource)
```
gauravkghildiyal, line 68: The other comment might make this one useless, but just so I don't forget, if we do keep this piece of code as is, there's a typo here since we're using the incorrect GVR

pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,221 @@
...
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/network"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+	"k8s.io/klog/v2"
+)
+
+var svcNegResource = schema.GroupVersionResource{Group: svcneg.SchemeGroupVersion.Group, Version: svcneg.SchemeGroupVersion.Version, Resource: "servicenetworkendpointgroups"}
```
gauravkghildiyal, line 30: nit: If the use of "Resource" sounds confusing, there's usual precedent to name such constants with the suffix "GVR", like "svcNegGVR" and so on.

pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,221 @@
...
+			providerConfigFilteredNetworkInformer.HasSynced() &&
+			providerConfigFilteredGkeNetworkParamsInformer.HasSynced() &&
+			providerConfigFilteredNodeTopologyInformer.HasSynced()
+	}
+
+	zoneGetter := zonegetter.NewZoneGetter(nodeInformer, providerConfigFilteredNodeTopologyInformer, cloud.SubnetworkURL())
+
+	// Create a channel to stop the controller for this specific provider config.
+	providerConfigStopCh := make(chan struct{})
+	defer close(providerConfigStopCh)
```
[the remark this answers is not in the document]
panslava: good catch, removed

pkg/multiproject/neg/neg.go
panslava: You are right, I think i had some types problems at some point, so I did GVR informers. But looks like now I can actually use just normal informers as we do in single project code. Reworked this part

pkg/multiproject/neg/neg.go
panslava: good catch, though, i just removed all GVR now

panslava:
I can squash commits after lgtm, if you prefer. Added fixes in the new commit for easier review

panslava:
ready for rereview @gauravkghildiyal

pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// joinedStopCh is a channel that will be closed when the global stop channel or the provider config stop channel is closed.
+	joinedStopCh := make(chan struct{})
+	go func() {
+		defer func() {
+			close(joinedStopCh)
+			logger.V(2).Info("NEG controller stop channel closed")
+		}()
+		select {
+		case <-globalStopCh:
```
gauravkghildiyal, line 119: Non-blocking: I'm curious if there's an advantage of making this part of the code aware of the `globalStopCh`. Because we are returning the caller of `StartNEGController` with the `providerConfigStopCh`, the caller could have just as easily handled closing `providerConfigStopCh` when `globalStopCh` gets closed, thereby making the code here unware of these coordination semantics. Anything that I'm missing?

gauravkghildiyal (commented):
Thanks for incorporating the changes.

Just one non-blocking comment. But besides that looks good.

pkg/multiproject/neg/neg.go
```diff
@@ -0,0 +1,231 @@
...
+	nodeTopologyClient nodetopologyclient.Interface,
+	kubeSystemUID types.UID,
+	clusterNamer *namer.Namer,
+	l4Namer *namer.L4Namer,
+	lpConfig labels.PodLabelPropagationConfig,
+	defaultCloudConfig string,
+	globalStopCh <-chan struct{},
+	logger klog.Logger,
+	providerConfig *providerconfig.ProviderConfig,
+) (chan struct{}, error) {
```
gauravkghildiyal, line 46: supernit: Maybe return a read-only channel

panslava: thank you, done!

pkg/multiproject/neg/neg.go
panslava: i thought about it, and i don't see real benefits of either approaches, so don't really want to change for now. Maybe it's just a bit easier to navigate what happens to the channels when the logic is in the same place....

panslava:
ready, squashed commits

pkg/multiproject/neg/neg.go
gauravkghildiyal: Seems like we mistakenly converted this to a write-only channel (which is definitely a no-no).

panslava: hehe, you are right, fixed

PR #2730, opened 2024-11-06
Remove unnecessary interfaces in pkg/backends

TortillaZHawaii:
As per [Go Wiki](https://go.dev/wiki/CodeReviewComments#interfaces) interfaces should generally belong in the package that uses values of the interface type, not the package that implements these values. In this case there is no real added value of using interfaces instead of concrete types. We can safely remove them and get rid of boilerplate.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -217,7 +217,8 @@ func (p *Pool) Delete(name string, version meta.Version, scope meta.KeyType, beL
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


mmamczur:
I does make the code follow go patterns a bit more, I'm adding Gaurav as well since this touches NEG pieces as well

/assign @gauravkghildiyal

gauravkghildiyal:
Thanks @TortillaZHawaii !

/approve
/assign @mmamczur

PR #2731, opened 2024-11-07
Extend retrieve current NEG endpoints to also look for NEGs in non-default subnets.

sawsa307:
/assign @gauravkghildiyal

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -270,8 +270,9 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		s.logger.Error(err, "Errored getting default subnet from NetworkInfo")
 		return err
 	}
+	subnetToNegMapping := map[string]string{defaultSubnet: s.NegSyncerKey.NegName}
 
-	currentMap, currentPodLabelMap, err := retrieveExistingZoneNetworkEndpointMap(s.NegSyncerKey.NegName, s.zoneGetter, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, defaultSubnet, s.logger)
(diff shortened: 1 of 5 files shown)


gauravkghildiyal:
Thanks!

Haven't checked the tests, but besides those, this one is looking relatively straightforward.

pkg/neg/syncers/transaction.go
```diff
@@ -883,6 +954,18 @@ func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.Endp
 	}
 }
 
+func (s *transactionSyncer) getNonDefaultSubnetName(subnet string) (string, error) {
```
mmamczur, line 957: this must be able to detect if it's running for L4 or L7 NEG and use the appropriate namer

sawsa307: Ahh I see, we can probably check the NEG type and use the correct namer(namer/l4namer). Both of them are available to NEG controller. Thanks for catching this!

gauravkghildiyal: Thanks for catching this!

pkg/neg/syncers/utils_test.go
```diff
@@ -891,6 +908,145 @@ func TestRetrieveExistingZoneNetworkEndpointMap(t *testing.T) {
...
+					"foo": "bar",
+				},
+				endpoint4: labels.PodLabelMap{
+					"foo": "bar",
+				},
+			},
+			expectErr: false,
+		},
+		{
+			desc: "2 empty negs in an addtional subnet, and 2 negs in the default subnet with multiple endpoints",
```
gauravkghildiyal, line 940: I didn't understand this description compared to what the test is doing or checking. Can you maybe add a comment about what that "2" figure is supposed to represent?

gauravkghildiyal (commented):
Almost ready!

pkg/neg/syncers/utils_test.go
sawsa307: Updated the description, thank you!

gauravkghildiyal (commented):
Thanks

/lgtm

PR #2732, opened 2024-11-07
Add flag for L4 Mixed Protocol feature

TortillaZHawaii:
Adds `EnableL4MixedProtocol` flag for mixed protocol feature. 

Also auto reformatted the `var ( F = ... )` to `var F = ...`.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -53,98 +53,97 @@ const (
 	DefaultLockObjectName = "ingress-gce-lock"
 )
 
-var (
-	// F are global flags for the controller.
-	F = struct {
-		APIServerHost                    string
-		ASMConfigMapBasedConfigCMName    string
-		ASMConfigMapBasedConfigNamespace string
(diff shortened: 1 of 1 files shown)


pkg/flags/flags.go
```diff
@@ -53,98 +53,97 @@ const (
...
+	EnableIGMultiSubnetCluster               bool
+	EnableMultiSubnetCluster                 bool
+	EnableMultiSubnetClusterPhase1           bool
+	NodeTopologyCRName                       string
+	EnableWeightedL4ILB                      bool
+	EnableWeightedL4NetLB                    bool
+	EnableDiscretePortForwarding             bool
+	EnableMultiProjectMode                   bool
+	MultiProjectCRDProjectNameLabel          string
+	EnableMixedProtocol                      bool
```
mmamczur, line 143: Let's put L4 in the name so it's clear it is only for L4 controllers, `EnableL4MixedProtocol`

PR #2733, opened 2024-11-07
ILB mixed protocol backend

TortillaZHawaii:
This PR allows the controller to create and migrate to L3 controllers. `backends.GetProtocol` will default to L3 protocol unless, there is already an existing service and ports that match the backend. For example if we have a running TCP backend and only services defined in the spec are TCP, it won't change protocol. However if we try to change from TCP to UDP, or mixed we will get L3. Once we are in L3 I don't think there is any reason to go out of it. `pkg/backends/protocol_test.go` contains an exhaustive list of scenarios and expected behavior.

~~In draft until the L4 Mixed Protocol feature flag is merged.~~

Alternative that does not take into account existing services can be found here: https://github.com/kubernetes/ingress-gce/pull/2739

### DIFF
--- a/pkg/backends/protocol.go
+++ b/pkg/backends/protocol.go
@@ -1,0 +1,63 @@
+package backends
+
+import (
+	api_v1 "k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/composite"
+)
+
+const (
+	L3Protocol = "UNSPECIFIED"
+)
+
+// GetProtocol returns the protocol for the BackendService based on Kubernetes Service port definitions
+// and existing backend service.
+//
+// If the service exists and service protocols match existingBS return that protocol.
+// Otherwise prefer L3.
+//
+// See https://cloud.google.com/load-balancing/docs/internal#forwarding-rule-protocols
+func GetProtocol(svcPorts []api_v1.ServicePort, existingBS *composite.BackendService) string {
+	if doesntExist := existingBS == nil; doesntExist {
+		return L3Protocol
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/backends/protocol.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// GetProtocol returns the protocol for the BackendService based on Kubernetes Service port definitions
+// and existing backend service.
+//
+// If the service exists and service protocols match existingBS return that protocol.
+// Otherwise prefer L3.
+//
+// See https://cloud.google.com/load-balancing/docs/internal#forwarding-rule-protocols
+func GetProtocol(svcPorts []api_v1.ServicePort, existingBS *composite.BackendService) string {
```
mmamczur, line 19: let's try to not use L3 by default
if we have non-mixed proto service then stick to the old behavior
if there is a mixed proto service then use L3
if service was single protocol but becomes multi protocol lets recreate the backend service

We can consider doing L3 Backend Services by default but only if we get 100% confirmation from the LB teams that there would be no limitations

TortillaZHawaii: Yeah, I agree.

I will mark it as in draft until we have confirmation.

TortillaZHawaii:
/hold

Waits for confirmation on feature differences between L3 and TCP/UDP backend services from L4 teams.

TortillaZHawaii:
Alternative solution that does not change how we handle single protocol backend services: https://github.com/kubernetes/ingress-gce/pull/2739

TortillaZHawaii:
Closed as we will pursue alternative solution https://github.com/kubernetes/ingress-gce/pull/2739 that doesn't default to L3 Backend Service

PR #2734, opened 2024-11-13
ILB mixed protocol forwarding rules

TortillaZHawaii:
When the mixed protocol flag is enabled this will use `forwardingrules.EnsureILBIPv4`. 

In draft because this PR will likely be split into multiple smaller PRs and also waits for the `backends.GetProtocol`.


TortillaZHawaii:
Moved with different solution

PR #2735, opened 2024-11-13
Add "SHARED_LOADBALANCER_VIP" to internal addresses

TortillaZHawaii:
When creating multiple internal forwarding rules for one address we need to use the address with .Purpose = "SHARED_LOADBALANCER_VIP". Otherwise we will receive an `Error syncing load balancer: googleapi: Error 409: IP_IN_USE_BY_ANOTHER_RESOURCE - IP '10.164.0.18' is already being used by another resource` .

### DIFF
--- a/pkg/address/manager.go
+++ b/pkg/address/manager.go
@@ -51,6 +51,9 @@ const (
 	// IPv6EndpointTypeNetLB is a value on Address.Ipv6EndpointType used specify
 	// that IPv6 address will be used for NetLB. Required for new IPv6 NetLB address creation.
 	IPv6EndpointTypeNetLB = "NETLB"
+	// Purpose for internal IP address that allows to use the same address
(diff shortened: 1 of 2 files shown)


TortillaZHawaii:
It looks like it is an issue for existing IP addresses. If an address was reserved before this change trying to reserve it again with .Purpose = "SHARED_LOADBALANCER_VIP" will lead to an error:
```
[manager.go:226] "Address reserve error" logger="Other controllers.L4Controller.L4Handler.AddressManager" serviceKey="default/hello-service" syncId=1608326150 err="googleapi: Error 409: IP_IN_USE_BY_ANOTHER_RESOURCE - IP '10.164.0.13' is already being used by another resource."
```

Not sure how to resolve it

TortillaZHawaii:
Closing, as we went with single forwarding rule for ILBs.

PR #2736, opened 2024-11-13
Quick fix RXLB transition test by increasing sleep and moving sleep before getting ingress

panslava:
/assign @swetharepakula


swetharepakula:
Thanks Slavik!

/lgtm
/approve

PR #2738, opened 2024-11-14
Add protocol aware GetPorts function

TortillaZHawaii:
Mixed protocol feature cannot rely on existing utils.GetPorts as it returns all ports in the spec, ignoring protocol. 

Part of the https://github.com/kubernetes/ingress-gce/pull/2734

### DIFF
--- a/pkg/forwardingrules/ports.go
+++ b/pkg/forwardingrules/ports.go
@@ -1,0 +1,31 @@
+package forwardingrules
+
+import (
+	"fmt"
+
+	api_v1 "k8s.io/api/core/v1"
+)
+
+const (
+	// Protocol field is optional.
+	// When it is empty the default is TCP
+	defaultProtocol = api_v1.ProtocolTCP
+)
+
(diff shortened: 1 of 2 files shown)


pkg/forwardingrules/ports.go
```diff
@@ -0,0 +1,37 @@
...
+		if proto == "" {
+			proto = defaultProtocol
+		}
+
+		if proto == protocol {
+			ports = append(ports, fmt.Sprint(p.Port))
+		}
+	}
+
+	if len(ports) > maxForwardedPorts {
```
mmamczur, line 33: so this is specifically for ILB only? I'd let the calling code decide if it wants to use max 5 or more

TortillaZHawaii: okay, I will return ports as they are

TortillaZHawaii: Resolved in latest push

PR #2740, opened 2024-11-18
Firewall Allowed for NetLB mixed protocol

TortillaZHawaii:
When specifying multi protocol service NetLB Controller needs to update firewall rules to allow traffic to flow on both TCP and UDP ports. This PR introduces `firewalls.AllowedForService` which adds that functionality, and invokes the function if the EnableL4MixedProtocol flag is enabled.

### DIFF
--- a/pkg/firewalls/allowed_test.go
+++ b/pkg/firewalls/allowed_test.go
@@ -1,0 +1,105 @@
+package firewalls
+
+import (
+	"testing"
+
+	compute "google.golang.org/api/compute/v1"
+	api_v1 "k8s.io/api/core/v1"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/firewalls/allowed_test.go
```diff
@@ -0,0 +1,106 @@
+package firewalls_test
```
mmamczur, line 1: this could still be the firewalls package. You won't have to export equal function  then

TortillaZHawaii: this is true, I will fix that

TortillaZHawaii: Fixed in pushed commit

PR #2743, opened 2024-11-20
Add L3 support for Forwarding rule namer

TortillaZHawaii:
Current implementation isn't able to simply add l3_default to the resource name as underscores are not allowed in the name. We can shorten the protocol to l3 to prevent that. It also doesn't take too much of the string that has to also fit service name and namespace.


pkg/utils/namer/l4_namer.go
```diff
@@ -107,6 +108,11 @@ func (namer *L4Namer) L4IPv6Firewall(namespace, name string) string {
 //
 // Output name is at most 63 characters.
 func (namer *L4Namer) L4ForwardingRule(namespace, name, protocol string) string {
+	protocol = strings.ToLower(protocol)
+	if protocol == "l3_default" {
```
mmamczur, line 112: when will this value be used as argument? Let's add a constant for it as well.

PR #2748, opened 2024-11-26
Add ability to push images with multiple tags

swetharepakula:
* Add one more tag for e2e-test images

/assign @gauravkghildiyal

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -31,6 +31,7 @@ steps:
   - HOME=/root
   - USER=root
   - BINARIES=e2e-test
+  - ADDITIONAL_TAGS="infrastructure-public-image-${_PULL_BASE_REF}"
   args:
   - -c
   - |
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -26,7 +26,7 @@ steps:
   env:
   - DOCKER_CLI_EXPERIMENTAL=enabled
   - REGISTRY=gcr.io/k8s-ingress-image-push
-  - VERSION=$_PULL_BASE_REF
+  - VERSION="${_PULL_BASE_REF} infrastructure-public-image-${_PULL_BASE_REF}"
```
gauravkghildiyal, line 29: nit: VERSIONS ?

swetharepakula: I am a little worried about changing the name of the variable incase there are other scripts built on top of push-multiarch.sh that we are unaware about.

By reusing `VERSION` and handling both single version or a list felt like a safer option. What do you think?

gauravkghildiyal: Yeah I can relate to this. Sure let's keep things as they are.

swetharepakula:
/hold remove 

Script has been tested

gauravkghildiyal:
/lgtm

Thanks!

PR #2750, opened 2024-11-26
Garbage collect NEGs in to-be-deleted state.

sawsa307:
/assign @swetharepakula 
/assign @gauravkghildiyal

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -548,16 +548,21 @@ func (manager *syncerManager) garbageCollectNEG() error {
 	return nil
 }
 
+type negDeletionCandidate struct {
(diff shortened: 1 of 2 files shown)


gauravkghildiyal (commented):
An initial look here makes me wonder: "Are we forcing this new pattern of deleting NEGs into our existing flow?"

More precisely, the code right handles the deletion of the SvcNEG and ALL of it's deleted NEGs. The part here where we also have to handle the deletion of the SvcNEG itself also adds other complications.

What we want to add now is the ability to just deleted targeted NEGs which have a specific state right? Have we thought about doing this in it's own piece of code? Would that be more easier or do we think that has it's own problems?

PR #2753, opened 2024-12-02
Add Ensure tests for mixed protocol ILB

TortillaZHawaii:
- Adds IPv6 and Dual Stack support for mixed protocol ILBs
- Adds test data in `pkg/loadbalancers/mixedprotocoltest` for {TCP, UDP, Mixed} x {IPv4, IPv6, DualStack}, including:
    -  annotations
    -  consts for names of nodes, services, namespaces, ip addresses
    - Kubernetes Ingress definitions
    - Cloud resources (fakeGCE): forwarding rules, backend services, firewalls, health checks
- Adds TestEnsureMixedILB:
    - Tests for {IPv4, IPv6, DualStack} x {TCP, UDP, Mixed}
    - create
    - transitions
    - delete
- Fixes forwarding rule resource leak when switching both protocol and stack f.e. IPv4 UDP to IPv6 TCP, by deleting all potential forwarding rules for a service. This leak will be created because the delete logic tries to create forwarding rule name for the old state stack based on the new protocol. So instead of deleting IPv4 UDP it tries to delete IPv4 TCP, which doesn't exist.

### DIFF
--- a/pkg/loadbalancers/mixedprotocoltest/resources.go
+++ b/pkg/loadbalancers/mixedprotocoltest/resources.go
@@ -1,0 +1,317 @@
+package mixedprotocoltest
+
+import (
+	"google.golang.org/api/compute/v1"
+)
+
+// GCEResources collects all GCE resources that are managed by the mixed protocol ILB
+type GCEResources struct {
+	FwdRuleIPv4    *compute.ForwardingRule
+	FirewallIPv4   *compute.Firewall
+	FwdRuleIPv6    *compute.ForwardingRule
+	FirewallIPv6   *compute.Firewall
+	HC             *compute.HealthCheck
+	HCFirewallIPv4 *compute.Firewall
+	HCFirewallIPv6 *compute.Firewall
+	BS             *compute.BackendService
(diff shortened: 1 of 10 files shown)


pkg/loadbalancers/mixedprotocoltest/resources.go
```diff
@@ -0,0 +1,317 @@
...
+		}),
+		BS:             BS("UNSPECIFIED"),
+		HC:             HC(),
+		HCFirewallIPv4: HCFirewall(),
+		HCFirewallIPv6: HCFirewallIPv6(),
+	}
+}
+
+// HC returns shared HealthCheck
+func HC() *compute.HealthCheck {
```
mmamczur, line 156: I'd suggest just naming this HealthCheck 
same for BS and others, also in the GCEL4Resources

GCEL4Resources could be GCEResources since this is under L4 package anyway

TortillaZHawaii: You're right, I will fix it

Issue #2724, opened 2024-10-31
Adding InstanceGroupController does not remove nodes from instance group after adding `node.kubernetes.io/exclude-from-external-load-balancers` label


VitusAcabado:
I am running `ingress-gce` on a self-managed kubernetes cluster on GCE.

Adding the label `node.kubernetes.io/exclude-from-external-load-balancers=true` to a node that is already in an instance group does not get removed because of this [line](https://github.com/kubernetes/ingress-gce/blob/master/pkg/instancegroups/manager.go#L312)

Since the list of nodes passed in the `sync` function already has the excluded nodes via `CandidateNodeFilter` [here](https://github.com/kubernetes/ingress-gce/blob/master/pkg/instancegroups/controller.go#L129).

It never proceeds into the for loop and these [lines](https://github.com/kubernetes/ingress-gce/blob/master/pkg/instancegroups/manager.go#L338-L361) wont get executed.

This is observed for tags v1.30.3 onwards.

swetharepakula:
cc/ @mmamczur

mmamczur:
/remove-lifecycle stale

this was fixed in [#2754](https://github.com/kubernetes/ingress-gce/pull/2754)

PR #2754, opened 2024-12-03
Fix excluding of nodes from NetLBs for last node in zone

anyush:
If no nodes for certain zone were provided to Sync, then all existing ones are removed.

Issue: https://github.com/kubernetes/ingress-gce/issues/2724

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -312,6 +312,16 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
 	zonedNodes := m.splitNodesByZone(nodes, iglogger)
 	iglogger.Info(fmt.Sprintf("Syncing nodes: %d nodes over %d zones", len(nodes), len(zonedNodes)))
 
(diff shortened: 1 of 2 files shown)


pkg/instancegroups/manager.go
```diff
@@ -312,6 +312,16 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
 	zonedNodes := m.splitNodesByZone(nodes, iglogger)
 	iglogger.Info(fmt.Sprintf("Syncing nodes: %d nodes over %d zones", len(nodes), len(zonedNodes)))
 
+	zones, err := m.ZoneGetter.ListZones(zonegetter.AllNodesFilter, iglogger)
+	if err != nil {
+		iglogger.Error(err, "Failed to list zones")
+	}
+	for _, zone := range zones {
```
mmamczur, line 319: I'm not sure if this would work. zones and nodes actually come from the same place

check out what `ZoneGetter.ListZones()` does. It lists nodes and extracts zones from them

so if there is no node in zone X then it won't even be in the result from ListZones

PR #2757, opened 2024-12-04
Trigger node sync on PodCIDR change in the IG controller.

mmamczur:
Nodes with missing PodCIDR can be skipped during IG sync. When the PodCIDR is added nodes should be reprocessed.

### DIFF
--- a/pkg/instancegroups/controller.go
+++ b/pkg/instancegroups/controller.go
@@ -111,6 +111,9 @@ func nodeStatusChanged(old, cur *apiv1.Node) bool {
 	if utils.NodeIsReady(old) != utils.NodeIsReady(cur) {
 		return true
 	}
+	if old.Spec.PodCIDR != cur.Spec.PodCIDR {
+		return true
+	}
 	return false
 }
 
--- a/pkg/instancegroups/controller_test.go
+++ b/pkg/instancegroups/controller_test.go
@@ -18,44 +18,68 @@ import (
 
 func TestNodeStatusChanged(t *testing.T) {
 	testCases := []struct {
-		desc   string
-		mutate func(node *api_v1.Node)
-		expect bool
+		desc        string
+		initialNode *api_v1.Node
+		mutate      func(node *api_v1.Node)
+		expect      bool
 	}{
 		{
-			"no change",
-			func(node *api_v1.Node) {},
-			false,
+			desc:   "no change",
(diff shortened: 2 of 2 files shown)


pkg/instancegroups/controller.go
```diff
@@ -111,6 +111,9 @@ func nodeStatusChanged(old, cur *apiv1.Node) bool {
 	if utils.NodeIsReady(old) != utils.NodeIsReady(cur) {
 		return true
 	}
+	if old.Spec.PodCIDR != cur.Spec.PodCIDR {
```
kl52752, line 114: there is also `PodCIDRs` field in pod spec, don't you need to compare this one too?

pkg/instancegroups/controller_test.go
```diff
@@ -49,6 +49,13 @@ func TestNodeStatusChanged(t *testing.T) {
 			},
 			false,
 		},
+		{
+			"PodCIDR_added",
+			func(node *api_v1.Node) {
```
kl52752, line 54: This test checks only case where PodCIDR was updated. What about updated? Is this possible that CIDR will be updated? If yes then please add more test cases if no, maybe you should change the logic and print the warning when podCIDR changed.

pkg/instancegroups/controller.go
mmamczur: no, the code later only depends on the man PodCIDR

pkg/instancegroups/controller_test.go
mmamczur: I added a test case but it involved modifying that test a bit more

pkg/instancegroups/controller.go
mmamczur: basically it's because of this part https://github.com/kubernetes/ingress-gce/blob/6a02d24d7f7210c77c6e0792ed6f85829f12f489/pkg/utils/zonegetter/zone_getter.go#L335

we have this 'safety fallback' when determining the subnet of the node (this is in context of MultiSubnet Clusters).
anyway, when podCIDR is missing and the subnet label is missing we skip the node. It should get the label with PodCIDR, or it may not get a label but get PodCIDR, in which case we say it's default subnet.

the problem is, initially nodes don't have PodCIDR. So we skip them but later when they get that PodCIDR we don't sync anymore, we don't add node as backend and we have a problem.

PR #2761, opened 2024-12-06
Add funcs to add/delete finalizer on ClusterSlice 

panslava:
- Added general functions to patch ClusterSlice ObjectMetadata
- Added functions to add/delete neg cleanup finalizer on ClusterSlice.
  This will allow to properly clean neg controllers when cluster slice
  is deleted

/assign @gauravkghildiyal

### DIFF
--- a/pkg/multiproject/finalizer/finalizer.go
+++ b/pkg/multiproject/finalizer/finalizer.go
@@ -1,0 +1,54 @@
+package finalizer
+
+import (
+	"fmt"
+	"slices"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	providerconfig "k8s.io/ingress-gce/pkg/apis/providerconfig/v1"
+	providerconfigclient "k8s.io/ingress-gce/pkg/providerconfig/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils/patch"
+	"k8s.io/ingress-gce/pkg/utils/slice"
+	"k8s.io/klog/v2"
+)
+
+const (
+	ProviderConfigNEGCleanupFinalizer = "multiproject.networking.gke.io/neg-cleanup"
+)
+
+func EnsureProviderConfigNEGCleanupFinalizer(cs *providerconfig.ProviderConfig, csClient providerconfigclient.Interface, logger klog.Logger) error {
+	return ensureProviderConfigFinalizer(cs, ProviderConfigNEGCleanupFinalizer, csClient, logger)
+}
+
+func DeleteProviderConfigNEGCleanupFinalizer(cs *providerconfig.ProviderConfig, csClient providerconfigclient.Interface, logger klog.Logger) error {
+	return deleteProviderConfigFinalizer(cs, ProviderConfigNEGCleanupFinalizer, csClient, logger)
+}
+
+func ensureProviderConfigFinalizer(pc *providerconfig.ProviderConfig, key string, csClient providerconfigclient.Interface, logger klog.Logger) error {
+	if HasGivenFinalizer(pc.ObjectMeta, key) {
+		return nil
+	}
+
+	// Make a deep copy of the ProviderConfig to avoid mutating the shared informer cache.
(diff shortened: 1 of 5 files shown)


pkg/utils/patch/patch.go
```diff
@@ -84,3 +88,40 @@ func PatchServiceLoadBalancerStatus(client coreclient.CoreV1Interface, svc *core
 	_, err := svchelpers.PatchService(client, svc, newSvc)
 	return err
 }
+
+// PatchClusterSliceObjectMetadata patches the given ClusterSlice's metadata based on new metadata.
+func PatchClusterSliceObjectMetadata(client clustersliceclient.Interface, cs *clusterslice.ClusterSlice, newObjectMetadata metav1.ObjectMeta) error {
+	// Reset Spec to ensure only ObjectMeta is patched.
+	newCS := cs.DeepCopy()
+	newCS.Spec = cs.Spec
```
gauravkghildiyal, line 96: If we deepcopied cs, why are we copying the spec again?

pkg/multiproject/finalizer/finalizer.go
```diff
@@ -0,0 +1,53 @@
...
+
+	updatedObjectMeta := cs.ObjectMeta.DeepCopy()
+	updatedObjectMeta.Finalizers = slice.RemoveString(updatedObjectMeta.Finalizers, key, nil)
+	logger.V(2).Info("Deleting finalizer from ClusterSlice", "finalizerKey", key, "clusterSlice", fmt.Sprintf("%s/%s", cs.Namespace, cs.Name))
+	return patch.PatchClusterSliceObjectMetadata(csClient, cs, *updatedObjectMeta)
+}
+
+// HasGivenFinalizer is true if the passed in meta has the specified finalizer.
+func HasGivenFinalizer(m metav1.ObjectMeta, key string) bool {
+	return slice.ContainsString(m.Finalizers, key, nil)
```
gauravkghildiyal, line 52: nit: `slices.Contains(...)` can avoid that extra `nil` argument.

pkg/multiproject/finalizer/finalizer_test.go
```diff
@@ -0,0 +1,141 @@
...
+			// Check if the finalizer is present as expected
+			hasFinalizer := slice.ContainsString(updatedCS.Finalizers, ClusterSliceNEGCleanupFinalizer, nil)
+			if hasFinalizer != tc.expectedFinalizerPresent {
+				t.Errorf("Finalizer presence mismatch: expected %v, got %v", tc.expectedFinalizerPresent, hasFinalizer)
+			}
+		})
+	}
+}
+
+func TestEnsureDeleteClusterSliceNEGCleanupFinalizer(t *testing.T) {
```
gauravkghildiyal, line 78: nit: Any reason why we have both `EnsureDelete` (like do we want to test both, or a typo?)

gauravkghildiyal (commented):
Thanks @panslava!

This LGTM (pending remaining)

pkg/utils/patch/patch.go
panslava: oh yeh, good catch, no reason for that...

pkg/multiproject/finalizer/finalizer.go
panslava: replaced

pkg/multiproject/finalizer/finalizer_test.go
panslava: Leftover, before I just called the actual function "EnsureDelete" (as done in old code), but then decided there is no reason to add "Ensure"

pkg/multiproject/finalizer/finalizer.go
panslava: also added one more commit to replace in old code

panslava:
@gauravkghildiyal ready for rereview

pkg/multiproject/finalizer/finalizer.go
gauravkghildiyal: Thanks for the cleanup!

PR #2762, opened 2024-12-06
Add multiproject manager which will be responsible for starting and tracking controllers per ProviderConfig

panslava:
/assign @gauravkghildiyal 

based on 3 previous prs

### DIFF
--- a/pkg/multiproject/manager/manager.go
+++ b/pkg/multiproject/manager/manager.go
@@ -1,0 +1,148 @@
+package manager
+
+import (
+	"fmt"
+	"sync"
+
+	networkclient "github.com/GoogleCloudPlatform/gke-networking-api/client/network/clientset/versioned"
+	nodetopologyclient "github.com/GoogleCloudPlatform/gke-networking-api/client/nodetopology/clientset/versioned"
+	"k8s.io/apimachinery/pkg/types"
+	informers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	providerconfig "k8s.io/ingress-gce/pkg/apis/providerconfig/v1"
+	"k8s.io/ingress-gce/pkg/multiproject/finalizer"
+	"k8s.io/ingress-gce/pkg/multiproject/neg"
+	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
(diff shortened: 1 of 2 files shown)


pkg/multiproject/manager/manager.go
```diff
@@ -0,0 +1,121 @@
...
+	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
+	providerconfigclient "k8s.io/ingress-gce/pkg/providerconfig/client/clientset/versioned"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/klog/v2"
+)
+
+type ProviderConfigControllersManager struct {
+	controllers          map[string]*ControllerSet
+	mu                   sync.Mutex
```
gauravkghildiyal, line 22: nit: Let's use a [mutex hat](https://dmitri.shuralyov.com/idiomatic-go#mutex-hat) for the variables we want this mutex to control

panslava: thanks, done

pkg/multiproject/manager/manager.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil
+	}
+
+	err := finalizer.EnsureProviderConfigNEGCleanupFinalizer(cs, cscm.providerConfigClient, cscm.logger)
+	if err != nil {
+		return fmt.Errorf("failed to ensure NEG cleanup finalizer for project %s: %v", csKey, err)
+	}
+	negControllerStopCh, err := neg.StartNEGController(cscm.informersFactory, cscm.kubeClient, cscm.svcNegClient, cscm.eventRecorderClient, cscm.kubeSystemUID, cscm.clusterNamer, cscm.l4Namer, cscm.lpConfig, cscm.defaultCloudConfig, cscm.globalStopCh, cscm.logger, cs)
+	if err != nil {
+		return fmt.Errorf("failed to start NEG controller for project %s: %v", csKey, err)
```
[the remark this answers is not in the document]
panslava: I think it's better to attempt cleaning finalizer inside this function in case of an error

Added this, thanks for the suggestion

panslava:
ready for rereview @gauravkghildiyal

PR #2763, opened 2024-12-10
Add flag for IPV6 only NEG's

zasweq:
This PR adds a flag for IPV6 only NEG's. If the flag is false, then the NEG Controller will still error if an IPV6 only service is configured.

The plan is to flag flip in a patch release once IPV6 NEG support is available in production (requires GCP LB support for IPV6 only NEG's).

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -143,6 +143,7 @@ var F = struct {
 	ProviderConfigAPIGroup                   string
 	EnableL4ILBMixedProtocol                 bool
 	EnableL4NetLBMixedProtocol               bool
+	EnableIPV6OnlyNEG                        bool
 }{
 	GCERateLimitScale: 1.0,
 }
@@ -334,6 +335,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableL4NetLBMixedProtocol, "enable-l4netlb-mixed-protocol", false, "Enable support for mixed protocol L4 external load balancers.")
 	flag.StringVar(&F.ProviderConfigAPIGroup, "provider-config-api-group", "", "The API group for the ProviderConfig CRD.")
 	flag.StringVar(&F.ProviderConfigNameLabelKey, "provider-config-name-label-key", "", "The label key for provider-config name, which is used to identify the provider-config of objects in multi-project mode.")
+	flag.BoolVar(&F.EnableIPV6OnlyNEG, "enable-ipv6-only-neg", false, "Enable support for IPV6 Only NEG's.")
 }
 
 func Validate() {
--- a/pkg/neg/controller_test.go
+++ b/pkg/neg/controller_test.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/cloud-provider-gcp/providers/gce"
 	"k8s.io/ingress-gce/pkg/annotations"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/neg/metrics/metricscollector"
 	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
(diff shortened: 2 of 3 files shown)


pkg/flags/flags.go
```diff
@@ -334,6 +335,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableL4ILBMixedProtocol, "enable-l4ilb-mixed-protocol", false, "Enable support for mixed protocol L4 internal load balancers.")
 	flag.BoolVar(&F.EnableL4NetLBMixedProtocol, "enable-l4netlb-mixed-protocol", false, "Enable support for mixed protocol L4 external load balancers.")
 	flag.StringVar(&F.ClusterSliceAPIGroup, "cluster-slice-api-group", "", "The API group for the ClusterSlice CRD.")
+	flag.BoolVar(&F.EnableIPV6NEG, "enable-ipv6-neg", false, "Enable support for IPV6 NEG's.")
```
gauravkghildiyal, line 338: nit: Can we use `enable-ipv6-only-neg` for clarity? (If we do, let's propagate the same to the variable names as well)

gauravkghildiyal (commented):
Thanks @zasweq

pkg/flags/flags.go
zasweq: Ah, good point. The error message returned from config verification says only and so do the tests. Switched.

pkg/neg/controller_test.go
```diff
@@ -1513,6 +1515,14 @@ func TestServiceIPFamilies(t *testing.T) {
 			ipFamilyPolicy: &singleStack,
 			expectNil:      false,
 		},
+		{
+			desc:              "ipv6 only service with l7 load balancer, ipFamilyPolicy is singleStack, ipv6 neg enabled",
+			serviceType:       v1.ServiceTypeClusterIP,
+			ipFamilies:        []v1.IPFamily{v1.IPv6Protocol},
```
krzykwas, line 1521: I'd just like to make sure the tests cover all the cases. Sorry if my comments are off, because I may not have all the context. A customer in a single stack IPv6 cluster could create a normal Service, where the Spec lacks `ipFamilies` or `ipFamilyPolicy`. Because the cluster is single stack IPv6 and IPv6 is the only family available, the Service will only have IPv6 addresses, and so the backends, and the NEGs would be only configured with an IPv6 address. So theoretically there is a case where these fields are not set, but a NEG in the end would only have an IPv6 address. I don't know if this matters, maybe it's not possible to add coverage just with unit tests. WDYT?

zasweq: I see what you're saying. For the scope of this specific unit test, which tests at the validation boundary, I think what to test here would be for an unset IP family to pass validation.

However, I wonder how to test this e2e. It would require however we determine the GKE Cluster's stack type and see if IPV6 + unset, and see if unset spins up a IPv6.

zasweq: Hmm ok I'll add a test for unset ipFamilies and ipFamily policy at the unit test layer. I'm assuming that for IPV4 only it'll simply create an IPV4 NEG, just as in the case of IPV6 only cluster it's valid and we want it to create an IPV6 only NEG.

gauravkghildiyal: Thanks @krzykwas, I wasn't aware about the _"theoretically there is a case where these fields are not set, but a NEG in the end would only have an IPv6 address"_ case. Good to have become aware about this.

But yes, this unit test here will not be able to exercise this check and we'll either need some kind of an integration test, or write a standard e2e test.

pkg/neg/controller_test.go
```diff
@@ -1590,9 +1590,9 @@ func TestServiceIPFamilies(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
 			if tc.enableIPV6OnlyNEG {
-				flags.F.EnableIPV6NEG = true
+				flags.F.EnableIPV6OnlyNEG = true
```
swetharepakula, line 1593: since this is a global variable, can you store the previous value and reset it in the defer. As the tests grow using this flag, I think that will be safer in terms of test pollution. Thanks

swetharepakula (commented):
Thanks for making the changes!

One nit in how the test was written.

Otherwise can you squash the commits as well? Or please expand the second commit message with more context.

pkg/neg/controller_test.go
zasweq: Ah, good point. That is how we set env vars in gRPC-Go tests too. Thanks.

zasweq:
Rebased and squashed all commits including those that address comments into one commit only.

pkg/neg/controller_test.go
zasweq: Ok sounds good. Let's leave this test to an integration or e2e test then.

swetharepakula: Service spec cannot have any empty `ipFamilies` and empty `ipFamilyPolicy`. If none are mentioned, the K8s API will put the default value based on the what is configured as the default stack: https://kubernetes.io/docs/concepts/services-networking/dual-stack/

swetharepakula (commented):
/lgtm
/approve

pkg/neg/controller_test.go
zasweq: Oh ok sounds good. Looks like there was no unit test here for this anyway, but important to keep in mind as the team thinks about writing integration and e2e tests. Thanks.

PR #2789, opened 2025-01-11
Refactor context health check implementation

panslava:
- Move Health checks out of context to systemhealth package. Context is already doing too much and systemhealth is self-sufficient struct that can live in separate package, which makes code less entangled
- Modify existing calls for AddHealthCheck to use new struct
- This will also allow to run healthchecks when controller is not initialising full context (for example in multiproject mode)

also I could've added this into L4 and L4netlb controller and add health check inside their constructor. Buuut. It would need to change too much of the test code to add everywhere new argument....

/assign @gauravkghildiyal

### DIFF
--- a/cmd/glbc/app/handlers.go
+++ b/cmd/glbc/app/handlers.go
@@ -29,14 +29,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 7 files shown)


gauravkghildiyal:
/cc @mmamczur 

There's some very minimal changes to L4 controllers here. Would appreciate your approval.

PR #2790, opened 2025-01-11
Refactor leader elect main

panslava:
based on https://github.com/kubernetes/ingress-gce/pull/2789

Should be no-op, but it was just a bit hard to read and understand the logic before

/assign @gauravkghildiyal

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -218,6 +218,19 @@ func main() {
 	var once sync.Once
 	stopCh := make(chan struct{})
 
+	rOption := runOption{
+		wg:     &sync.WaitGroup{},
+		stopCh: stopCh,
+		// This ensures that stopCh is only closed once.
+		// Right now, we have three callers.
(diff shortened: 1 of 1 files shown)


cmd/glbc/main.go
```diff
@@ -217,6 +218,22 @@ func main() {
 	var once sync.Once
 	stopCh := make(chan struct{})
 
+	rOption := runOption{
+		wg:     &sync.WaitGroup{},
+		stopCh: stopCh,
+		// This ensures that stopCh is only closed once.
+		// Right now, we have three callers.
+		// One is triggered when the ASM configmap changes, and the other two are
+		// triggered by the SIGTERM handler.
```
gauravkghildiyal, line 227: nit: Does the "two" here mean the two sigterm handlers that we converted to one? If yes, let's update the comment

gauravkghildiyal (commented):
Looking good. Waiting for the parent.

panslava:
@gauravkghildiyal rebase on new master and changed comment

PR #2764, opened 2024-12-11
NetLB IPv4 Mixed Protocol 

TortillaZHawaii:

PR #2793, opened 2025-01-13
Update address.HoldExternal to allow holding IPv6 as well

TortillaZHawaii:
This is done in order to reuse address holding logic in future mixed protocol NetLB IPv6 implementation. Similar thing has been already done in IPv4.

* Updates config with fields that are IPv4/IPv6 specific.
* Adds tests for HoldExternal IPv6 addresses.
* Updates forwarding_rules_ipv6.go to use address.HoldExternal.
* Removes tearDownResourcesWithWrongNetworkTier from l4netlb as it is already implemented inside address.HoldExternal.

Waits for #2764 to be merged.


TortillaZHawaii:
Duplicate of #2734

PR #2801, opened 2025-01-31
Remove ProviderConfigAPIGroup flag. Regenerate clients with providerconfig.gke.io API Group

panslava:
/assign @gauravkghildiyal


panslava:
@gauravkghildiyal fixed tests and lost lgtm

PR #2805, opened 2025-02-14
Introduce a new lock for L4 and related controllers.

mmamczur:
This is the first part of Phase 1 of the effort to split L4 from ingress-gce.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -71,6 +71,7 @@ import (
 )
 
 const negLockName = "ingress-gce-neg-lock"
+const l4LockName = "l4-lb-controller-gce-lock"
 
 func main() {
 	flags.Register()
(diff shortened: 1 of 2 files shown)


swetharepakula (commented):
LGTM, waiting on @gauravkghildiyal to give final LGTM.

cmd/glbc/main.go
```diff
@@ -71,6 +71,7 @@ import (
 )
 
 const negLockName = "ingress-gce-neg-lock"
+const l4LockName = "l4-lb-controller-gce-lock"
```
gauravkghildiyal, line 74: nitpick: Do we need "-gce-"?

gauravkghildiyal (commented):
I did have a nitpick or a question, but don't want that to block this.

/lgtm

PR #2806, opened 2025-02-18
ProviderConfig fixes and integration test

panslava:
/assign @gauravkghildiyal

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -32,6 +32,7 @@ import (
 	flag "github.com/spf13/pflag"
 	crdclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	informers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	clientset "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -43,8 +44,10 @@ import (
 	frontendconfigclient "k8s.io/ingress-gce/pkg/frontendconfig/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/instancegroups"
 	"k8s.io/ingress-gce/pkg/l4lb"
+	multiprojectgce "k8s.io/ingress-gce/pkg/multiproject/gce"
 	multiprojectstart "k8s.io/ingress-gce/pkg/multiproject/start"
 	"k8s.io/ingress-gce/pkg/network"
+	providerconfigclient "k8s.io/ingress-gce/pkg/providerconfig/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/psc"
 	"k8s.io/ingress-gce/pkg/serviceattachment"
 	serviceattachmentclient "k8s.io/ingress-gce/pkg/serviceattachment/client/clientset/versioned"
@@ -243,17 +246,26 @@ func main() {
 		rootLogger.Info("Multi-project mode is enabled, starting project-syncer")
 
 		runWithWg(func() {
+			gceCreator, err := multiprojectgce.NewGCEFromFileStringCreator(rootLogger)
+			if err != nil {
+				klog.Fatalf("Failed to create GCE creator: %v", err)
+			}
+			providerConfigClient, err := providerconfigclient.NewForConfig(kubeConfig)
+			if err != nil {
+				klog.Fatalf("Failed to create ProviderConfig client: %v", err)
+			}
 			if flags.F.LeaderElection.LeaderElect {
 				err := multiprojectstart.StartWithLeaderElection(
 					context.Background(),
 					leaderElectKubeClient,
 					hostname,
-					kubeConfig,
 					rootLogger,
 					kubeClient,
 					svcNegClient,
 					kubeSystemUID,
 					eventRecorderKubeClient,
+					providerConfigClient,
+					gceCreator,
 					namer,
 					stopCh,
 				)
@@ -261,13 +273,16 @@ func main() {
 					rootLogger.Error(err, "Failed to start multi-project syncer with leader election")
 				}
 			} else {
+				informersFactory := informers.NewSharedInformerFactory(kubeClient, flags.F.ResyncPeriod)
 				multiprojectstart.Start(
-					kubeConfig,
 					rootLogger,
 					kubeClient,
 					svcNegClient,
 					kubeSystemUID,
 					eventRecorderKubeClient,
+					providerConfigClient,
+					informersFactory,
+					gceCreator,
 					namer,
 					stopCh,
 				)
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -246,7 +246,7 @@ func main() {
 		rootLogger.Info("Multi-project mode is enabled, starting project-syncer")
 
 		runWithWg(func() {
-			gceCreator, err := multiprojectgce.NewGCEFromFileStringCreator(rootLogger)
+			gceCreator, err := multiprojectgce.NewDefaultGCECreator(rootLogger)
 			if err != nil {
 				klog.Fatalf("Failed to create GCE creator: %v", err)
 			}
@@ -254,6 +254,7 @@ func main() {
 			if err != nil {
 				klog.Fatalf("Failed to create ProviderConfig client: %v", err)
 			}
+			informersFactory := informers.NewSharedInformerFactory(kubeClient, flags.F.ResyncPeriod)
 			if flags.F.LeaderElection.LeaderElect {
 				err := multiprojectstart.StartWithLeaderElection(
 					context.Background(),
@@ -265,6 +266,7 @@ func main() {
 					kubeSystemUID,
 					eventRecorderKubeClient,
 					providerConfigClient,
+					informersFactory,
 					gceCreator,
 					namer,
 					stopCh,
@@ -273,7 +275,6 @@ func main() {
 					rootLogger.Error(err, "Failed to start multi-project syncer with leader election")
 				}
 			} else {
-				informersFactory := informers.NewSharedInformerFactory(kubeClient, flags.F.ResyncPeriod)
 				multiprojectstart.Start(
 					rootLogger,
 					kubeClient,
--- a/pkg/multiproject/controller/controller.go
+++ b/pkg/multiproject/controller/controller.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"runtime/debug"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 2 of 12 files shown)


pkg/neg/controller.go
```diff
@@ -178,6 +178,12 @@ func NewController(
 		apiv1.EventSource{Component: "neg-controller"})
 
 	syncerMetrics := syncMetrics.NewNegMetricsCollector(flags.F.NegMetricsExportInterval, logger)
+
+	allNodes := nodeInformer.GetIndexer().List()
+	klog.Infof("All Nodes in NewController: %+v", allNodes)
```
gauravkghildiyal, line 183: If we're logging individual nodes, maybe we don't need to log allNodes

pkg/neg/controller.go
```diff
@@ -178,6 +178,12 @@ func NewController(
 		apiv1.EventSource{Component: "neg-controller"})
 
 	syncerMetrics := syncMetrics.NewNegMetricsCollector(flags.F.NegMetricsExportInterval, logger)
+
+	allNodes := nodeInformer.GetIndexer().List()
```
gauravkghildiyal, line 182: Curious to know why we're doing this. Also worth documenting.

I'm guessing we're fine with logging whatevers there before hasSynced, right?

pkg/neg/controller.go
```diff
@@ -394,10 +400,12 @@ func (c *Controller) Run() {
 		go wait.Until(c.nodeTopologyWorker, time.Second, c.stopCh)
 	}
 	go func() {
+		c.logger.Info("Waiting for gcPeriod to run the first GC", "gcPeriod", c.gcPeriod)
 		// Wait for gcPeriod to run the first GC
 		// This is to make sure that all services are fully processed before running GC.
 		time.Sleep(c.gcPeriod)
 		wait.Until(c.gc, c.gcPeriod, c.stopCh)
+		c.logger.Info("First GC completed")
```
gauravkghildiyal, line 408: It doesn't look like this is just for the first GC, since it's after the wait.Until

cmd/glbc/main.go
```diff
@@ -243,31 +246,43 @@ func main() {
...
+					providerConfigClient,
+					gceCreator,
 					namer,
 					stopCh,
 				)
 				if err != nil {
 					rootLogger.Error(err, "Failed to start multi-project syncer with leader election")
 				}
 			} else {
+				informersFactory := informers.NewSharedInformerFactory(kubeClient, flags.F.ResyncPeriod)
```
gauravkghildiyal, line 276: For consistency, let's also pass the informersFactory in `StartWithLeaderElection()`. That avoids the possibility of the informer factory construction mistakenly diverging in the future.

(I can imagine that the tests only needed you to inject it through `Start()`, but overall doing in both places may be useful.)

pkg/multiproject/controller/controller.go
```diff
@@ -98,7 +106,8 @@ func (pcc *ProviderConfigController) syncWrapper(key string) error {
 
 	defer func() {
 		if r := recover(); r != nil {
-			svcLogger.Error(fmt.Errorf("panic in ProviderConfig sync worker goroutine: %v", r), "Recovered from panic")
+			stack := string(debug.Stack())
```
gauravkghildiyal, line 109: Nice!

pkg/multiproject/neg/neg.go
```diff
@@ -229,6 +221,14 @@ func initializeInformers(
 	return informers, hasSynced, nil
 }
 
+func addIndexerIfNotPresent(indexer cache.Indexer, indexName string, indexFunc cache.IndexFunc) error {
```
gauravkghildiyal, line 224: I'm assuming this is intended to enable injection in tests. Please document the rationale for doing it.

pkg/multiproject/gce/gce.go
```diff
@@ -21,11 +21,29 @@ func init() {
 	ini.PrettySection = true
 }
 
-// NewGCEForProviderConfig returns a new GCE client for the given project.
+type GCECreator interface {
+	GCEForProviderConfig(providerConfig *v1.ProviderConfig, logger klog.Logger) (*cloudgce.Cloud, error)
+}
+
+type GCEFromFileStringCreator struct {
```
gauravkghildiyal, line 28: Please document the intent of this struct.

I understand the desire with the naming is to clarify the internals of this creator, but perhaps those are better left as internal details (given the `defaultConfigFileString` field is already private, the "fileString" part is simply not configurable). Maybe simply name this DefaultGCECreator (or other ideas)?

pkg/multiproject/gce/fake.go
```diff
@@ -0,0 +1,110 @@
...
+		defaultTestClusterValues:  cloudgce.DefaultTestClusterValues(),
+		clientsForProviderConfigs: make(map[string]*cloudgce.Cloud),
+	}
+}
+
+func providerConfigKey(providerConfig *v1.ProviderConfig) string {
+	return fmt.Sprintf("%s/%s", providerConfig.Namespace, providerConfig.Name)
+}
+
+func (g *GCEFake) GetGCEForProviderConfig(providerConfig *v1.ProviderConfig) (*cloudgce.Cloud, error) {
```
gauravkghildiyal, line 31: nit: I know this is a Fake, but we have two similar named functions `GCEForProviderConfig()` and `GetGCEForProviderConfig()`. The first one seems to be required, since it implements the interface. The second one seems a bit unclear. Judging by its implementation, thus having a function like `HasGCEForProviderConfig(...) bool` satisfy the requirement?

pkg/multiproject/gce/fake.go
```diff
@@ -0,0 +1,110 @@
...
+	pcKey := providerConfigKey(providerConfig)
+	if g.clientsForProviderConfigs[pcKey] == nil {
+		return nil, fmt.Errorf("GCEFake does not have a client for provider config %q", pcKey)
+	}
+	return g.clientsForProviderConfigs[pcKey], nil
+}
+
+func (g *GCEFake) GCEForProviderConfig(providerConfig *v1.ProviderConfig, logger klog.Logger) (*cloudgce.Cloud, error) {
```
gauravkghildiyal, line 39: Please document what this is doing / creating.

pkg/multiproject/testutil/providerconfigwebhook.go
```diff
@@ -0,0 +1,36 @@
...
+	Add(obj runtime.Object) error
+}
+
+// EmulateProviderConfigLabelingWebhook is a helper function that emulates the behaviour
+// of the providerconfig webhook.
+// It will set the providerconfig name label on the object, on creation.
+// This is needed, as in unit/integration tests, when we create objects, we expect
+// the providerconfig name label to be set.
+// The function takes a fake client and the name of the CRD.
+func EmulateProviderConfigLabelingWebhook(tracker FakeTracker, fake *testing.Fake, crName string) {
```
gauravkghildiyal, line 21: Thanks for documenting for this webhook is supposed to emulate. Let's also mention the use of the "namespace" as a signal about which providerconfig to tie the resource to.

pkg/multiproject/start/start_test.go
```diff
@@ -0,0 +1,596 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			flags.F.ResyncPeriod = 10000 * time.Second
+
+			// Build fake clients.
+			kubeClient := fake.NewSimpleClientset()
+			pcClient := pcclientfake.NewSimpleClientset()
+			svcNegClient := svcnegfake.NewSimpleClientset()
+
+			// This simulates the automatic labeling that the real environment does.
```
gauravkghildiyal, line 182: Let's leave a similar comment here about the implicit use of the "namespace" and how it identifies the providerConfig

gauravkghildiyal (commented):
Awesome test Slavik!

pkg/neg/controller.go
panslava: thanks for catching, this was a debugging leftover, removed

pkg/neg/controller.go
panslava: debugging leftover, removed

pkg/neg/controller.go
panslava: this was a debugging leftover, removed

cmd/glbc/main.go
panslava: Thanks, Done

pkg/multiproject/controller/controller.go
panslava: 😋

pkg/multiproject/neg/neg.go
panslava: Actually, not only for tests. Looks like down there, indexers are shared for the same informer factory, that's why it's needed (otherwise adding already existing indexer returns an error)

pkg/multiproject/gce/gce.go
panslava: Renamed to DefaultGCECreator

pkg/multiproject/gce/fake.go
panslava: Actually I don't need `GetGCEForProviderConfig`, left just GCEForProviderConfig

pkg/multiproject/gce/fake.go
panslava: Done

pkg/multiproject/testutil/providerconfigwebhook.go
panslava: added notion about namespace

pkg/multiproject/start/start_test.go
panslava: added comment

panslava:
Thank you Gaurav! Applied review comments fixes, ready for re-review @gauravkghildiyal

pkg/neg/controller.go
```diff
@@ -178,6 +178,7 @@ func NewController(
 		apiv1.EventSource{Component: "neg-controller"})
 
 	syncerMetrics := syncMetrics.NewNegMetricsCollector(flags.F.NegMetricsExportInterval, logger)
+
```
gauravkghildiyal, line 181: I guess this remaining diff is not intentional (let's not leave a footprint of touching this file in this PR)

pkg/multiproject/testutil/providerconfigwebhook.go
gauravkghildiyal: Thanks!

pkg/multiproject/neg/neg.go
gauravkghildiyal: Thanks for clarifying. Saw the other comment at the function invocation, that'll be really helpful for future readers.

gauravkghildiyal (commented):
Looks good. Just one minor nit about a random modification in pkg/neg/controller.go

pkg/neg/controller.go
panslava: 🧹 done

gauravkghildiyal:
Thanks!

/lgtm

PR #2808, opened 2025-02-21
Fix firewall cr creation

maciejriedl:
While testing gcpfirewalls cr creation inside ingress-gce we encountered an issue when Allow - TPC:all rule is requested.

The former implementation would create a cr with Allow:all instead, which is a violation of the specification:
https://cloud.google.com/kubernetes-engine/docs/concepts/firewall-rules#ingress-fws

### DIFF
--- a/pkg/firewalls/firewalls_test.go
+++ b/pkg/firewalls/firewalls_test.go
@@ -131,12 +131,29 @@ func TestFirewallPoolSyncSrcRanges(t *testing.T) {
 func TestFirewallPoolSyncPorts(t *testing.T) {
 	fwp := NewFakeFirewallsProvider(false, false)
 	fwClient := firewallclient.NewSimpleClientset()
-	fp := NewFirewallPool(fwp, defaultNamer, srcRanges, portRanges(), klog.TODO())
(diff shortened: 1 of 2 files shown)


pkg/firewalls/firewalls_test.go
```diff
@@ -131,13 +131,30 @@ func TestFirewallPoolSyncSrcRanges(t *testing.T) {
...
 
 	if err := fp.Sync(nodes, nil, nil, true); err != nil {
 		t.Fatal(err)
 	}
+	verifyFirewallRule(fwp, ruleName, nodes, srcRanges, emptyPortRanges, t)
+
+	if err := fcrp.Sync(nodes, nil, nil, true); err != nil {
+		t.Fatal(err)
+	}
+	verifyFirewallCR(fwClient, ruleName, srcRanges, emptyPortRanges, true, t)
```
swetharepakula, line 149: Should there be a corresponding change in the verify firewall CR that checks for the tcp:all?

swetharepakula (commented):
Can you add some context in the commit message of why this change is necessary and in what situations this applies?

Thanks!

swetharepakula (commented):
/lgtm
/approve

PR #2809, opened 2025-02-21
Look for correct neg when using multi-networking

swetharepakula:
Multi-networking and multi subnet clusters are currently incompatible features and approach neg naming differently. Therefore, ensure that the neg syncer understands which naming scheme to use and how to detect multi-networking versus multi subnet clusters correctly

/assign @gauravkghildiyal 
/assign @mmamczur

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -255,8 +255,10 @@ func (s *transactionSyncer) syncInternal() error {
 }
 
 func (s *transactionSyncer) syncInternalImpl() error {
-	if s.syncer.IsStopped() || s.syncer.IsShuttingDown() {
-		s.logger.V(3).Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String())
+	isStopped := s.syncer.IsStopped()
+	isShuttingDown := s.syncer.IsShuttingDown()
+	if isStopped || isShuttingDown {
+		s.logger.Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String(), "syncerStopped", isStopped, "syncerShuttingDown", isShuttingDown)
 		return nil
 	}
 	if s.needInit || s.isZoneChange() {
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/transaction_test.go
```diff
@@ -507,6 +507,68 @@ func TestTransactionSyncNetworkEndpointsMSC(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to initialize transaction syncer: %v", err)
+	}
+
+	// Start syncer without starting syncer goroutine
+	(transactionSyncer.syncer.(*syncer)).stopped = false
+	if err := transactionSyncer.ensureNetworkEndpointGroups(); err != nil {
+		t.Errorf("Expect error == nil, but got %v", err)
+	}
+	// var targetPort string
```
mmamczur, line 529: remove the commented code?

pkg/neg/syncers/transaction.go
```diff
@@ -256,7 +256,7 @@ func (s *transactionSyncer) syncInternal() error {
 
 func (s *transactionSyncer) syncInternalImpl() error {
 	if s.syncer.IsStopped() || s.syncer.IsShuttingDown() {
-		s.logger.V(3).Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String())
+		s.logger.Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String(), "syncerStopper", s.syncer.IsStopped(), "syncerShuttingDown", s.syncer.IsShuttingDown())
```
mmamczur, line 259: "syncerStopper" -> "syncerStopped"?

pkg/neg/syncers/transaction_test.go
swetharepakula: fixed. Thanks for catching!

pkg/neg/syncers/transaction.go
swetharepakula: fixed. Thanks for catching!

pkg/neg/syncers/transaction.go
```diff
@@ -256,7 +256,7 @@ func (s *transactionSyncer) syncInternal() error {
 
 func (s *transactionSyncer) syncInternalImpl() error {
 	if s.syncer.IsStopped() || s.syncer.IsShuttingDown() {
-		s.logger.V(3).Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String())
+		s.logger.Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String(), "syncerStopped", s.syncer.IsStopped(), "syncerShuttingDown", s.syncer.IsShuttingDown())
```
gauravkghildiyal, line 259: Thanks, this is a useful log message.

This may be overthinking, but should we record the state returned by the functions in a variable, and then log the variables? Basically trying to avoid a confusing scenario where the two independent executions returned a different result and our log message ends up confusing us.

gauravkghildiyal (commented):
Thanks!
Looks good, just one minor comment

pkg/neg/syncers/transaction.go
swetharepakula: Good catch! The lock should be held when this is run, so this should be safe, but making the change is better to make sure the future refactors don't report incorrectly

gauravkghildiyal: Thanks for incorporating the suggestion. I believe we missed updating the condition to use the variable, right? (to make sure both the condition and log are based on the same source of truth)

gauravkghildiyal:
Thanks :)

/lgtm

pkg/neg/syncers/transaction.go
swetharepakula: Thanks for catching! Should be fixed now.

PR #2810, opened 2025-02-21
Update ProviderConfig CRD types

panslava:
/assign @gauravkghildiyal


pkg/multiproject/gce/gce_test.go
```diff
@@ -165,9 +165,11 @@ subnetwork-name = default-subnetwork
 				Spec: v1.ProviderConfigSpec{
 					ProjectID:     "providerconfig-project-id",
 					ProjectNumber: 654321,
-					NetworkConfig: &v1.NetworkConfig{
-						Network:           "providerconfig-network-url",
-						DefaultSubnetwork: "providerconfig-subnetwork-url",
+					NetworkConfig: v1.ProviderNetworkConfig{
+						Network: "providerconfig-network-url",
```
gauravkghildiyal, line 169: nit: I suspect in the actual Object, this may not be a URL, so perhaps good to change the test to correctly reflect that.

gauravkghildiyal (commented):
Looks good, just one minor nit.

pkg/multiproject/gce/gce_test.go
gauravkghildiyal: Update: I'm trying to ensure that we get a full path here for both the network and subnet. So do make sure if you need to extract just the name, it's taken care of. Let's tackle that in a separate PR.

PR #2811, opened 2025-02-24
script: Use instance zone instead of cluster location in setup-local script

arvindbr8:
Fixes a bug that prevented the script from working with regional clusters. 

Previously, the script incorrectly used the cluster's region as the instance group zone. This change now extracts the correct zone from the instance group's URL, ensuring compatibility with regional clusters.

### DIFF
--- a/hack/setup-local.sh
+++ b/hack/setup-local.sh
@@ -78,4 +78,4 @@ node-tags = ${nodeTag}
 local-zone = ${zone}
 EOF
 
-echo "Run glbc with hack/run-glbc.sh"
+echo "Run glbc with hack/run-local-glbc.sh"
--- a/hack/setup-local.sh
+++ b/hack/setup-local.sh
@@ -15,12 +15,6 @@ parseCluster() {
     project=$(echo ${selfLink} | sed 's+.*/projects/\([-a-z0-9]*\)/.*+\1+')
 }
 
-parseInstance() {
-    local name=$1
-    # Globals.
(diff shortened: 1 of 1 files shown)


hack/setup-local.sh
```diff
@@ -37,7 +31,7 @@ if [ -z "$clusterLocation" ]; then
 else
     clusters=$(gcloud container clusters list --format="${fmt}" --filter="name:${clusterName} location:${clusterLocation}")
 fi
-if [ $(echo "${cluster}" | wc -l) -gt 1 ]; then
+if [ "$(echo "${clusters}" | wc -l)" -gt 1 ]; then
```
gauravkghildiyal, line 34: nit: Maybe change the variables to "cluster" since we really want this to have one value and use it in that manner in the rest of the script.

gauravkghildiyal (commented):
Thank you for the improvement

hack/setup-local.sh
arvindbr8: Makes sense. A followup to that: why do we use a `gcloud container clusters list`, instead of `gcloud container clusters describe`?  That way we can get rid of the `-gt 1` check and check for $? instead?

gauravkghildiyal: I can only guess that maybe someone thought the --filter only works when you list. I agree it's not needed.

gauravkghildiyal: I'll go ahead with the submit for now and any nitpicks we can address separately. Overall I still this this is a functionally sound change.

arvindbr8:
Thanks for the review @gauravkghildiyal

PR #2815, opened 2025-02-26
Add zonal affinity support

08volt:
## Overview
This PR implements zonal affinity capabilities for L4 Internal Load Balancers, allowing traffic to stay within zones when possible for improved performance and reduced cross-zone data transfer costs.

## Changes
- Add zonal affinity flag for L4 ILB controllers
- Implement backendService zonal affinity control logic
- Add backendService minimum API version control for zonal features
- Update backendService tests to include zonal affinity cases with expected NetworkPassThroughLbTrafficPolicy
- Add handler for service spec Spec.TrafficDistribution change events
- Add L4 TestZonalAffinity integration tests

## Usage
Users can enable zonal affinity by configuring their service with `Spec.TrafficDistribution=PreferClose`.

## Testing
- Added unit tests for backendService with zonal affinity configurations
- Added L4 TestZonalAffinity integration tests to verify controller functionality

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -35,6 +35,9 @@ const (
 	DefaultConnectionDrainingTimeoutSeconds = 30
 	defaultTrackingMode                     = "PER_CONNECTION"
 	PerSessionTrackingMode                  = "PER_SESSION" // the only one supported with strong session affinity
+	ZonalAffinityEnabledSpillover           = "ZONAL_AFFINITY_SPILL_CROSS_ZONE"
+	DefaultZonalAffinitySpilloverRatio      = 0
+	ZonalAffinityDisabledSpillover          = "ZONAL_AFFINITY_DISABLED"
 )
 
 // LocalityLBPolicyType is the type of locality lb policy the backend service should use.
@@ -90,6 +93,36 @@ type L4BackendServiceParams struct {
 	NetworkInfo              *network.NetworkInfo
 	ConnectionTrackingPolicy *composite.BackendServiceConnectionTrackingPolicy
 	LocalityLbPolicy         LocalityLBPolicyType
+	EnableZonalAffinity      bool
+}
+
+var versionPrecedence = map[meta.Version]int{
+	meta.VersionAlpha: 2,
+	meta.VersionBeta:  1,
+	meta.VersionGA:    0,
+}
+
+// selectApiVersionForUpdate selects the appropriate API version for updating a backend service.
+// It compares the current API version (potentially retrieved from the service's description) with the expected API version.
+// The function returns the API version with the higher feature set to ensure no functionality is lost during the update.
+func selectApiVersionForUpdate(current, expected meta.Version) meta.Version {
+	precedenceCurrent, okCurrent := versionPrecedence[current]
+	precedenceExpected, okExpected := versionPrecedence[expected]
+
+	if !okCurrent {
+		precedenceCurrent = 0
+		current = meta.VersionGA
+	}
+
+	if !okExpected {
+		precedenceExpected = 0
+		expected = meta.VersionGA
+	}
+
+	if precedenceCurrent > precedenceExpected {
+		return current
+	}
+	return expected
 }
 
 // ensureDescription updates the BackendService Description with the expected value
@@ -327,6 +360,14 @@ func (p *Pool) DeleteSignedURLKey(be *composite.BackendService, keyName string, 
 	return nil
 }
 
+// apiVersionRequiredbyServiceFeatures to create a backend service with the given params
+func apiVersionRequiredbyServiceFeatures(params L4BackendServiceParams) meta.Version {
+	if params.EnableZonalAffinity {
+		return meta.VersionBeta
+	}
+	return meta.VersionGA
+}
+
 // EnsureL4BackendService creates or updates the backend service with the given name.
 func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger klog.Logger) (*composite.BackendService, utils.ResourceSyncStatus, error) {
 	start := time.Now()
@@ -342,11 +383,14 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
 	if err != nil {
 		return nil, utils.ResourceResync, err
 	}
-	currentBS, err := composite.GetBackendService(p.cloud, key, meta.VersionGA, beLogger)
+	expectedVersion := apiVersionRequiredbyServiceFeatures(params)
+
+	currentBS, err := composite.GetBackendService(p.cloud, key, expectedVersion, beLogger)
 	if err != nil && !utils.IsNotFoundError(err) {
 		return nil, utils.ResourceResync, err
 	}
-	desc, err := utils.MakeL4LBServiceDescription(params.NamespacedName.String(), "", meta.VersionGA, false, utils.ILB)
+
+	expectedDesc, err := utils.MakeL4LBServiceDescription(params.NamespacedName.String(), "", expectedVersion, false, utils.ILB)
 	if err != nil {
 		beLogger.Info("EnsureL4BackendService: Failed to generate description for BackendService", "err", err)
 	}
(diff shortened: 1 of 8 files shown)


pkg/flags/flags.go
```diff
@@ -329,6 +330,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.NodeTopologyCRName, "node-topology-cr-name", "default", "The name of the Node Topology CR.")
 	flag.BoolVar(&F.EnableWeightedL4ILB, "enable-weighted-l4-ilb", false, "Enable Weighted Load balancing for L4 ILB.")
 	flag.BoolVar(&F.EnableWeightedL4NetLB, "enable-weighted-l4-netlb", false, "EnableWeighted Load balancing for  L4 NetLB .")
+	flag.BoolVar(&F.EnableZonalAffinity, "enable-zonal-affinity", false, "Enable Zonal Affinity for L4 ILB.")
```
TortillaZHawaii, line 333: If it is for L4 ILB, shouldn't the flag have L4ILB at the end, like weighted?

pkg/loadbalancers/l4_test.go
```diff
@@ -2388,6 +2388,107 @@ func TestWeightedILB(t *testing.T) {
...
+			trafficDistributionSpec:         &zonalAffinityTrafficDistributionKey,
+			wantZonalAffinityEnabled:        false,
+			wantZonalAffinitySpilloverRatio: 0,
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
+			// t.Parallel()
```
TortillaZHawaii, line 2430: Can this be run in parallel? If so please remove the dashes, if not remove the line

pkg/flags/flags.go
TortillaZHawaii:
```suggestion
	flag.BoolVar(&F.EnableZonalAffinityL4ILB, "enable-zonal-affinity-l4-ilb", false, "Enable Zonal Affinity for L4 ILB.")
```

pkg/backends/backends.go
```diff
@@ -447,9 +506,31 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
...
+}
+
+func zonalAffinityEqual(a, b *composite.BackendService) bool {
+	aZonalAffinity := convertNetworkLbTrafficPolicyToZonalAffinity(a.NetworkPassThroughLbTrafficPolicy)
+	bZonalAffinity := convertNetworkLbTrafficPolicyToZonalAffinity(b.NetworkPassThroughLbTrafficPolicy)
+
+	// Compare Spillover values
+	spilloverEqual := aZonalAffinity.Spillover == bZonalAffinity.Spillover
+	// Compare SpilloverRatio values
+	spilloverRatioEqual := aZonalAffinity.SpilloverRatio == bZonalAffinity.SpilloverRatio
```
TortillaZHawaii, line 529: nit: Are those // compare comments necessary?

pkg/backends/backends.go
```diff
@@ -90,6 +93,31 @@ type L4BackendServiceParams struct {
...
+}
+
+var versionPrecedence = map[meta.Version]int{
+	meta.VersionAlpha: 2,
+	meta.VersionBeta:  1,
+	meta.VersionGA:    0,
+}
+
+// maxVersion returns the higher version based on precedence.
+func maxVersion(a, b meta.Version) meta.Version {
```
TortillaZHawaii, line 106: Is this tested?

TortillaZHawaii (changes requested):

pkg/flags/flags.go
08volt: Thats correct, thanks for the suggetion

pkg/loadbalancers/l4_test.go
08volt: Yes I will uncomment

pkg/backends/backends.go
08volt: Adding the tests right now, thanks

pkg/flags/flags.go
```diff
@@ -329,6 +330,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.NodeTopologyCRName, "node-topology-cr-name", "default", "The name of the Node Topology CR.")
 	flag.BoolVar(&F.EnableWeightedL4ILB, "enable-weighted-l4-ilb", false, "Enable Weighted Load balancing for L4 ILB.")
 	flag.BoolVar(&F.EnableWeightedL4NetLB, "enable-weighted-l4-netlb", false, "EnableWeighted Load balancing for  L4 NetLB .")
+	flag.BoolVar(&F.EnableL4ILBZonalAffinity, "enable-zonal-affinity", false, "Enable Zonal Affinity for L4 ILB.")
```
TortillaZHawaii, line 333:
```suggestion
	flag.BoolVar(&F.EnableL4ILBZonalAffinity, "enable-zonal-affinity-l4-ilb", false, "Enable Zonal Affinity for L4 ILB.")
```

pkg/backends/backends.go
```diff
@@ -327,6 +358,14 @@ func (p *Pool) DeleteSignedURLKey(be *composite.BackendService, keyName string,
 	return nil
 }
 
+// minRequiredVersion to create a backend service with the given params
+func minRequiredVersion(params L4BackendServiceParams) meta.Version {
```
mmamczur, line 362: could we name this so that it clearly states it's for the apiVersion? maybe something like `apiVersionRequiredbyServiceFeatures()`

pkg/backends/backends.go
```diff
@@ -398,6 +449,17 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
...
+		// Use the version with most priority if the BackendService was already using one
+		currentVersion := meta.VersionGA
+		var currentDesc utils.L4LBResourceDescription
+		err = currentDesc.Unmarshal(currentBS.Description)
+		if err != nil {
+			beLogger.V(0).Error(err, "EnsureL4BackendService: error unmarshaling backend service description")
+		} else {
+			currentVersion = currentDesc.APIVersion
+		}
+		expectedBS.Version = maxVersion(currentVersion, expectedBS.Version)
```
mmamczur, line 462: a question: so with this solution once the LB starts using Beta it can't ever come back to GA, so my question is, if you have a BackendService created using beta with the zonal affinity option set. What will happen if you turn the featue off and update the BackendService with the GA API? will zonal affinity be turned off or will it keep what was set using beta?

if using GA can turn off zonal affinity maybe we can make it possible to go back to GA?

pkg/backends/backends.go
```diff
@@ -342,25 +381,37 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
...
 		LoadBalancingScheme: params.Scheme,
 		LocalityLbPolicy:    string(params.LocalityLbPolicy),
 	}
 
+	if params.EnableZonalAffinity {
+		beLogger.V(2).Info("EnsureL4BackendService: using Zonal Affinity", "spillover", ZonalAffinityEnabledSpillover, "spilloverRatio", DefaultZonalAffinitySpilloverRatio)
+		expectedBS.NetworkPassThroughLbTrafficPolicy = zonalAffinityEnabledTrafficPolicy()
+	} else {
+		beLogger.V(2).Info("EnsureL4BackendService: not using Zonal Affinity", "spillover", ZonalAffinityDisabledSpillover, "spilloverRatio", DefaultZonalAffinitySpilloverRatio)
+		expectedBS.NetworkPassThroughLbTrafficPolicy = zonalAffinityDisabledTrafficPolicy()
```
mmamczur, line 412: Pay extra attention here not to trigger updates of services that we don't want to update.
I see that later the comparison treats empty value as not set at all which should be ok

I'd consider putting this whole if else into a block guarded by the feature flag just in case.

pkg/backends/backends.go
```diff
@@ -90,6 +93,34 @@ type L4BackendServiceParams struct {
...
+}
+
+var versionPrecedence = map[meta.Version]int{
+	meta.VersionAlpha: 2,
+	meta.VersionBeta:  1,
+	meta.VersionGA:    0,
+}
+
+// maxVersion returns the higher version based on precedence.
+func maxVersion(a, b meta.Version) meta.Version {
```
mmamczur, line 106: can we name this differently?

My understanding looking at the code is that this is the version that should be used based on the API version used to create the service and the thing that is required by the features we use.

It would be best if the function name explained what it is doing exactly

pkg/backends/backends.go
08volt: It is not possible to turnOff ZA with GA, it will simply ignore the field and don't update. In order to turn off zonal affinity the field "Spillover" needs to be set to ZONAL_AFFINITY_DISABLED

pkg/backends/backends.go
08volt: The feature is always compared with a  Disabled/Enabled conversion  ( empty means Disblaed, Explicitly Disabled means Disabled )

pkg/backends/backends.go
08volt: Sure

pkg/backends/backends.go
08volt: Sure

pkg/backends/backends.go
08volt: It actually is a neutral function that is used to compare the "priority" of the versions

pkg/backends/backends.go
08volt: By the way, once the zonal affinity is disabled, also the description will be set to GA so it will use GA as baseline for the next update

pkg/backends/backends_test.go
```diff
@@ -734,3 +833,205 @@ func TestUpdateLocalityLBPolicy(t *testing.T) {
 		})
 	}
 }
+
+func TestHigherPrecedenceAPIVersion(t *testing.T) {
```
mmamczur, line 837: can you also rename the tests to match the changed function name?

mmamczur:
looks good now, let me know when you want to merge this

08volt:
@mmamczur looks ready

PR #2816, opened 2025-02-27
Garbage collect TO_BE_DELETED NEGs

swetharepakula:
This PR is dependent on the merging of https://github.com/kubernetes/ingress-gce/pull/2812.

/assign @gauravkghildiyal

### DIFF
--- a/pkg/neg/manager.go
+++ b/pkg/neg/manager.go
@@ -547,16 +547,23 @@ func (manager *syncerManager) garbageCollectNEG() error {
 	return nil
 }
 
+type deletionCandidate struct {
+	neg *negv1beta1.ServiceNetworkEndpointGroup
+	// tbdOnly indicates that only the NEGs that are in the TBD state should be deleted
+	tbdOnly bool
+}
+
 // garbageCollectNEGWithCRD uses the NEG CRs and the svcPortMap to determine which NEGs
 // need to be garbage collected. Neg CRs that do not have a configuration in the svcPortMap will deleted
 // along with all corresponding NEGs in the CR's list of NetworkEndpointGroups. If NEG deletion fails in
 // the cloud, the corresponding Neg CR will not be deleted
 func (manager *syncerManager) garbageCollectNEGWithCRD() error {
-	deletionCandidates := map[string]*negv1beta1.ServiceNetworkEndpointGroup{}
+
+	deletionCandidates := map[string]deletionCandidate{}
 	negCRs := manager.svcNegLister.List()
 	for _, obj := range negCRs {
 		neg := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
-		deletionCandidates[neg.Name] = neg
+		deletionCandidates[neg.Name] = deletionCandidate{neg: neg, tbdOnly: false}
 	}
 
 	func() {
@@ -570,8 +577,13 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
 				// In the situation a neg config is in the svcPortMap but the CR has a deletion timestamp, then
 				// neither the neg nor CR will not be deleted. In the situation a neg config is not in the svcPortMap,
 				// but the CR does not have a deletion timestamp, both CR and neg will be deleted.
-				if _, ok := deletionCandidates[portInfo.NegName]; ok {
-					delete(deletionCandidates, portInfo.NegName)
+				candidate, ok := deletionCandidates[portInfo.NegName]
+				if ok {
+					if !containsTBDNeg(candidate.neg) {
+						delete(deletionCandidates, portInfo.NegName)
+					} else {
+						deletionCandidates[portInfo.NegName] = deletionCandidate{neg: candidate.neg, tbdOnly: true}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/neg/manager.go
```diff
@@ -620,20 +632,34 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
...
+func containsTBDNeg(negCR *negv1beta1.ServiceNetworkEndpointGroup) bool {
+	for _, negRef := range negCR.Status.NetworkEndpointGroups {
+		if negRef.State == negv1beta1.ToBeDeletedState {
+			return true
+		}
+	}
+	return false
+}
+
 // processNEGDeletionCandidate attempts to delete `svcNegCR` and all NEGs
```
gauravkghildiyal, line 644: Please update the comment as well.

pkg/neg/manager_test.go
```diff
@@ -1532,6 +1533,190 @@ func TestGarbageCollectionNegCrdEnabled(t *testing.T) {
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: testServiceNamespace,
+			Name:      testServiceName,
+		},
+	}
+	svc.SetUID("svc-uid")
```
gauravkghildiyal, line 1547: nit: Why not directly set the uid in the ObjectMeta?

pkg/neg/manager_test.go
```diff
@@ -1532,6 +1533,190 @@ func TestGarbageCollectionNegCrdEnabled(t *testing.T) {
...
+
+			expectedNegCount := len(tc.expectedActiveNegs) + len(tc.expectedTBDNegs)
+
+			if len(negs) != expectedNegCount {
+				t.Errorf("got %d NEGs, expected %d Negs", len(negs), expectedNegCount)
+			}
+
+			var expectedNegRefs []negv1beta1.NegObjectReference
+			for _, negRes := range tc.expectedActiveNegs {
+				count := checkForNegDeletions(negs, negRes.Name)
```
gauravkghildiyal, line 1686: This function name seems misleading?

pkg/neg/manager.go
```diff
@@ -620,20 +632,34 @@ func (manager *syncerManager) garbageCollectNEGWithCRD() error {
...
+
 	deleteByZone := len(svcNegCR.Status.NetworkEndpointGroups) == 0
 	// Change this to a map from NEG name to sets once we allow multiple NEGs
 	// in a specific zone(multi-subnet cluster).
 	deletedNegs := make(map[negtypes.NegInfo]struct{})
 
 	for _, negRef := range svcNegCR.Status.NetworkEndpointGroups {
+		if candidate.tbdOnly && negRef.State != negv1beta1.ToBeDeletedState {
+			continue
+		}
```
gauravkghildiyal, line 662: Would it be overthinking to consider what happens if we fail to ParseResourceURL for a TBD NEG? I can't imagine how that would happen but if it does, apparently it sets `deleteByZone` which ends up deleting all of the NEGs.

swetharepakula: This is an interesting scenario that I was thinking about as well. My initial thought was that maybe we attempt by zone and if the NEGs are being used, they will fail deletion and should be okay.

However, I just realized in this situation we have lost the NEG name for non default subnets, so we guaranteed are trying to delete the wrong negs. So perhaps we should instead emit an event on the svcneg or on the service to ask the user to manually delete NEGs of that subnet?

pkg/neg/manager_test.go
swetharepakula: changed the name in a separate commit.

pkg/neg/manager_test.go
swetharepakula: set in the object meta.Copied this from the previous test, so not sure why this was done separately.

pkg/neg/manager.go
swetharepakula: Comment has been updated to reflect TBD NEGs and that svcneg is updated accordingly.

pkg/neg/manager.go
swetharepakula: Added an event on the svc neg for now. Do you think it makes more sense to have it on the service object? I feel like it is little more difficult to direct users to the svcneg then.

gauravkghildiyal: My initial thoughts were also the same that having it on the Service object is more visible. But I'm thinking that what we're concerned with here is the GC, which could very well have been triggered by the Service itself being deleted. So it anyways makes more sense to definitely send GC related events on an object that we own and control and which will exist to receive the event. (Yes one could argue that maybe it's even better to send events on both SvcNEG and Service, but for this particular, it seems like an overkill, unless we really see this happening)

Sound fair?

gauravkghildiyal:
Thanks! Please unhold if we agree regarding https://github.com/kubernetes/ingress-gce/pull/2816#discussion_r1976255332 

/lgtm
/hold

pkg/neg/manager.go
swetharepakula: That sounds reasonable to me. I am okay with this, thanks for the discussion!

swetharepakula:
Agree with the rationale above, so remove hold.
/unhold

/retest

PR #2817, opened 2025-03-03
Make the subset sizes configurable for L4 NetLB NEG.

mmamczur:
Add `--l4-netlb-local-subset-size` and `--l4-netlb-local-subset-size` flags that allow to configure the subset sizes for the L4 NetLB endpoint calculators.

### DIFF
--- a/pkg/neg/syncers/subsets.go
+++ b/pkg/neg/syncers/subsets.go
@@ -35,10 +35,6 @@ const (
 	maxSubsetSizeLocal = 250
 	// Max number of subsets in ExternalTrafficPolicy:Cluster, which is the default mode.
 	maxSubsetSizeDefault = 25
-	// Max number of subsets for NetLB in ExternalTrafficPolicy:Local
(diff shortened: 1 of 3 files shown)


pkg/neg/syncers/subsets.go
```diff
@@ -35,10 +35,6 @@ const (
 	maxSubsetSizeLocal = 250
 	// Max number of subsets in ExternalTrafficPolicy:Cluster, which is the default mode.
 	maxSubsetSizeDefault = 25
-	// Max number of subsets for NetLB in ExternalTrafficPolicy:Local
```
swetharepakula, line 38: instead of removing completely, should there be a default if the flags are not provided?

mmamczur: the flags have a default baked in. Maybe it would be nice to use these constants used in the flag def but I'm guessing it will make a circular dep then?

mmamczur:
/retest

weird faiure, locally tests seem to work bu then the failure looks like could potentially be caused by flags change?

PR #2820, opened 2025-03-04
[Cherry-pick #2819 -> 1.33] Increase max number of subsets NetLB for ETP:local

FelipeYepez:
Match the max subset size with IG per zone for NEG backed NetLB.


swetharepakula (commented):
/lgtm
/approve

PR #2821, opened 2025-03-04
Temporarily disable linters

mmamczur:
The recent upgrade to go 1.24 in kubernetes/kubernetes causes linters to fail in many ways for code that is correct. Temporarily disable not to block development, the linters will be enabled once we determine the cause of the issue.


swetharepakula:
Sorry missed this PR, but same thing as this PR https://github.com/kubernetes/ingress-gce/pull/2823/

PR #2822, opened 2025-03-04
Change multiproject auth API to support per-provider-config token

panslava:
/assign @gauravkghildiyal

### DIFF
--- a/pkg/multiproject/gce/gce.go
+++ b/pkg/multiproject/gce/gce.go
@@ -80,9 +80,15 @@ func generateConfigForProviderConfig(defaultConfigContent string, providerConfig
 	oldValue := globalSection.Key(tokenURLKey).String()
 	// Extract location from the old token URL
 	location := extractLocationFromTokenURL(oldValue)
-	// Format: https://gkeauth.googleapis.com/v1/projects/{TENANT_PROJECT_NUMBER}/locations/{TENANT_LOCATION}/tenants/{TENANT_ID}:generateTenantToken"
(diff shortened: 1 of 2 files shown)


pkg/multiproject/gce/gce.go
```diff
@@ -77,16 +77,18 @@ func generateConfigForProviderConfig(defaultConfigContent string, providerConfig
...
-	projectNumberInt := providerConfig.Spec.ProjectNumber
-	projectNumberStr := fmt.Sprintf("%d", projectNumberInt)
-	newTokenURL := replaceProjectNumberInTokenURL(tokenURL, projectNumberStr)
-	globalSection.Key(tokenURLKey).SetValue(newTokenURL)
+	oldValue := globalSection.Key(tokenURLKey).String()
+	// Extract location from the old token URL
+	location := extractLocationFromTokenURL(oldValue)
+	// Format: https://gkeauth.googleapis.com/v1/projects/{TENANT_PROJECT_NUMBER}/locations/{TENANT_LOCATION}/tenants/{TENANT_ID}:generateTenantToken"
+	formatString := "https://gkeauth.googleapis.com/v1/projects/%d/locations/%s/tenants/%s:generateTenantToken"
+	tokenURL := fmt.Sprintf(formatString, providerConfig.Spec.ProjectNumber, location, providerConfig.Name)
```
gauravkghildiyal, line 85: The "gkeauth.googlapis.com" address would vary across different environments (staging, prod, etc). This API address hasn't changed, so we should use whatever address is set in gce.conf and only modify the "projects/...." part.

panslava: great idea, fixed

PR #2824, opened 2025-03-05
Rework per-provider-config auth. Assume cluster project if no label set

panslava:
/assign @gauravkghildiyal


gauravkghildiyal:
Thanks

/lgtm

PR #2825, opened 2025-03-05
Subnet change triggers ensureNetworkEndpointGroups

swetharepakula:
When subnet change is detected, we need to update the SvcNegCR, therefore we should trigger ensureNetworkEndpointGroups

/assign @gauravkghildiyal


pkg/neg/syncers/transaction.go
```diff
@@ -856,6 +860,34 @@ func (s *transactionSyncer) isZoneChange() bool {
 	return !currZones.Equal(existingZones)
 }
 
+func (s *transactionSyncer) isSubnetChange() bool {
```
gauravkghildiyal, line 863: I think we would want to flag-guard the behaviour of this function, since it depends on the existing of the SubnetURL in the SvcNEG? In the absence of a flag-guard, if the SubnetURL isn't populated, this always return true. Maybe we should always return false when the flag is not enabled to maintain?

swetharepakula: That's a good point. SubnetChange uses ListSubnets which is flag gated, but we need to handle the fact that SubnetURL might not be populated

swetharepakula: New version handles the empty Subnet URL properly

gauravkghildiyal:
Thanks. This implementation looks quite flexible.

/lgtm

PR #2826, opened 2025-03-06
Regenerate ProviderConfig types to be Cluster scoped, not Namespace scoped

panslava:
/assign @gauravkghildiyal


gauravkghildiyal:
Thanks Slavik, we'll have to adjust some usages and logging as well:

E.g. https://github.com/kubernetes/ingress-gce/blob/0f70024e0383ab3ed62c65d068c9fcacc98303cb/pkg/utils/patch/patch.go#L104

panslava:
@gauravkghildiyal thank you, fixed

gauravkghildiyal:
Thanks @panslava 

/lgtm

PR #2829, opened 2025-03-17
Setup script for GLBC configuration on regional clusters

08volt:
This PR introduces a bash script (setup-local-regional.sh) to simplify GLBC configuration for regional GKE clusters.

The script automates the process of gathering necessary information about a regional cluster, with logic to find a running instance in the cluster's instance groups. including:

- Network name    
- Subnetwork name    
- Project ID    
- Node tags    
- Instance zone    

It then generates a gce.conf file containing these parameters, which is required for running GLBC.

### DIFF
--- a/hack/setup-local.sh
+++ b/hack/setup-local.sh
@@ -8,11 +8,12 @@ parseCluster() {
     # These are all globals.
     net=$1
     subnet=$2
-    zone=$3
-    selfLink=$4
+    selfLink=$3
+    name=$4
     net=$(echo ${net} | sed 's+.*networks/\([-a-z0-9]*\).*$+\1+')
     subnet=$(echo ${subnet} | sed 's+.*subnetworks/\([-a-z0-9]*\)$+\1+')
     project=$(echo ${selfLink} | sed 's+.*/projects/\([-a-z0-9]*\)/.*+\1+')
+    clusterLocation=$(echo ${selfLink} | sed 's+.*/locations/\([-a-z0-9]*\)/.*+\1+')
 }
 
 clusterName="$1"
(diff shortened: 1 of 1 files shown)


hack/setup-local-regional.sh
```diff
@@ -0,0 +1,111 @@
...
+}
+
+clusterName="$1"
+clusterRegion="$2"
+
+if [ -z "${clusterName}" ]; then
+    echo "Usage: $0 CLUSTER_NAME [REGION]"
+    echo
+    echo "REGION should be a region for regional clusters (e.g., us-central1)."
+    echo "REGION is optional if there is only one cluster with CLUSTER_NAME"
```
TortillaZHawaii, line 35: I tried with only cluster that has a unique name and got an error:
```sh
./hack/setup-local.sh kluster-xl            
ERROR: (gcloud.container.clusters.describe) argument --region: expected one argument
Usage: gcloud container clusters describe NAME [optional flags]
  optional flags may be  --help | --location | --region | --zone

For detailed information on this command and its flags, run:
  gcloud container clusters describe --help
```

With region set it works well.

TortillaZHawaii (changes requested):
I was fighting this today. Good to see someone working on it.

Do we need a separate file to setup regional? Can't previous one be extended?

hack/setup-local-regional.sh
```diff
@@ -0,0 +1,111 @@
...
+token-url = nil
+project-id = ${project}
+network-name = ${net}
+subnetwork-name = ${subnet}
+node-tags = ${nodeTag}
+local-zone = ${instance_zone}
+EOF
+
+echo "----"
+echo "Run glbc with hack/run-glbc.sh"
```
TortillaZHawaii, line 111: I know this is copied from `hack/setup-local.sh`, however `hack/run-glbc.sh` doesn't exist.
```suggestion
echo "Run glbc with hack/run-local-glbc.sh"
```

08volt: Fixed

hack/setup-local-regional.sh
08volt: Thanks Dawid for testing it, should be fixed now

TortillaZHawaii: Looks good!

hack/setup-local-regional.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
mmamczur, line 1: would it be possible to merge these 2 scripts so that they can figure out if the cluster is zonal or regional?

TortillaZHawaii (changes requested):

PR #2830, opened 2025-03-18
Fixed setup script for zonal cluster

08volt:
Fixed setup script, now working for both zonal and regional clusters


08volt:
@mmamczur @TortillaZHawaii

TortillaZHawaii:
Verified with zonal and regional, with and without location.

Haven't checked for clusters with the same name, but different locations.

/lgtm

08volt:
Tested also with multiple clusters with same name and different zones

PR #2831, opened 2025-03-18
Updated google.golang.org/api to v0.226.0

08volt:
Updated go version to 1.23.7
Updated google.golang.org/api to v0.226.0

run
```
go mod tidy
go mod vendor
```

updated generated code using
```
./hack/update-codegen.sh
```


go.mod
```diff
@@ -1,23 +1,23 @@
 module k8s.io/ingress-gce
 
-go 1.22.8
+go 1.23.0
```
TortillaZHawaii, line 3: For context, it can't be v1.24 due to https://github.com/kubernetes/ingress-gce/pull/2823.

However, shouldn't we use latest minor v1.23.7 with latest security patches? Or use v1.23 instead of v1.23.0?
https://go.dev/doc/devel/release#go1.23.minor

08volt: updated to 1.23.7 thanks

PR #2843, opened 2025-03-26
Add zonal affinity to L4 load balancer metrics

08volt:
# Add zonal affinity to L4 load balancer metrics

## Description
This PR adds support for tracking zonal affinity usage in L4 load balancer metrics. The zonal affinity feature allows users to configure traffic distribution policies for L4 ILB by setting `spec.trafficDistribution` to `PreferClose` which creates load balancers that prefer routing traffic to backends in the same zone.

## Changes
- Added `zonal_affinity` label to both L4 metrics
- Updated metric initialization to capture zonal affinity status
- Added zonal affinity verification to metric test functions

## Testing
- Updated unit tests to verify proper reporting of zonal affinity metrics

### DIFF
--- a/pkg/l4lb/metrics/metrics.go
+++ b/pkg/l4lb/metrics/metrics.go
@@ -44,6 +44,7 @@ const (
 	l4LBControllerPanicsMetricName                 = "l4_controllers_panics_count"
 	L4SyncDetailsMetricName                        = "l4_sync_details_count"
 	l4WeightedLBPodsPerNodeMetricName              = "l4_weighted_lb_pods_per_node"
+	l4ILBZonalAffinityMetricName                   = "l4_ilb_zonal_affinity"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/metrics/l4_metrics.go
```diff
@@ -165,6 +168,7 @@ func (im *ControllerMetrics) exportL4NetLBsMetrics() {
 			l4LabelStrongSessionAffinity: strconv.FormatBool(svcState.StrongSessionAffinity),
 			l4LabelWeightedLBPodsPerNode: strconv.FormatBool(svcState.WeightedLBPodsPerNode),
 			l4LabelBackendType:           string(svcState.BackendType),
+			l4LabelZonalAffinity:         strconv.FormatBool(svcState.ZonalAffinity),
```
FelipeYepez, line 171: We only need it for ILB

08volt:
@mmamczur

pkg/l4lb/metrics/metrics.go
```diff
@@ -57,6 +58,10 @@ var (
 		"backend_type", // type of the backends of the LB (IG or NEG)
 	}
 
+	l4SyncLatencyILBSpecificMetricLabels = []string{
```
mmamczur, line 61: this shouldn't be named `l4SyncLatencyILBSpecificMetricLabels` if it's used in a non latency metric

maybe something like just ilbSpecificMetricLabels (just don't say it's for sync if it isn't just for sync)

08volt: Thats right, named l4SyncILBSpecificMetricLabels since it is for Sync Latency and Errors

PR #2852, opened 2025-04-18
Return atleast the default subnet if the NodeTopology has empty subnets.

gauravkghildiyal:
This slightly increases the robustness with the handling of NodeTopology CR.

/assign @swetharepakula


swetharepakula:
This LGTM! Can you add a unit test for this as well. Thanks!

PR #2855, opened 2025-04-29
Implement loadBalancerClass support for Services of `type: LoadBalancer`

FelipeYepez:
This pull request introduces support for the `spec.loadBalancerClass` field for Services of `type: LoadBalancer`. The `spec.loadBalancerClass` field allows users to explicitly specify which load balancer controller should manage their LoadBalancer Services. Users can now choose the specific load balancer implementation that best suits their needs.


pkg/annotations/gce_annotations.go
```diff
@@ -37,11 +37,33 @@ const (
...
 )
 
 // GetLoadBalancerAnnotationType returns the type of GCP load balancer which should be assembled.
 func GetLoadBalancerAnnotationType(service *v1.Service) LoadBalancerType {
 	var lbType LoadBalancerType
+	// Check LoadBalancerClass before load balancer type annotation since it has precedence.
+	if HasLoadBalancerClass(service, RegionalInternalLoadBalancerClass) {
+		return LBTypeInternal
+	} else if HasLoadBalancerClass(service, RegionalExternalLoadBalancerClass) {
+		return lbType
```
mmamczur, line 65: What is the value returned here? looks like the default value for LoadBalancerType type straight from the variable declaration. Let's return the external lb constant here

FelipeYepez: The value was the default empty. Modified the function to return a new "External" constant

PR #2856, opened 2025-04-29
Update gke networking api to gke-networking-api to v0.2.1-020250318085121-e88f4ed9f50a (main)

panslava:
/assign @gauravkghildiyal 

Idk how you prefer, i can squash into single commit, so we have single "buildable and test-passing" commit

Otherwise, now it's better to see what I did, but first 2 commits are not passing tests

This update is required to add CRD informer factories for MultiProject informers


gauravkghildiyal:
Thanks

/lgtm

PR #2860, opened 2025-05-01
fix: NEG Readiness gate for MSC is no longer prematurely marked ready

gauravkghildiyal:
Previously, when we didn't create NEGs for the additinoal subnets, we had the logic in place to make sure that the pods which exist in nodes from the additional subnet are always marked ready for the NEG readiness gate. This was done since the pod will never get added to a NEG and hence won't have it's readiness gate ready in a natural way.

But now that we DO have NEGs in additional subnets, we no longer need this explicit handling.

/assign @swetharepakula

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -206,9 +206,7 @@ func NewController(
 			podInformer.GetIndexer(),
 			cloud,
 			manager,
-			zoneGetter,
 			enableDualStackNEG,
-			flags.F.EnableMultiSubnetCluster,
 			logger,
 		)
 	} else {
--- a/pkg/neg/readiness/poller_test.go
+++ b/pkg/neg/readiness/poller_test.go
@@ -68,7 +68,7 @@ func (p *testPatcher) Eval(t *testing.T, pod string, negKey, bsKey *meta.Key) {
 }
 
 func newFakePoller() (*poller, error) {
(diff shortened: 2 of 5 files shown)


gauravkghildiyal:
> One question, do the existing tests cover the additional subnet cases? Please remove the hold if that is the case.

So all of the readiness reflector code is agnostic of all things related to subnet and does not have any MSC specific tests.

Yes the invocation of the Readiness reflector is different for MSC. Essentially, it's like the invoker of the readiness package tells it to watch for a {NEG, Pod} pair, and it is the invokers responsibility to plug in the correct NEG corresponding to the Pod (depending on the Node where the Pod lies, the NEG passed could either be from the default or non-default subnet) I see that this part has it's own tests in the form of: https://github.com/kubernetes/ingress-gce/blob/8b365e30fb1d64e79118016c0512331ab8b04851/pkg/neg/syncers/transaction_test.go#L1376

Thanks to your suggestion, I've also taken this opportunity to do a more thorough cleanup of any unused variables and code., essentially ending up doing a revert of most of the changes from this specific commit: https://github.com/kubernetes/ingress-gce/pull/2560/commits/f9895f1beb92fc81ad81915565e7dbc7e3add38c

swetharepakula (commented):
/lgtm
/approve

gauravkghildiyal:
Thanks!

/unhold

PR #2862, opened 2025-05-05
Ensure all nodes are considered default subnet for legacy networks

swetharepakula:
In legacy GCE networks, no subnets exist and therefore the controller & zone getter are initialized with an empty defaultSubnetURL. If the defaultSubnetURL is empty, we should assume that the cluster is running on a legacy network and always return true when checking if the node is in the default subnet.

/cc @sypakine
/assign @gauravkghildiyal
/assign @mmamczur

### DIFF
--- a/pkg/utils/zonegetter/zone_getter.go
+++ b/pkg/utils/zonegetter/zone_getter.go
@@ -354,8 +354,17 @@ func (z *ZoneGetter) IsDefaultSubnetNode(nodeName string, logger klog.Logger) (b
 // For any new nodes created after multi-subnet cluster is enabled, they are
 // guaranteed to have the subnet label if PodCIDR is populated. For any
 // existing nodes, they will not have label and can only be in the default
-// subnet.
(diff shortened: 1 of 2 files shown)


pkg/utils/zonegetter/zone_getter.go
```diff
@@ -354,8 +354,16 @@ func (z *ZoneGetter) IsDefaultSubnetNode(nodeName string, logger klog.Logger) (b
 // For any new nodes created after multi-subnet cluster is enabled, they are
 // guaranteed to have the subnet label if PodCIDR is populated. For any
 // existing nodes, they will not have label and can only be in the default
-// subnet.
+// subnet. If defaultSubnetURL is empty, then
```
mmamczur, line 357: ... it's a legacy GCE network?

gauravkghildiyal:
I'm guessing we can close this out?

swetharepakula:
Closing this pr in favor https://github.com/kubernetes/ingress-gce/pull/2865

PR #2863, opened 2025-05-05
Fix race when changing between NetLB and ILB

TortillaZHawaii:
When we change from NetLB to ILB or vice versa independent loops will try to update resources at the same time. This commit adds an additional check to wait until the other controller removes its finalizer from the service.

It also moves some helper functions from pkg/utils to pkg/annotations to prevent import cycle.

### DIFF
--- a/pkg/annotations/service.go
+++ b/pkg/annotations/service.go
@@ -24,6 +24,8 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/cloud-provider-gcp/providers/gce"
+	"k8s.io/ingress-gce/pkg/utils/common"
+	"k8s.io/ingress-gce/pkg/utils/slice"
 )
 
 const (
@@ -256,10 +258,17 @@ func WantsL4ILB(service *v1.Service) (bool, string) {
 	if service.Spec.Type != v1.ServiceTypeLoadBalancer {
 		return false, fmt.Sprintf("Type : %s", service.Spec.Type)
 	}
+
+	ltype := GetLoadBalancerAnnotationType(service)
+	// Prevents race condition when switching from NetLB to ILB
+	if IsRBSL4NetLB(service) {
+		return false, fmt.Sprintf("Type : %s, LBType : %s, HasNetLBResources : %t", service.Spec.Type, ltype, true)
+	}
+
 	if service.Spec.LoadBalancerClass != nil {
 		return HasLoadBalancerClass(service, RegionalInternalLoadBalancerClass), fmt.Sprintf("Type : %s", service.Spec.Type)
 	}
-	ltype := GetLoadBalancerAnnotationType(service)
+
 	if ltype == LBTypeInternal {
 		return true, fmt.Sprintf("Type : %s, LBType : %s", service.Spec.Type, ltype)
 	}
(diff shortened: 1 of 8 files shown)


pkg/annotations/service.go
```diff
@@ -492,3 +507,39 @@ func (svc *Service) GetInternalLoadBalancerAnnotationSubnet() string {
 	}
 	return ""
 }
+
+// IsLegacyL4ILBService returns true if the given LoadBalancer service is managed by service controller.
```
mmamczur, line 511: let's put these in some other file like maybe finalizers

having this in annotations pkg may suggest that these function looks at annotations

pkg/annotations/service.go
```diff
@@ -256,10 +258,17 @@ func WantsL4ILB(service *v1.Service) (bool, string) {
...
+
+	ltype := GetLoadBalancerAnnotationType(service)
+	// Prevents race condition when switching from NetLB to ILB
+	if IsRBSL4NetLB(service) {
+		return false, fmt.Sprintf("Type : %s, LBType : %s, HasNetLBResources : %t", service.Spec.Type, ltype, true)
+	}
+
 	if service.Spec.LoadBalancerClass != nil {
 		return HasLoadBalancerClass(service, RegionalInternalLoadBalancerClass), fmt.Sprintf("Type : %s", service.Spec.Type)
 	}
```
FelipeYepez, line 270: Lets keep this check for loadBalancerClass on top of what you added since it has precedence and will never conflict with NetLB -> ILB

pkg/annotations/service.go
```diff
@@ -274,10 +283,16 @@ func WantsL4NetLB(service *v1.Service) (bool, string) {
...
+	if IsSubsettingL4ILBService(service) {
+		return false, fmt.Sprintf("Type : %s, LBType : %s, HasILBResources : %t", service.Spec.Type, ltype, true)
+	}
+
 	if service.Spec.LoadBalancerClass != nil {
 		return HasLoadBalancerClass(service, RegionalExternalLoadBalancerClass), fmt.Sprintf("Type : %s", service.Spec.Type)
 	}
-	ltype := GetLoadBalancerAnnotationType(service)
 	return ltype != LBTypeInternal, fmt.Sprintf("Type : %s, LBType : %s", service.Spec.Type, ltype)
 }
```
FelipeYepez, line 297: Same here

FelipeYepez (changes requested):

pkg/annotations/service.go
TortillaZHawaii: As noted in the comment below by @mmamczur, the naming is a bit confusing. `IsRBSL4NetLB` actually verifies that the LB has the NetLBs finalizers, and if they do they shouldn't be put to the ILB work queue. Once the NetLB controller removes the finalizer, we can check if this really should be a ILB.

> will never conflict with NetLB -> ILB

[The k8s docs specify that the `.spec.loadBalancerClass` field _Once set, it cannot be changed_.](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class) The user can have an NetLB with annotation, then move to using ILB by setting the class (once) and we will have similar race condition between NetLB and ILB. Once the finalizer is removed, the class will in fact take precedence over annotation.

pkg/annotations/service.go
TortillaZHawaii: See https://github.com/kubernetes/ingress-gce/pull/2863#discussion_r2075182010

pkg/annotations/service.go
FelipeYepez: If a service is created without the loadBalancerClass `nil` there is no way to set it in an update so you don't have to worry about it.

TortillaZHawaii: got it, thanks!

pkg/annotations/service_test.go
```diff
@@ -21,10 +21,104 @@ import (
...
+						gce.ServiceAnnotationLoadBalancerType: string(gce.LBTypeInternal),
+					},
+				},
+			},
+			want: wants{},
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+
```
TortillaZHawaii, line 117: todo: fix

TortillaZHawaii:
/close 

the work continued on PR https://github.com/kubernetes/ingress-gce/pull/2866

PR #2864, opened 2025-05-06
Add annotations for enabling logging for L4 ILB/NetLB

anyush:
Allow to enable logging for L4 ILB/NetLB backends and provide sample rate for logging using service's annotations.


pkg/loadbalancers/l4_test.go
```diff
@@ -3042,3 +3042,140 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, name, tag string)
...
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+
+			svc := test.NewL4ILBService(false, 8080)
+			if tc.includeLoggingAnnotation {
+				if tc.enableLogging {
+					svc.Annotations[annotations.L4LBLoggingAnnotationKey] = annotations.L4LBLoggingEnabled
+				} else {
+					svc.Annotations[annotations.L4LBLoggingAnnotationKey] = ""
+				}
```
TortillaZHawaii, line 3120:
```suggestion
				svc.Annotations[annotations.L4LBLoggingAnnotationKey] = ""
				if tc.enableLogging {
					svc.Annotations[annotations.L4LBLoggingAnnotationKey] = annotations.L4LBLoggingEnabled
				}
```

pkg/loadbalancers/l4_test.go
```diff
@@ -3042,3 +3042,140 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, name, tag string)
...
+			}
+
+			result := l4.EnsureInternalLoadBalancer(nodeNames, svc)
+			if result.Error != nil {
+				if !tc.errorExpected {
+					t.Errorf("Failed to ensure internal loadBalancer, err %v", result.Error)
+				}
+				return
+			}
+			if result.Error == nil && tc.errorExpected {
```
TortillaZHawaii, line 3155:
```suggestion
			if tc.errorExpected {
```
we return if the error was not nil above, so it has to be nil here

pkg/loadbalancers/l4netlb.go
```diff
@@ -368,6 +368,17 @@ func (l4netlb *L4NetLB) provideBackendService(syncResult *L4NetLBSyncResult, hcL
...
+	sampleRate, err := annotations.GetL4LBLoggingSampleRate(l4netlb.Service)
+	if err != nil {
+		syncResult.GCEResourceInError = annotations.BackendServiceResource
+		syncResult.Error = utils.NewUserError(err)
+		return ""
+	}
+	if annotations.HasL4LBLoggingAnnotation(l4netlb.Service) {
+		backendParams.EnableLogging = true
+		backendParams.LoggingSampleRate = sampleRate
+	}
```
TortillaZHawaii, line 380: as mentioned in l4.go it looks like we could somewhat benefit for encapsulating that logic in a separate func, and only update syncResult struct (specific to netlb) here

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1744,3 +1744,119 @@ func assertAddressOldReservedHook(t *testing.T, gceCloud *gce.Cloud) func(ctx co
...
+			l4NetLB := mustSetupNetLBTestHandler(t, svc, nodeNames)
+
+			result := l4NetLB.EnsureFrontend(nodeNames, svc)
+			if result.Error != nil {
+				if !tc.errorExpected {
+					t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+				}
+				return
+			}
+			if result.Error == nil && tc.errorExpected {
```
TortillaZHawaii, line 1836:
```suggestion
			if tc.errorExpected {
```

pkg/loadbalancers/l4netlb.go
```diff
@@ -368,6 +368,17 @@ func (l4netlb *L4NetLB) provideBackendService(syncResult *L4NetLBSyncResult, hcL
...
 	}
 
+	sampleRate, err := annotations.GetL4LBLoggingSampleRate(l4netlb.Service)
+	if err != nil {
+		syncResult.GCEResourceInError = annotations.BackendServiceResource
+		syncResult.Error = utils.NewUserError(err)
+		return ""
+	}
+	if annotations.HasL4LBLoggingAnnotation(l4netlb.Service) {
+		backendParams.EnableLogging = true
```
TortillaZHawaii, line 378: Does changing that fields trigger recreate of the ILB/NetLB? Or is it patchable?

TortillaZHawaii (changes requested):

pkg/loadbalancers/l4.go
```diff
@@ -553,6 +553,18 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+	loggingSampleRate, err := annotations.GetL4LBLoggingSampleRate(l4.Service)
+	if err != nil {
+		result.GCEResourceInError = annotations.BackendServiceResource
+		result.Error = utils.NewUserError(err)
+		return result
+	}
+	if annotations.HasL4LBLoggingAnnotation(l4.Service) {
+		backendParams.EnableLogging = true
+		backendParams.LoggingSampleRate = loggingSampleRate
+	}
```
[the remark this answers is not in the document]
anyush: Moved.

I originally thought the same way, but wasn't sure if it's good to ignore invalid value without any notification to user. On the other side, this field really doesn't have any sense if logging is disabled, so fixed

pkg/loadbalancers/l4netlb.go
anyush: As I see it's patched, not recreated

pkg/backends/backends.go
```diff
@@ -516,6 +519,21 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
 	return svcsEqual
 }
 
+func backendServiceLogConfigEqual(oldLC, newLC *composite.BackendServiceLogConfig) bool {
+	if oldLC == nil && newLC == nil {
+		return true
+	}
+	if oldLC == nil && newLC != nil || oldLC != nil && newLC == nil {
+		return false
+	}
```
TortillaZHawaii, line 528: based on the GetL4LBLoggingConfig we know that when we call backendServiceLogConfigEqual, the newLC won't be nil. My question is when can oldLC be nil. Is it possible for the API call for it to be nil, and if so how is it different to not being nil and having zero values inside?

If the API returns oldLC := nil, and newLC has default zero values `backendServiceLogConfigEqual` will return false, causing BS update, even though we didn't explicitly specified that we want logging or even a change in a BS. I think it's safer to create `oldLC := &composite.BackendServiceLogConfig{}` if its nil, and then compare individual fields against `newLC`, and skip the nil checks alltogether.

TortillaZHawaii (changes requested):

pkg/annotations/service.go
```diff
@@ -141,6 +142,12 @@ const (
 	WeightedL4AnnotationKey = "networking.gke.io/weighted-load-balancing"
 	// Service annotation value for using pods-per-node Weighted load balancing in both ILB and NetlB
 	WeightedL4AnnotationPodsPerNode = "pods-per-node"
+
+	// Service annotation key for enabling logging for both ILB and NetLB backends
```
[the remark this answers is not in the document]
TortillaZHawaii: Close those or replace Service annotation key with actual key name
```suggestion
	// L4LBLoggingAnnotationKey for enabling logging for both ILB and NetLB backends
```

pkg/annotations/service.go
```diff
@@ -141,6 +142,12 @@ const (
...
+	L4LBLoggingAnnotationKey           = "networking.gke.io/l4lb-logging"
+	L4LBLoggingEnabled                 = "enabled"
+	// Service annotation key for specifying sample rate for both ILB and NetLB backends
```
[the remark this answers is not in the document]
TortillaZHawaii:
```suggestion
	// L4LBLoggingSampleRateAnnotationKey for specifying sample rate for both ILB and NetLB backends
```

pkg/backends/backends.go
```diff
@@ -516,6 +519,21 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
...
+	return oldLC.Enable == newLC.Enable &&
+		oldLC.SampleRate == newLC.SampleRate &&
+		oldLC.OptionalMode == newLC.OptionalMode &&
+		utils.EqualStringSets(oldLC.OptionalFields, newLC.OptionalFields) &&
+		utils.EqualStringSets(oldLC.ForceSendFields, newLC.ForceSendFields) &&
+		utils.EqualStringSets(oldLC.NullFields, newLC.NullFields)
```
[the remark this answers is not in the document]
anyush: I think it might cause bugs in future (e.g. if later it would be possible to specify optional field), therefore it would be better to leave them. However there rather won't be problems with generic fields - removing.

Changes of mocks were caused by using explicit disabled config if no provided, not related anymore.

pkg/backends/backends.go
anyush: Reworked a bit, now if any of configs is nil, it's treated as `&composite.BackendServiceLogConfig{}`

pkg/annotations/service.go
```diff
@@ -326,6 +334,55 @@ func HasLoadBalancerClass(service *v1.Service, key string) bool {
 	return false
 }
 
+// HasL4LBLoggingAnnotation checks if given service has the L4 ILB/NetLB logging annotation
+func HasL4LBLoggingAnnotation(service *v1.Service) bool {
```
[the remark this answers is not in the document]
TortillaZHawaii: I don't like it but okay

pkg/backends/backends.go
```diff
@@ -493,12 +495,22 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
...
 // The list of backends is not checked, since that is handled by the neg-linker.
 func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTracking bool) bool {
+	oldLC := &composite.BackendServiceLogConfig{}
+	if oldBS.LogConfig != nil {
+		oldLC = oldBS.LogConfig
+	}
+	newLC := &composite.BackendServiceLogConfig{}
+	if newBS.LogConfig != nil {
+		newLC = newBS.LogConfig
+	}
```
TortillaZHawaii, line 505: nit: I think this would fit better in `backendServiceLogConfigEqual`, as both `oldLC` and `newLC` aren't passed anywhere else in backendSvcEqual

So backendServiceLogConfigEqual would have a:
```go
func backendServiceLogConfigEqual(old, new *composite.BackendServiceLogConfig) bool {
     oldLC, newLC := &composite.BackendServiceLogConfig{}, &composite.BackendServiceLogConfig{}
     if old != nil {
        oldLC = old
    }
    if new != nil {
        newLC = new
    }
    

	return oldLC.Enable == newLC.Enable &&
		oldLC.SampleRate == newLC.SampleRate &&
		oldLC.OptionalMode == newLC.OptionalMode &&
		utils.EqualStringSets(oldLC.OptionalFields, newLC.OptionalFields)
}
```

pkg/loadbalancers/l4_test.go
```diff
@@ -3042,3 +3042,139 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, name, tag string)
...
+
+func TestILBLogging(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		desc                        string
+		includeLoggingAnnotation    bool
+		enableLogging               bool
+		includeSampleRateAnnotation bool
+		sampleRate                  string
```
TortillaZHawaii, line 3054: I think it would be clearer to the reader to have annotations map[string]string instead of those four fields

pkg/loadbalancers/l4_test.go
```diff
@@ -3042,3 +3042,139 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, name, tag string)
...
+		logConfigExpected           bool
+		expectedSampleRate          float64
+		errorExpected               bool
```
TortillaZHawaii, line 3057: Golang prefers `want` to `expected`. Either way why is one a prefix and the other two a suffix?

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1744,3 +1744,118 @@ func assertAddressOldReservedHook(t *testing.T, gceCloud *gce.Cloud) func(ctx co
...
+
+	testCases := []struct {
+		desc                        string
+		includeLoggingAnnotation    bool
+		enableLogging               bool
+		includeSampleRateAnnotation bool
+		sampleRate                  string
+		logConfigExpected           bool
+		expectedSampleRate          float64
+		errorExpected               bool
```
TortillaZHawaii, line 1759: same notes as in comments for ILB tests

TortillaZHawaii (changes requested):

pkg/loadbalancers/l4_test.go
anyush: It stands for "logConfig expected to be set", "expected value of the sample rate" and "error expected to be raised".

Will replace by "wantLogConfig", "wantSampleRate" and "wantError"

pkg/loadbalancers/l4_test.go
anyush: Done

pkg/backends/backends.go
```diff
@@ -516,6 +519,20 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
...
+		oldLC = &composite.BackendServiceLogConfig{}
+	}
+	if newLC == nil {
+		newLC = &composite.BackendServiceLogConfig{}
+	}
+
+	return oldLC.Enable == newLC.Enable &&
+		oldLC.SampleRate == newLC.SampleRate &&
+		oldLC.OptionalMode == newLC.OptionalMode &&
+		utils.EqualStringSets(oldLC.OptionalFields, newLC.OptionalFields)
```
mmamczur, line 539: I'm not sure if we should compare optional fields. If they really really want to set them manually maybe let them do it and let's not overwrite

pkg/context/context.go
```diff
@@ -312,6 +313,7 @@ func (ctx *ControllerContext) Start(stopCh <-chan struct{}) {
 	go ctx.PodInformer.Run(stopCh)
 	go ctx.NodeInformer.Run(stopCh)
 	go ctx.EndpointSliceInformer.Run(stopCh)
+	go ctx.ConfigMapInformer.Run(stopCh)
```
TortillaZHawaii: Shouldn't this be flag gated?

pkg/l4lb/l4controller.go
```diff
@@ -178,9 +179,37 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}, lo
...
+			configMap := obj.(*v1.ConfigMap)
+			l4c.enqueueServicesReferencingConfigMap(configMap)
+		},
+	})
+
 	return l4c
 }
 
+func (l4c *L4Controller) enqueueServicesReferencingConfigMap(configMap *v1.ConfigMap) {
+	services := operator.Services(l4c.ctx.Services().List(), l4c.logger).ReferencesL4LoggingConfigMap(*configMap).AsList()
```
TortillaZHawaii: Why `ReferencesL4LoggingConfigMap(*configMap)` dereferences configMap?

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -173,9 +174,37 @@ func NewL4NetLBController(
 		},
 	})
 
+	ctx.ConfigMapInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			configMap := obj.(*v1.ConfigMap)
```
TortillaZHawaii: I assume this AddEventHandler was created before Go generics. Can we guarantee that this will always be (*v1.ConfigMap)? Please add a check that this was actually ok, otherwise we might get a panic: https://go.dev/ref/spec#Type_assertions

Same below and above

pkg/common/operator/service.go
```diff
@@ -46,6 +47,23 @@ func (op *ServicesOperator) ReferencesBackendConfig(beConfig *backendconfigv1.Ba
 	return Services(s, op.logger)
 }
 
+// ReferencesL4LoggingConfigMap returns the Services that reference the given ConfigMap.
+func (op *ServicesOperator) ReferencesL4LoggingConfigMap(configMap api_v1.ConfigMap) *ServicesOperator {
```
TortillaZHawaii: can we test this?

pkg/backends/backends.go
```diff
@@ -525,6 +531,22 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
 	return svcsEqual
 }
 
+// backendServiceLogConfigEqual returns true if both elements are equal
+// and return false if at least one parameter is different
+func backendServiceLogConfigEqual(oldLC, newLC *composite.BackendServiceLogConfig) bool {
```
TortillaZHawaii, line 528: I can't find tests for this

pkg/annotations/service.go
```diff
@@ -471,6 +474,16 @@ func (svc *Service) getBackendConfigAnnotation() (string, bool) {
 	return "", false
 }
 
+// GetL4ConfigMapAnnotation returns name of the config map which contains logging config.
+// Returns false if annotation with the name is not specified.
+func (svc *Service) GetL4ConfigMapAnnotation() (string, bool) {
```
TortillaZHawaii: Please add tests

TortillaZHawaii (changes requested):
/unapprove

Tests are missing

pkg/annotations/service.go
anyush: It only gets value of annotation (unlike getBackendConfigAnnotation, which unmarshals value later), so I don't think that we should explicitly cover it (so as e.g. GetExternalLoadBalancerAnnotationSubnet)

pkg/backends/backends.go
anyush: Added

pkg/context/context.go
anyush: Gated

pkg/l4lb/l4controller.go
anyush: Fixed

pkg/l4lb/l4netlbcontroller.go
anyush: ConfigMapInformer shouldn't return anything except ConfigMap objects, but added for double check

pkg/common/operator/service.go
anyush: Like ReferencesBackendConfig - I don't think it can be covered by unit testing

pkg/backends/backends_test.go
```diff
@@ -349,6 +349,7 @@ func TestBackendSvcEqual(t *testing.T) {
 		newBackendService         *composite.BackendService
 		compareConnectionTracking bool
 		withZonalAffinityEnabled  bool
+		withL4LBManagementEnabled bool
```
TortillaZHawaii:
```suggestion
		withLoggingManagementEnabled bool
```

pkg/backends/backends_test.go
```diff
@@ -546,6 +547,81 @@ func TestBackendSvcEqual(t *testing.T) {
 			withZonalAffinityEnabled: true,
 			wantEqual:                false,
 		},
+		{
```
TortillaZHawaii, line 550: Please add tests with nil values that would represent enabling/disabling logging on the Backend Service

pkg/loadbalancers/l4_logging.go
```diff
@@ -0,0 +1,166 @@
...
+	availableOptions := []string{
+		OptionalModeExcludeAll,
+		OptionalModeIncludeAll,
+		OptionalModeCustom,
+	}
+	for _, option := range availableOptions {
+		if val == option {
+			return val, nil
+		}
+	}
```
TortillaZHawaii: nit, no change needed as this is small slice,
but this is a textbook set usage

pkg/loadbalancers/l4_logging.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return "", fmt.Errorf("invalid optional mode: \"%s\", available options are: %v", val, availableOptions)
```
TortillaZHawaii: nit: you can use %q instead of escaping quotes and adding %s:
https://pkg.go.dev/fmt#:~:text=%25q%09a%20double%2Dquoted%20string%20safely%20escaped%20with%20Go%20syntax

pkg/loadbalancers/l4_logging.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return logConfig, nil
+}
+
+// ParseL4LoggingConfigMap parses logging configs from ConfigMap. Extra fields ignored.
+// If any of fields doesn't pass validation - error returned.
+func ParseL4LoggingConfigMap(configMap *corev1.ConfigMap) (*composite.BackendServiceLogConfig, error) {
```
TortillaZHawaii, line 74: tests

pkg/loadbalancers/l4_logging.go
```diff
@@ -0,0 +1,166 @@
...
+	OptionalModeKey        = "optional-mode"
+	OptionalModeExcludeAll = "EXCLUDE_ALL_OPTIONAL"
+	OptionalModeIncludeAll = "INCLUDE_ALL_OPTIONAL"
+	OptionalModeCustom     = "CUSTOM"
+	OptionalFieldsKey      = "optional-fields"
+)
+
+// GetL4LoggingConfig retrieves config map responsible for logging config for service if contains annotation.
+// If annotation not specified or config map doesn't exist returns nil.
+func GetL4LoggingConfig(service *corev1.Service, configMapLister cache.Store) (*composite.BackendServiceLogConfig, error) {
```
TortillaZHawaii, line 45: tests

TortillaZHawaii (changes requested):

pkg/backends/backends_test.go
anyush: Changed to `withL4LoggingManagementEnabled`

pkg/loadbalancers/l4_logging.go
anyush: Changed to map

pkg/loadbalancers/l4_logging.go
anyush: Added

pkg/loadbalancers/l4_logging.go
anyush: Added

pkg/loadbalancers/l4_logging_test.go
```diff
@@ -0,0 +1,342 @@
...
+				Data: map[string]string{
+					SampleRateKey:     "1",
+					OptionalModeKey:   OptionalModeCustom,
+					OptionalFieldsKey: "field1",
+				},
+			},
+			wantConfig: nil,
+		},
+		{
+			desc: "Excplicitly disabled logging",
```
TortillaZHawaii: typo
```suggestion
			desc: "Explicitly disabled logging",
```

pkg/loadbalancers/l4_logging_test.go
```diff
@@ -0,0 +1,342 @@
...
+		},
+		{
+			desc: "Optional mode - invalid",
+			configMap: &v1.ConfigMap{
+				Data: map[string]string{
+					LoggingEnableKey: LoggingEnabledValue,
+					OptionalModeKey:  "invalid",
+				},
+			},
+			wantError: "invalid optional mode: \"invalid\", available options are: \"EXCLUDE_ALL_OPTIONAL\", \"INCLUDE_ALL_OPTIONAL\", \"CUSTOM\"",
```
TortillaZHawaii: no change needed, but in the future you can use `` so you don't have to escape " later.

pkg/loadbalancers/l4_logging_test.go
```diff
@@ -0,0 +1,342 @@
...
+				OptionalFields: []string{"field1", "field2"},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			loggingConfig, err := ParseL4LoggingConfigMap(tc.configMap)
+
+			if err != nil {
```
TortillaZHawaii: We skip `wantError` when the error is `nil`. `!reflect.DeepEqual` also will skip this case because `tc.wantConfig` is `nil` and `ParseL4LoggingConfigMap` returns `nil, err`. This can cause false positives.

Please fix.

pkg/loadbalancers/l4_logging_test.go
anyush: fixed

pkg/loadbalancers/l4_logging_test.go
anyush: Fixed

TortillaZHawaii:
/lgtm

Feel free to `unhold` if you don't need other reviewers. Tide will merge the PR with just one lgtm
/hold

pkg/backends/backends.go
mmamczur, line 528: when you create a Backend Service with no LogConfig and then retrieve it, do we know what the value for BackendServiceLogsConfig will be? will it be nil?

Waht will happen if you configure logging and then disable it, will it be empty or nil or something else?

just asking to be sure that we don't trigger unnecessary Backend Service updates

anyush: If logging disabled / was never enabled - BackendServiceLogConfig will be nil

pkg/loadbalancers/l4.go
```diff
@@ -386,7 +388,7 @@ func (l4 *L4) subnetName() string {
 
 // EnsureInternalLoadBalancer ensures that all GCE resources for the given loadbalancer service have
 // been created. It returns a LoadBalancerStatus with the updated ForwardingRule IP address.
-func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service) *L4ILBSyncResult {
+func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service, configMapLister cache.Store) *L4ILBSyncResult {
```
mmamczur: I'd consider making this a field in l4 (and the same for netlb)

anyush: Done

PR #2866, opened 2025-05-09
Improve L4 NEG controller handoff, finalizer logic and NetLB <-> ILB controllers race.

FelipeYepez:
This PR refactors the L4 NEG processing logic within the NEG controller to ensure smoother transitions and proper resource management, particularly when services migrate between NetLB and ILB, or during their creation/deletion.

When a Service is configured for L4 ILB (requiring `L4ILBFinalizerV2`) or L4 NetLB on the default network (requiring `L4NetLBFinalizerV3`), the NEG controller will check for the presence of these finalizers.

If a required finalizer is missing and the Service is *not* being deleted, the controller will requeue the Service. This allows the respective L4 controller (ILB or NetLB) time to add its finalizer and take ownership, preventing the NEG controller from acting prematurely.

If a required finalizer is missing but the Service *is* being deleted (`DeletionTimestamp` is set), the NEG controller will assume the L4 resources are being torn down and will not wait for the finalizer. This allows NEG cleanup to proceed correctly.

When migrating a service from NetLB to ILB (or vice versa), the L4 ILB/NetLB controllers will effectively wait for the previous controller to remove its finalizer before processing the service.


pkg/annotations/gce_annotations.go
```diff
@@ -18,6 +18,7 @@ package annotations
 
 import (
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/utils/common"
```
bowei, line 21: Can we put the refactor changes in a separate commit (includes tests)

fivemeyestore:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -185,9 +185,9 @@ func (lc *L4NetLBController) needsAddition(newSvc, oldSvc *v1.Service) bool {
 // needsDeletion return true if svc required deleting RBS based NetLB
 func (lc *L4NetLBController) needsDeletion(svc *v1.Service, svcLogger klog.Logger) bool {
 	// Check if service was provisioned by RBS controller before -- if it has rbs finalizer, rbs loadBalancerClass or rbs forwarding rule
-	if !utils.HasL4NetLBFinalizerV2(svc) && !utils.HasL4NetLBFinalizerV3(svc) &&
+	if !common.HasL4NetLBFinalizerV2(svc) && !common.HasL4NetLBFinalizerV3(svc) &&
 		!lc.hasRBSForwardingRule(svc, svcLogger) &&
-		!annotations.HasLoadBalancerClass(svc, annotations.RegionalExternalLoadBalancerClass) {
+		!common.HasLoadBalancerClass(svc, common.RegionalExternalLoadBalancerClass) {
```
TortillaZHawaii, line 190: nit: we can remove that comment by using a helper variable
```go
wasProvisionedByRBS := common.HasL4NetLBFinalizerV2(svc) || common.HasL4NetLBFinalizerV3(svc) ||
    lc.hasRBSForwardingRule(svc, svcLogger) || common.HasLoadBalancerClass(svc, common.RegionalExternalLoadBalancerClass)
    
if !wasProvisionedByRBS {
    return false
}
```

pkg/neg/controller.go
```diff
@@ -672,61 +673,99 @@ func (c *Controller) mergeStandaloneNEGsPortInfo(service *apiv1.Service, name ty
...
+				return fmt.Errorf(msg)
+			}
+		} else {
+			// Only process ILB services after L4 controller has marked it with v2 finalizer.
+			effectiveNeedsILB = true
+		}
+	}
+
+	effectiveNeedsNetLB := false
+	if wantsNetLB {
```
TortillaZHawaii, line 717: In ILB we have `c.runL4ForILB`

why isn't there a symetric flag for NetLB?

pkg/neg/controller_test.go
```diff
@@ -381,7 +381,9 @@ func TestEnableNEGServiceWithL4ILB(t *testing.T) {
 	defer controller.stop()
 	var prevSyncerKey, updatedSyncerKey negtypes.NegSyncerKey
 	t.Logf("Creating L4 ILB service with ExternalTrafficPolicy:Cluster")
-	controller.serviceLister.Add(newTestILBService(controller, false, 80))
+	svcILB := newTestILBService(controller, false, 80)
+	svcILB.Finalizers = []string{gce.ILBFinalizerV1}
```
TortillaZHawaii, line 385: does depending on the ILBFinalizerV1 being present would mean that this cannot be potentially backported to the versions with CCM that doesn't add ILBFinalizerV1?

pkg/neg/controller.go
```diff
@@ -672,61 +673,99 @@ func (c *Controller) mergeStandaloneNEGsPortInfo(service *apiv1.Service, name ty
...
+							return fmt.Errorf(msg)
+						}
+					} else {
+						// Only process NetLB services after L4NetLB controller has marked it with v3 finalizer.
+						effectiveNeedsNetLB = true
+					}
+				}
+			}
+		}
+	}
```
TortillaZHawaii, line 752: Please make it easier to read, there is far too many nesting levels. Same goes for ILB.

[go/tott/733](https://goto.google.com/tott/733)

We would also want to have this as a seperate func really - as it is around 40 lines long each. It could return (bool, error) tuple.

pkg/l4lb/l4netlbcontroller.go
TortillaZHawaii: plus we have `common.IsRBSL4NetLB` helper func

pkg/neg/controller_test.go
```diff
@@ -1751,6 +1753,97 @@ func TestEnableNEGServiceWithL4NetLB(t *testing.T) {
...
+		},
+		{
+			desc:             "NetLB service",
+			finalizers:       []string{common.NetLBFinalizerV3},
+			wantL4ExternalLB: true,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			kubeClient := fake.NewSimpleClientset()
```
TortillaZHawaii, line 1778:
```suggestion
	for _, tc := range testCases {
	    tc := tc
		t.Run(tc.desc, func(t *testing.T) {
		    t.Parallel()
			kubeClient := fake.NewSimpleClientset()
```

TortillaZHawaii (changes requested):

pkg/neg/controller.go
FelipeYepez: For NetLB we have `c.runL4ForNetLB` in line 727. We check for this after checking the network info since Multinet services also use NEGs even if this field is false.

pkg/neg/controller_test.go
FelipeYepez: This can be backported since the first check done in the neg controller after verifying it is a NetLB service is checking if it has the required annotations like RBS or loadBalancerClass. What would happen in the worst case is the neg controller requeuing the service if it has the RBS annotation with no finalizer

TortillaZHawaii: got it, thanks!

pkg/l4lb/l4netlbcontroller.go
FelipeYepez: Done. Thanks for pointing this out

pkg/neg/controller.go
```diff
@@ -672,61 +673,126 @@ func (c *Controller) mergeStandaloneNEGsPortInfo(service *apiv1.Service, name ty
...
+			// Return an error to trigger RateLimited requeue via handleErr in case NetLB -> ILB migration.
+			msg := fmt.Sprintf("ILB service %s/%s might require NEGs but is missing the L4 ILB v2 finalizer. Will retry.", service.Namespace, service.Name)
+			c.logger.Info(msg)
+			c.recorder.Eventf(service, apiv1.EventTypeNormal, "WaitingForL4Finalizer", msg)
+			return false, fmt.Errorf(msg)
+		}
+	}
+
+	// Only process ILB services after L4 controller has marked it with v2 finalizer.
+	return true, nil
```
TortillaZHawaii, line 746: If we reverse if statement we will have one less nesting level:
```go
if common.HasL4ILBFinalizerV2(service) {
    // Only process ILB services after L4 controller has marked it with v2 finalizer.
    return true, nil
}

if service.ObjectMeta.DeletionTimestamp != nil {
// rest of the code from 728-742
```

bowei:
This is going to age out if there is no activity

PR #2868, opened 2025-05-13
Clean up Neg controller code

avntyagi-google:
#cleanup Clean up Neg controller code

Removed unnecessary checks in the code for checking whether svcNegClient is null or not. It is initialized properly in the main function at ingress-gce/blob/master/cmd/glbc/main.go#L176, if initialization fails it causes the program to exit.

It is safe to assume in the other parts of code that it is fully initialized and value is not null.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -377,7 +377,10 @@ func main() {
 		logger.Info("Start running the enabled controllers",
 			"NEG controller", flags.F.EnableNEGController,
 		)
-		runNEGController(ctx, systemHealth, rOption, logger)
+		err := runNEGController(ctx, systemHealth, rOption, logger)
+		if err != nil {
+			klog.Fatalf("failed to run NEG controller: %s", err)
+		}
 	}
 	runIngress := func() {
 		logger := rootLogger.WithName("Other controllers")
@@ -461,7 +464,10 @@ func makeNEGRunnerWithLeaderElection(
 		leOption,
 		negLockName,
 		func(context.Context) {
-			runNEGController(ctx, systemHealth, runOption, logger)
+			err := runNEGController(ctx, systemHealth, runOption, logger)
+			if err != nil {
+				klog.Fatalf("failed to run NEG controller: %s", err)
+			}
 		},
 		func() {
 			logger.Info("Stop running NEG Leader election")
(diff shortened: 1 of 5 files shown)


panslava:
I thought about it for a second

Can we add validation in constructors of those structs, and return an error if svcNegClient is nil? 

It will require a bit more changes, but I think it is worth it, as with MultiProject we have second path of code that calls all of that, and it is good to fail fast then wait for error to happen

cmd/glbc/main.go
```diff
@@ -616,7 +622,10 @@ func runNEGController(ctx *ingctx.ControllerContext, systemHealth *systemhealth.
 	}
 
 	if flags.F.EnableNEGController {
-		negController := createNEGController(ctx, systemHealth, option.stopCh, logger)
+		negController, err := createNEGController(ctx, systemHealth, option.stopCh, logger)
+		if err != nil {
+			return fmt.Errorf("failed to create NEG controller: %s", err)
```
panslava, line 627: nit: can you use %w for error instead of %s when constructing using fmt.Errorf

cmd/glbc/main.go
```diff
@@ -701,9 +711,12 @@ func createNEGController(ctx *ingctx.ControllerContext, systemHealth *systemheal
 		stopCh,
 		logger,
 	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create NEG controller: %s", err)
```
panslava, line 715: nit: can you use %w for error instead of %s when constructing using fmt.Errorf

avntyagi-google: Done.

pkg/multiproject/neg/neg.go
```diff
@@ -308,6 +312,8 @@ func createNEGController(
 		stopCh,
 		logger,
 	)
-
-	return negController
+	if err != nil {
+		return nil, fmt.Errorf("failed to create NEG controller: %v", err)
```
panslava, line 316: %w

pkg/multiproject/neg/neg.go
```diff
@@ -105,6 +105,10 @@ func StartNEGController(
 		logger,
 	)
 
+	if err != nil {
+		return nil, fmt.Errorf("failed to create NEG controller: %v", err)
```
panslava, line 109: %w

panslava:
/lgtm

thank you!

PR #2869, opened 2025-05-13
Allow ability to override firewall source ranges

panslava:

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -144,6 +144,7 @@ var F = struct {
 	EnableL4NetLBForwardingRulesOptimizations bool
 	EnableIPV6OnlyNEG                         bool
 	MultiProjectOwnerLabelKey                 string
+	OverrideHealthCheckSourceCIDRs            string
 
 	// DEPRECATED: ASM ConfigMap based config is no longer used and will be removed in a future release.
 	EnableASMConfigMapBasedConfig    bool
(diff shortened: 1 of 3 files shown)


pkg/flags/flags.go
```diff
@@ -342,6 +343,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableL4NetLBForwardingRulesOptimizations, "enable-l4netlb-forwarding-rules-optimizations", false, "Enable optimized processing of forwarding rules for L4 NetLB.")
 	flag.BoolVar(&F.EnableIPV6OnlyNEG, "enable-ipv6-only-neg", false, "Enable support for IPV6 Only NEG's.")
 	flag.StringVar(&F.MultiProjectOwnerLabelKey, "multi-project-owner-label-key", "multiproject.gke.io/owner", "The label key for multi-project owner, which is used to identify the owner of objects in multi-project mode.")
+	flag.StringVar(&F.OverrideLoadBalancerSourceRanges, "override-load-balancer-source-ranges", "", "Override the default load balancer source ranges. These are used to set up firewalls")
```
gauravkghildiyal, line 346: nit: Lets specify format, and if we are doing that, then let's also use a more precise name?

>flag.StringVar(&F.OverrideHealthCheckSourceCIDRs, "override-health-check-src-cidrs", "", "Override the default source IP ranges used when configuring firewall rules to allow health check probes in L7 LBs. The source IP ranges are passed as a comma separated list of CIDRs")

gauravkghildiyal (commented):
Thanks for the wait. Looks good. Just one nit

pkg/flags/flags.go
panslava: That's great suggestion, I also think it is more clear like that. Changed

PR #2876, opened 2025-05-16
Check for FirewallXPNError in IsUserError

TortillaZHawaii:
Since utils package is dependend upon practicaly everywhere, it causes import cycles when trying to add additional `error.As` there. Since I don't want to grow this black hole pkg this commit moves IsUserError to pkg/loadbalancers as we actually use it there.

Also adds tests which were missing for that method.


TortillaZHawaii:
\assign @mmamczur

PR #2878, opened 2025-05-16
Fix some of the golangci-lint staticcheck errors

TortillaZHawaii:
Each commit fixes errors caused by specific error code


cmd/404-server-with-metrics/server-with-metrics.go
```diff
@@ -225,7 +225,9 @@ func (s *server) shutdownHandler() http.HandlerFunc {
 
 // notFoundHandler uses the default http NotFoundHandler which returns a 404 status code
 func (s *server) notFoundHandler() http.HandlerFunc {
-	rand.Seed(1)
+	src := rand.NewSource(1)
```
mmamczur, line 228: won't this now use the same val each time the function is called?

mmamczur: ok nvm this handler is called once and then the inline function below uses it, ignore this comment

PR #2879, opened 2025-05-19
Add new namer funcs for Forwarding Rule sets

TortillaZHawaii:
New naming convention suffixes "-{num}" to the end of Forwarding Rule name or "-{num}-ipv6" for IPv6 ones. We use base36 to limit the space used up by num.

Also refactors some of the existing TestL4Namer to simplify repetetive code. Personally I'm not a huge fan of having to check each field's length, as the `want`ed result tests that as well, but for sanity check we can leave it there.

### DIFF
--- a/pkg/utils/namer/l4_namer.go
+++ b/pkg/utils/namer/l4_namer.go
@@ -2,6 +2,7 @@ package namer
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


TortillaZHawaii:
\assign @mmamczur 
\lint

PR #2880, opened 2025-05-19
NetLB: workaround for the finalizer removal case

mmamczur:
When determining which backend to use for a NetLB service in addition to the finalizers also check the actual backend service backends.

This is to work around the situations where the customers remove finalizers and can cause an unexpected move to NEG backends.

### DIFF
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/ingress-gce/pkg/utils/zonegetter"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -550,7 +551,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
 		svcLogger.Info("Finished syncing L4 NetLB RBS service", "timeTaken", time.Since(startTime))
 	}()
 
-	usesNegBackends := lc.shouldUseNEGBackends(service)
+	usesNegBackends := lc.shouldUseNEGBackends(service, svcLogger)
 
 	l4NetLBParams := &loadbalancers.L4NetLBParams{
 		Service:                          service,
@@ -645,7 +646,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
 	return syncResult
 }
 
-func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
+func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service, svcLogger klog.Logger) bool {
 	if !lc.enableNEGSupport {
 		return false
 	}
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
 	if utils.HasL4NetLBFinalizerV3(service) {
 		return true
 	}
-	return lc.enableNEGAsDefault
+	if !lc.enableNEGAsDefault {
+		return false
+	}
+	// Do an extra check in case the customer removes the required finalizers
+	// Get the backend service if one exists and see what kind of backend
+	// is attached. If it's an IG keep using IGs.
+	backendType, err := lc.getBackendLinkType(service, svcLogger)
+	if err != nil {
+		svcLogger.Error(err, "Could not determine the backend type of the existing backend service, deferring to default")
+		return true
+	}
+	if backendType != nil && *backendType == instanceGroupLink {
+		return false
+	}
+	return backendType == nil || *backendType == negLink
+}
+
+func (lc *L4NetLBController) getBackendLinkType(service *v1.Service, svcLogger klog.Logger) (*backendLinkType, error) {
+	bsName := lc.namer.L4Backend(service.Namespace, service.Name)
+	backendService, err := lc.backendPool.Get(bsName, meta.VersionGA, meta.Regional, svcLogger)
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if backendService == nil {
+		return nil, nil
+	}
+	// When moving from subsetting ILB the BackendService could exist with
+	// a different load balancing scheme. Here we should treat this as
+	// a non-existent Backend Service.
+	if backendService.LoadBalancingScheme != string(cloud.SchemeExternal) {
+		return nil, nil
+	}
+	if len(backendService.Backends) > 0 {
+		firstBackend := backendService.Backends[0]
+		if strings.Contains(firstBackend.Group, "/instanceGroups/") {
+			return ptr.To(instanceGroupLink), nil
+		}
+		if strings.Contains(firstBackend.Group, "/networkEndpointGroups/") {
+			return ptr.To(negLink), nil
+		}
+	}
+	return nil, nil
 }
 
 func (lc *L4NetLBController) emitEnsuredDualStackEvent(service *v1.Service) {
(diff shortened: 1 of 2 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+		// is attached. If it's an IG keep using IGs.
+		backendType, err := lc.backendTypeOfTheBackendService(service, svcLogger)
+		if err != nil {
+			svcLogger.Error(err, "Could not determine the backend type of the existing backend service, deferring to default")
+			return true
+		}
+		if backendType != nil && *backendType == instanceGroupLink {
+			return false
+		}
+		return true
```
TortillaZHawaii, line 671: If in the future there would be something different then NEG or IG this statement will return false and we will try provision NEGs. 

```suggestion
		return backendType == nil || *backendType == negGroupLink
```

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+	}
+	return false
+}
+
+func (lc *L4NetLBController) backendTypeOfTheBackendService(service *v1.Service, svcLogger klog.Logger) (*backendLinkType, error) {
```
TortillaZHawaii, line 676: I suggest using simpler name
```suggestion
func (lc *L4NetLBController) getBackendLinkType(service *v1.Service, svcLogger klog.Logger) (*backendLinkType, error) {
```

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+	if len(backendService.Backends) > 0 {
+		firstBackend := backendService.Backends[0]
+		if strings.Contains(firstBackend.Group, "/instanceGroups/") {
+			return ptr.To(instanceGroupLink), nil
+		}
+		if strings.Contains(firstBackend.Group, "/networkEndpointGroups/") {
+			return ptr.To(negLink), nil
+		}
+	}
+	return nil, nil
```
TortillaZHawaii, line 703: nit: reverse if to reduce nesting

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+func (lc *L4NetLBController) backendTypeOfTheBackendService(service *v1.Service, svcLogger klog.Logger) (*backendLinkType, error) {
+	bsName := lc.namer.L4Backend(service.Namespace, service.Name)
+	backendService, err := lc.backendPool.Get(bsName, meta.VersionGA, meta.Regional, svcLogger)
```
TortillaZHawaii, line 678: Do we need to fetch it here and do additional API call? I believe l4netlbcontroller already fetches it somewhere.

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -778,6 +778,109 @@ func TestProcessIGServiceWhenNEGIsEnabled(t *testing.T) {
...
+		t.Errorf("%v", err)
+	}
+	deleteNetLBService(lc, svc)
+}
+
+// Test the rollback scenario to a version where NEGs are not default but enabled.
+func TestFinalizerRemoveNEGServiceUpdate(t *testing.T) {
+	lc := newL4NetLBServiceController()
+	lc.enableNEGSupport = true
+	lc.enableNEGAsDefault = true
```
TortillaZHawaii, line 846: I believe this is a typo. The comment mentions that NEGs are not default. Am I missing something?

TortillaZHawaii (changes requested):
First round, I will take another look at `pkg/l4lb/l4netlbcontroller_test.go` tommorow

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -778,6 +778,109 @@ func TestProcessIGServiceWhenNEGIsEnabled(t *testing.T) {
...
+		t.Errorf("the service %s should not have the V3 finalizer but instead it had %v", svc.Name, svc.Finalizers)
+	}
+	validateNetLBSvcStatus(svc, t)
+	if err := checkBackendService(lc, svc); err != nil {
+		t.Errorf("UnexpectedError %v", err)
+	}
+	if err := validateAnnotations(svc); err != nil {
+		t.Errorf("%v", err)
+	}
+	deleteNetLBService(lc, svc)
```
TortillaZHawaii, line 839: Can we skip checking resources being removed? This func only deletes the service definition.

pkg/l4lb/l4netlbcontroller_test.go
```diff
@@ -778,6 +778,109 @@ func TestProcessIGServiceWhenNEGIsEnabled(t *testing.T) {
...
+		t.Errorf("the service %s should not have the V2 finalizer but instead it had %v", svc.Name, svc.Finalizers)
+	}
+	// validate that NEG is still attached
+	if err := checkBackendServiceWithNEG(lc, svc); err != nil {
+		t.Errorf("UnexpectedError %v", err)
+	}
+	if err := validateAnnotations(svc); err != nil {
+		t.Errorf("%v", err)
+	}
+	deleteNetLBService(lc, svc)
```
TortillaZHawaii, line 881: Same as above
https://github.com/kubernetes/ingress-gce/pull/2880#discussion_r2097409774

pkg/l4lb/l4netlbcontroller.go
mmamczur: it does but it's late unfortunately and we need it early. we could push it down but this would mean a lot of refactoring :/

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,51 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+		if err != nil {
+			svcLogger.Error(err, "Could not determine the backend type of the existing backend service, deferring to default")
+			return true
+		}
+		if backendType != nil && *backendType == instanceGroupLink {
+			return false
+		}
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
mmamczur: done

pkg/l4lb/l4netlbcontroller.go
mmamczur: done

pkg/l4lb/l4netlbcontroller.go
mmamczur: done

pkg/l4lb/l4netlbcontroller_test.go
mmamczur: changed the comment. NEGs should be default now

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -655,7 +656,52 @@ func (lc *L4NetLBController) shouldUseNEGBackends(service *v1.Service) bool {
...
+	backendType, err := lc.getBackendLinkType(service, svcLogger)
+	if err != nil {
+		svcLogger.Error(err, "Could not determine the backend type of the existing backend service, deferring to default")
+		return true
+	}
+	if backendType != nil && *backendType == instanceGroupLink {
+		return false
+	}
+	return backendType == nil || *backendType == negLink
+
```
TortillaZHawaii, line 674: empty line, surprised this wasn't cleaned by fmt

TortillaZHawaii:
/lgtm

/hold 
- please remove empty line and run unhold to merge

PR #2881, opened 2025-05-19
Add List func to ForwardingRules

TortillaZHawaii:
~ListForwardingRules in `pkg/composite/gen.go` doesn't let us pass filters which necessitates copying most of the code from there and adding an option to pass filter.~

This will be used for optimized NetLB FRs in the future.

### DIFF
--- a/pkg/forwardingrules/forwarding_rules.go
+++ b/pkg/forwardingrules/forwarding_rules.go
@@ -3,6 +3,7 @@ package forwardingrules
 import (
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


TortillaZHawaii:
I'll check if we can do something with generator itself to avoid manually writing this.

TortillaZHawaii:
Ok, it looks like we can. https://github.com/kubernetes/ingress-gce/pull/2887 has to be merged first

/hold

TortillaZHawaii:
Fixed

PR #2884, opened 2025-05-20
[Cherrypick #2882 to release-1.35] NEGController: rely on the presence of the V3 finalizer vs the RBS annotation

mmamczur:
When deciding if a NetLB requires NEGs rely on the presence of the V3 finalizer and not of the RBS annotation.

Sometimes customers create an RBS service but then remove annotations and/or finalizers. In that case the L4 NetLB can detect it is an RBS service and restore the V3 finalizer. The NEG controller always expected the RBS annotation, so it could result in missing NEGs.


TortillaZHawaii:
/lgtm

This was reviewed here: https://github.com/kubernetes/ingress-gce/pull/2882

PR #2886, opened 2025-05-20
Remove all ASM functionality

panslava:
- Keep the flags to not break existing releases, but deprecate them


swetharepakula (commented):
/lgtm
/approve

PR #2888, opened 2025-05-22
Add RemoveUnusedDiscretePorts step

TortillaZHawaii:
This is 4b part of the optimized fr design.

### DIFF
--- a/pkg/forwardingrules/optimized/steps/step.go
+++ b/pkg/forwardingrules/optimized/steps/step.go
@@ -1,0 +1,13 @@
+package steps
+
+import (
+	api_v1 "k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/composite"
+)
+
+// ResourceName is a type alias for string, representing the name of a GCE resource.
+type ResourceName string
+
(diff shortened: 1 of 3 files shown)


pkg/forwardingrules/optimized/steps/step.go
```diff
@@ -0,0 +1,13 @@
...
+	api_v1 "k8s.io/api/core/v1"
+	"k8s.io/ingress-gce/pkg/composite"
+)
+
+// ResourceName is a type alias for string, representing the name of a GCE resource.
+type ResourceName string
+
+// Step represents single transformation of Forwarding Rules.
+// Multiple steps are chained to achieve desired transformation.
+type Step func(ports []api_v1.ServicePort, frs map[ResourceName]*composite.ForwardingRule) error
```
mmamczur, line 13: do you later intend to have a list of these and apply in sequence? will we have to dynamically change the order of steps or add/remove steps?

TortillaZHawaii:
/hold

@mmamczur I actually cannot use the map with names as keys here, as I add the names for non-existing FRs at the end, and I don't want to pollute the logic by combining those two steps

PR #2895, opened 2025-05-28
Add Compare for sets of Forwarding Rules

TortillaZHawaii:
This allows to compare two slices of Forwarding Rules. It has Equal and Patchable as parameters to allow it to be used for both IPv4 and IPv6 which have different func implementations.

We also add tests for verifying logic specific to compare, as exact equal/patchable logic are tested elsewhere.


TortillaZHawaii:
Added comments in Compare func to describe why update returns not only Updates

PR #2900, opened 2025-06-04
NEG controller, generateSubnetToNegNameMap multinet fix

mmamczur:
NEG Controller, when generating subnet to NEG name map, for mutlinet services return one element mapping of cluster default subnetwork to default NEG name. Do not use the network from networkInfo since it will not contain the cluster default network but the service multinet network.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -396,7 +396,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 }
 
 func (s *transactionSyncer) generateSubnetToNegNameMap(subnetConfigs []nodetopologyv1.SubnetConfig) (map[string]string, error) {
-	defaultSubnet, err := utils.KeyName(s.networkInfo.SubnetworkURL)
+	defaultSubnet, err := utils.KeyName(s.cloud.SubnetworkURL())
 	if err != nil {
 		s.logger.Error(err, "Errored getting default subnet from NetworkInfo when retrieving existing endpoints")
 		return nil, err
--- a/pkg/neg/syncers/transaction_test.go
+++ b/pkg/neg/syncers/transaction_test.go
@@ -589,6 +589,86 @@ func TestNegNameMultiNetworking(t *testing.T) {
 	}
 }
 
+func TestGenerateSubnetToNegNameMapMultiNetworking(t *testing.T) {
+	fakeGCE := gce.NewFakeGCECloud(test.DefaultTestClusterValues())
+	fakeCloud := negtypes.NewAdapter(fakeGCE)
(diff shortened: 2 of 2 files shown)


pkg/neg/syncers/transaction.go
```diff
@@ -396,7 +396,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 }
 
 func (s *transactionSyncer) generateSubnetToNegNameMap(subnetConfigs []nodetopologyv1.SubnetConfig) (map[string]string, error) {
-	defaultSubnet, err := utils.KeyName(s.networkInfo.SubnetworkURL)
+	defaultSubnet, err := utils.KeyName(s.cloud.SubnetworkURL())
```
swetharepakula, line 399: Though this is called default subnet, really what we are looking it is this the default subnet defined in the networkInfo (for multi net, this means not the default subnetwork for the cluster).

If you look to the comments on line 405, that indicates why we are comparing against networkInfo instead of cloud.SubnetworkURL

swetharepakula:
Looking through the code I think here is the mismatch that is occurring in the multinet scenario:

1. When we query the current, we get the correct information because we query for the base Neg name (because this is the only thing in the map): 

https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/utils.go#L701

Since we use the map to determine the subnet name we get the correct subnet name in the logs.

2. When we calculate the endpoints, we depend on the zoneGetter which has the defaultSubnet "hardcoded" in, so when we generate the map, all the endpoints are shown as part of the defaultSubnet instead of the multi nic one:
https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/endpoints_calculator.go#L108-L114 and https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/endpoints_calculator.go#L189-L194.

3. The subsets that get returned use subnets from the map: https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/subsets.go#L216-L217

4. when calculating the diff, everything will be considered different: https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/utils.go#L92-L109.

5. When attaching and detaching, we get lucky that we do not need to provide the subnet: https://github.com/kubernetes/ingress-gce/blob/master/pkg/neg/syncers/transaction.go#L687-L692.

The fix in this PR will make (1) match the subnet distribution as (2), however the logs will be incorrect. Both current and desired will show endpoints being attached to the NEG as attaching the default subnet in the multi-net case which is incorrect. My other concern is that everywhere else in the NEG code we are using networkInfoSubnetwork to calculate defaultSubnet and this will become confusing as a special case.

Therefore, I think we should find a more correct fix. Likely something in (2) or we teach the zoneGetter about multinet.

PR #2903, opened 2025-06-09
NEG Controller: fix the multinet NEG location to prevent constant NEG detach calls

mmamczur:
NEG Controller, when generating desired endpoints in the endpoints calculator for Multi Networking service, use the MN network as as the NEG subnet instead of the subnet of the node.

This is to make sure the location from the calculator matches the 'current' state calculated before and also so that the location actually matches the actual subnet the NEG will reside in.
Without this change the NEG controller calculation will be different for current state and for desired, so the controller will think it has to detach and attach endpoints. This will result in the controller issuing constant detach calls, exhausting the quota in the project for the NEG API.


gauravkghildiyal:
Thanks @mmamczur. This ....
/lgtm

PR #2905, opened 2025-06-13
Safeguard gcpfirewalls cr creation to not return sync errors in Shadow run

maciejriedl:
Make sure that during gcpfirewalls cr creation, sync errors are not returned in Shadow run (when l7LB still reconciles the FWs).

Also fix dryRun flag to properly mean "testing" flow (the logic of the flag usage has not been changed)

### DIFF
--- a/pkg/firewalls/firewalls_l7_cr.go
+++ b/pkg/firewalls/firewalls_l7_cr.go
@@ -41,7 +41,10 @@ type FirewallCR struct {
 	// all the port ranges to open with each call to Sync()
 	nodePortRanges []string
 	firewallClient firewallclient.Interface
-	dryRun         bool
+	// If set to true, CRs are marked as "disabled" and errors are not
+	// propagated. Firewalls should be still created/updated by L7LB, not the
+	// Platform Firewall
+	dryRun bool
 
 	logger klog.Logger
 }
@@ -85,13 +88,13 @@ func (fr *FirewallCR) Sync(nodeNames, additionalPorts, additionalRanges []string
 	if err != nil {
 		return err
 	}
-	return ensureFirewallCR(fr.firewallClient, expectedFirewallCR, fr.logger)
+	return ensureFirewallCR(fr.firewallClient, expectedFirewallCR, fr.logger, fr.dryRun)
 }
 
 // ensureFirewallCR creates/updates the firewall CR
 // On CR update, it will read the conditions to see if there are errors updated by PFW controller.
 // If the Spec was updated by others, it will reconcile the Spec.
-func ensureFirewallCR(client firewallclient.Interface, expectedFWCR *gcpfirewallv1.GCPFirewall, logger klog.Logger) error {
+func ensureFirewallCR(client firewallclient.Interface, expectedFWCR *gcpfirewallv1.GCPFirewall, logger klog.Logger, dryRun bool) error {
 	fw := client.NetworkingV1().GCPFirewalls()
 	currentFWCR, err := fw.Get(context.Background(), expectedFWCR.Name, metav1.GetOptions{})
 	logger.V(3).Info("ensureFirewallCR Get CR", "currentFirewallCR", fmt.Sprintf("%+v", currentFWCR), "err", err)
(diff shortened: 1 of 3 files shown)


pkg/firewalls/firewalls_l7_cr.go
```diff
@@ -41,7 +41,8 @@ type FirewallCR struct {
 	// all the port ranges to open with each call to Sync()
 	nodePortRanges []string
 	firewallClient firewallclient.Interface
-	dryRun         bool
+	// If firewalls are still reconciled by l7LB.
```
swetharepakula: nit: this comment isn't clear. Can you expand this?

pkg/firewalls/firewalls_l7_cr.go
```diff
@@ -116,7 +121,11 @@ func ensureFirewallCR(client firewallclient.Interface, expectedFWCR *gcpfirewall
 			con.Reason == string(gcpfirewallv1.FirewallRuleReasonSyncError) {
 			// Use recorder to emit the cmd in Sync()
 			logger.V(3).Info("ensureFirewallCR: Could not enforce Firewall CR", "currentFirewallCRName", currentFWCR.Name, "reason", con.Reason)
-			return fmt.Errorf(con.Reason)
+			if dryRun {
+				return nil
```
swetharepakula, line 127: why do we not want to return this error?

swetharepakula (commented):
The change looks okay to me. However wanted to clarify why we aren't returning an error.

The flag flip should result in a test change. Do we have an unit tests?

pkg/firewalls/firewalls_l7_cr.go
maciejriedl: This is the point of the dryRun - If we are in the "testing" flow, we don't want CRs to eventually cause disruptions to the L7LB (error in PFW could cause L7LB to receive infinte errors)

pkg/firewalls/firewalls_l7_cr.go
maciejriedl: done

pkg/firewalls/firewalls_l7_cr.go
```diff
@@ -41,7 +41,10 @@ type FirewallCR struct {
 	// all the port ranges to open with each call to Sync()
 	nodePortRanges []string
 	firewallClient firewallclient.Interface
-	dryRun         bool
+	// If set to true, CRs are marked as "disabled" and errors are not
+	// propagated. Firewalls should be still created/updated by l7LB, not the
```
yswe:
```suggestion
	// propagated. Firewalls should be still created/updated by L7LB, not the
```

pkg/firewalls/firewalls_l7_cr.go
```diff
@@ -81,17 +84,17 @@ func (fr *FirewallCR) Sync(nodeNames, additionalPorts, additionalRanges []string
 	ranges.Insert(additionalRanges...)
 
 	fr.logger.V(3).Info("Firewall CR is enabled.")
-	expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{}, fr.dryRun)
+	expectedFirewallCR, err := NewFirewallCR(name, ports.List(), ranges.UnsortedList(), []string{}, !fr.dryRun)
```
yswe: Should we use `dryRun` instead of `enforced` in the `NewFirewallCR` to improve the readability?

pkg/firewalls/firewalls_l7_cr.go
maciejriedl: done. thank you!

pkg/firewalls/firewalls_l7_cr.go
maciejriedl: Makes sense. thank you!

pkg/firewalls/firewalls_l7_cr.go
maciejriedl: Also added a unit test for the changes

yswe:
/lgtm

Please align with @swetharepakula on the approach of not returning error, in case we're missing something and this error still has to be returned for any reason.

pkg/firewalls/firewalls_l7_cr.go
swetharepakula: Should whether we use the error or not happen at the caller instead of here?

Also wondering should we add a metric for the error so we can monitor this. There is the log line, but it will be hard to see that we are doing okay like that

PR #2907, opened 2025-06-17
Multiproject: stop controllers when leadership ends

panslava:
- Scope: changes are confined to the multiproject leader-election flow.
- Controllers & informers now inherit the leader-election context, exiting immediately when the lease is lost or on SIGTERM.
- ReleaseOnCancel=true enables instant lease hand-off for smoother rollouts.

/assign @swetharepakula

### DIFF
--- a/pkg/multiproject/start/start.go
+++ b/pkg/multiproject/start/start.go
@@ -33,7 +33,7 @@ const multiProjectLeaderElectionLockName = "ingress-gce-multi-project-lock"
 
 // StartWithLeaderElection starts the ProviderConfig controller with leader election.
 func StartWithLeaderElection(
-	ctx context.Context,
+	parentCtx context.Context,
 	leaderElectKubeClient kubernetes.Interface,
 	hostname string,
 	logger klog.Logger,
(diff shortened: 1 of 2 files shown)


pkg/multiproject/start/start.go
```diff
@@ -99,13 +103,18 @@ func makeLeaderElectionConfig(
...
-		LeaseDuration: flags.F.LeaderElection.LeaseDuration.Duration,
-		RenewDeadline: flags.F.LeaderElection.RenewDeadline.Duration,
-		RetryPeriod:   flags.F.LeaderElection.RetryPeriod.Duration,
+		Lock:            rl,
+		LeaseDuration:   flags.F.LeaderElection.LeaseDuration.Duration,
+		RenewDeadline:   flags.F.LeaderElection.RenewDeadline.Duration,
+		RetryPeriod:     flags.F.LeaderElection.RetryPeriod.Duration,
+		ReleaseOnCancel: true,
 		Callbacks: leaderelection.LeaderCallbacks{
-			OnStartedLeading: func(context.Context) {
```
aojea, line 107: why do not use the `OnStartedLeading` and `OnStoppedLeading` hooks instead of overloading the `OnStartedLeading` hook?

panslava: Hey, I am not sure exactly, which changes does this require? I see we use very similar pattern across other controllers in ingress-gce... 

Do you have an example?

aojea: see how this is implemented https://github.com/kubernetes/kubernetes/blob/885db534cabb9c1975553f43e32c90f0be256a51/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L211-L221

```go
func (le *LeaderElector) Run(ctx context.Context) {
	defer runtime.HandleCrash()
	defer le.config.Callbacks.OnStoppedLeading()

	if !le.acquire(ctx) {
		return // ctx signalled done
	}
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	go le.config.Callbacks.OnStartedLeading(ctx)
	le.renew(ctx)
}
```

one the context is canceled it calls ` le.config.Callbacks.OnStoppedLeading()`

aojea: I mean, this works, but is an odd usage

panslava: Hope I got it, is it more conventional now?

aojea: you can decide one or the other, I just found it odd

gauravkghildiyal:
As discussed, while I think this fix makes sense and this has brought to light issues not specific to multi-project, the current plan is to dig a bit more whether this fix solves the immediate problem

We do need to separately fix leader election in general for both multi-project and standard modes.

/hold for further evaluation.

swetharepakula:
/unhold

We discussed and agree that we need to fix the two sides separately (multiproject, single controller). 

Thanks for discovering and fixing Slavik!

PR #2909, opened 2025-06-19
Fix shadowing error in tests

panslava:
Problem was, we had `err := verifyLogging` but should be `err = verifyLogging`

### DIFF
--- a/cmd/e2e-test/logging_test.go
+++ b/cmd/e2e-test/logging_test.go
@@ -170,9 +170,11 @@ func TestLogging(t *testing.T) {
 				gclb, err = fuzz.GCLBForVIP(context.Background(), Framework.Cloud, params)
 				if err != nil {
 					t.Logf("Failed to GCP resources for LB with IP = %q: %v", vip, err)
+					err = fmt.Errorf("failed to GCP resources for LB with IP = %q: %v", vip, err)
 					return false, nil
 				}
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/logging_test.go
```diff
@@ -170,9 +170,11 @@ func TestLogging(t *testing.T) {
 				gclb, err = fuzz.GCLBForVIP(context.Background(), Framework.Cloud, params)
 				if err != nil {
 					t.Logf("Failed to GCP resources for LB with IP = %q: %v", vip, err)
+					err = fmt.Errorf("failed to GCP resources for LB with IP = %q: %v", vip, err)
```
gauravkghildiyal, line 173: Sorry I am a bit confused. I understand shadowing in general. But could you share where exactly this helps us here? Like where is this new "err" used?

panslava: On this line https://github.com/kubernetes/ingress-gce/blob/1dc3a4106afc5fb5b33cdf547a7489b838f58bc1/cmd/e2e-test/logging_test.go#L182

Now, it says lastErr is nil, which basically means `fuzz.GCLBForVIP(context.Background(), Framework.Cloud, params)` returned no error, but `verifyLogging` call error is stored in a new var. And we never log that error, so it's completely lost actually

gauravkghildiyal: Thanks. I'm sorry I didn't see that.

PR #2914, opened 2025-07-02
Compare health checks via resource path to support multi-version APIs

08volt:
This PR refactors the equality check between backend services (backendSvcEqual) to correctly handle health check links that may have been generated using different API versions (e.g., v1, beta).

Previously, the comparison was done on the full health check URL. This caused a mismatch if the existing backend service referenced a health check with a beta link while the controller generated an expected link using the GA API (v1), even though both URLs pointed to the exact same GCP resource. This would trigger unnecessary and redundant BackendService.Update calls.

This change makes the comparison logic more robust by normalizing the health check links into a canonical resource path, ensuring that we only compare the resource's identity.

- A new utility function utils.GetResourceFromBasePath has been added. It takes a full resource URL (e.g., https://www.googleapis.com/compute/v1/...) and extracts the stable resource path (e.g., projects/my-project/global/healthChecks/...).

- The backendSvcEqual function now uses this utility to convert both the new and old health check URL slices into slices of resource paths.

- The comparison is then performed on these normalized resource paths using utils.EqualStringSets, ensuring that differences in API versions or endpoints are ignored.

### DIFF
--- a/pkg/backends/backends.go
+++ b/pkg/backends/backends.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/cloud-provider-gcp/providers/gce"
 	"k8s.io/ingress-gce/pkg/backends/features"
 	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/network"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
@@ -493,12 +494,19 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
 // since that is handled by the neg-linker.
 // The list of backends is not checked, since that is handled by the neg-linker.
 func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTracking bool) bool {
+
 	svcsEqual := newBS.Protocol == oldBS.Protocol &&
 		newBS.Description == oldBS.Description &&
 		newBS.SessionAffinity == oldBS.SessionAffinity &&
 		newBS.LoadBalancingScheme == oldBS.LoadBalancingScheme &&
-		utils.EqualStringSets(newBS.HealthChecks, oldBS.HealthChecks) &&
 		newBS.Network == oldBS.Network
+
+	if flags.F.EnableL4ILBZonalAffinity {
+		// Compare healthChecks sets ignoring api version
+		svcsEqual = svcsEqual && healthChecksEqual(newBS.HealthChecks, oldBS.HealthChecks)
+	} else {
+		svcsEqual = svcsEqual && utils.EqualStringSets(newBS.HealthChecks, oldBS.HealthChecks)
+	}
 
 	// Compare only for backendSvc that uses Strong Session Affinity feature
 	if compareConnectionTracking {
@@ -516,12 +524,31 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
 	return svcsEqual
 }
 
+// removeAPIVersionFromHealthChecks converts a slice of full health check URLs
+// into a slice of their URL without the API version
+func removeAPIVersionFromHealthChecks(hcLinks []string) []string {
+	hcResourcePaths := make([]string, 0, len(hcLinks))
+	for _, hcLink := range hcLinks {
+		resourcePath := utils.FilterAPIversionFromResourcePath(hcLink)
+		hcResourcePaths = append(hcResourcePaths, resourcePath)
+	}
+	return hcResourcePaths
+}
+
 func convertNetworkLbTrafficPolicyToZonalAffinity(trafficPolicy *composite.BackendServiceNetworkPassThroughLbTrafficPolicy) composite.BackendServiceNetworkPassThroughLbTrafficPolicyZonalAffinity {
 	if trafficPolicy == nil || trafficPolicy.ZonalAffinity == nil {
 		return *zonalAffinityDisabledTrafficPolicy().ZonalAffinity
 	}
 
 	return *trafficPolicy.ZonalAffinity
+}
+
+// healthCheckEqual compare healthcheck URL ignoring the API version used
+func healthChecksEqual(hcLinksA, hcLinksB []string) bool {
+	healthChecksA := removeAPIVersionFromHealthChecks(hcLinksA)
+	healthChecksB := removeAPIVersionFromHealthChecks(hcLinksB)
+
+	return utils.EqualStringSets(healthChecksA, healthChecksB)
 }
 
 func zonalAffinityEqual(a, b *composite.BackendService) bool {
--- a/pkg/backends/backends_test.go
+++ b/pkg/backends/backends_test.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cloud-provider-gcp/providers/gce"
 	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/network"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
@@ -347,6 +348,7 @@ func TestBackendSvcEqual(t *testing.T) {
 		oldBackendService         *composite.BackendService
 		newBackendService         *composite.BackendService
 		compareConnectionTracking bool
+		withZonalAffinityEnabled  bool
 		wantEqual                 bool
 	}{
 		{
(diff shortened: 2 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -908,3 +908,27 @@ func GetDomainFromGABasePath(basePath string) string {
 	domain = strings.TrimSuffix(domain, "/compute/v1")
 	return domain
 }
+
+// GetResourceFromBasePath takes any API version base path of the form <path>/compute/<version>/<resource_path> and returns the resource_path.
+func GetResourceFromBasePath(basePath string) string {
```
TortillaZHawaii: Tests are missing in `pkg/utils`. I see that part of it is tested in `pkg/backends/backends_test.go`, but since this is utils we could use it in other contexts as well.

pkg/backends/backends_test.go
```diff
@@ -478,6 +478,26 @@ func TestBackendSvcEqual(t *testing.T) {
...
+		{
+			desc: "Test with changed health-checks version beta-v1",
+			oldBackendService: &composite.BackendService{
+				HealthChecks: []string{"https://www.googleapis.com/compute/beta/abc"},
+			},
+			newBackendService: &composite.BackendService{
+				HealthChecks: []string{"https://www.googleapis.com/compute/v1/abc"},
+			},
+			wantEqual: true,
+		},
```
TortillaZHawaii: Please add a tests where `wantEqual: false`, but we have those long links. Like:
* "https://www.googleapis.com/compute/beta/abc" and "https://www.googleapis.com/compute/beta/abcd"
* "https://www.googleapis.com/compute/v1/abc" and "https://www.googleapis.com/compute/beta/abcd"

pkg/backends/backends.go
```diff
@@ -516,6 +521,18 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
 	return svcsEqual
 }
 
+// convertHealthCheckURLstoResourcePaths converts a slice of full health check URLs
+// into a slice of their corresponding resource paths (e.g., "projects/my-project/global/healthChecks/my-hc").
+func convertHealthCheckURLstoResourcePaths(hcLinks []string) []string {
+	hcResourcePaths := make([]string, 0, len(hcLinks))
+	for _, hcLink := range hcLinks {
+		// Use the utility function to extract the path.
```
TortillaZHawaii: Do we need this comment?

pkg/backends/backends.go
```diff
@@ -497,9 +497,14 @@ func backendSvcEqual(newBS, oldBS *composite.BackendService, compareConnectionTr
...
 		newBS.SessionAffinity == oldBS.SessionAffinity &&
 		newBS.LoadBalancingScheme == oldBS.LoadBalancingScheme &&
-		utils.EqualStringSets(newBS.HealthChecks, oldBS.HealthChecks) &&
 		newBS.Network == oldBS.Network
 
+	// Compare health checks by their parsed resource IDs.
+	newHealthChecks := convertHealthCheckURLstoResourcePaths(newBS.HealthChecks)
+	oldHealthChecks := convertHealthCheckURLstoResourcePaths(oldBS.HealthChecks)
+
+	svcsEqual = svcsEqual && utils.EqualStringSets(newHealthChecks, oldHealthChecks)
```
TortillaZHawaii: nit: I would move this code into a private func like healthChecksEqual(newBS.HealthChecks, oldBS.HealthChecks) and just call it with introduction of svcsEqual, rather than mutating `svcsEqual = svcEqual && ...`, but since this is already done below this might be fine as well.

pkg/utils/utils.go
08volt: Done thanks

pkg/backends/backends_test.go
08volt: done thanks

pkg/backends/backends.go
08volt: removed

pkg/backends/backends.go
08volt: Thats right, fixed

pkg/utils/utils_test.go
```diff
@@ -1754,3 +1754,49 @@ func TestGetDomainFromGABasePath(t *testing.T) {
...
 	}
 }
+
+func TestGetResourceFromBasePath(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		desc     string
+		basePath string
+		want     string
+	}{
```
FelipeYepez, line 1764: Can you also test what would happen with "https://www.googleapis.com/compute/"

pkg/utils/utils.go
FelipeYepez: It would be better to compare all the path except for the API used (v1, beta, etc) instead than only from "/compute/"

pkg/backends/backends_test.go
```diff
@@ -729,6 +786,9 @@ func TestBackendSvcEqual(t *testing.T) {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
 			t.Parallel()
+			if tc.withZonalAffinityEnabled {
```
FelipeYepez: You can try to use 
```
oldFlag := flags.F.EnableL4ILBZonalAffinity
flags.F.EnableL4ILBZonalAffinity = true
defer func() { flags.F.EnableL4ILBZonalAffinity = oldFlag }()
```

pkg/backends/backends_test.go
```diff
@@ -729,6 +786,9 @@ func TestBackendSvcEqual(t *testing.T) {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
 			t.Parallel()
```
FelipeYepez: Please check if `t.Parallel()` will affect testing since you are modifying `flags.F.EnableL4ILBZonalAffinity`

pkg/backends/backends_test.go
```diff
@@ -478,6 +480,61 @@ func TestBackendSvcEqual(t *testing.T) {
...
+			desc: "Test with changed health-checks version beta-v1",
+			oldBackendService: &composite.BackendService{
+				HealthChecks: []string{"https://www.googleapis.com/compute/v1/abc"},
+			},
+			newBackendService: &composite.BackendService{
+				HealthChecks: []string{"https://www.googleapis.com/compute/beta/abcd"},
+			},
+			withZonalAffinityEnabled: true,
+			wantEqual:                false,
+		},
```
FelipeYepez, line 548: Try checking with a different link start
`https://www.googleapis.com/compute/v1/abc`
`https://www.google.com/compute/v1/abc`
Should have `wantEqual: false`

pkg/backends/backends.go
```diff
@@ -524,6 +543,14 @@ func convertNetworkLbTrafficPolicyToZonalAffinity(trafficPolicy *composite.Backe
 	return *trafficPolicy.ZonalAffinity
 }
 
+func healthChecksEqual(hcLinksA, hcLinksB []string) bool {
+	// Compare health checks by their parsed resource IDs.
+	newHealthChecks := convertHealthCheckURLstoResourcePaths(hcLinksA)
+	oldHealthChecks := convertHealthCheckURLstoResourcePaths(hcLinksB)
```
FelipeYepez: Seems like old/new is irrelevant in this function, please rename `newHealthChecks` and `oldHealthChecks`

pkg/backends/backends.go
```diff
@@ -524,6 +543,14 @@ func convertNetworkLbTrafficPolicyToZonalAffinity(trafficPolicy *composite.Backe
 	return *trafficPolicy.ZonalAffinity
 }
 
+func healthChecksEqual(hcLinksA, hcLinksB []string) bool {
```
FelipeYepez, line 547: You can mention that this function ignores the api version when comparing or try to rename the function

FelipeYepez (changes requested):

pkg/backends/backends_test.go
08volt: Would they really be different resources in this case that it is necessary to trigger an update or would they reference the same HC?

pkg/utils/utils_test.go
08volt: done

pkg/utils/utils.go
08volt: done

pkg/backends/backends_test.go
08volt: it did, removed, thanks

pkg/backends/backends_test.go
08volt: Fixed

pkg/backends/backends.go
08volt: Naming is similar to the other functions comparing other BackendService properties, I added a comment

pkg/backends/backends_test.go
08volt: fixed thanks

pkg/backends/backends.go
08volt: fixed thanks

pkg/utils/utils_test.go
```diff
@@ -1754,3 +1754,49 @@ func TestGetDomainFromGABasePath(t *testing.T) {
...
+			desc:     "path ends after version",
+			basePath: "https://www.googleapis.com/compute/v1/",
+			want:     "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			if got := FilterAPIversionFromResourcePath(tc.basePath); got != tc.want {
+				t.Errorf("GetResourceFromBasePath(%q) = %q, want %q", tc.basePath, got, tc.want)
```
TortillaZHawaii, line 1798: name mismatch, I think I prefer `Get` rather then `Filter`

PR #2918, opened 2025-07-07
Avoid odd reservation calls for user specified IP addresses in "networking.gke.io/load-balancer-ip-addresses" annotation

midavadim:
Avoid odd reservation calls for user specified IP addresses in "networking.gke.io/load-balancer-ip-addresses" annotation.

Such IP address are not managed by ILB controller. 

Previously controller ran 3 additional calls for such IP address:
m.svc.GetRegionAddress(m.name, m.region)
m.svc.ReserveRegionAddress(newAddr, m.region)
addr, err := m.svc.GetRegionAddressByIP(m.region, m.targetIP)


pkg/address/to_use.go
```diff
@@ -18,30 +18,30 @@ import (
 //  2. Use .Spec.LoadBalancerIP (old field, was deprecated).
 //  3. Use existing forwarding rule IP. If subnetwork was changed (or no existing IP),
 //     reset the IP (by returning empty string).
-func IPv4ToUse(cloud *gce.Cloud, recorder record.EventRecorder, svc *v1.Service, fwdRule *composite.ForwardingRule, requestedSubnet string) (string, error) {
+func IPv4ToUse(cloud *gce.Cloud, recorder record.EventRecorder, svc *v1.Service, fwdRule *composite.ForwardingRule, requestedSubnet string) (string, string, error) {
 	// Get value from new annotation which support both IPv4 and IPv6
-	ipv4FromAnnotation, err := annotations.FromService(svc).IPv4AddressAnnotation(cloud)
+	ipv4FromAnnotation, ipNameFromAnnotation, err := annotations.FromService(svc).IPv4AddressAnnotation(cloud)
```
midavadim, line 23: here we have already done a call
m.svc.GetRegionAddress(m.name, m.region)
which gets IpAddress by its name specified by in annotation.

So if ipNameFromAnnotation exists then we no longer need to make further address related calls to ensure given IP.

pkg/address/to_use.go
```diff
@@ -50,33 +50,33 @@ func IPv4ToUse(cloud *gce.Cloud, recorder record.EventRecorder, svc *v1.Service,
 //  1. Use static addresses annotation "networking.gke.io/load-balancer-ip-addresses".
 //  2. Use existing forwarding rule IP. If subnetwork was changed (or no existing IP),
 //     reset the IP (by returning empty string).
-func IPv6ToUse(cloud *gce.Cloud, svc *v1.Service, ipv6FwdRule *composite.ForwardingRule, requestedSubnet string, logger klog.Logger) (string, error) {
+func IPv6ToUse(cloud *gce.Cloud, svc *v1.Service, ipv6FwdRule *composite.ForwardingRule, requestedSubnet string, logger klog.Logger) (string, string, error) {
```
TortillaZHawaii: Please use a struct with named fields instead of specifying another string, as the next reader would have to go to the function to determine what is returned in which order.

pkg/loadbalancers/l4.go
```diff
@@ -445,21 +445,26 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+			isReserved = true
+		}
 
+		if !l4.cloud.IsLegacyNetwork() && !isReserved {
+			l4.svcLogger.V(2).Info("EnsureInternalLoadBalancer, reserve existing IPv4 address before making any changes")
+			ipv4AddressName = l4.GetFRName()
 			nm := types.NamespacedName{Namespace: l4.Service.Namespace, Name: l4.Service.Name}.String()
 			// ILB can be created only in Premium Tier
-			addrMgr := address.NewManager(l4.cloud, nm, l4.cloud.Region(), subnetworkURL, expectedFRName, ipv4AddressToUse, cloud.SchemeInternal, cloud.NetworkTierPremium, address.IPv4Version, l4.svcLogger)
+			addrMgr := address.NewManager(l4.cloud, nm, l4.cloud.Region(), subnetworkURL, ipv4AddressName, ipv4AddressToUse, cloud.SchemeInternal, cloud.NetworkTierPremium, address.IPv4Version, l4.svcLogger)
```
mmamczur: here we start to use ipv4AddressName instead of the expectedFRName

however, this value is never the thing that we get from the IPv4ToUse from above

In fact, we never care about the value returned from that call, other than it being nil or not. Maybe we should just return isReserved bool?

pkg/loadbalancers/l4.go
```diff
@@ -479,20 +484,27 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
 		if err != nil {
 			result.Error = fmt.Errorf("EnsureInternalLoadBalancer error: address.IPv6ToUse returned error: %w", err)
 			return result
 		}
-		expectedIPv6FRName := l4.getIPv6FRName()
+
+		isReserved := false
+		if ipv6AddrToUse != "" && ipv6AddressName != "" {
+			isReserved = true
+		}
```
TortillaZHawaii:
```suggestion
		isReserved := ipv6AddrToUse != "" && ipv6AddressName != ""
```

pkg/loadbalancers/l4.go
TortillaZHawaii: Address manager will use that name to create static IP address for the duration of recreation of Forwarding Rules. Which name will the static address get if the field here is empty?

https://github.com/kubernetes/ingress-gce/blob/master/pkg/address/manager.go#L107

midavadim: logic has changed slightly.

pkg/address/manager.go
```diff
@@ -103,12 +105,17 @@ func (m *Manager) HoldAddress() (string, IPAddressType, error) {
 	// case of using a controller address, retrieving the address by name results in the fewest API
 	// calls since it indicates whether a Delete is necessary before Reserve.
 	m.frLogger.V(4).Info("Attempting hold of IP", "ip", m.targetIP, "addressType", m.addressType)
-	// Get the address in case it was orphaned earlier
+
+	if m.addressName != "" {
```
midavadim, line 114: if addressName is specified it means that we have a deal with IP reserved by customer. We only need to call HoldAddress to handle possible orphaned Address with default "ForwardingRule" based name

pkg/address/hold_test.go
```diff
@@ -266,6 +266,7 @@ func arrange(t *testing.T, existingRules []*composite.ForwardingRule, service *a
 		Region:      vals.Region,
 		Address:     reservedIPv4,
 		AddressType: "EXTERNAL",
+		NetworkTier: "PREMIUM",
```
midavadim, line 269: it is required, otherwise validation fails. Previously test worked becasue GCEMock allow reservation of the same IP address but with different name and network tier

pkg/address/hold_test.go
```diff
@@ -65,7 +65,7 @@ func TestHoldExternalIPv4(t *testing.T) {
 			},
 			want: address.HoldResult{
 				IP:      reservedIPv4,
-				Managed: address.IPAddrManaged,
+				Managed: address.IPAddrUnmanaged,
```
midavadim, line 68: Address exists. So it is unmanaged Address. It was wrong data in the test

pkg/address/hold.go
```diff
@@ -45,10 +45,11 @@ func HoldExternalIPv4(cfg HoldConfig) (HoldResult, error) {
 	subnet := ""
 	name := utils.LegacyForwardingRuleName(cfg.Service)
 
+	var ipv4AddressName string
 	// Determine IP which will be used for this LB. If no forwarding rule has been established
 	// or specified in the Service spec, then requestedIP = "".
 	rule := pickForwardingRuleToInferIP(cfg.ExistingRules)
-	res.IP, err = IPv4ToUse(cfg.Cloud, cfg.Recorder, cfg.Service, rule, subnet)
+	res.IP, ipv4AddressName, err = IPv4ToUse(cfg.Cloud, cfg.Recorder, cfg.Service, rule, subnet)
```
mmamczur, line 52: how is the .spec.loadBalancerIP handled?

looking at the code from IPv4ToUse the address name would be empty, even though that address may be actually reserverd by the user and named

pkg/address/manager.go
mmamczur: ok I see later that the call to GetRegionAddress should sort this out

pkg/address/manager.go
```diff
@@ -103,8 +105,28 @@ func (m *Manager) HoldAddress() (string, IPAddressType, error) {
...
+		// Name for custom unmanaged static Address which is reserved by user and specified in annotation
+		addressName = m.addressName
+		addressManagementType = IPAddrUnmanaged
+		m.tryRelease = false
+
+		// Custom reserved Address is specified so no possible orphaned default address should exist
+		if addressName != defaultName {
+			if addrDefault, _ := m.svc.GetRegionAddress(defaultName, m.region); addrDefault != nil {
+				_ = m.removeAddress(defaultName, "address is orphaned and not needed anymore")
```
mmamczur: if we can't do anything about the error maybe we should at least log it, if we ever have to debug this area it could make it easier

pkg/address/hold.go
mmamczur: ok I see later that the call to GetRegionAddress should sort this out

pkg/address/to_use.go
mmamczur: or maybe name return params

pkg/loadbalancers/l4_test.go
```diff
@@ -2214,7 +2214,7 @@ func TestDualStackILBStaticIPAnnotation(t *testing.T) {
 	}
 	ipv6Address := &compute.Address{
```
mmamczur, line 2215: I'm wondering if just for future we maybe should set PrefixLength for IPv6, since all created addresses do have it. For now we don't look at this field here but maybe make it consistent with GCE behavior?

pkg/address/hold.go
midavadim: yes, correct.

pkg/address/to_use.go
mmamczur: It would be great to do one of the suggested things here

named return params should be very easy to introduce

mmamczur: It would be great to do one of the suggested things here

named return params should be very easy to introduce

mmamczur:
looks good overall, I'd consider naming the return params in the where we introduced many. If you decide to not do that let us know and we can merge

PR #2919, opened 2025-07-08
Add validation for health check source CIDRs and refactor related functions

panslava:
- Introduced `validateHealthCheckSourceCIDRs` function to validate CIDR format for health check source ranges.
- Refactored `lbSourceRanges` to utilize a new `parseOverrideRanges` function for better clarity and validation.
- Added `isValidIPRange` function to check the validity of IP ranges.
- Created unit tests for health check source CIDR validation in `flags_test.go`.

/assign @gauravkghildiyal 
/cc @bowei

### DIFF
--- a/pkg/firewalls/controller.go
+++ b/pkg/firewalls/controller.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
-	"strings"
 	"time"
 
 	gcpfirewallv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/gcpfirewall/v1"
@@ -39,6 +38,7 @@ import (
 	"k8s.io/ingress-gce/pkg/loadbalancers/features"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+	"k8s.io/ingress-gce/pkg/validation"
 	"k8s.io/klog/v2"
 )
 
@@ -100,10 +100,13 @@ func NewFirewallController(
 	enableCR, disableFWEnforcement, enableRegionalXLB bool,
 	stopCh <-chan struct{},
 	logger klog.Logger,
-) *FirewallController {
+) (*FirewallController, error) {
 	logger = logger.WithName("FirewallController")
 	compositeFirewallPool := &compositeFirewallPool{}
-	sourceRanges := lbSourceRanges(logger, flags.F.OverrideHealthCheckSourceCIDRs)
+	sourceRanges, err := lbSourceRanges(logger, flags.F.OverrideHealthCheckSourceCIDRs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get source ranges: %w", err)
+	}
 	logger.Info("Using source ranges", "ranges", sourceRanges)
 	if enableCR {
 		firewallCRPool := NewFirewallCRPool(ctx.FirewallClient, ctx.Cloud, ctx.ClusterNamer, sourceRanges, portRanges, disableFWEnforcement, logger)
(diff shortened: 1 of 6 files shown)


pkg/firewalls/controller.go
```diff
@@ -201,16 +202,33 @@ func NewFirewallController(
...
+		return parseOverrideRanges(overrideRanges, logger)
+	}
 
-		ranges := strings.Split(overrideRanges, ",")
-		for _, sourceRange := range ranges {
-			sourceRanges = append(sourceRanges, strings.TrimSpace(sourceRange))
+	return gce.L7LoadBalancerSrcRanges()
+}
+
+func parseOverrideRanges(overrideRanges string, logger klog.Logger) []string {
```
bowei: prefer logger as first argument

pkg/firewalls/controller.go
```diff
@@ -201,16 +202,33 @@ func NewFirewallController(
...
+	var sourceRanges []string
+	ranges := strings.Split(overrideRanges, ",")
+	for _, sourceRange := range ranges {
+		trimmed := strings.TrimSpace(sourceRange)
+		if trimmed == "" {
```
bowei: isn't this an error (empty strings is not valid)

pkg/firewalls/controller.go
```diff
@@ -201,16 +202,33 @@ func NewFirewallController(
...
+			continue
+		}
+
+		if isValidIPRange(trimmed) {
+			sourceRanges = append(sourceRanges, trimmed)
+		} else {
+			logger.Error(fmt.Errorf("invalid IP range '%s' in override ranges, skipping", trimmed), "Invalid IP range in override ranges, skipping", "range", trimmed)
```
bowei: Do we just log or the error will cause something to fail?
As this is a startup option, does it make more sense to have it fail so people know it needs to be fixed?

pkg/flags/flags.go
```diff
@@ -362,6 +363,29 @@ func Validate() {
...
+
+// validateHealthCheckSourceCIDRs validates the override-health-check-src-cidrs flag
+func validateHealthCheckSourceCIDRs(cidrsFlag string) error {
+	if cidrsFlag == "" {
+		return nil
+	}
+
+	cidrs := strings.Split(cidrsFlag, ",")
+	for _, cidr := range cidrs {
+		cidr = strings.TrimSpace(cidr)
```
bowei: seems like duplicate check to the one outside

pkg/flags/flags_test.go
```diff
@@ -0,0 +1,179 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
bowei: date

pkg/firewalls/controller.go
panslava: fixed

pkg/firewalls/controller.go
panslava: I didn't want initially to change signature of constructor to add returning an error

Additionally, I have verification in flags.go too, which will make program fail fast

However, it firewall controller constructor is not used in that too many places, so I actually changed now the signature, to return an error, instead of logging, please take a look

pkg/flags/flags.go
panslava: I'd like to leave this check in both places, as those units could change independently, and it is nice to have an independent checks

pkg/flags/flags_test.go
panslava: fixed

panslava:
I reworked PR a little bit, consolidated validations into single package that is imported

pkg/validation/cidr_test.go
```diff
@@ -0,0 +1,297 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			err := ValidateCIDR(tc.input)
+			if (err != nil) != tc.wantError {
+				t.Errorf("ValidateCIDR(%q) expected error=%v, got error=%v", tc.input, tc.wantError, err != nil)
+			}
+		})
+	}
+}
+
+func TestValidateHealthCheckSourceCIDRs(t *testing.T) {
```
gauravkghildiyal: If this function is just a wrapper, any reason to have separate tests for it, compared to combining (not removign) the tests with `ParseHealthCheckSourceCIDRs`

panslava: Removed TestValidateHealthCheckSourceCIDRs, combined into TestParseHealthCheckSourceCIDRs

PR #2922, opened 2025-07-09
Read-Only mode for controllers

08volt:


pkg/instancegroups/controller.go
```diff
@@ -40,6 +40,9 @@ type Controller struct {
 
 	enableMultiSubnetCluster bool
 
+	//
```
TortillaZHawaii: missing comment?

08volt: fixed

08volt:
@FelipeYepez

pkg/l4lb/l4controller.go
```diff
@@ -290,6 +290,11 @@ func (l4c *L4Controller) processServiceCreateOrUpdate(service *v1.Service, svcLo
 		svcLogger.Info("Finished syncing L4 ILB service", "timeTaken", time.Since(startTime))
 	}()
 
+	if l4c.ctx.ReadOnlyMode {
+		svcLogger.Info("Skipping syncing L4 ILB service since the controller is in read-only mode", "service", service.Name)
+		return nil
+	}
+
```
FelipeYepez: You don't need this code since this function is only called from the sync() which has already this condition

pkg/l4lb/l4controller.go
```diff
@@ -520,6 +532,10 @@ func (l4c *L4Controller) sync(key string, svcLogger klog.Logger) error {
 }
 
 func (l4c *L4Controller) needsDeletion(svc *v1.Service) bool {
+	if l4c.ctx.ReadOnlyMode {
+		return false
+	}
+
```
FelipeYepez: You don't need this either, please determine the critical places to add the read only check

FelipeYepez (changes requested):
Please verify we add the read only mode only in the places we need it. Review all controllers and ask for review when done

pkg/flags/flags.go
```diff
@@ -352,6 +353,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableIPV6OnlyNEG, "enable-ipv6-only-neg", false, "Enable support for IPV6 Only NEG's.")
 	flag.StringVar(&F.MultiProjectOwnerLabelKey, "multi-project-owner-label-key", "multiproject.gke.io/owner", "The label key for multi-project owner, which is used to identify the owner of objects in multi-project mode.")
 	flag.StringVar(&F.OverrideHealthCheckSourceCIDRs, "override-health-check-src-cidrs", "", "Overrides the default source IP ranges used when configuring firewall rules to allow health check probes for L7 load balancers. Provide the ranges as a comma-separated list of CIDRs. Example: --override-health-check-src-cidrs=130.211.0.0/22,35.191.0.0/16")
+	flag.BoolVar(&F.ReadOnlyMode, "read-only", false, "When enabled, the controller runs in a read-only mode, preventing it from executing any mutating API calls (create, update, delete). This is useful for safely observing controller behavior without modifying resources.")
```
FelipeYepez: Can you rename it to `read-only-controllers` and specify that this applies to all controllers except for ingress controller

swetharepakula (commented):
Commenting on a closed PR, but we should convert this into metrics on the API calls instead of logs.

Also is there a read out of what this would look like? Will the controller keep erroring on its reads (because no writes have been made)?

swetharepakula:
Actually diving deeper, this doesn't feel like read only mode. This is more just logging when a sync is triggered

PR #2923, opened 2025-07-14
Add handler functions for watching updates to ServiceNetworkEndpointG…

mmamczur:
…roup CR for L4 services.

Whenever the ServiceNetworkEndpointGroup CR changes it's referred NEGs a resync for related services should be triggered.

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/backends"
 	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/forwardingrules"
 	l4metrics "k8s.io/ingress-gce/pkg/l4lb/metrics"
 	"k8s.io/ingress-gce/pkg/loadbalancers"
@@ -177,6 +178,17 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}, lo
 			}
 		},
 	})
+
+	enableMultiSubnetClusterPhase1 := flags.F.EnableMultiSubnetClusterPhase1
+	if enableMultiSubnetClusterPhase1 {
+		ctx.SvcNegInformer.AddEventHandler(&svcNEGEventHandler{
+			ServiceInformer: ctx.ServiceInformer,
+			svcQueue:        l4c.svcQueue,
+			svcFilterFunc:   isSubsettingILBService,
+			logger:          logger,
+		})
+		logger.V(3).Info("set up SvcNegInformer event handlers")
+	}
 
 	return l4c
 }
--- a/pkg/l4lb/l4netlbcontroller.go
+++ b/pkg/l4lb/l4netlbcontroller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/ingress-gce/pkg/annotations"
 	"k8s.io/ingress-gce/pkg/backends"
 	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/forwardingrules"
 	"k8s.io/ingress-gce/pkg/instancegroups"
 	l4metrics "k8s.io/ingress-gce/pkg/l4lb/metrics"
@@ -172,6 +173,17 @@ func NewL4NetLBController(
 			}
 		},
 	})
+
+	enableMultiSubnetClusterPhase1 := flags.F.EnableMultiSubnetClusterPhase1
+	if l4netLBc.enableNEGSupport && enableMultiSubnetClusterPhase1 {
+		ctx.SvcNegInformer.AddEventHandler(&svcNEGEventHandler{
+			ServiceInformer: ctx.ServiceInformer,
+			svcQueue:        l4netLBc.svcQueue,
+			svcFilterFunc:   isRBSNetLBServiceWithNEGs,
+			logger:          logger,
+		})
+		logger.V(3).Info("set up SvcNegInformer event handlers")
+	}
 
 	return l4netLBc
 }
--- a/pkg/l4lb/svcnegeventhandler.go
+++ b/pkg/l4lb/svcnegeventhandler.go
@@ -1,0 +1,154 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package l4lb
+
+import (
+	"reflect"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/annotations"
+	negv1beta1 "k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+// svcNEGEventHandler is a handler used to trigger service processing on update
+// of a related ServiceNetworkEndpointGroup.
+// It should be added as an event handler to the ServiceNetworkEndpointGroup informer.
+// Then it will enqueue a matching service to a given queue, if the service is matched by the svcFilterFunc.
+type svcNEGEventHandler struct {
+	logger          klog.Logger
+	ServiceInformer cache.SharedIndexInformer
+	svcQueue        utils.TaskQueue
+	svcFilterFunc   func(*v1.Service) bool
+}
+
+func (s *svcNEGEventHandler) OnAdd(obj interface{}, isInInitialList bool) {
+	// This function is directly executed by the event handler routine,
+	// it should not do heavy computation but rather focus on short checks and enqueuing.
+	svcNEG, ok := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
+	if !ok {
+		s.logger.Error(nil, "ServiceNetworkEndpointGroup invalid type, got %+v", obj)
+		return
+	}
+	svcKeys := getSvcOwnersOfSvcNEG(svcNEG, s.logger)
+	for _, svcKey := range svcKeys {
+		svcObj, exists, err := s.ServiceInformer.GetStore().GetByKey(svcKey)
+		if err != nil {
+			s.logger.Error(err, "ServiceNetworkEndpointGroup getting owner service", "svcKey", svcKey, "svcNeg", klog.KObj(svcNEG))
+			continue
+		}
+		if !exists {
+			s.logger.V(3).Info("ServiceNetworkEndpointGroup owner service not found", "svcKey", svcKey, "svcNeg", klog.KObj(svcNEG))
+			continue
+		}
+		svc, ok := svcObj.(*v1.Service)
+		if !ok {
+			continue
+		}
+		isHandledByTheL4Controller := s.svcFilterFunc(svc)
+		if !isHandledByTheL4Controller {
+			continue
+		}
+		svcLogger := s.logger.WithValues("serviceKey", svcKey)
+		svcLogger.V(3).Info("NEGs added for service, triggering resync")
+		s.svcQueue.Enqueue(svc)
+	}
+}
+func (s *svcNEGEventHandler) OnUpdate(old, cur interface{}) {
+	// This function is directly executed by the event handler routine,
+	// it should not do heavy computation but rather focus on short checks and enqueuing.
+	curSvcNEG, ok := cur.(*negv1beta1.ServiceNetworkEndpointGroup)
+	if !ok {
+		s.logger.Error(nil, "ServiceNetworkEndpointGroup invalid type, got %+v", cur)
+		return
+	}
+	oldSvcNEG, ok := old.(*negv1beta1.ServiceNetworkEndpointGroup)
+	if !ok {
+		s.logger.V(3).Info("ServiceNetworkEndpointGroup invalid type, got %+v", old)
+		return
+	}
+	svcKeys := getSvcOwnersOfSvcNEG(curSvcNEG, s.logger)
+	for _, svcKey := range svcKeys {
+		obj, exists, err := s.ServiceInformer.GetStore().GetByKey(svcKey)
+		if err != nil {
+			s.logger.Error(err, "ServiceNetworkEndpointGroup getting owner service", "svcKey", svcKey, "svcNeg", klog.KObj(curSvcNEG))
+			continue
+		}
+		if !exists {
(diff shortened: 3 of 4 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -178,6 +179,18 @@ func NewILBController(ctx *context.ControllerContext, stopCh <-chan struct{}, lo
...
+	if enableMultiSubnetClusterPhase1 {
+		ctx.SvcNegInformer.AddEventHandler(&svcNEGEventHandler{
+			ServiceInformer: ctx.ServiceInformer,
+			svcQueue:        l4c.svcQueue,
+			svcFilterFunc:   isSubsettingILBService,
+			logger:          logger,
+		})
+	}
+
+	logger.V(3).Info("set up SvcNegInformer event handlers")
```
TortillaZHawaii: I believe this log should have been inside the if (?)
```suggestion
	if enableMultiSubnetClusterPhase1 {
		ctx.SvcNegInformer.AddEventHandler(&svcNEGEventHandler{
			ServiceInformer: ctx.ServiceInformer,
			svcQueue:        l4c.svcQueue,
			svcFilterFunc:   isSubsettingILBService,
			logger:          logger,
		})
		logger.V(3).Info("set up SvcNegInformer event handlers")
	}
```

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -173,6 +174,16 @@ func NewL4NetLBController(
...
 	})
 
+	enableMultiSubnetClusterPhase1 := flags.F.EnableMultiSubnetClusterPhase1
+	if l4netLBc.enableNEGSupport && enableMultiSubnetClusterPhase1 {
+		ctx.SvcNegInformer.AddEventHandler(&svcNEGEventHandler{
+			ServiceInformer: ctx.ServiceInformer,
+			svcQueue:        l4netLBc.svcQueue,
+			svcFilterFunc:   isRBSNetLBServiceWithNEGs,
+			logger:          logger,
+		})
```
TortillaZHawaii, line 184: Let's add similar log line here to be consistent with ILB
```suggestion
        logger.V(3).Info("set up SvcNegInformer event handlers for NetLB")
		})
```

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+// Then it will enqueue a matching service to a given queue, if the service is matched by the svcFilterFunc.
+type svcNEGEventHandler struct {
+	logger          klog.Logger
+	ServiceInformer cache.SharedIndexInformer
+	svcQueue        utils.TaskQueue
+	svcFilterFunc   func(*v1.Service) bool
+}
+
+func (s *svcNEGEventHandler) OnAdd(obj interface{}, isInInitialList bool) {
+	svcNEG := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
```
TortillaZHawaii: Please add a check that this was actually ok, otherwise we might get a panic: https://go.dev/ref/spec#Type_assertions

I assume it will always be the correct type, however this is a simple check to do

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, svcKey := range svcKeys {
+		svcLogger := s.logger.WithValues("serviceKey", svcKey)
+		svcObj, exists, err := s.ServiceInformer.GetStore().GetByKey(svcKey)
+		if err != nil {
+			continue
+		}
+		if !exists {
+			continue
+		}
+		svc := svcObj.(*v1.Service)
```
TortillaZHawaii: Same here, please check for ok: https://go.dev/ref/spec#Type_assertions

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+		if !isHandledByTheL4Controller {
+			continue
+		}
+		svcLogger.V(3).Info("NEGs added for service, triggering resync")
+		s.svcQueue.Enqueue(svc)
+	}
+}
+func (s *svcNEGEventHandler) OnUpdate(old, cur interface{}) {
+	curSvcNEG := cur.(*negv1beta1.ServiceNetworkEndpointGroup)
+	oldSvcNEG := old.(*negv1beta1.ServiceNetworkEndpointGroup)
```
TortillaZHawaii: https://go.dev/ref/spec#Type_assertions

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, svcKey := range svcKeys {
+		svcLogger := s.logger.WithValues("serviceKey", svcKey)
+		obj, exists, err := s.ServiceInformer.GetStore().GetByKey(svcKey)
+		if err != nil {
+			continue
+		}
+		if !exists {
+			continue
+		}
+		svc := obj.(*v1.Service)
```
TortillaZHawaii: https://go.dev/ref/spec#Type_assertions

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+}
+func (s *svcNEGEventHandler) OnDelete(obj interface{}) {
+	// no need for any action. The service is likely being deleted anyway.
+}
+
+func getSvcOwnersOfSvcNEG(svcNEG *negv1beta1.ServiceNetworkEndpointGroup, logger klog.Logger) []string {
+	logger.V(3).Info("getSvcOwnerOfSvcNEG", "ownersLen", len(svcNEG.OwnerReferences))
+	var resultKeys []string
+	for _, ownersRef := range svcNEG.OwnerReferences {
+		logger.V(3).Info("getSvcOwnerOfSvcNEG.OwnerReferences[0]", "kind", ownersRef.Kind, "name", ownersRef.Name)
```
TortillaZHawaii: why is the [0] here?

pkg/l4lb/svcnegeventhandler_test.go
```diff
@@ -0,0 +1,556 @@
...
+
+	testCases := []struct {
+		name           string
+		svcNEG         *negv1beta1.ServiceNetworkEndpointGroup
+		oldSvcNEG      *negv1beta1.ServiceNetworkEndpointGroup
+		services       []*v1.Service
+		expectEnqueues int
+	}{
+		{
+			name:   "SvcNEG_unchanged_dont_equeue",
```
TortillaZHawaii:
```suggestion
			name:   "SvcNEG_unchanged_dont_enqueue",
```

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		if !exists {
+			continue
+		}
+		svc := obj.(*v1.Service)
+		isHandledByTheL4Controller := s.svcFilterFunc(svc)
+		if !isHandledByTheL4Controller {
+			continue
+		}
+		if !reflect.DeepEqual(curSvcNEG.Status.NetworkEndpointGroups, oldSvcNEG.Status.NetworkEndpointGroups) {
```
TortillaZHawaii: Those two are slices, do we care about the order here? If not it might be more safe to use `cmp.Equal` with `cmpopts.SortSlices` based on Id or SelfLink.

pkg/l4lb/svcnegeventhandler_test.go
```diff
@@ -0,0 +1,556 @@
...
+
+	if len(svcKeys) != 2 {
+		t.Fatalf("getSvcOwnersOfSvcNEG returned unexpected number of results, wanted 2 but got list=%v", svcKeys)
+	}
+	if svcKeys[0] != "testNamespace/testService" {
+		t.Errorf("getSvcOwnersOfSvcNEG svcKeys[0], want=%s got=%s", "testNamespace/testService", svcKeys[0])
+	}
+	if svcKeys[1] != "testNamespace/testService2" {
+		t.Errorf("getSvcOwnersOfSvcNEG svcKeys[1], want=%s got=%s", "testNamespace/testService2", svcKeys[1])
+	}
```
TortillaZHawaii: nit for future, no change needed

I think this stuff would be more readable and quicker to write if we just used cmp.Diff

pkg/l4lb/svcnegeventhandler_test.go
```diff
@@ -0,0 +1,556 @@
...
+			NetworkEndpointGroups: []negv1beta1.NegObjectReference{},
+		},
+	}
+
+	testCases := []struct {
+		name           string
+		svcNEG         *negv1beta1.ServiceNetworkEndpointGroup
+		oldSvcNEG      *negv1beta1.ServiceNetworkEndpointGroup
+		services       []*v1.Service
+		expectEnqueues int
```
TortillaZHawaii, line 221: do we want to test multiple enqueues?

TortillaZHawaii (changes requested):

pkg/l4lb/l4controller.go
mmamczur: moved

pkg/l4lb/svcnegeventhandler.go
mmamczur: done

pkg/l4lb/svcnegeventhandler_test.go
mmamczur: It's a bit artificial case since there should be only 1 owner per SvcNEG but I added a test case just for coverage's sake

pkg/l4lb/svcnegeventhandler.go
mmamczur: yeah I was thinking of doing something similar, I changed this to follow your suggestion
should reduce unnecessary enqueues

pkg/l4lb/svcnegeventhandler.go
mmamczur: it shouldn't be here, removed

pkg/l4lb/svcnegeventhandler_test.go
TortillaZHawaii: kk thanks

TortillaZHawaii:
If you don't need another review feel free to unhold
/hold

/lgtm

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,140 @@
...
+	svcNEG, ok := obj.(*negv1beta1.ServiceNetworkEndpointGroup)
+	if !ok {
+		s.logger.V(3).Info("ServiceNetworkEndpointGroup invalid type, got %+v", obj)
+		return
+	}
+	svcKeys := getSvcOwnersOfSvcNEG(svcNEG, s.logger)
+	for _, svcKey := range svcKeys {
+		svcObj, exists, err := s.ServiceInformer.GetStore().GetByKey(svcKey)
+		if err != nil {
+			s.logger.Error(err, "ServiceNetworkEndpointGroup getting owner service", "svcKey", svcKey, "svcNeg", svcNEG.Namespace+"/"+svcNEG.Name)
```
gauravkghildiyal: nit: (Non blocking) You can simplify `curSvcNEG.Namespace+"/"+curSvcNEG.Name` to `klog.KObj(curSvcNEG)`

pkg/l4lb/svcnegeventhandler.go
gauravkghildiyal: Sorry that I don't present a better alternative, but I would like to highlight this note which I recall from https://pkg.go.dev/github.com/google/go-cmp/cmp

_"It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal."_

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,140 @@
...
+// It should be added as an event handler to the ServiceNetworkEndpointGroup informer.
+// Then it will enqueue a matching service to a given queue, if the service is matched by the svcFilterFunc.
+type svcNEGEventHandler struct {
+	logger          klog.Logger
+	ServiceInformer cache.SharedIndexInformer
+	svcQueue        utils.TaskQueue
+	svcFilterFunc   func(*v1.Service) bool
+}
+
+func (s *svcNEGEventHandler) OnAdd(obj interface{}, isInInitialList bool) {
```
gauravkghildiyal, line 42: For any future modifications, it'd be helpful to drop in a note here that because these functions are directly executed by the event handler, don't add complex logic other than some simple enqueuing.

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		svcLogger := s.logger.WithValues("serviceKey", svcKey)
+		svcLogger.V(3).Info("NEGs added for service, triggering resync")
+		s.svcQueue.Enqueue(svc)
+	}
+}
+func (s *svcNEGEventHandler) OnUpdate(old, cur interface{}) {
+	curSvcNEG, ok := cur.(*negv1beta1.ServiceNetworkEndpointGroup)
+	if !ok {
+		s.logger.V(3).Info("ServiceNetworkEndpointGroup invalid type, got %+v", cur)
```
gauravkghildiyal: nit: (Non blocking) Since this is a log which we ideally don't expect to show up, maybe lets log this as an error.

pkg/l4lb/svcnegeventhandler.go
```diff
@@ -0,0 +1,140 @@
...
+}
+func (s *svcNEGEventHandler) OnDelete(obj interface{}) {
+	// no need for any action. The service is likely being deleted anyway.
+}
+
+func getSvcOwnersOfSvcNEG(svcNEG *negv1beta1.ServiceNetworkEndpointGroup, logger klog.Logger) []string {
+	logger.V(3).Info("getSvcOwnerOfSvcNEG", "ownersLen", len(svcNEG.OwnerReferences))
+	var resultKeys []string
+	for _, ownersRef := range svcNEG.OwnerReferences {
+		if ownersRef.Kind != "Service" {
```
gauravkghildiyal, line 124: Thank you for being thoughtful about this. Since we are anyways trying to check the Kind, maybe lets verify the Group (empty) and Version (v1) as well?

gauravkghildiyal (commented):
Looking good. Just some minor stuff. Please go ahead with submitting even without my approval after resolving the few things.

/lgtm

pkg/l4lb/svcnegeventhandler.go
mmamczur: looks like there is no Group in owner ref somehow. For version, probably we shouldn't expect anything else than v1 but then if v1beta or whatever showed up there we could just enqueue, at worst it won't do anything.

pkg/l4lb/svcnegeventhandler.go
mmamczur: you are right, thanks for pointing out the cmp thing since it's not obvious. I convert manually to a map and then do a reflect deep equal, does what I wanted it to do.

pkg/l4lb/svcnegeventhandler.go
mmamczur: thanks, done!

pkg/l4lb/l4netlbcontroller.go
mmamczur: done

pkg/l4lb/svcnegeventhandler.go
mmamczur: done

pkg/l4lb/svcnegeventhandler.go
mmamczur: done

pkg/l4lb/svcnegeventhandler.go
gauravkghildiyal: I don't want us to spend time changing this, but for future references, the "apiVersion" field in ownerRef YAML that you may be seeing is actually <GROUP>/<VERSION> (like is the case for the top level apiVersion). For Service, it's "v1" because the Group is "empty" (i.e. core) but the fact that it's empty is still a value and can be checked through something like `GroupVersion.Group`

PR #2927, opened 2025-07-17
This PR introduces three new binaries to be built inside cmd folder

rino-gaelico:
The purpose is to create three new binaries that willl be subsituting the monolithic binary that existed until today (e2e-test). The new binaries are ingress-controller-e2e-test, neg-e2e-test, and psc-e2e-test.

In particular each new folder hosts: the code of the specific tests for the respective controller; an entry point for the container/binary; an upgrade folder with the targets for the upgrade_test.go.

In addition three new Docker templates are introduced to build the three images.

Finally Makefile is updated with the new binaries.

The old e2e is preserved and it will be deleted later once the new images are tested and proven to work correctly within the CI/CD pipeline.

**Note** that some of the files inside cmd/ingress-controller-e2e-test will be renamed and the git history will be restored once we get rid of the cmd/e2e-test folder


rino-gaelico:
**Note for reviewers**

At the moment the files in cmd/ingress-controller-e2e-test appear as new files. This is not ideal because the blame history data would be lost. 
To fix this, in the following PR when e2e-test will be finally substituted by ingress-controller it will be important to rename those files and preserve the historical blame data. 
This intermediary step is required so that a new image (ingress-controller) can be created and tested by the new orchestrators.

PR #2929, opened 2025-07-22
Fix MT NEG controller panic, caused by AddEventHandler on a nil nodeTopologyInformer

panslava:


bowei:
no test?

PR #2937, opened 2025-07-24
Calculate new subsetPerZone with the exclusion of existing endpoints (map version)

TortillaZHawaii:
This change is there to fix edge cases when the number of nodes in the zone fluctuates while being nearly equal. Those fluctations cause the `pickSubsetMinRemovals` zone ordering to change - causing constant detach/attach of an endpoint. This "moving" endpoint then causes constant connection draining for ~1/25 of traffic.

The fix is to keep existing endpoints as long as the node/subnet pair attached to the endpoint still exists, and only consider unused endpoints for `pickSubsetMinRemovals`.


TortillaZHawaii:
/close 

#2942 was a better option

PR #2948, opened 2025-08-19
Gracefully handle empty default backend service

FelipeYepez:
This commit prevents controllers initialization from crashing and controllers performing unnecessary work by ensuring the DefaultBackendServicePort function returns an empty utils.ServicePort when the Ingress controller is disabled. This change modifies the core behavior of the DefaultBackendServicePort function, making it resilient to cases where the default backend service does not exist when the Ingress controller is not running. It also ensures that all controllers that rely on this function are now able to gracefully handle the empty utils.ServicePort, preventing panics and redundant processing.

### DIFF
--- a/cmd/glbc/app/init.go
+++ b/cmd/glbc/app/init.go
@@ -35,6 +35,11 @@ import (
 // DefaultBackendServicePort returns the ServicePort which will be
 // used as the default backend for load balancers.
 func DefaultBackendServicePort(kubeClient kubernetes.Interface, logger klog.Logger) utils.ServicePort {
+	if !flags.F.RunIngressController && !flags.F.EnableNEGsForIngress {
+		noDefaultBackendServicePort := utils.ServicePort{}
+		return noDefaultBackendServicePort
+	}
+
 	if flags.F.DefaultSvc == "" {
 		klog.Fatalf("Please specify --default-backend-service")
 	}
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -113,8 +113,11 @@ type Controller struct {
 	// runL4ForNetLB indicates if the controller can create NEGs for L4 NetLB services.
 	runL4ForNetLB bool
 
-	// readOnlyMode indicates wheter or not the controller will run in read only mode
+	// readOnlyMode indicates whether or not the controller will run in read only mode
 	readOnlyMode bool
+
+	// enableNEGsForIngress indicates whether the NEG controller will create NEGs for Ingress services
+	enableNEGsForIngress bool
 
 	stopCh <-chan struct{}
 	logger klog.Logger
@@ -153,6 +156,7 @@ func NewController(
 	enableIngressRegionalExternal bool,
 	runL4ForNetLB bool,
 	readOnlyMode bool,
+	enableNEGsForIngress bool,
 	stopCh <-chan struct{},
 	logger klog.Logger,
 ) (*Controller, error) {
@@ -253,6 +257,7 @@ func NewController(
 		enableMultiSubnetClusterPhase1: enableMultiSubnetClusterPhase1,
 		runL4ForNetLB:                  runL4ForNetLB,
 		readOnlyMode:                   readOnlyMode,
+		enableNEGsForIngress:           enableNEGsForIngress,
 		stopCh:                         stopCh,
 		logger:                         logger,
 	}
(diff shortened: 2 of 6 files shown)


pkg/neg/controller.go
```diff
@@ -727,6 +727,11 @@ func (c *Controller) netLBServiceNeedsNEG(service *apiv1.Service, networkInfo *n
 // it is passed to the controller via a command line flag.
 // Additionally, supporting NEGs for default backends is only for L7-ILB
 func (c *Controller) mergeDefaultBackendServicePortInfoMap(key string, service *apiv1.Service, portInfoMap negtypes.PortInfoMap, networkInfo *network.NetworkInfo) error {
+	emptyDefaultBackendServicePort := utils.ServicePort{}
+	if c.defaultBackendService.ID.Service.String() == emptyDefaultBackendServicePort.ID.Service.String() {
```
mmamczur: is the value of `emptyDefaultBackendServicePort.ID.Service.String()` emty here? we init an empty struct above

FelipeYepez: The function returns "/" when empty.

type NamespacedName struct {
	Namespace string
	Name      string
}

const (
	Separator = '/'
)

// String returns the general purpose string representation
func (n NamespacedName) String() string {
	return n.Namespace + string(Separator) + n.Name
}

mmamczur: I'm wondering if this is going to work

so this part here executes for NEG controller. For the NEG controller the flag RunIngress will always be false

maybe we need another flag for the NEG controller like create_NEGs_for_ingress? if it's false then skip the whole ingress related logic.

FelipeYepez: True, thanks. Adding another flag which will use the same logic to gate NEGs for Ingress

pkg/neg/controller.go
```diff
@@ -755,6 +755,11 @@ func (c *Controller) netLBServiceNeedsNEG(service *apiv1.Service, networkInfo *n
 // it is passed to the controller via a command line flag.
 // Additionally, supporting NEGs for default backends is only for L7-ILB
 func (c *Controller) mergeDefaultBackendServicePortInfoMap(key string, service *apiv1.Service, portInfoMap negtypes.PortInfoMap, networkInfo *network.NetworkInfo) error {
```
mmamczur, line 768: we should skip running this based on the flag, don't do the comparison for the empty name

also you need to skip mergeIngressPortInfo() if EnableNEGsForIngress is false

mmamczur: best to do something like `if !EnableNEGsForIngress { return nil }` in here and in `mergeIngressPortInfo()`

FelipeYepez: In a separate PR I will disable the NEG functionality for ingress using the `EnableNEGsForIngress` flag

FelipeYepez: Added the changes to this PR

cmd/glbc/app/init.go
```diff
@@ -35,6 +35,11 @@ import (
 // DefaultBackendServicePort returns the ServicePort which will be
 // used as the default backend for load balancers.
 func DefaultBackendServicePort(kubeClient kubernetes.Interface, logger klog.Logger) utils.ServicePort {
+	if !flags.F.RunIngressController && !flags.F.EnableNEGsForIngress {
```
mmamczur, line 38: remember that with this setup you will have to explicitly set EnableNEGsForIngress to false for the L4 controllers (normally you'd not even set that flag when NEG controller is disabled adn the default is true)

FelipeYepez: Yes, will be set to false. The logic will only be applied from the NEG controller

pkg/neg/controller_test.go
```diff
@@ -2304,3 +2305,174 @@ func newTestServiceCustomNamedNeg(c *Controller, negSvcPorts map[int32]string, i
...
+	// Verify syncers are stopped after the service is updated
+	controller.serviceLister.Update(newTestService(controller, false, []int32{}))
+	err = controller.processService(utils.ServiceKeyFunc(testServiceNamespace, testServiceName))
+	if err != nil {
+		t.Fatalf("Failed to process service: %v", err)
+	}
+	validateSyncers(t, controller, 3, true)
+}
+
+func TestMergeDefaultBack(t *testing.T) {
```
mmamczur: is this only for cases where ingress is disabled? best to name it so that it's clear

pkg/neg/controller_test.go
```diff
@@ -2304,3 +2305,174 @@ func newTestServiceCustomNamedNeg(c *Controller, negSvcPorts map[int32]string, i
...
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+
+			controller, err := newTestController(fake.NewSimpleClientset())
+			if err != nil {
+				t.Fatalf("failed to create test controller %s", err)
+			}
+			controller.enableNEGsForIngress = false
+			controller.enableIngressRegionalExternal = true
```
mmamczur, line 2457: why do we enable `controller.enableIngressRegionalExternal = true`?

FelipeYepez: It is used to evaluate the third case with `annotations.GceL7XLBRegionalIngressClass`

pkg/neg/controller_test.go
FelipeYepez: Good catch, thanks

pkg/neg/controller.go
```diff
@@ -921,8 +936,9 @@ func (c *Controller) enqueueIngressServices(ing *v1.Ingress) {
 		c.enqueueService(cache.ExplicitKey(key))
 	}
 
-	// enqueue default backend service
-	if ing.Spec.DefaultBackend == nil {
+	// Enqueue default backend service if it is not empty.
```
mmamczur, line 939: this is probably not needed if you already do the check at the beginning of this function

PR #2949, opened 2025-08-19
PSC Controller: avoid odd patch requests

midavadim:
ConsumerRejectLists should be set to nil for needsUpdate() if there is no data to avoid odd patch requests.

When needUpdate() compare gceSvcAttachment and existingSA, ConsumerRejectLists fields are different:
 nil != [] 
 
 so it caused odd Patch request for PSC SA.

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -699,6 +699,10 @@ func needsUpdate(existingSA, desiredSA *ga.ServiceAttachment) (bool, error) {
 	// Set region to avoid selflink mismatches
 	desiredCopy.Region = existingSA.Region
 
+	// convertRejectList should be a nil (but not empty list) if there is no data
+	if len(desiredCopy.ConsumerRejectLists) == 0 {
+		desiredCopy.ConsumerRejectLists = nil
+	}
 	return !reflect.DeepEqual(desiredCopy, existingSA), nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/psc/controller.go
```diff
@@ -340,7 +340,7 @@ func (c *Controller) processServiceAttachment(key string) error {
 	gceSvcAttachment.Description = desc.String()
 	gceSvcAttachment.EnableProxyProtocol = updatedCR.Spec.ProxyProtocol
 	gceSvcAttachment.ConsumerAcceptLists = convertAllowList(updatedCR.Spec)
-	gceSvcAttachment.ConsumerRejectLists = updatedCR.Spec.ConsumerRejectList
+	gceSvcAttachment.ConsumerRejectLists = convertRejectList(updatedCR.Spec)
```
midavadim: updatedCR.Spec.ConsumerRejectList contained empty slice which goes to gceSvcAttachment.

Then when needUpdate compare gceSvcAttachment and existingSA nil != [] so it caused odd Patch request for PSC SA.

mmamczur: I'd consider putting this inside `needsUpdate()` at the very bottom. This setup here is done purely for comparison logic and that kind of stuff belongs down there I think.

PR #2950, opened 2025-08-20
Added ProviderConfigId label and refactored NEG Controller metrics emitted to use new NegMetrics struct 

priyapande:
To allow providerConfigId to be set for multi-project use-cases, added a new NegMetrics struct that registers to the existing metrics methods so we can use it as an additional label in all our metrics to support per NEG controller metrics.


priyapande:
[metrics-added.txt](https://github.com/user-attachments/files/21904314/metrics-added.txt)
prometheus metrics with provider config ID labels tested locally

PR #2952, opened 2025-08-20
Added ProviderConfigId label for NEG metrics

priyapande:
To allow providerConfigId to be set for multi-project use-cases, added a new NegMetrics struct that registers to the existing metrics methods so we can use it as an additional label in all our metrics to support per NEG controller metrics.


cmd/glbc/main.go
```diff
@@ -711,6 +717,14 @@ func createNEGController(ctx *ingctx.ControllerContext, systemHealth *systemheal
 		adapter = ctx.Cloud
 	}
 
+	// register NEG prometheus metrics
+	metrics.RegisterMetrics()
```
panslava: I think this can be a tricky bug

we need to move RegisterMetrics to main func I think

Looks like we now register NEG metrics only inside createNEGController. 

When --enable-neg-controller=false, we still build L4/PSC controllers that call negtypes.NewAdapter(...) and emit NEG GCE/K8s request/error metrics, but the collectors are no longer registered so those exports disappear. 

Could we register both metrics.RegisterMetrics() and syncMetrics.RegisterMetrics() once up front in main() (before we branch on EnableNEGController) and drop the per-controller registrations? That keeps the metrics live even when the NEG controller itself is disabled.

priyapande: Thanks for the catch, moved it up in the main.

pkg/multiproject/neg/neg.go
```diff
@@ -28,6 +30,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
+func init() {
```
panslava: I don't think this needed anymore as we register in main

pkg/multiproject/start/start.go
```diff
@@ -31,6 +34,12 @@ import (
 
 const multiProjectLeaderElectionLockName = "ingress-gce-multi-project-lock"
 
+func init() {
```
panslava: I don't think this needed anymore as we register in main

pkg/neg/metrics/metrics.go
panslava: Sorry, just a question, that might not be a bug -- why this and the next function are still global, not a struct methods?

pkg/neg/metrics/metrics.go
```diff
@@ -280,29 +280,37 @@ func RegisterMetrics() {
 	})
 }
 
+// NegMetrics is a struct that holds the labels for multi-project NEG metrics
```
panslava: I think we should remove this comment, cause for now we don't store any labels

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -53,6 +54,7 @@ func TestLocalGetEndpointSet(t *testing.T) {
 	prevFlag := flags.F.EnableMultiSubnetCluster
 	defer func() { flags.F.EnableMultiSubnetCluster = prevFlag }()
 	flags.F.EnableMultiSubnetCluster = false
+	negMetrics := metrics.NewNegMetrics()
```
panslava: I think we can just initialize those metrics inside t.Run()

Cause now metrics are shared between testcases, and it is a bit unusual to store negMetrics in tc struct

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -211,6 +220,7 @@ func TestClusterGetEndpointSet(t *testing.T) {
...
```
panslava: Same comment as in previous test

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -347,6 +364,7 @@ func TestClusterGetEndpointSet(t *testing.T) {
 
 // TestClusterWantedNEGsCount verifies logic for
 func TestClusterWantedNEGsCount(t *testing.T) {
+	negMetrics := metrics.NewNegMetrics()
```
panslava: Same comment as in previous test

panslava (changes requested):
I might missed some places -- in unit tests move metrics struct initialization inside t.Run so don't store it in testCases struct

pkg/multiproject/start/start.go
priyapande: yes, I had removed it but I think some rebasing went wrong. Fixed it. Thanks

pkg/neg/metrics/metrics.go
priyapande: yes ofcourse -- when I was making the change for labels for provider config, I was not sure if these are relevant metrics for NEG to add the provider config ID label. For now, to maintain consistency across the NEG metrics I have registered these to the struct too.

panslava:
/lgtm

great job, thank you!

PR #2954, opened 2025-08-22
Prevent L4 ILB controller race conditions on service creation

FelipeYepez:
Introduces a wait and re-fetch mechanism when a service has no L4 ILB finalizers (v1 or v2) through a flag `L4ILBLegacyHeadStartTime`. This prevents race conditions between controllers during initial service creation, ensuring the L4 controller processes the most recent service state and avoids conflicting operations. This to prevent race during cluster update when enabling GKE subsetting.

This will only be applied to regional clusters since zonal clusters don't have the issue.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -148,6 +148,7 @@ var F = struct {
 	ManageL4LBLogging                         bool
 	EnableNEGsForIngress                      bool
 	EnableIPv6OnlyL4                          bool
+	L4ILBLegacyHeadStartTime                  time.Duration
 
 	// ===============================
 	// DEPRECATED FLAGS
@@ -360,6 +361,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.ReadOnlyMode, "read-only-controllers", false, "When enabled, this flag runs the IG, NEG, L4 ILB, and L4 NetLB controllers in a read-only mode. This prevents them from executing any mutating API calls (e.g., create, update, delete), allowing you to safely observe controller behavior without modifying resources. The Ingress controller is exempt from this mode.")
 	flag.BoolVar(&F.EnableNEGsForIngress, "enable-negs-for-ingress", true, "Allow the NEG controller to create NEGs for Ingress services.")
 	flag.BoolVar(&F.EnableIPv6OnlyL4, "enable-ipv6-only-l4", false, "Enables IPv6-only mode for the L4 ILB and NetLB controllers, disabling all IPv4-related logic and resource management.")
+	flag.DurationVar(&F.L4ILBLegacyHeadStartTime, "prevent-legacy-race-l4-ilb", 0*time.Second, "Delay before processing new L4 ILB services without existing finalizers. This gives the legacy controller a head start to claim the service, preventing a race condition upon service creation.")
 }
 
 func Validate() {
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -288,9 +288,16 @@ func (l4c *L4Controller) shouldProcessService(service *v1.Service, svcLogger klo
 			return false
 		}
 	}
+
+	// Prevent race condition with legacy controller
+	service = l4c.handleCreationRace(service, svcLogger)
+	if service == nil {
+		return false
+	}
+
 	// skip services that are being handled by the legacy service controller.
 	if utils.IsLegacyL4ILBService(service) {
-		svcLogger.Info("Ignoring update for service managed by service controller")
+		svcLogger.Info("Ignoring update for service managed by service controller, has finalizer v1")
 		return false
 	}
 	frName := utils.LegacyForwardingRuleName(service)
@@ -716,4 +723,34 @@ func (l4c *L4Controller) publishMetrics(result *loadbalancers.L4ILBSyncResult, n
 	default:
 		svcLogger.Info("Unknown sync type, skipping metrics for service", "syncType", result.SyncType)
 	}
+}
+
+// handleCreationRace prevents a race condition between the legacy and new L4 ILB controllers
+// when a service is created, ensuring the L4 controller processes the most recent service
+// state and avoids conflicting operations..
+func (l4c *L4Controller) handleCreationRace(service *v1.Service, svcLogger klog.Logger) *v1.Service {
+	hasLegacyILBFinalizer := utils.HasLegacyL4ILBFinalizerV1(service)
+	hasILBFinalizerV2 := utils.HasL4ILBFinalizerV2(service)
+	l4ILBLegacyHeadStartTime := flags.F.L4ILBLegacyHeadStartTime
+
+	// Prevent controllers race on creation
+	if !hasLegacyILBFinalizer && !hasILBFinalizerV2 && l4ILBLegacyHeadStartTime > 0*time.Second {
+		svcLogger.Info("Service has no finalizers, waiting %d seconds to prevent controllers race on creation.", l4ILBLegacyHeadStartTime/time.Second)
+		time.Sleep(l4ILBLegacyHeadStartTime)
+
+		// Get current service from store, so we can verify most recent state.
+		svcKey := utils.ServiceKeyFunc(service.Namespace, service.Name)
+		svc, exists, err := l4c.ctx.Services().GetByKey(svcKey)
+		if err != nil {
+			svcLogger.Info("Could not get service from store, using existing one, error: ", err)
+		} else if exists {
+			service = svc
+			svcLogger.Info("finalizer Found service in informer store after wait, using it.")
+		} else {
+			// Service might have been deleted during the wait, return to avoid processing a non-existing service.
+			svcLogger.Info("Service not found in informer store after wait, ignoring processing.")
+			return nil
+		}
+	}
+	return service
 }
--- a/pkg/l4lb/l4controller_test.go
+++ b/pkg/l4lb/l4controller_test.go
@@ -46,6 +46,7 @@ import (
 	negv1beta1 "k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/flags"
 	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned/fake"
 	"k8s.io/ingress-gce/pkg/test"
 	"k8s.io/ingress-gce/pkg/utils/common"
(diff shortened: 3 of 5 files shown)


pkg/l4lb/l4controller.go
```diff
@@ -288,9 +288,33 @@ func (l4c *L4Controller) shouldProcessService(service *v1.Service, svcLogger klo
...
+			svcLogger.Info("Could not get service from store, using existing one, error: ", err)
+		} else if exists {
+			service = svc
+			svcLogger.Info("finalizer Found service in informer store after wait, using it.")
+		} else {
+			// Service might have been deleted during the wait, return to avoid processing a non-existing service.
+			svcLogger.Info("Service not found in informer store after wait, ignoring processing.")
+			return false
+		}
+	}
```
TortillaZHawaii: please move this into a separate func

pkg/l4lb/l4controller.go
```diff
@@ -288,9 +288,33 @@ func (l4c *L4Controller) shouldProcessService(service *v1.Service, svcLogger klo
...
 		}
 	}
+
+	hasLegacyILBFinalizer := utils.HasLegacyL4ILBFinalizerV1(service)
+	hasILBFinalizerV2 := utils.HasL4ILBFinalizerV2(service)
+
+	// Prevent controllers race on creation
+	if !hasLegacyILBFinalizer && !hasILBFinalizerV2 && flags.F.EnableL4ILBPreventLegacyRace {
+		svcLogger.Info("Service has no finalizers, waiting 3 seconds to prevent controllers race on creation.")
+		time.Sleep(3 * time.Second)
```
TortillaZHawaii: Can we make this a parameter instead? It's hard for me to understand why 3s was picked, but fine tuning it later will be more difficult if its hardcoded. I suggest instead of having `EnableL4ILBPreventLegacyRace bool` flag to have a `L4ILBLegacyHeadStartTime *time.Time`, or something better sounding. If it's `nil` we would treat it as current `false`, otherwise we will use it in `time.Sleep(...)`

TortillaZHawaii: Also the wording is not fully accurate. We cannot prevent them from racing, we can only give the CCM a head start.

pkg/l4lb/l4controller_test.go
```diff
@@ -1211,3 +1212,62 @@ func createLegacyForwardingRule(t *testing.T, svc *api_v1.Service, cloud *gce.Cl
...
+	origEnableL4ILBPreventLegacyRace := flags.F.EnableL4ILBPreventLegacyRace
+	flags.F.EnableL4ILBPreventLegacyRace = true
+	defer func() {
+		flags.F.EnableL4ILBPreventLegacyRace = origEnableL4ILBPreventLegacyRace
+	}()
+
+	backoff := retry.DefaultRetry
+	// Increase the duration including the wait before fetching the service.
+	backoff.Duration = 5 * time.Second
+	for _, onlyLocal := range []bool{true, false} {
```
TortillaZHawaii: what is `onlyLocal` used for?

pkg/l4lb/l4controller_test.go
```diff
@@ -1211,3 +1212,62 @@ func createLegacyForwardingRule(t *testing.T, svc *api_v1.Service, cloud *gce.Cl
...
+		t.Run(fmt.Sprintf("L4 with LocalMode=%v", onlyLocal), func(t *testing.T) {
+			l4c := newServiceController(t, newFakeGCE(), false)
+			prevMetrics, err := test.GetL4ILBLatencyMetric()
+			if err != nil {
+				t.Errorf("Error getting L4 ILB latency metrics %v", err)
+			}
+			go l4c.Run()
```
TortillaZHawaii: what will stop and clean up the controller? newServiceController doesn't send anything to stop channel.

pkg/l4lb/l4controller_test.go
```diff
@@ -1211,3 +1212,62 @@ func createLegacyForwardingRule(t *testing.T, svc *api_v1.Service, cloud *gce.Cl
...
+				addILBService(l4c, newSvc)
+				// add the NEG so that link to backendService works.
+				addNEGAndSvcNegL4Controller(l4c, newSvc)
+				l4c.svcQueue.Enqueue(newSvc)
+			}
+			if err := retry.OnError(backoff, func(error) bool { return true }, func() error {
+				for _, name := range svcNames {
+					newSvc, err := l4c.client.CoreV1().Services(testNs).Get(context2.TODO(), name, v1.GetOptions{})
+					if err != nil {
+						return fmt.Errorf("Failed to lookup service %s, err: %v", name, err)
```
TortillaZHawaii: go errors should start with a lowercase

pkg/l4lb/l4controller_test.go
```diff
@@ -1211,3 +1212,62 @@ func createLegacyForwardingRule(t *testing.T, svc *api_v1.Service, cloud *gce.Cl
 		t.Errorf("Failed to create fake forwarding rule %s, err %v", frName, err)
 	}
 }
+
+func TestMultipleServicesProcessingWithPreventLegacyRace(t *testing.T) {
```
TortillaZHawaii: I don't see how this test verifies the newly added logic. Can you run the test with coverage and implement test cases for branches that weren't run before?

pkg/l4lb/l4controller_test.go
FelipeYepez: Fixed, thanks. It adds externalTrafficPolicy "local" to the service

pkg/l4lb/l4controller.go
FelipeYepez: Done

pkg/l4lb/l4controller.go
FelipeYepez: Done

pkg/l4lb/l4controller_test.go
FelipeYepez: Thanks

pkg/l4lb/l4controller_test.go
FelipeYepez: Modified the test to use finalizers and verify that there's no extra wait when the service already exists.

pkg/l4lb/l4controller_test.go
FelipeYepez: Good catch, modified for existing tests as well

pkg/utils/utils.go
```diff
@@ -739,6 +739,11 @@ func IsLegacyL4ILBService(svc *api_v1.Service) bool {
 	if svc.Spec.LoadBalancerClass != nil {
 		return annotations.HasLoadBalancerClass(svc, annotations.LegacyRegionalInternalLoadBalancerClass)
 	}
+	return HasLegacyL4ILBFinalizerV1(svc)
+}
+
+// HasLegacyL4ILBFinalizerV1 returns true if the given Service has ILBFinalizerV1
+func HasLegacyL4ILBFinalizerV1(svc *api_v1.Service) bool {
```
TortillaZHawaii, line 746: test this please

pkg/l4lb/l4controller_test.go
```diff
@@ -546,7 +545,8 @@ func TestProcessMultipleServices(t *testing.T) {
 	backoff.Duration = 3 * time.Second
 	for _, onlyLocal := range []bool{true, false} {
 		t.Run(fmt.Sprintf("L4 with LocalMode=%v", onlyLocal), func(t *testing.T) {
-			l4c := newServiceController(t, newFakeGCE(), false)
+			l4c, stopCh := newServiceController(t, newFakeGCE(), false)
+			defer close(stopCh)
```
TortillaZHawaii, line 549: why defer is only here?

pkg/flags/flags.go
```diff
@@ -356,6 +358,8 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.StringVar(&F.OverrideHealthCheckSourceCIDRs, "override-health-check-src-cidrs", "", "Overrides the default source IP ranges used when configuring firewall rules to allow health check probes for L7 load balancers. Provide the ranges as a comma-separated list of CIDRs. Example: --override-health-check-src-cidrs=130.211.0.0/22,35.191.0.0/16")
 	flag.BoolVar(&F.ManageL4LBLogging, "manage-l4lb-logging", false, "Manage L4 ILB/NetLB logging.")
 	flag.BoolVar(&F.ReadOnlyMode, "read-only-controllers", false, "When enabled, this flag runs the IG, NEG, L4 ILB, and L4 NetLB controllers in a read-only mode. This prevents them from executing any mutating API calls (e.g., create, update, delete), allowing you to safely observe controller behavior without modifying resources. The Ingress controller is exempt from this mode.")
+
+	flag.DurationVar(&F.L4ILBLegacyHeadStartTime, "prevent-legacy-race-l4-ilb", 0*time.Second, "When enabled, this flag fetches the last state of the ILB after a delay to prevent controllers race when the service is created.")
```
TortillaZHawaii: please update the flag name and description as well

pkg/l4lb/l4controller_test.go
```diff
@@ -1211,3 +1212,98 @@ func createLegacyForwardingRule(t *testing.T, svc *api_v1.Service, cloud *gce.Cl
...
+		desc            string
+		backoffDuration time.Duration
+		svcFinalizers   []string
+		shouldProcess   bool
+	}{
+		{
+			desc:            "New service without finalizers",
+			backoffDuration: 5 * time.Second,
+			shouldProcess:   true,
+		},
```
TortillaZHawaii: Please add a negative testcase that shows that we actually need to wait more than 3 seconds for the service to be processed, otherwise nothing will happen. Otherwise based on this test we cannot prove that the wait was actually triggered. I think this should work, maybe a bit less than 3s is required:
```suggestion
		{
			desc:            "New service without finalizers, after head start",
			backoffDuration: 5 * time.Second,
			shouldProcess:   true,
		},
		{
		    desc:            "New service without finalizers, head start for CCM",
			backoffDuration: 3 * time.Second, 
			shouldProcess:   false, 
		},
```

pkg/flags/flags.go
FelipeYepez: Done

pkg/l4lb/l4controller_test.go
FelipeYepez: I used `defer close(stopCh)` for tests that run the controller asynchronously with `go l4c.Run()` like this one. Other tests call `l4c.sync()` directly which is a synchronous with no long-running background process so there is nothing to stop.

pkg/l4lb/l4controller_test.go
FelipeYepez: Sure, done

pkg/l4lb/l4controller_test.go
TortillaZHawaii: got it, thanks!

PR #2958, opened 2025-08-26
Keep shared informers alive across ProviderConfig shutdown

panslava:
Previously, MultiProject NEG controller informers were started with a per-ProviderConfig stop channel. When the first ProviderConfig was deleted, its stopCh closed and killed the shared caches — leaving subsequent ProviderConfigs unprocessed until a new leader restart.

Fix by starting informers using global stopCh so they survive individual ProviderConfig lifecycles.

### DIFF
--- a/pkg/multiproject/neg/neg.go
+++ b/pkg/multiproject/neg/neg.go
@@ -31,6 +31,17 @@ import (
 // StartNEGController creates and runs a NEG controller for the specified ProviderConfig.
 // The returned channel is closed by StopControllersForProviderConfig to signal a shutdown
 // specific to this ProviderConfig's controller.
+//
+// Channel lifecycle management:
+//   - globalStopCh: Closes on process/leader shutdown. Shared resources (base informers) use this
+//     to survive individual ProviderConfig deletions.
+//   - providerConfigStopCh: The returned channel that closes when this specific ProviderConfig
+//     is deleted. Used to stop PC-specific controllers.
+//   - joinedStopCh: Internal channel that closes when EITHER globalStopCh OR providerConfigStopCh
+//     closes. Used for PC-specific resources that should stop in either case.
+//
+// IMPORTANT: Base informers from factories use globalStopCh to remain alive across PC changes.
+// Only ProviderConfig-specific controllers and resources should use joinedStopCh.
 func StartNEGController(
 	informersFactory informers.SharedInformerFactory,
 	svcNegFactory informersvcneg.SharedInformerFactory,
@@ -71,7 +82,7 @@ func StartNEGController(
 		}
 	}()
 
-	informers, hasSynced, err := initializeInformers(informersFactory, svcNegFactory, networkFactory, nodeTopologyFactory, providerConfigName, logger, joinedStopCh)
+	informers, hasSynced, err := initializeInformers(informersFactory, svcNegFactory, networkFactory, nodeTopologyFactory, providerConfigName, logger, globalStopCh)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


panslava:
working on fixing unit test

panslava:
ready for review

pkg/multiproject/neg/neg.go
```diff
@@ -71,7 +71,7 @@ func StartNEGController(
 		}
 	}()
 
-	informers, hasSynced, err := initializeInformers(informersFactory, svcNegFactory, networkFactory, nodeTopologyFactory, providerConfigName, logger, joinedStopCh)
+	informers, hasSynced, err := initializeInformers(informersFactory, svcNegFactory, networkFactory, nodeTopologyFactory, providerConfigName, logger, globalStopCh)
```
swetharepakula, line 85: Can't comment on the lines above, but can you add some commends earlier in the function that explains the 3 channels and how they should be used?

pkg/multiproject/start/start_test.go
```diff
@@ -297,6 +308,181 @@ func TestStartProviderConfigIntegration(t *testing.T) {
...
+	validateService(ctx, t, kubeClient, svcNegClient, gceCreator, svc1, pc1)
+
+	// --- pc-2: create and validate service ---
+	pc2 := createPC(ctx, t, pcClient, "pc-2", "owner-2", "proj-2", 2222, "net-2", "subnet-2")
+	seedAll(t, kubeClient, informersFactory, "pc-2", "svc2-a", "20.100")
+	svc2a := createNEGService(ctx, t, kubeClient, "pc-2", "svc2-a", "demo")
+	validateService(ctx, t, kubeClient, svcNegClient, gceCreator, svc2a, pc2)
+
+	// Stop pc-1 (the first PC, which owns the stopCh).
+	markPCDeletingAndWait(ctx, t, pcClient, pc1)
```
swetharepakula, line 366: should there be an expectation on the service here? Should it get cleaned up?

pkg/multiproject/start/start_test.go
```diff
@@ -297,6 +308,181 @@ func TestStartProviderConfigIntegration(t *testing.T) {
...
+	ctx context.Context,
+	t *testing.T,
+	pcClient *pcclientfake.Clientset,
+	pc *providerconfigv1.ProviderConfig,
+) {
+	t.Helper()
+	cp := pc.DeepCopy()
+	now := metav1.NewTime(time.Now())
+	cp.DeletionTimestamp = &now
+	if _, err := pcClient.CloudV1().ProviderConfigs().Update(ctx, cp, metav1.UpdateOptions{}); err != nil {
```
swetharepakula, line 435: Can you add a comment here on why we are doing this versus the deletion? I believe this is because of our mocks work right?

pkg/multiproject/start/start_test.go
```diff
@@ -620,3 +820,187 @@ func getTestEndpointSlices(
...
+				if _, err := kubeClient.CoreV1().Services(tc.service.Namespace).Create(ctx, tc.service, metav1.CreateOptions{}); err != nil {
+					if tc.expectError {
+						t.Logf("Got expected error creating service: %v", err)
+						return
+					}
+					t.Fatalf("Unexpected error creating service: %v", err)
+				}
+
+				// Give some time for processing
+				time.Sleep(2 * time.Second)
```
swetharepakula: can we make this a poll instead a of sleep?

swetharepakula (commented):
Mostly nits, the change LGTM. Thanks Slavik!

pkg/multiproject/start/start_test.go
panslava: I can add more tests later, but currently, deleting ProviderConfig is not triggering neg controller to delete all it's resources. We rely on other automation that deletes namespace, and waits for it to be clean, before deleting providerConfig

pkg/multiproject/start/start_test.go
panslava: Added comment explaining that this is simulates api server behaviour

pkg/multiproject/start/start_test.go
panslava: Added some extra logic to pol

panslava:
/unhold @swetharepakula ready for rereview

I understood there is a way to clean up how we use factories/informers in multiproject neg, but it will require bigger change (that should be no-op) and I will work on it after quick fix

pkg/multiproject/neg/neg.go
swetharepakula: Thanks for adding the comment!

pkg/multiproject/start/start_test.go
swetharepakula: We should make our controller more robust to handle this case, but doesn't need to be handled here.

swetharepakula (commented):
/lgtm
/approve

PR #2965, opened 2025-08-27
[Cherry-pick #2958 to 1.36] Keep shared informers alive across ProviderConfig shutdown

panslava:
Previously, MultiProject NEG controller informers were started with a per-ProviderConfig stop channel. When the first ProviderConfig was deleted, its stopCh closed and killed the shared caches — leaving subsequent ProviderConfigs unprocessed until a new leader restart.

Fix by starting informers using global stopCh so they survive individual ProviderConfig lifecycles.

/assign @swetharepakula


swetharepakula (commented):
/lgtm
/approve

PR #2967, opened 2025-08-28
feat(l4): Add support for enabling L4 NetLB Regional Backend Services by default

08volt:
feat(l4): Enable L4 NetLB Regional Backend Services by default

This change introduces a feature flag `--enable-l4-netlb-rbs-by-default` to enable the use of L4 NetLB Regional Backend Services (RBS) by default for newly created L4 NetLB services, instead of relying on CCM as default implementation.

**Changes:**

*   Added `EnableL4NetLBRBSByDefault` field to `ControllerContextConfig` in `pkg/context/context.go`.
*   Added the `--enable-l4-netlb-rbs-by-default` boolean flag in `pkg/flags/flags.go`, defaulting to `false`.
*   Updated `L4NetLBController` in `pkg/l4lb/l4netlbcontroller.go` to use the `EnableL4NetLBRBSByDefault` setting from the context.
*   Modified `isRBSBasedService` logic to return `true` if the `enableRBSDefault` flag is set, in addition to existing checks (annotations, finalizers, existing FR).
*   Updated unit tests in `pkg/l4/l4netlbcontroller_test.go` to test the new default behavior.

This change allows clusters to default to using the more modern RBS for L4 NetLBs, while still allowing individual services to use the legacy instance-based backend services if explicitly setting the LoadBalancerClass spec in the Service.

#L4NetLB #RBS

### DIFF
--- a/pkg/l4/controllers/l4netlbcontroller_test.go
+++ b/pkg/l4/controllers/l4netlbcontroller_test.go
@@ -312,7 +312,7 @@ func getFakeGCECloud(vals gce.TestClusterValues) *gce.Cloud {
 	return fakeGCE
 }
 
-func buildContext(vals gce.TestClusterValues, readOnlyMode bool) (*ingctx.ControllerContext, error) {
+func buildContext(vals gce.TestClusterValues, readOnlyMode bool, isRBSdefault bool) (*ingctx.ControllerContext, error) {
 	fakeGCE := getFakeGCECloud(vals)
 	kubeClient := fake.NewSimpleClientset()
 	networkClient := netfake.NewSimpleClientset()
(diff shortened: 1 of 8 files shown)


pkg/l4/controllers/l4netlbcontroller_test.go
```diff
@@ -1661,13 +1671,18 @@ func TestIsRBSBasedService(t *testing.T) {
 			frHook:           test.GetLegacyForwardingRule,
 			expectRBSService: false,
 		},
+		{
+			desc:             "Should detect RBS by default",
+			isRBSDefault:     true,
```
TortillaZHawaii, line 1719: What about other testcases? We need to check if this doesn't modify existing Target pool LBs. I think it might be easier to add `expectRBSServiceWithDefault` to existing testcases and run with flag "true" and "false"

08volt: I added those cases and added the check for ForwardingRule legacy in the controller code

08volt: @TortillaZHawaii

PR #2970, opened 2025-09-01
Move to independent locks for L4 and Ingress controllers.

mmamczur:
Acquiring the L4 lock will no longer happen 'inside' the Ingress lock. With this change the controllers can be run as separate containers/pods.


cmd/glbc/main.go
```diff
@@ -364,7 +364,7 @@ func main() {
 		id: fmt.Sprintf("%v_%x", hostname, rand.Intn(1e6)),
 	}
 
-	enableOtherControllers := flags.F.RunIngressController || flags.F.RunL4Controller || flags.F.RunL4NetLBController || flags.F.EnableIGController || flags.F.EnablePSC
+	enableOtherControllers := flags.F.RunL4Controller || flags.F.RunL4NetLBController || flags.F.EnableIGController || flags.F.EnablePSC
```
swetharepakula: nit: rename to enableL4Controllers

mmamczur: done

cmd/glbc/main.go
```diff
@@ -401,28 +408,44 @@ func main() {
...
 			if err != nil {
 				klog.Fatalf("makeLeaderElectionConfig()=%v, want nil", err)
 			}
 			leaderelection.RunOrDie(context.Background(), *ingressRunner)
 		}
+		if flags.F.GateL4ByLock {
+			runL4 = func() {
+				logger := rootLogger.WithName("L4 controller")
+				logger.Info("Start running L4 leader election",
+					"Ingress controller", flags.F.RunIngressController,
```
swetharepakula: remove ingress from the runL4

cmd/glbc/main.go
```diff
@@ -479,14 +502,35 @@ func makeIngressRunnerWithLeaderElection(
...
+	lockLogger := logger.WithValues("lock", l4LockName)
+	return makeRunnerWithLeaderElection(
+		leOption,
+		l4LockName,
+		func(context.Context) {
+			lockLogger.V(0).Info("Acquired L4 Leader election lock")
+			runL4Controllers(ctx, systemHealth, runOption, leOption, logger)
+		},
+		func() {
+			lockLogger.Info("lost master")
```
swetharepakula, line 530: we should also make sure the controllers shutdown here.

cc/ @panslava 

We discovered a bug and Slavik has a change to fix for the other runners: https://github.com/kubernetes/ingress-gce/pull/2977

cmd/glbc/main.go
```diff
@@ -401,28 +408,44 @@ func main() {
...
-				"InstanceGroup controller", flags.F.EnableIGController,
-				"PSC controller", flags.F.EnablePSC,
 			)
 			ingressRunner, err := makeIngressRunnerWithLeaderElection(ctx, systemHealth, rOption, leOption, logger)
 			if err != nil {
 				klog.Fatalf("makeLeaderElectionConfig()=%v, want nil", err)
 			}
 			leaderelection.RunOrDie(context.Background(), *ingressRunner)
 		}
+		if flags.F.GateL4ByLock {
```
swetharepakula: what is the expected behavior when this is false? Should this be included as part of the ingress runner?

mmamczur: let's not over complicate, let's make it a requirement to use L4 locks if LeaderElection is enabled overall.

mmamczur: it's like with NEG flag which is now unused

cmd/glbc/main.go
mmamczur: done

cmd/glbc/main.go
mmamczur: can we merge this and then fix for all in the other PR? I'll be out for most of next week

PR #2972, opened 2025-09-03
multiproject: centralize informers; harden manager; add docs

panslava:
- Centralize informer lifecycle with InformerSet for simpler, shared wiring.
- Harden ProviderConfigControllersManager:
  - Add NEG cleanup finalizer on start; roll back on failures.
  - Idempotent, concurrency-safe start; stop closes channel and removes finalizer using latest object.
  - Introduce startNEGController test seam; unexport controllerSet; wrap errors; add docs.
- Add README explaining multiproject structure and flow.
- Tests: pkg/multiproject/manager/manager_test.go covers finalizer add/rollback, idempotent start, stop behavior, and concurrent single-shot.
- No happy-path behavior change; improves robustness and testability.

/assign @swetharepakula

### DIFF
--- a/pkg/multiproject/manager/manager.go
+++ b/pkg/multiproject/manager/manager.go
@@ -116,27 +116,22 @@ func (pccm *ProviderConfigControllersManager) StartControllersForProviderConfig(
 		return nil
 	}
 
+	cloud, err := pccm.gceCreator.GCEForProviderConfig(pc, logger)
+	if err != nil {
+		return fmt.Errorf("failed to create GCE client for provider config %+v: %w", pc, err)
+	}
+
 	// Track if finalizer already exists so we only roll it back if we added it.
 	// If the PC already had a finalizer (e.g., from a previous controller run),
 	// we must not remove it on failure since NEGs from that run may still exist.
 	hadFinalizer := finalizer.HasGivenFinalizer(pc.ObjectMeta, finalizer.ProviderConfigNEGCleanupFinalizer)
 
-	// Add the cleanup finalizer up front to avoid a window where controllers
-	// may create external resources without a finalizer present. If deletion
-	// happens in that window, cleanup could be skipped. We roll this back on
-	// any subsequent startup error only if we added it.
-	err := finalizer.EnsureProviderConfigNEGCleanupFinalizer(pc, pccm.providerConfigClient, logger)
+	// Add the cleanup finalizer right before starting the controller to minimize
+	// the window where external resources could be created without a finalizer.
+	// We roll this back on subsequent startup error only if we added it.
+	err = finalizer.EnsureProviderConfigNEGCleanupFinalizer(pc, pccm.providerConfigClient, logger)
 	if err != nil {
 		return fmt.Errorf("failed to ensure NEG cleanup finalizer for project %s: %w", pcKey, err)
-	}
-
-	cloud, err := pccm.gceCreator.GCEForProviderConfig(pc, logger)
-	if err != nil {
-		// If GCE client creation fails after finalizer was added, roll it back.
-		if !hadFinalizer {
-			pccm.rollbackFinalizerOnStartFailure(pc, logger, err)
-		}
-		return fmt.Errorf("failed to create GCE client for provider config %+v: %w", pc, err)
 	}
 
 	negControllerStopCh, err := startNEGController(
@@ -173,7 +168,9 @@ func (pccm *ProviderConfigControllersManager) StartControllersForProviderConfig(
 	return nil
 }
 
-func (pccm *ProviderConfigControllersManager) StopControllersForProviderConfig(pc *providerconfig.ProviderConfig) {
+// StopControllersForProviderConfig stops controllers for the given ProviderConfig and removes the cleanup finalizer.
+// Returns an error to allow callers to requeue when cleanup fails.
+func (pccm *ProviderConfigControllersManager) StopControllersForProviderConfig(pc *providerconfig.ProviderConfig) error {
 	logger := pccm.logger.WithValues("providerConfigId", pc.Name)
 
 	csKey := providerConfigKey(pc)
@@ -182,39 +179,34 @@ func (pccm *ProviderConfigControllersManager) StopControllersForProviderConfig(p
 	cs, exists := pccm.controllers.Delete(csKey)
 	if !exists {
 		logger.Info("Controllers for provider config do not exist")
-		return
+		return nil
 	}
 
 	// Perform cleanup operations without holding the lock
 	close(cs.stopCh)
 
-	// Ensure we remove the finalizer from the latest object state.
-	pcLatest := pccm.latestPCWithCleanupFinalizer(pc)
-	err := finalizer.DeleteProviderConfigNEGCleanupFinalizer(pcLatest, pccm.providerConfigClient, logger)
+	// Fetch the latest ProviderConfig to ensure we have current finalizer state.
+	pcLatest, err := pccm.providerConfigClient.CloudV1().ProviderConfigs().Get(context.Background(), pc.Name, metav1.GetOptions{})
 	if err != nil {
-		logger.Error(err, "failed to delete NEG cleanup finalizer for project")
+		return fmt.Errorf("failed to get latest ProviderConfig for finalizer removal: %w", err)
+	}
+	err = finalizer.DeleteProviderConfigNEGCleanupFinalizer(pcLatest, pccm.providerConfigClient, logger)
+	if err != nil {
+		return fmt.Errorf("failed to delete NEG cleanup finalizer for project: %w", err)
 	}
 	logger.Info("Stopped controllers for provider config")
+	return nil
 }
 
 // rollbackFinalizerOnStartFailure removes the NEG cleanup finalizer after a
 // start failure so that ProviderConfig deletion is not blocked.
 func (pccm *ProviderConfigControllersManager) rollbackFinalizerOnStartFailure(pc *providerconfig.ProviderConfig, logger klog.Logger, cause error) {
-	pcLatest := pccm.latestPCWithCleanupFinalizer(pc)
+	pcLatest, err := pccm.providerConfigClient.CloudV1().ProviderConfigs().Get(context.Background(), pc.Name, metav1.GetOptions{})
+	if err != nil {
+		logger.Error(err, "failed to get latest ProviderConfig for finalizer rollback", "originalError", cause)
+		return
+	}
 	if err := finalizer.DeleteProviderConfigNEGCleanupFinalizer(pcLatest, pccm.providerConfigClient, logger); err != nil {
 		logger.Error(err, "failed to clean up NEG finalizer after start failure", "originalError", cause)
 	}
-}
-
-// latestPCWithCleanupFinalizer returns the latest ProviderConfig from the API server.
-// If the Get fails, it returns a local copy of the provided pc with the cleanup
-// finalizer appended to ensure a subsequent delete attempt is not a no-op.
-func (pccm *ProviderConfigControllersManager) latestPCWithCleanupFinalizer(pc *providerconfig.ProviderConfig) *providerconfig.ProviderConfig {
-	pcLatest, err := pccm.providerConfigClient.CloudV1().ProviderConfigs().Get(context.Background(), pc.Name, metav1.GetOptions{})
-	if err != nil {
-		pcCopy := pc.DeepCopy()
-		pcCopy.Finalizers = append(pcCopy.Finalizers, finalizer.ProviderConfigNEGCleanupFinalizer)
-		return pcCopy
-	}
-	return pcLatest
 }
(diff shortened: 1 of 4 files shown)


pkg/multiproject/informerset/informerset.go
```diff
@@ -0,0 +1,281 @@
...
+	startInformer(i.Service, stopCh)
+	startInformer(i.Pod, stopCh)
+	startInformer(i.Node, stopCh)
+	startInformer(i.EndpointSlice, stopCh)
+
+	// Start optional informers
+	startInformer(i.SvcNeg, stopCh)
+	startInformer(i.Network, stopCh)
+	startInformer(i.GkeNetworkParams, stopCh)
+	startInformer(i.NodeTopology, stopCh)
```
swetharepakula, line 161: since we are replicating the check on whether the client is nil or not multiple times. Can we consolidate in initialization the list of informers that needs to be started

pkg/multiproject/informerset/informerset.go
```diff
@@ -0,0 +1,281 @@
...
+			if !hasSynced() {
+				return false
+			}
+		}
+		return true
+	}
+}
+
+// hasSyncedFuncs returns a list of HasSynced functions for all non-nil informers.
+func (i *InformerSet) hasSyncedFuncs() []func() bool {
```
swetharepakula, line 238: can we consolidate these checks perhaps at initialization?

So we save the set of functions and don't have to re-calculate every time?

pkg/multiproject/informerset/informerset_test.go
```diff
@@ -0,0 +1,260 @@
...
+		inf := NewInformerSet(kubeClient, nil, nil, nil, metav1.Duration{Duration: 0})
+
+		stop := make(chan struct{})
+		defer close(stop)
+
+		err := inf.Start(stop, logger)
+		if err != nil {
+			t.Fatalf("Start() returned error: %v", err)
+		}
+		if ok := cache.WaitForCacheSync(stop, inf.CombinedHasSynced()); !ok {
```
swetharepakula, line 192: can you expand this to show that it is waiting for multiple informers

pkg/multiproject/informerset/informerset_test.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		t.Fatalf("Start() returned error: %v", err)
+	}
+
+	filteredAfter := inf.FilterByProviderConfig("pc-1")
+	if filteredAfter == nil {
+		t.Fatalf("FilterByProviderConfig returned nil (after start)")
+	}
+	if !filteredAfter.started {
+		t.Fatalf("expected filtered InformerSet to have started=true after base Start")
```
swetharepakula, line 242: should we check all the informers?

pkg/multiproject/manager/manager.go
```diff
@@ -161,9 +183,34 @@ func (pccm *ProviderConfigControllersManager) StopControllersForProviderConfig(p
...
+}
+
+// latestPCWithCleanupFinalizer returns the latest ProviderConfig from the API server.
+// If the Get fails, it returns a local copy of the provided pc with the cleanup
+// finalizer appended to ensure a subsequent delete attempt is not a no-op.
+func (pccm *ProviderConfigControllersManager) latestPCWithCleanupFinalizer(pc *providerconfig.ProviderConfig) *providerconfig.ProviderConfig {
+	pcLatest, err := pccm.providerConfigClient.CloudV1().ProviderConfigs().Get(context.Background(), pc.Name, metav1.GetOptions{})
+	if err != nil {
+		pcCopy := pc.DeepCopy()
+		pcCopy.Finalizers = append(pcCopy.Finalizers, finalizer.ProviderConfigNEGCleanupFinalizer)
```
swetharepakula: We should optionally add this finalizer if it doesn't exist.
Is this to cover the case our cache hasn't updated to after us putting it on ?

pkg/multiproject/manager/manager_test.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		t.Fatalf("create pc: %v", err)
+	}
+	return pc
+}
+
+// TestStart_AddsFinalizer_AndIsIdempotent verifies that starting controllers
+// for a ProviderConfig adds the NEG cleanup finalizer and that repeated starts
+// are idempotent (the underlying controller is launched only once).
+func TestStart_AddsFinalizer_AndIsIdempotent(t *testing.T) {
```
swetharepakula, line 103: can you add a testcase for starting a controller for a PC that already has a finalizer (in the case of controller restart)

pkg/multiproject/manager/manager_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if len(calls) != 1 || calls[0].pcName != pc.Name {
+		t.Fatalf("unexpected start calls: %#v", calls)
+	}
+
+}
+
+// TestStart_FailureCleansFinalizer verifies that when starting controllers
+// fails, the added finalizer is rolled back and removed from the object.
+func TestStart_FailureCleansFinalizer(t *testing.T) {
```
swetharepakula, line 143: what about the case that the PC already had a finalizer to begin with?

pkg/multiproject/neg/neg_test.go
```diff
@@ -0,0 +1,180 @@
...
+		stopCh <-chan struct{}, l klog.Logger, negMetrics *metrics.NegMetrics, syncerMetrics *syncMetrics.SyncerMetrics) (*neg.Controller, error) {
+		capturedStopCh = stopCh
+		return neg.NewController(kc, sc, ec, uid, ing, svc, pod, node, es, sn, netInf, gke, nt, synced, l4, defSP, cloud, zg, nm,
+			resync, gc, workers, enableRR, runL4, nonGCP, dualStack, lp, multiNetworking, ingressRegional, runNetLB, readOnly, enableNEGsForIngress, stopCh, l, negMetrics, syncerMetrics)
+	}
+	t.Cleanup(func() { newNEGController = orig })
+
+	testCases := []struct {
+		name          string
+		closeProvider bool
```
swetharepakula, line 93: add a comment to what this bool dictates in the test

pkg/multiproject/informerset/informerset_test.go
panslava: Done, if i correctly understood what is needed

pkg/multiproject/informerset/informerset.go
panslava: Same as the previous comment... If you insist I can do it but seems a bit like overengineering, and I don't think re-calculation is too bad.... and nil checks are cheap

pkg/multiproject/informerset/informerset.go
panslava: tbh, tried to do it, but felt like overengineering. Adding a new abstraction to store with not so much gain, and makes it easier to miss and just seems even less readable to store same informers in additional list... I think we should leave it like this...

pkg/multiproject/informerset/informerset_test.go
panslava: Done

pkg/multiproject/manager/manager.go
panslava: Maybe this is a bit overengineering, but this function is only used to get PC on deletion

Idea is -- if by some case "get" failed (idk, some transient error), we don't want to leave PC hanging and still want to try to remove finalizer, so that's the logic besides this function

pkg/multiproject/manager/manager_test.go
panslava: Added TestStart_WithExistingFinalizer_IsIdempotent

pkg/multiproject/manager/manager_test.go
panslava: Good catch - was a bug. We were removing pre-existing finalizers on failure. Fixed + added test.

pkg/multiproject/neg/neg_test.go
panslava: Done

panslava:
Ready for re-review @swetharepakula

pkg/multiproject/informerset/informerset.go
swetharepakula: I am okay to leave as is

pkg/multiproject/informerset/informerset.go
swetharepakula: okay to leave as is for now

pkg/multiproject/informerset/informerset_test.go
swetharepakula: Can you add multiple informers to show it takes all of them?

pkg/multiproject/informerset/informerset_test.go
swetharepakula: I didn't see a diff for this, do you have some unpushed changes?

pkg/multiproject/manager/manager.go
swetharepakula: I think we should instead throw an error in this situation. Is there a reason we don't think the version in the cache is not up to date?

pkg/multiproject/manager/manager.go
```diff
@@ -116,10 +116,15 @@ func (pccm *ProviderConfigControllersManager) StartControllersForProviderConfig(
...
+	// If the PC already had a finalizer (e.g., from a previous controller run),
+	// we must not remove it on failure since NEGs from that run may still exist.
+	hadFinalizer := finalizer.HasGivenFinalizer(pc.ObjectMeta, finalizer.ProviderConfigNEGCleanupFinalizer)
+
 	// Add the cleanup finalizer up front to avoid a window where controllers
 	// may create external resources without a finalizer present. If deletion
 	// happens in that window, cleanup could be skipped. We roll this back on
-	// any subsequent startup error.
+	// any subsequent startup error only if we added it.
 	err := finalizer.EnsureProviderConfigNEGCleanupFinalizer(pc, pccm.providerConfigClient, logger)
```
swetharepakula: move this to right before we start the controller, so we ensure all clients are up. This way we don't have to remove because GCE client didn't initialize.

pkg/multiproject/informerset/informerset_test.go
panslava: Sorry, maybe I didn't get what was needed exactly. 

I added checks in two places:
TestStart_Semantics (lines 174-190) - verifies each individual informer has synced:
```
if !inf.Ingress.HasSynced() { ... }
if !inf.Service.HasSynced() { ... }
if !inf.Pod.HasSynced() { ... }
if !inf.Node.HasSynced() { ... }
if !inf.EndpointSlice.HasSynced() { ... }
```

TestFilterByProviderConfig_WrappingAndState (lines 244-251) - verifies all informers are accessible after start:

```

	if filteredAfter.Ingress == nil || filteredAfter.Service == nil || filteredAfter.Pod == nil || filteredAfter.Node == nil || filteredAfter.EndpointSlice == nil {
		t.Fatalf("expected core filtered informers to be non-nil after base Start")
	}
	if filteredAfter.SvcNeg == nil {
		t.Fatalf("expected SvcNeg filtered informer to be non-nil after base Start")
	}
	if filteredAfter.Network != nil || filteredAfter.GkeNetworkParams != nil || filteredAfter.NodeTopology != nil {
		t.Fatalf("expected absent optional informers to remain nil after base Start")
	}
```

pkg/multiproject/manager/manager.go
panslava: Yeah, I agree actually, removed this, inlined getting providerconfig

pkg/multiproject/manager/manager.go
panslava: Done

panslava:
Ready for re-review @swetharepakula

pkg/multiproject/manager/manager.go
```diff
@@ -182,39 +179,34 @@ func (pccm *ProviderConfigControllersManager) StopControllersForProviderConfig(p
...
-	err := finalizer.DeleteProviderConfigNEGCleanupFinalizer(pcLatest, pccm.providerConfigClient, logger)
+	// Fetch the latest ProviderConfig to ensure we have current finalizer state.
+	pcLatest, err := pccm.providerConfigClient.CloudV1().ProviderConfigs().Get(context.Background(), pc.Name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get latest ProviderConfig for finalizer removal: %w", err)
+	}
+	err = finalizer.DeleteProviderConfigNEGCleanupFinalizer(pcLatest, pccm.providerConfigClient, logger)
 	if err != nil {
-		logger.Error(err, "failed to delete NEG cleanup finalizer for project")
+		return fmt.Errorf("failed to delete NEG cleanup finalizer for project: %w", err)
```
swetharepakula, line 195: Can you add PC information like the name so it is clear which tenant in the case of multiple tenants in the same project.

swetharepakula (commented):
One more nit and this PR lgtm. Thanks!

pkg/multiproject/manager/manager.go
panslava: TBH, we log provideconfig name on the controller.go level when we return errors. Do you think it is actually needed here (that would be repetitive)?

swetharepakula: got it, as long as it is logged, that is okay

PR #2975, opened 2025-09-04
Avoid odd ig nodes sync

midavadim:
Avoid running of add if nodes sync during subsequent NetLB ensuring.

But still we keep full IG ennure precedure during initial NetLB creation when IG are not present yet.

Otherwise, during cluster scale in,  multiple NetLB tried to sync nodes and all NetLB were waiting for long time for node removal (2 .. 6 mins).

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -82,6 +82,30 @@ func NewManager(config *ManagerConfig) Manager {
 	}
 }
 
+// InstanceGroupsExist checks if IGs with given name exist in all cluster zones
+func (m *manager) InstanceGroupsExist(name string, logger klog.Logger) (exist bool, err error) {
+	iglogger := logger.WithName("InstanceGroupsManager")
+
+	// Instance groups should exist in all zones that nodes are in.
+	zones, err := m.ZoneGetter.ListZonesInDefaultSubnet(zonegetter.AllNodesFilter, iglogger)
+	if err != nil {
+		return false, err
+	}
+
+	// Find instance groups in each zone
+	for _, zone := range zones {
+		_, err := m.Get(name, zone)
+		if err != nil {
+			if !utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+				return false, err
+			}
+			logger.V(3).Info("Instance group is not found", "key", klog.KRef(zone, name))
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
 // EnsureInstanceGroupsAndPorts creates or gets an instance group if it doesn't exist
 // and adds the given ports to it. Returns a list of one instance group per zone,
 // all of which have the exact same named ports.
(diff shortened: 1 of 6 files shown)


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -829,9 +829,20 @@ func (lc *L4NetLBController) ensureInstanceGroups(service *v1.Service, nodeNames
 		svcLogger.V(2).Info("Finished ensuring instance groups for L4 NetLB Service", "timeTaken", time.Since(start))
 	}()
 
+	igName := lc.ctx.ClusterNamer.InstanceGroup()
+	exist, _ := lc.instancePool.InstanceGroupsExist(igName, svcLogger)
```
TortillaZHawaii: Why do we skip error?

pkg/l4lb/l4netlbcontroller.go
```diff
@@ -829,9 +829,20 @@ func (lc *L4NetLBController) ensureInstanceGroups(service *v1.Service, nodeNames
 		svcLogger.V(2).Info("Finished ensuring instance groups for L4 NetLB Service", "timeTaken", time.Since(start))
 	}()
 
+	igName := lc.ctx.ClusterNamer.InstanceGroup()
```
mmamczur, line 832: can we have a test for ensureInstanceGroups() that checks if we skip creation and node sync if IG is already there?

pkg/l4lb/l4netlbcontroller.go
midavadim: it is not needed here, but can be useful in another possible use cases

pkg/l4lb/l4netlbcontroller.go
midavadim: added

pkg/l4lb/l4netlbcontroller.go
mmamczur: it would be good to at least log itso that in case it fails for whatever reason we may someday detect it

TortillaZHawaii: > but can be useful in another possible use cases

Since we only call it here, and having an error always means `exist = false` (and `true` is always without an error), let's just make it less confusing and return just a boolean, without any error and add log line wherever it happens.

TortillaZHawaii:
/approve

Ask @mmamczur for the LGTM

pkg/l4lb/l4netlbcontroller.go
midavadim: I added log message

pkg/l4lb/l4netlbcontroller.go
mmamczur: let's log the actual error message so that we know what the error was

pkg/instancegroups/manager.go
```diff
@@ -82,6 +82,27 @@ func NewManager(config *ManagerConfig) Manager {
...
+	// Instance groups should exist in all zones that nodes are in.
+	zones, err := m.ZoneGetter.ListZonesInDefaultSubnet(zonegetter.AllNodesFilter, iglogger)
+	if err != nil {
+		return false, err
+	}
+
+	// Find instance groups in each zone
+	for _, zone := range zones {
+		_, err := m.Get(name, zone)
+		if err != nil {
```
mmamczur, line 98: sorry for a late comment but I only realized now, so this part will log errors when the IG does not exist, even though this is not an error condition since not found is an error.

we could do something like this https://github.com/kubernetes/ingress-gce/blob/3222987665e75e51ece1f2362def8a66f09639dc/pkg/instancegroups/manager.go#L111-L110 to detect not found and just return false here.

PR #2980, opened 2025-09-05
replace 5s poll with WaitForCacheSync

panslava:
- Use client-go cache.WaitForCacheSync to wait for informer caches instead of polling every 5s


swetharepakula (commented):
/lgtm
/approve

PR #2984, opened 2025-09-15
Add a Deny All Firewall Rule

TortillaZHawaii:
A long list of changes, I want to squash the PRs after the reviews to make it easier to backport.

### DIFF
--- a/pkg/loadbalancers/l4netlb.go
+++ b/pkg/loadbalancers/l4netlb.go
@@ -70,9 +70,10 @@ type L4NetLB struct {
 	NamespacedName types.NamespacedName
 	healthChecks   healthchecksl4.L4HealthChecks
 	// mixedManager is responsible for managing forwarding rules for mixed protocol lbs
-	mixedManager    *forwardingrules.MixedManagerNetLB
-	forwardingRules ForwardingRulesProvider
-	enableDualStack bool
+	mixedManager     *forwardingrules.MixedManagerNetLB
+	forwardingRules  ForwardingRulesProvider
+	enableDualStack  bool
+	useDenyFirewalls bool
 	// represents if `enable strong session affinity` flag was set
 	enableStrongSessionAffinity      bool
 	networkInfo                      network.NetworkInfo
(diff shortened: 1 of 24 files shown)


TortillaZHawaii:
/hold

I want to squash before merge, ping me on chat if ready

pkg/loadbalancers/l4netlb.go
```diff
@@ -544,6 +549,60 @@ func (l4netlb *L4NetLB) ensureIPv4NodesFirewall(nodeNames []string, ipAddress st
...
 	}
 	result.Annotations[annotations.FirewallRuleKey] = firewallName
+
+	l4netlb.ensureDeny(result, nodeNames, ipAddress, fwLogger)
+}
+
+func (l4netlb *L4NetLB) ensureDeny(result *L4NetLBSyncResult, nodeNames []string, ipAddress string, fwLogger klog.Logger) {
+	// if we don't use deny flag, make sure that a leftover deny rule is cleaned up
+	if !l4netlb.useDenyFirewalls {
+		if err := cleanUpDenyFirewallRule(l4netlb.cloud, l4netlb.namer.L4FirewallDeny(l4netlb.Service.Namespace, l4netlb.Service.Name), fwLogger); err != nil {
```
mmamczur: so here we'll start doing an extra get once this is merged even before we enable the flag?

TortillaZHawaii: affirm

I'm not sure whether there is a better solution for a rollback. We could have a separate flag for this but I haven't seen much benefit in that. 

The more worrysome is the 1000->999 priority without the flag, which might actually cause some trouble (check the doc for comments). We could flag gate this as well to have a logic like:
```
EnsureFirewall() {
    priority := 999
    if !useDenyFirewalls {
        priority = 1000
        clearUpDeny()
    }
    ensureAllow(priority)
    
    if useDenyFirewalls {
         ensureDeny() // at 1000
    }
}
```

TortillaZHawaii: we would need also to flag gate the priority comparison in equal.go as well, because there might be an edge case that controller with a new version dies and controller running old code (or without the flag) will update the priority of the allow back to 1000, without removing deny, which would block all traffic.

TortillaZHawaii:
due to rebase this is superseded by #3025 

/close

PR #2985, opened 2025-09-15
[Cherrypick #2970 to release-1.37] Move to independent locks for L4 and Ingress controllers.

swetharepakula:
Acquiring the L4 lock will no longer happen 'inside' the Ingress lock. With this change the controllers can be run as separate containers/pods.


gauravkghildiyal:
Thanks!

Seems like a pure cherry-pick of #2970 

/lgtm
/approve

PR #2989, opened 2025-09-22
filter out inactive negs in isZoneChange

midavadim:
1. Upgrade Starts: The E2E test triggers the cluster upgrade. The node in us-central1-a is cordoned and becomes temporarily NotReady.
2. Controller Restarts: The NEG controller pod is rescheduled onto the new, upgraded master.
3. The Race Condition: The new controller starts up and immediately runs a sync loop. Crucially, this happens in the small time window when the node is still NotReady.
4. Incorrect State Detection: The controller's syncInternal() logic queries for the current state of the cluster. Since the only node is NotReady, the zoneGetter reports that there are zero active nodes in us-central1-a.
5. NEG Marked INACTIVE: The controller compares the current state (zero active nodes) with the pre-upgrade state (one active NEG in us-central1-a). It concludes that the zone is no longer active and correctly, based on this limited information, patches the ServiceNetworkEndpointGroup CR to set the NEG's state to INACTIVE.
6. Node Becomes Ready: A moment later, the node finishes its upgrade and becomes Ready.
7. The Bug is Exposed: On the next sync, the controller should see that us-central1-a is active again and switch the NEG back to ACTIVE. This is where it fails. The isZoneChange() function, which is supposed to detect this, has a flaw. It checks the zones of all NEGs in the CR status, including INACTIVE ones.
- existingZones from the CR status is {"us-central1-a"} (from the INACTIVE NEG).
- currZones from the now-ready node is {"us-central1-a"}.
The function sees no difference ({"us-central1-a"} == {"us-central1-a"}) and returns false.
8. Stuck State: Because isZoneChange() returns false, the controller doesn't trigger the reconciliation logic needed to switch the NEG back to ACTIVE. The test waits for an active NEG, which never happens, and times out.

Solution

This is why the proposed change to isZoneChange is so critical. By making it ignore INACTIVE NEGs, it can correctly detect the transition from "no active zones" to "one active zone," triggering the necessary reconciliation to bring the NEG back to an ACTIVE state and fix the load balancer.

### DIFF
--- a/pkg/neg/syncers/transaction_test.go
+++ b/pkg/neg/syncers/transaction_test.go
@@ -2728,6 +2728,7 @@ func TestIsZoneChange(t *testing.T) {
 		zoneDeleted                    bool
 		zoneAdded                      bool
 		nodeWithNoProviderIDAdded      bool
+		negCRStateInactive             bool
 		nodeWithInvalidProviderIDAdded bool
 		expectedResult                 bool
 	}{
@@ -2761,6 +2762,11 @@ func TestIsZoneChange(t *testing.T) {
 			desc:           "no zone change occurred",
 			expectedResult: false,
 		},
+		{
+			desc:               "neg cr state is inactive",
+			negCRStateInactive: true,
+			expectedResult:     true,
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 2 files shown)


midavadim:
@swetharepakula  please check if provided update looks fine for you  and can fix this problem. If yes I will proceed with testing

midavadim:
@swetharepakula 

I have added unittest for this update.
So we can proceed with review.
Please let me know if you are fine with this update.

pkg/neg/syncers/transaction_test.go
```diff
@@ -2793,6 +2799,9 @@ func TestIsZoneChange(t *testing.T) {
 			for _, neg := range negRefMap {
 				refs = append(refs, neg)
 			}
+			if tc.negCRStateInactive {
+				refs[0].State = negv1beta1.InactiveState
```
swetharepakula, line 2803: Can we make all of the NEGs inactive instead of just 1?

This makes the what the flag does consistent. Otherwise, it is a little non-obvious that we are only affecting the first one to be inactive.

Or I would change `negCRStateInactive` to be instead be a list of zones where inactive negs exist. And we can make the refs considering that.

PR #2804, opened 2025-02-14
Refactor ListSubnets

swetharepakula:
* As part of ZoneGetter initialization, generate the default subnet config to stop re-generating the config in every ListSubnets call
 * ListSubnets will always return the current list of subnets
 * TransactionSyncer syncInternalImpl depends on ListSubnets to determine which set of subnets are valid (flag gate occurs in ListSubnet). This abstracts the subnet determination to the zone getter
 * Add unit tests for ListSubnets

/assign @gauravkghildiyal

PR #2990, opened 2025-09-23
Implement a legacy mode for the zonegetter

swetharepakula:
Legacy networks do not have the concept of subnets and naturally are not compatible with the multi subnet feature. Therefore zonegetter must be able to handle legacy networks and not error out. Legacy mode zonegetter still validates that the Nodes have the pod cidr and the provider ID set before using it for any calculation

Legacy Network Assumptions:
* All Nodes are considered part of the default network
* The subnetwork is always empty

Adds a readme on what the zonegetter does at a high level

/assign @gauravkghildiyal

### DIFF
--- a/pkg/utils/zonegetter/zone_getter.go
+++ b/pkg/utils/zonegetter/zone_getter.go
@@ -42,6 +42,10 @@ const (
 	// In NonGCP mode, the ZoneGetter will only return the zone specified
 	// in gce.conf directly.
 	NonGCP
+	// In Legacy mode, the ZoneGetter assumes that all nodes are in the default
+	// subnet. Legacy networks have no concept of subnets, so the subnet returned
+	// is always empty.
+	Legacy
 )
 
 const (
@@ -63,7 +67,7 @@ var providerIDRE = regexp.MustCompile(`^` + "gce" + `://([^/]+)/([^/]+)/([^/]+)$
 // ZoneGetter manages lookups for GCE instances to zones.
 type ZoneGetter struct {
 	nodeLister cache.Indexer
-	// Mode indicates if the ZoneGetter is in GCP or Non-GCP mode
+	// Mode indicates if the ZoneGetter is in GCP, Non-GCP, or Legacy mode
 	// GCP mode ZoneGetter fetches zones from k8s node resource objects.
 	// Non-GCP mode ZoneGetter always return its one single stored zone
 	mode Mode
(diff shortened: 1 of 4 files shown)


gauravkghildiyal (commented):
Thank you for spending time with this. The logic here makes sense. But I had this one question:

I was wondering why do we have some minor differences between the changes here and what's on master. I guess a pure cherry-pick may not have been possible, because of the code differences, but I'm wondering why things like `getSubnet()` behave differently between the two (and that impacts the callers of `getSubnet()` as well)

This the diff between the changes from this PR (left hand side) versus the current master (right hand side): https://www.diffchecker.com/mHqgOs0d/

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -365,6 +381,19 @@ func NewZoneGetter(nodeInformer cache.SharedIndexInformer, defaultSubnetURL stri
 	}
 }
 
+// NewLegacyZoneGetter initialize a ZoneGetter in legacy network mode.
+// Legacy Networks do not have subnets and therefore the subnet should
+// always be empty. Regardless of the labels on the nodes, consider every
+// node as part of the default subnet. onlyIncludeDefaultSubnetNodes is
+// considered true as with Legacy Networks everything is in the default subnet.
+func NewLegacyZoneGetter(nodeInformer cache.SharedIndexInformer) *ZoneGetter {
```
gauravkghildiyal, line 389: Seems like we never use this, and likely we should be using it like we do in our master branch? [https://github.com/kubernetes/ingress-gce/blob/576c1a1e0c5777568e84081d82c5a09fe7fcbd61/pkg/context/context.go#L257](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Maybe that file got left out of the commit.

swetharepakula: Thanks for catching!

Added a commit for the last piece which is to initialize a legacy zone getter when subnetwork is detected to be empty.

swetharepakula:
Zone getter was partially refactored between the two versions which is why the old cherrypick couldn't be cleanly applied.  https://github.com/kubernetes/ingress-gce/pull/2804 is the biggest reason for the diff in the branches. I was refactoring to make the zonegetter abstraction cleaner.

gauravkghildiyal:
Sounds good.

/lgtm
/approve

PR #2991, opened 2025-09-23
Calculate zones from all the NEGs in the SvcNeg

swetharepakula:
The zones here are used to find where all created NEGs exist. We should be including the inactive ones as well, since if endpoints show up in that zone, functionally those zones do have NEGs that can have endpoints that are maintained. Therefore only filter out the ToBeDeleted NEGs.

### DIFF
--- a/pkg/e2e/helpers.go
+++ b/pkg/e2e/helpers.go
@@ -854,9 +854,12 @@ func populateZonesFromCR(serviceName string, svcNegs []negv1beta1.ServiceNetwork
 		}
 
 		for _, negRef := range svcNeg.Status.NetworkEndpointGroups {
-			// We only want zones of the negs that are in an active state.
-			// If a NEG is not active likely the cluster is not in that zone anymore, and we no longer guarantee a NEG in that zone
-			if negRef.State != "" && negRef.State != negv1beta1.ActiveState {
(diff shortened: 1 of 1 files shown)


pkg/e2e/helpers.go
```diff
@@ -856,7 +856,7 @@ func populateZonesFromCR(serviceName string, svcNegs []negv1beta1.ServiceNetwork
 		for _, negRef := range svcNeg.Status.NetworkEndpointGroups {
 			// We only want zones of the negs that are in an active state.
 			// If a NEG is not active likely the cluster is not in that zone anymore, and we no longer guarantee a NEG in that zone
-			if negRef.State != "" && negRef.State != negv1beta1.ActiveState {
+			if negRef.State != "" && negRef.State != negv1beta1.ToBeDeletedState {
```
midavadim: Probably you meant that we should continue if neg == ToBeDeletedState, so !=. should changed to ==

if negRef.State != "" && negRef.State == negv1beta1.ToBeDeletedState {

swetharepakula: Ahh, yes, thank you!

gauravkghildiyal:
/lgtm

Thanks!

PR #2992, opened 2025-09-27
Ensure Ingress CRDS only when Ingress ctrl is enabled

swetharepakula:
* BackendConfig and FrontendConfig are only used by Ingress and do not need to be ensured by other controllers.

/assign @gauravkghildiyal

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -128,32 +128,34 @@ func main() {
 		klog.Fatalf("Failed to create kubernetes client config: %v", err)
 	}
 
-	var backendConfigClient backendconfigclient.Interface
 	crdClient, err := crdclient.NewForConfig(kubeConfig)
 	if err != nil {
 		klog.Fatalf("Failed to create kubernetes CRD client: %v", err)
 	}
 	// TODO(rramkumar): Reuse this CRD handler for other CRD's coming.
 	crdHandler := crd.NewCRDHandler(crdClient, rootLogger)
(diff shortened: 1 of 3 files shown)


cmd/glbc/main.go
```diff
@@ -128,32 +128,34 @@ func main() {
...
+		backendConfigCRDMeta := backendconfig.CRDMeta()
+		if _, err := crdHandler.EnsureCRD(backendConfigCRDMeta, true); err != nil {
+			klog.Fatalf("Failed to ensure BackendConfig CRD: %v", err)
+		}
 
-	backendConfigClient, err = backendconfigclient.NewForConfig(kubeConfig)
-	if err != nil {
-		klog.Fatalf("Failed to create BackendConfig client: %v", err)
-	}
+		backendConfigClient, err = backendconfigclient.NewForConfig(kubeConfig)
```
gauravkghildiyal, line 145: I guess this would cause `BackendConfigInformer.HasSynced` to fail: https://github.com/kubernetes/ingress-gce/blob/576c1a1e0c5777568e84081d82c5a09fe7fcbd61/pkg/context/context.go#L282

We need to decouple that as well? Or maybe it will even fail earlier where the informer never gets created. (Similar thing for FrontendConfigInformer)

swetharepakula: Thanks for catching!

Made a couple fixes -- such as not initializing the informers if the backend config or frontend config clients are nil as well as not adding them to the HasSynced if the informers don't exist.

PR #2994, opened 2025-09-30
Move PSC related metrics into PSC metrics directory

swetharepakula:
* remove PSC Metrics Collector interface and use the PSC metrics collector type directly

* This should be a no-op in regards to the type of metrics and the frequency they are collected.

/assign @gauravkghildiyal 
/assign @TortillaZHawaii


swetharepakula:
/unhold

I have validated that the metrics are equivalent to before and running as expected

PR #2995, opened 2025-09-30
Enable IPv6-only L4 LoadBalancer 

zuzqqa:
This change adds support to the controller to provision L4 LoadBalancers configured as IPv6-only.

### DIFF
--- a/pkg/l4lb/l4controller.go
+++ b/pkg/l4lb/l4controller.go
@@ -77,8 +77,7 @@ type L4Controller struct {
 	syncTracker     utils.TimeTracker
 	forwardingRules ForwardingRulesGetter
 	enableDualStack bool
-
-	hasSynced func() bool
+	hasSynced       func() bool
 
 	serviceVersions *serviceVersionsTracker
 
--- a/pkg/loadbalancers/l4.go
+++ b/pkg/loadbalancers/l4.go
@@ -590,6 +590,7 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	} else {
 		l4.ensureIPv4Resources(result, nodeNames, options, bs, existingIPv4FR, subnetworkURL, ipv4AddressToUse)
 	}
+
 	if result.Error != nil {
 		return result
 	}
@@ -624,6 +625,7 @@ func (l4 *L4) provideHealthChecks(nodeNames []string, result *L4ILBSyncResult) s
 
 func (l4 *L4) provideDualStackHealthChecks(nodeNames []string, result *L4ILBSyncResult) string {
 	sharedHC := !helpers.RequestsOnlyLocalTraffic(l4.Service)
+
 	hcResult := l4.healthChecks.EnsureHealthCheckWithDualStackFirewalls(l4.Service, l4.namer, sharedHC, meta.Global, utils.ILB, nodeNames, utils.NeedsIPv4(l4.Service), utils.NeedsIPv6(l4.Service), l4.network, l4.svcLogger)
 	if hcResult.Err != nil {
 		result.GCEResourceInError = hcResult.GceResourceInError
@@ -633,10 +635,16 @@ func (l4 *L4) provideDualStackHealthChecks(nodeNames []string, result *L4ILBSync
 
 	if hcResult.HCFirewallRuleName != "" {
 		result.Annotations[annotations.FirewallRuleForHealthcheckKey] = hcResult.HCFirewallRuleName
+	} else {
+		delete(result.Annotations, annotations.FirewallRuleForHealthcheckKey)
 	}
+
 	if hcResult.HCFirewallRuleIPv6Name != "" {
 		result.Annotations[annotations.FirewallRuleForHealthcheckIPv6Key] = hcResult.HCFirewallRuleIPv6Name
+	} else {
+		delete(result.Annotations, annotations.FirewallRuleForHealthcheckIPv6Key)
 	}
+
 	result.Annotations[annotations.HealthcheckKey] = hcResult.HCName
 	return hcResult.HCLink
 }
--- a/pkg/loadbalancers/l4_test.go
+++ b/pkg/loadbalancers/l4_test.go
@@ -1290,6 +1290,11 @@ func TestDualStackILBBadCustomSubnet(t *testing.T) {
 			subnetIpv6AccessType: subnetExternalIPv6AccessType,
 			subnetStackType:      "IPV4_IPV6",
 		},
+		{
+			desc:                 "External IPv6 subnet",
+			subnetIpv6AccessType: subnetExternalIPv6AccessType,
+			subnetStackType:      "IPV6_ONLY",
+		},
 	}
 
 	for _, tc := range testCases {
@@ -1851,6 +1856,95 @@ func TestEnsureInternalLoadBalancerAllPorts(t *testing.T) {
 	assertILBResourcesDeleted(t, l4)
 }
 
+func TestIPv6OnlyEnsureInternalLoadBalancerAllPorts(t *testing.T) {
+	t.Parallel()
+
+	nodeNames := []string{"test-node-1"}
+	svc := test.NewL4ILBDualStackService(8080, v1.ProtocolTCP, []v1.IPFamily{v1.IPv6Protocol}, v1.ServiceExternalTrafficPolicyTypeCluster)
+	l4 := mustSetupILBTestHandler(t, svc, nodeNames)
+
+	// Initial creation with port 8080
+	result := l4.EnsureInternalLoadBalancer(nodeNames, svc)
+	if result.Error != nil {
+		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+	}
+	l4.Service.Annotations = result.Annotations
+	assertDualStackILBResources(t, l4, nodeNames)
+
+	// Verify initial port on IPv6 FR
+	frName := l4.getIPv6FRNameWithProtocol("TCP")
+	frKey, err := composite.CreateKey(l4.cloud, frName, meta.Regional)
+	if err != nil {
+		t.Errorf("Unexpected error when creating key - %v", err)
+	}
+	fwdRule, err := composite.GetForwardingRule(l4.cloud, frKey, meta.VersionGA, klog.TODO())
+	if err != nil {
+		t.Errorf("Unexpected error when looking up forwarding rule - %v", err)
+	}
+	if fwdRule.Ports[0] != "8080" {
+		t.Errorf("Unexpected port value %v, expected '8080'", fwdRule.Ports)
+	}
+
+	// Add more than 5 ports to service spec
+	svc.Spec.Ports = []v1.ServicePort{
+		{Name: "testport", Port: int32(8080), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8090), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8100), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8200), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8300), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8400), Protocol: "TCP"},
+	}
+	result = l4.EnsureInternalLoadBalancer(nodeNames, svc)
+	if result.Error != nil {
+		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+	}
+	l4.Service.Annotations = result.Annotations
+	assertDualStackILBResources(t, l4, nodeNames)
+
+	fwdRule, err = composite.GetForwardingRule(l4.cloud, frKey, meta.VersionGA, klog.TODO())
+	if err != nil {
+		t.Errorf("Unexpected error when looking up forwarding rule - %v", err)
+	}
+	if !fwdRule.AllPorts {
+		t.Errorf("Expected AllPorts field to be set in forwarding rule - %+v", fwdRule)
+	}
+	if len(fwdRule.Ports) != 0 {
+		t.Errorf("Unexpected port value %v, expected empty list", fwdRule.Ports)
+	}
+
+	// Modify the service to use less than 5 ports
+	svc.Spec.Ports = []v1.ServicePort{
+		{Name: "testport", Port: int32(8090), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8100), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8300), Protocol: "TCP"},
+		{Name: "testport", Port: int32(8400), Protocol: "TCP"},
+	}
+	expectPorts := []string{"8090", "8100", "8300", "8400"}
+	result = l4.EnsureInternalLoadBalancer(nodeNames, svc)
+	if result.Error != nil {
+		t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
+	}
+	l4.Service.Annotations = result.Annotations
+	assertDualStackILBResources(t, l4, nodeNames)
+
+	fwdRule, err = composite.GetForwardingRule(l4.cloud, frKey, meta.VersionGA, klog.TODO())
+	if err != nil {
+		t.Errorf("Unexpected error when looking up forwarding rule - %v", err)
+	}
+	if !utils.EqualStringSets(fwdRule.Ports, expectPorts) {
+		t.Errorf("Unexpected port value %v, expected %v", fwdRule.Ports, expectPorts)
+	}
+	if fwdRule.AllPorts {
+		t.Errorf("Expected AllPorts field to be unset in forwarding rule - %+v", fwdRule)
+	}
+	// Delete the service
+	result = l4.EnsureInternalLoadBalancerDeleted(svc)
+	if result.Error != nil {
+		t.Errorf("Unexpected error %v", result.Error)
+	}
+	assertDualStackILBResourcesDeleted(t, l4)
+}
+
 func TestEnsureInternalDualStackLoadBalancer(t *testing.T) {
 	t.Parallel()
 	nodeNames := []string{"test-node-1"}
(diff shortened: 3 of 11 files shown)


pkg/loadbalancers/l4.go
```diff
@@ -451,7 +454,7 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	var existingIPv4FR *composite.ForwardingRule
 	var ipv4AddressToUse string
 	var ipv4AddressName string
-	if !l4.enableDualStack || utils.NeedsIPv4(l4.Service) {
+	if l4.enableDualStack || utils.NeedsIPv4(l4.Service) {
```
FelipeYepez: Is there any reason to change this logic? You could add `if (!l4.enableDualStack && !l4.enableIPv6Only) || utils.NeedsIPv4(l4.Service) {`

pkg/l4lb/l4controller.go
```diff
@@ -436,6 +438,7 @@ func (l4c *L4Controller) processServiceDeletion(key string, svc *v1.Service, svc
 		Namer:                            l4c.namer,
 		Recorder:                         l4c.ctx.Recorder(svc.Namespace),
 		DualStackEnabled:                 l4c.enableDualStack,
+		// IPv6OnlyEnabled:                  l4c.enableIPv6Only,
```
FelipeYepez: Seems like you left this commented by accident

pkg/loadbalancers/l4.go
```diff
@@ -135,6 +137,7 @@ func NewL4Handler(params *L4ILBParams, logger klog.Logger) *L4 {
 		healthChecks:                     healthchecksl4.NewL4HealthChecks(params.Cloud, params.Recorder, logger),
 		forwardingRules:                  forwardingrules.New(params.Cloud, meta.VersionGA, scope, logger),
 		enableDualStack:                  params.DualStackEnabled,
+		enableIPv6Only:                   params.IPv6OnlyEnabled,
```
FelipeYepez: No change needed here but due to not including the  `IPv6OnlyEnabled` param for serviceDeletion (was commented out) in the `l4controller.go`, this value might have been false all the time. Still, `enableDualStack` was enabled so maybe that's why testing worked

pkg/loadbalancers/l4.go
FelipeYepez, line 257: I think we don't need this check here, having `if l4.enableMixedProtocol` might be enough. Can you please take a look @TortillaZHawaii

pkg/loadbalancers/l4.go
```diff
@@ -589,12 +587,11 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 		return result
 	}
 	result.Annotations[annotations.BackendServiceKey] = bsName
-	if bs.LogConfig != nil {
-		result.MetricsState.LoggingEnabled = bs.LogConfig.Enable
-	}
 
 	if l4.enableDualStack {
 		l4.ensureDualStackResources(result, nodeNames, options, bs, existingIPv4FR, existingIPv6FR, subnetworkURL, ipv4AddressToUse, ipv6AddrToUse)
+	} else if l4.enableIPv6Only {
```
FelipeYepez: I think it would be better to simply say 
```
    if l4.enableDualStack || l4.enableIPv6Only {
		l4.ensureDualStackResources(result, nodeNames, options, bs, existingIPv4FR, existingIPv6FR, subnetworkURL, ipv4AddressToUse, ipv6AddrToUse)
	} else {
		l4.ensureIPv4Resources(result, nodeNames, options, bs, existingIPv4FR, subnetworkURL, ipv4AddressToUse)
	}
```

No need to worry about IPv4 checks inside `ensureDualStackResources`

pkg/l4lb/l4controller.go
```diff
@@ -77,8 +77,8 @@ type L4Controller struct {
 	syncTracker     utils.TimeTracker
 	forwardingRules ForwardingRulesGetter
 	enableDualStack bool
```
mmamczur, line 79: as a side note - dual stack is enabled since ingress-gce 1.22. We could likely just make it true permanently and clean up the code
But let's not do it in this change

pkg/loadbalancers/l4netlb.go
```diff
@@ -266,7 +270,7 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 	}
 
 	// If service requires IPv6 LoadBalancer -- verify that Subnet with External IPv6 ranges is used.
-	if l4netlb.enableDualStack && utils.NeedsIPv6(svc) {
+	if l4netlb.enableIPv6Only || (l4netlb.enableDualStack && utils.NeedsIPv6(svc)) {
```
mmamczur: will `utils.NeedsIPv6()` be true for each service if the cluster is IPv6?

maybe it is enough to just leave this as it was before

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -961,7 +1532,8 @@ func mustSetupNetLBTestHandler(t *testing.T, svc *v1.Service, nodeNames []string
 		Cloud:            fakeGCE,
 		Namer:            namer,
 		Recorder:         record.NewFakeRecorder(100),
-		DualStackEnabled: true,
+		DualStackEnabled: enableDualStack,
```
mmamczur: I'm wondering if there is much benefit in testing variants without dual stack - I'd go towards assuming it's actually always true.

Especially that this makes the signature less readable if we have `mustSetupNetLBTestHandler(..., true, true)` and you need to check what these bools mean

one way could be to make this function accept an "Options" struct, so that you could call it like `mustSetupNetLBTestHandler(..., setupHandlerOptions{enableIPv6: true})`

or maybe a bit overkill:
```
type mustSetupNetLBtestHandlerOption func(params *L4NetLBParams)

func enableIPv6() mustSetupNetLBtestHandlerOption {
   return func(params *L4NetLBParams){
      params.IPv6OnlyEnabled = true
   }
}

func mustSetupNetLBTestHandler(t *testing.T, svc *v1.Service, nodeNames []string, opts ...mustSetupNetLBtestHandlerOption)
```

and then you do 
`mustSetupNetLBTestHandler(..., enableIPv6)`

pkg/loadbalancers/l4.go
```diff
@@ -624,27 +621,37 @@ func (l4 *L4) requireZonalAffinity(svc *corev1.Service) bool {
...
 		return l4.provideDualStackHealthChecks(nodeNames, result)
 	}
 	return l4.provideIPv4HealthChecks(nodeNames, result)
 }
 
 func (l4 *L4) provideDualStackHealthChecks(nodeNames []string, result *L4ILBSyncResult) string {
 	sharedHC := !helpers.RequestsOnlyLocalTraffic(l4.Service)
-	hcResult := l4.healthChecks.EnsureHealthCheckWithDualStackFirewalls(l4.Service, l4.namer, sharedHC, meta.Global, utils.ILB, nodeNames, utils.NeedsIPv4(l4.Service), utils.NeedsIPv6(l4.Service), l4.network, l4.svcLogger)
+
+	needsIPv4ForHC := !l4.enableIPv6Only && utils.NeedsIPv4(l4.Service)
```
FelipeYepez: The existing logic should work so you can leave `needsIPv4ForHC := utils.NeedsIPv4(l4.Service)` here

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -949,7 +949,578 @@ func TestDualStackNetLBStaticIPAnnotation(t *testing.T) {
...
+			expectedStaticLoadBalancerIP: "2::2",
+			expectError:                  false,
+		},
+		{
+			desc:                         "Annotation with two reserved IPv6 addresses",
+			staticAnnotationVal:          "ipv6-address,ipv6-address1",
+			addressesToReserve:           []*ga.Address{ipv6Address, ipv6Address1},
+			expectedStaticLoadBalancerIP: "2::2",
+			expectError:                  false,
+		},
```
mmamczur, line 1161: do we have a test case where this is an invalid name in the annotation?
what would happen if we give it a IPv4 address?

pkg/loadbalancers/l4.go
```diff
@@ -624,27 +621,37 @@ func (l4 *L4) requireZonalAffinity(svc *corev1.Service) bool {
...
 	}
 
-	if hcResult.HCFirewallRuleName != "" {
+	if needsIPv4ForHC && hcResult.HCFirewallRuleName != "" {
 		result.Annotations[annotations.FirewallRuleForHealthcheckKey] = hcResult.HCFirewallRuleName
+	} else {
+		delete(result.Annotations, annotations.FirewallRuleForHealthcheckKey)
 	}
+
 	if hcResult.HCFirewallRuleIPv6Name != "" {
```
FelipeYepez, line 642: Try to use the same logic as above `if needsIPv6ForHC && ...` or remove it `if hcResult.HCFirewallRuleName != ""` for consistency

pkg/loadbalancers/l4netlb_test.go
zuzqqa: I added test cases to cover:
- Using an invalid/non-existent address name in the annotation.
- Using an IPv4 address name in the annotation.

In both scenarios, the current implementation does not return an error. Instead, it falls back and selects an available IPv6 address.

I discussed this behavior with Felipe, and an error is not returned for now. I will add a note about this in the design doc, including the potential for stricter validation and error handling in the future, as I agree that surfacing an error for invalid user input would be ideal.

pkg/loadbalancers/l4.go
```diff
@@ -417,7 +420,7 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	l4.network = *svcNetwork
 
 	// If service requires IPv6 LoadBalancer -- verify that Subnet with Internal IPv6 ranges is used.
-	if l4.enableDualStack && utils.NeedsIPv6(l4.Service) {
+	if l4.enableIPv6Only || (l4.enableDualStack && utils.NeedsIPv6(l4.Service)) {
```
FelipeYepez: With your findings you could say `if (l4.enableIPv6Only || l4.enableDualStack) && utils.NeedsIPv6(l4.Service) {` or simply leave the existing logic

pkg/loadbalancers/l4netlb.go
```diff
@@ -286,6 +290,8 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
 
 	if l4netlb.enableDualStack {
 		l4netlb.ensureDualStackResources(result, nodeNames, bsLink)
+	} else if l4netlb.enableIPv6Only {
```
FelipeYepez: Same here, might be better to say
```
	if l4netlb.enableDualStack || l4netlb.enableIPv6Only {
		l4netlb.ensureDualStackResources(result, nodeNames, bsLink)
	} else {
		l4netlb.ensureIPv4Resources(result, nodeNames, bsLink)
	}
```

pkg/loadbalancers/l4netlb.go
```diff
@@ -294,15 +300,19 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service)
...
 		return l4netlb.provideDualStackHealthChecks(nodeNames, result)
 	}
 	return l4netlb.provideIPv4HealthChecks(nodeNames, result)
 }
 
 func (l4netlb *L4NetLB) provideDualStackHealthChecks(nodeNames []string, result *L4NetLBSyncResult) string {
 	sharedHC := !helpers.RequestsOnlyLocalTraffic(l4netlb.Service)
-	hcResult := l4netlb.healthChecks.EnsureHealthCheckWithDualStackFirewalls(l4netlb.Service, l4netlb.namer, sharedHC, l4netlb.scope, utils.XLB, nodeNames, utils.NeedsIPv4(l4netlb.Service), utils.NeedsIPv6(l4netlb.Service), l4netlb.networkInfo, l4netlb.svcLogger)
+
+	needsIPv4ForHC := !l4netlb.enableIPv6Only && utils.NeedsIPv4(l4netlb.Service)
```
FelipeYepez: The existing logic should work so you can leave `needsIPv4ForHC := utils.NeedsIPv4(l4netlb.Service)` here

pkg/loadbalancers/l4netlb.go
```diff
@@ -311,11 +321,16 @@ func (l4netlb *L4NetLB) provideDualStackHealthChecks(nodeNames []string, result
 		return ""
 	}
 
-	if hcResult.HCFirewallRuleName != "" {
+	if needsIPv4ForHC && hcResult.HCFirewallRuleName != "" {
```
FelipeYepez: Also here, keep the check consistent between `if hcResult.HCFirewallRuleName != "" {` and `if hcResult.HCFirewallRuleIPv6Name != "" {`

If you add `needsIPv4ForHC` for one, add it for the other `needsIPv6ForHC` or don't add it to any of them

pkg/utils/utils.go
```diff
@@ -87,6 +87,7 @@ const (
 	// be removed in 1.18.
 	LabelAlphaNodeRoleExcludeBalancer = "alpha.service-controller.kubernetes.io/exclude-balancer"
 	DualStackSubnetStackType          = "IPV4_IPV6"
+	IPv6SubnetStackType               = "IPV6"
```
FelipeYepez: Apparently the possible values of Subnetwork.StackType are `IPV4_IPV6`, `IPV4_ONLY` and `IPV6_ONLY`. Can you please try with `IPV6_ONLY`.

Was `IPV6` working for you?

pkg/utils/utils.go
```diff
@@ -861,7 +862,7 @@ func SubnetHasIPv6Range(cloud *gce.Cloud, subnetName, ipv6AccessType string) (bo
 	if err != nil {
 		return false, fmt.Errorf("failed getting subnet: %w", err)
 	}
-	return subnet.StackType == DualStackSubnetStackType && subnet.Ipv6AccessType == ipv6AccessType, nil
+	return ((subnet.StackType == DualStackSubnetStackType) || (subnet.StackType == IPv6SubnetStackType)) && subnet.Ipv6AccessType == ipv6AccessType, nil
```
FelipeYepez: Can you please break this logic into separate lines. Something like:

```
if subnet.Ipv6AccessType != ipv6AccessType {
    return false, nil
}

isCorrectStackType := (subnet.StackType == DualStackSubnetStackType) || (subnet.StackType == IPv6SubnetStackType)
if !isCorrectStackType {
    return false, nil
}
return true, nil
```

pkg/utils/utils.go
FelipeYepez:
```
    // StackType: The stack type for the subnet. If set to IPV4_ONLY, new VMs in
	// the subnet are assigned IPv4 addresses only. If set to IPV4_IPV6, new VMs in
	// the subnet can be assigned both IPv4 and IPv6 addresses. If not specified,
	// IPV4_ONLY is used. This field can be both set at resource creation time and
	// updated using patch.
	//
	// Possible values:
	//   "IPV4_IPV6" - New VMs in this subnet can have both IPv4 and IPv6
	// addresses.
	//   "IPV4_ONLY" - New VMs in this subnet will only be assigned IPv4 addresses.
	//   "IPV6_ONLY" - New VMs in this subnet will only be assigned IPv6 addresses.
	StackType string `json:"stackType,omitempty"`
```

pkg/loadbalancers/l4netlb.go
zuzqqa: I will leave the existing logic here

pkg/loadbalancers/l4netlb.go
zuzqqa: Same here, I will leave an existing logic here so needsIPv4ForHC and needsIPv6ForHC won't be used

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1444,6 +1718,66 @@ func assertDualStackNetLBResourcesWithCustomIPv6Subnet(t *testing.T, l4NetLB *L4
...
+		t.Errorf("verifyNetLBIPv6HealthCheckFirewall() = %v, want nil", err)
+	}
+
+	if err := verifyNetLBIPv6NodesFirewall(l4NetLB, nodeNames); err != nil {
+		t.Errorf("verifyNetLBIPv6NodesFirewall() = %v, want nil", err)
+	}
+
+	if err := verifyNetLBIPv6NodesFirewall(l4NetLB, nodeNames); err != nil {
+		t.Errorf("verifyNetLBIPv6NodesFirewall() = %v, want nil", err)
+	}
```
FelipeYepez: Seems like this verification is duplicated

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1444,6 +1718,66 @@ func assertDualStackNetLBResourcesWithCustomIPv6Subnet(t *testing.T, l4NetLB *L4
...
+
+	if err := verifyNetLBIPv4ResourcesDeletedOnSync(l4NetLB); err != nil {
+		t.Errorf("verifyNetLBIPv4ResourcesDeletedOnSync(_) returned error %v, want nil (no IPv4 resources)", err)
+	}
```
FelipeYepez: Since you verified this at the top of the function you can remove this check

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1444,6 +1718,66 @@ func assertDualStackNetLBResourcesWithCustomIPv6Subnet(t *testing.T, l4NetLB *L4
...
+
+	isSharedHC = !servicehelper.RequestsOnlyLocalTraffic(l4NetLB.Service)
+	hcFwName = l4NetLB.namer.L4HealthCheckFirewall(l4NetLB.Service.Namespace, l4NetLB.Service.Name, isSharedHC)
+	if err := verifyFirewallNotExists(l4NetLB.cloud, hcFwName); err != nil {
+		t.Errorf("verifyFirewallNotExists(_, %s) for IPv4 HC FW returned error %v, want nil", hcFwName, err)
+	}
```
FelipeYepez: This is as well duplicated, please remove

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -949,6 +954,275 @@ func TestDualStackNetLBStaticIPAnnotation(t *testing.T) {
...
+	// Copy result annotations to the service, as assertion verifies that service got proper annotations.
+	svc.Annotations = result.Annotations
+	assertDualStackNetLBResourcesWithCustomIPv6Subnet(t, l4NetLB, nodeNames, l4NetLB.cloud.SubnetworkURL())
+
+	// create custom subnet
+	subnetKey := meta.RegionalKey("test-subnet", l4NetLB.cloud.Region())
+	subnetToCreate := &ga.Subnetwork{
+		Ipv6AccessType: subnetExternalIPv6AccessType,
+		StackType:      "IPV4_IPV6",
```
FelipeYepez: Might be better to use `IPV6_ONLY` since it is already tested for dual_stack

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -949,6 +954,275 @@ func TestDualStackNetLBStaticIPAnnotation(t *testing.T) {
...
+		AddressType:      "EXTERNAL",
+		Ipv6EndpointType: "NETLB",
+	}
+
+	testCases := []struct {
+		desc                         string
+		staticAnnotationVal          string
+		addressesToReserve           []*ga.Address
+		expectedStaticLoadBalancerIP string
+		expectError                  bool
```
FelipeYepez: Seems like you don't need `expectError` since it is always false

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -949,6 +954,275 @@ func TestDualStackNetLBStaticIPAnnotation(t *testing.T) {
...
+			svc.Annotations[annotations.StaticL4AddressesAnnotationKey] = tc.staticAnnotationVal
+
+			result := l4NetLB.EnsureFrontend(nodeNames, svc)
+			if tc.expectError {
+				if result.Error == nil {
+					t.Fatalf("EnsureFrontend(%+v) = nil, want error", svc.Spec)
+				}
+				t.Logf("EnsureFrontend(%+v) returned expected error: %v", svc.Spec, result.Error)
+				return
+			}
```
FelipeYepez: Once you remove this please make sure to validate no error was returned
```
if result.Error != nil {
	t.Errorf("Failed to ensure loadBalancer, err %v", result.Error)
}
```

pkg/loadbalancers/l4_test.go
```diff
@@ -2299,6 +2393,96 @@ func TestDualStackILBStaticIPAnnotation(t *testing.T) {
...
+	}{
+		{
+			desc:                         "Valid static IPv6 address",
+			staticAnnotationVal:          "ipv6-address",
+			addressesToReserve:           []*compute.Address{ipv6Address},
+			expectedStaticLoadBalancerIP: "2::2",
+		},
+		{
+			desc:                         "Annotation with two reserved IPv6 addresses",
+			staticAnnotationVal:          "ipv4-address,ipv6-address",
```
FelipeYepez: `ipv4-address` doesn't exist here, I guess you wanted to use `ipv6-address1`?

pkg/loadbalancers/l4netlb_test.go
```diff
@@ -1444,6 +1711,54 @@ func assertDualStackNetLBResourcesWithCustomIPv6Subnet(t *testing.T, l4NetLB *L4
...
+	if err := verifyNetLBIPv6HealthCheckFirewall(l4NetLB, nodeNames); err != nil {
+		t.Errorf("verifyNetLBIPv6HealthCheckFirewall() = %v, want nil", err)
+	}
+
+	if err := verifyNetLBIPv6NodesFirewall(l4NetLB, nodeNames); err != nil {
+		t.Errorf("verifyNetLBIPv6NodesFirewall() = %v, want nil", err)
+	}
+
+	isSharedHC = !servicehelper.RequestsOnlyLocalTraffic(l4NetLB.Service)
+	hcFwName = l4NetLB.namer.L4HealthCheckFirewall(l4NetLB.Service.Namespace, l4NetLB.Service.Name, isSharedHC)
```
FelipeYepez: Forgot to remove this

FelipeYepez (approved):

PR #2996, opened 2025-10-06
Prevent flakiness in instance groups controller test

anyush:
This test originally was added to test that creation/update/deletion of multiple nodes won't cause multiple IG controller resyncs: https://github.com/kubernetes/ingress-gce/pull/2523

However as soon as node created/updated/deleted controller tries to take event out of the queue and process it, therefore if 2 actions performed one after another there is still a possibility that event processing will start before second action pushes event on the queue, which will increase `len(fakeManager.syncedNodes)` value.

### DIFF
--- a/pkg/instancegroups/controller_test.go
+++ b/pkg/instancegroups/controller_test.go
@@ -2,6 +2,7 @@ package instancegroups
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/instancegroups/controller_test.go
```diff
@@ -192,9 +225,12 @@ func verifyExpectedSyncerCount(t *testing.T, syncedNodes [][]string, expectedCou
 
 type IGManagerFake struct {
 	syncedNodes [][]string
+	SyncLock    sync.Locker
```
TortillaZHawaii, line 228: why not use sync.Mutex directly? you don't need to initialize it explicitly

PR #3002, opened 2025-10-15
Include ensuring instance groups in latency metric

midavadim:


pkg/l4lb/l4netlbcontroller.go
```diff
@@ -671,6 +671,7 @@ func (lc *L4NetLBController) syncInternal(service *v1.Service, svcLogger klog.Lo
 		}
 		return syncResult
 	}
+	syncResult.StartTime = startTime
```
08volt: Why is it after the Error check? We would retur different startTime if there was an Error or not

mmamczur: it feels like a hack here. I'd try to either refactor how the syncResult is created (create it in this function instead of inside `l3netln.EnsureFrontend()`) or maybe pass start time into EnsureFrontend().
It's very confusing now since time will be sent inside EnsureFrontend and then overridden here

mmamczur: we could also move away from tracking this time in L4SyncResult. This could be a better approach really. We could put somewhere around this place https://github.com/kubernetes/ingress-gce/blob/13ae861eebc2ce9c55644c2048f8358f53d09769/pkg/l4lb/l4netlbcontroller.go#L541-L540

and then pass the start time into https://github.com/kubernetes/ingress-gce/blob/13ae861eebc2ce9c55644c2048f8358f53d09769/pkg/l4lb/l4netlbcontroller.go#L585

make sure that the times from L4Result are not used anywhere else and remove that startTime field

midavadim: I updated [l4netlbcontroller.go](https://github.com/kubernetes/ingress-gce/blob/13ae861eebc2ce9c55644c2048f8358f53d09769/pkg/l4lb/l4netlbcontroller.go#L585) and passwd starttime to it becuase [l4netlbcontroller.go](https://github.com/kubernetes/ingress-gce/blob/13ae861eebc2ce9c55644c2048f8358f53d09769/pkg/l4lb/l4netlbcontroller.go#L585) becuase startime is used in :
ingress-gce/pkg/loadbalancers/l4netlb.go (9872615) lines 115-116
		MetricsLegacyState: metrics.InitL4NetLBServiceLegacyState(&startTime),
		MetricsState:       metrics.InitServiceMetricsState(svc, &startTime, isMultinet, enabledStrongSessionAffinity, isWeightedLBPodsPerNode, isLBWithZonalAffinity, backendType),

pkg/loadbalancers/l4netlb.go
```diff
@@ -98,8 +98,8 @@ type L4NetLBSyncResult struct {
 	GCEResourceUpdate  ResourceUpdates
 }
 
-func NewL4SyncResult(syncType string, svc *corev1.Service, isMultinet bool, enabledStrongSessionAffinity bool, isWeightedLBPodsPerNode bool, useNEGs bool) *L4NetLBSyncResult {
-	startTime := time.Now()
+func NewL4SyncResult(syncType string, startTime time.Time, svc *corev1.Service, isMultinet bool, enabledStrongSessionAffinity bool, isWeightedLBPodsPerNode bool, useNEGs bool) *L4NetLBSyncResult {
+
```
midavadim, line 102: startTime is passed from parent function here.

midavadim: and now it is aligng with code for ILB

08volt (approved):

PR #3022, opened 2025-12-05
Add configurable connection draining timeout for L4 LoadBalancers

mattrobenolt:
Adds `networking.gke.io/connection-draining-timeout-sec` annotation to configure connection draining timeout for L4 LoadBalancers with RBS enabled.

Currently, L4 LoadBalancers use a hardcoded 30 second timeout. For workloads with long-lived connections, this is quite insufficient. GCE supports up to 3600 seconds, but there is no way to configure this in the Service definition.

Note: for disclosure, I used the assistance of Claude Code to help implement this PR, but I have reviewed everything personally to the best of my knowledge.

### DIFF
--- a/pkg/l4annotations/service.go
+++ b/pkg/l4annotations/service.go
@@ -18,6 +18,7 @@ package l4annotations
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
@@ -99,6 +100,9 @@ const (
 
 	// Service annotation key for specifying config map which contains logging config
 	L4LoggingConfigMapKey = "networking.gke.io/l4-logging-config-map"
+
+	// Service annotation key for specifying connection draining timeout in seconds for L4 backend services
+	ConnectionDrainingTimeoutKey = "networking.gke.io/connection-draining-timeout-sec"
 )
 
 // Service represents Service annotations.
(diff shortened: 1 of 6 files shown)


pkg/l4annotations/service.go
```diff
@@ -99,6 +100,9 @@ const (
 
 	// Service annotation key for specifying config map which contains logging config
 	L4LoggingConfigMapKey = "networking.gke.io/l4-logging-config-map"
+
+	// Service annotation key for specifying connection draining timeout in seconds for L4 backend services
+	ConnectionDrainingTimeoutKey = "networking.gke.io/connection-draining-timeout-sec"
```
TortillaZHawaii: We could use "networking.gke.io/connection-draining-timeout" and use `time.Duration` to allow specifying time with unit, like in https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#configurable-container-restart-delay

TortillaZHawaii (commented):
Hi, thanks for the PR. 

I would like to understand what happens, if the user has modified the timeout on the created Backend Service resource through Cloud Console or CLI. Unfortunately we did somewhat allow this to happen (even if it wasn't supported) and this PR would break the expectations for those users.

It's also not clear what the implications are for the rest of the stack, I'm not sure if there won't be any issues somewhere unexpected.

/hold
/ok-to-test

pkg/l4annotations/service.go
mattrobenolt: I'm not entirely sure that makes sense in this context when GCP only accepts seconds as input in the API.

So what would it mean for a spec to say `networking.gke.io/connection-draining-timeout = 100ms`?

We could error on anything that is sub-second, and that'd enable us to use `5m` or something else?

Is that more what you were thinking?

mattrobenolt:
> I would like to understand what happens, if the user has modified the timeout on the created Backend Service resource through Cloud Console or CLI. Unfortunately we did somewhat allow this to happen (even if it wasn't supported) and this PR would break the expectations for those users.

The expected behavior with this PR is that value is preserved, and that's exactly our setup. We create the LoadBalancer, then need to modify it after with `gcloud` to change this attribute. pkg/backends/backends.go:472

pkg/l4annotations/service.go
TortillaZHawaii: > I'm not entirely sure that makes sense in this context when GCP only accepts seconds as input in the API.
> So what would it mean for a spec to say networking.gke.io/connection-draining-timeout = 100ms?

We need validation regardless of using int or duration. 

> and that'd enable us to use 5m or something else?

yes exactly

mattrobenolt: Done, I swapped this to just `connection-draining-timeout` with some validation for the acceptable range.

mattrobenolt:
@TortillaZHawaii if you'd like to take another look, I think everything is addressed. Also added more testing around pre-existing values not being clobbered.

PR #3060, opened 2026-03-13
refactor: update informer set to use SharedInformerFactory for GKE Networking API and core Kubernetes resources.

iromanycheva-eng:
Refactor InformerSet to use component SharedInformerFactory instead of handling each Informer individually. 
This change is aligned with standard controller idiomatic patterns.


pkg/multiproject/neg/informerset/informerset.go
```diff
@@ -51,80 +53,48 @@ func NewInformerSet(
...
+	if nodeTopologyClient != nil {
+		infSet.nodetopologyFactory = nodetopologyinformers.NewSharedInformerFactory(nodeTopologyClient, resyncPeriod.Duration)
+	}
 
-	informers.Node = informerv1.NewNodeInformer(
-		kubeClient,
-		resyncPeriod.Duration,
-		utils.NewNamespaceIndexer(),
-	)
+	// Create core Kubernetes informers from factory
```
swetharepakula, line 69: for all of the informers we create, set the namespace indexer on them

swetharepakula (commented):
This change mostly looks good to me.

A couple of comments, to add back the namespace indexer for the informers.

For the commits, can you squash them. If you would like to keep each commit separate, then update the messages to indicate what each commit is doing. Also can you separate out the vendor and the go.mod changes. I would make that the first commit and then have the code changes after that.

pkg/multiproject/neg/informerset/informerset.go
iromanycheva-eng: They are already set because we are using SharedInformerFactory.

PR #3081, opened 2026-04-10
metrics: Rename logging_control_enabled label to logging_enabled

08volt:
## Description
This PR renames the Prometheus metric label `logging_control_enabled` to `logging_enabled` in the L4 IPv4 metrics implementation. This makes the label name more concise.

## Changes
*   `pkg/l4/metrics/ipv4_metrics.go`: Updated the value of the `l4LabelLoggingControlEnabled` constant from `"logging_control_enabled"` to `"logging_enabled"`.
*   `pkg/l4/metrics/metrics_test.go`: Updated the corresponding test cases to use the new label name `"logging_enabled"`.

## Verification
*   The unit tests in `pkg/l4/metrics/metrics_test.go` were updated to reflect this change.

### DIFF
--- a/pkg/l4/metrics/ipv4_metrics.go
+++ b/pkg/l4/metrics/ipv4_metrics.go
@@ -35,7 +35,7 @@ const (
 	l4LabelZonalAffinity         = "zonal_affinity"
 	l4LabelBackendType           = "backend_type"
 	l4LabelProtocol              = "protocol" // Either "TCP", "UDP", or "MIXED". The default "" means that there is an error.
-	l4LabelLoggingControlEnabled = "logging_control_enabled"
+	l4LabelLoggingControlEnabled = "logging_enabled"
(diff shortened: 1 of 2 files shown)


TortillaZHawaii:
/hold

How many versions of this label are in PROD? Is there both "logging_enabled" AND "logging_control_enabled" in the wild at the same time?

08volt:
> /hold
> 
> How many versions of this label are in PROD? Is there both "logging_enabled" AND "logging_control_enabled" in the wild at the same time?

I don't think so as release branch 1.38 still has logging_enabled

pkg/l4/backends/backends.go
TortillaZHawaii, line 291: Do you need to rebase it? This was part of other PR

08volt: fixed

PR #3084, opened 2026-04-13
Stabilize multiproject controller tests

rino-gaelico:
Continuation of the work started with [PR#3080](https://github.com/kubernetes/ingress-gce/pull/3080).

[In client-go v0.35+](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/3157-watch-list/README.md), informers now set `AllowWatchBookmarks: true` in their watch options and expect a `watch.Bookmark` event to signal the end of the initial event stream.

Because standard fake clients do not emit this bookmark event by default, calls to `WaitForCacheSync` hang indefinitely in unit tests.

As a fix, we Injected `test.PrependBookmarkReactor` into all fake clients (kube, pc, svcNeg, network, nodeTopology) across all test cases to satisfy the new client-go requirements for cache synchronization.

In addition to that we added a 2-second sleep to allow the controller to initialize and register event handlers before resources are created in the test.

### DIFF
--- a/pkg/multiproject/start/start_test.go
+++ b/pkg/multiproject/start/start_test.go
@@ -8,12 +8,16 @@ import (
 	"testing"
 	"time"
 
+	networkv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/network/v1"
+	nodetopologyv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/nodetopology/v1"
 	networkfake "github.com/GoogleCloudPlatform/gke-networking-api/client/network/clientset/versioned/fake"
 	nodetopologyfake "github.com/GoogleCloudPlatform/gke-networking-api/client/nodetopology/clientset/versioned/fake"
 	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/multiproject/start/start_test.go
```diff
@@ -351,6 +429,11 @@ func TestSharedInformers_PC1Stops_PC2AndPC3KeepWorking(t *testing.T) {
 		gceCreator, rootNamer, globalStop, syncMetrics.FakeSyncerMetrics(),
 	)
 
+	// Without the time.Sleep: the main test goroutine would create resources
+	// before the background goroutine had actually started and registered
+	// event handlers, causing the test to fail to register some event handlers.
+	time.Sleep(2 * time.Second)
```
mmamczur, line 435: this is somewhat surprising and maybe only a property of the fake informer, but normally, aren't events for all resources emitted when you create a new informer?
For example when the controller is updated or restarted, it should reprocess everything

rino-gaelico: I agree on the perplexity. 
I think that since the informer starts the Watch after the create event, the catching up is too slow, while unit test asserts expectations instantly.
An alternative way, perhaps more robust, would be to sync on the informer to start watching.

rino-gaelico: Also, this is happening only after the update to 1.35. Probably because the new falling back mechanism (it checks if watchlist is supported or not first) adds enough overhead, to miss the addition of the event handler in time for the first creation event

PR #3090, opened 2026-04-17
Fix warning related to Printf like formatting that became build error…

mmamczur:
…s in go1.26 (code won't build without fixing this on go 1.26)

This is done in preparation for k8s 1.36 that requires go 1.26.

### DIFF
--- a/pkg/psc/controller.go
+++ b/pkg/psc/controller.go
@@ -240,7 +240,7 @@ func (c *Controller) handleErr(err error, key interface{}) {
 		c.logger.Info("failed to retrieve service attachment from the store", "serviceKey", key.(string), "err", err)
 	} else if exists {
 		svcAttachment := obj.(*sav1.ServiceAttachment)
-		c.recorder(svcAttachment.Namespace).Eventf(svcAttachment, v1.EventTypeWarning, "ProcessServiceAttachmentFailed", eventMsg)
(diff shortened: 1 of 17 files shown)


pkg/psc/controller.go
```diff
@@ -240,7 +240,7 @@ func (c *Controller) handleErr(err error, key interface{}) {
 		c.logger.Info("failed to retrieve service attachment from the store", "serviceKey", key.(string), "err", err)
 	} else if exists {
 		svcAttachment := obj.(*sav1.ServiceAttachment)
-		c.recorder(svcAttachment.Namespace).Eventf(svcAttachment, v1.EventTypeWarning, "ProcessServiceAttachmentFailed", eventMsg)
+		c.recorder(svcAttachment.Namespace).Eventf(svcAttachment, v1.EventTypeWarning, "ProcessServiceAttachmentFailed", "error processing service attachment %q: %q", key, err)
```
08volt: Nit: we could keep using "eventMsg" and just use "Event" instead of Eventf:
```
c.recorder(svcAttachment.Namespace).Event(svcAttachment, v1.EventTypeWarning, "ProcessServiceAttachmentFailed", eventMsg)
```

08volt (commented):
/lgtm

pkg/psc/controller.go
mmamczur: changed it to use `Event`

PR #3095, opened 2026-04-22
bump go version to 1.25.9

kl52752:
This is a cherry-pick form master, fixing vulnerabilities in previous go libraries


mmamczur:
newer GO has some stricter requirements on what you feed to printf like functions, this is probably the cause of these failures. Fixes would likely be very mechanical so potentially should fall under vulnfix anyway

PR #3096, opened 2026-04-23
Modify the push-multiarch.sh script to tag all sub-images instead of …

mmamczur:
…just the main multiarch manifest.

The script will now use krane to get the manifest and then extract the information about per-arch digests. Then the digests will be used to tag the per-arch images. Later we can reliably remove all tagless images.


kl52752:
Thanks for that! 
/lgtm

PR #3097, opened 2026-04-23
Install jq in cloudbuild.yaml for the build that needs it.

mmamczur:
this is a follow up to #3096

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -39,7 +39,9 @@ steps:
     # This image is built separately because it has
     # additional dependencies (curl, gcloud-cli), and
     # requires `docker buildx` for multiarch building process
-    gcloud auth configure-docker  \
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -39,7 +39,9 @@ steps:
     # This image is built separately because it has
     # additional dependencies (curl, gcloud-cli), and
     # requires `docker buildx` for multiarch building process
-    gcloud auth configure-docker  \
+    # also install jq for manifest parsing
+    apk add jq \
```
08volt: The base image uses `apk --no-cache add` when it's built, meaning there is no local package index left in the container. If you just run `apk add jq` here, it will likely fail. Please change this to `apk add --no-cache jq \` so it fetches the index before installing

PR #3103, opened 2026-05-04
Add NegBinding CRD

anyush:
This CRD will be used to bind already created NetworkEndpointGroups to Services providing an alternative API for the current neg-annotation(`cloud.google.com/neg`)-based.


hack/update-codegen.sh
```diff
@@ -165,4 +165,19 @@ echo "Generating openapi for L4LBConfig v1"
...
   --go-header-file "${SCRIPT_ROOT}"/boilerplate.go.txt \
-  k8s.io/ingress-gce/pkg/apis/l4lbconfig/v1
\ No newline at end of file
+  k8s.io/ingress-gce/pkg/apis/l4lbconfig/v1
+
+  echo "Performing code generation for NetworkEndpointGroupBinding CRD"
+"${CODEGEN_PKG}"/generate-groups.sh \
+  "deepcopy,client,informer,lister" \
+  k8s.io/ingress-gce/pkg/negbinding/client k8s.io/ingress-gce/pkg/apis \
+  "negbinding:v1beta1" \
```
TortillaZHawaii: I want to check with someone with more experience what the version should be (`v1beta1`, `v1` or something else)

pkg/apis/negbinding/v1beta1/types.go
```diff
@@ -0,0 +1,127 @@
...
+	// +k8s:validation:cel[2]:message="Field 'name' must be set"
+	// +k8s:validation:cel[3]:rule="has(self.port)"
+	// +k8s:validation:cel[3]:message="Field 'port' must be set"
+	BackendRef *BackendRefConfig `json:"backendRef"`
+	// NetworkEndpointGroups references NEGs which should be managed by controller
+	// +listType=map
+	// +listMapKey=name
+	// +k8s:validation:maxItems=50
+	// +k8s:validation:cel[0]:rule="self.all(i, self.filter(j, j.subnet == i.subnet).size() <= 1)"
+	// +k8s:validation:cel[0]:message="NegBinding can contain only one NEG name per subnet"
```
swetharepakula, line 53: should we also add a validation that every subnet is only provided once as well?

pkg/apis/negbinding/v1beta1/types.go
```diff
@@ -0,0 +1,127 @@
...
+	NetworkEndpointGroups []SpecNegRef `json:"networkEndpointGroups"`
+}
+
+// BackendRefConfig specifies backendRef for the NetworkEndpointGroupBinding resource
+// +k8s:openapi-gen=true
+type BackendRefConfig struct {
```
swetharepakula, line 59: Should we embed the existing object reference type in K8s core? Then you just add the port field?

pkg/apis/negbinding/v1beta1/types.go
anyush: This is what CEL[0] rule validates - every subnet should be unique (list [negs[0].subnet, negs[1].subnet, ...] can't have duplicates)

Worth to mention that as `listType=map` and `listMapKey=name` - neg names are also unique in this list, so if there is any duplicated subnets, then it means that multiple NEGs specified for same subnet - that's where error message comes from

pkg/apis/negbinding/v1beta1/types.go
anyush: I think it might be a bit misleading as technically provides ability to add namespace, which won't affect anything (as NEGBinding have to be in the same namespace as its backendRef). There is `LocalObjectReference`, but in fact it's just `Name` field, so I don't think that it makes much sense to replace single field by this

pkg/apis/negbinding/v1beta1/types.go
swetharepakula: Thats, I think I meant Name, but you are correct the map validation will ensure that as well. Thanks!

pkg/apis/negbinding/v1beta1/types.go
swetharepakula: I am okay to use the custom type

swetharepakula (commented):
/lgtm
/approve

PR #3104, opened 2026-05-04
Neg controller's syncer refactor

anyush:
WIP on an API which will utilize existing syncer's functionality for syncing NetworkEnpointGroup's endpoints without managing theirs lifecycle (Negs won't be created/deleted by the Neg controller) and using ServiceNetworkEndpointGroup (SvcNeg) CR, therefore refactoring code to decouple interaction with resources from the logic of the syncer.

New API will also make Negs' names per subnet, subnets list, zones per subnet adjustable, which is why e.g. `GenerateSubnetToNegNameMap` also hidden under interface and `isZoneChange`'s and `isSubnetChange`'s logic also adjusted.

Note: this is refactor only - current functionality should not be changed.


pkg/neg/syncers/resourcemanager/neg_resource_manager.go
```diff
@@ -0,0 +1,42 @@
...
+	nodetopologyv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/nodetopology/v1"
+	discovery "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/ingress-gce/pkg/composite"
+
+	"k8s.io/ingress-gce/pkg/network"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+)
+
+type NegResourceManager interface {
```
TortillaZHawaii: https://go.dev/wiki/CodeReviewComments#interfaces

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
+/*
```
TortillaZHawaii, line 1: no tests?

pkg/neg/syncers/resourcemanager/neg_resource_manager.go
anyush: Moved to syncers pkg

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
anyush: All current tests test this implementation (as the only existing one)

pkg/neg/syncers/transaction.go
```diff
@@ -137,34 +136,30 @@ type transactionSyncer struct {
...
 	syncerMetrics *metricscollector.SyncerMetrics,
-	customName bool,
 	log klog.Logger,
 	lpConfig labels.PodLabelPropagationConfig,
 	enableDualStackNEG bool,
 	networkInfo network.NetworkInfo,
-	namer namer.NonDefaultSubnetNEGNamer,
 	negMetrics *metrics.NegMetrics,
+	negResourceManager NegResourceManager,
+
```
TortillaZHawaii, line 158: nit: unnecessary empty line

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+	negName, err := s.GetNegNameForSubnet(subnet)
+	if err != nil {
+		return nil, err
+	}
+	networkEndpointsWithHealthStatus, err := s.cloud.ListNetworkEndpoints(negName, zone, showHealthStatus, s.version, s.logger)
+	if err != nil {
+		// It is possible for a NEG to be missing in a zone without candidate nodes. Log and ignore this error.
+		// NEG not found in a candidate zone is an error.
+		if utils.IsNotFoundError(err) && !candidateZonesMap.Has(zone) {
+			s.logger.Info("Ignoring NotFound error for NEG", "negName", negName, "zone", zone)
```
TortillaZHawaii, line 346: Please add a reason why we ignore it to the logs. This will help support and us when looking at the logs.

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
TortillaZHawaii: I would expect that the tests that verify logic of this should be put in the same `resourcemanager` pkg. 

Currently it looks like the two are too tightly coupled. In an ideal world we would have:
* `SvcNegResourceManager` with the whole contract (methods) tested for different cases in this pkg
* `Syncer` testing it's own logic with mocks of the common interface. If there is an issue with syncer that is common for both implementation of ResourceManager it should be caught here. This will be in `syncers` pkg.
* `SvcNegResourceManager` x `Syncer` that could be done in `resourcemanager_test` package to avoid import cycles or in pkg/neg near manager.

When the new implementation comes we would just add the ResourceManager tests.

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+// getSvcNEGFromStore returns the neg associated with the provided namespace and neg name if it exists otherwise throws an error
+func (s *SvcNegResourceManager) getSvcNEGFromStore() (*negv1beta1.ServiceNetworkEndpointGroup, error) {
+	n, exists, err := s.svcNegLister.GetByKey(fmt.Sprintf("%s/%s", s.negSyncerKey.Namespace, s.negSyncerKey.NegName))
+	if err != nil {
+		return nil, fmt.Errorf("error getting SvcNeg %s/%s from cache: %w", s.negSyncerKey.Namespace, s.negSyncerKey.NegName, err)
+	}
+	if !exists {
+		return nil, fmt.Errorf("SvcNeg %s/%s is not in store", s.negSyncerKey.Namespace, s.negSyncerKey.NegName)
```
TortillaZHawaii, line 383: nit, errors should be lowercase

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	return n.(*negv1beta1.ServiceNetworkEndpointGroup), nil
```
TortillaZHawaii, line 386: can we guarantee that this is always of type `*negv1beta1.ServiceNetworkEndpointGroup`? It will be safer to check the cast with `ok`

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+		if c.Type == conditionType {
+			return c, i, true
+		}
+	}
+
+	return negv1beta1.Condition{}, -1, false
+}
+
+// getInitializedCondition returns the expected initialized condition based on given error
+func (s *SvcNegResourceManager) getInitializedCondition(err error) negv1beta1.Condition {
```
TortillaZHawaii, line 428: nit: you can skip get https://go.dev/doc/effective_go#Getters

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	return s.namer.NonDefaultSubnetNEG(s.negSyncerKey.Namespace, s.negSyncerKey.Name, subnet, s.negSyncerKey.PortTuple.Port), nil
+}
+
+func (s *SvcNegResourceManager) ListSubnets() []nodetopologyv1.SubnetConfig {
+	if !s.networkInfo.IsDefault {
+		subnet, err := utils.KeyName(s.networkInfo.SubnetworkURL)
+		if err != nil {
+			s.logger.Error(err, "Failed to get subnet from NetworkInfo", "subnetworkURL", s.networkInfo.SubnetworkURL)
```
TortillaZHawaii, line 282: Why is it not visible to user? Is it actionable by us?

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+		negMetrics:    negMetrics,
+		logger:        logger,
+		serviceLister: serviceLister,
+		svcNegLister:  svcNegLister,
+		svcNegClient:  svcNegClient,
+	}
+}
+
+func (s *SvcNegResourceManager) UpdateStatusNegsEnsured(negObjs []*composite.NetworkEndpointGroup, errList []error) {
+	if s.svcNegClient == nil {
```
TortillaZHawaii, line 101: Why is it here? When is it expected to happen? Do we want to silently fail here?

pkg/neg/syncers/resourcemanager/svcneg_resource_manager.go
```diff
@@ -0,0 +1,543 @@
...
+	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/network"
+	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/ingress-gce/pkg/utils/patch"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+	"k8s.io/klog/v2"
+)
+
+type SvcNegResourceManager struct {
```
TortillaZHawaii, line 49: nit: https://go.dev/doc/effective_go#package-names

you could have `svcneg` as a package and then use `svcneg.NewManager`

TortillaZHawaii (changes requested):

anyush:
This refactor complicates NEG controller even more - closing in favor of smaller distinct changes.

PR #3109, opened 2026-05-07
fix: classify GCE 403 Forbidden firewall errors as user errors

08volt:
Classifies raw GCE `403 Forbidden` errors on firewall resources (e.g., missing `compute.firewalls.get` IAM permissions or Shared VPC setup issues) as User Errors instead of system errors.

This stops active requeuing loops and loud error-level logs on permission-related sync failures, preventing false-positive SLO burns for L4 ILB availability.

What was changed:
- pkg/utils: Added IsFirewallForbiddenError utility and corresponding unit tests.
- pkg/l4/resources: Integrated IsFirewallForbiddenError into IsUserError and added unit tests.

### DIFF
--- a/pkg/l4/resources/user_error.go
+++ b/pkg/l4/resources/user_error.go
@@ -21,6 +21,7 @@ func IsUserError(err error) bool {
 		utils.IsInvalidLoadBalancerSourceRangesAnnotationError(err) ||
 		utils.IsUnsupportedNetworkTierError(err) ||
 		utils.IsConstraintViolationError(err) ||
+		utils.IsFirewallForbiddenError(err) ||
 		errors.As(err, &firewallErr) ||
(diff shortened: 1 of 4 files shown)


pkg/l4/resources/user_error.go
```diff
@@ -21,6 +21,7 @@ func IsUserError(err error) bool {
 		utils.IsInvalidLoadBalancerSourceRangesAnnotationError(err) ||
 		utils.IsUnsupportedNetworkTierError(err) ||
 		utils.IsConstraintViolationError(err) ||
+		utils.IsFirewallForbiddenError(err) ||
 		errors.As(err, &firewallErr) ||
```
TortillaZHawaii, line 25: It's already there `*firewalls.FirewallXPNError`

PR #3110, opened 2026-05-07
Reduce very noisy logs that bring very little benefit for debugging.

mmamczur:
The resource lock export log was emitted every second and doesnt bring any signal. The Health messages were emitted very often - changed that one to log only if there is a failed check.


kl52752 (commented):
LGTM

PR #3112, opened 2026-05-08
Upgrade Go toolchain to 1.26.3 and Kubernetes dependencies to v0.36.0

FelipeYepez:
Bumps Go toolchain to 1.26.3 and core Kubernetes dependencies to v0.36.0. Applies Go formatting updates and fixes test mock store compatibility with the updated client-go Cache APIs.


pkg/backendconfig/backendconfig_test.go
```diff
@@ -27,6 +27,13 @@ import (
 	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
 )
 
+type fakeStore struct {
```
mmamczur, line 30: I'd just create an interface with only the methods we need and chencte the type of the argument in GetBackendConfigForServicePort()

but this should also work

FelipeYepez: Thanks I'd prefer to avoid modifying those files in this PR since it is just a dependency upgrade.

The compilation failure actually comes from an upstream bug in the Kubernetes `client-go` library. They updated their  cache.Store  interface in K8s 1.36 but forgot to update their own FakeCustomStore test mock utility.

Since it is just an upstream mock utility bug, keeping the fix strictly isolated to our test files using a local wrapper seems like the safest and cleanest approach for now. Let me know what you think

mmamczur: it is a good practice to for the clients in go to use an interface they define themselves with the only stuff they need
I'm not sure if that's a bug or they intended to do it this way

PR #3113, opened 2026-05-11
feat(instancegroups): mitigate GCE API quota exhaustion with TTL-based cache for non-existent instance groups

08volt:
## Motivation and Context
In large or highly active GKE clusters, the L4 controller's background InstanceGroupsController is triggered continuously by Kubernetes node status, heartbeat, and scheduling updates to keep node membership synchronized in the cluster's default unmanaged instance groups (k8s-ig--...).

When an unmanaged instance group is deleted or missing in a zone (e.g., during standard resource updates or before initial provisioning):
* GCE API ListInstancesInInstanceGroup calls return a 404 Not Found error.
* Because node updates occur multiple times per second, the controller repeatedly spams GCE API requests for the missing group, causing a severe 403 Quota Exceeded error storm.
* When a 404 occurred in one zone, the sync loop aborted immediately, leaving all other healthy zones unsynchronized.
* Widespread GCE API quota exhaustion starves all other controllers (Ingress, NetLB, and ILB controllers) from performing operations, leading to cascading service failures and significant cluster availability degradation.

To ensure a safe and progressive rollout in production, this caching optimization is fully gated behind a feature flag (`--enable-ig-ttl-cache`), defaulting to `false`. This ensures zero risk of runtime regression under default configurations.

## Proposed Changes
This PR introduces a robust, thread-safe, TTL-based caching mechanism within the InstanceGroupsManager to suppress redundant queries for missing instance groups, gated behind a feature flag.

* **Feature-Gated Caching**: Wired up the existing `--enable-ig-ttl-cache` flag from `pkg/flags/flags.go` through `pkg/context/context.go`. The TTL caching behavior is fully bypassed unless explicitly enabled.
* **Thread-Safe Cache**: Added a mutex-protected `igNotFoundTime` map (zone -> timestamp) to `pkg/instancegroups/manager.go` (active only when the flag is enabled).
* **Cache Check & Bypass**: Integrated cache checking into the synchronization path. If a zone's unmanaged instance group is missing and still within its 1-minute TTL window (`igNotFoundTTL`), GCE API queries are bypassed for that zone.
* **Fault Tolerance & Partial Success**: Instead of aborting the entire sync cycle when a zone's group is missing, the controller logs the event, skips the zone, and successfully synchronizes all other healthy zones.
* **Self-Healing**: When a zone's instance group list successfully returns VM instances, any existing "missing" cache entry for that zone is instantly cleared.
* **Unit Tests**: Added full coverage in `pkg/instancegroups/manager_test.go`:
  * `TestNodePoolSyncCachingNotFound` verifies cache entry creation, query suppression, expiration, and self-clearing behaviors with the flag enabled.
  * `TestNodePoolSyncCachingDisabled` validates that the GCE API is repeatedly queried on every sync (no caching) when the flag is disabled.

### DIFF
--- a/pkg/instancegroups/manager.go
+++ b/pkg/instancegroups/manager.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	metrics "k8s.io/ingress-gce/pkg/instancegroups/metrics"
@@ -40,6 +41,9 @@ import (
 const (
 	// State string required by gce library to list all instances.
 	allInstances = "ALL"
+
+	// igNotFoundTTL is the duration to cache the non-existence of an instance group.
+	igNotFoundTTL = 1 * time.Minute
 )
 
 // manager implements Manager.
@@ -51,6 +55,10 @@ type manager struct {
 	instanceLinkFormat string
 	maxIGSize          int
 	readOnlyMode       bool
+	enableIGTTLCache   bool
+
+	mu             sync.Mutex
+	igNotFoundTime map[string]time.Time // key: zone, value: last 404 time
 }
 
 type recorderSource interface {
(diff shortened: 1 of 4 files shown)


pkg/instancegroups/manager.go
```diff
@@ -371,14 +405,24 @@ func (m *manager) Sync(nodes []string, logger klog.Logger) (err error) {
...
 
 		gceNodes := sets.NewString()
 		instances, err := m.cloud.ListInstancesInInstanceGroup(igName, zone, allInstances)
 		if err != nil {
+			if utils.IsHTTPErrorCode(err, http.StatusNotFound) {
+				iglogger.V(3).Info("Instance group not found, caching", "zone", zone, "igName", igName)
+				m.setIGNotFoundCache(zone)
+			}
 			iglogger.Error(err, "Failed to list instance from instance group", "zone", zone, "igName", igName)
 			return err
```
mmamczur, line 436: shouldn't we not return error if this happens? just continue for remaining zones?

08volt: Yes, fixed.

PR #3121, opened 2026-05-27
Create TopologyProvider for NEG controller's syncers

anyush:
Create interface for providing target subnets/zones where NEGs should be created/managed by NEG controller, so it will be possible to provide these in future APIs instead of solely rely on the NodeTopology CR.

Also makes zones list subnet-aware for syncer - for different subnets different subsets of zones can be provided.

EDIT:
Overview:
* `ZoneGetter` accessed through `TopologyProvider` by the `transactionSyncer`
* Zones list provided by `TopologyProvider` is subnet-dependent (possible to have different zones for different subnets)
* `transactionSyncer.IsZoneChange` checks zones per subnet, which leads to verifying subnets' change as well - `transactionSyncer.IsSubnetChange` removed, `transactionSyncer.IsZoneChange` updated and renamed into `transactionSyncer.IsLocationChange`

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -86,7 +86,7 @@ type transactionSyncer struct {
 	svcNegLister        cache.Indexer
 	recorder            record.EventRecorder
 	cloud               negtypes.NetworkEndpointGroupCloud
-	zoneGetter          *zonegetter.ZoneGetter
+	topologyProvider    negtypes.TopologyProvider
 	endpointsCalculator negtypes.NetworkEndpointsCalculator
 
 	// retry handles back off retry for NEG API operations
@@ -148,7 +148,7 @@ func NewTransactionSyncer(
 	negSyncerKey negtypes.NegSyncerKey,
 	recorder record.EventRecorder,
 	cloud negtypes.NetworkEndpointGroupCloud,
-	zoneGetter *zonegetter.ZoneGetter,
+	topologyProvider negtypes.TopologyProvider,
 	podLister cache.Indexer,
 	serviceLister cache.Indexer,
 	endpointSliceLister cache.Indexer,
@@ -182,7 +182,7 @@ func NewTransactionSyncer(
 		svcNegLister:              svcNegLister,
 		recorder:                  recorder,
 		cloud:                     cloud,
-		zoneGetter:                zoneGetter,
+		topologyProvider:          topologyProvider,
 		endpointsCalculator:       epc,
 		reflector:                 reflector,
 		kubeSystemUID:             kubeSystemUID,
@@ -300,14 +300,14 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 	s.logger.V(2).Info("Sync NEG", "negSyncerKey", s.NegSyncerKey.String(), "endpointsCalculatorMode", s.endpointsCalculator.Mode())
 
-	subnetConfigs := s.zoneGetter.ListSubnets(s.logger)
+	subnetConfigs := s.topologyProvider.ListSubnets(s.logger)
 	subnetToNegMapping, err := s.generateSubnetToNegNameMap(subnetConfigs)
 	if err != nil {
 		s.logger.Error(err, "failed to generate subnet to neg name mapping")
 		return err
 	}
 
-	currentMap, currentPodLabelMap, drainingEndpoints, err := retrieveExistingZoneNetworkEndpointMap(subnetToNegMapping, s.zoneGetter, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, s.logger, s.negMetrics, needInitDrainStatus, s.NegSyncerKey.IncludeDrainNodesL4Local)
+	currentMap, currentPodLabelMap, drainingEndpoints, err := retrieveExistingZoneNetworkEndpointMap(subnetToNegMapping, s.topologyProvider, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, s.logger, s.negMetrics, needInitDrainStatus, s.NegSyncerKey.IncludeDrainNodesL4Local)
 	if err != nil {
 		return fmt.Errorf("%w: %w", negtypes.ErrCurrentNegEPNotFound, err)
 	}
@@ -532,7 +532,7 @@ func (s *transactionSyncer) candidateNodeFilter() zonegetter.Filter {
 // ensureNetworkEndpointGroups ensures NEGs are created and configured correctly in the corresponding zones.
 func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	// NEGs should be created in zones with candidate nodes only.
-	zonesPerSubnet, err := s.zoneGetter.ListZonesPerSubnet(s.candidateNodeFilter(), s.logger)
+	zonesPerSubnet, err := s.topologyProvider.ListZonesPerSubnet(s.candidateNodeFilter(), s.logger)
 	if err != nil {
 		return err
 	}
@@ -556,7 +556,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 		// zoneGetter.
 
 		// List all existing subnets from the cluster.
-		subnetConfigs = s.zoneGetter.ListSubnets(s.logger)
+		subnetConfigs = s.topologyProvider.ListSubnets(s.logger)
 	} else {
 		// This is the multi-networking case where the VPC under consideration
 		// is not the default. Use the pre configured subnet from the
@@ -573,7 +573,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	for _, subnetConfig := range subnetConfigs {
 		zones, ok := zonesPerSubnet[subnetConfig.Name]
 		if !ok {
-			// s.zoneGetter.ListSubnets and s.zoneGetter.ListZonesPerSubnet should return same set of subnets.
+			// s.topologyProvider.ListSubnets and s.topologyProvider.ListZonesPerSubnet should return same set of subnets.
 			// Therefore this condition should be true only for multi-networking where we don't want NEGs in default subnet
 			continue
 		}
@@ -940,7 +940,7 @@ func (s *transactionSyncer) isTopologyChange() bool {
 		existingSubnetZones[subnetName].Insert(id.Key.Zone)
 	}
 
-	wantSubnetZones, err := s.zoneGetter.ListZonesPerSubnet(s.candidateNodeFilter(), s.logger)
+	wantSubnetZones, err := s.topologyProvider.ListZonesPerSubnet(s.candidateNodeFilter(), s.logger)
 	if err != nil {
 		s.logger.Error(err, "unable to list zones")
 		s.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
@@ -1045,7 +1045,7 @@ func (s *transactionSyncer) updateInitStatus(negObjRefs []negv1beta1.NegObjectRe
 
 	neg := origNeg.DeepCopy()
 	if flags.F.EnableMultiSubnetClusterPhase1 {
-		nonActiveNegRefs := getNonActiveNegRefs(origNeg.Status.NetworkEndpointGroups, negObjRefs, s.zoneGetter.ListSubnets(s.logger), s.networkInfo.SubnetworkURL, s.logger)
+		nonActiveNegRefs := getNonActiveNegRefs(origNeg.Status.NetworkEndpointGroups, negObjRefs, s.topologyProvider.ListSubnets(s.logger), s.networkInfo.SubnetworkURL, s.logger)
 		negObjRefs = append(negObjRefs, nonActiveNegRefs...)
 	}
 	neg.Status.NetworkEndpointGroups = negObjRefs
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -532,7 +532,7 @@ func (s *transactionSyncer) candidateNodeFilter() zonegetter.Filter {
 // ensureNetworkEndpointGroups ensures NEGs are created and configured correctly in the corresponding zones.
 func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	// NEGs should be created in zones with candidate nodes only.
-	zones, err := s.zoneGetter.ListZones(s.candidateNodeFilter(), s.logger)
+	zonesPerSubnet, err := s.zoneGetter.ListZonesPerSubnet(s.candidateNodeFilter(), s.logger)
 	if err != nil {
 		return err
 	}
@@ -571,6 +571,12 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	}
 
 	for _, subnetConfig := range subnetConfigs {
+		zones, ok := zonesPerSubnet[subnetConfig.Name]
+		if !ok {
+			// s.zoneGetter.ListSubnets and s.zoneGetter.ListZonesPerSubnet should return same set of subnets.
+			// Therefore this condition should be true only for multi-networking where we don't want NEGs in default subnet
+			continue
+		}
 		negName := s.NegSyncerKey.NegName
 		networkInfo := s.networkInfo
 
@@ -594,7 +600,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 			networkInfo.SubnetworkURL = cloud.SelfLink(meta.VersionGA, resourceID.ProjectID, resourceID.Resource, resourceID.Key)
 		}
 
-		for _, zone := range zones {
+		for zone := range zones {
 			var negObj negv1beta1.NegObjectReference
 			negObj, err = ensureNetworkEndpointGroup(
 				s.Namespace,
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -54,6 +54,7 @@ import (
 	"k8s.io/ingress-gce/pkg/neg/syncers/dualstack"
 	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
+	"k8s.io/ingress-gce/pkg/neg/types/shared"
 	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
 	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/ingress-gce/pkg/utils/zonegetter"
@@ -285,14 +286,13 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		s.logger.Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String(), "syncerStopped", isStopped, "syncerShuttingDown", isShuttingDown)
 		return nil
 	}
-	zoneChange := s.isZoneChange()
-	subnetChange := s.isSubnetChange()
+	topologyChange := s.isTopologyChange()
 	// Decide if endpoint health should be retrieved when listing endpoints.
 	// Only matters for L4 Local mode.
 	needInitDrainStatus := s.needInit && s.enableL4NEGDetachCancel && s.endpointsCalculator.Mode() == negtypes.L4LocalMode
 
-	if s.needInit || zoneChange || subnetChange {
-		s.logger.Info("Need to ensure network endpoint groups", "needInit", s.needInit, "zoneChange", zoneChange, "subnetChange", subnetChange)
+	if s.needInit || topologyChange {
+		s.logger.Info("Need to ensure network endpoint groups", "needInit", s.needInit, "topologyChange", topologyChange)
 		if err := s.ensureNetworkEndpointGroups(); err != nil {
 			return fmt.Errorf("%w: %v", negtypes.ErrNegNotFound, err)
 		}
@@ -900,9 +900,9 @@ func (s *transactionSyncer) commitPods(endpointMap map[negtypes.NEGLocation]negt
 	}
 }
 
-// isZoneChange returns true if a zone change has occurred by comparing which zones the nodes are in
-// with the zones that NEGs are initialized in
-func (s *transactionSyncer) isZoneChange() bool {
+// isTopologyChange returns true if a topology (zones/subnets) change has occurred by comparing which zones/subnets have NEGs are initialized in
+// to zones/subnets NEGs are expected to be in
+func (s *transactionSyncer) isTopologyChange() bool {
 	negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.NegSyncerKey.NegName)
 	if err != nil {
 		s.logger.Error(err, "unable to retrieve neg from the store", "neg", klog.KRef(s.Namespace, s.NegName))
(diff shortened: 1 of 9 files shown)


TortillaZHawaii:
/ok-to-test

🥔

pkg/neg/syncers/transaction_test.go
```diff
@@ -321,12 +338,7 @@ func TestTransactionSyncNetworkEndpointsMSC(t *testing.T) {
 		// Verify the NEGs are created as expected
 		ret, _ := transactionSyncer.cloud.AggregatedListNetworkEndpointGroup(transactionSyncer.NegSyncerKey.GetAPIVersion(), klog.TODO())
 		// Though the test cases below only add instances in zone1 and zone2, NEGs will be created in zone3 or zone4 as well since fakeZoneGetter includes those zones.
```
TortillaZHawaii: move this comment with expectZones

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -1567,3 +1567,62 @@ func TestLegacyListSubnets(t *testing.T) {
...
 }
+
+func TestListZonesForSubnet(t *testing.T) {
+	t.Parallel()
+
+	nodeInformer := FakeNodeInformer()
+	PopulateFakeNodeInformer(nodeInformer, false)
+	zoneGetter, err := NewFakeZoneGetter(nodeInformer, FakeNodeTopologyInformer(), defaultTestSubnetURL, false)
+	if err != nil {
+		t.Fatalf("failed to initialize zone getter")
```
TortillaZHawaii, line 1692: adding err message would be nice

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -1567,3 +1567,62 @@ func TestLegacyListSubnets(t *testing.T) {
...
+	}
+	testCases := []struct {
+		desc      string
+		filter    Filter
+		subnet    string
+		expectLen int
```
TortillaZHawaii: why len not content? you can use cmp.Diff with https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -220,19 +220,59 @@ func (z *ZoneGetter) ListNodesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+		allZones = allZones.Union(zones)
+	}
+	return allZones.UnsortedList(), nil
 }
 
 // ListZonesInDefaultSubnet returns a list of zones containing nodes in the default subnet that satisfy the given
 // node filtering mode.
 func (z *ZoneGetter) ListZonesInDefaultSubnet(filter Filter, logger klog.Logger) ([]string, error) {
-	return z.listZones(filter, true, logger)
+	if z.mode == NonGCP {
```
TortillaZHawaii: are the different mode behaviors covered in unit tests? I couldn't find any for NonGCP mode tests for ListZonesInDefaultSubnet nor ListZonesForSubnet, just for ZoneAndSubnetForNode.

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -220,19 +220,59 @@ func (z *ZoneGetter) ListNodesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+	if err != nil {
+		return nil, err
+	}
+	if zones, ok := subnetToZones[subnet]; ok {
+		return zones.UnsortedList(), nil
+	}
+	return []string{}, nil
+}
+
+func (z *ZoneGetter) listZones(filter Filter, defaultSubnetOnly bool, logger klog.Logger) (map[string]sets.Set[string], error) {
```
TortillaZHawaii: add a comment on what it returns, it is not obvious from just the signature which string is which

pkg/neg/types/interfaces.go
```diff
@@ -94,3 +97,12 @@ type NetworkEndpointsCalculator interface {
...
 }
+
+// TopologyProvider is an interface for looking up zone and subnet information for NEG syncer.
+type TopologyProvider interface {
+	ListZonesForSubnet(filter zonegetter.Filter, subnet string, logger klog.Logger) ([]string, error)
+	ListNodes(filter zonegetter.Filter, logger klog.Logger) ([]*apiv1.Node, error)
+	ZoneAndSubnetForNode(name string, logger klog.Logger) (string, string, error)
+	IsNodeSelectedByFilter(node *apiv1.Node, filter zonegetter.Filter, filterLogger klog.Logger) bool
+	ListSubnets(logger klog.Logger) []nodetopologyv1.SubnetConfig
+}
```
TortillaZHawaii, line 107: This whole file should not exist, let's not add to it.

https://dave.cheney.net/2016/08/20/solid-go-design
https://go.dev/wiki/CodeReviewComments#interfaces

pkg/neg/syncers/transaction.go
TortillaZHawaii: Explain in the comment that it doesn't cover when zone is changed in a subnet that ceased to exist or run `deepequal` on two maps.

TortillaZHawaii (changes requested):
small changes required, I hope Vadym catches anything NEG domain related

pkg/neg/syncers/transaction.go
anyush: Used `DeepEqual`. In fact `isSubnetChange` is irrelevant (subnet changes also covered by `isZoneChange`) - will remove in future PR, so it won't complicate this one more.

pkg/neg/syncers/transaction_test.go
anyush: Done

pkg/utils/zonegetter/zone_getter.go
anyush: Adjusted test of NonGCPZoneGetter - now covered

pkg/utils/zonegetter/zone_getter.go
anyush: Done

pkg/utils/zonegetter/zone_getter_test.go
anyush: Done

pkg/neg/types/interfaces.go
anyush: I strongly believe it's much cleaner than adding potentially tens of interfaces in transaction.go.

Think we can treat `pkg/neg` as single package as everything under it is tightly coupled and is a part of NEG controller (quite unlikely to be used independently)

pkg/utils/zonegetter/zone_getter_test.go
anyush: Updated

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -245,18 +286,31 @@ func (z *ZoneGetter) listZones(filter Filter, defaultSubnetOnly bool, logger klo
...
+	subnetToZones := make(map[string]sets.Set[string])
 	for _, n := range nodes {
 		zone, err := getZone(n)
 		if err != nil || zone == EmptyZone {
 			filterLogger.Error(err, "Failed to get zone from providerID", "nodeName", n.Name)
 			continue
 		}
-		zones.Insert(zone)
+
+		subnet := ""
```
midavadim: here we can endup with a map where a key is "", 

subnetToZones[""] = [zone]

it is correct behabiour, do we handle it correctly in caller functions?

pkg/neg/syncers/transaction.go
```diff
@@ -888,26 +889,43 @@ func (s *transactionSyncer) isZoneChange() bool {
 		return false
 	}
 
-	existingZones := sets.NewString()
+	existingSubnetZones := make(map[string]sets.Set[string])
 	for _, ref := range negCR.Status.NetworkEndpointGroups {
+		subnetURL := s.networkInfo.SubnetworkURL
```
midavadim: I suppose you wanted to do something like this:

		subnetURL := ref.SubnetURL
		if subnetURL == "" {
			subnetURL = s.networkInfo.SubnetworkURL
		}

pkg/neg/types/interfaces.go
```diff
@@ -94,3 +97,9 @@ type NetworkEndpointsCalculator interface {
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+// TopologyProvider is an interface for looking up zone and subnet information for NEG syncer.
+type TopologyProvider interface {
+	ListZonesForSubnet(filter zonegetter.Filter, subnet string, logger klog.Logger) ([]string, error)
+	ListSubnets(logger klog.Logger) []nodetopologyv1.SubnetConfig
```
midavadim, line 105: consider to have methid which returns sunbet - > zones mapping, so it could be directly used in caller functions. it should simplify them and avoid extra calls of NodeListers

pkg/neg/types/interfaces.go
```diff
@@ -94,3 +97,9 @@ type NetworkEndpointsCalculator interface {
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+// TopologyProvider is an interface for looking up zone and subnet information for NEG syncer.
```
midavadim, line 102: extend description slightly more

pkg/neg/syncers/transaction.go
anyush: But
```
subnetURL := s.networkInfo.SubnetworkURL
if ref.SubnetURL != "" {
	subnetURL = ref.SubnetURL
}
```
will do pretty much the same:
* ref.SubnetURL != "" -> subnetURL == ref.SubnetURL
* ref.SubnetURL == "" -> subnetURL == s.networkInfo.SubnetworkURL

pkg/utils/zonegetter/zone_getter.go
anyush: It will be "" for `z.mode == Legacy` only, which later will return all zones despite the subnets (including with this empty subnet)

pkg/neg/types/interfaces.go
anyush: Done

pkg/neg/syncers/transaction.go
```diff
@@ -270,29 +271,28 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		s.logger.Info("Skip syncing NEG", "negSyncerKey", s.NegSyncerKey.String(), "syncerStopped", isStopped, "syncerShuttingDown", isShuttingDown)
 		return nil
 	}
-	zoneChange := s.isZoneChange()
-	subnetChange := s.isSubnetChange()
+	locationChange := s.isLocationChange()
```
midavadim: I would change it to "isTopologyChange"

pkg/neg/syncers/transaction.go
```diff
@@ -548,63 +542,73 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
-				s.logger.Error(err, "Unable to get the name of the additional NEG based on the subnet name", "subnetName", subnetConfig.Name)
-				errList = append(errList, err)
+	// NEGs should be created in zones with candidate nodes only.
+	zonesPerSubnet, err := s.topologyProvider.ListZonesPerSubnet(negtypes.NodeFilterForEndpointCalculatorMode(s.EpCalculatorMode), s.logger)
+	if err != nil {
+		errList = append(errList, err)
+	} else {
+		for _, subnetConfig := range subnetConfigs {
+			zones, ok := zonesPerSubnet[subnetConfig.Name]
+			if !ok {
```
midavadim: Add comment when it could happen (in multinet)

pkg/neg/types/interfaces.go
anyush: Done

pkg/neg/syncers/transaction.go
anyush: Done

pkg/neg/syncers/transaction.go
anyush: Done

midavadim:
LGTM

midavadim (approved):

pkg/neg/syncers/transaction.go
```diff
@@ -910,26 +911,43 @@ func (s *transactionSyncer) isZoneChange() bool {
...
+			s.logger.Error(err, "unable to list zones for subnet", "subnet", subnet.Name)
+			s.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
+			return false
+		}
+		expectedSubnetZones[subnet.Name] = sets.New[string](zones...)
 	}
-	currZones := sets.NewString(zones...)
 
-	return !currZones.Equal(existingZones)
+	return !reflect.DeepEqual(expectedSubnetZones, existingSubnetZones)
```
swetharepakula: this logic is not optimized for the general use case where zones are standardized across the subnets.

2 approaches:
1. We make it so that zones need to be uniform across subnets in NEG Binding
2. Move this logic to the topology provider interface (so add a method on zonegetter), that allows us to optimize the logic based

My preference is for (1), but we can make the controller a bit more robust and do (2)

pkg/neg/syncers/transaction.go
```diff
@@ -531,12 +532,6 @@ func (s *transactionSyncer) candidateNodeFilter() zonegetter.Filter {
 
 // ensureNetworkEndpointGroups ensures NEGs are created and configured correctly in the corresponding zones.
 func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
```
swetharepakula, line 533: is this function going to be used in the neg binding case? In that case, we shouldn't be ensuring Network Endpoint Groups. Since that is the case, I would not move the zone listing down on a per subnet basis as in the non neg binding case, zones are the same for all subnets.

pkg/neg/syncers/transaction.go
```diff
@@ -571,6 +566,12 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	}
 
 	for _, subnetConfig := range subnetConfigs {
+		// NEGs should be created in zones with candidate nodes only.
+		zones, err := s.topologyProvider.ListZonesForSubnet(s.candidateNodeFilter(), subnetConfig.Name, s.logger)
+		if err != nil {
+			errList = append(errList, err)
+			continue
+		}
```
swetharepakula, line 579: in the general case, this is a subtle change in behavior. It means that when we add a subnet in the MSC case, we won't automatically create a NEG in all the zones the cluster is in, wait for the zones to be expanded to in the subnet.

I would be careful about this change, because if any customer is expecting this / depending on this subtle behavior it can affect their workflows.

We also should be moving zone getter to depend on NodeTopology for the zones and remove the need to calculate it ourselves. This will further codify that in the general case we expect the list of zones to be uniform for every subnet

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -229,19 +229,60 @@ func (z *ZoneGetter) ListNodesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+		return nil, err
+	}
+	if zones, ok := subnetToZones[subnet]; ok {
+		return zones.UnsortedList(), nil
+	}
+	return []string{}, nil
+}
+
+// list zones returns zones where filtered nodes are found by subnet (map[subnetName]sets.Set[zoneName])
+func (z *ZoneGetter) listZones(filter Filter, defaultSubnetOnly bool, logger klog.Logger) (map[string]sets.Set[string], error) {
```
swetharepakula: perhaps keep listZones and create a subnetToZones helper instead?

ListZones is a bit misleading in terms of what is being returned here.

swetharepakula (commented):
A few general comments (I have only reviewed the first commit):

* The commit messages need to be more descriptive. For example in the first commit, you don't mention the zone getter refactor or the why behind the changes. Or the changes of looking at zones per subnet. The why needs to be documented
* I like to review each commit individually (as when you look at history, this is what you will take a look at). So if you create something in one and delete it a following commit, please properly squash or reorganize your commits
* when refactoring or introducing the interface (like in the first commit), I would suggesting making an in place change first and then adding logic changes in a subsequent commit. Try not to mix refactoring with other changes.

pkg/utils/zonegetter/zone_getter.go
anyush: Done

pkg/neg/syncers/transaction.go
anyush: `transactionSyncer.ensureNetworkEndpointGroups` in fact collects topology and calls  `pkg/neg/syncers/utils.go:ensureNetworkEndpointGroup`, which is not only creates, but:
* verifies existence
* verifies description
* creates/recreates if needed
* verifies existence

This highly overlap with what we expect from binding as well (collect topology, verify existence, verify description (different one, but still needs verification)), therefore I think it's better to reuse logic with minor flag-gated changes. There you can see my proposition to this: https://github.com/kubernetes/ingress-gce/pull/3142

pkg/neg/syncers/transaction.go
anyush: As I see we can't go with 1st option, as NEGBinding API created in a way which allows providing different sets of zones (and if I remember correctly it was discussed directly).

Moving this check into topology provider will slightly speed up check for current API (mainly by not needing to create subnet-zone map, but we still will need to compare zones' set for every subnet with expected list of zones), but I think this optimization will be negligible and only will produce almost duplicated code.

I've only changed comparison a bit, so now it won't add reflect's overhead.

pkg/neg/syncers/transaction.go
anyush: I've changed how ZoneGetter will return zones per subnet - in fact it will be the way it was before: it will get subnets, zones and return product of them (will return list of all zones for every subnet)

pkg/neg/syncers/transaction.go
TortillaZHawaii: I agree with Vadym on this one.  Extra `not` increases cognitive load or in other words not having `not` will decrease cognitive load.

pkg/neg/syncers/transaction.go
```diff
@@ -900,73 +905,67 @@ func (s *transactionSyncer) commitPods(endpointMap map[negtypes.NEGLocation]negt
...
 		}
-		existingZones.Insert(id.Key.Zone)
+		if _, ok := existingSubnetZones[subnetName]; !ok {
+			existingSubnetZones[subnetName] = sets.New[string]()
+		}
+		existingSubnetZones[subnetName].Insert(id.Key.Zone)
 	}
 
-	zones, err := s.zoneGetter.ListZones(s.candidateNodeFilter(), s.logger)
+	expectedSubnetZones := make(map[string]sets.Set[string])
```
TortillaZHawaii: I might be too tired to review this, but `existingSubnetZones` looks very similar to `expectedSubnetZones`. Consider changing one prefix to `have`, `old` or `want`, so there is more difference.

pkg/neg/syncers/transaction.go
```diff
@@ -900,73 +905,67 @@ func (s *transactionSyncer) commitPods(endpointMap map[negtypes.NEGLocation]negt
...
-	return !currZones.Equal(existingZones)
+	return !subnetZonesEqual(expectedSubnetZones, existingSubnetZones)
 }
 
-func (s *transactionSyncer) isSubnetChange() bool {
-	negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.NegSyncerKey.NegName)
-	if err != nil {
-		s.logger.Error(err, "unable to retrieve neg from the store", "neg", klog.KRef(s.Namespace, s.NegName))
-		s.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
+func subnetZonesEqual(a, b map[string]sets.Set[string]) bool {
```
TortillaZHawaii: `map[string]sets.Set[string]` is used so often as `subnetZones`, maybe it should be a type

pkg/neg/syncers/transaction_test.go
```diff
@@ -301,10 +301,28 @@ func TestTransactionSyncNetworkEndpointsMSC(t *testing.T) {
...
-		zonegetter.SetNodeTopologyHasSynced(transactionSyncer.zoneGetter, func() bool { return true })
+		zonegetter.SetNodeTopologyHasSynced(transactionSyncer.topologyProvider.(*zonegetter.ZoneGetter), func() bool { return true })
+
+		// Though the test cases below only add instances in zone1 and zone2, NEGs will be created in zone3 or zone4 as well since fakeZoneGetter includes those zones.
+		var expectZones []string
+		if testNegType == negtypes.VmIpEndpointType {
+			expectZones = []string{negtypes.TestZone1, negtypes.TestZone2, negtypes.TestZone3}
+		} else {
+			expectZones = []string{negtypes.TestZone1, negtypes.TestZone2, negtypes.TestZone4}
+		}
```
TortillaZHawaii: go way to write this is:

```suggestion
		expectZones := []string{negtypes.TestZone1, negtypes.TestZone2, negtypes.TestZone4}
		if testNegType == negtypes.VmIpEndpointType {
			expectZones = []string{negtypes.TestZone1, negtypes.TestZone2, negtypes.TestZone3}
		}
```

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -1639,3 +1661,111 @@ func TestLegacyListSubnets(t *testing.T) {
...
+			expectZonesAllSubnets: map[string][]string{
+				"default":     {"zone1", "zone2", "zone3", "zone5", "zone6", "zone7", "zone8"},
+				"non-default": {"zone1", "zone2", "zone3", "zone5", "zone6", "zone7", "zone8"},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			for _, onlyIncludeDefaultSubnetNodes := range []bool{true, false} {
```
TortillaZHawaii: you can add a t.Run under onlyIncludeDefaultSubnetNodes for default and all subnets, so that one doesn't block the other

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -1639,3 +1661,111 @@ func TestLegacyListSubnets(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("expect err = nil, but got %v", err)
+				}
+
+				var expected map[string][]string
+				if onlyIncludeDefaultSubnetNodes {
+					expected = tc.expectZonesOnlyDefault
+				} else {
+					expected = tc.expectZonesAllSubnets
+				}
```
TortillaZHawaii: go way to write this is


```suggestion
				expected := tc.expectZonesAllSubnets
				if onlyIncludeDefaultSubnetNodes {
					expected = tc.expectZonesOnlyDefault
				}
```

or even more common for tests is `want`

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -1639,3 +1661,111 @@ func TestLegacyListSubnets(t *testing.T) {
...
+
+				if !reflect.DeepEqual(expected, zonesPerSubnet) {
```
TortillaZHawaii: In tests you can use https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Diff-Testing

pkg/utils/zonegetter/zone_getter_test.go
```diff
@@ -724,13 +724,35 @@ func TestNonGCPZoneGetter(t *testing.T) {
...
-		t.Errorf("expect list zones = %v, but got %v", expectZones, ret)
+		t.Errorf("expect ListZones = %v, but got %v", expectZones, ret)
+	}
+
+	// ListZonesInDefaultSubnet
+	retInDefault, err := zoneGetter.ListZonesInDefaultSubnet(AllNodesFilter, klog.TODO())
+	if err != nil {
+		t.Errorf("expect err = nil for ListZonesInDefaultSubnet, but got %v", err)
+	}
+	if !reflect.DeepEqual(expectZones, retInDefault) {
```
TortillaZHawaii, line 746: what does `ret` stand for?

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -238,6 +238,25 @@ func (z *ZoneGetter) ListZonesInDefaultSubnet(filter Filter, logger klog.Logger)
 	return z.listZones(filter, true, logger)
 }
 
+// ListZonesPerSubnet returns a list of zones containing nodes that satisfy the given node filtering mode per subnet.
+func (z *ZoneGetter) ListZonesPerSubnet(filter Filter, logger klog.Logger) (map[string][]string, error) {
```
TortillaZHawaii: this return value strongly makes me think `map[string][]string` should be a type

pkg/neg/syncers/transaction_test.go
```diff
@@ -4695,3 +4662,131 @@ func generateExpectedNegObjReferences(t *testing.T, cloud negtypes.NetworkEndpoi
...
+			a:              map[string]sets.Set[string]{},
+			b:              map[string]sets.Set[string]{},
+			expectedResult: true,
+		},
+		{
+			desc:           "one nil, one empty",
+			a:              nil,
+			b:              map[string]sets.Set[string]{},
+			expectedResult: true,
+		},
```
TortillaZHawaii: I see in the other file that you handle the case when `len(zones) == 0` for subnet by not adding them, but let's add it for completeness sake as a test case as well

TortillaZHawaii (changes requested):
if this is is time sensitive I suggest adding someone else to follow up on reviews, otherwise I can revisit this next week

pkg/utils/zonegetter/zone_getter_test.go
anyush: Based on existed code `ret, err := zoneGetter.ListZones(AllNodesFilter, klog.TODO())`, which, guess, stands for "returned value"

pkg/neg/syncers/transaction.go
anyush: K, changed

pkg/neg/syncers/transaction.go
anyush: renamed expected to want

pkg/neg/syncers/transaction_test.go
anyush: Changed

pkg/utils/zonegetter/zone_getter_test.go
anyush: changed

pkg/utils/zonegetter/zone_getter_test.go
anyush: Done

pkg/neg/syncers/transaction.go
anyush: Done

pkg/utils/zonegetter/zone_getter_test.go
anyush: Done

pkg/utils/zonegetter/zone_getter.go
anyush: Used ZonesPerSubnetMap (map[string]sets.Set[string])

pkg/neg/syncers/transaction_test.go
anyush: Now instead of adding check for len not populating subnet if there're no zones

PR #3122, opened 2026-05-28
Retry deletion loop when resource is used by another resource.

kl52752:
When BackendService deletion failed with cause resource is used by another resource we should return the error and retry the loop. In this change we also are fixing typo in chekcing error.


mmamczur:
I recommend adding a unit test but otherwise lgtm (unhold when you want to merge)

/hold
/lgtm

PR #3123, opened 2026-05-28
update go to 1.25.10

kl52752:


build/build.sh
```diff
@@ -47,7 +47,7 @@ BIN_PKG="$PKG/cmd/$(basename ${TARGET})"
 LD_FLAGS="-X ${PKG}/pkg/version.Version=${VERSION} -X ${PKG}/pkg/version.GitCommit=${GIT_COMMIT}"
 
 if echo "${TARGET}" | grep '.*-test$'; then
-  go test -c -ldflags "${LD_FLAGS}" -o "${TARGET}" "${BIN_PKG}"
+  go test -buildvcs=false -c -ldflags "${LD_FLAGS}" -o "${TARGET}" "${BIN_PKG}"
```
mmamczur, line 50: why do we need this change?
I think there could be some issue that only manifests during local development, like discepancy of git versions between the build image and local env. I'm not sure we need this flag

build/test.sh
```diff
@@ -51,24 +51,24 @@ echo "PASS"
...
 # Set gobin here, install dependencies after this
-cd "hack/tools"
-export GOBIN=$PWD/bin
-export PATH=$GOBIN:$PATH
-# Install golangci-lint
-echo "Installing golangci-lint"
-echo
-go install github.com/golangci/golangci-lint/cmd/golangci-lint > /dev/null
-cd "../.."
+# cd "hack/tools"
```
mmamczur, line 54: why do we need to comment this out?

PR #3124, opened 2026-05-29
Wrap NEG syncer status handling with interface

anyush:
Allow reporting managed NEGs and errors raised during the sync in different way than into ServiceNEG CR (e.g. into different CR) - to do so different implementation of the introduced `NegStatusHandler`.

Done as preparation for the NEG Binding feature which will manage NEGs using NEG syncer, but will use `NetworkEndpointGroupBinding` CR for status reporting instead of SvcNEG.  NEGBinding CRD: https://github.com/kubernetes/ingress-gce/pull/3103

EDIT:
Overview:
* All current `ServiceNetworkEndpointGroup`-related functionality hidden under `NegStatusHandler` interface with `SvcNegStatusHandler` implementation
  * `updateInitStatus` (called after NEGs ensured) -> `SvcNegStatusHandler.ReportStatus`
  * `updateStatus` (called after sync is completed) -> `SvcNegStatusHandler.ReportSyncStatus`
  * getting list of already ensured (and reported on `SvcNEG`) NEGs in `isTopologyChange` -> `SvcNegStatusHandler.SubnetToZonesMap`
  * getting last sync time in `transactionSyncer.computeEPSStaleness` -> `SvcNegStatusHandler.LastSyncTime`
  * all helper functions (`getNegFromStore`, `findCondition`, `patchSvcNegStatus, etc.) -> receivers on `SvcNegStatusHandler` (used only for the above functionality, therefore private and not accessible outside from status handler)
* Code related to `SvcNEGs` (listed above) mostly moved from `transaction.go` to `svcneg.go` without changes (`SubnetToZonesMap` is an exception because of returned format, idea - getting `SvcNEG` and going through `Status.NetworkEndpointGroups` - same as was previously in `isTopologyChange`).


pkg/neg/syncers/transaction.go
```diff
@@ -1101,19 +1009,17 @@ func convertUntypedToEPS(endpointSliceUntyped []interface{}) []*discovery.Endpoi
 }
 
 func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.EndpointSlice) {
-	negCR, err := getNegFromStore(s.svcNegLister, s.Namespace, s.NegSyncerKey.NegName)
+	lastSyncTimestamp, err := s.statusHandler.LastSyncTime()
 	if err != nil {
-		s.logger.Error(err, "unable to retrieve neg from the store", "neg", klog.KRef(s.Namespace, s.NegName))
-		s.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
```
midavadim, line 1110: is it expected that metric call is removed from here?

pkg/neg/syncers/negstatushandler/svcneg.go
```diff
@@ -0,0 +1,349 @@
...
+		logger:       logger.WithValues("svcneg", klog.KRef(namespace, svcNegName)),
+	}
+}
+
+// SubnetToZonesMap parses the stored GKE SvcNEG object references inside status,
+// and resolves their subnetwork URL names mapping to the set of their GCE zone names.
+func (h *SvcNegStatusHandler) SubnetToZonesMap() (map[string]sets.Set[string], error) {
+	svcNegCR, err := h.getSvcNegFromStore()
+	if err != nil {
+		return nil, err
```
midavadim, line 85: should we add 

h.negMetrics.PublishNegControllerErrorCountMetrics(err, true) ?

midavadim: please reivew another err/mteric call after getSvcNegFromStore

Maybe it is better to move 
PublishNegControllerErrorCountMetrics
to getSvcNegFromStore func ?

pkg/neg/types/interfaces.go
```diff
@@ -94,3 +98,19 @@ type NetworkEndpointsCalculator interface {
...
+
+	// NEGsEnsured reports the status after NEGs are ensured.
+	NEGsEnsured(negs []*composite.NetworkEndpointGroup, errList []error) error
+
+	// SubnetToZonesMap returns the current subnet-to-zones mapping reported in status.
+	SubnetToZonesMap() (map[string]sets.Set[string], error)
+
+	// LastSyncTime returns the last time the NEG syncer synced associated NEGs.
+	LastSyncTime() (time.Time, error)
+}
```
midavadim, line 126: type NegStatusHandler interface {
	// SyncFinished reports the result of a sync operation.
	// It returns true if the syncer needs to re-initialize NEGs on next sync.
	**ReportSyncResult**(syncErr error) (**needInit** bool, error)

	// NEGsEnsured reports the status after NEGs are ensured.
	**ReportStatus**(negs []*composite.NetworkEndpointGroup, errList []error) error

	// SubnetToZonesMap returns the current subnet-to-zones mapping reported in status.
	SubnetToZonesMap() (map[string]sets.Set[string], error)

	// LastSyncTime returns the last time the NEG syncer synced associated NEGs.
	LastSyncTime() (time.Time, error)
}

midavadim: as discussed near whiteboard

pkg/neg/syncers/transaction.go
anyush: returned metric calls everywhere it was removed

pkg/neg/types/interfaces.go
anyush: Renamed

pkg/neg/syncers/negstatushandler/svcneg.go
anyush: I think it's better to do on-place, so simple getter won't have unexpected side effects - added metric calls everywhere these're needed

anyush: Not added there, as there are metric calls for SubnetToZonesMap's errors in the places this method is used (isZoneChange, isSubnetChange)

pkg/neg/syncers/transaction.go
```diff
@@ -516,7 +515,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	}
 
 	var errList []error
-	var negObjRefs []negv1beta1.NegObjectReference
+	var negObjRefs []*composite.NetworkEndpointGroup
```
midavadim: rename variable to negObjs

anyush: Done

midavadim:
LGTM

midavadim (approved):

pkg/neg/syncers/negstatushandler/svcneg.go
```diff
@@ -0,0 +1,349 @@
...
+		if _, ok := subnetZones[subnetName]; !ok {
+			subnetZones[subnetName] = sets.New[string]()
+		}
+		subnetZones[subnetName].Insert(id.Key.Zone)
+	}
+	return subnetZones, nil
+}
+
+// ReportStatus reports the updated list of successfully configured GCE NEGs inside SvcNEG status.
+func (h *SvcNegStatusHandler) ReportStatus(negs []*composite.NetworkEndpointGroup, errList []error) error {
```
mmamczur, line 115: so this patches status in the SvcNEG CR? I was thinking that with this name it would return some report

PR #3128, opened 2026-06-03
feat: keep draining nodes in L4 ETP Local NEGs while pods remain

08volt:
## Overview
Adds the `--enable-l4-neg-local-include-drain-nodes` (`EnableL4NEGLocalIncludeDrainNodes`) CLI flag. 

When enabled, nodes carrying the GKE drain label (`operation.gke.io/type=drain`) are kept in Layer 4 (L4) External Traffic Policy (`ETP=Local`) Network Endpoint Groups (NEGs) as long as they still host at least one backing pod. Without this flag, the drain label causes a node to be evicted from the NEG immediately, which drops in-flight connections before the pods have finished draining. 

> ⚠️ **Note:** This flag has no effect on `ETP=Cluster` NEGs or Layer 7 (L7) NEGs.

---

### 1. New Filter
* **`CandidateAndDrainingNodesFilter`** is added to `ZoneGetter`. 
* It functions as a strict superset of `CandidateAndUnreadyNodesFilter`, explicitly including both **unready nodes** and **drain-labeled nodes**. 
* All other exclusions (such as `ToBeDeletedTaint`, the `exclude-balancer` label, and subnet membership) continue to apply unconditionally.

### 2. Flag Threading
* The `includeDrainNodesL4Local` configuration flows from the CLI flag through the following lifecycle: 
    $$\text{Controller} \longrightarrow \text{syncerManager} \longrightarrow \text{NegSyncerKey} \longrightarrow \text{LocalL4EndpointsCalculator}$$
* Placing the value on `NegSyncerKey` preserves syncer identity, map keying, and filter derivation consistency without requiring direct flags access inside syncers.

### 3. LocalL4EndpointsCalculator
* The constructor has been replaced by a `LocalL4EndpointsCalculatorParams` struct to accommodate the new field without growing an already long positional argument list.
* The per-node filter check inside `CalculateEndpoints` evaluates the flag to dynamically select between:
    * `CandidateAndUnreadyNodesFilter` (flag off)
    * `CandidateAndDrainingNodesFilter` (flag on)

### 4. Node Update Handler
* The controller's node `UpdateFunc` now checks three distinct filters (one per calculator mode plus `CandidateNodesFilter`).
* This ensures that a node transitioning from drain-labeled to ready (or vice versa) reliably triggers a re-sync regardless of which NEG types the cluster is using.

### 5. NodeFilterForEndpointCalculatorMode
* Accepts a second `includeDrainNodesL4Local` boolean parameter.
* When `mode == L4LocalMode && includeDrainNodesL4Local` is true, it returns `CandidateAndDrainingNodesFilter`.
* All other mode/flag combinations return the same filters as before.

### 6. GetEndpointsCalculator
* Reads `syncerKey.IncludeDrainNodesL4Local` directly instead of accepting a separate boolean parameter, maintaining the flag value in one canonical location per syncer.

### 7. Zone Listing
* `ensureNetworkEndpointGroups`, `isZoneChange`, `syncNegStatusAnnotation`, and `retrieveExistingZoneNetworkEndpointMap` all derive their zone filter through a shared `candidateNodeFilter()` helper (or the same `NodeFilterForEndpointCalculatorMode` call).
* This ensures drain-zone NEGs are created, health-checked, and cleaned up consistently.

### 8. NegSyncerKey.String()
* The `-includeDrainNodesL4Local=true` suffix is appended only when the flag is true, ensuring the log format of all existing syncers remains unchanged.

---

## Testing Matrix

The following test suites have been introduced or updated to validate the feature:

| Test Case / Suite | Coverage & Validation Details |
| :--- | :--- |
| **`TestListZones`** | Adds `CandidateAndDrainingNodesFilter` case; verifies `zone4` (drain nodes) is included. |
| **`TestIsNodeSelectedByFilter`**<br>`TestLegacyIsNodeSelectedByFilter` | Adds an `expectAcceptByDraining` validation column and tests both a `ready+drain` and an `unready+drain` node case for each filter. |
| **`TestLocalGetEndpointSet`** | Adds four new test cases covering flag-on behavior explicitly with unready, draining, and `exclude-balancer` nodes. |
| **`TestSyncL4NEGs`** | Adds two end-to-end transaction syncer test cases covering flag-off behavior (drain node excluded) and flag-on behavior (drain node included). |
| **`TestNodeInformerFilterWithIncludeDrainNodesL4Local`** | Verifies that the controller properly enqueues a node when its drain label is removed while the flag is enabled. |
| **`TestRetrieveExistingZoneNetworkEndpointMapWithDrainNodes`** | Verifies `NotFound` suppression semantics for `zone4` both with and without the flag enabled. |
| **`TestGetSyncerKeyIncludeDrainNodesL4Local`** | Verifies that `getSyncerKey` correctly embeds the configuration flag into the `NegSyncerKey` struct. |
| **`TestNodePredicateForEndpointCalculatorMode`** | Updated to support the new flag parameter and ensure correct evaluation of the new filter variant. |

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -153,6 +153,7 @@ var F = struct {
 	EnableIPv6NodeNEGEndpoints                  bool
 	EnableL4NEGDetachCancel                     bool
 	EnablePSCReconcileConnections               bool
+	EnableL4NEGLocalIncludeDrainNodes           bool
 	// EnableL4DenyFirewallExplicitlySet will be set to true if the argument was explicitly set by the user.
 	EnableL4DenyFirewallExplicitlySet bool
 	// ===============================
@@ -371,6 +372,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableIPv6NodeNEGEndpoints, "enable-ipv6-node-neg-endpoints", false, "Enable populating IPv6 addresses for Node IPs in GCE_VM_IP NEGs.")
 	flag.BoolVar(&F.EnableL4NEGDetachCancel, "enable-l4-neg-detach-cancel", false, "Enable cancelling NEG detach when endpoints need to be re-attached.")
 	flag.BoolVar(&F.EnablePSCReconcileConnections, "enable-psc-reconcile-connections", false, "Enable support for PSC ServiceAttachment reconcile connections field")
+	flag.BoolVar(&F.EnableL4NEGLocalIncludeDrainNodes, "enable-l4-neg-local-include-drain-nodes", false, "For L4 LB NEGs with externalTrafficPolicy=Local, keep nodes carrying the GKE drain label in the NEG as long as they still host a backing pod. Unready-node behavior is unchanged.")
 }
 
 func Validate() {
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -116,11 +116,54 @@ type Controller struct {
 	// enableNEGsForIngress indicates whether the NEG controller will create NEGs for Ingress services
 	enableNEGsForIngress bool
 
+	// includeDrainNodesL4Local indicates whether to include draining nodes for NEGs with L4Local mode
+	includeDrainNodesL4Local bool
+
+	// nodeMembershipFilters is the deduplicated set of zone filters that decide
+	// whether a node belongs in a NEG. The node update handler treats a node
+	// change as relevant when the node's inclusion under any of these filters
+	// flips. Precomputed at controller initialization so the handler has a single
+	// source of truth and does not rebuild the set on every event.
+	nodeMembershipFilters []zonegetter.Filter
+
 	stopCh <-chan struct{}
 	logger klog.Logger
 
 	// negMetrics is used to collect metrics for NEG
 	negMetrics *metrics.NegMetrics
+}
+
+// buildNodeMembershipFilters returns the deduplicated set of node filters that
+// the NEG syncers use to decide whether a node belongs in a NEG. A node update
+// is only worth a resync if it changes the node's inclusion under one of these.
+// Dedup matters because several modes resolve to the same filter (L4 Cluster
+// and VM_IP_PORT both use CandidateNodesFilter; with includeDrainNodesL4Local
+// the L4 Local filter may coincide too), so we never evaluate a filter twice.
+func buildNodeMembershipFilters(includeDrainNodesL4Local bool) []zonegetter.Filter {
+	set := map[zonegetter.Filter]struct{}{
+		negtypes.NodeFilterForEndpointCalculatorMode(negtypes.L4LocalMode, includeDrainNodesL4Local): {},
+		negtypes.NodeFilterForEndpointCalculatorMode(negtypes.L4ClusterMode, false):                  {},
+		zonegetter.CandidateNodesFilter: {},
+	}
+	filters := make([]zonegetter.Filter, 0, len(set))
+	for f := range set {
+		filters = append(filters, f)
+	}
+	return filters
+}
+
+// nodeUpdateRequiresResync reports whether the change from oldNode to
+// currentNode could alter NEG membership and therefore needs a resync. The
+// change is relevant when the node's inclusion under any membership filter
+// flips.
+func (c *Controller) nodeUpdateRequiresResync(oldNode, currentNode *apiv1.Node) bool {
+	for _, filter := range c.nodeMembershipFilters {
+		if c.zoneGetter.IsNodeSelectedByFilter(oldNode, filter, c.logger) !=
+			c.zoneGetter.IsNodeSelectedByFilter(currentNode, filter, c.logger) {
+			return true
+		}
+	}
+	return false
 }
 
 // NewController returns a network endpoint group controller.
@@ -157,6 +200,7 @@ func NewController(
 	runL4ForNetLB bool,
 	readOnlyMode bool,
 	enableNEGsForIngress bool,
+	includeDrainNodesL4Local bool,
 	stopCh <-chan struct{},
 	logger klog.Logger,
 	negMetrics *metrics.NegMetrics,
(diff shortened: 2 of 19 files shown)


pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -58,13 +59,14 @@ func NewLocalL4EndpointsCalculator(nodeLister listers.NodeLister, zoneGetter *zo
...
-		subsetSizeLimit: subsetSize,
-		svcId:           svcId,
-		logger:          logger.WithName("LocalL4EndpointsCalculator"),
-		networkInfo:     networkInfo,
-		negMetrics:      negMetrics,
+		nodeLister:            nodeLister,
+		zoneGetter:            zoneGetter,
+		subsetSizeLimit:       subsetSize,
+		svcId:                 svcId,
+		excludeOnlyCandidates: flags.F.NegNodeFilteringExcludeOnlyCandidates,
```
TortillaZHawaii: can we make it so that we don't hide this global variable in the constructor? this would make the testing easier

pkg/flags/flags.go
```diff
@@ -371,6 +372,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.EnableIPv6NodeNEGEndpoints, "enable-ipv6-node-neg-endpoints", false, "Enable populating IPv6 addresses for Node IPs in GCE_VM_IP NEGs.")
 	flag.BoolVar(&F.EnableL4NEGDetachCancel, "enable-l4-neg-detach-cancel", false, "Enable cancelling NEG detach when endpoints need to be re-attached.")
 	flag.BoolVar(&F.EnablePSCReconcileConnections, "enable-psc-reconcile-connections", false, "Enable support for PSC ServiceAttachment reconcile connections field")
+	flag.BoolVar(&F.NegNodeFilteringExcludeOnlyCandidates, "neg-node-filtering-exclude-only-candidates", false, "Exclude nodes that are not candidates for NEG endpoints and include upgrading nodes.")
```
TortillaZHawaii: I wonder if this should be a boolean or an enum with default being the old value.

This flag should also state that this is behavior for etp local

pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -104,7 +106,13 @@ func (l *LocalL4EndpointsCalculator) CalculateEndpoints(eds []types.EndpointsDat
 				l.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
 				continue
 			}
-			if ok := l.zoneGetter.IsNodeSelectedByFilter(node, zonegetter.CandidateAndUnreadyNodesFilter, l.logger); !ok {
+
+			filter := zonegetter.CandidateAndUnreadyNodesFilter
+			if l.excludeOnlyCandidates {
+				filter = zonegetter.CandidateNodesFilter
```
TortillaZHawaii: let's maybe make it an enum

TortillaZHawaii (changes requested):

pkg/neg/syncers/endpoints_calculator.go
08volt: right

pkg/neg/syncers/endpoints_calculator.go
08volt: boolean flag is probably safer

pkg/neg/controller.go
```diff
@@ -237,6 +248,18 @@ func NewController(
 	}
 	enableMultiSubnetClusterPhase1 := flags.F.EnableMultiSubnetClusterPhase1
 
+	// Precompute the deduplicated filter set for the node update handler so the
```
mmamczur: why do we need this change?

08volt: this is just a simple optimization to create a slice of filter to loop through it on every sync ( line 370 ) instead of:
- having all filters hardcoded there like it is now ( +1 added with this PR )
- allocating a new map during every sync to loop through it

pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -168,19 +184,30 @@ type ClusterL4EndpointsCalculator struct {
 	negMetrics *metrics.NegMetrics
 }
 
-func NewClusterL4EndpointsCalculator(nodeLister listers.NodeLister, zoneGetter *zonegetter.ZoneGetter, svcId string, logger klog.Logger, networkInfo *network.NetworkInfo, l4LBtype types.L4LBType, negMetrics *metrics.NegMetrics) *ClusterL4EndpointsCalculator {
+// ClusterL4EndpointsCalculatorParams contains the parameters for the ClusterL4EndpointsCalculator.
```
mmamczur, line 186: I'm not sure if it's a good idea to do this params refactor together together with the logic change. I'd at least split to separate commits

pkg/neg/syncers/endpoints_calculator.go
```diff
@@ -168,19 +184,30 @@ type ClusterL4EndpointsCalculator struct {
...
+type ClusterL4EndpointsCalculatorParams struct {
+	NodeLister  listers.NodeLister
+	ZoneGetter  *zonegetter.ZoneGetter
+	SvcId       string
+	Logger      klog.Logger
```
mmamczur: I'd keep the logger as a separate argument anyway

pkg/neg/controller.go
mmamczur: the naming of this is very confusing. I guess the goal is to decide if a node update should be processed or not
it would be best to restrucute it so that it's clear what it's doing

nodeUdateFilters doesn't explain it to me

08volt: Refactored with 2 functions:
- buildNodeMembershipFilters: aggregate required filters
- nodeUpdateRequiresResync: decide whether a resync is needed

pkg/neg/syncers/endpoints_calculator.go
08volt: Split 2 commits, first one to add the params and second with logic change

pkg/neg/syncers/endpoints_calculator.go
08volt: done

PR #3131, opened 2026-06-09
Implement NEGBinding-based NEG namer

anyush:
Implements NEG namer which will return NEG name for subnet based solely on the Spec of the NetworkEndpointGroupBinding CR ([CRD](https://github.com/kubernetes/ingress-gce/pull/3103)).

`NonDefaultSubnetNEGNamer`'s `NonDefaultSubnetNEG` was updated to be able to return error. In practice it will allow to return not found error if trying to receive NEG name from NEGBinding-based namer for subnet which is not specified in the Spec. Also becomes more consistent with `NonDefaultSubnetCustomNEG` receiver.

### DIFF
--- a/pkg/utils/namer/negbinding_namer.go
+++ b/pkg/utils/namer/negbinding_namer.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namer
+
+import (
+	"errors"
+	"fmt"
+
+	"k8s.io/client-go/tools/cache"
+	negbindingv1beta1 "k8s.io/ingress-gce/pkg/apis/negbinding/v1beta1"
+)
+
+var (
+	ErrNegBindingNotFound       = errors.New("negbinding is not in store")
+	ErrNegNameNotFound          = errors.New("NEG name for subnet not found in NegBinding Spec")
+	ErrNBNamerNotImplemented    = errors.New("not implemented for NegBindingNamer")
+	ErrNBNamerInvalidBackendRef = errors.New("NEGBinding's backendRef doesn't match")
+)
+
+// NegBindingNamer resolves NEG names based on NetworkEndpointGroupBinding CR.
+// It implements NonDefaultSubnetNEGNamer.
+type NegBindingNamer struct {
+	namespace        string
+	name             string
+	negBindingLister cache.Indexer
+}
+
+// NewNegBindingNamer constructs a new NegBindingNamer
+func NewNegBindingNamer(namespace, name string, negBindingLister cache.Indexer) *NegBindingNamer {
+	return &NegBindingNamer{
+		namespace:        namespace,
+		name:             name,
+		negBindingLister: negBindingLister,
+	}
+}
+
+func (n *NegBindingNamer) getBinding() (*negbindingv1beta1.NetworkEndpointGroupBinding, error) {
+	key := fmt.Sprintf("%s/%s", n.namespace, n.name)
+	obj, exists, err := n.negBindingLister.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("error getting negbinding from cache: %w", err)
+	}
+	if !exists {
+		return nil, fmt.Errorf("%w: key %s", ErrNegBindingNotFound, key)
+	}
+
+	binding, ok := obj.(*negbindingv1beta1.NetworkEndpointGroupBinding)
+	if !ok {
+		return nil, fmt.Errorf("unexpected object type %T in negbinding cache for key %s", obj, key)
+	}
+	return binding, nil
+}
+
+// negNameForSubnet returns the NEG name defined in the NegBinding CR Spec for the given subnet.
+func (n *NegBindingNamer) negNameForSubnet(subnet, namespace, svcName string, port int32) (string, error) {
+	binding, err := n.getBinding()
+	if err != nil {
+		return "", err
+	}
+
+	if binding.Namespace != namespace ||
+		binding.Spec.BackendRef.Name != svcName ||
+		binding.Spec.BackendRef.Port != port {
+		gotRef := fmt.Sprintf("%s/%s:%v", namespace, svcName, port)
+		expectedRef := fmt.Sprintf("%s/%s:%v", binding.Namespace, binding.Spec.BackendRef.Name, port)
+		return "", fmt.Errorf("%w: backendRef %s, negBinding.Spec.BackendRef %s", ErrNBNamerInvalidBackendRef, gotRef, expectedRef)
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/utils/namer/negbinding_namer.go
```diff
@@ -0,0 +1,80 @@
...
+		if ref.Subnet == subnet {
+			return ref.Name, nil
+		}
+	}
+	key := fmt.Sprintf("%s/%s", n.namespace, n.name)
+	return "", fmt.Errorf("NEG name for subnet %s not found in NegBinding %s Spec", subnet, key)
+}
+
+// NonDefaultSubnetNEG returns the NEG name for the given subnet based on NegBinding's Spec.
+func (n *NegBindingNamer) NonDefaultSubnetNEG(namespace, name, subnetName string, port int32) (string, error) {
```
08volt: Parameters are not being used here. Prefer one of:
1. Drop namespace/name from the struct and use the parameters in getBinding / negNameForSubnet, mirroring Namer and L4Namer; or
2. Keep stored fields but return an error in NonDefaultSubnetNEG when namespace != n.namespace || name != n.name.

pkg/utils/namer/negbinding_namer.go
```diff
@@ -0,0 +1,80 @@
...
+func (n *NegBindingNamer) getBinding() (*negbindingv1beta1.NetworkEndpointGroupBinding, error) {
+	key := fmt.Sprintf("%s/%s", n.namespace, n.name)
+	obj, exists, err := n.negBindingLister.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("error getting negbinding from cache: %w", err)
+	}
+	if !exists {
+		return nil, fmt.Errorf("negbinding %s is not in store", key)
+	}
+	return obj.(*negbindingv1beta1.NetworkEndpointGroupBinding), nil
```
08volt: Unchecked type assertion will panic on cache.DeletedFinalStateUnknown (delivered during cache eviction races) or any wrong-type entry. 
```
binding, ok := obj.(*negbindingv1beta1.NetworkEndpointGroupBinding)
if !ok {
 return nil, fmt.Errorf("unexpected object type %T in negbinding cache for key %s", obj, key)
 }
return binding, nil
```

pkg/utils/namer/negbinding_namer.go
```diff
@@ -0,0 +1,80 @@
...
+		name:             name,
+		negBindingLister: negBindingLister,
+	}
+}
+
+func (n *NegBindingNamer) getBinding() (*negbindingv1beta1.NetworkEndpointGroupBinding, error) {
+	key := fmt.Sprintf("%s/%s", n.namespace, n.name)
+	obj, exists, err := n.negBindingLister.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("error getting negbinding from cache: %w", err)
```
08volt, line 55: Consider exported errors
For example:

```
var (
    ErrBindingNotFound      = errors.New("negbinding not in store")
    ErrSubnetNotInBinding   = errors.New("subnet not present in negbinding spec")
    ErrCustomNEGUnsupported = errors.New("custom NEG names not supported for NegBindingNamer")
)
```

pkg/utils/namer/negbinding_namer.go
```diff
@@ -0,0 +1,80 @@
...
+	key := fmt.Sprintf("%s/%s", n.namespace, n.name)
+	return "", fmt.Errorf("NEG name for subnet %s not found in NegBinding %s Spec", subnet, key)
+}
+
+// NonDefaultSubnetNEG returns the NEG name for the given subnet based on NegBinding's Spec.
+func (n *NegBindingNamer) NonDefaultSubnetNEG(namespace, name, subnetName string, port int32) (string, error) {
+	return n.negNameForSubnet(subnetName)
+}
+
+// NonDefaultSubnetCustomNEG implements NonDefaultSubnetNEGNamer.NonDefaultSubnetNEGNamer.
```
08volt: // NonDefaultSubnetCustomNEG implements NonDefaultSubnetNEGNamer.NonDefaultSubnetNEGNamer. → the second NonDefaultSubnetNEGNamer should be "NonDefaultSubnetCustomNEG."

pkg/utils/namer/negbinding_namer_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	negbindingv1beta1 "k8s.io/ingress-gce/pkg/apis/negbinding/v1beta1"
+)
+
+func TestNegBindingNamer(t *testing.T) {
```
08volt, line 29: Test gaps:

1. **Multi-Entry Binding Evaluation**
   * **Scenario:** A multi-entry binding where the matching subnet is located at the second (or later) entry.
   * **Reasoning:** The current test suite only validates instances where `len == 1`. As a result, the for-loop iteration logic past index `0` remains completely uncovered.

2. **Type Safety Guard Validation**
   * **Scenario:** Injecting an object of the wrong type into the indexer.
   * **Reasoning:** This is required to validate the robustness of the `comma-ok` type assertion guard suggested in `negbinding_namer.go:51`.

3. **Sentinel Error Assertions**
   * **Scenario:** Refactor existing error assertions.
   * **Reasoning:** If the proposed sentinel errors are adopted, switch existing `err == nil` or `err != nil` assertions to precise `errors.Is(err, ErrSubnetNotInBinding)` and `errors.Is(err, ErrBindingNotFound)` checks. This ensures that any regressions in specific error types are automatically caught by the test suite.

pkg/utils/namer/negbinding_namer.go
anyush: name from struct is NegBinding's name, and argument is service's name. Updated name of the argument to distinguish them + added validation of the backendRef based on the arguments

pkg/utils/namer/negbinding_namer.go
anyush: Done

pkg/utils/namer/negbinding_namer_test.go
anyush: Done

pkg/utils/namer/negbinding_namer.go
anyush: Done

pkg/utils/namer/negbinding_namer.go
anyush: Done

pkg/utils/namer/negbinding_namer.go
```diff
@@ -0,0 +1,100 @@
...
+
+// NonDefaultSubnetNEG returns the NEG name for the given subnet based on NegBinding's Spec.
+func (n *NegBindingNamer) NonDefaultSubnetNEG(namespace, svcName, subnetName string, port int32) (string, error) {
+	return n.negNameForSubnet(subnetName, namespace, svcName, port)
+}
+
+// NonDefaultSubnetCustomNEG implements NonDefaultSubnetNEGNamer.NonDefaultSubnetCustomNEG.
+// Returns a not implemented error as custom names should not be set for NegBinding-based syncers.
+func (n *NegBindingNamer) NonDefaultSubnetCustomNEG(customNEGName, subnetName string) (string, error) {
+	return "", ErrNBNamerNotImplemented
```
midavadim, line 99: Lets implement it the same as for l4 

func (n *L4Namer) NonDefaultSubnetCustomNEG(customNEGName, subnetName string) (string, error) {
	return "", fmt.Errorf("Custom NEG is not allowed for L4")
}

midavadim (approved):

midavadim:
LGTM

PR #3136, opened 2026-06-10
Fix data race in instancegroups TestSync

08volt:
IGManagerFake.syncedNodes was appended from the controller's task queue goroutine while the test goroutine read it directly, causing 'go test -race ./pkg/instancegroups/' to fail. Guard the fake's state with a mutex and read it through a synchronized accessor.

### DIFF
--- a/pkg/instancegroups/controller_test.go
+++ b/pkg/instancegroups/controller_test.go
@@ -2,6 +2,7 @@ package instancegroups
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/instancegroups/controller_test.go
```diff
@@ -191,24 +192,41 @@ func verifyExpectedSyncerCount(t *testing.T, syncedNodes [][]string, expectedCou
 }
 
 type IGManagerFake struct {
+	// mu guards syncedNodes, which is written by the controller's queue
+	// goroutine and read by the test goroutine.
+	mu          sync.Mutex
```
kl52752, line 197: why not RWMutex?

PR #3137, opened 2026-06-15
Sync endpoints of the ensured NEGs before resync if error caused by fail while ensuring only part of the NEGs

anyush:
If during the ensuring of NetworkEndpointGroups one (or more) NEGs failed to be ensured first endpoints of the correctly ensured will be synced and only then resync will start. Before changes - such failure will result in immediate resync, which potentially can lead to endpoints never attached/detached in case one (or more) NEGs can't be ensured.

### DIFF
--- a/pkg/neg/syncers/transaction.go
+++ b/pkg/neg/syncers/transaction.go
@@ -38,6 +38,7 @@ import (
 
 	nodetopologyv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/nodetopology/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
@@ -288,12 +289,31 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	// Only matters for L4 Local mode.
 	needInitDrainStatus := s.needInit && s.enableL4NEGDetachCancel && s.endpointsCalculator.Mode() == negtypes.L4LocalMode
 
+	var ensureErr error
+	var ensuredSubnetZones map[string]sets.Set[string]
 	if s.needInit || topologyChange {
 		s.logger.Info("Need to ensure network endpoint groups", "needInit", s.needInit, "topologyChange", topologyChange)
-		if err := s.ensureNetworkEndpointGroups(); err != nil {
-			return fmt.Errorf("%w: %v", negtypes.ErrNegNotFound, err)
+
+		// Passing ensured NEGs forward from ensureNetworkEndpointGroups() if called during this sync as reading from statusHandler in the same sync might result in reading stale data.
+		ensuredSubnetZones, ensureErr = s.ensureNetworkEndpointGroups()
+		if ensureErr == nil {
+			s.needInit = false
+		} else {
+			// Resync will be triggered only after this iteration will complete syncing ensured NEGs.
+			ensureErr = fmt.Errorf("%w: %v", negtypes.ErrNegNotFound, ensureErr)
+			s.needInit = true
+
+			// if not a single NEG ensured - no need to continue resync
+			if len(ensuredSubnetZones) == 0 {
+				return ensureErr
+			}
 		}
-		s.needInit = false
+	} else {
+		var err error
+		ensuredSubnetZones, err = s.statusHandler.SubnetToZonesMap()
+		if err != nil {
+			return fmt.Errorf("failed to get subnet to zones map from status handler: %w", err)
+		}
 	}
 	s.logger.V(2).Info("Sync NEG", "negSyncerKey", s.NegSyncerKey.String(), "endpointsCalculatorMode", s.endpointsCalculator.Mode())
 
@@ -304,7 +324,7 @@ func (s *transactionSyncer) syncInternalImpl() error {
 		return err
 	}
 
-	currentMap, currentPodLabelMap, drainingEndpoints, err := retrieveExistingZoneNetworkEndpointMap(subnetToNegMapping, s.topologyProvider, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.endpointsCalculator.Mode(), s.enableDualStackNEG, s.networkInfo, s.logger, s.negMetrics, needInitDrainStatus, s.NegSyncerKey.IncludeDrainNodesL4Local)
+	currentMap, currentPodLabelMap, drainingEndpoints, err := retrieveExistingZoneNetworkEndpointMap(subnetToNegMapping, s.topologyProvider, ensuredSubnetZones, s.cloud, s.NegSyncerKey.GetAPIVersion(), s.enableDualStackNEG, s.networkInfo, s.logger, s.negMetrics, needInitDrainStatus)
 	if err != nil {
 		return fmt.Errorf("%w: %w", negtypes.ErrCurrentNegEPNotFound, err)
 	}
@@ -366,6 +386,10 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			s.logger.Info("Using normal mode endpoint calculation")
 		}
 	}
+
+	// Filter out locations without NEGs to prevent attaching endpoints in locations where is no NEG
+	targetMap = s.dropLocationsWithoutNEGs(targetMap, currentMap)
+
 	// When the flags are not enabled, error state should be reset when no
 	// error occurs in the sync.
 	// notInDegraded and onlyInDegraded are not populated when the flags are
@@ -419,13 +443,16 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	}
 
 	if len(addEndpoints) == 0 && len(removeEndpoints) == 0 {
-		s.logger.V(3).Info("No endpoint change. Skip syncing NEG. ", s.Namespace, s.Name)
-		return nil
+		s.logger.V(3).Info("No endpoint change. Skip syncing NEG.", s.Namespace, s.Name)
+		return ensureErr
 	}
+
 	s.logEndpoints(addEndpoints, "adding endpoint")
 	s.logEndpoints(removeEndpoints, "removing endpoint")
-
-	return s.syncNetworkEndpoints(addEndpoints, removeEndpoints, endpointPodLabelMap, migrationZone)
+	if syncErr := s.syncNetworkEndpoints(addEndpoints, removeEndpoints, endpointPodLabelMap, migrationZone); syncErr != nil {
+		return utilerrors.NewAggregate([]error{ensureErr, syncErr})
+	}
+	return ensureErr
 }
 
 // reAddDrainingEndpointsThatAreInTargetMap will make sure that endpoints that are draining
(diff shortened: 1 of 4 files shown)


pkg/neg/syncers/utils.go
```diff
@@ -698,37 +688,45 @@ func podBelongsToService(pod *apiv1.Pod, service *apiv1.Service) error {
...
-	// Include zones that have non-candidate nodes currently. It is possible that NEGs were created in those zones previously and the endpoints now became non-candidates.
-	// Endpoints in those NEGs now need to be removed. This mostly applies to VM_IP_NEGs where the endpoints are nodes.
-	zones, err := zoneGetter.ListZones(zonegetter.AllNodesFilter, logger)
+func retrieveExistingZoneNetworkEndpointMap(subnetToNegMapping map[string]string, topologyProvider negtypes.TopologyProvider, negStatusHandler negtypes.NegStatusHandler, cloud negtypes.NetworkEndpointGroupCloud, version meta.Version, mode negtypes.EndpointsCalculatorMode, enableDualStackNEG bool, logger klog.Logger, negMetrics *metrics.NegMetrics, retrieveDrainStatus bool) (map[negtypes.NEGLocation]negtypes.NetworkEndpointSet, labels.EndpointPodLabelMap, map[negtypes.NetworkEndpoint]string, error) {
+	zoneNetworkEndpointMap := map[negtypes.NEGLocation]negtypes.NetworkEndpointSet{}
+	endpointPodLabelMap := labels.EndpointPodLabelMap{}
+	drainingEndpoints := make(map[negtypes.NetworkEndpoint]string)
+
+	ensuredZonesPerSubnet, err := negStatusHandler.SubnetToZonesMap()
```
08volt: `SubnetToZonesMap()` calls `getSvcNegFromStore()`, which reads the informer cache (`svcNEGLister`), while `ReportStatus()` earlier in the same `syncInternalImpl` call wrote the status through the API server (`patchSvcNegStatus`). Since the cache updates asynchronously, this read can still observe the pre-ensure status within the same sync.

For the first sync of a NEG this looks self-correcting: the cached status shows 0 NEGs, so `ReportSyncStatus` returns `needInit = true` and the next cycle re-runs. My concern is the incremental case where a new zone added to an already-initialized NEG: the cache shows a non-empty status, `needInit` stays false, and the new zone is dropped by `dropLocationsWithoutNEGs` this cycle. Recovery then depends on the SvcNEG-status watch re-enqueuing the key (or the periodic resync), and `isTopologyChange()` reads the same stale cache.

`patchSvcNegStatus` already returns the updated object,feeding it back so the same sync reads its own write would make this deterministic.

pkg/neg/syncers/transaction.go
```diff
@@ -407,13 +422,16 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
+		s.logger.V(3).Info("No endpoint change. Skip syncing NEG.", s.Namespace, s.Name)
+		return ensureErr
 	}
+
 	s.logEndpoints(addEndpoints, "adding endpoint")
 	s.logEndpoints(removeEndpoints, "removing endpoint")
-
-	return s.syncNetworkEndpoints(addEndpoints, removeEndpoints, endpointPodLabelMap, migrationZone)
+	if syncErr := s.syncNetworkEndpoints(addEndpoints, removeEndpoints, endpointPodLabelMap, migrationZone); syncErr != nil {
+		return utilerrors.NewAggregate([]error{ensureErr, syncErr})
```
08volt, line 453: `utilerrors.NewAggregate` implements `Is` but not `As`/`Unwrap`, so `ClassifyError`'s `errors.As(err, &NegSyncError{})` can't see the `ErrNegNotFound` carried by `ensureErr` once it's wrapped here — classification falls back to `ReasonOtherError`.

if `syncNetworkEndpoints` ever returns a NegSyncError with `IsErrorState: true`, aggregating it here would silently suppress the error-state entry. Classifying before aggregating (or a wrapper that preserves `errors.As`) would be safer

pkg/neg/syncers/utils.go
```diff
@@ -698,37 +688,45 @@ func podBelongsToService(pod *apiv1.Pod, service *apiv1.Service) error {
...
 			networkEndpointsWithHealthStatus, err := cloud.ListNetworkEndpoints(negName, zone, retrieveDrainStatus, version, logger)
 			if err != nil {
 				// It is possible for a NEG to be missing in a zone without candidate nodes. Log and ignore this error.
 				// NEG not found in a candidate zone is an error.
-				if utils.IsNotFoundError(err) && !candidateZonesMap.Has(zone) {
-					logger.Info("Ignoring NotFound error for NEG", "negName", negName, "zone", zone, "subnet", subnet)
-					negMetrics.PublishNegControllerErrorCountMetrics(err, true)
-					continue
+				if utils.IsNotFoundError(err) {
+					candidateZonesMap := sets.New(candidateZonesPerSubnet[subnet]...)
```
08volt: `candidateZonesMap` is now rebuilt on each NotFound error inside the `for zone` loop, whereas before it was computed once per subnet. It's only on the error path so the cost is negligible, but hoisting it back above the zone loop would be cleaner and match the prior structure

pkg/neg/syncers/transaction.go
```diff
@@ -548,69 +560,84 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
+				}
 			}
 		}
 	}
 
 	if updateNEGStatus {
-		s.updateInitStatus(negObjRefs, errList)
+		if err := s.statusHandler.ReportStatus(negObjs, errList); err != nil {
+			s.logger.Error(err, "Failed to report ensured NEGs on Status reporter")
+			return EnsuredNEGsReportingFailedError
```
08volt: Returning `EnsuredNEGsReportingFailedError` here changes behavior: previously a `ReportStatus` failure was logged and swallowed, now it aborts the sync. Is this intentional?

pkg/neg/syncers/transaction.go
```diff
@@ -270,29 +273,37 @@ func (s *transactionSyncer) syncInternalImpl() error {
...
-			return fmt.Errorf("%w: %v", negtypes.ErrNegNotFound, err)
+	var ensureErr error
+	if s.needInit || topologyChange {
+		s.logger.Info("Need to ensure network endpoint groups", "needInit", s.needInit, "topologyChange", topologyChange)
+		ensureErr = s.ensureNetworkEndpointGroups()
+		if ensureErr == nil {
+			s.needInit = false
+		} else if errors.Is(ensureErr, EnsuredNEGsReportingFailedError) {
+			// EnsuredNEGsReportingFailedError treated as fatal error as further in code
+			// current state will be read from statusHandler, so it should be up to date.
```
08volt: The comment reads as a justification for proceeding ("...so it should be up to date"), but the reason for the early return is the opposite: because reporting failed, the status is *not* trustworthy, so we must abort. Suggest rewording, e.g.:

```
    // ReportStatus failed, so the status we'd read downstream is stale/unreliable.
    // Treat as fatal and abort the sync so it's retried.
```

pkg/neg/syncers/transaction.go
```diff
@@ -354,6 +365,10 @@ func (s *transactionSyncer) syncInternalImpl() error {
 			s.logger.Info("Using normal mode endpoint calculation")
 		}
 	}
+
+	// Filter out locations where is no NEG
```
08volt: "there is no NEG"

pkg/neg/syncers/transaction.go
```diff
@@ -878,73 +905,41 @@ func (s *transactionSyncer) commitPods(endpointMap map[negtypes.NEGLocation]negt
...
-			s.negMetrics.PublishNegControllerErrorCountMetrics(err, true)
-			continue
+// dropLocationsWithoutNEGs excludes locations from targetMap that do not exist in currentMap.
+func (s *transactionSyncer) dropLocationsWithoutNEGs(targetMap, currentMap map[negtypes.NEGLocation]negtypes.NetworkEndpointSet) map[negtypes.NEGLocation]negtypes.NetworkEndpointSet {
+	filteredMap := make(map[negtypes.NEGLocation]negtypes.NetworkEndpointSet)
+	for loc, endpoints := range targetMap {
+		if _, ok := currentMap[loc]; ok {
+			filteredMap[loc] = endpoints
+		} else {
+			s.logger.Info("Excluding target endpoints for location, where is no NEG", "location", loc)
```
08volt: "there is no NEG"

pkg/neg/syncers/transaction.go
```diff
@@ -54,12 +52,14 @@ import (
...
 	"k8s.io/ingress-gce/pkg/neg/syncers/labels"
 	negtypes "k8s.io/ingress-gce/pkg/neg/types"
-	svcnegclient "k8s.io/ingress-gce/pkg/svcneg/client/clientset/versioned"
-	"k8s.io/ingress-gce/pkg/utils/patch"
 	"k8s.io/ingress-gce/pkg/utils/zonegetter"
 	"k8s.io/klog/v2"
 )
 
+var (
+	EnsuredNEGsReportingFailedError = errors.New("failed to report ensured NEGs")
```
08volt: no need to be exported

pkg/neg/syncers/utils.go
anyush: As discussed passing ensured locations from ensureNetworkEndpointGroups everywhere in the current iteration and reading status for iterations without ensuring NEGs.

pkg/neg/syncers/transaction.go
anyush: As I see syncNetworkEndpoints never return NegSyncError, so I think it can be safely be returned as is

pkg/neg/syncers/utils.go
anyush: Updated

pkg/neg/syncers/transaction.go
anyush: Removed err caused by fail in status reporting - irrelevant

pkg/neg/syncers/transaction.go
anyush: Updated

pkg/neg/syncers/transaction.go
anyush: Updated

pkg/neg/syncers/transaction.go
anyush: No more need for this error - removed

pkg/neg/syncers/transaction.go
anyush: Reverted this behaviour

pkg/neg/syncers/transaction.go
```diff
@@ -638,6 +658,7 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
 	if updateNEGStatus {
```
midavadim, line 698: probably we should move thid change below s.syncMetricsCollector call

pkg/neg/syncers/transaction.go
```diff
@@ -56,6 +56,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var (
+	EnsuredNEGsReportingFailedError = errors.New("failed to report ensured NEGs")
```
midavadim: rename to something like

ErrNEGsReportStatus = errors.New("failed to report NEGs status")

Err... is more common in our codebase

pkg/neg/syncers/transaction.go
midavadim: still needs to be repfrased

pkg/neg/syncers/transaction.go
midavadim: still not updated

pkg/neg/syncers/transaction.go
```diff
@@ -294,20 +294,26 @@ func (s *transactionSyncer) syncInternalImpl() error {
 	needInitDrainStatus := s.needInit && s.enableL4NEGDetachCancel && s.endpointsCalculator.Mode() == negtypes.L4LocalMode
 
 	var ensureErr error
+	var ensuredSubnetZones map[string]sets.Set[string]
```
midavadim, line 293: probably we should consider to use custom type, since we use it quite often

type SubnetToZonesMap map[string]sets.Set[string]

pkg/neg/syncers/transaction.go
anyush: Seems like this is comment to older version, where if there is an error during reporting, than we returned before metrics collector even called. closing as irrelevant

pkg/neg/syncers/utils.go
```diff
@@ -687,39 +687,36 @@ func podBelongsToService(pod *apiv1.Pod, service *apiv1.Service) error {
...
-		zones := allZonesPerSubnet[subnet]
-		candidateZonesMap := candidateZonesPerSubnet[subnet]
+		ensuredZones, ok := ensuredZonesPerSubnet[subnet]
+		if !ok {
+			continue
+		}
+		candidateZones := candidateZonesPerSubnet[subnet]
 
-		for zone := range zones {
+		for zone := range ensuredZones {
```
08volt: This loop now iterates ensuredZones instead of the previous AllNodesFilter set. 

The removed block specifically pulled in zones that have non-candidate nodes but still hold a NEG, to detach endpoints that are no longer desired. 

After a zone drops out of the candidate set (e.g. its nodes get cordoned/drained but aren't deleted), it won't be in ensuredSubnetZones on the ensure path, and ReportStatus then rewrites status without it, so no later sync re-scans it either. 

Whole-NEG GC (processNEGDeletionCandidate) only fires when the entire SvcNEG is undesired, not when a single zone drains. What detaches those stale endpoints now?

anyush: Yeah, already discovered this issue myself - `ensuredZones` should replace `candidateZones` there, not `allZones`, as these are in fact filtered candidates (non-existent filtered out) - fixed

PR #3139, opened 2026-06-16
Add a new controller for L4 Standalone NEG

mmamczur:
This PR adds the L4 Standalone NEG controller - a controller that allows the user to request a GCE_VM_IP NEG and set up a kubernetes cluster with a custom created external passthrough load balancer.

### DIFF
--- a/cmd/glbc/main.go
+++ b/cmd/glbc/main.go
@@ -662,6 +662,12 @@ func runL4Controllers(ctx *ingctx.ControllerContext, systemHealth *systemhealth.
 		logger.V(0).Info("L4NetLB controller started")
 	}
 
+	if flags.F.RunL4StandaloneNEGLBController {
+		standaloneNEGLBController := controllers.NewStandaloneNEGLBController(ctx, option.stopCh, logger)
+		runWithWg(standaloneNEGLBController.Run, option.wg)
+		logger.V(0).Info("L4 Standalone NEG LB controller started")
+	}
+
 	ctx.Start(option.stopCh)
 }
 
--- a/pkg/l4/annotations/service.go
+++ b/pkg/l4/annotations/service.go
@@ -107,6 +107,9 @@ const (
 
 	// Service annotation key for specifying L4LBConfig
 	L4LBConfigKey = "networking.gke.io/l4lb-config"
+
+	// CustomForwardingRuleKey is the annotation key for custom forwarding rule name.
+	CustomForwardingRuleKey = "networking.gke.io/custom-forwarding-rule"
 )
 
 // Service represents Service annotations.
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -1,0 +1,259 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/ingress-gce/pkg/composite"
+	ccontext "k8s.io/ingress-gce/pkg/context"
+	"k8s.io/ingress-gce/pkg/l4/annotations"
+	"k8s.io/ingress-gce/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+type parsedForwardingRule struct {
+	key     *meta.Key
+	rawName string
+}
+
+// StandaloneNEGLBController manages services with CustomNegLoadBalancerClass.
+type StandaloneNEGLBController struct {
+	ctx       *ccontext.ControllerContext
+	svcQueue  utils.TaskQueue
+	logger    klog.Logger
+	stopCh    <-chan struct{}
+	hasSynced func() bool
+}
+
+// NewStandaloneNEGLBController creates a new instance of StandaloneNEGLBController.
+func NewStandaloneNEGLBController(ctx *ccontext.ControllerContext, stopCh <-chan struct{}, logger klog.Logger) *StandaloneNEGLBController {
+	logger = logger.WithName("StandaloneNEGLBController")
+	lc := &StandaloneNEGLBController{
+		ctx:       ctx,
+		stopCh:    stopCh,
+		hasSynced: ctx.HasSynced,
+		logger:    logger,
+	}
+	lc.svcQueue = utils.NewPeriodicTaskQueueWithMultipleWorkers("standalone-l4-neg-lb", "services", 1, lc.sync, logger)
+
+	ctx.ServiceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			svc := obj.(*v1.Service)
+			if lc.shouldProcess(svc) {
+				lc.enqueue(svc)
+			}
+		},
+		UpdateFunc: func(old, cur interface{}) {
+			curSvc := cur.(*v1.Service)
+			if lc.shouldProcess(curSvc) {
+				lc.enqueue(curSvc)
+			}
+		},
+	})
+
+	return lc
+}
+
+func (lc *StandaloneNEGLBController) shouldProcess(svc *v1.Service) bool {
+	if svc.Spec.Type != v1.ServiceTypeLoadBalancer {
+		return false
+	}
+	return svc.Spec.LoadBalancerClass != nil && *svc.Spec.LoadBalancerClass == annotations.StandalonePassthroughNegLoadBalancerClass
+}
+
+func (lc *StandaloneNEGLBController) enqueue(svc *v1.Service) {
+	lc.svcQueue.Enqueue(svc)
+}
+
+func (lc *StandaloneNEGLBController) Run() {
+	lc.logger.Info("Starting StandaloneNEGLBController")
+
+	err := wait.PollUntilContextCancel(wait.ContextForChannel(lc.stopCh), 5*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		lc.logger.V(2).Info("Waiting for initial cache sync before starting L4 Standalone NEG controller")
+		return lc.hasSynced(), nil
+	})
+	if err != nil {
+		// This will fail if the context is canceled which means the channel was closed. We should return in that case.
+		lc.logger.Error(err, "Failed to wait for initial cache sync")
+		return
+	}
+
+	defer lc.svcQueue.Shutdown()
+	lc.svcQueue.Run()
+	<-lc.stopCh
+}
+
+func (lc *StandaloneNEGLBController) sync(key string) error {
+	obj, exists, err := lc.ctx.ServiceInformer.GetIndexer().GetByKey(key)
+	if err != nil {
+		return fmt.Errorf("failed to lookup service for key %s: %w", key, err)
+	}
+	if !exists || obj == nil {
+		return nil
+	}
+	svc := obj.(*v1.Service)
+	if svc.DeletionTimestamp != nil {
+		return nil
+	}
+	if !lc.shouldProcess(svc) {
+		return nil
+	}
+	svcLogger := lc.logger.WithValues("service", klog.KObj(svc))
+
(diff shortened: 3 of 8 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return nil
+}
+
+func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLogger klog.Logger) error {
+	frNamesStr, ok := svc.Annotations[annotations.CustomForwardingRuleKey]
+	if !ok || frNamesStr == "" {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "NoForwardingRuleRef", "Service has no forwarding rule reference")
+		svcLogger.V(4).Info("Service has no forwarding rule reference, skipping")
+		return nil
```
08volt, line 179: Previously published LB Ingress IP is never cleared

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+	var errs []error
+	errs = append(errs, parseErrs...)
+
+	if len(parsedRules) == 0 {
+		if len(errs) > 0 {
+			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use any Forwarding Rule %s", errors.Join(errs...).Error())
+			return errors.Join(errs...)
+		}
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "NoForwardingRuleRef", "Service has no forwarding rule reference")
+		return nil
```
08volt, line 196: Same here, previously published LB Ingress IP is never cleared

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+		lbIngresses = append(lbIngresses, v1.LoadBalancerIngress{IP: fr.IPAddress, IPMode: &vipMode})
+
+	}
+
+	if len(errs) > 0 {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use all Forwarding Rules %s", errors.Join(errs...).Error())
+	}
+	// if at least one FR was ok then we use it
+	if len(lbIngresses) == 0 {
+		return errors.Join(errs...)
```
08volt, line 230: Same here, never clears the LB status

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+			frKey = resourceID.Key
+		} else {
+			frKey = meta.RegionalKey(frName, lc.ctx.Cloud.Region())
+		}
+		parsedRules = append(parsedRules, parsedForwardingRule{key: frKey, rawName: frName})
+	}
+	return parsedRules, errs
+}
+
+func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
```
08volt, line 160: It never verifies the FR's backend service targets this service's NEGs, any external FR's VIP can be published. Is it possible to validate the linkage?

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+	ctx       *ccontext.ControllerContext
+	svcQueue  utils.TaskQueue
+	logger    klog.Logger
+	stopCh    <-chan struct{}
+	hasSynced func() bool
+}
+
+// NewStandaloneNEGLBController creates a new instance of StandaloneNEGLBController.
+func NewStandaloneNEGLBController(ctx *ccontext.ControllerContext, stopCh <-chan struct{}, logger klog.Logger) *StandaloneNEGLBController {
+	logger = logger.WithName("CustomLBController")
```
08volt: CustomLBController naming is inconsistent with type and flag

cmd/glbc/main.go
```diff
@@ -662,6 +662,12 @@ func runL4Controllers(ctx *ingctx.ControllerContext, systemHealth *systemhealth.
 		logger.V(0).Info("L4NetLB controller started")
 	}
 
+	if flags.F.RunL4StandaloneNEGLBController {
+		standaloneNEGLBController := controllers.NewStandaloneNEGLBController(ctx, option.stopCh, logger)
+		runWithWg(standaloneNEGLBController.Run, option.wg)
+		logger.V(0).Info("Custom NEG LB controller started")
```
08volt: Log not consistent should be something like "L4 Standalone NEG LB controller"

pkg/neg/controller.go
```diff
@@ -714,11 +715,13 @@ func (c *Controller) mergeVmIpNEGsPortInfo(service *apiv1.Service, name types.Na
 		return nil
 	}
 
-	// Ignore services with LoadBalancerClass different than "networking.gke.io/l4-regional-external" or
-	// "networking.gke.io/l4-regional-internal" used for L4 controllers that use GCE_VM_IP NEGs.
+	// Ignore services with LoadBalancerClass different than "networking.gke.io/l4-regional-external",
+	// "networking.gke.io/l4-regional-internal" or "networking.gke.io/custom-neg-load-balancer"
```
08volt: class for standalone negs doesn't match the actual one

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "NoForwardingRuleRef", "Service has no forwarding rule reference")
+		return nil
+	}
+
+	var lbIngresses []v1.LoadBalancerIngress
+	vipMode := v1.LoadBalancerIPModeVIP
+
+	for _, parsed := range parsedRules {
+		fr, err := composite.GetForwardingRule(lc.ctx.Cloud, parsed.key, meta.VersionGA, lc.logger)
```
08volt: should we pass svcLogger ?

pkg/l4/annotations/service.go
```diff
@@ -107,6 +107,12 @@ const (
 
 	// Service annotation key for specifying L4LBConfig
 	L4LBConfigKey = "networking.gke.io/l4lb-config"
+
+	// CustomForwardingRuleKey is the annotation key for custom forwarding rule name.
+	CustomForwardingRuleKey = "networking.gke.io/custom-forwarding-rule"
+	// ExternalHealthCheckPortKey is the annotation key for the port of the external health check.
+	// Deprecated: No longer used by GKE Ingress standalone NEG controller.
+	ExternalHealthCheckPortKey = "networking.gke.io/external-hc-port"
```
08volt: Is it already deprecated?

pkg/l4/controllers/standalonenegcontroller_test.go
```diff
@@ -0,0 +1,529 @@
...
+	testCases := []struct {
+		desc               string
+		svc                *v1.Service
+		frs                map[string]*composite.ForwardingRule
+		expectIPs          []string
+		expectEventReasons []string
+		expectError        bool
+	}{
+		{
+			desc: "Valid HTTP health check",
```
08volt: Is there an Health Check?

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -0,0 +1,229 @@
...
+			if lc.shouldProcess(curSvc) {
+				lc.enqueue(curSvc)
+			}
+		},
+	})
+
+	return lc
+}
+
+func (lc *StandaloneNEGLBController) shouldProcess(svc *v1.Service) bool {
```
08volt, line 81: We could check also that Service type is LoadBalancer

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: we will not verify this initially. Later we will add some validation but it will NOT reject the setup, only emit an event.
Making just the setup is correct is the users responsibility in this case.

pkg/l4/annotations/service.go
mmamczur: wtf, this AI...

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: you can't have the class set on a non-lb but adding for completeness

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: btw this veirification will be added in a followup

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: thanks for catching this, fixed

pkg/neg/controller.go

pkg/l4/controllers/standalonenegcontroller.go

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: done

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: done

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: done

cmd/glbc/main.go
mmamczur: changed

pkg/l4/controllers/standalonenegcontroller_test.go
```diff
@@ -0,0 +1,688 @@
...
+			} else {
+				for i, ip := range tc.expectIPs {
+					if updatedSvc.Status.LoadBalancer.Ingress[i].IP != ip {
+						t.Errorf("Expected IP %s, got %v", ip, updatedSvc.Status.LoadBalancer.Ingress[i].IP)
+					}
+				}
+			}
+
+			if len(tc.expectEventReasons) > 0 {
+				time.Sleep(100 * time.Millisecond)
```
08volt: sleep could cause flakyness.
Suggestion:
```
for _, expectedReason := range tc.expectEventReasons {
	expectedReason := expectedReason
	err := wait.PollUntilContextTimeout(context.Background(), 10*time.Millisecond, 5*time.Second, true,
		func(ctx context.Context) (bool, error) {
			events, err := kubeClient.CoreV1().Events(tc.svc.Namespace).List(ctx, metav1.ListOptions{})
			if err != nil {
				return false, err
			}
			for _, e := range events.Items {
				if e.Reason == expectedReason {
					return true, nil
				}
			}
			return false, nil
		})
	if err != nil {
		t.Errorf("Expected %s event, but none found within timeout", expectedReason)
	}
}
```

pkg/neg/controller.go
```diff
@@ -771,7 +731,9 @@ func (c *Controller) mergeVmIpNEGsPortInfo(service *apiv1.Service, name types.Na
 	// Update usage metrics.
 	negUsage.VmIpNeg = metricscollector.NewVmIpNegType(onlyLocal)
 	var l4LBType negtypes.L4LBType
-	if needsNEGForILB {
+	if wantsStandaloneNEGLB {
+		l4LBType = negtypes.L4ExternalLB // for now treat this as external since it has higher limits for backend subset
```
FelipeYepez, line 735: How will we know in the future if they want ILB or NetLB for Standalone neg

pkg/neg/controller.go
```diff
@@ -757,11 +715,13 @@ func (c *Controller) mergeVmIpNEGsPortInfo(service *apiv1.Service, name types.Na
...
-	// Ignore services with LoadBalancerClass different than "networking.gke.io/l4-regional-external" or
-	// "networking.gke.io/l4-regional-internal" used for L4 controllers that use GCE_VM_IP NEGs.
+	// Ignore services with LoadBalancerClass different than "networking.gke.io/l4-regional-external",
+	// "networking.gke.io/l4-regional-internal" or "networking.gke.io/standalone-passthrough-lb-neg"
+	// used for L4 controllers that use GCE_VM_IP NEGs.
 	if service.Spec.LoadBalancerClass != nil &&
 		!l4annotations.HasLoadBalancerClass(service, l4annotations.RegionalExternalLoadBalancerClass) &&
-		!l4annotations.HasLoadBalancerClass(service, l4annotations.RegionalInternalLoadBalancerClass) {
+		!l4annotations.HasLoadBalancerClass(service, l4annotations.RegionalInternalLoadBalancerClass) &&
+		!wantsStandaloneNEGLB {
```
FelipeYepez: We could just do the check for the loadBalancerClass here since we already return above if the service doesn't want standalone neg. I understand the logic but just in case we abstract it to other things in the future above

pkg/l4/controllers/standalonenegcontroller_test.go
mmamczur: this is a leftover, shouldn;t be here, removing

pkg/l4/controllers/standalonenegcontroller_test.go
mmamczur: done

mmamczur: interesting that AI suggested this and I guess corrected itself in your review

pkg/neg/controller.go

pkg/neg/controller.go
mmamczur: I think we'll have to allow the user to set the subset size themselves and they will have to be aware what they are configuring.
I see not good way, I don't want the NEG controller to have to check the ForwardingRule if it's internal or not.

PR #3140, opened 2026-06-17
fix(check-gke-ingress): dedupe repeated service and BackendConfig checks

immanuwell:
this fixes duplicate results in `check-gke-ingress` when one Ingress points to the same Service more than once.

what changed
- dedupe referenced Service names before service checks run
- dedupe referenced BackendConfig names before backendconfig checks run

repro
1. create an Ingress that uses `svc-1` in `defaultBackend` and again in one or more `rules[].http.paths[]`
2. add `cloud.google.com/backend-config: {"default":"beconfig-1","ports":{"80":"beconfig-1"}}` to that Service
3. run `check-gke-ingress -n test`
4. before this patch it spits duplicate `ServiceExistenceCheck`, `BackendConfigAnnotationCheck`, `BackendConfigExistenceCheck`, and `HealthCheckTimeoutCheck`
5. after this patch each of those shows once per ingress

why this is real
GKE docs support one Ingress reusing the same Service across multiple host or path rules, so this isnt a lab-only edge case.

tests
`go test ./cmd/check-gke-ingress/...`

### DIFF
--- a/cmd/check-gke-ingress/app/ingress/ingress.go
+++ b/cmd/check-gke-ingress/app/ingress/ingress.go
@@ -107,15 +107,16 @@ func RunChecks(ingresses []networkingv1.Ingress, client kubernetes.Interface, be
 
 		// Get the names of the services referenced by the ingress.
 		svcNames := []string{}
+		seenServices := make(map[string]struct{})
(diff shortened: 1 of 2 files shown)


immanuwell:
@bowei @spencerhance please take a look

PR #3142, opened 2026-06-19
Enable NEGs creation by flag on the NEG controller's syncer

anyush:
Allow disablement of the lifecycle management by NEG controller's syncer by setting a flag.

Allows to manage endpoints of the NEGs, which lifecycle is not managed by NEG controller.


pkg/neg/syncers/utils.go
08volt, line 160: Shouldnt the manageLifecycle check be added here as well?

anyush: Should be - added

pkg/neg/syncers/utils.go
```diff
@@ -164,11 +163,23 @@ func ensureNetworkEndpointGroup(svcNamespace, svcName, negName, zone, negService
 			(!utils.EqualResourceIDs(neg.Network, networkInfo.NetworkURL) ||
 				!utils.EqualResourceIDs(neg.Subnetwork, networkInfo.SubnetworkURL)) {
 
+			if !manageLifecycle {
+				negLogger.Error(
+					nil,
+					"NEG does not match network and subnetwork of the cluster, but NEGs' lifecycle not managed",
```
midavadim: network and/or subnetwork

midavadim: Found unmanaged NEG does not match network and/or subnetwork of the cluster

midavadim: somthing like this

anyush: updated

PR #3143, opened 2026-06-19
Move L4-specific error types and helpers to pkg/l4/utils

zuzqqa:
Part two of the ongoing effort to split `pkg/utils`. This extracts all L4-specific error types, error checkers, and their corresponding tests out of `pkg/utils/utils.go` and into a dedicated `pkg/l4/utils/errors.go` file.

### DIFF
--- a/pkg/l4/address/manager.go
+++ b/pkg/l4/address/manager.go
@@ -23,9 +23,10 @@ import (
 
 	compute "google.golang.org/api/compute/v1"
 	"k8s.io/cloud-provider-gcp/providers/gce"
-	"k8s.io/ingress-gce/pkg/utils"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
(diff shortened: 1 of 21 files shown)


pkg/l4/address/manager.go
```diff
@@ -23,9 +23,10 @@ import (
 
 	compute "google.golang.org/api/compute/v1"
 	"k8s.io/cloud-provider-gcp/providers/gce"
-	"k8s.io/ingress-gce/pkg/utils"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	l4utils "k8s.io/ingress-gce/pkg/l4/utils"
+	utils "k8s.io/ingress-gce/pkg/utils"
```
TortillaZHawaii: Nit, do you need to use named import if the PKG name matches

zuzqqa: Good catch, I updated the import

PR #3144, opened 2026-06-22
Topology provider negbinding

anyush:
`TopologyProvider`'s (https://github.com/kubernetes/ingress-gce/pull/3121) implementation based on the NEGBinding CR (https://github.com/kubernetes/ingress-gce/pull/3103) instead of the SvcNEG.

As NEGBinding's subnets and zones should be managed by the feature's user, it's expected that is location (subnet + zone) is provided in the CR, then controller should expect and manage it there without checking if there are (or are no any) candidate nodes in this location, as adding location to CR is an explicit request to manage NEG there.

### DIFF
--- a/pkg/neg/syncers/negbinding_topology.go
+++ b/pkg/neg/syncers/negbinding_topology.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncers
+
+import (
+	"fmt"
+
+	nodetopologyv1 "github.com/GoogleCloudPlatform/gke-networking-api/apis/nodetopology/v1"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+	negbindingv1beta1 "k8s.io/ingress-gce/pkg/apis/negbinding/v1beta1"
+	"k8s.io/ingress-gce/pkg/neg/types/shared"
+	"k8s.io/ingress-gce/pkg/network"
+	"k8s.io/ingress-gce/pkg/utils/zonegetter"
+	"k8s.io/klog/v2"
+)
+
+// NEGBindingTopologyProvider provides subnets and zones where NEGs should be managed
+// based on NetworkEndpointGroupBinding CR
+type NEGBindingTopologyProvider struct {
+	negBindingName   string
+	namespace        string
+	negBindingLister cache.Indexer
+	defaultSubnetID  *cloud.ResourceID
+}
+
+// NewNEGBindingTopologyProvider constructs a new NEGBindingTopologyProvider
+func NewNEGBindingTopologyProvider(namespace, negBindingName string, negBindingLister cache.Indexer, defaultSubnetURL string) (*NEGBindingTopologyProvider, error) {
+	defaultSubnetID, err := cloud.ParseResourceURL(defaultSubnetURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse default subnetwork URL %q: %w", defaultSubnetURL, err)
+	}
+
+	return &NEGBindingTopologyProvider{
+		negBindingName:   negBindingName,
+		namespace:        namespace,
+		negBindingLister: negBindingLister,
+		defaultSubnetID:  defaultSubnetID,
+	}, nil
+}
+
+func (p *NEGBindingTopologyProvider) getBinding() (*negbindingv1beta1.NetworkEndpointGroupBinding, error) {
+	key := fmt.Sprintf("%s/%s", p.namespace, p.negBindingName)
+	obj, exists, err := p.negBindingLister.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("error getting negbinding from cache: %w", err)
+	}
+	if !exists {
+		return nil, fmt.Errorf("negbinding %s is not in store", key)
+	}
+	binding, ok := obj.(*negbindingv1beta1.NetworkEndpointGroupBinding)
+	if !ok {
+		return nil, fmt.Errorf("cached object %q is of type %T, expected *NetworkEndpointGroupBinding", key, obj)
+	}
+	return binding, nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/neg/syncers/negbinding_topology.go
```diff
@@ -0,0 +1,102 @@
...
+		configs = append(configs, nodetopologyv1.SubnetConfig{
+			Name:       ref.Subnet,
+			SubnetPath: subnetPath,
+		})
+	}
+	return configs
+}
+
+// ListZonesPerSubnet returns a map of subnet to zones defined inside the NegBinding CR Spec.
+func (p *NEGBindingTopologyProvider) ListZonesPerSubnet(filter zonegetter.Filter, logger klog.Logger) (map[string][]string, error) {
```
08volt: filter param is ignored, it would be better to add a comment about this

pkg/neg/syncers/negbinding_topology_test.go
08volt, line 1: Consider adding a case for 
- the "negbinding is not in store" error path
- Spec.NetworkEndpointGroups is nil or empty

pkg/neg/syncers/negbinding_topology.go
```diff
@@ -0,0 +1,102 @@
...
+func (p *NEGBindingTopologyProvider) getBinding() (*negbindingv1beta1.NetworkEndpointGroupBinding, error) {
+	key := fmt.Sprintf("%s/%s", p.namespace, p.negBindingName)
+	obj, exists, err := p.negBindingLister.GetByKey(key)
+	if err != nil {
+		return nil, fmt.Errorf("error getting negbinding from cache: %w", err)
+	}
+	if !exists {
+		return nil, fmt.Errorf("negbinding %s is not in store", key)
+	}
+	return obj.(*negbindingv1beta1.NetworkEndpointGroupBinding), nil
```
08volt: this can panic, better to use "binding, ok := " and check if it is ok

pkg/neg/syncers/negbinding_topology.go
anyush: Added

pkg/neg/syncers/negbinding_topology.go
anyush: Added

pkg/neg/syncers/negbinding_topology_test.go
anyush: Added test cases for not in the store and empty NEGs list. Spec.NetworkEndpointGroups is a required field and can't be set to nil

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -229,19 +229,60 @@ func (z *ZoneGetter) ListNodesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+	}
+	if zones, ok := subnetToZones[subnet]; ok {
+		return zones.UnsortedList(), nil
+	}
+	return []string{}, nil
+}
+
+// list zones returns zones where filtered nodes are found by subnet (map[subnetName]sets.Set[zoneName])
+func (z *ZoneGetter) listZones(filter Filter, defaultSubnetOnly bool, logger klog.Logger) (map[string]sets.Set[string], error) {
 	if z.mode == NonGCP {
```
midavadim, line 274: we should rename this func

pkg/utils/zonegetter/zone_getter.go
```diff
@@ -229,19 +229,74 @@ func (z *ZoneGetter) ListNodesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+	}
+	zonesPerSubnet := map[string][]string{}
+	for subnet, zones := range subnetToZones {
+		zonesPerSubnet[subnet] = zones.UnsortedList()
+	}
+	return zonesPerSubnet, nil
+}
+
+// list zones returns zones where filtered nodes are found by subnet (map[subnetName]sets.Set[zoneName])
+func (z *ZoneGetter) listZones(filter Filter, defaultSubnetOnly bool, logger klog.Logger) (map[string]sets.Set[string], error) {
```
midavadim: is it possible to use the  function name  everyhtiherer jere in binding topology provider 

ListZonesPerSubnet and return the same type ?

pkg/neg/syncers/negbinding_topology.go
```diff
@@ -0,0 +1,108 @@
...
+			SubnetPath: subnetPath,
+		}
+	}
+	return configs
+}
+
+// ListZonesPerSubnet returns a map of subnet to zones defined inside the NegBinding CR Spec.
+// NEGBinding contains explicit locations (subnet + zone pairs), where NEG controller is expected to
+// manage NEGs ignoring if any endpoints available there. Therefore ignoring filtering.
+func (p *NEGBindingTopologyProvider) ListZonesPerSubnet(_ zonegetter.Filter, logger klog.Logger) (map[string][]string, error) {
```
midavadim: it it possible to use the same type verytwherefpr subettoxone map ?

map[string][]string
map[string]set ...

pkg/neg/types/interfaces.go
```diff
@@ -94,3 +97,10 @@ type NetworkEndpointsCalculator interface {
 	// ValidateEndpoints validates the NEG endpoint information is correct
 	ValidateEndpoints(endpointData []EndpointsData, endpointPodMap EndpointPodMap, endpointsExcludedInCalculation int) error
 }
+
+// TopologyProvider is an interface for looking up zone and subnet information for NEG syncer.
+// It helps to determine GCE locations (subnets and zones) where NEGs should be created and/or synced.
+type TopologyProvider interface {
+	ListSubnets(logger klog.Logger) []nodetopologyv1.SubnetConfig
```
midavadim: actually "nodetopologyv1.SubnetConfig" is not needed . Evehere in the code just subnet name is used from this strcture

midavadim: Actually, it is used only once to get project info from subneturl ... which sound strange... I am not sure If I mssing something


/usr/local/google/home/ratniuk/src/k8s.io/ingress-gce/pkg/neg/syncers/transaction.go
				// Determine the networkInfo for the non-default subnet NEGs.
				resourceID, err := cloud.ParseResourceURL(subnetConfig.SubnetPath)
				if err != nil {
					s.logger.Error(err, "Failed to parse subnet path", "subnetPath", subnetConfig.SubnetPath)
					errList = append(errList, err)
					continue
				}
				// Add compute and version GA prefix.
				networkInfo.SubnetworkURL = cloud.SelfLink(meta.VersionGA, resourceID.ProjectID, resourceID.Resource, resourceID.Key)

pkg/neg/syncers/transaction.go
```diff
@@ -570,63 +564,75 @@ func (s *transactionSyncer) ensureNetworkEndpointGroups() error {
...
-			)
-			if err != nil {
-				errList = append(errList, err)
-				// Do not modify NEG Status if there is conflict within the same cluster
-				// and namespace because the CR is owned by a different syncer.
-				if errors.Is(err, utils.ErrNEGUsedByAnotherSyncer) {
-					updateNEGStatus = false
-					break
+				// Determine the networkInfo for the non-default subnet NEGs.
+				resourceID, err := cloud.ParseResourceURL(subnetConfig.SubnetPath)
```
midavadim: basically , it is not needed here since

resourceID.ProjectID and Region (which part of resourceID.Key) we can take directly from NetworkInfo

project and region must be the same on the cluster.

so only thing we need is 
subnetname

anyush: Not sure honestly, but as I see project can differ in shared VPC case

PR #3147, opened 2026-06-30
vuln fixes for 1.34

kl52752:


hack/tools/go.mod
```diff
@@ -1,200 +1,209 @@
 module k8s.io/ingress-gce/hack/tools
 
-go 1.18
+go 1.23.0
```
TortillaZHawaii: this is not supported anymore, right? https://go.dev/doc/devel/release#policy

TortillaZHawaii: and there were some security fixes in later 1.23 patches

kl52752: fixed

PR #3148, opened 2026-06-30
Check for address in forwarding rules before reserving it.

JumiDeluxe:
This change aims to improve static IP assignment by avoiding triggering quota `filtered_list_cost_overhead_per_region` (https://docs.cloud.google.com/compute/api-quota).

### DIFF
--- a/pkg/l4/address/manager.go
+++ b/pkg/l4/address/manager.go
@@ -20,11 +20,15 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strings"
 
 	compute "google.golang.org/api/compute/v1"
 	"k8s.io/cloud-provider-gcp/providers/gce"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/filter"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"k8s.io/ingress-gce/pkg/composite"
 	l4utils "k8s.io/ingress-gce/pkg/l4/utils"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
@@ -168,7 +172,6 @@ func (m *Manager) ReleaseAddress() error {
 }
 
 func (m *Manager) removeAddress(name, reason string) error {
-
 	m.frLogger.V(2).Info("Releasing existing address", "name", name, "reason", reason)
 	err := m.svc.DeleteRegionAddress(name, m.region)
 	if err != nil {
@@ -180,6 +183,26 @@ func (m *Manager) removeAddress(name, reason string) error {
 	}
 	m.frLogger.V(4).Info("Successfully released address", "addressName", name)
 	return nil
+}
+
+// DecompressIPv6 re-adds ommited hextets in IPv6 address.
+// For example, running the function with argument "2001:db8::ff00:42:8329"
+// returns "2001:db8:0:0:0:ff00:42:8329".
+func DecompressIPv6(addr string) string {
+	if !strings.Contains(addr, "::") {
+		// no compression
+		return addr
+	}
+	hextets := strings.Split(addr, ":")
+	toAdd := 9 - len(hextets)
+	expanded := strings.Repeat(":0", toAdd) + ":"
+	if strings.HasPrefix(addr, "::") {
+		expanded = "0" + expanded
+	}
+	if strings.HasSuffix(addr, "::") {
+		expanded += "0"
+	}
+	return strings.Replace(addr, "::", expanded, 1)
 }
 
 // ensureAddressReservation reserves ip address and returns address as a string,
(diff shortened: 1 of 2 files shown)


pkg/l4/address/manager.go
```diff
@@ -270,6 +293,34 @@ func (m *Manager) ensureAddressReservation() (string, IPAddressType, error) {
...
 
+	// Check if the address is in forwarding rules. If it is, we don't need to reserve it.
+	if gceCloud, ok := m.svc.(*gce.Cloud); ok {
+		// Create a filter for the IPAddress.
+		// ForwardingRules keep IP addresses in expanded form, mask may be included.
+		ipFilter := filter.Regexp("IPAddress", DecompressIPv6(m.targetIP)+"(/\\d+)?")
+		key := meta.RegionalKey("", m.region)
+
+		// List regional forwarding rules to determine if the target IP is already in use.
+		frList, err := composite.ListForwardingRules(gceCloud, key, meta.VersionGA, m.frLogger, ipFilter)
```
mmamczur: in this case we'd start listing Forwarding rules with a client side filter. It could lead us to similar problems like listing Addresses. I would be best to try to look at the FR currently used by the LB, and only that one. If the IPs don't match then we may have to execute the old logic.

JumiDeluxe: right. Replaced, now we only check the current one via `get`, and check for its IP address and service name, because I don't know if we can assume that it will always have current service name.

pkg/l4/address/manager.go
```diff
@@ -205,6 +205,39 @@ func DecompressIPv6(addr string) string {
 	return strings.Replace(addr, "::", expanded, 1)
 }
 
+func (m *Manager) isAddressInForwardingRules() error {
+	// Check if the address is in forwarding rules. If it is, we don't need to reserve it.
+	if gceCloud, ok := m.svc.(*gce.Cloud); ok {
+		key := meta.RegionalKey(m.name, m.region)
+		fr, err := composite.GetForwardingRule(gceCloud, key, meta.VersionGA, m.frLogger)
+		if err != nil {
+			return fmt.Errorf("failed to lookup forwarding rules, err: %v", err)
```
mmamczur: if you do a return like this you can drop the else clause below, always recommended in go

```
        fr, err := composite.GetForwardingRule(gceCloud, key, meta.VersionGA, m.frLogger)
        if err != nil {
			return fmt.Errorf("failed to lookup forwarding rules, err: %v", err)
		} 
		// We might need to match de-compressed address with mask.
		ok, err := regexp.MatchString(DecompressAddr(m.targetIP)+"(/\\d+)?", fr.IPAddress)
        ...
```

pkg/l4/address/manager.go
```diff
@@ -205,6 +205,39 @@ func DecompressIPv6(addr string) string {
...
+		} else {
+			// We might need to match de-compressed address with mask.
+			ok, err := regexp.MatchString(DecompressAddr(m.targetIP)+"(/\\d+)?", fr.IPAddress)
```
mmamczur: what is the trailing piece that we need to match with a regex?

JumiDeluxe: it is an optional group for mask, that is `/` and one or more digits

pkg/l4/address/manager.go
JumiDeluxe: dropped

pkg/l4/address/manager.go
```diff
@@ -270,6 +333,12 @@ func (m *Manager) ensureAddressReservation() (string, IPAddressType, error) {
 		return "", IPAddrUndefined, fmt.Errorf("failed to reserve address %q with no specific IP, err: %v", m.name, reserveErr)
 	}
 
+	if ok := m.IsAddressInForwardingRules(); ok {
+		return m.targetIP, IPAddrManaged, nil
```
mmamczur: most likely if we reach here the address is not managed

does this value (managed/non managed) matter at all later? If not maybe we don't have to worry about it too much.

pkg/l4/address/manager.go
```diff
@@ -182,6 +185,65 @@ func (m *Manager) removeAddress(name, reason string) error {
...
+
+		// If the forwarding rule service name matches.
+		var desc utils.L4LBResourceDescription
+		descErr := desc.Unmarshal(fr.Description)
+		if descErr != nil {
+			m.frLogger.V(4).Info(m.targetIP, fmt.Sprintf("failed to unmarshal forwarding rule, err: %v", err))
+			return false
+		}
+		if desc.ServiceName != m.serviceName {
+			m.frLogger.V(4).Info(m.targetIP, fmt.Sprintf("is already in use by forwarding rule %s of service %s", fr.Name, desc.ServiceName))
```
mmamczur, line 248: this would actually be a very weird state where we have a FR name conflict. I'm not sure if the code later even cares about the description, if it doesn't match probably it would be updated to match. but returning false here might make sense

pkg/l4/address/manager.go
```diff
@@ -182,6 +185,65 @@ func (m *Manager) removeAddress(name, reason string) error {
...
+// DecompressAddr re-adds ommited hextets in IPv6 address.
+// For example, running the function with argument "2001:db8::ff00:42:8329"
+// returns "2001:db8:0:0:0:ff00:42:8329".
+func DecompressAddr(addr string) string {
+	if !strings.Contains(addr, "::") {
+		// no compression
+		return addr
+	}
+	hextets := strings.Split(addr, ":")
+	toAdd := 9 - len(hextets)
```
mmamczur: could you add a comment about why we use `9` here? Also maybe guard against this becoming negative number. Should never happen but who knows, not sure what strings.Repeat does with negative values

pkg/l4/address/manager.go
JumiDeluxe: I see it only used as a result in https://github.com/kubernetes/ingress-gce/blob/071cb7e6e21f4e49f4cba67cdc1f8b2797b2a99e/pkg/l4/address/hold.go#L91 which is then used for metrics in https://github.com/kubernetes/ingress-gce/blob/071cb7e6e21f4e49f4cba67cdc1f8b2797b2a99e/pkg/l4/resources/l4netlb.go#L505 and https://github.com/kubernetes/ingress-gce/blob/071cb7e6e21f4e49f4cba67cdc1f8b2797b2a99e/pkg/l4/metrics/legacy_metrics.go#L173-L175

pkg/l4/address/manager.go
JumiDeluxe: we are returning `false` here already, unless you mean also another location?

pkg/l4/address/manager.go
JumiDeluxe: right, this may not be self explanatory at first sight, added comments and checks

pkg/l4/address/manager.go
JumiDeluxe: changed to `IPAddrUnmanaged`

PR #3150, opened 2026-07-03
Standalone L4 NEG controller metrics

mmamczur:
Added a metric `l4_standalone_neg_count` that tracks the number of standalone NEG services. The metric can be broken down by the type of FR attached and the status. The status could be success, user error (bad configuration) or system error (mainly problems reaching GCE API).

### DIFF
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -25,15 +25,44 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"golang.org/x/exp/slices"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/ingress-gce/pkg/composite"
 	ccontext "k8s.io/ingress-gce/pkg/context"
 	"k8s.io/ingress-gce/pkg/l4/annotations"
+	l4metrics "k8s.io/ingress-gce/pkg/l4/metrics"
+	"k8s.io/ingress-gce/pkg/l4/resources"
+	l4utils "k8s.io/ingress-gce/pkg/l4/utils"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
 )
+
+var (
+	// supportedLBSchemes is a list of supported LB schemes. Only Forwarding rules
+	// with these schemes can be referenced in a L4 standalone NEG service.
+	supportedLBSchemes = []string{
+		string(cloud.SchemeExternal),
+		"EXTERNAL_PASSTHROUGH",
+	}
+	// supportedFRProtocols is a list of supported protocols. Only Forwarding
+	// rules with these protocols can be referenced in a L4 standalone NEG
+	// service.
+	supportedFRProtocols = []string{
+		"TCP",
+		"UDP",
+		"L3_DEFAULT",
+	}
+)
+
+func isSchemeSupported(lbScheme string) bool {
+	return slices.Contains(supportedLBSchemes, lbScheme)
+}
+
+func isFRProtocolSupported(protocol string) bool {
+	return slices.Contains(supportedFRProtocols, protocol)
+}
 
 type parsedForwardingRule struct {
 	key     *meta.Key
@@ -62,15 +91,40 @@ func NewStandaloneNEGLBController(ctx *ccontext.ControllerContext, stopCh <-chan
 
 	ctx.ServiceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			svc := obj.(*v1.Service)
+			svc, ok := obj.(*v1.Service)
+			if !ok {
+				return
+			}
 			if lc.shouldProcess(svc) {
 				lc.enqueue(svc)
 			}
 		},
 		UpdateFunc: func(old, cur interface{}) {
-			curSvc := cur.(*v1.Service)
-			if lc.shouldProcess(curSvc) {
+			curSvc, ok := cur.(*v1.Service)
+			if !ok {
+				return
+			}
+			oldSvc, okOld := old.(*v1.Service)
+			if okOld && lc.shouldProcess(oldSvc) || lc.shouldProcess(curSvc) {
 				lc.enqueue(curSvc)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			svc, ok := obj.(*v1.Service)
+			if !ok {
+				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+				if !ok {
+					logger.Error(nil, "unexpected object type in DeleteFunc", "type", fmt.Sprintf("%T", obj))
+					return
+				}
+				svc, ok = tombstone.Obj.(*v1.Service)
+				if !ok {
+					logger.Error(nil, "unexpected object type in tombstone in DeleteFunc", "type", fmt.Sprintf("%T", tombstone.Obj))
+					return
+				}
+			}
+			if lc.shouldProcess(svc) {
+				lc.enqueue(svc)
 			}
 		},
 	})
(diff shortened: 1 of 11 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -68,11 +71,30 @@ func NewStandaloneNEGLBController(ctx *ccontext.ControllerContext, stopCh <-chan
 			}
 		},
 		UpdateFunc: func(old, cur interface{}) {
+			oldSvc := old.(*v1.Service)
```
TortillaZHawaii: Are there guarantees that there will always be `old` and `cur` present (not nil)? If they could be nil does lc.shouldProcess handle that case?

Could they be always asserted to `*v1.Service`?

I see below that it could potentially be something else for Delete. It will probably be safer to add an okay here as well to prevent from panic.

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -158,7 +189,9 @@ func (lc *StandaloneNEGLBController) parseForwardingRuleKeys(frNamesStr string,
 }
 
 func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
-	if fr.LoadBalancingScheme != string(cloud.SchemeExternal) {
+	if fr.LoadBalancingScheme != string(cloud.SchemeExternal) &&
+		fr.LoadBalancingScheme != "EXTERNAL_PASSTHROUGH" &&
```
TortillaZHawaii: Is `External_Passthrough` new?

Also if this is visible to the user:
* We should state which are supported. Probably useful to have supported values defined once, and use `!slices.Contains(XSupported, fr.X)` and include `XSupported` in message.
*  We should return all the errors at once (`errors.Join()`), so that users don't have to recreate the forwarding rule potentially two times if both `protocol` and `scheme` is wrong

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -167,21 +200,23 @@ func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
 	return nil
 }
 
-func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLogger klog.Logger) error {
+func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLogger klog.Logger) ([]string, error) {
```
TortillaZHawaii: either use a type or write a comment on what the returned []string means

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -257,3 +297,33 @@ func (lc *StandaloneNEGLBController) clearStatusIngressIP(svc *v1.Service, svcLo
...
+	}
+
+	state.LBSchemeExternal = contains(schemes, "EXTERNAL")
+	state.LBSchemeExternalPassthrough = contains(schemes, "EXTERNAL_PASSTHROUGH")
+	state.LBSchemeInternal = contains(schemes, "INTERNAL")
+
+	lc.ctx.L4Metrics.SetL4StandaloneNEGService(key, state)
+}
+
+func contains(slice []string, s string) bool {
```
TortillaZHawaii: nit: slices.Contains does the same

pkg/l4/controllers/standalonenegcontroller_test.go
```diff
@@ -506,6 +511,28 @@ func TestStandaloneNEGLBSync(t *testing.T) {
 			expectError:        true,
 			expectEventReasons: []string{"ForwardingRuleUnusable"},
 		},
+		{
+			desc: "Clear status ingress IP when Service class is no longer standalone",
```
TortillaZHawaii: I thought class is immutable, is this is for deletes?

pkg/l4/controllers/standalonenegcontroller_test.go
```diff
@@ -664,3 +699,449 @@ func TestValidateForwardingRule(t *testing.T) {
...
+	state, ok := lc.ctx.L4Metrics.StandaloneNEGServiceState(svcKey)
+	if !ok {
+		t.Fatalf("Expected service %s in metrics map", svcKey)
+	}
+	if state.Status != l4metrics.StatusSuccess {
+		t.Errorf("Expected status %s, got %s", l4metrics.StatusSuccess, state.Status)
+	}
+	if !state.LBSchemeExternal {
+		t.Errorf("Expected LBSchemeExternal to be true")
+	}
```
TortillaZHawaii, line 765: nit: We could compare all values using cmp.Diff and having whole object there, so that both trues and falses are compared. same below

EDIT: I see state.FirstSyncErrorTime below, you can disregard it

pkg/l4/metrics/collector.go
```diff
@@ -40,6 +42,7 @@ func NewCollector(exportInterval, l4NetLBProvisionDeadline time.Duration, enable
 		l4NetLBServiceLegacyMap:                 make(map[string]L4NetLBServiceLegacyState),
 		l4ILBServiceMap:                         make(map[string]L4ServiceState),
 		l4NetLBServiceMap:                       make(map[string]L4ServiceState),
+		l4StandaloneNEGServiceMap:               make(map[string]L4StandaloneServiceState),
```
TortillaZHawaii: The naming is inconsistent, the ones I've found in this PR:
* L4StandaloneNEG
* StandaloneNEG
* L4Standalone
* L4StandaloneNEGLB

RN it's clear that it's the same thing, but with it mixing within the l4/ package it might be more difficult to onboard new persons and explain that it means same thing. Same as NetLB, XLB, ELB.

I would especially like to avoid LB, as IIUC there can be multiple GCE LBs pointing to single L4StandaloneNEGLB.

TortillaZHawaii (changes requested):
in the next prs let's split it into separate package

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: added defensive checks

pkg/l4/controllers/standalonenegcontroller.go

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: removed

pkg/l4/controllers/standalonenegcontroller_test.go
mmamczur: ok noted

pkg/l4/metrics/collector.go
mmamczur: ok now most should be `L4StandaloneNEG`

pkg/l4/controllers/standalonenegcontroller_test.go
mmamczur: removed this case as it doesn't make sense

pkg/l4/controllers/standalonenegcontroller.go
TortillaZHawaii: I meant that the XSupported should be included in the error message. So if you use PROTOCOL_SCTP you should get error message something along the lines: "forwarding rule ABC has unsupported protocol: SCTP. **Expected TCP, UDP, L3_DEFAULT**".

TortillaZHawaii: same goes for protocol and scheme

pkg/l4/controllers/standalonenegcontroller.go
TortillaZHawaii: this is not resolved

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -257,3 +317,24 @@ func (lc *StandaloneNEGLBController) clearStatusIngressIP(svc *v1.Service, svcLo
...
+		} else {
+			state.Status = l4metrics.StatusError
+			now := time.Now()
+			state.FirstSyncErrorTime = &now
+		}
+	}
+
+	state.LBSchemeExternal = slices.Contains(schemes, "EXTERNAL")
+	state.LBSchemeExternalPassthrough = slices.Contains(schemes, "EXTERNAL_PASSTHROUGH")
+	state.LBSchemeInternal = slices.Contains(schemes, "INTERNAL")
```
TortillaZHawaii, line 351: this is not needed anymore

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: now it is

pkg/l4/controllers/standalonenegcontroller.go

PR #3151, opened 2026-07-03
Ensure svcneg is available on local lister after creation

midavadim:
ensure svcneg is avaialble on local lister after creation, 
run syncers start in paralel

Previously, 
on 1st sync: on Service creation - EnsureSyncers creates svcnegs
on 2nd consequential sync (caused by Service.annotation update): EnsureSyncers checks again if svcneg exists (it did not) so it creates svcneg again. It has been already created in previous sync but it was not available on local lister yet.

So, almost on each service creation we get a Warning message::
```
~/src/k8s.io/ingress-gce$ k describe service/svc-standalone-neg-xxx
Events:
  Type     Reason                Age   From            Message
  ----     ------                ----  ----            -------
  Warning  ProcessServiceFailed  20s   neg-controller  error processing service "default/svc-standalone-neg-xxx": failed to ensure svc neg cr default/xneg/80 for port: servicenetworkendpointgroups.networking.gke.io "xneg" already exists
  Normal   Create                14s   neg-controller  Created NEG "xneg" for default/svc-standalone-neg-xxx-xneg-tcp-port/80-8080-GCE_VM_IP_PORT-L7 in "us-central1-a".
  Normal   Attach                11s   neg-controller  Attach 3 network endpoint(s) (NEG "xneg" in zone "us-central1-a")

```

Solution is: wait for short period of time when created svcneg appears in local lister.

### DIFF
--- a/pkg/multiproject/start/start_test.go
+++ b/pkg/multiproject/start/start_test.go
@@ -728,11 +728,18 @@ func verifyCloudNEG(
 	t.Helper()
 
 	const defaultRegion = "us-central1"
-	neg, err := gceCloud.GetNetworkEndpointGroup(negName, defaultRegion)
+	var lastErr error
+	err := wait.PollImmediate(shortTimeout, defaultTimeout, func() (bool, error) {
+		_, lastErr = gceCloud.GetNetworkEndpointGroup(negName, defaultRegion)
+		if lastErr != nil {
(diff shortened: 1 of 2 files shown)


mmamczur:
I recommend adding some more elaborate commit description so that it's clear why are we doing this change

PR #3152, opened 2026-07-06
Add NEG pre-provisioning support

midavadim:
1. NEG Pre-provisioning Annotation Support

In K8s Service, extend existing “cloud.google.com/neg” annotation with “zones” field:
```
{"exposed_ports": {...}, "zones":["*"]}  # all zones in the region
   OR
{"exposed_ports": {...}, "zones":["us-central1-a", "us-central1-b"]}
```
It forces the controller to create empty NEGs in specific zones (or all zones in the region) even if there is no workload in the zone.

(by default) If “zones” field is missing or set to empty list (controller works as usual):
NEGs will be created only in zones with workload.
Unused NEGs will be marked as inactive

If “zones” field is  ["*"] then:
NEGs will be created for all zones in the region

If “zones” field is ["us-central1-a", "us-central1-b"] then:
NEGs will be created in “zones with workload” and “zones from neg annotation”

If “zones” field contains incorrect values then field will be ignored and warning event will appear:
```
  Type     Reason                               Age              From            Message
  ----     ------                               ----             ----            -------
  Warning  IgnoreZonePreprovisioningAnnotation  7s (x4 over 8s)  neg-controller  Ignore zone pre-provisioning annotation because of incorrect value: zone "incorrect_value" is not in the cluster's region zones: [us-central1-c us-central1-a us-central1-f us-central1-b]
```

2. Internal behaviour:
Controller run Syncer.sync() on each change of 'zone' field in k8s Service neg annotation.
The change is identified by comparing with 'stored' value in manager.svcPreprovisioningZonesMap.
Values in 'zones' field are verified by checking with output of cloud.Zones(), which also cache successful results

3. Feature Flag & Verification
Feature can be enabled using controller flag: --enable-neg-preprovisioning = true

4. Metrics Collection
Introduced metrics to track the number of pre-provisioned NEGs.
Updated the metrics collector in pkg/neg/metrics/metricscollector/ and added corresponding unit tests.

```
curl http://localhost:8081/metrics | grep NEG

# HELP number_of_negs Number of NEGs
number_of_negs{feature="CustomNamedNEG"} 1
number_of_negs{feature="IngressNEG"} 0
number_of_negs{feature="NEG"} 2
number_of_negs{feature="PreprovisionedNEG"} 1                 <---
number_of_negs{feature="StandaloneNEG"} 2
number_of_negs{feature="VmIpNEG"} 0
number_of_service_ports{feature="NEG"} 0
```


pkg/neg/types/cloudprovideradapter.go
```diff
@@ -189,3 +192,22 @@ func (a *cloudProviderAdapter) Region() string {
 func (a *cloudProviderAdapter) GetNetwork(networkName string) (*compute.Network, error) {
 	return a.c.GetNetwork(networkName)
 }
+
+func (a *cloudProviderAdapter) Zones() ([]string, error) {
+	a.regionZonesCacheMu.Lock()
```
mmamczur, line 197: I wonder if it could be possible to use Once here https://go.dev/cs/go/+/go1.26.4:src/sync/once.go;l=52

pkg/neg/controller.go
```diff
@@ -901,6 +903,32 @@ func (c *Controller) syncNegStatusAnnotation(namespace, name string, portMap neg
...
 	}
 
+	// Get zones with nodes
+	zones, err := c.zoneGetter.ListZones(negtypes.NodeFilterForEndpointCalculatorMode(portMap.EndpointsCalculatorMode(), c.includeDrainNodesL4Local), c.logger)
+	if err != nil {
+		return err
+	}
+
+	if flags.F.EnableNEGPreprovisioning {
+		// Get preprovisioning zones from neg annotation
```
mmamczur, line 913: I'd recommend exctracting this part to a separate function

midavadim: moved

PR #3153, opened 2026-07-08
Topology provider fix for multinet

anyush:
Introduced in https://github.com/kubernetes/ingress-gce/pull/3121 `TopologyProvider.ListZonesPerSubnet`  in case when Multi-Net feature is used still returned zones for primary network's subnets, which resulted in empty zones set for non-primary networks' subnets. Making this receiver aware of which feature is used: Multi-Subnet or Multi-Net - and build `ZonesPerSubnet` map accordingly.

Also renamed `ListSubnets` into `ListSubnetsInDefaultNetwork` as it's better describes what it does.

### DIFF
--- a/pkg/utils/zonegetter/zone_getter.go
+++ b/pkg/utils/zonegetter/zone_getter.go
@@ -241,7 +241,7 @@ func (z *ZoneGetter) ListZonesInDefaultSubnet(filter Filter, logger klog.Logger)
 
 // ListZonesPerSubnet returns a list of zones containing nodes that satisfy the given node filtering mode per subnet.
 func (z *ZoneGetter) ListZonesPerSubnet(filter Filter, logger klog.Logger) (shared.ZonesPerSubnetMap, error) {
-	subnetConfigs := z.ListSubnets(logger)
+	subnetConfigs := z.ListSubnetsInDefaultNetwork(logger)
 	if subnetConfigs == nil {
 		return shared.ZonesPerSubnetMap{}, nil
 	}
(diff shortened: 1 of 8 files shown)


pkg/utils/zonegetter/zone_getter.go
```diff
@@ -240,19 +241,25 @@ func (z *ZoneGetter) ListZonesInDefaultSubnet(filter Filter, logger klog.Logger)
...
+	zones, err := z.ListZones(filter, logger)
+	// Don't return subnets with empty zones set. As ZoneGetter returns all zones per all subnets - don't return any.
+	if err != nil || len(zones) == 0 {
+		return shared.ZonesPerSubnetMap{}, err
 	}
 
-	zones, err := z.ListZones(filter, logger)
-	if err != nil {
-		return nil, err
+	// Multi-Net
```
mmamczur, line 251: so for multinet it's always going to be 1 subnet, the one the service is in?

anyush: Yes, Multi-Net and Multi-Subnet are mutually exclusive, so if Multi-Net is used, then there is only one subnet managed. That's also can be seen in ensuring NEGs: https://github.com/kubernetes/ingress-gce/blob/525d911a5b5a4e2a2a244cae7b4d58bca12455a9/pkg/neg/syncers/transaction.go#L565

Worth mention that for multi-subnet case we get subnets from `NodeTopology` CR, but it only exists for default network, so there seems to be no API for getting subnets for different network (which is Multi-Net case, which in fact is just Not-Default-Net).

PR #3157, opened 2026-07-09
Add conditions to the Standalone NEG

mmamczur:
This changes the L4 Standalone NEG Service controller to add conditions to the status of the service to signal the state of the service to the user.

### DIFF
--- a/pkg/l4/controllers/l4lbcommon.go
+++ b/pkg/l4/controllers/l4lbcommon.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"fmt"
 	"reflect"
+	"sort"
 
 	"k8s.io/ingress-gce/pkg/l4/resources"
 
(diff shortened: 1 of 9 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -253,7 +274,17 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 	}
 
 	if len(parsedRules) == 0 {
-		err := lc.clearStatusIngressIP(svc, svcLogger)
+		var reason lbConditionReason
+		var err error
+		if len(parseErrs) > 0 {
+			reason = InvalidForwardingRule
+			err = errors.Join(parseErrs...)
```
FelipeYepez: this err is never used

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -253,7 +274,17 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
...
+		} else {
+			reason = NoForwardingRuleRef
+			err = errors.New("service has no valid forwarding rule reference in annotation")
```
FelipeYepez: same for this err, it is never used

pkg/l4/controllers/l4lbcommon.go
```diff
@@ -106,24 +107,36 @@ func deleteAnnotation(ctx *context.ControllerContext, svc *v1.Service, annotatio
...
-
+// Conditions listed in conditionsToRemove will be deleted.
+func mergeConditions(existing, newConditions []metav1.Condition, conditionsToRemove []string) []metav1.Condition {
 	existingMap := make(map[string]metav1.Condition)
 	for _, cond := range existing {
 		existingMap[cond.Type] = cond
 	}
 
+	for _, condType := range conditionsToRemove {
+		delete(existingMap, condType)
```
FelipeYepez, line 121: I wonder what should we do or if it would happen that we try to delete and add a same condition type. Deletion would happen first making the condition lose its original LastTransitionTime. Maybe we should add it as a comment for the function to document this behavior

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -297,7 +328,8 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 	// if at least one FR was ok then we use it
 	if len(lbIngresses) == 0 {
 		// if none of the FRs is usable remove any that is possibly there
-		err := lc.clearStatusIngressIP(svc, svcLogger)
+		cond := NewConditionExternalIPProgrammedFalse(classifyError(errs[0]))
```
FelipeYepez, line 329: If multiple FRs fail validation for different reasons this would select only the first error and hide the rest possible errors. This makes sense since we can only add 1 reason but lets document it to acknowledge the limitation

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -319,14 +361,11 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
...
 func (lc *StandaloneNEGLBController) clearStatusIngressIP(svc *v1.Service, svcLogger klog.Logger) error {
-	if len(svc.Status.LoadBalancer.Ingress) == 0 {
-		return nil
-	}
 	newStatus := &v1.LoadBalancerStatus{
 		Ingress: nil,
 	}
 
-	if err := updateServiceStatus(lc.ctx, svc, newStatus, nil, svcLogger); err != nil {
+	if err := updateServiceStatus(lc.ctx, svc, newStatus, nil, []string{ExternalIPProgrammed}, svcLogger); err != nil {
```
FelipeYepez: For easier understanding I think it would be better to first save []string{ExternalIPProgrammed} in conditionsToRemove and then pass this variable as argument

FelipeYepez (changes requested):

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: it's not needed, removing. The condition doesn't need the error message, the event will have it

pkg/l4/controllers/standalonenegcontroller.go
mmamczur: it's not needed, removing. The condition doesn't need the error message, the event will have it

PR #3163, opened 2026-07-10
Read multiple IP addresses from Forwarding Rule in Standalone NEG

JumiDeluxe:
support multiple IPs per ForwardingRule for Global NetLB integration

- Iterate over the plural `IPAddresses` array in `composite.ForwardingRule` to support Global NetLB  setups.
- Populate all extracted IPs into the Service `status.loadBalancer.ingress`.
- Maintain backward compatibility by falling back to singular `IPAddress` for legacy/Regional NetLB forwarding rules.


pkg/l4/address/manager.go
```diff
@@ -47,8 +47,9 @@ const (
 type IPVersion = string
 
 const (
-	IPv4Version IPVersion = "IPV4"
-	IPv6Version IPVersion = "IPV6"
+	UnknownVersion IPVersion = "UNKNOWN"
```
mmamczur: usually in k8s world the naming was IPFamily

we ould also follow that here, so it would be IPv4Family and so on
but that's not very important

mmamczur: oh I see this type was defined earlier, lets forget about this then

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -293,13 +343,18 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "NoForwardingRuleRef", "Service has no forwarding rule reference")
 		return nil, l4utils.NewUserError(fmt.Errorf("service has no valid forwarding rule reference in annotation"))
 	}
+	if len(parsedRules) > ForwardingRulesLimit {
+		err := fmt.Errorf("number of forwarding rules (%d) exceeds the limit of %d", len(parsedRules), ForwardingRulesLimit)
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use Forwarding Rules")
+		return nil, err
```
mmamczur: let's not fail here, lets use at most 10 from the provided list and emit a warning that we skipped some of them

JumiDeluxe: Ok, removed fail and now we are using the first 10 ones. We will need to document that so it is clear that we are skipping remaining ones

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -243,6 +246,35 @@ func (lc *StandaloneNEGLBController) parseForwardingRuleKeys(frNamesStr string,
 	return parsedRules, errs
 }
 
+func getIPVersion(s string) (address.IPVersion, error) {
```
mmamczur: overall GO style guide discourages using functions named getX but this is debatable, just leaving FYI

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -243,6 +246,35 @@ func (lc *StandaloneNEGLBController) parseForwardingRuleKeys(frNamesStr string,
...
+		return address.UnknownVersion, err
+	}
+	if addr.Is4() {
+		return address.IPv4Version, nil
+	}
+	// Due to ParseAddr we assume that address will always be a valid IPv6 version if it is not IPv4.
+	return address.IPv6Version, nil
+}
+
+func validateIPVersions(fr *composite.ForwardingRule, frName string) error {
```
mmamczur: so this checks if we have mixed IPv4 and IPv6? let's add a comment on what it's doing. Also I'm wondering if we should do this. I don't know if it is possible for a FR to have IPv4 and v6 but if that was ever supported then we'd rather check if the service has the specific family and if so just use it

pkg/l4/controllers/standalonenegcontroller.go
JumiDeluxe: removed the function as it is no longer needed

pkg/l4/controllers/standalonenegcontroller.go
JumiDeluxe: removed the check and related tests, it should be fine to not check it here

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -318,13 +331,17 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "NoForwardingRuleRef", "Service has no forwarding rule reference")
 		return nil, l4utils.NewUserError(fmt.Errorf("service has no valid forwarding rule reference in annotation"))
 	}
+	if len(parsedRules) > ForwardingRulesLimit {
+		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Up to %d forwarding rules are supported. Skipping remaining forwarding rules.", ForwardingRulesLimit)
+		parsedRules = parsedRules[:ForwardingRulesLimit]
```
TortillaZHawaii, line 355: Is the order of parsedRules consistent between resyncs? If I have 20 FR, will I get always the same 10 FR here, or is there a chance it will flip flop?

Also we should probably go which Forwarding Rules where taken into account or which were skipped in Warning message.

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -276,9 +277,21 @@ func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
 	if !isFRProtocolSupported(fr.IPProtocol) {
 		errs = append(errs, l4utils.NewUnsupportedProtocolError(frName, fr.IPProtocol, supportedFRProtocols))
 	}
+	if len(fr.IPAddresses) > 2 {
```
TortillaZHawaii, line 281: while it may be unexpected to have more then 2 addresses right now, is it something that we should block on our controller? Is there something in future logic that depends on having max 2 addresses?

pkg/l4/address/manager.go
TortillaZHawaii: Is it referenced anywhere?

JumiDeluxe: Thanks, good catch, this will not be useful anymore because we are no longer checking for IP families. Might be useful in the future but not now, removing it.

pkg/l4/controllers/standalonenegcontroller.go
TortillaZHawaii: discussed f2f, resolving

pkg/l4/controllers/standalonenegcontroller.go
JumiDeluxe: Added alphabetical sort

PR #3167, opened 2026-07-13
Fix standalone NEG flag collision

FelipeYepez:
Decouple Standalone NEG controller flag from L4 leader election.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -86,6 +86,7 @@ var F = struct {
 	RunL4Controller              bool
 	RunL4NetLBController         bool
 	RunL4StandaloneNEGController bool
+	EnableL4StandaloneNEGs       bool
 	EnableIGController           bool
 	Version                      bool
 	WatchNamespace               string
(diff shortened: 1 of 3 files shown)


pkg/flags/flags.go
```diff
@@ -311,6 +312,7 @@ L7 load balancing. CSV values accepted. Example: -node-port-ranges=80,8080,400-5
 	flag.BoolVar(&F.RunL4Controller, "run-l4-controller", false, `Optional, whether or not to run L4 Service Controller as part of glbc. If set to true, services of Type:LoadBalancer with Internal annotation will be processed by this controller.`)
 	flag.BoolVar(&F.RunL4NetLBController, "run-l4-netlb-controller", false, `Optional, if enabled then the L4NetLbController will be run.`)
 	flag.BoolVar(&F.RunL4StandaloneNEGController, "run-l4-standalone-neg-controller", false, `Optional, if enabled then the Standalone NEG L4 LB controller will be run.`)
+	flag.BoolVar(&F.EnableL4StandaloneNEGs, "enable-l4-standalone-negs", false, `Optional, if enabled then the NEG controller will isolate generic NEG processing from L4 standalone NEG processing.`)
```
TortillaZHawaii: I'm not sure about *isolate generic NEG processing from L4 standalone NEG processing.*, the logic is the same based on the pkg/neg/controller.go#L752

The only thing different so far is that it is enabled for StandalonePassthroughNegLoadBalancerClass lbclass

FelipeYepez: Done

PR #3168, opened 2026-07-13
BYOIPv6: Read and use IP Collection annotation for LBs

linpu01:
GKE is introducing support for [Bring Your Own IP (BYOIP) IPv6](http://go/byoip-ipv6-gke) which allows users to assign external non-Google IPv6 addresses to clusters and load balancers (LBs). 

User can specify their BYOIP IPv6 addresses through the "networking.gke.io/ip-collection", the controller reserves an address with it and use it for Forwarding Rule creation.

User will get an error when specifying subnet and IP collection at the same time. When creating Forwarding Rules, the defaulted subnetURL will be omitted if ip-collection annotation is specified.

### DIFF
--- a/pkg/l4/address/manager.go
+++ b/pkg/l4/address/manager.go
@@ -29,6 +29,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/flags"
 	l4utils "k8s.io/ingress-gce/pkg/l4/utils"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/klog/v2"
@@ -60,19 +61,25 @@ const (
 
 // Original file in https://github.com/kubernetes/legacy-cloud-providers/blob/6aa80146c33550e908aed072618bd7f9998837f6/gce/gce_address_manager.go
 type Manager struct {
-	svc         gce.CloudAddressService
-	name        string
-	serviceName string
-	addressName string
-	targetIP    string
-	addressType cloud.LbScheme
-	region      string
-	subnetURL   string
-	tryRelease  bool
-	networkTier cloud.NetworkTier
-	ipVersion   IPVersion
+	svc          gce.CloudAddressService
+	name         string
+	serviceName  string
+	addressName  string
+	targetIP     string
+	addressType  cloud.LbScheme
+	region       string
+	subnetURL    string
+	tryRelease   bool
+	networkTier  cloud.NetworkTier
+	ipVersion    IPVersion
+	ipCollection string
 
 	frLogger klog.Logger
+}
+
+// SetIPCollection sets the ipCollection for the manager.
+func (m *Manager) SetIPCollection(ipCollection string) {
+	m.ipCollection = ipCollection
 }
 
 func NewManager(svc gce.CloudAddressService, serviceName, region, subnetURL, name, addressName, targetIP string, addressType cloud.LbScheme, networkTier cloud.NetworkTier, ipVersion IPVersion, frLogger klog.Logger) *Manager {
@@ -261,11 +268,12 @@ func (m *Manager) ensureAddressReservation() (string, IPAddressType, error) {
 	// Try reserving the IP with controller-owned address name
 	// If am.targetIP is an empty string, a new IP will be created.
 	newAddr := &compute.Address{
-		Name:        m.name,
-		Description: fmt.Sprintf(`{"kubernetes.io/service-name":"%s"}`, m.serviceName),
-		Address:     m.targetIP,
-		AddressType: string(m.addressType),
-		Subnetwork:  m.subnetURL,
+		Name:         m.name,
+		Description:  fmt.Sprintf(`{"kubernetes.io/service-name":"%s"}`, m.serviceName),
+		Address:      m.targetIP,
+		AddressType:  string(m.addressType),
+		Subnetwork:   m.subnetURL,
+		IpCollection: m.ipCollection,
 	}
 	// NetworkTier is supported only for External IP Address
 	if m.addressType == cloud.SchemeExternal {
@@ -386,6 +394,9 @@ func (m *Manager) validateAddress(addr *compute.Address) error {
 	}
 	if addr.NetworkTier != m.networkTier.ToGCEValue() {
 		return l4utils.NewNetworkTierErr(fmt.Sprintf("Static IP (%v)", m.name), m.networkTier.ToGCEValue(), addr.NetworkTier)
+	}
+	if flags.F.EnableBYOIPv6 && m.ipCollection != addr.IpCollection {
+		return fmt.Errorf("ipCollection mismatch, expected %q, actual: %q", m.ipCollection, addr.IpCollection)
 	}
 	return nil
 }
--- a/pkg/l4/forwardingrules/equal.go
+++ b/pkg/l4/forwardingrules/equal.go
@@ -52,7 +52,8 @@ func EqualIPv6(fr1, fr2 *composite.ForwardingRule) (bool, error) {
 		fr1.AllowGlobalAccess == fr2.AllowGlobalAccess &&
 		fr1.AllPorts == fr2.AllPorts &&
 		fr1.Subnetwork == fr2.Subnetwork &&
-		fr1.NetworkTier == fr2.NetworkTier, nil
+		fr1.NetworkTier == fr2.NetworkTier &&
+		equalResourcePaths(fr1.IpCollection, fr2.IpCollection), nil
 }
 
 // equalPorts compares two port ranges or slices of ports. Before comparison,
--- a/pkg/l4/resources/forwarding_rules_ipv6.go
+++ b/pkg/l4/resources/forwarding_rules_ipv6.go
@@ -190,6 +190,14 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 	if err != nil {
 		return nil, l4utils.ResourceResync, fmt.Errorf("error getting ipv6 forwarding rule subnet: %w", err)
 	}
+
+	var ipCollection string
+	if flags.F.EnableBYOIPv6 {
+		ipCollection = annotations.FromService(l4netlb.Service).GetIPCollectionV6()
+	}
+	if ipCollection != "" {
+		subnetworkURL = ""
+	}
 	frLogger.V(2).Info("subnetworkURL for service", "subnetworkURL", subnetworkURL)
 
 	// Determine IP which will be used for this LB. If no forwarding rule has been established
@@ -199,6 +207,12 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 		frLogger.Error(err, "address.IPv6ToUse for service returned error")
 		return nil, l4utils.ResourceResync, err
 	}
+
+	// If the IP collection changed, we cannot reuse the old forwarding rule's IP.
+	if fwdRuleToDetermineIP != nil && fwdRuleToDetermineIP.IpCollection != ipCollection {
+		ipv6AddrToUse = ""
+	}
+
 	frLogger.V(2).Info("ipv6AddressToUse for service", "ipv6AddressToUse", ipv6AddrToUse)
 
 	netTier, isFromAnnotation := annotations.NetworkTier(l4netlb.Service)
@@ -214,6 +228,7 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 	if !l4netlb.cloud.IsLegacyNetwork() && netTier == cloud.NetworkTierPremium {
 		nm := types.NamespacedName{Namespace: l4netlb.Service.Namespace, Name: l4netlb.Service.Name}.String()
 		addrMgr := address.NewManager(l4netlb.cloud, nm, l4netlb.cloud.Region(), subnetworkURL, expectedIPv6FrName, ipv6AddressName, ipv6AddrToUse, cloud.SchemeExternal, netTier, address.IPv6Version, frLogger)
+		addrMgr.SetIPCollection(ipCollection)
 
 		// If network tier annotation in Service Spec is present
 		// check if it matches network tiers from forwarding rule and external ip Address.
@@ -333,6 +348,11 @@ func (l4netlb *L4NetLB) buildExpectedIPv6ForwardingRule(bsLink, ipv6AddressToUse
 		}
 	}
 
+	var ipCollection string
+	if flags.F.EnableBYOIPv6 {
+		ipCollection = annotations.FromService(l4netlb.Service).GetIPCollectionV6()
+	}
+
 	fr := &composite.ForwardingRule{
 		Name:                frName,
 		Description:         frDesc,
@@ -346,6 +366,7 @@ func (l4netlb *L4NetLB) buildExpectedIPv6ForwardingRule(bsLink, ipv6AddressToUse
 		Subnetwork:          subnetworkURL,
 		AllPorts:            allPorts,
 		Ports:               ports,
+		IpCollection:        ipCollection,
 	}
 
 	return fr, nil
--- a/pkg/l4/resources/l4.go
+++ b/pkg/l4/resources/l4.go
@@ -424,6 +424,12 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	}
 	l4.network = *svcNetwork
 
+	// Check if ip-collection is specified for IPv4 service
+	if flags.F.EnableBYOIPv6 && annotations.FromService(svc).GetIPCollectionV6() != "" && l4utils.NeedsIPv4(svc) {
+		err := fmt.Errorf("%s is currently only supported for IPv6-only external LBs", annotations.IPCollectionV6AnnotationKey)
+		l4.recorder.Event(l4.Service, corev1.EventTypeWarning, "IPCollectionV6Error", err.Error())
+	}
+
 	// If service requires IPv6 LoadBalancer -- verify that Subnet with Internal IPv6 ranges is used.
 	if l4.enableDualStack && l4utils.NeedsIPv6(l4.Service) {
 		err := l4.serviceSubnetHasInternalIPv6Range()
@@ -494,6 +500,7 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	var ipv6AddressName string
 	if l4.enableDualStack && l4utils.NeedsIPv6(l4.Service) {
 		existingIPv6FR, err = l4.getOldIPv6ForwardingRule(existingBS)
+
 		ipv6AddrToUse, ipv6AddressName, err = address.IPv6ToUse(l4.cloud, l4.Service, existingIPv6FR, subnetworkURL, l4.svcLogger)
 		if err != nil {
 			result.Error = fmt.Errorf("EnsureInternalLoadBalancer error: address.IPv6ToUse returned error: %w", err)
--- a/pkg/l4/resources/l4_test.go
+++ b/pkg/l4/resources/l4_test.go
@@ -3439,3 +3439,50 @@ func TestEnsureInternalLoadBalancer_L4LBConfigLogging(t *testing.T) {
 		})
 	}
 }
+
+func TestEnsureInternalLoadBalancerIPCollectionError(t *testing.T) {
+	flags.F.EnableBYOIPv6 = true
+	defer func() { flags.F.EnableBYOIPv6 = false }()
+	nodeNames := []string{"test-node-1"}
+	vals := gce.DefaultTestClusterValues()
+	fakeGCE := getFakeGCECloud(vals)
+
+	svc := test.NewL4ILBService(false, 8080)
+	if svc.Annotations == nil {
+		svc.Annotations = make(map[string]string)
+	}
+	svc.Annotations[annotations.IPCollectionV6AnnotationKey] = "my-collection"
+
+	namer := namer_util.NewL4Namer(kubeSystemUID, namer_util.NewNamer(vals.ClusterName, "cluster-fw", klog.TODO()))
+
+	l4Params := &L4ILBParams{
+		Service:         svc,
+		Cloud:           fakeGCE,
+		Namer:           namer,
+		Recorder:        record.NewFakeRecorder(100),
+		NetworkResolver: network.NewFakeResolver(network.DefaultNetwork(fakeGCE)),
+	}
+	l4 := NewL4Handler(l4Params, klog.TODO())
+
+	if _, err := test.CreateAndInsertNodes(l4.cloud, nodeNames, vals.ZoneName); err != nil {
+		t.Errorf("Unexpected error when adding nodes %v", err)
+	}
+
+	result := l4.EnsureInternalLoadBalancer(nodeNames, svc)
+	if result.Error != nil {
+		t.Errorf("Expected no error for conflicting annotations, but got %v", result.Error)
+	}
+
+	fakeRecorder := l4.recorder.(*record.FakeRecorder)
+	warningFound := false
+	for len(fakeRecorder.Events) > 0 {
+		event := <-fakeRecorder.Events
+		if strings.Contains(event, "Warning IPCollectionV6Error") {
+			warningFound = true
+			break
+		}
+	}
+	if !warningFound {
+		t.Errorf("Expected IPCollectionV6Error warning event, but got none")
+	}
+}
--- a/pkg/l4/resources/l4netlb.go
+++ b/pkg/l4/resources/l4netlb.go
@@ -271,6 +271,47 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service, 
 
 	l4netlb.networkInfo = *networkInfo
 
+	var ipCollectionV6 string
+	if flags.F.EnableBYOIPv6 {
+		ipCollectionV6 = annotations.FromService(svc).GetIPCollectionV6()
+	}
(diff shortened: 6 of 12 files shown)


pkg/l4/address/hold.go
gnossen, line 38: This method seems to be IPv4 specific?

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,15 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
 
 	l4netlb.networkInfo = *networkInfo
 
+	// Check conflicts between custom subnet and ip-collection annotations.
+	if annotations.FromService(svc).GetIPCollection() != "" && annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet() != "" {
+		err := fmt.Errorf("cannot specify both custom subnet and ip-collection for LoadBalancer")
```
gnossen: Please include the actual keys and actual values. This is going to be user-facing and users need to know what specific actions to take to resolve errors.

pkg/l4/resources/forwarding_rules_ipv6.go
```diff
@@ -204,6 +209,11 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 	netTier, isFromAnnotation := annotations.NetworkTier(l4netlb.Service)
 	frLogger.V(2).Info("network tier for service", "networkTier", netTier, "isFromAnnotation", isFromAnnotation)
 
+	ipCollection := annotations.FromService(l4netlb.Service).GetIPCollection()
+	if ipCollection != "" {
+		subnetworkURL = ""
```
gnossen: It looks like you've added a validation that presumably runs before this function in `pkg/l4/resources/l4.go`. Is this line of code actually unreachable? If so, can we delete it? Or perhaps return an error instead? Same comment for the similar pattern lower in this file.

linpu01: The validation before this is for user specified subnet, however even if user doesn't specify subnet, the controller defaults to use cluster's subnet, here is for removing the defaulted subnet.

pkg/l4/resources/l4netlb.go
linpu01: Done, added annotation key and value

gnossen (approved):

pkg/l4/address/hold_test.go
```diff
@@ -37,6 +37,7 @@ func TestHoldExternalIPv4(t *testing.T) {
 		service       *api_v1.Service
 		want          address.HoldResult
 		wantTearDown  bool
+		wantIpCol     string
```
mmamczur: there is no new test case where we expect a different behavior, is it needed?

linpu01: Done, removed

pkg/l4/resources/forwarding_rules.go
```diff
@@ -246,6 +246,7 @@ func (l4netlb *L4NetLB) ensureIPv4ForwardingRule(bsLink string) (*composite.Forw
 		LoadBalancingScheme: string(cloud.SchemeExternal),
 		BackendService:      bsLink,
 		NetworkTier:         netTier.ToGCEValue(),
+		IpCollection:        annotations.FromService(l4netlb.Service).GetIPCollectionV6(),
```
FelipeYepez: This shouldn't be used here since this is used for IPv4 only and there is no need to assign an IPv6 IpCollection to an IPv4 resource

pkg/l4/resources/l4.go
```diff
@@ -424,6 +424,36 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
...
+
+	// Check if ip-collection-v6 is specified for IPv4 service
+	if annotations.FromService(svc).GetIPCollectionV6() != "" && l4utils.NeedsIPv4(svc) {
+		err := fmt.Errorf("networking.gke.io/ip-collection-v6 is currently only supported for IPv6-only Services")
+		result.Error = l4utils.NewUserError(err)
+		result.MetricsLegacyState.IsUserError = true
+		result.MetricsState.Status = metrics.StatusUserError
+		return result
+	}
+
```
FelipeYepez: If I understand correctly all the changes in this file should be discarded since this is used for the Internal LoadBalancer implementation. Internal LoadBalancers should not require the ip-collection because they pick IPs directly from the subnet.

pkg/l4/resources/l4.go
```diff
@@ -424,6 +424,36 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	}
 	l4.network = *svcNetwork
 
+	// Check conflicts between custom subnet and ip-collection
```
FelipeYepez: What we could do here is create a Warning event if the internal LoadBalancer service specifies an ip-collection

Something like `l4.recorder.Eventf(l4.Service, corev1.EventTypeWarning, REASON, MESSAGE)`

pkg/l4/resources/l4_test.go
```diff
@@ -3439,3 +3440,161 @@ func TestEnsureInternalLoadBalancer_L4LBConfigLogging(t *testing.T) {
 		})
 	}
 }
+
+func TestEnsureInternalLoadBalancerConflictingAnnotations(t *testing.T) {
```
FelipeYepez: Same here, since these tests are for ILB my understanding is that they should be discarded

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
 
 	l4netlb.networkInfo = *networkInfo
 
+	// Check conflicts between custom subnet and ip-collection annotations.
+	if annotations.FromService(svc).GetIPCollectionV6() != "" && annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet() != "" {
+		subnet := annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet()
+		ipCollection := annotations.FromService(svc).GetIPCollectionV6()
+		err := fmt.Errorf("cannot specify both networking.gke.io/load-balancer-subnet (%q) and networking.gke.io/ip-collection-v6 (%q) for LoadBalancer", subnet, ipCollection)
```
FelipeYepez: Please use the exported constants for these annotation keys since they could change in the future and it would be better not to hard code them in the error message

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
...
+		result.Error = l4utils.NewUserError(err)
+		result.MetricsLegacyState.IsUserError = true
+		result.MetricsState.Status = metrics.StatusUserError
+		return result
```
FelipeYepez, line 284: Not sure what is the intention here (and below) but I would say that it would be better not to return an error since this would prevent the service from being created. Maybe just logging a warning event on the conflict and ignoring the subnet should be enough? wdyt @mmamczur @08volt

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
...
+	}
+
+	// Check conflicts between spec.loadBalancerIP and ip-collection
```
FelipeYepez, line 287: we should also check for the `networking.gke.io/load-balancer-ip-addresses` annotation

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
...
+	if annotations.FromService(svc).GetIPCollectionV6() != "" && svc.Spec.LoadBalancerIP != "" {
+		ipCollection := annotations.FromService(svc).GetIPCollectionV6()
+		err := fmt.Errorf("cannot specify both spec.LoadBalancerIP (%q) and networking.gke.io/ip-collection-v6 (%q) for LoadBalancer", svc.Spec.LoadBalancerIP, ipCollection)
```
FelipeYepez: same here, please don't hard code annotation keys as part of error, import them

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
...
+		return result
+	}
+
+	// Check if ip-collection-v6 is specified for an IPv4 service
+	if annotations.FromService(svc).GetIPCollectionV6() != "" && l4utils.NeedsIPv4(svc) {
+		err := fmt.Errorf("networking.gke.io/ip-collection-v6 is currently only supported for IPv6-only Services")
+		result.Error = l4utils.NewUserError(err)
+		result.MetricsLegacyState.IsUserError = true
+		result.MetricsState.Status = metrics.StatusUserError
+		return result
```
FelipeYepez, line 312: same here, using a warning event could be enough instead of returning an error and not processing the service

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,36 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
 
 	l4netlb.networkInfo = *networkInfo
 
+	// Check conflicts between custom subnet and ip-collection annotations.
+	if annotations.FromService(svc).GetIPCollectionV6() != "" && annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet() != "" {
+		subnet := annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet()
```
FelipeYepez: Declare and assign variable at the top of the validation block to reduce overhead and improve readability. 

```
subnet := annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet()
ipCollection := annotations.FromService(svc).GetIPCollectionV6()

if ipCollection != "" && subnet != "" {
...
```

FelipeYepez: same for other places in the modified code

pkg/l4/resources/l4netlb_test.go
```diff
@@ -2231,3 +2245,201 @@ func TestEnsureL4NetLB_L4LBConfigLogging(t *testing.T) {
...
+		t.Errorf("Expected an error for conflicting annotations, but got nil")
+	} else if !strings.Contains(result.Error.Error(), "cannot specify both networking.gke.io/load-balancer-subnet (\"my-subnet\") and networking.gke.io/ip-collection-v6 (\"my-collection\") for LoadBalancer") {
+		t.Errorf("Expected conflict error, got %v", result.Error)
+	}
+	if result.MetricsState.Status != metrics.StatusUserError {
+		t.Errorf("Expected UserError status, got %v", result.MetricsState.Status)
+	}
+}
+
+func TestEnsureFrontendConflictingIPAnnotations(t *testing.T) {
```
FelipeYepez, line 2324: We should also check for conflicts between ip-collection and networking.gke.io/load-balancer-ip-addresses

FelipeYepez (changes requested):

pkg/l4/resources/forwarding_rules.go
linpu01: Done, removed

pkg/l4/resources/l4.go
linpu01: Done, removed, only kept the one for the warning event when ip-collection is specified

pkg/l4/resources/l4.go
linpu01: Done

pkg/l4/resources/l4_test.go
linpu01: Done, removed

pkg/l4/resources/l4netlb.go
linpu01: Done

pkg/l4/resources/l4netlb.go
linpu01: Done

pkg/l4/resources/l4netlb.go
linpu01: Added the check

pkg/l4/resources/l4netlb.go
linpu01: Done

pkg/l4/resources/l4netlb.go
linpu01: Done

pkg/l4/resources/l4netlb_test.go
linpu01: Added test

pkg/l4/resources/forwarding_rules_ipv6.go
```diff
@@ -99,6 +99,10 @@ func (l4 *L4) buildExpectedIPv6ForwardingRule(bsLink string, options gce.ILBOpti
 		}
 	}
 
+	if annotations.FromService(l4.Service).GetIPCollectionV6() != "" {
```
08volt: This is related to ILBs and should be removed

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,33 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
 
 	l4netlb.networkInfo = *networkInfo
 
+	// Check conflicts between custom subnet and ip-collection annotations.
+	ipCollectionV6 := annotations.FromService(svc).GetIPCollectionV6()
+	subnet := annotations.FromService(svc).GetExternalLoadBalancerAnnotationSubnet()
+	if ipCollectionV6 != "" && subnet != "" {
+		err := fmt.Errorf("cannot specify both %s (%q) and %s (%q) for LoadBalancer", annotations.CustomSubnetAnnotationKey, subnet, annotations.IPCollectionV6AnnotationKey, ipCollectionV6)
+		l4netlb.recorder.Event(l4netlb.Service, corev1.EventTypeWarning, "IPCollectionV6Error", err.Error())
```
08volt: I think it would be best to return an error (`l4utils.NewUserError `) in this case to avoid a conflict in the API call later.

pkg/l4/resources/forwarding_rules_ipv6.go
08volt, line 61: EqualIPv6 does not check for a change in IPCollection. If the user changes the annotation's value, the FR is never recreated.

pkg/l4/resources/l4netlb.go
08volt: I think we should return an error here to avoid a conflict in the API call later

pkg/l4/resources/forwarding_rules_ipv6.go
08volt: This is related to ILBs where IPCollection should not be supported

pkg/l4/resources/forwarding_rules_ipv6.go
```diff
@@ -190,6 +195,11 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 	if err != nil {
 		return nil, l4utils.ResourceResync, fmt.Errorf("error getting ipv6 forwarding rule subnet: %w", err)
 	}
+
+	ipCollection := annotations.FromService(l4netlb.Service).GetIPCollectionV6()
+	if ipCollection != "" {
+		subnetworkURL = ""
```
08volt, line 199: The Subnetwork field is checked when comparing Forwarding Rules in the EqualIPv6 function. 
We need to be sure this field is not populated on the GCE side to avoid a delete-recreation loop

pkg/l4/resources/l4netlb.go
```diff
@@ -271,6 +271,33 @@ func (l4netlb *L4NetLB) EnsureFrontend(nodeNames []string, svc *corev1.Service,
...
+	}
+
+	// Check conflicts between spec.loadBalancerIP and ip-collection
+	if ipCollectionV6 != "" && svc.Spec.LoadBalancerIP != "" {
+		err := fmt.Errorf("cannot specify both spec.LoadBalancerIP (%q) and %s (%q) for LoadBalancer", svc.Spec.LoadBalancerIP, annotations.IPCollectionV6AnnotationKey, ipCollectionV6)
+		l4netlb.recorder.Event(l4netlb.Service, corev1.EventTypeWarning, "IPCollectionV6Error", err.Error())
+	}
+
+	// Check conflicts between load-balancer-ip-addresses and ip-collection
+	staticL4Addresses, _ := svc.Annotations[annotations.StaticL4AddressesAnnotationKey]
```
08volt: no need for ", _"

pkg/l4/resources/forwarding_rules_ipv6.go
FelipeYepez: Good point, @linpu01 can you please make sure that when performing the manual verification after th service is created the subnetworkURL of the Forwarding Rule is also empty.

pkg/l4/resources/forwarding_rules_ipv6.go
```diff
@@ -346,6 +352,7 @@ func (l4netlb *L4NetLB) buildExpectedIPv6ForwardingRule(bsLink, ipv6AddressToUse
 		Subnetwork:          subnetworkURL,
 		AllPorts:            allPorts,
 		Ports:               ports,
+		IpCollection:        annotations.FromService(l4netlb.Service).GetIPCollectionV6(),
```
FelipeYepez: It would be better if you save the value of the annotation in a variable and then use that variable here

pkg/l4/forwardingrules/equal.go
```diff
@@ -52,7 +52,8 @@ func EqualIPv6(fr1, fr2 *composite.ForwardingRule) (bool, error) {
 		fr1.AllowGlobalAccess == fr2.AllowGlobalAccess &&
 		fr1.AllPorts == fr2.AllPorts &&
 		fr1.Subnetwork == fr2.Subnetwork &&
-		fr1.NetworkTier == fr2.NetworkTier, nil
+		fr1.NetworkTier == fr2.NetworkTier &&
+		equalResourcePaths(fr1.IpCollection, fr2.IpCollection), nil
```
FelipeYepez, line 56: Is the IpCollection an URL? I see you now use equalResourcePaths and from what I understand this expects URLs

pkg/l4/resources/l4netlb.go
FelipeYepez: My bad, @linpu01 we indeed need to return an error as you first did to prevent conflicting configurations between ipCollection and custom subnet annotation / LoadBalancerIP / load-balancer-ip-addresses annotation. Please remove the warning I suggested and lets return an error for the multiple verifications in this file. For ILB it is fine to keep the warning

pkg/l4/address/manager.go
```diff
@@ -261,11 +267,12 @@ func (m *Manager) ensureAddressReservation() (string, IPAddressType, error) {
...
-		Address:     m.targetIP,
-		AddressType: string(m.addressType),
-		Subnetwork:  m.subnetURL,
+		Name:         m.name,
+		Description:  fmt.Sprintf(`{"kubernetes.io/service-name":"%s"}`, m.serviceName),
+		Address:      m.targetIP,
+		AddressType:  string(m.addressType),
+		Subnetwork:   m.subnetURL,
+		IpCollection: m.ipCollection,
 	}
```
FelipeYepez, line 277: From what I understood in your doc, if a specific IP address is provided, we must omit both the subnetwork and the IP collection fields so GCE can find the source. However, this code passes both m.targetIP and m.ipCollection. 

What should really be done here to reserve the address? Wasn't this giving errors?

pkg/l4/address/manager.go
FelipeYepez, line 388: This function checks if the address matches the target IP, type, and tier, but it completely misses checking the IpCollection. Is it possible to add a check to ensure the existing IP address belongs to the requested IP collection?

pkg/l4/resources/l4netlb_test.go
```diff
@@ -2241,3 +2245,275 @@ func TestEnsureL4NetLB_L4LBConfigLogging(t *testing.T) {
 		})
 	}
 }
+
+func TestIPv6OnlyNetLBIPCollectionAnnotation(t *testing.T) {
```
FelipeYepez, line 2249: Sorry for the confusion, we indeed need to return an error with the conflicting annotations, can you please update the tests back to checking the LoadBalancer is not provisioned

pkg/l4/resources/forwarding_rules_ipv6.go
FelipeYepez, line 216: If a user updates an existing Service to use a new ip-collection, the old IP address extracted from the existing Forwarding Rule will be invalid. If we pass this old IP to the Address Manager, GCP will throw an error because the old IP doesn't belong to the new collection.

We need to clear ipv6AddrToUse when the collection changes so the Address Manager allocates a new IP. Please add this check right after address.IPv6ToUse returns:

```
ipv6AddrToUse, ipv6AddressName, err := address.IPv6ToUse(l4netlb.cloud, l4netlb.Service, fwdRuleToDetermineIP, subnetworkURL, frLogger)
if err != nil {
	// ... existing error handling
}

// If the IP collection changed, we cannot reuse the old forwarding rule's IP.
if fwdRuleToDetermineIP != nil && fwdRuleToDetermineIP.IpCollection != ipCollection {
	ipv6AddrToUse = ""
}
```

FelipeYepez (changes requested):
Please safely gate this new logic behind a feature flag (e.g., flags.F.EnableBYOIPv6).

You need to apply the flag guard at every point where the ip-collection annotation changes the controller's default behavior.

pkg/l4/forwardingrules/equal.go
linpu01: I checked the log and the IP collection in the FR we get from GCE is a URL starts with 'https://' so this function is used, or there will be a sync loop if user provided URL doesn't start with 'https://' but with 'projects/' that keep recreating the FR.

pkg/l4/resources/forwarding_rules_ipv6.go
linpu01: Yes I checked the log and did a gcloud describe, the FR we created doesn't contain subnetworkURL

pkg/l4/resources/forwarding_rules_ipv6.go
linpu01: Done, added change in EqualIPv6

linpu01: Done, added code

pkg/l4/resources/l4netlb_test.go
linpu01: Done, switched back to returning errors

pkg/l4/address/manager.go
linpu01: We disallow user to provide both IP address and IP collection and from the log I see during creation this m.targetIP is empty and has no error.

pkg/l4/address/manager.go
linpu01: Done, added the check

pkg/l4/resources/forwarding_rules_ipv6.go
linpu01: Done

pkg/l4/address/manager.go
FelipeYepez: Yes, it would be safer to explicitly omit m.ipCollection if m.targetIP is populated or the other way around ensuring GCE can always properly determine the IP source

pkg/l4/resources/l4.go
```diff
@@ -424,6 +424,12 @@ func (l4 *L4) EnsureInternalLoadBalancer(nodeNames []string, svc *corev1.Service
 	}
 	l4.network = *svcNetwork
 
+	// Check if ip-collection is specified for IPv4 service
+	if flags.F.EnableBYOIPv6 && annotations.FromService(svc).GetIPCollectionV6() != "" && l4utils.NeedsIPv4(svc) {
+		err := fmt.Errorf("%s is currently only supported for IPv6-only Services", annotations.IPCollectionV6AnnotationKey)
```
08volt: Should warn also in IPv6-only clusters that it is only available for external LBs

pkg/l4/resources/forwarding_rules_ipv6.go
08volt, line 242: I am not sure if this would work in case of the user setting a new ipCollection.

- ipv6AddrToUse is "".
- ipv6AddressName defaults to the address name (e.g., k8s2-tcp-...) because there is no static IP annotation provided (which is blocked from being used with ip-collection-v6).

It gets the address by its name, which still exists:
```
addr, err := m.svc.GetRegionAddress(addressName, m.region)
```
than attempts to validate it:
```
	if addr != nil {
		// If address exists, check if the address had the expected attributes.
		validationError := m.validateAddress(addr)
```

Inside validateAddress, it throws an error because the collections do not match:

```
	if flags.F.EnableBYOIPv6 && m.ipCollection != addr.IpCollection {
		return fmt.Errorf("ipCollection mismatch, expected %q, actual: %q", m.ipCollection, addr.IpCollection)
	}
```

Which will return a sync fail instead of removing the forwarding rule

linpu01: I have tested the case where I update ip-collection from A to B and it worked, the previous FR is removed and the new address comes from new ip-collection. [gpaste](https://paste.googleplex.com/5430910102142976)

Probably because the address is released on GCE side after the previous FR was created 
```
defer func() {
	// Release the address that was reserved, in all cases. If the forwarding rule was successfully created,
	// the ephemeral IP is not needed anymore. If it was not created, the address should be released to prevent leaks.
	if err := addrMgr.ReleaseAddress(); err != nil {
		frLogger.Error(err, "ensureIPv6ForwardingRule: failed to release address reservation, possibly causing an orphan")
	}
}()
```

Then GetRegionAddress gets an not found error
```
addr, err := m.svc.GetRegionAddress(addressName, m.region)
```

pkg/l4/resources/l4.go
linpu01: Done, added to mention that this is only for IPv6 external LBs

PR #3169, opened 2026-07-16
NEG Binding Core

anyush:
Core functionality for NEG binding - if NEGBinding CR is created and references existing service and existing NEGs - manage NEGs' endpoints (sync them with service's ones).

Not covered by core, but will be in further PRs:
* Conflicts resolving (force only one NB CR to manage single NEG name) (https://github.com/kubernetes/ingress-gce/pull/3177)
* Cleanup (if NEG was managed by NB CR, but not anymore - remove all endpoints) (https://github.com/kubernetes/ingress-gce/pull/3180)
* Verifying NEGs' description
* Reporting bound NEGs' number in metrics
* Readiness gate of the pods, which are added to NEGs using binding feature

Feature is disabled by default using `enable-neg-binding` flag.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -45,6 +45,9 @@ import (
 	l4lbconfigclient "k8s.io/ingress-gce/pkg/l4lbconfig/client/clientset/versioned"
 	informerl4lbconfig "k8s.io/ingress-gce/pkg/l4lbconfig/client/informers/externalversions/l4lbconfig/v1"
 	"k8s.io/ingress-gce/pkg/metrics"
+	"k8s.io/ingress-gce/pkg/neg"
+	negbindingclient "k8s.io/ingress-gce/pkg/negbinding/client/clientset/versioned"
+	informernegbinding "k8s.io/ingress-gce/pkg/negbinding/client/informers/externalversions/negbinding/v1beta1"
 	"k8s.io/ingress-gce/pkg/recorders"
 	serviceattachmentclient "k8s.io/ingress-gce/pkg/serviceattachment/client/clientset/versioned"
 	informerserviceattachment "k8s.io/ingress-gce/pkg/serviceattachment/client/informers/externalversions/serviceattachment/v1"
@@ -69,6 +72,7 @@ const (
 type ControllerContext struct {
 	KubeClient          kubernetes.Interface
 	SvcNegClient        svcnegclient.Interface
+	NEGBindingClient    negbindingclient.Interface
 	SAClient            serviceattachmentclient.Interface
 	FirewallClient      firewallclient.Interface
 	EventRecorderClient kubernetes.Interface
@@ -91,6 +95,7 @@ type ControllerContext struct {
 	NodeInformer             cache.SharedIndexInformer
 	EndpointSliceInformer    cache.SharedIndexInformer
 	SvcNegInformer           cache.SharedIndexInformer
+	NEGBindingInformer       cache.SharedIndexInformer
 	SAInformer               cache.SharedIndexInformer
 	FirewallInformer         cache.SharedIndexInformer
 	NetworkInformer          cache.SharedIndexInformer
@@ -154,6 +159,7 @@ func NewControllerContext(
 	frontendConfigClient frontendconfigclient.Interface,
 	firewallClient firewallclient.Interface,
 	svcnegClient svcnegclient.Interface,
+	negBindingClient negbindingclient.Interface,
 	saClient serviceattachmentclient.Interface,
 	networkClient networkclient.Interface,
 	nodeTopologyClient nodetopologyclient.Interface,
@@ -182,6 +188,7 @@ func NewControllerContext(
 		KubeClient:              kubeClient,
 		FirewallClient:          firewallClient,
 		SvcNegClient:            svcnegClient,
+		NEGBindingClient:        negBindingClient,
 		SAClient:                saClient,
 		EventRecorderClient:     eventRecorderClient,
 		NodeTopologyClient:      nodeTopologyClient,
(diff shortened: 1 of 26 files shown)


pkg/neg/negbinding_manager.go
```diff
@@ -0,0 +1,537 @@
...
+		if !ok {
+			errs = append(errs, fmt.Errorf("unexpected object type %T in binding index for service %s", obj, svcKey))
+			continue
+		}
+		if networkInfoErr != nil {
+			_ = m.updateBackendRefCondition(binding, networkInfoErr)
+			continue
+		}
+
+		_, err := m.ensureSyncerForNEGBinding(binding, svc, networkInfo)
```
08volt: Shouldn't get the syncer and call the Sync()?
```
syncer, err := m.ensureSyncerForNEGBinding(binding, svc, networkInfo)
if (syncer != nil) {
syncer.Sync()
}
```

anyush: Yeah it should - fixed

pkg/neg/syncers/transaction.go
```diff
@@ -1108,6 +1087,26 @@ func (s *transactionSyncer) computeEPSStaleness(endpointSlices []*discovery.Endp
 	}
 }
 
+// getNEGName returns the name of the NEG for the given subnet.
+func (s *transactionSyncer) getNEGName(subnet string) (string, error) {
+	if s.NegSyncerKey.IsBindingKey() {
```
midavadim, line 1094: add comment for each step with exaplnation when it is appied

pkg/neg/controller.go
```diff
@@ -591,49 +661,58 @@ func (c *Controller) processService(key string) error {
...
+		}
 		return nil
 	}
 	service := obj.(*apiv1.Service)
 	if service == nil {
 		return fmt.Errorf("cannot convert to Service (%T)", obj)
 	}
+
+	var bindingErr error
+	if c.enableNEGBinding && c.negBindingManager != nil {
```
midavadim: c.negBindingManager != nil  - is not needed I think

pkg/neg/controller.go
midavadim, line 648: should we stop negbinding related things ?

pkg/neg/controller.go
```diff
@@ -591,49 +661,58 @@ func (c *Controller) processService(key string) error {
...
+		bindingErr = c.negBindingManager.EnsureSyncersForService(namespace, name)
```
midavadim, line 673: so now it could mean that in case bindingErr, service will be added to queue again and processService will be called again for both: usual Syncer and for NegBinding, 

We should double check if it is not heavy operation. 
Alternative is to have separate services queue

pkg/context/context.go
```diff
@@ -218,6 +225,11 @@ func NewControllerContext(
 		context.SAInformer = informerserviceattachment.NewServiceAttachmentInformer(saClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer())
 	}
 
+	if negBindingClient != nil {
+		context.NEGBindingInformer = informernegbinding.NewNetworkEndpointGroupBindingInformer(negBindingClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer())
+		_ = context.NEGBindingInformer.AddIndexers(cache.Indexers{neg.ServiceKeyIndex: neg.ServiceKeyIndexFunc})
```
midavadim, line 230: nice  ... I didn't know about thid functionality

pkg/neg/negbinding_manager.go
```diff
@@ -0,0 +1,541 @@
...
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (m *negBindingManager) ensureSyncerForNEGBinding(
+	binding *negbindingv1beta1.NetworkEndpointGroupBinding,
+	svc *apiv1.Service,
+	networkInfo *network.NetworkInfo,
+) (negtypes.NegSyncer, error) {
+	svcName := binding.Spec.BackendRef.Name
```
midavadim, line 250: svcName := binding.Spec.BackendRef.Name
	svcPort := binding.Spec.BackendRef.Port
could be un safe .,... probably err := m.validateBackendRef(binding) is needed

pkg/neg/negbinding_manager.go
```diff
@@ -0,0 +1,541 @@
...
+		nbNamer,
+		m.negMetrics,
+	)
+
+	if err := syncer.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start syncer for binding %s: %w", bindingKey, err)
+	}
+
+	m.syncerMap[bindingKey] = syncer
+	m.syncerConfigs[bindingKey] = newConfig
```
midavadim, line 365: newConfig contains poiner to *network.NetworkInfo which is not recomended.

consider: "Dereferencing *networkInfo (newConfig := syncerConfig{portTuple: portTuple, networkInfo: *networkInfo}) guarantees thread-safety, immutability, and panic-free equality checks."

pkg/neg/controller.go
```diff
@@ -525,6 +589,9 @@ func (c *Controller) processNode() {
 	}
 
 	c.manager.SyncNodes()
+	if c.enableNEGBinding {
+		c.negBindingManager.SyncAllSyncers()
```
midavadim: lets discuss if we need run it on node chnage

pkg/neg/controller.go
anyush: I guess in read-only we should not do anything, so rather yes. In fact as controller won't ever execute later code in processService (or processNEGBinding) - there should be no syncers at all.

pkg/neg/negbinding_manager.go
anyush: It's already dereferenced.

pkg/neg/syncers/transaction.go
anyush: Added

pkg/neg/controller.go
anyush: Removed

pkg/neg/controller.go
anyush: As checked offline - it's quite lightweight, so leaving as is. Separate queues will make code too messy.

pkg/neg/negbinding_manager.go
anyush: Added validation to EnsureSyncersForService (already was there for EnsureSyncerForNEGBinding) to preserve order of BackendRef condition's errors

pkg/context/context.go
```diff
@@ -154,6 +159,7 @@ func NewControllerContext(
 	frontendConfigClient frontendconfigclient.Interface,
 	firewallClient firewallclient.Interface,
 	svcnegClient svcnegclient.Interface,
+	negBindingClient negbindingclient.Interface,
```
DamianSawicki, line 162: At some point, parts of this long parameter list may be worth refactoring into a struct, but this is not a blocking comment.

DamianSawicki (approved):
Approval for `pkg/controller/controller_test.go`.

/lgtm
/approve

PR #3170, opened 2026-07-17
Refactor L4 standalone NEG controller logic

FelipeYepez:
Apply fixes for error handling

### DIFF
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -142,12 +142,12 @@ func NewStandaloneNEGLBController(ctx *ccontext.ControllerContext, stopCh <-chan
 			if !ok {
 				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
 				if !ok {
-					logger.Error(nil, "unexpected object type in DeleteFunc", "type", fmt.Sprintf("%T", obj))
+					logger.Error(nil, "Unexpected object type in DeleteFunc", "type", fmt.Sprintf("%T", obj))
 					return
 				}
 				svc, ok = tombstone.Obj.(*v1.Service)
 				if !ok {
-					logger.Error(nil, "unexpected object type in tombstone in DeleteFunc", "type", fmt.Sprintf("%T", tombstone.Obj))
+					logger.Error(nil, "Unexpected object type in tombstone in DeleteFunc", "type", fmt.Sprintf("%T", tombstone.Obj))
 					return
 				}
 			}
@@ -257,7 +257,7 @@ func (lc *StandaloneNEGLBController) parseForwardingRuleKeys(frNamesStr string, 
 		if strings.Contains(frName, "/") {
 			resourceID, err := cloud.ParseResourceURL(frName)
 			if err != nil {
-				svcLogger.Error(err, "failed to parse forwarding rule reference URL", "frRef", frName)
+				svcLogger.Error(err, "Failed to parse forwarding rule reference URL", "frRef", frName)
 				errs = append(errs, fmt.Errorf("failed to parse forwarding rule reference URL %s: %w", frName, err))
 				continue
 			}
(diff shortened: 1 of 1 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -321,17 +328,25 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
...
 	// if at least one FR was ok then we use it
 	if len(lbIngresses) == 0 {
 		// if none of the FRs is usable remove any that is possibly there
 		cond := NewConditionExternalIPProgrammedFalse(classifyError(errs[0]))
-		err := updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{Ingress: nil}, []metav1.Condition{cond}, nil, svcLogger)
-		if err != nil {
-			errs = append(errs, err)
+		clearErr := updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{Ingress: nil}, []metav1.Condition{cond}, nil, svcLogger)
+		joinedErr := errors.Join(errs...)
+		var hasSystemError bool
```
mmamczur: lets make a function for this hasSystemError, I see this repeated below

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -275,23 +275,31 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
...
-		}
+		clearErr := updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{Ingress: nil}, []metav1.Condition{cond}, nil, svcLogger)
 		if len(errs) > 0 {
-			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use any Forwarding Rule %s", errors.Join(errs...).Error())
-			return nil, errors.Join(errs...)
+			joinedErr := errors.Join(errs...)
+			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use any Forwarding Rule: %v", joinedErr)
+
+			var hasSystemError bool
+			for _, e := range errs {
```
mmamczur: why do we treat these errors differently? So the only difference is that if there is a non-user error we say "multiple errors occurred" but other than that it's the same list

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -275,23 +275,31 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
...
 		cond := NewConditionExternalIPProgrammedFalse(reason)
-		err = updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{Ingress: nil}, []metav1.Condition{cond}, nil, svcLogger)
-		if err != nil {
-			errs = append(errs, err)
-		}
+		clearErr := updateServiceStatus(lc.ctx, svc, &v1.LoadBalancerStatus{Ingress: nil}, []metav1.Condition{cond}, nil, svcLogger)
 		if len(errs) > 0 {
-			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use any Forwarding Rule %s", errors.Join(errs...).Error())
-			return nil, errors.Join(errs...)
+			joinedErr := errors.Join(errs...)
```
TortillaZHawaii: this will lead to nesting of error messages, when joinedErr is joined again later

FelipeYepez: Done — cleaned this up to just append clearErr with the rest of the slice and errors.Join() them exactly once to prevent nesting

pkg/l4/controllers/standalonenegcontroller.go
FelipeYepez: Done, eliminated the hasSystemError logic

pkg/l4/controllers/standalonenegcontroller.go
FelipeYepez: Removed

PR #3175, opened 2026-07-21
Canonicalize IPv6 addresses in L4StandaloneNEG to prevent k8s apiserv…

TortillaZHawaii:
…er warnings

It's not released yet so it's a trivial change. For NetLB/ILB we also should do this, but at this point I'm afraid it might break something. The warning comes from https://github.com/kubernetes/kubernetes/blob/74c594bf1a429c0b6cdc33e48a32596c978dfa7b/staging/src/k8s.io/apimachinery/pkg/util/validation/ip.go#L126


FelipeYepez:
Thanks!

PR #3176, opened 2026-07-21
Fix omitted fields in XPN gcloud firewall command generation

marzdrel:
## What this PR does / why we need it
In Google Cloud Shared VPC (XPN) clusters, the `ingress-gce` controller service account in the service project lacks IAM permissions to directly modify GCP firewall rules in the host project.

When firewall API calls fail with `403 Forbidden`, `ingress-gce` catches `IsForbiddenError` and raises a Kubernetes `Event` on the Ingress/Service resource containing a generated `gcloud` CLI command. Host network administrators execute these generated `gcloud` commands in the host project to grant the required network access.

Previously, `firewallToGcloudArgs` in `vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_util.go` omitted several fields from the internal `compute.Firewall` struct:
- **`DestinationRanges`**: Pinhole firewall rules (`EnablePinhole=true`) omitted `--destination-ranges`, causing admin-applied rules to be overly permissive exposing node IPs.
- **`Denied`**: L4 NetLB Deny rules formatted `--allow ""` instead of `--deny`.
- **`Priority`**: Non-default rule priorities fell back to GCP default (`1000`).
- **`Direction`**: Explicit directions (e.g. `EGRESS`) were omitted.
- **`Disabled`**: Rule disabled state was omitted.

This PR refactors `firewallToGcloudArgs` to conditionally append `--destination-ranges`, `--deny`, `--priority`, `--direction`, and `--disabled` flags when set on the `compute.Firewall` object.

Additionally, to prevent future regression bugs when new fields are added to `compute.Firewall` in the GCP SDK, this PR introduces `TestFirewallToGCloud_FieldExhaustiveness`. Using Go reflection (`reflect.TypeOf`), the test actively fails the build if an unmapped field is added to `compute.Firewall`.


pkg/firewalls/firewalls_l4_test.go
```diff
@@ -235,3 +237,298 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, tag string) {
...
+			desc: "pinhole rule with destination ranges",
+			fw: &compute.Firewall{
+				Name:              "k8s-fw-pinhole",
+				Network:           "projects/test-project/global/networks/default",
+				Description:       "pinhole rule",
+				SourceRanges:      []string{"10.0.0.0/8"},
+				DestinationRanges: []string{"192.168.1.2/32", "192.168.1.1/32"},
+				TargetTags:        []string{"node-tag"},
+				Allowed: []*compute.FirewallAllowed{
+					{IPProtocol: "tcp", Ports: []string{"80"}},
```
TortillaZHawaii, line 257: can we also test it:
* with both tcp and udp ports being present?
* with multiple ports defined for each entry?

pkg/firewalls/firewalls_l4_test.go
```diff
@@ -235,3 +237,298 @@ func createVMInstanceWithTag(t *testing.T, fakeGCE *gce.Cloud, tag string) {
...
+		{
+			desc: "deny rule with priority, direction, and disabled",
+			fw: &compute.Firewall{
+				Name:         "k8s-fw-deny",
+				Network:      "projects/test-project/global/networks/default",
+				Description:  "deny rule",
+				SourceRanges: []string{"10.0.0.0/8"},
+				TargetTags:   []string{"node-tag"},
+				Denied: []*compute.FirewallDenied{
+					{IPProtocol: "tcp", Ports: []string{"8080"}},
```
TortillaZHawaii: Deny rules are exclusively created with protocol "all" and without ports. Worth checking if it formats it corrrectly. https://github.com/kubernetes/ingress-gce/blob/ccbd96ae5d95921f210692cda9932e7a5b9c8ed7/pkg/firewalls/allowed.go#L36-L40

vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_util.go
```diff
@@ -177,21 +177,62 @@ func FirewallToGCloudDeleteCmd(fwName, projectID string) string {
 }
 
 func firewallToGcloudArgs(fw *compute.Firewall, projectID string) string {
```
TortillaZHawaii, line 200: This is vendored from another package, but I see no changes to go.mod. This probably needs upstreaming to https://github.com/kubernetes/cloud-provider-gcp instead of updating it here. Future updates to packages will overwrite those changes.

TortillaZHawaii (changes requested):
/hold

vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_util.go
TortillaZHawaii: Alternative is to have separate implementation in both places and stop importing it.

pkg/firewalls/firewalls_l4_test.go
marzdrel: Good point, I've updated formatted plus added some more test cases. 👍

pkg/firewalls/firewalls_l4_test.go
marzdrel: Added. ✅

marzdrel:
Submitted upstream PR: https://github.com/kubernetes/cloud-provider-gcp/pull/1265

vendor/k8s.io/cloud-provider-gcp/providers/gce/gce_util.go
marzdrel: I've added https://github.com/kubernetes/cloud-provider-gcp/pull/1265 and I'll update this repo once the upstream PR gets merged and released. I'll also reduced amount of tests here, as it doesn't make much sense to test vendored dependency in such detail (especially since those tests were moved to the source).

PR #3177, opened 2026-07-21
NEGBinding - NEGs ownership

anyush:
Forcing controller to manage single NEG by at most single NEGBinding-based syncer based on NEG name.

If multiple NEGBinding CRs will specify the same NEG name (subnets/zones are not verified there), then only one will have the ability to attach/detach endpoints from all NEGs with this name.

### DIFF
--- a/pkg/neg/negbinding_manager.go
+++ b/pkg/neg/negbinding_manager.go
@@ -308,11 +308,13 @@ func (m *negBindingManager) EnsureSyncersForService(svcNamespace, svcName string
 			errs = append(errs, fmt.Errorf("unexpected object type %T in binding index for service %s", obj, svcKey))
 			continue
 		}
+		binding = binding.DeepCopy()
 
 		err := m.validateBackendRef(binding)
 		if err != nil {
 			_ = m.updateBackendRefCondition(binding, err)
-			return err
+			errs = append(errs, err)
+			continue
 		}
 
 		if networkInfoErr != nil {
@@ -635,10 +637,10 @@ func (m *negBindingManager) updateBackendRefCondition(binding *negbindingv1beta1
 		}
 	}
 
-	origBinding := binding.DeepCopy()
-	m.ensureCondition(binding, cond)
+	newBinding := binding.DeepCopy()
+	m.ensureCondition(newBinding, cond)
 
-	patchBytes, err := patch.MergePatchBytes(negbindingv1beta1.NetworkEndpointGroupBinding{Status: origBinding.Status}, negbindingv1beta1.NetworkEndpointGroupBinding{Status: binding.Status})
+	patchBytes, err := patch.MergePatchBytes(negbindingv1beta1.NetworkEndpointGroupBinding{Status: binding.Status}, negbindingv1beta1.NetworkEndpointGroupBinding{Status: newBinding.Status})
 	if err != nil {
 		return fmt.Errorf("failed to prepare patch bytes for status update: %w", err)
 	}
--- a/pkg/neg/negbinding_manager.go
+++ b/pkg/neg/negbinding_manager.go
@@ -408,6 +408,7 @@ func (m *negBindingManager) ensureSyncerForNEGBinding(
 		m.negBindingClient,
 		m.negBindingLister,
 		m.negMetrics,
+		m.ownershipRegistry,
 		m.logger,
 	)
 
--- a/pkg/neg/negbinding_manager.go
+++ b/pkg/neg/negbinding_manager.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	negbindingv1beta1 "k8s.io/ingress-gce/pkg/apis/negbinding/v1beta1"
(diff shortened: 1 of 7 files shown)


pkg/neg/syncers/negstatushandler/negbinding.go
```diff
@@ -256,3 +269,34 @@ func (h *NEGBindingStatusHandler) patchNegBindingStatus(oldStatus, newStatus neg
...
+	if len(conflicts) == 0 {
+		return negbindingv1beta1.Condition{
+			Type:               ManagedCondition,
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.Now(),
+		}
+	}
+
+	message := conflicts[0]
+	for i := 1; i < len(conflicts); i++ {
```
midavadim: := strings.Join(conflicts, "; ")

pkg/neg/negbinding_manager.go
```diff
@@ -0,0 +1,673 @@
...
+func (c syncerConfig) Equals(other syncerConfig) bool {
+	return c.portTuple.Port == other.portTuple.Port &&
+		c.portTuple.Name == other.portTuple.Name &&
+		c.portTuple.TargetPort == other.portTuple.TargetPort &&
+		c.networkInfo.IsDefault == other.networkInfo.IsDefault &&
+		c.networkInfo.NetworkURL == other.networkInfo.NetworkURL &&
+		c.networkInfo.SubnetworkURL == other.networkInfo.SubnetworkURL &&
+		c.networkInfo.K8sNetwork == other.networkInfo.K8sNetwork
+}
+
```
midavadim, line 88: // negOwnershipRegistry tracks exclusive ownership of NEG names across NEGBinding CR syncers 
//
// Ownership Lifecycle & Conflict Resolution:
//
// 1. First-Come, First-Served Acquisition:
//    - When a NEGBinding CR is processed, its syncer calls Acquire(negName, ownerKey).
//    - The first CR to request an unowned NEG gains exclusive ownership in the registry.
//
// 2. Handling Conflicting CRs:
//    - If a second NEGBinding CR requests the same NEG name, Acquire() fails and returns
//      (false, currentOwnerKey).
//    - The conflicting CR skips endpoint/subnet management for that NEG.
//    - The status handler queries GetOwner(negName) and reports a status condition:
//      Managed = False (Reason: Conflict) with a message identifying the owning CR.
//
// 3. Automatic Failover on Release:
//    - When an owning CR is deleted or updated (removing a NEG from its spec),
//      ReleaseAllOwnedExcept() removes ownership and fires the onRelease(negName) callback.
//    - The callback triggers tryAssignNEGToBinding(negName), re-syncing any waiting CR
//      so it can attempt to acquire the newly freed NEG.

```
Binding A           negOwnershipRegistry             Binding B          Status Reporter (Binding B)
    |                         |                          |                            |
    |-- Acquire(neg-shared) ->|                          |                            |
    |   (owner: ns/binding-a) |                          |                            |
    |<-- (true, "") ----------|                          |                            |
    |                         |                          |                            |
    |                         |<-- Acquire(neg-shared) --|                            |
    |                         |    (owner: ns/binding-b) |                            |
    |                         |--- (false, "binding-a") ->|                            |
    |                         |                          |                            |
    |                         |<-------------------------|----------------------------|
    |                         |                          |  GetOwner("neg-shared")    |
    |                         |--------------------------+--------------------------->|
    |                         |                          |  Returns: "ns/binding-a"   |
    |                         |                          |  Condition: Conflict       |
    |                         |                          |                            |
    |-- Delete/Release ------>|                          |                            |
    |                         |-- delete(owners, neg)    |                            |
    |                         |-- go onRelease(neg)      |                            |
    |                         |       |                  |                            |
    |                         |       +----------------->|                            |
    |                         |   Trigger Re-sync        |                            |
    |                         |                          |                            |
    |                         |<-- Acquire(neg-shared) --|                            |
    |                         |--- (true, "") ---------->| (Acquisition Successful)   |
    v                         v                          v                            v
```

pkg/neg/negbinding_manager.go
```diff
@@ -0,0 +1,673 @@
...
+				Name:       sp.Name,
+				TargetPort: sp.TargetPort.String(),
+			}
+			return portTuple, nil
+		}
+	}
+	return negtypes.SvcPortTuple{}, fmt.Errorf("port %d not found in service %s/%s spec", port, svc.Namespace, svc.Name)
+}
+
+// InitializeOwnershipRegistry reads NEG names from existing NEGBinding CRs' statuses and acquires them.
```
midavadim, line 576: I have a question.

What if there 2 Bindings. One was succesfull and another in 'failed' state.

InitializeOwnershipRegistry runs once after controller restart. So it is possbile that Second Binding could acquire NEGname first.

Is it possible/worth to check NEG Binding contditions to avoid this situtation ?

midavadim: or will ninding.status.NetworkEndpointGroups  contain only 'correctly  owned NEGs' ?

anyush: Status contains only NEGs, which were actively managed, which (after this PR is merged) requires previous ownership. So in the situation you're describing second binding will never have conflicting name in it's status, therefore InitializeOwnershipRegistry won't try to acquire ownership for it

pkg/neg/negbinding_manager.go
anyush: Yeah, that's how it works - do you need any response from me about it or any changes?

pkg/neg/syncers/negstatushandler/negbinding.go
anyush: Applied

pkg/neg/negbinding_manager.go
```diff
@@ -546,3 +648,26 @@ func (m *negBindingManager) ensureCondition(binding *negbindingv1beta1.NetworkEn
...
+		if !ok {
+			continue
+		}
+
+		for _, ref := range binding.Spec.NetworkEndpointGroups {
+			if ref.Name == negName {
+				bindingKey := fmt.Sprintf("%s/%s", binding.Namespace, binding.Name)
+				// It's not guaranteed that this binding will have ownership if conflict with other binding still exists
+				m.logger.Info("Triggering ensure/sync for binding which refers to released NEG", "binding", bindingKey, "negName", negName)
+				if err := m.EnsureSyncerForNEGBinding(binding); err != nil {
```
08volt: Should deep copy `binding` before passing it to avoid race

pkg/neg/syncers/negstatushandler/negbinding.go
```diff
@@ -256,3 +270,30 @@ func (h *NEGBindingStatusHandler) patchNegBindingStatus(oldStatus, newStatus neg
...
+		if owner != "" && owner != ownerKey {
+			conflicts = append(conflicts, fmt.Sprintf("NEG %q is owned by %q", ref.Name, owner))
+		}
+	}
+	if len(conflicts) == 0 {
+		return negbindingv1beta1.Condition{
+			Type:               ManagedCondition,
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.Now(),
+		}
```
08volt, line 291: I think Condition Reason and Message are required when creating a Condition

pkg/neg/syncers/negbinding_topology.go
```diff
@@ -71,6 +78,28 @@ func (p *NEGBindingTopologyProvider) getBinding() (*negbindingv1beta1.NetworkEnd
...
+			acquiredRefs = append(acquiredRefs, ref)
+		} else {
+			logger.Info("NEG name is owned by another binding, skipping", "negName", ref.Name, "owner", ownerKey, "currentOwner", owner)
+		}
+	}
+	return acquiredRefs
+}
+
 // ListSubnetsInDefaultNetwork returns the list of subnets declared inside the NegBinding CR Spec.
 func (p *NEGBindingTopologyProvider) ListSubnetsInDefaultNetwork(logger klog.Logger) []nodetopologyv1.SubnetConfig {
```
08volt, line 97: Both `ListSubnetsInDefaultNetwork` and `ListZonesPerSubnet` call `acquireSpecNEGRefs` which spawn gorutines from the manager. 

Each go rutines does full scan and full ensure of another binding if im not mistaken, and this happens twice every sync, once for subnets and once for zones, doing the same ownership work.

These methods should be read-only but instead they actually work on the neg ownerahip.

I think a solution would be to use the queue `negBindingQueue` and enqueue keys instead of spawn with `go r.onRelease`.

Suggested fix:
1. Inject an enqueue function for the bindingKey into the manager 
2. in ReleaseAllOwnedExcept collect the released NEG names and return them ( registry becomes just a data structure, without callbacks or gorutines)
3. the manager releases released names -> referencing binding keys -> calls the enqueue function for each of them
4. the processNegBinding ensure on the serialized worker

pkg/neg/syncers/negbinding_topology.go
```diff
@@ -79,17 +108,24 @@ func (p *NEGBindingTopologyProvider) ListSubnetsInDefaultNetwork(logger klog.Log
 		return nil
 	}
 
-	configs := make([]nodetopologyv1.SubnetConfig, len(binding.Spec.NetworkEndpointGroups))
-	for i, ref := range binding.Spec.NetworkEndpointGroups {
+	// Return only subnets where NEGs are owned
+	subnets := sets.New[string]()
+	ownedSpecRefs := p.acquireSpecNEGRefs(binding, logger)
```
08volt: move the aquireSpecNegRefs out of the "getters" list functions

pkg/neg/syncers/negbinding_topology.go
```diff
@@ -107,8 +143,10 @@ func (p *NEGBindingTopologyProvider) ListZonesPerSubnet(_ zonegetter.Filter, net
 		return nil, fmt.Errorf("failed to get NegBinding from store: %w", err)
 	}
 
+	// Return only zones of subnets, where NEGs are owned
+	ownedSpecRefs := p.acquireSpecNEGRefs(binding, logger)
```
08volt: move the aquireSpecNegRefs out of the "getters" list functions

08volt: You could add a `ownedRefs` field to `NEGBindingTopologyProvider` and a `RefereshOwnership` function that runs the single Release+Acquire

pkg/neg/negbinding_manager.go
08volt: here you could have `newBinding := ...` and use the newBinding in the following function calls

08volt: one bad binding in the for loop aborts all the bindings for the service, we could continue and append to errs

pkg/neg/negbinding_manager.go
anyush: Done

pkg/neg/syncers/negstatushandler/negbinding.go
anyush: Not required (will be "" otherwise), but added

pkg/neg/syncers/negbinding_topology.go
anyush: `ListSubnetsInDefaultNetwork ` and `ListZonesPerSubnet ` calling `GetOwner` instead of `Acquire` in registry from now on

pkg/neg/syncers/negbinding_topology.go
anyush: Used `GetOwner` instead

pkg/neg/syncers/negbinding_topology.go
anyush: NEGs acquired and released by manager

pkg/neg/syncers/negbinding_topology.go
anyush: Calling `GetOwner` instead of acquiring NEGs

pkg/neg/negbinding_manager.go
anyush: Done

anyush: Fixed

PR #3179, opened 2026-07-22
Fix neg cpu spike

iromanycheva-eng:
Fix a severe 100% CPU lock and goroutine leak issue that occurs when ProviderConfig tenants are deleted in multi-tenant mode.

**Root Cause**
When an active tenant was deleted, uncontrolled CPU starvation caused the Kubernetes Control Plane (KCP) to rapidly and continuously vertically scale (resize) the cluster's master VMs up to enormous sizes. This was triggered by the goroutine Leak in NEG worker loops. The worker polling loops in pkg/neg/controller.go (serviceWorker, endpointWorker, etc.) were wrapped in anonymous functions inside an infinite for loop. When the workqueue was shut down on tenant deletion, the queue's Get() unblocked sending a quit=true flag. The return statement successfully exited the anonymous function, but instantly dropped execution back to the top of the infinite loop—spinning wildly on the shutdown queue without blocking.

**Fixes**
Refactored all NEG worker loops to use canonical Kubernetes processNextWorkItem() bool handler methods. When the workqueue is shut down, false is returned, natively breaking the outer infinite for loop and terminating the goroutine safely.

### DIFF
--- a/pkg/neg/controller.go
+++ b/pkg/neg/controller.go
@@ -486,29 +486,33 @@ func (c *Controller) stop() {
 }
 
 func (c *Controller) endpointWorker() {
-	for {
-		func() {
-			key, quit := c.endpointQueue.Get()
-			if quit {
-				return
-			}
-			c.processEndpoint(key.(string))
-			c.endpointQueue.Done(key)
-		}()
+	for c.processNextEndpointWorkItem() {
 	}
 }
 
-func (c *Controller) nodeWorker() {
-	for {
-		func() {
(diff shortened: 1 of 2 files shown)


midavadim:
probably serviceAttachmentWorker has the same problem

midavadim:
pkg/experimental/workload/controller.go 

bas the same problem ... but it seams like not used anywhere

swetharepakula (commented):
Great find Yulia!

/lgtm
/approve

PR #3181, opened 2026-07-22
Fix L4 Address Manager IP validation and regex matching bugs

FelipeYepez:
* Change unanchored 'regexp.MatchString' to 'net.ParseIP().Equal()', preventing false positive substring matches (e.g., 10.0.0.1 matching 10.0.0.123).
* Enforce explicit Network Tier and LoadBalancingScheme validation inside 'IsAddressInForwardingRules' to prevent incorrectly paired internal/external configurations.
* Remove brittle manual IPv6 string manipulation ('DecompressAddr') by leveraging native 'net.ParseIP' functionality.
* Expand 'TestIsAddressInForwardingRules' to cover new negative validation paths and re-incorporate legacy IPv6 compression matrix test cases.

### DIFF
--- a/pkg/l4/address/manager.go
+++ b/pkg/l4/address/manager.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"regexp"
 	"strings"
 
 	compute "google.golang.org/api/compute/v1"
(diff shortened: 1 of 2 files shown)


pkg/l4/address/manager.go
```diff
@@ -188,34 +187,6 @@ func (m *Manager) removeAddress(name, reason string) error {
 // DecompressAddr re-adds ommited hextets in IPv6 address.
 // For example, running the function with argument "2001:db8::ff00:42:8329"
 // returns "2001:db8:0:0:0:ff00:42:8329".
```
JumiDeluxe: Remove the function documentation too

JumiDeluxe (changes requested):

pkg/l4/address/manager.go
FelipeYepez: Removed and added a comment for `IsAddressInForwardingRules`, thanks

JumiDeluxe (approved):

PR #3187, opened 2026-07-28
NEGBinding - NEG Metrics

anyush:
Report NEGs bound by NEGBinding CRs in NEG metrics.

NEGs normally (not binding-related) count as set across zones, which in fact means number of unique NEG names, not number of NEGs. According to this when reporting binding-related NEGs, also reporting number of unique NEG names + towards which service count NEG is decided based on the ownership of NEG names by bindings (https://github.com/kubernetes/ingress-gce/pull/3177).

Depends on https://github.com/kubernetes/ingress-gce/pull/3185 and should not be merged before it. Contains commits from it - please review only ones, which are not in the dependency.

### DIFF
--- a/pkg/neg/metrics/metricscollector/metrics_collector.go
+++ b/pkg/neg/metrics/metricscollector/metrics_collector.go
@@ -448,6 +448,11 @@ func (sm *SyncerMetrics) SetNegService(svcKey string, negState NegServiceState) 
 	if sm.negMap == nil {
 		klog.Fatalf("Ingress Metrics failed to initialize correctly.")
 	}
+	if existing, ok := sm.negMap[svcKey]; ok {
+		negState.NegBindingNeg = existing.NegBindingNeg
+		negState.BindingSuccessfulNeg = existing.BindingSuccessfulNeg
+		negState.BindingErrorNeg = existing.BindingErrorNeg
+	}
 	sm.negMap[svcKey] = negState
 }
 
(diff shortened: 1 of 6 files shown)


pkg/neg/negbinding_manager.go
```diff
@@ -253,14 +263,32 @@ func newNEGBindingManager(
...
+	)
 	return m
 }
 
 // EnsureSyncerForNEGBinding ensures corresponding syncer is started for the binding.
-func (m *negBindingManager) EnsureSyncerForNEGBinding(binding *negbindingv1beta1.NetworkEndpointGroupBinding) error {
+func (m *negBindingManager) EnsureSyncerForNEGBinding(binding *negbindingv1beta1.NetworkEndpointGroupBinding) (err error) {
+	if binding != nil {
+		defer func() {
+			svcKey := fmt.Sprintf("%s/%s", binding.Namespace, binding.Spec.BackendRef.Name)
```
08volt: nil pointer panic when BackendRef is nil

pkg/neg/negbinding_manager.go
```diff
@@ -253,14 +263,32 @@ func newNEGBindingManager(
...
 		logger:              logger.WithName("NEGBindingManager"),
 	}
-	m.ownershipRegistry = newNEGOwnershipRegistry(m.tryAssignReleasedNEGs)
+	m.ownershipRegistry = newNEGOwnershipRegistry(
+		m.tryAssignReleasedNEGs,
+		func(ownerKey string) {
+			obj, exists, err := m.negBindingLister.GetByKey(ownerKey)
+			if err == nil && exists {
+				if binding, ok := obj.(*negbindingv1beta1.NetworkEndpointGroupBinding); ok {
+					svcKey := fmt.Sprintf("%s/%s", binding.Namespace, binding.Spec.BackendRef.Name)
```
08volt: nil pointer panic when BackendRef is nil

pkg/neg/negbinding_manager.go
```diff
@@ -119,6 +121,9 @@ func (r *negOwnershipRegistry) Acquire(negName string, owner string) (bool, stri
 	currentOwner, ok := r.owners[negName]
 	if !ok {
 		r.setOwnerLocked(negName, owner)
+		if r.onChange != nil {
+			go r.onChange(owner)
```
08volt: Are you sure we need this "onChange" ? we can return the affected owners and recompute the metrics on the worker after the locks are released, sequentially without go routines.

pkg/neg/negbinding_manager.go
```diff
@@ -144,8 +149,13 @@ func (r *negOwnershipRegistry) ReleaseAllOwnedExcept(owner string, keep sets.Set
 		}
 	}
 
-	if len(released) > 0 && r.onRelease != nil {
-		go r.onRelease(released)
+	if len(released) > 0 {
+		if r.onRelease != nil {
+			go r.onRelease(released)
```
08volt: Same here, we can return `released` and re-enqueue the affected bindings into `negBindingQueue` so failover runs on serialized workers

pkg/neg/negbinding_manager.go
anyush: BackendRef is forced to not being nil on the kubernetes side - if it's not set CR won't even be created

pkg/neg/metrics/metricscollector/metrics_collector.go
```diff
@@ -456,7 +461,69 @@ func (sm *SyncerMetrics) DeleteNegService(svcKey string) {
...
+// DeleteNegBindingService resets NegBindingNeg for a service key and removes the service if no other NEGs exist.
+func (sm *SyncerMetrics) DeleteNegBindingService(svcKey string) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+
+	existing, ok := sm.negMap[svcKey]
+	if !ok {
+		return
+	}
+	if existing.SuccessfulNeg == 0 && existing.ErrorNeg == 0 {
```
08volt: Shouldn't we check a Neg exists here? 
Such as:
```
existing.StandaloneNeg != 0 || existing.IngressNeg != 0 || existing.VmIpNeg != nil
```

anyush: AI continues to suggest this as well, but if there're any NEGs (ingress/standalone/vm_ip) - shouldn't all of these be either successful or error?

08volt: I think its easier to understand and more "robust" to use the other fields, but its fine either way

anyush: Ok, will add these to make it more clear (adding more checks with OR won't prevent check to pass either way)

anyush: Done

pkg/neg/negbinding_manager.go
anyush: Reworked, so instead of using callback just reporting metrics from manager if NEGs acquired/released

PR #3188, opened 2026-07-29
Implement allowMissing in FilteredInformer and enable it for Node informer

priyapande:
Fixing gap between gke-enterprise-mt implementation for filtered informers for Node for supervisor not having labels.


swetharepakula (commented):
/lgtm
/approve

PR #3189, opened 2026-08-04
Prevent RESOURCE_IN_USE reconciliation loop on ip-collection-v6 updates

FelipeYepez:
When updating networking.gke.io/ip-collection-v6 on an existing Service, GCE rejects deleting the Address if an existing ForwardingRule still references it, causing an infinite RESOURCE_IN_USE loop. This change:
- Deletes the existing IPv6 ForwardingRule before calling HoldAddress() when ip-collection-v6 changes.
- Resets the address in IPv6ToUse when the IP collection changes on an ephemeral BYOIPv6 Service.
- Adds positive and negative unit tests verifying collection updates, removals, subnetwork independence, and loop prevention.

### DIFF
--- a/pkg/l4/address/to_use.go
+++ b/pkg/l4/address/to_use.go
@@ -6,6 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/ingress-gce/pkg/composite"
+	"k8s.io/ingress-gce/pkg/flags"
 	"k8s.io/ingress-gce/pkg/l4/annotations"
 	"k8s.io/klog/v2"
 
@@ -70,6 +71,10 @@ func IPv6ToUse(cloud *gce.Cloud, svc *v1.Service, ipv6FwdRule *composite.Forward
 	}
 	if requestedSubnet != ipv6FwdRule.Subnetwork {
 		logger.V(2).Info("ipv6AddressToUse: reset IPv6 Address due to changed subnet")
+		return "", "", nil
+	}
+	if flags.F.EnableBYOIPv6 && annotations.FromService(svc).GetIPCollectionV6() != ipv6FwdRule.IpCollection {
+		logger.V(2).Info("ipv6AddressToUse: reset IPv6 Address due to changed IP collection")
 		return "", "", nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/l4/address/to_use.go
```diff
@@ -72,6 +73,10 @@ func IPv6ToUse(cloud *gce.Cloud, svc *v1.Service, ipv6FwdRule *composite.Forward
 		logger.V(2).Info("ipv6AddressToUse: reset IPv6 Address due to changed subnet")
 		return "", "", nil
 	}
+	if flags.F.EnableBYOIPv6 && annotations.FromService(svc).GetIPCollectionV6() != ipv6FwdRule.IpCollection {
```
08volt, line 76: this is not only for netlbs, in case on ilb with annotation the "ipv6FwdRule.IpCollection" is never set and will reset the address at every sync

pkg/l4/resources/l4netlb_test.go
08volt, line 1: New tests pass even without the fix

pkg/l4/resources/forwarding_rules_ipv6.go
```diff
@@ -239,6 +234,16 @@ func (l4netlb *L4NetLB) ensureIPv6ForwardingRule(bsLink string) (*composite.Forw
 			}
 		}
 
+		if flags.F.EnableBYOIPv6 && existingIPv6FwdRule != nil && existingIPv6FwdRule.IpCollection != ipCollection {
+			frLogger.V(2).Info("deleting forwarding rule for service due to ip collection mismatch", "existingIpCollection", existingIPv6FwdRule.IpCollection, "expectedIpCollection", ipCollection)
+			if err := l4netlb.forwardingRules.Delete(existingIPv6FwdRule.Name); err != nil {
+				frLogger.Error(err, "l4netlb.forwardingRules.Delete returned error, want nil")
+				return nil, l4utils.ResourceResync, err
```
08volt, line 241: line 257 is very similar to this one and uses "l4utils.ResourceUpdate" , not sure whats the right one but should be consistent

FelipeYepez:
This PR is no longer needed, thanks for the review @08volt

PR #3194, opened 2026-08-06
Add gke-enterprise-mt dependency for syncing metrics & update local framework copy with DeleteFunc to handle delete events for ProviderConfig

priyapande:


pkg/multiproject/framework/controller_test.go
```diff
@@ -381,3 +398,48 @@ func TestPanicRecovery(t *testing.T) {
...
+	err = tc.pcInformer.GetIndexer().Delete(pc)
+	if err != nil {
+		t.Fatalf("failed to delete ProviderConfig from indexer: %v", err)
+	}
+	// Trigger Delete handler/queue
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(pc)
+	if err == nil {
+		tc.pcController.providerConfigQueue.Enqueue(key)
+	}
+
```
swetharepakula: shouldn't the controller do this automatically?

pkg/multiproject/framework/manager.go
```diff
@@ -25,6 +33,9 @@ type manager struct {
 	providerConfigClient providerconfigclient.Interface
 	finalizerName        string
 	controllerStarter    ControllerStarter
+
+	deletingMu sync.Mutex
+	deleting   map[string]*deletingTenantState
```
swetharepakula: add a comment to what these are for

iromanycheva-eng:
lgtm

PR #3201, opened 2026-08-18
Standalone L4 NEG - validate NEG attachment 

mmamczur:
before writing LB VIP into service status, check if the service NEGs are attached to the LB.
LB VIPs for LBs that do not use the controller created NEGs will not be programmed in the cluster.

### DIFF
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -31,8 +31,10 @@ import (
 	"golang.org/x/exp/slices"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
+	negv1beta1 "k8s.io/ingress-gce/pkg/apis/svcneg/v1beta1"
 	"k8s.io/ingress-gce/pkg/composite"
 	ccontext "k8s.io/ingress-gce/pkg/context"
 	"k8s.io/ingress-gce/pkg/l4/annotations"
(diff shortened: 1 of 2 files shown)


mmamczur:
> /assign @JumiDeluxe

looks like this won't work until you have k8s membership - just look through and comment if needed

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -286,6 +291,114 @@ func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
...
+		return l4utils.NewUserError(fmt.Errorf("failed to parse backend service URL %s: %w", bsURL, err))
+	}
+	if resourceID == nil || resourceID.Key == nil || resourceID.Key.Name == "" {
+		return l4utils.NewUserError(fmt.Errorf("invalid backend service URL %s: missing resource key", bsURL))
+	}
+
+	bs, err := composite.GetBackendService(lc.ctx.Cloud, resourceID.Key, meta.VersionBeta, svcLogger)
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			return l4utils.NewUserError(fmt.Errorf("failed to get backend service %s: %w", bsURL, err))
```
JumiDeluxe: we can construct this error msg before the `if`, as it is the same for both errors

pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -286,6 +291,114 @@ func validateForwardingRule(fr *composite.ForwardingRule, frName string) error {
...
+		}
+		groupKey, err := cloud.ParseResourceURL(be.Group)
+
+		if err == nil && groupKey != nil && targetNEGs.Has(groupKey.MapKey()) {
+			hasMatchingBackend = true
+			break
+		}
+	}
+
+	if !hasMatchingBackend {
```
JumiDeluxe: Can we move it to a separate function? As this is simple code but due to size it reduces readability

pkg/l4/controllers/standalonenegcontroller.go

JumiDeluxe:
lgtm

PR #3218, opened 2026-08-21
Fix user error classification and IP sanitization in Standalone NEG LB controller

FelipeYepez:
Filter nil errors in joinMaybeUserErrors to ensure user-caused errors are properly attributed in metrics. Sanitize forwarding rule IP addresses to prevent empty string entries from being written to service status ingress.

### DIFF
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -407,11 +407,13 @@ func backendServiceHasNEGAttached(bs *composite.BackendService, targetNEGs sets.
 
 // IPAddress field is used for creating Regional NetLB forwarding rules, IPAddresses[] field is used for creating Global NetLB forwarding rules.
 func frAddresses(fr *composite.ForwardingRule) []string {
-	ipAddrs := fr.IPAddresses
-	if len(ipAddrs) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -548,17 +557,28 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 // join errors but if all are UserErrors make the wrapping error a UserError as
 // well. This is required to properly attribute sync status in metrics.
 func joinMaybeUserErrors(errs ...error) error {
-	allUserErrors := true
+	var nonNilErrs []error
 	for _, err := range errs {
+		if err != nil {
```
mmamczur, line 562: is there a chance that we have a nil error?

FelipeYepez: Yes, clearErr is prepended to the error slice (allErrs := append([]error{clearErr}, errs...)).

When updateServiceStatus succeeds, clearErr is nil. Without filtering nils, resources.IsUserError(nil) evaluates to false, which would incorrectly flip allUserErrors = false and cause user errors to be misclassified as provider/system errors in metrics.

PR #3219, opened 2026-08-21
Improve event messaging and rule sorting in Standalone NEG LB

FelipeYepez:
Differentiate total vs partial forwarding rule failure messages, use ForwardingRulesLimitExceeded when skipping rules over the limit, clarify success event text, and ensure deterministic sorting of parsed rules.

### DIFF
--- a/pkg/l4/controllers/standalonenegcontroller.go
+++ b/pkg/l4/controllers/standalonenegcontroller.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud"
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	"golang.org/x/exp/slices"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -44,6 +43,7 @@ import (
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
 	"k8s.io/klog/v2"
+	"slices"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/l4/controllers/standalonenegcontroller.go
```diff
@@ -516,7 +516,11 @@ func (lc *StandaloneNEGLBController) syncStandaloneNEGLB(svc *v1.Service, svcLog
 	}
 
 	if len(errs) > 0 {
-		lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use all Forwarding Rules: %v", errors.Join(errs...))
+		if len(lbIngresses) == 0 {
+			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use any Forwarding Rule: %v", errors.Join(errs...))
+		} else {
+			lc.ctx.Recorder(svc.Namespace).Eventf(svc, v1.EventTypeWarning, "ForwardingRuleUnusable", "Could not use all Forwarding Rules: %v", errors.Join(errs...))
```
mmamczur: I wonder if this could be understood as unable to use all as in all were bad. But I'm not sure how to word it well in English so it's not confusing

pkg/l4/controllers/standalonenegcontroller_test.go
```diff
@@ -2102,14 +2103,15 @@ func TestStandaloneNEGLBControllerEventHandlers_Add(t *testing.T) {
...
 
+	lbClass := annotations.StandalonePassthroughNegLoadBalancerClass
 	svc := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "svc1",
 			Namespace: "default",
 		},
 		Spec: v1.ServiceSpec{
 			Type:              v1.ServiceTypeLoadBalancer,
-			LoadBalancerClass: new(annotations.StandalonePassthroughNegLoadBalancerClass),
```
mmamczur, line 2112: what is wrong with using new() here?

pkg/l4/controllers/standalonenegcontroller.go
JumiDeluxe: At the first glance it is even difficult to notice that these messages are different. Maybe for the second one use something like `"Some Forwarding Rules were unusable: %v"`?

pkg/l4/controllers/standalonenegcontroller_test.go
FelipeYepez: In Go, new(T) takes a type rather than a value, so passing the const string value annotations.StandalonePassthroughNegLoadBalancerClass causes a compilation error (is not a type). Using &lbClass (or ptr.To(...)) allows us to point directly to the string value. I saw these errors highlighted in VSCode.

pkg/l4/controllers/standalonenegcontroller.go
FelipeYepez: Thanks fixed the wording

PR #3234, opened 2026-08-27
Add flags to override L4 LB HealthCheck source CIDRs

midavadim:
This introduces new flags

--override-l4-ilb-health-check-src-cidrs  - for L4 ILB contains both IPv4/v6
--override-l4-netlb-health-check-src-cidrs -  for L4 NetLB contains both IPv4/v6
to support custom health check source ranges for Trusted Partner Cloud environments.

if flag is missing or needed IP ranges are missing, then default hardcoded IP ranges are used.

Controllerbehaviour for different L4 LB type (ILB/NetLB):
- if flag is specified: only LB_TYPE related ranges of needed type (IPv4 or IPv6)
- without flag (existing behaviour):
       - IPv4 - always contains ranges ILB + NETLB  from "k8s.io/**cloud-provider-gcp**/providers/gce":
                           using func: L4LoadBalancerSrcRanges()
       - IPv6 - only LB_TYPE related ranges
  
"shared" FR (for ETP:Cluster) contains ILB + NETLB ranges

### DIFF
--- a/pkg/l4/healthchecks/healthchecksl4.go
+++ b/pkg/l4/healthchecks/healthchecksl4.go
@@ -31,12 +31,16 @@ import (
 	"k8s.io/cloud-provider/service/helpers"
 	"k8s.io/ingress-gce/pkg/composite"
 	"k8s.io/ingress-gce/pkg/firewalls"
+	"k8s.io/ingress-gce/pkg/flags"
+	"k8s.io/ingress-gce/pkg/l4/address"
 	"k8s.io/ingress-gce/pkg/l4/annotations"
 	l4utils "k8s.io/ingress-gce/pkg/l4/utils"
 	"k8s.io/ingress-gce/pkg/network"
 	"k8s.io/ingress-gce/pkg/utils"
 	"k8s.io/ingress-gce/pkg/utils/namer"
+	"k8s.io/ingress-gce/pkg/validation"
 	"k8s.io/klog/v2"
+	netutils "k8s.io/utils/net"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/l4/healthchecks/healthchecksl4.go
```diff
@@ -215,19 +218,32 @@ func (l4hc *l4HealthChecks) ensureHealthCheck(hcName string, svcName types.Names
...
 }
 
-// ensureIPv4Firewall rule for `svc`.
+// ensureFirewall creates a firewall rule for `svc` allowing health check probes.
 //
 // L4 ILB and L4 NetLB Services with ExternalTrafficPolicy=Cluster use the same firewall
 // rule at global scope.
-func (l4hc *l4HealthChecks) ensureIPv4Firewall(svc *corev1.Service, namer namer.L4ResourcesNamer, hcPort int32, isSharedHC bool, nodeNames []string, hcResult *EnsureHealthCheckResult, svcNetwork network.NetworkInfo, svcLogger klog.Logger) {
-	start := time.Now()
+func (l4hc *l4HealthChecks) ensureFirewall(l4Type utils.L4LBType, isIPv6 bool, isSharedHC bool, svc *corev1.Service, namer namer.L4ResourcesNamer, hcPort int32, nodeNames []string, hcResult *EnsureHealthCheckResult, svcNetwork network.NetworkInfo, svcLogger klog.Logger) {
```
mmamczur: I'd suggest using some type for the IP version. With a bool you then end up with calls like ``l4hc.ensureFirewall(l4Type, true, ...` and you wonder what that means

PR #3235, opened 2026-09-02
Mitigate informer cache staleness in L4 and NEG controllers using ConsistencyStore

08volt:
Controllers reconcile from informer caches, which lag behind the API server. Under load, a controller that patches a Service and then reconciles it again from a not-yet-updated cache can revert its own writes or loop forever.

This PR introduces a **ConsistencyStore** to mitigate that in the L4 controllers (ILB, NetLB, standalone NEG LB) and the NEG controller, gated behind `--enable-consistency-store` (default off):

- New `pkg/utils/consistency`: records the `ResourceVersion` returned by each write and reports staleness against the informer's `LastSyncResourceVersion`. Records are keyed by `(GroupResource, NamespacedName)`, so each resource is checked against its own informer. Keeps the newest RV per object UID, replaces records on object recreation, and fails open on non-numeric (opaque) ResourceVersions.
- Each controller's `sync` calls `EnsureReady` first and requeues (rate limited) with a `ConsistencyError` until the cache has observed the controller's own writes.
- Every Service write is recorded: status/annotation patches, finalizer add/remove, and RBS annotation removal.
- Records are cleared when a Service disappears (informer `DeleteFunc` handlers and the non-existent-service sync paths), so entries don't leak and a recreated Service isn't stalled by its predecessor.
- A single store is shared by all three controllers: they mutate the same Services through the same informer, so a Service migrating between controllers (e.g. NetLB → ILB) also waits for the other controller's writes.
- `pkg/utils/patch` and finalizer helpers now return `(*v1.Service, error)` so callers can observe the written RV; the L4 mutation helpers never return nil, returning the original Service on failure so error paths keep operating on a valid object.

**NEG controller (second commit):** SvcNeg CR writes are recorded by a thin wrapper around the SvcNeg client - so no write site can forget to record - and the reads that feed them are gated, in `ensureSvcNegCR` and in the status handler. The status patches are the reason: they are JSON merge patches built from a cached read, and merge patch replaces arrays wholesale, so one computed against a stale `Status` can drop a newer condition. Those patches also carry the observed `resourceVersion` as a precondition, so the API server rejects a stale patch with a Conflict even when the store cannot see the conflicting write (e.g. another instance's during a leader transition). Recording the create also replaces the hand-rolled poll that used to block in `ensureSvcNegCR` waiting for the CR to appear in the local store; with the store disabled (the default) the poll remains as a fallback, so default behavior is unchanged. Records are cleared when the informer observes a CR's deletion, which also covers deletions the controller did not perform. A stall is requeued without a warning event or an error metric.

**Out of scope:** Pod readiness gates. The condition written there comes from the NEG poller rather than from the cached pod, and the patch is a strategic merge keyed by condition type, so a stale read costs a redundant identical patch rather than a lost write - and stalling that path would delay pods becoming Ready. The multi-project NEG controller is wired in a third commit, building its store from the per-ProviderConfig filtered SvcNeg informer.

**Tests:** unit tests for the consistency package (incl. `-race`), finalizer and patch helper tests (incl. new `TestPatchServiceStatus` and returned-service assertions), and per-controller integration tests verifying syncs record writes, stall on a stale cache, resume when it catches up, and clear on deletion.


TortillaZHawaii:
Hi, I will review it on Monday as I would need to familiarize myself with how those watches work

pkg/utils/patch/patch_test.go
```diff
@@ -237,7 +237,7 @@ func TestPatchServiceLoadBalancerStatus(t *testing.T) {
 				t.Fatalf("Create(%s) = %v, want nil", svcKey, err)
 			}
 			expectSvc := tc.newMetaFunc(tc.svc)
-			err := PatchServiceLoadBalancerStatus(coreClient, tc.svc, expectSvc.Status.LoadBalancer)
+			_, err := PatchServiceLoadBalancerStatus(coreClient, tc.svc, expectSvc.Status.LoadBalancer)
```
TortillaZHawaii: Why there is no test coverage on the returned service?

pkg/utils/patch/patch_test.go
```diff
@@ -183,7 +183,7 @@ func TestPatchServiceObjectMetadata(t *testing.T) {
 				t.Fatalf("Create(%s) = %v, want nil", svcKey, err)
 			}
 			expectSvc := tc.newMetaFunc(tc.svc)
-			err := PatchServiceObjectMetadata(coreClient, tc.svc, expectSvc.ObjectMeta)
+			_, err := PatchServiceObjectMetadata(coreClient, tc.svc, expectSvc.ObjectMeta)
```
TortillaZHawaii: Why there is no test coverage on the returned service?

pkg/utils/patch/patch.go
```diff
@@ -105,9 +103,8 @@ func PatchProviderConfigObjectMetadata(client providerconfigclient.Interface, pc
 
 // PatchServiceStatus patches the given service's ServiceStatus
 // based on new service's status.
-func PatchServiceStatus(client coreclient.CoreV1Interface, svc *corev1.Service, newStatus corev1.ServiceStatus) error {
+func PatchServiceStatus(client coreclient.CoreV1Interface, svc *corev1.Service, newStatus corev1.ServiceStatus) (*corev1.Service, error) {
```
TortillaZHawaii, line 112: can you fill in the missing tests for this func?

cmd/glbc/main.go
```diff
@@ -19,6 +19,9 @@ package main
 import (
```
TortillaZHawaii, line 19: what about neg controller? it's mentioned in the description and title, yet it doesn't include consistency store.

pkg/utils/consistency/consistency.go
```diff
@@ -0,0 +1,122 @@
+package consistency
```
TortillaZHawaii, line 34: Why there are no tests?

TortillaZHawaii, line 34: Also why not import https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/util/consistency?

TortillaZHawaii (changes requested):
> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

1. Did you run it against the test suite? 
1. Can we make sure we have a mock in place to really test the consistency implementation, with test cases that ensure stale data hits during different parts of the code? This would be to make sure that ILB, NetLB, NEG, and StandaloneNEG controller work or give up as expected.
1. Can you flag gate it?
1. Does ILB, NetLB and StandaloneNEG controllers have separate services cache? Is the race condition possible - for example when moving from NetLB to ILB, when each of them uses separate consistency store?

pkg/utils/patch/patch_test.go
08volt: Done

pkg/utils/patch/patch.go
08volt: Done

pkg/utils/patch/patch_test.go
08volt: Fixed

cmd/glbc/main.go
08volt: Re-scoped to L4 controllers only, the NEG controller's only service write is the neg-status annotation, so I don't think we need the implementation there

pkg/utils/consistency/consistency.go
08volt: Fixed

08volt: Not really possible to import only that part, it would pull in the whole k8s.io/kubernetes module, which isn't consumable as a library. I think a small implementation here is better

08volt:
> > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.
> 
> 1. Did you run it against the test suite?
> 2. Can we make sure we have a mock in place to really test the consistency implementation, with test cases that ensure stale data hits during different parts of the code? This would be to make sure that ILB, NetLB, NEG, and StandaloneNEG controller work or give up as expected.
> 3. Can you flag gate it?
> 4. Does ILB, NetLB and StandaloneNEG controllers have separate services cache? Is the race condition possible - for example when moving from NetLB to ILB, when each of them uses separate consistency store?

1. Yes 
2. Done, added a fake informer RV getter plus per-controller integration tests (ILB, NetLB, standalone NEG)
3. Done, -enable-l4-consistency-store
4. All three controllers read the same ServiceInformer cache

cmd/glbc/main.go
TortillaZHawaii: cc @anyush, @midavadim

There are a couple of CRs that NEG controller updates, as well as readiness gate for Pods.

PR #3236, opened 2026-09-02
Enable client-go informer metrics for cache monitoring

08volt:
This PR enables `client-go` shared informer metrics in `ingress-gce`, allowing us to track queue backlogs, event latency, and cache sync state for core resources.

**Changes:**
- Generated a globally unique `cache.InformerName` ("ingress-gce") during the controller context initialization.
- Replaced the direct core informer creation (Pod, Node, Ingress, Service) with their `...WithOptions` equivalents, passing the initialized `InformerName`.
- Added the `k8s.io/component-base/metrics/prometheus/clientgo/fifo` blank import to `cmd/glbc/main.go` to register the internal Prometheus metrics providers.
- Updated `vendor/` to include the downloaded package dependencies.

*(Note: Custom CRD informers were skipped as their generated clients do not currently support `internalinterfaces.InformerOptions`)*.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	informerv1 "k8s.io/client-go/informers/core/v1"
 	discoveryinformer "k8s.io/client-go/informers/discovery/v1"
+	"k8s.io/client-go/informers/internalinterfaces"
 	informernetworking "k8s.io/client-go/informers/networking/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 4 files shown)


pkg/context/context.go
```diff
@@ -173,8 +174,23 @@ func NewControllerContext(
 ) (*ControllerContext, error) {
 	logger = logger.WithName("ControllerContext")
 
-	podInformer := informerv1.NewPodInformer(kubeClient, config.Namespace, config.ResyncPeriod, utils.NewNamespaceIndexer())
-	nodeInformer := informerv1.NewNodeInformer(kubeClient, config.ResyncPeriod, utils.NewNamespaceIndexer())
+	informerName, err := cache.NewInformerName("ingress-gce")
+	if err != nil {
+		logger.Error(err, "Failed to create InformerName")
+	}
+	optsNamespace := internalinterfaces.InformerOptions{
```
kl52752: why optsNamespace and optsCluster has the same values? I would expect that they will have at least different name. If that's not the case then we don't need 2 variables

08volt: Fixed

PR #3238, opened 2026-09-03
Fix dual-stack nodes being silently dropped from L4 GCE_VM_IP NEGs

08volt:
/kind bug

`GetNodePrimaryIP` and `GetNodeIPForNetwork` returned the first `NodeInternalIP` without checking its family. A dual-stack node listing IPv6 first was therefore classified "not IPv4" and **silently dropped from its GCE_VM_IP NEG**, losing an L4 load balancer backend, despite having a valid IPv4 address later in the list.

Node IPs are now resolved by family, not by list position. Endpoints stay single-stack: `--enable-ipv6-node-neg-endpoints` exists for IPv6-only clusters, so IPv6 is a fallback, never an addition. That flag now also requires `--enable-dual-stack-neg`, without which desired and observed endpoints disagree and churn forever. A zone that loses every node now logs an error.

Dual-stack NEG endpoints remain future work.

```release-note
Fixed dual-stack nodes whose IPv6 address is listed first being silently dropped from L4 GCE_VM_IP NEGs, removing them from the load balancer backends. The --enable-ipv6-node-neg-endpoints flag now requires --enable-dual-stack-neg.
```


pkg/neg/syncers/subsets.go
```diff
@@ -215,34 +278,27 @@ func getSubsetPerZone(nodesPerZone map[string][]*nodeWithSubnet, totalLimit int,
...
-			}
-
 			result[egi].Insert(newEndpoint)
 		}
+		if picked == 0 && len(skipped) > 0 {
+			// Every candidate node in this zone was dropped, so the zone's NEG ends up
+			// empty and the service loses all of its backends there.
+			for reason, count := range skipped {
+				logger.Error(nil, "All candidate nodes in zone were skipped", "zone", zone.Name, "skippedNodes", count, "reason", string(reason), "svcID", svcID)
+			}
```
anyush, line 306: Maybe instead of multiple logs it's better to have a single one, which will contain all reasons (along with skipped number per reason), as otherwise it might be misleading if during debug one (or more) logs were missed. E.g. I'd expect `"skippedNodes"` to be equal to all candidate nodes number within log `"All candidate nodes ... skipped"`.

pkg/neg/syncers/subsets.go
```diff
@@ -178,6 +176,71 @@ func newNodeWithSubnet(node *v1.Node, subnet string) *nodeWithSubnet {
...
+// case the node must be left out of the NEG.
+//
+// The endpoint is always single-stack, carrying one address and never both.
+// EnableIPv6NodeNEGEndpoints exists to support IPv6-only clusters, where a node
+// has no IPv4 address at all, so IPv6 is a fallback rather than an addition:
+// whenever a node has an IPv4 address that is the one used. Emitting both
+// families would turn every dual-stack node's existing endpoint into a
+// different endpoint, and NetworkEndpoint is a set key, so the syncer would
+// have to detach and re-attach every node before the NEG converged again.
+func nodeEndpoint(node *v1.Node, networkInfo *network.NetworkInfo, populateIPv6 bool, logger klog.Logger) (negtypes.NetworkEndpoint, nodeSkipReason, bool) {
```
anyush: I don't really like `populateIPv6` name as it only manages fallback, but sounds like something additional (setting `populateX` and then receiving something which doesn't have `X` is misleading)

pkg/neg/syncers/endpoints_calculator_test.go
```diff
@@ -1794,3 +1794,102 @@ func updateNodes(t *testing.T, nodeNames []string, nodeLabels map[string]map[str
...
+				Name:      testServiceName,
+				NegType:   negtypes.VmIpEndpointType,
+			}
+			ec := GetEndpointsCalculator(
+				nil, nodeIndexer, nil,
+				zoneGetter,
+				syncerKey,
+				negtypes.L4LocalMode,
+				klog.TODO(),
+				false, // enableDualStackNEG
```
anyush: Doesn't it create an impossible situation with 1 test case? `enableDualStackNEG == false` and `enableIPv6NodeNEGEndpoints == true`, which will fail in pkg/flags/flag.go

PR #3240, opened 2026-09-07
Fix security vulnerabilities for release-1.36 (b/556932907)

kl52752:
- Update Go to 1.25.14 (fixes CVE-2026-33818, CVE-2026-33814)
- Upgrade google.golang.org/grpc to v1.83.1 (fixes CVE-2026-84304)
- Regenerate composite types and update backend service zonal affinity handling
- Re-vendor dependencies


pkg/backends/backends.go
```diff
@@ -411,7 +411,7 @@ func (p *Pool) EnsureL4BackendService(params L4BackendServiceParams, beLogger kl
 	if params.EnableZonalAffinity {
 		beLogger.V(2).Info("EnsureL4BackendService: using Zonal Affinity", "spillover", ZonalAffinityEnabledSpillover, "spilloverRatio", DefaultZonalAffinitySpilloverRatio)
 		expectedBS.NetworkPassThroughLbTrafficPolicy = zonalAffinityEnabledTrafficPolicy()
-	} else {
+	} else if currentBS != nil && currentBS.NetworkPassThroughLbTrafficPolicy != nil {
```
TortillaZHawaii: /hold 

Why are the checks introduced? They are not present in master https://github.com/kubernetes/ingress-gce/blob/master/pkg/l4/backends/backends.go#L301

kl52752: removed, thanks for spotting that

PR #3244, opened 2026-09-11
Add IPv6 FR support for PSC SA

midavadim:
Add support for IPv6 ForwardingRules in GKE SA

Currently, the PSC controller only uses the IPv4 ForwardingRule during Service Attachment creation. 
This causes failures when customers attempt to use IPv6-only Kubernetes Services or IPv6-only NAT subnets.

Solution:
This PR implements logic for the PSC controller to check the Kubernetes Service's IP families and select the appropriate ForwardingRule (IPv4 or IPv6) based on the provided NAT Subnets.

Key Changes:
Adds support for IPv6-only Services using IPv6/Dual-stack subnets.
Adds support for Dual-stack Services using IPv6-only subnets.
Ensures dual-stack Services and Subnets default to the more universal IPv4 ForwardingRule.
Adds validation to prevent GCE errors by verifying that all provided NAT subnets support the selected IP stack type.


pkg/psc/controller.go
```diff
@@ -90,6 +90,130 @@ var (
...
+	case "IPV6_ONLY":
+		st.IPv6Only = append(st.IPv6Only, subnetObj.Name)
+	default:
+		// Covers "IPV4_ONLY", or missing value or any unknown stack types
+		st.IPv4Only = append(st.IPv4Only, subnetObj.Name)
+	}
+}
+
+// GetSubnetStackMap creates a new SubnetStackMap from a slice of subnets
+func GetSubnetStackMap(subnets []*ga.Subnetwork) *SubnetStackMap {
```
TortillaZHawaii: nit
```suggestion
// NewSubnetStackMap creates a new SubnetStackMap from a slice of subnets
func NewSubnetStackMap(subnets []*ga.Subnetwork) *SubnetStackMap {
```

pkg/psc/controller.go
```diff
@@ -90,6 +90,130 @@ var (
...
+func GetForwardingRuleStackMap(svc *v1.Service) (ForwardingRuleStackMap, error) {
+	var frNames ForwardingRuleStackMap
+	var ok bool
+
+	if _, ok = svc.Annotations[annotations.L3ForwardingRuleKey]; ok {
+		return frNames, fmt.Errorf("Service %s manages L3 ForwardingRule and cannot be used as PSC backend", svc.Name)
+	}
+	if _, ok := svc.Annotations[annotations.L3ForwardingRuleIPv6Key]; ok {
+		return frNames, fmt.Errorf("Service %s manages L3 ForwardingRule and cannot be used as PSC backend", svc.Name)
+	}
```
TortillaZHawaii, line 196: might be worth to return both `namespace/name` as using common name between different namespaces is quite common
```suggestion
	if _, ok = svc.Annotations[annotations.L3ForwardingRuleKey]; ok {
		return frNames, fmt.Errorf("service %s/%s manages L3 ForwardingRule and cannot be used as PSC backend", svc.Namespace, svc.Name)
	}
	if _, ok := svc.Annotations[annotations.L3ForwardingRuleIPv6Key]; ok {
		return frNames, fmt.Errorf("service %s/%s manages L3 ForwardingRule and cannot be used as PSC backend", svc.Namespace, svc.Name)
	}
```

pkg/psc/controller.go
```diff
@@ -324,21 +448,61 @@ func (c *Controller) processServiceAttachment(key string) error {
 		return err
 	}
 
-	var frURL string
-	frURL, err = c.getForwardingRule(namespace, updatedCR.Spec.ResourceRef.Name)
+	// Get the K8s Service referenced by the ServiceAttachment.
+	svcName := updatedCR.Spec.ResourceRef.Name
+	svcKey := fmt.Sprintf("%s/%s", namespace, svcName)
+	svc0bj, exists, err := c.serviceLister.GetByKey(svcKey)
```
TortillaZHawaii, line 454: is this 0 or O?

pkg/psc/controller.go
```diff
@@ -537,7 +678,8 @@ func (c *Controller) getForwardingRule(namespace, svcName string) (string, error
 	// Verify that the forwarding rule found has the IP expected in Service.Status
 	foundMatchingIP := false
 	for _, ing := range svc.Status.LoadBalancer.Ingress {
-		if ing.IP == fwdRule.IPAddress {
+		// IPv6 addr could contain /96 part, so split by "/" and take the first part to remove the CIDR notation
+		if ing.IP == strings.Split(fwdRule.IPAddress, "/")[0] {
```
TortillaZHawaii, line 682: This might be problematic if one of the two IPv6 is using canonical format and the other one is not. I think the L4 controllers are using the non canonical form (with lots of zeros), but should really use canonical one.

You can probably find in the repo functions that compare 2 addresses that work on IPv4 and both forms of IPv6.

pkg/psc/controller.go
```diff
@@ -550,24 +692,44 @@ func (c *Controller) getForwardingRule(namespace, svcName string) (string, error
...
-		subnet, err := c.cloud.Compute().Subnetworks().Get(context2.Background(), meta.RegionalKey(subnetName, c.cloud.Region()))
-		if err != nil {
-			return subnetURLs, fmt.Errorf("failed to find Subnetwork %s/%s: %w", c.cloud.Region(), subnetName, err)
+			// For shared vpc cases (could be cross-project), users must specify full resource path of the subnet.
+			var opts []cloud.Option
+			opts = append(opts, cloud.ForceProjectID(id.ProjectID))
+			subnet, err = c.cloud.Compute().Subnetworks().Get(context2.Background(), id.Key, opts...)
+			if err != nil {
+				if utils.IsNotFoundError(err) {
+					return nil, fmt.Errorf("failed to find Subnetwork %s/%s: %w", region, subnetName, err)
```
TortillaZHawaii, line 710: if the users must specify full resource path for xpn, shouldn't the project be also included in the err message?

pkg/psc/controller.go
```diff
@@ -550,24 +692,44 @@ func (c *Controller) getForwardingRule(namespace, svcName string) (string, error
...
+				// Proceed with an unvalidated subnet and let GCE SA API validate it during SA creation.
+				c.logger.V(2).Info("Unable to get Subnetwork from shared project, falling back to using subnet as is without validation", "subnetName", subnetName, "error", err)
+				subnet = &ga.Subnetwork{
+					Name:      id.Key.Name,
+					SelfLink:  subnetName,
+					StackType: "IPV4_IPV6", // use dual stack to fit any possible FR type
+				}
+			}
+		} else {
+
```
TortillaZHawaii, line 722: nit format empty line

pkg/psc/controller.go
```diff
@@ -550,24 +692,44 @@ func (c *Controller) getForwardingRule(namespace, svcName string) (string, error
...
+			var opts []cloud.Option
+			opts = append(opts, cloud.ForceProjectID(id.ProjectID))
+			subnet, err = c.cloud.Compute().Subnetworks().Get(context2.Background(), id.Key, opts...)
+			if err != nil {
+				if utils.IsNotFoundError(err) {
+					return nil, fmt.Errorf("failed to find Subnetwork %s/%s: %w", region, subnetName, err)
+				}
+				// The controller may lack permission to read shared VPC subnets (but Subnet still can be "used" for PSC)
+				// Proceed with an unvalidated subnet and let GCE SA API validate it during SA creation.
+				c.logger.V(2).Info("Unable to get Subnetwork from shared project, falling back to using subnet as is without validation", "subnetName", subnetName, "error", err)
```
TortillaZHawaii, line 714: Is 403 returned in this case?

TortillaZHawaii (changes requested):
I'll check the tests tommorow

pkg/psc/controller_test.go
```diff
@@ -1993,3 +2009,265 @@ func TestCheckUnsyncedFields(t *testing.T) {
...
+				if got != tc.wantFRName {
+					t.Errorf("GetCompatibleForwardingRule() = %s, want %s", got, tc.wantFRName)
+				}
+			}
+		})
+	}
+}
+
+func TestGetForwardingRuleStackMap(t *testing.T) {
+	flags.F.EnablePSCIPv6Support = true
```
TortillaZHawaii: Global flag should go back to previous state after the test finishes.

pkg/psc/controller_test.go
```diff
@@ -1993,3 +2009,265 @@ func TestCheckUnsyncedFields(t *testing.T) {
...
 }
+
+func TestSubnetStackMap(t *testing.T) {
+	testcases := []struct {
+		desc     string
+		subnets  []*ga.Subnetwork
+		wantIPv4 bool
+		wantIPv6 bool
+	}{
+		{
```
TortillaZHawaii, line 2042: Can we add a singular ipv4 only without StackType edge case? I want to make sure that the behavior is not regressed in the future

pkg/psc/controller_test.go
```diff
@@ -1494,40 +1510,40 @@ func TestFilterError(t *testing.T) {
...
-			createSubnet:   true,
+			createName:     "my-subnet",
 			expectErr:      false,
 			expectedSubnet: "https://www.googleapis.com/compute/v1/projects/test-project/regions/us-central1/subnetworks/my-subnet",
 		},
 		{
 			desc:           "subnet is the resource url path",
 			subnet:         "projects/test-project/regions/us-central1/subnetworks/subnet-1",
-			createSubnet:   false,
+			createName:     "subnet-1",
```
TortillaZHawaii, line 1532: why do we need to create them now? is this possibly a breaking change?

pkg/psc/controller_test.go
```diff
@@ -138,6 +139,17 @@ func TestServiceAttachmentCreation(t *testing.T) {
...
 			readOnlyMode:         false,
 		},
+		{
+			desc:                 "forwarding rule has ipv6 address with cidr suffix",
+			annotationKey:        annotations.TCPForwardingRuleKey,
+			svcExists:            true,
+			fwdRuleExists:        true,
+			connectionPreference: "ACCEPT_AUTOMATIC",
+			resourceRef:          validRef,
+			customFwdRuleIPAddr:  "1.2.3.4/96", // svc creates with frIPAddr (1.2.3.4), fwd rule gets created with customFwdRuleIPAddr
```
TortillaZHawaii, line 149: Let's add a test case with actual IPv6 addresses with both canonical and non-canonical form.

TortillaZHawaii (changes requested):

pkg/psc/controller.go
midavadim: good spoted

pkg/psc/controller.go
midavadim: update log message, subnetName already contains full subnet path

pkg/psc/controller.go
midavadim: yes, during manual testing I saw 403

pkg/psc/controller_test.go
midavadim: Now, we have to check for stacktype for each specified Subnet.

So yes, now we need to create each Subnet. The only exception if for subnets from "shared" vpc, which are specified with full subnet url

pkg/psc/controller.go
```diff
@@ -90,6 +91,130 @@ var (
...
+}
+
+// GetSubnetStackMap creates a new SubnetStackMap from a slice of subnets
+func GetSubnetStackMap(subnets []*ga.Subnetwork) *SubnetStackMap {
+	st := &SubnetStackMap{}
+	for _, subnetObj := range subnets {
+		st.Add(subnetObj)
+	}
+	return st
+}
```
TortillaZHawaii, line 132: Just an observation, you can disregard it as the benefit of changing it is probably not worth it.

The purpose of this struct is to just evaluate `AllSupportIPv6` and `AllSupportIPv6`. I don't think you need to store each subnet name and have O(n) storage, and you could get away with just using booleans `HasIPv4Only`, ... and skip `append`s. Then `st.AllSupportIPv6()` is  `!st.HasIPv4Only`.

But the controller isn't really hard on memory itself so the current version is fine as well.

pkg/psc/controller.go
```diff
@@ -324,21 +449,61 @@ func (c *Controller) processServiceAttachment(key string) error {
...
+	svcName := updatedCR.Spec.ResourceRef.Name
+	svcKey := fmt.Sprintf("%s/%s", namespace, svcName)
+	svcObj, exists, err := c.serviceLister.GetByKey(svcKey)
+	if err != nil {
+		return fmt.Errorf("errored getting service %s/%s: %w", namespace, svcName, err)
+	}
+	if !exists {
+		return fmt.Errorf("failed to get Service %s/%s: %w", namespace, svcName, ServiceNotFoundError)
+	}
+	svc := svcObj.(*v1.Service)
```
TortillaZHawaii, line 462: I see that this has been copied from old code and I don't think there is any chance that this will not be a service, but if for whatever reason it's not this will panic. 

Generally the safer way to catch against it is to use `svc, ok := svcObj.(*v1.Service)` https://go.dev/doc/effective_go#interface_conversions

pkg/psc/controller_test.go
```diff
@@ -1993,3 +2031,376 @@ func TestCheckUnsyncedFields(t *testing.T) {
...
+		})
+	}
+	flags.F.EnablePSCIPv6FRSupport = false
+}
+
+func TestGetForwardingRule(t *testing.T) {
+	testcases := []struct {
+		desc      string
+		fwdRuleIP string
+		svcIPs    []string
```
TortillaZHawaii, line 2311: can we add a test case with 2 IPs - IPv4 and IPv6 to match the second one?

pkg/psc/controller_test.go
TortillaZHawaii: ahh yeah makes sense, thanks for clarifying

TortillaZHawaii (approved):
non blocking comments, can be changed in future PR if necessary

/lgtm

pkg/psc/controller.go
midavadim: names are added to use in log and event message, to indicate exact subnet names which are incorrect.

PR #3246, opened 2026-09-14
Validate the content of the ```networking.gke.io/load-balancer-ip-addresses``` annotation

aymenkhelifi-1:
### What this PR does / why we need it:

If a user puts an invalid value (such as an IP literal) into the `networking.gke.io/load-balancer-ip-addresses` annotation, the controller previously passed it directly to GCE and failed with:
```text
"Requeuing due to error" err="EnsureInternalLoadBalancer error: address.IPv4ToUse returned error: googleapi: Error 400: Invalid value for field 'address': '1.2.3.4'. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}', invalid"
```

This PR validates the annotation value rules before calling GCE:

- Emits a clear error indicating the annotation value is invalid.
- Marks it as a UserError to prevent requeue loops and protect SLO metrics.

### Does this PR introduce a user-facing change?
```release-note
Fix: Validate `networking.gke.io/load-balancer-ip-addresses` annotation to reject invalid address names as a UserError instead of failing with a GCE API 400 error.

### CODE REVIEW

File: pkg/l4/annotations/l4_static_address.go [hunk 1]
```diff
@@ -20,6 +22,11 @@ const (
 	IPv6Version                    IPVersion = "IPV6"
 )
 
+var (
+	addressNameRegex       = regexp.MustCompile(`^[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?$`)
+	ErrInvalidStaticIPName = errors.New("invalid static IP address name")
```
TortillaZHawaii, line 27: Can you include a suggestion on why the address is invalid and what should be used to make it a valid field? 

It should be okay to include `See https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#spd-static-ip` in the message.

TortillaZHawaii (changes requested):

pkg/l4/annotations/l4_static_address.go
aymenkhelifi-1: Done! The error message has been updated with the explanation and the documentation link as requested.

PR #3251, opened 2026-09-23
Add custom org constraints to user error

TortillaZHawaii:


pkg/l4/utils/errors.go
```diff
@@ -105,7 +105,14 @@ func IsUnsupportedNetworkTierError(err error) bool {
 
 // IsConstraintViolationError checks if the error is a constraint violation error returned by GCP.
 func IsConstraintViolationError(err error) bool {
-	return gceutils.IsHTTPErrorCode(err, http.StatusPreconditionFailed) && strings.Contains(err.Error(), "Constraint") && strings.Contains(err.Error(), "violated")
+	if !gceutils.IsHTTPErrorCode(err, http.StatusPreconditionFailed) {
+		return false
+	}
+
+	reOrgPolicy := regexp.MustCompile(`Operation denied by org policy: \[customConstraints/custom\..+\] : .*`)
```
08volt, line 112: These get recompiled on every call. Please move them to package-level vars, like `networkTierErrorRegexp` above

08volt, line 112: Requiring `custom`. and the trailing `] : <description>` could be too strict as we would miss custom constraints with no description, and built-in constraints that use the same phrasing. Since we already gate on 412, strings.Contains(errMsg, "Operation denied by org policy") would be enough

TortillaZHawaii: great call, reverted to substrings

TortillaZHawaii: updated, thank you!

